PR #16385, opened 2023-08-09
feat: add ignore as recommended by React docs

Lantianyou:


apps/docs/pages/guides/getting-started/tutorials/with-react.mdx
```diff
@@ -155,7 +156,13 @@ export default function Account({ session }) {
       setLoading(false)
     }
 
-    getProfile()
+    if (!ignore) {
+      getProfile()
+    }
```
gregnr, line 161: I think `ignore` would always be `false` here, since this always runs before the cleanup function. Let's move this logic to inside the `getProfile()` function, after getting the results back from `supabase-js`.

gregnr (commented):
Thanks @Lantianyou! Just a small change.

charislam (approved):

PR #16347, opened 2023-08-08
Chore: Convert all independent PATCH and DELETE requests to RQ mutations

joshenlim:
Looks into converting all independent PATCH and DELETE requests on the dashboard to RQ mutations (Except StorageExplorerStore)

PR #16438, opened 2023-08-11
Convert all independent GET requests to RQ

joshenlim:


studio/components/interfaces/BillingV2/Invoices.tsx
```diff
@@ -1,92 +1,57 @@
...
       })
     }
   }
 
   return (
     <div className="container my-4 max-w-4xl space-y-1">
-      <Loading active={loading}>
+      {isLoading && <GenericSkeletonLoader />}
+
+      {isError && <AlertError error={error} subject="Failed to retrieve project invoices" />}
```
ivasilov, line 52: We should use the new alerts here, right?

studio/components/interfaces/BillingV2/Invoices.tsx
```diff
@@ -103,7 +68,7 @@ const Invoices = ({ projectRef }: InvoicesProps) => {
               <Table.tr>
                 <Table.td colSpan={6} className="p-3 py-12 text-center">
                   <p className="text-scale-1000">
-                    {loading ? 'Checking for invoices' : 'No invoices for this project yet'}
+                    {isLoading ? 'Checking for invoices' : 'No invoices for this project yet'}
```
ivasilov, line 71: `isLoading` and `isSuccess` can't both be true, right?

studio/components/interfaces/Organization/BillingSettings/BillingAddress/BillingAddress.tsx
```diff
@@ -54,15 +61,15 @@ const BillingAddress = ({ loading, address, onAddressUpdated }: BillingAddressPr
   const onSubmit = async (values: any, { setSubmitting }: any) => {
     try {
       setSubmitting(true)
-      const updatedCustomer = await patch(`${API_URL}/organizations/${orgSlug}/customer`, {
+      const updatedCustomer = await patch(`${API_URL}/organizations/${slug}/customer`, {
```
ivasilov, line 64: This refactor would've been unnecessary once we have the this PATCH call in a mutation, right? It will invalidate the cache key and the GET query will happen on its own.

studio/components/interfaces/BillingV2/Invoices.tsx
joshenlim: Ahh yup! AlertError is already using the new Alert component - its just a standardized component on how we want to show UI Alerts across the dashboard - an error message and a quick button to direct users to the support form the title prefilled and error message added to the message body

studio/components/interfaces/BillingV2/Invoices.tsx
joshenlim: nope - the states from RQ are all mutually exclusive (isLoading, isSuccess, isError for the one's that we commonly use, and all the other states too)

studio/components/interfaces/Organization/BillingSettings/BillingAddress/BillingAddress.tsx
joshenlim: yup! the PATCH call mutation change is added in https://github.com/supabase/supabase/pull/16347/files

studio/components/interfaces/BillingV2/Invoices.tsx
joshenlim: ohh sorry Alaister just pointed out that isLoading will never be true here since its within the isSuccess conditional - will update! 🙏

studio/data/invoices/project-invoices-query.ts
```diff
@@ -0,0 +1,51 @@
...
+export type ProjectInvoicesError = ResponseError
+
+export const useProjectInvoicesQuery = <TData = ProjectInvoicesData>(
+  { projectRef, offset, limit }: ProjectInvoicesVariables,
+  {
+    enabled = true,
+    ...options
+  }: UseQueryOptions<ProjectInvoicesData, ProjectInvoicesError, TData> = {}
+) =>
+  useQuery<ProjectInvoicesData, ProjectInvoicesError, TData>(
```
alaister, line 40: For the future we could look at making this `useInfiniteQuery` since it takes a limit/offset.

studio/data/invoices/project-invoices-query.ts
```diff
@@ -0,0 +1,51 @@
+import { useQuery, UseQueryOptions } from '@tanstack/react-query'
+import { get } from 'data/fetchers'
+import { ResponseError } from 'types'
+import { invoicesKeys } from './keys'
+
+export type ProjectInvoicesVariables = {
+  projectRef?: string
+  offset: number
+  limit: number
```
alaister, line 9: Seems strange to me that these are required – can't we provide defaults? e.g. `offset = 0` and `limit = 50`?

studio/data/invoices/project-invoices-query.ts
```diff
@@ -0,0 +1,51 @@
...
+) =>
+  useQuery<ProjectInvoicesData, ProjectInvoicesError, TData>(
+    invoicesKeys.projectInvoices(projectRef),
+    ({ signal }) => getProjectInvoices({ projectRef, offset, limit }, signal),
+    {
+      enabled:
+        enabled &&
+        typeof projectRef !== 'undefined' &&
+        typeof limit !== 'undefined' &&
+        typeof offset !== 'undefined',
```
alaister, line 48: If they weren't required these wouldn't be needed either

studio/data/invoices/invoices-query.ts
```diff
@@ -0,0 +1,51 @@
+import { useQuery, UseQueryOptions } from '@tanstack/react-query'
+import { get } from 'data/fetchers'
+import { ResponseError } from 'types'
+import { invoicesKeys } from './keys'
+
+export type InvoicesVariables = {
+  customerId?: string
+  slug?: string
+  offset: number
+  limit: number
```
alaister, line 10: Same comment here about these being required

studio/data/fetchers.ts
```diff
@@ -89,9 +89,11 @@ export const del: typeof _del = async (url, init) => {
 export const head: typeof _head = async (url, init) => {
   const headers = await constructHeaders(init?.headers)
 
+  // @ts-ignore
```
alaister, line 92: Why is this needed here?

alaister (changes requested):
Looking good! The pagination on invoices doesn't seem to be working, though. I think changing it to an infinite query might be the way to go

studio/data/fetchers.ts
joshenlim: ahh i'm only running into a ts error when i added the parseAs: 'text' - wondering if you might know whats the cause as i'm not really sure :/ 
image

studio/data/invoices/project-invoices-query.ts
joshenlim: ahh yeap makes sense

studio/data/fetchers.ts
alaister: I think it's because it doesn't know the endpoint, and so it doesn't know if it supports `parseAs` yet. If you pass in `parseAt` at the call site instead of in this file, it shouldn't cause a type error.

alaister (approved):
LGTM!

PR #16451, opened 2023-08-11
Typesafe forms using Formik

ivasilov:
This is a proof of concept on how to use Formik to have more typesafe forms. It's better than nothing but other libs (`react-hook-form` for example) have better ergonomics.


MildTomato:
Are we just going with react-hook-form ?

awalias:
closing as stale feel free to reopen as needed

PR #16459, opened 2023-08-11
Fix: Regex for destructive operation detection in SQL editor

gregnr:
Improve the regex for destructive operation detection to reduce false positives.

eg. when you perform a select on a table with a column called `deleted` or `deleted_timestamp` - this shouldn't flag the warning:
image

### DIFF
--- a/studio/components/interfaces/SQLEditor/SQLEditor.constants.ts
+++ b/studio/components/interfaces/SQLEditor/SQLEditor.constants.ts
@@ -1256,4 +1256,4 @@ export const sqlAiDisclaimerComment = stripIndent`
 
 export const untitledSnippetTitle = 'Untitled query'
 
-export const destructiveSqlRegex = [/(drop|delete|truncate) /i]
+export const destructiveSqlRegex = [/(drop|delete|truncate)\s/i]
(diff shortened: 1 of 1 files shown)


studio/components/interfaces/SQLEditor/SQLEditor.constants.ts
```diff
@@ -1256,4 +1256,4 @@ export const sqlAiDisclaimerComment = stripIndent`
 
 export const untitledSnippetTitle = 'Untitled query'
 
-export const destructiveSqlRegex = [/drop/i, /delete/i, /truncate/i]
+export const destructiveSqlRegex = [/(drop|delete|truncate) /i]
```
gregnr, line 1259: Specifically - adding the space ` ` at the end should reduce a lot of false positives.

kevcodez (approved):

studio/components/interfaces/SQLEditor/SQLEditor.constants.ts
gregnr: Updated to use `\s` instead to account for line breaks, tabs, etc.

PR #16476, opened 2023-08-12
chore: redirect project pausing button

pamelachia:

Explain to free plan users that they can avoid project pausing by upgrading to pro at the project paused page. Pro plan users button remains the same. 

Screenshot 2023-08-12 at 2 53 14 PM

### DIFF
--- a/studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
+++ b/studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
@@ -13,6 +13,7 @@ import { post } from 'lib/common/fetch'
 import { API_URL, PROJECT_STATUS } from 'lib/constants'
 import { Button, IconPauseCircle, Modal } from 'ui'
 import { useProjectContext } from './ProjectContext'
+import { useProjectSubscriptionV2Query } from 'data/subscriptions/project-subscription-v2-query'
 
 export interface ProjectPausedStateProps {
   product?: string
@@ -25,6 +26,8 @@ const ProjectPausedState = ({ product }: ProjectPausedStateProps) => {
   const selectedOrganization = useSelectedOrganization()
   const { project } = useProjectContext()
   const orgSlug = selectedOrganization?.slug
+  const { data: subscription } = useProjectSubscriptionV2Query({ projectRef: ref })
+  const isFreePlan = subscription?.plan.id === 'free'
 
   const { data: membersExceededLimit } = useFreeProjectLimitCheckQuery({ slug: orgSlug })
   const hasMembersExceedingFreeTierLimit = (membersExceededLimit || []).length > 0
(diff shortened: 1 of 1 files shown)


studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
```diff
@@ -128,11 +136,21 @@ const ProjectPausedState = ({ product }: ProjectPausedStateProps) => {
...
                   )}
                 </Tooltip.Root>
-                <Link href={`/project/${ref}/settings/general`}>
-                  <a>
-                    <Button type="default">View project settings</Button>
-                  </a>
-                </Link>
+                {isFreePlan ? (
+                  <Link
+                    href={`/project/${ref}/settings/billing/subscription?panel=subscriptionPlan`}
```
kevcodez, line 141: Let's have the correct URL depending on org/project level billing. We've set up some redirects, but it's still a longer route to go through the project first and then be redirected to the org. So a better UX to get the URL right

Org-Billing URL:

`/org/${slug}/billing?panel=subscriptionPlan`

To check if the org uses org level billing

```ts
 const billedViaOrg = Boolean(currentOrg?.subscription_id)
```

studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
```diff
@@ -25,6 +26,8 @@ const ProjectPausedState = ({ product }: ProjectPausedStateProps) => {
   const selectedOrganization = useSelectedOrganization()
   const { project } = useProjectContext()
   const orgSlug = selectedOrganization?.slug
+  const { data: subscription } = useProjectSubscriptionV2Query({ projectRef: ref })
+  const isFreePlan = subscription?.plan.id === 'free'
```
kevcodez, line 30:
```suggestion
  const isFreePlan = subscription?.plan?.id === 'free'
```

studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
```diff
@@ -128,11 +136,21 @@ const ProjectPausedState = ({ product }: ProjectPausedStateProps) => {
...
-                  <a>
-                    <Button type="default">View project settings</Button>
-                  </a>
-                </Link>
+                {isFreePlan ? (
+                  <Link
+                    href={`/project/${ref}/settings/billing/subscription?panel=subscriptionPlan`}
+                  >
+                    <a>
+                      <Button type="default">Change Plan</Button>
```
kevcodez, line 144: No hard feelings, just food for thought to think about different CTAs for this one

"Upgrade Plan"
"Upgrade to Pro"

Can keep it as it is though

kevcodez (approved):
Just a few nits, not blocking

studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
pamelachia: Good idea will change

PR #16512, opened 2023-08-14
Allow for manual re-arrangement of columns when editing

### LINKED ISSUE #16509: Allow manually sorting columns from side panel in create table and update table modes
We should be able to reorder columns during the edit table and create table stage in the side panel. Currently, we can't change the order on existing tables. Ideally, we'd be able to sort columns from both.

CleanShot 2023-08-14 at 10 54 37@2x

faizanraso:

This PR allows users to manually change column order when editing an existing table. This PR closes https://github.com/supabase/supabase/issues/16509

Currently users are only able to manually edit the order of columns when creating a new table - as discussed in Issue https://github.com/supabase/supabase/issues/16509

## What is the new behavior?

Users can now change the column order when editing a table
Screenshot 2023-08-14 at 10 43 56 AM


joshenlim:
hey there @faizanraso! thank you so much for taking on the issue :) unfortunately the changes in your PR will not work as updating the order of SQL columns requires more than a UI change.

With reference to this document [here](https://wiki.postgresql.org/wiki/Alter_column_position#:~:text=Postgres%20currently%20defines%20column%20order,you%20reach%20the%20desired%20layout), there's technically no straight forward way to reorder the columns in a postgres table.

What I'd suggest to tackle this issue, however, is to leverage on local storage - which is what the dashboard currently does when you drag and re-order the columns via the table editor. You can refer to these 2 pieces of code for reference [[1]](https://github.com/supabase/supabase/blob/790395dddf6dc0cd41e6a4c40d3d97b13976e67f/studio/components/grid/components/grid/ColumnHeader.tsx#L109), [[2]](https://github.com/supabase/supabase/blob/790395dddf6dc0cd41e6a4c40d3d97b13976e67f/studio/components/grid/store/reducers/column.ts#L45)

Do keep in mind as well that this issue will also involve ensuring that the TableEditor side panel for an existing table shows the columns in the right order as well based on the UI sorting of the columns

e.g if the columns of the table are originally id, name and inserted_at, and i reordered via the UI to be id, inserted_at and name, the TableEditor side panel will need to show the columns in the order top to bottom as id, inserted_at name to be consistent

faizanraso:
Hi @joshenlim, thanks so much for the explanation! I'll give it a go and see if I can get it working.

faizanraso:
Hi @joshenlim, I've taken the feedback you've provided and tried to implement it in my solution. I'm leveraging the existing method to re-arrange columns on the dashboard and applying it to the side panel editor. I believe I am heading in the right direction, but the issue I'm having is using the `dispatch` function. I was wondering if you have time and could check over the changes I made and advise me on how to go about resolving this issue.

PR #16514, opened 2023-08-14
feat: allow manual sorting cols on update table modal 

### LINKED ISSUE #16509: Allow manually sorting columns from side panel in create table and update table modes

SuhelMakkad:

fixes: #16509 
this will allow manual sorting columns on the update table modal 

Screenshot 2023-08-14 at 8 58 25 PM

## Additional context

Showing drag handles, originally hidden for new table

### DIFF
--- a/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/Column.tsx
+++ b/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/Column.tsx
@@ -42,7 +42,6 @@ import { ColumnField } from '../SidePanelEditor.types'
 interface ColumnProps {
   column: ColumnField
   enumTypes: PostgresType[]
-  isNewRecord: boolean
   hasImportContent: boolean
   dragHandleProps?: any
   onEditRelation: (column: any) => void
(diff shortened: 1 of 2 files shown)


joshenlim:
hey there @SuhelMakkad! thank you so much for taking on the issue :) unfortunately the changes in your PR will not work as updating the order of SQL columns requires more than a UI change.

With reference to this document [here](https://wiki.postgresql.org/wiki/Alter_column_position#:~:text=Postgres%20currently%20defines%20column%20order,you%20reach%20the%20desired%20layout), there's technically no straight forward way to reorder the columns in a postgres table.

What I'd suggest to tackle this issue, however, is to leverage on local storage - which is what the dashboard currently does when you drag and re-order the columns via the table editor. You can refer to these 2 pieces of code for reference [[1]](https://github.com/supabase/supabase/blob/790395dddf6dc0cd41e6a4c40d3d97b13976e67f/studio/components/grid/components/grid/ColumnHeader.tsx#L109), [[2]](https://github.com/supabase/supabase/blob/790395dddf6dc0cd41e6a4c40d3d97b13976e67f/studio/components/grid/store/reducers/column.ts#L45)

SuhelMakkad:
Hi @joshenlim, thanks for explaining this issue. I will take a shot at it.

PR #16516, opened 2023-08-14
implements expression filter above table grids

Filsommer:

This is a new feature that I missed quite a lot from other SQL editor tools like DBeaver or Oracle Developer: Being able to filter rows based on an input field visible from the table view.
**I've added a demo to showcase this feature below.**

Currently, I have to click LOTS of times to add multiple filters. 
Let's take an example where I want to filter my products by price (above 100$) and stock (below 50 units).
To achieve this currently in the Table Editor, I would have to click on at least 5 dropdowns `(Filter -> price -> greater than -> 100 -> Add filter -> stock -> less than -> 50 -> Apply filter`.

## What is the new behavior?

With this change, users would be able to write the query "price > 100 AND stock < 50" and apply is directly above the rows, without ever opening a single dropdown.
[Here is a screen recording](https://imgur.com/a/RbE47qj) showing the new feature in action.
Technically, this method simply mimics the process of adding multiple filters in the Filter dropdown, i.e. pushes the filters into the URL directly - aka. there are no hidden tricks behind this.

## Additional context

⚠️ _Important_: 
There are still a few things that need to be improved (contributors are welcome!):
- The query input needs a robust validation mechanism. Some examples of possible improvements:
    - "price > 100" is valid, but "price>100" is invalid, as I separate the operators and values by whitespaces. 
    - Right now, it's only possible to chain multiple filters with AND, not OR.
    - The queries should. be more consistent with SQL language. E.g. The SQL expression "name in ('John','Marie')" is currently written as "name in John,Marie", which is shorter, but not standard. 
- Styles can be improved, e.g. less rounded input? Feedback is appreciated.
- Ideally, users would be able to hide (toggle) this input. Could be hidden by default and be toggled inside the Filter dropdown.


saltcod:
Quite clever @Filsommer!   Will review with the team asap and report back. Thanks so much for this!

studio/components/grid/components/header/filter/ExpressionFilter.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { isEqual } from 'lodash'
+import * as Tooltip from '@radix-ui/react-tooltip'
+
+function convertToSingleSpace(input: string) {
+  // Replace multiple spaces with a single space
+  return input.replace(/\s+/g, ' ')
+}
+
+function isValidSearch(searchText: string) {
+  return searchText
```
Filsommer, line 15: // TODO create a more robust search validation. See "Additional Context" section on https://github.com/supabase/supabase/pull/16516

alaister:
Hey @Filsommer,

Thanks for all the work you've put in here!

Unfortunately, I'm going to go ahead and close this PR. We definitely want to improve this UI, but we're going to have to go through a full design process with the whole team. If we reference any of your work here, we'll be sure to give you the credit!

PR #16519, opened 2023-08-15
Modify gotrue.json doc to map to auth.user right paths

theodufort:

Doc update

Simply linking to wrong database column

## What is the new behavior?

Linked to updated database column

### DIFF
--- a/spec/enrichments/tsdoc_v2/gotrue.json
+++ b/spec/enrichments/tsdoc_v2/gotrue.json
@@ -3140,7 +3140,7 @@
 														"isOptional": true
 													},
 													"comment": {
-														"shortText": "A custom data object to store additional metadata about the user. This maps to the `auth.users.user_metadata` column."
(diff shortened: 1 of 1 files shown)


gregnr:
Thanks @theodufort!

cc @supabase/auth to confirm this change - if so @theodufort do you mind also opening a PR on supabase/gotrue-js which is the original source of these docs?

https://github.com/supabase/gotrue-js/blob/master/src/GoTrueAdminApi.ts

kangmingtay:
hi @theodufort, thanks for taking the time to make this PR, unfortunately, this file is generated from the tsdocs based on the [gotrue-js](https://github.com/supabase/gotrue-js) library so we would need to make the change there first and then generate it here

alaister:
Hey @theodufort ,

Thanks for the PR, and sorry for the delay!

Unfortunately, this has got quite stale, so I'm going to go ahead and close it. Please feel free to make the requested changes to gotrue-js and then open up a new PR.

Thanks again!

PR #16527, opened 2023-08-15
hotfix: storage list abort toast error

alaister:
In browsers that are not Chrome an error toast is displayed when trying to list a buckets contents:

Screenshot 2023-08-15 at 5 01 18 pm


darora (approved):

pcnc:
Looks like there's an identical code block further down the file - could we also handle it the same way?

https://github.com/supabase/supabase/blob/56b3a6e125e2f8ee2823e72dbbd511e9696ab63e/studio/localStores/storageExplorer/StorageExplorerStore.js#L1166-L1171

PR #16540, opened 2023-08-15
fix: project plan when switching project in support form

kevcodez:
When switching the project in the support form, it wouldn't refresh the plan the project is on, leading to possibly showing the wrong plan, i.e. your Pro plan would show "Your project is on the free plan".


mansueli (approved):

mansueli:
LGTM

PR #16547, opened 2023-08-15
fix: remove unnecessary env var checks in self-hosted studio

Ziinc:
This PR removes some checks in the self-hosted studio, which are no longer required as logs are always enabled now, and the logs explorer is compatible with postgres backend (via translation)


filipecabaco (approved):

filipecabaco:
Please remove it from docker-compose.yml. Maybe in the context of this PR 🙏

joshenlim (approved):

PR #16555, opened 2023-08-16
Convert independent POST requests to RQ P1

joshenlim:
- Refactored AddNewPaymentMethod modal, and remove repeated AddNewPaymentMethod modal component
- Convert all independent requests in org join page to RQ
- Convert project restore request to RQ and remove all calls to restore in useEffects
  - There’s already an appropriate UI to handle paused state, and its better that users explicitly click restore, rather than we try to restore in the BG


studio/components/interfaces/BillingV2/AddNewPaymentMethodModal/AddNewPaymentMethodModal.tsx
```diff
@@ -52,7 +66,7 @@ const AddNewPaymentMethodModal = ({
           return
         }
 
-        await setupIntent(token ?? undefined)
+        await initSetupIntent(token ?? undefined)
```
alaister, line 69: Can you define `initSetupIntent` _before_ this `useEffect` and then call it via `useLatest`? Or put it as an inline function inside the `useEffect`

alaister (changes requested):
Looking good!

One potential issue: when pausing a project there doesn't seem to be any sort of polling, so unless the user tabs back and forth, they'll be stuck on the polling screen forever.

studio/components/interfaces/BillingV2/AddNewPaymentMethodModal/AddNewPaymentMethodModal.tsx
joshenlim: actually yeah i'll shift it inside the useEffect since its only used in that block

joshenlim: RE pausing polling - hmm super odd i remember there used to be some polling logic but i actually can't find it now too, not part of the changes in the PR as well. i'll reinstate it

joshenlim: @alaister just pushed - added long polling logic with RQ in ProjectPausingState

studio/components/layouts/ProjectLayout/PausingState.tsx
```diff
@@ -1,11 +1,42 @@
...
+    { projectRef: ref },
+    {
+      enabled: startPolling,
+      onSuccess: async (res) => {
+        if ([PROJECT_STATUS.GOING_DOWN, PROJECT_STATUS.PAUSING].includes(res.status)) {
+          await timeout(2000)
+          refetch()
+        } else if (PROJECT_STATUS.INACTIVE) {
+          await invalidateProjectsQuery(queryClient)
+        }
```
alaister, line 31: I'm still getting stuck on the pausing state.

react-query has built-in support for refetching automatically btw: look for `refetchInterval` in the docs.

It can also be a function which returns a number (interval ms) or `false` to disable the refetching, so you can move this logic into that :)

joshenlim: gotcha - hmm is the request not getting triggered when you hit pause? the only problem i can think of is that the async process to toggle a project to pausing takes longer than the initial 4 seconds before the first call to project's status

alaister: Yeah the pause seems to work, but then I don't see any requests for the polling

joshenlim: oh wait lol sorry! the else if logic is wrong 🤦

studio/components/layouts/ProjectLayout/PausingState.tsx
```diff
@@ -1,11 +1,41 @@
...
+  useProjectStatusQuery(
+    { projectRef: ref },
+    {
+      enabled: startPolling,
+      refetchInterval: (res) => {
+        return res?.status === PROJECT_STATUS.INACTIVE ? false : 2000
+      },
+      onSuccess: async (res) => {
+        if (res.status === PROJECT_STATUS.INACTIVE) {
+          await invalidateProjectsQuery(queryClient)
```
alaister, line 29: All the requests seem to be working now, but the pausing state remains... Should this be `setProjectStatus` or invalidate project **detail** query instead of `invalidateProjectsQuery`?

joshenlim: ohhh yes you're right!

joshenlim: okay just pushed again 🙏  - I also update BuildingState to do the same too, was wondering why locally the page doesnt automatically update after the project is active_healthy

alaister (approved):
LGTM!
Although please make sure the "return to project" button works without having to switch tabs – not sure if it needs the same invalidation treatment, or it's just my project taking a while 😅

joshenlim:
> Although please make sure the "return to project" button works without having to switch tabs – not sure if it needs the same invalidation treatment

ahh good catch! i think its not invalidating actually, but rather just retrieving the project details and setting it in meta store
i'll add in the invalidation here cause ProjectLayout is dependent on the state from RQ, but it'll all clean up once we've moved away from mobx

PR #16557, opened 2023-08-16
Convert independent POST requests to RQ P2

joshenlim:
- Refactored PauseProjectButton for RQ
- Refactored RestartProjectButton for RQ
- Refactored BackupItem for RQ
- Refactored PITRSelection for RQ


alaister (approved):
LGTM!

PR #16566, opened 2023-08-16
fix: Add joins into counts sql generation

Ziinc:
This PR fixes a bug where the counts sql was not generated with the correct cross joins when a filter was selected.

Before:
Screenshot 2023-08-17 at 3 20 39 AM

After:
Screenshot 2023-08-17 at 3 20 56 AM

### DIFF
--- a/studio/components/interfaces/Settings/Logs/Logs.utils.ts
+++ b/studio/components/interfaces/Settings/Logs/Logs.utils.ts
@@ -203,7 +203,7 @@ const genCrossJoinUnnests = (table: LogsTableName) => {
   cross join unnest(m.request) as request`
 
     default:
-      return ""
+      return ''
   }
 }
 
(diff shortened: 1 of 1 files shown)


filipecabaco (approved):

studio/components/interfaces/Settings/Logs/Logs.utils.ts
```diff
@@ -195,8 +228,9 @@ export const maybeShowUpgradePrompt = (from: string | null | undefined, planId?:
 }
 
 export const genCountQuery = (table: LogsTableName, filters: Filters): string => {
-  const where = _genWhereStatement(table, filters)
-  return `SELECT count(*) as count FROM ${table} ${where}`
+  const where = genWhereStatement(table, filters)
+  const joins = genCrossJoinUnnests(table)
+  return `SELECT count(*) as count FROM ${table} ${joins} ${where}`
```
chasers, line 233: @Ziinc can we replace this with it's own endpoint now? Or do you want to do that later?

joshenlim (approved):

studio/components/interfaces/Settings/Logs/Logs.utils.ts
Ziinc: Will merge this first, as need more thought for the counts query issue. 

Quick fix would be to copy the entire logs.all endpoint query and maintain two sets of endpoint configs, but it increases chance of mistakes and higher maintenance burden.

Been thinking of how to handle that with minimal endpoint copying to reduce maintenance.

Crux of the issue is that we need to support the sql generated filters, so it needs to be "sandboxable" and accept the sql query param (and we cant use the "sandboxed" endpoints feature to solve this as the sandboxed sql would be fixed. Sandboxed endpoints was meant for handling report queries, though the reports would also need to handle dynamic filters so a little on the fence on moving forward with sandboxed endpoints) 

Got a few ideas for the above:
- "cache-only" endpoints that have identical queries but with different cache settings. 
- endpoint query composition, on the fence for this one, not very convinced that it will be intuitive enough because query logic will be split across multiple pages, making maintenance harder
- sandboxed endpoints with fixed filters

Will flesh themout more in Notion. All of these aren't mutually exclusive, we could use a combination of these.

PR #16604, opened 2023-08-18
feat: enable `navigatorLock` completely, track random logouts

hf:
Enables `navigatorLock` completely, which can be disabled per-browser by setting the `supabase.dashboard.auth.navigatorLock.disabled` flag to `true`.

Yet another prefix is added to track random logouts using the sign-in clicks counter.


J0 (approved):

hf:
Let's queue this for Thursday's deploy. I'll merge it tomorrow.

PR #16610, opened 2023-08-18
Install `@supabase/auth-ui-shared` before using it

Silverethical:
## docs update
Based on the example provided, @supabase/auth-ui-shared package is also needed to be installed. However, it is not mentioned in the previous section.


gregnr (approved):
Good catch, thanks @Silverethical!

PR #16652, opened 2023-08-21
Fixed unable to switch to light mode without login

AbhijeetBiswalCodes:

Unable to Switch to Light Mode without Login.

## What is the new behavior?

Now You Can Switch to Light Mode without Login.


fsansalvadore:
Thanks for the suggestion @AbhijeetBiswalCodes but the behaviour is intended.
We'll keep the home page locked to dark mode until we remove the Launch Week 8 branding in the hero.

I'll close this PR for now, but thanks for the effort!

Screenshot 2023-08-21 at 12 05 08

PR #16662, opened 2023-08-21
Trim column headers to avoid saving names with whitespaces

tonibardina:
It can be dangerous, or at least, annoying to save column headers with whitespaces by mistake


tonibardina:
> @tonibardina is this everywhere a column name is saved? What happens if someone has one saved and they edit it, losing whitespace as a result?

I don't see any reason for someone to add white spaces at the end or beginning of the name. But yes this will trim the name on every update.

alaister:
Hey @tonibardina,

Thanks for your PR, and sorry for the delay here.

It can be annoying to accidentally save whitespace into a column name. However, since this is a supported feature of Postgres, we want to also support it in the dashboard.

We'd love to accept a PR that warns a user that there's whitespace at the end of their column name, for example.

Thanks again 🙌🏻

Issue #16506, opened 2023-08-14
Bug: Entering Double Quotation Mark in the Input Field of the Command Menu Causes Application Error


ratnaraj7:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application.
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com).

## Describe the bug

When entering a double quotation mark (") in the input field of the command menu, it throws an error: "Application error: a client-side exception has occurred (see the browser console for more information)."

## To Reproduce

Steps to reproduce the behavior on the Supabase documentation website:

1. Go to the [Supabase documentation](https://supabase.com/docs).
2. Open the command menu.
3. In the input field of command menu, type a double quotation mark (").
4. Observe the error message.

## Expected behavior

The input should accept double quotation marks without throwing an error.

## Screenshots

https://github.com/supabase/supabase/assets/102022870/cafc1c10-96fb-4a6b-b35f-93cd2b1bedd8

## System information

- OS: Arch
- Browser: Chrome
- Version of supabase-js: Not applicable (issue observed on Supabase documentation website)
- Version of Node.js: Not applicable (issue observed on Supabase documentation website)

nwaliaez:
Hello everyone,

After investigating, I believe this might be a limitation within the library itself (cmdk-supabase), rather than an issue with this repo. As a quick workaround, I've managed to restrict users from inputting double quotes, which seems to prevent the error. However, I'm aware that this isn't the ideal approach.

Since I'm relatively new to open source, I'm unsure about the best way to proceed. Should I attempt to make changes to the cmdk-supabase library to address this issue directly? if anyone has suggestions for a better workaround, I'd greatly appreciate the advice.

Thank you in advance for your help and guidance!

saltcod:
hey @nwaliaez would be ideal to commit to this repo instead of cmdk-supabase if possible. We had to fork that repo because of a compatibility issue, but intend to eventually deprecate it and move back to using the cmdk package.

PR #16670, opened 2023-08-21
Fix double quote issue in docs search

ghost:
Summary of Pull Request
This pull request addresses a bug related to entering double quotation marks in the input field of the command menu, which currently causes an application error. The issue can be found [here](https://github.com/supabase/supabase/issues/16506).

Current Behavior
Currently, when users input double quotation marks in the command menu's input field, it triggers a bug that results in an application error.

Proposed New Behavior
The proposed change restricts the use of double quotation marks in the input field for the documentation search. This fix is temporary and will be removed once we transition back to the original command menu (from cmdk to cmdk-supabase).


packages/ui/src/components/Command/CommandMenu.tsx
```diff
@@ -67,6 +67,12 @@ const CommandMenu = ({ projectRef }: CommandMenuProps) => {
     useCommandMenu()
   const showCommandInput = !currentPage || !CHAT_ROUTES.includes(currentPage)
 
+  // This function has been added to prevent the use of double quotes in the search docs input due to an issue with the cmdk-supabase module. This function can be removed when we transition to using cmdk.
+  const handleInputChange = (value: string) => {
+    const newValue = value.replace(/"/g, '') // Remove double quotes
+    setSearch(newValue)
+  }
+
```
gregnr, line 75: Thanks @nwaliaez! Can you help me understand what specifically triggers the error? Is it something in `CommandDialog` that crashes when `value` contains a double quote character?

nwaliaez: Hello @gregnr, based on my interpretation, the problem lies within the `CommandInput`. It appears to be causing issues with the `data-value` attribute when a double quote is entered. This leads to a `SyntaxError: The execution of 'querySelector' on 'Element' has failed because '[cmdk-item=""][data-value="search the docs: ""]' is an invalid selector.`

gregnr: Got it thank you. I think this is a good interim solution until it's fixed in the lib. Thanks!

gregnr (approved):

packages/ui/src/components/Command/CommandMenu.tsx
nwaliaez: Thank you @gregnr

PR #16681, opened 2023-08-22
Passing JWT_SECRET and JWT_EXPIRY to postgrest APP_SETTINGS

Lan-Hekary:

This PR allow the user to get the JWT Secret and expiry inside postgres using `current_setting('app.settings.jwt_secret')` and `current_setting('app.settings.jwt_expiry')`

Supabase managed instances can get the JWT_SECRET through executing the query `current_setting('app.settings.jwt_secret')` , But Self-hosted instances can't do that because the these values are not passed in the application setting for postgrest.

## What is the new behavior?

Passing the Values as environment variables to postgrest services as mentioned [here](https://postgrest.org/en/stable/references/configuration.html#app-settings) align the behavior of self-hosted services with supabase managed instances.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -168,6 +168,8 @@ services:
       PGRST_DB_ANON_ROLE: anon
       PGRST_JWT_SECRET: ${JWT_SECRET}
       PGRST_DB_USE_LEGACY_GUCS: "false"
+      PGRST_APP_SETTINGS_JWT_SECRET: ${JWT_SECRET}
+      PGRST_APP_SETTINGS_JWT_EXP: ${JWT_EXPIRY}
 
   realtime:
     container_name: realtime-dev.supabase-realtime


soedirgo (approved):

sweatybridge (changes requested):
Could you confirm these env vars are working as expected for you? I'm still getting an error from my local self-hosted setup.

```bash
$ psql -h localhost -U postgres -p 5432
Password for user postgres:
psql (15.3, server 15.1 (Ubuntu 15.1-1.pgdg20.04+1))
Type "help" for help.

postgres=> select current_setting('app.settings.jwt_secret');
ERROR:  unrecognized configuration parameter "app.settings.jwt_secret"
```

steve-chavez:
> postgres=> select current_setting('app.settings.jwt_secret');
ERROR:  unrecognized configuration parameter "app.settings.jwt_secret"

@sweatybridge The above would only work if you're inside a [PostgREST transaction](https://postgrest.org/en/stable/references/transactions.html). Like when executing a function/RPC.

Lan-Hekary:
Hello @sweatybridge,
as @steve-chavez  said, its working only inside a PostggREST Transaction. 
meaning that it works only if called from the REST Endpoint.

sweatybridge (approved):

Lan-Hekary:
It satisfies my use case for a degree, but I am thinking that it's better to be exposed through out the DB like with the Supabase hosted version.

So, Should we add an initialisation script to the DB instead ?
I see you already did that in the CLI, if it's okay let me know And I will work on it, or if you prefer to do it its okay as well.

soedirgo:
I'll merge this for now, but feel free to add a db init script in another PR if you need to 👍

PR #16687, opened 2023-08-22
Fix loading state

joshenlim:


studio/components/to-be-cleaned/Dropdown/OrganizationDropdown.js
```diff
@@ -3,9 +3,8 @@ import { useRouter } from 'next/router'
 import { useFlag } from 'hooks'
 import { Button, Dropdown, IconPlus } from 'ui'
 
-const OrganizationDropdown = ({ organizations }) => {
+const OrganizationDropdown = ({ organizations = [] }) => {
```
alaister, line 6: Can you use `EMPTY_ARR` here instead of `[]` in case we use `organizations` in a dependency array somewhere.

`import { EMPTY_ARR } from 'lib/void'`

alaister (approved):

PR #16692, opened 2023-08-22
feat: update supabase-js on www, use `navigatorLock`

hf:
Website's GoTrue must be setup the same way it's setup for studio, as multiple windows / tabs open need to synchronize over the local storage medium.


hf:
Not necessary at this time.

Issue #16671, opened 2023-08-21
Twitter logo may be need to update as new 'X' logo


EngHasanSabbir:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

On July 23, 2023, the social media platform twitter has been updated their logo in new abstract 'X' shape after that most of online site who used Twitter traditional bird logo they have changed their logo to new 'X' shape but superbase did not update it yet. I think that is important to aware them about this. 

 ## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to 'https://supabase.com/'
2. In home page  Scroll down to 'Join the community'
3. You can see the old bird logo with People Avatar
4. You can find the similar issue in footer icon.

## Expected behavior
![image](https://github.com/supabase/supabase/assets/60088050/bd0e4e19-866f-482d-823c-fbea49cd8889)

## Screenshots
![image](https://github.com/supabase/supabase/assets/60088050/a904917d-bfdb-4817-a1b0-053c8903b0f8)
![image](https://github.com/supabase/supabase/assets/60088050/b34c5ee2-aaf0-4a71-a3ab-18a94bd547f4)

## Additional context

It's also important to regularly update the site frontend as well as backend.

rishabhbizzle:
hey, can you assign this issue to me I would like to work on it

rishabhbizzle:
The project is using react-feather package for icons but the new X logo is not yet available in this package so what should we do now?

AbhijeetBiswalCodes:
#16688

Aakash-Rajbhar:
hello can you assign this issue to me, I want to contribute in it passionately

fsansalvadore:
Fixed by:
- #16688
- #16768
- #16805

PR #16701, opened 2023-08-22
Twitter logo may be need to update as new 'X' logo

po1so9:

Bug fix, fixed the twitter logo (X) at the homepage

https://github.com/supabase/supabase/issues/16671

## What is the new behavior?
Showing new Twitter logo( X ) at the home page.
Supabase _ The Open Source Firebase Alternative - Brave 22-08-2023 18_22_24


hieuwu (commented):
It would be nice if you find other places that are using the old Twitter icon and replace them as well. Also double check for light theme

po1so9:
> It would be nice if you find other places that are using the old Twitter icon and replace them as well. Also double check for light theme

Sure, I'll replace it if I see one also it is working as intended with the light theme as well.

saltcod:
Hey there — thanks for this!

Looks like there's this one as well, if you could please have a look!

CleanShot 2023-08-22 at 12 01 49@2x

EngHasanSabbir:
In dark mode, I think this logo is looking more gorgeous

image

po1so9:
> In dark mode, I think this logo is looking more gorgeous
> 
> image

yeah, I also noticed that but the logo is very small and the X lines are very thin which makes it hard to see also you can see that the screenshot you added has a black background which makes it easier to add the logo and there are not a lot of X logos out there may be in future if I find a better one I'll update it. I'm still going through the whole file structure to understand the project better.

PR #16704, opened 2023-08-22
Docs: RLS and Roles

kiwicopple:
- Database section: 
  - new [RLS page](https://docs-git-doc-rlsrecommendations-supabase.vercel.app/docs/guides/database/postgres/row-level-security). Adds a lot of Gary's performance considerations
  - new [Roles page](https://docs-git-doc-rlsrecommendations-supabase.vercel.app/docs/guides/database/postgres/roles)
  - new [Configuration page](https://docs-git-doc-rlsrecommendations-supabase.vercel.app/docs/guides/database/postgres/configuration) . consolidated the "configuration" into a single page
- Auth section:
  - Updated [RLS page](https://docs-git-doc-rlsrecommendations-supabase.vercel.app/docs/guides/auth/row-level-security). Moves a lot of RLS content out, to the Database RLS doc above
  - removes "Deprecated section" - it is now in our Discussions: https://github.com/orgs/supabase/discussions/16703


fsansalvadore (approved):

apps/docs/pages/guides/database/postgres/roles.mdx
```diff
@@ -0,0 +1,134 @@
+import Layout from '~/layouts/DefaultGuideLayout'
+
+export const meta = {
+  id: 'postgres-roles',
+  title: 'Postgres Roles',
+  description: 'Managing access to your Postgres database and configuring permissions.',
+  subtitle: 'Managing access to your Postgres database and configuring permissions.',
+}
+
+PostgreSQL manages database access permissions using the concept of roles. Generally you wouldn't use these roles for your own application - they are mostly for configuring _system access_ to your database. If you want to confgure _application access_, then you should use Row Level Security (although in some cases, you can mix RLS and Postgres Roles to build an extremely powerful Auth system).
```
kangmingtay, line 10: Should we omit this if we're not gonna mention how?
```suggestion
PostgreSQL manages database access permissions using the concept of roles. Generally you wouldn't use these roles for your own application - they are mostly for configuring _system access_ to your database. If you want to confgure _application access_, then you should use Row Level Security.
```

apps/docs/pages/guides/database/postgres/triggers.mdx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+In Postgres, a trigger executes a set of actions automatically on table events such as INSERTs, UPDATEs, DELETEs, or TRUNCATE operations.
+
+## Creating a Trigger
+
+Creating triggers involve 2 parts:
+
+1. A [Function](docs/guides/database/functions) which will be executed (called the Trigger Function)
+2. The actualTtrigger object, with parameters around when the trigger should be run.
```
kangmingtay, line 17:
```suggestion
2. The actual trigger object, with parameters around when the trigger should be run.
```

kangmingtay (approved):

apps/docs/pages/guides/auth/row-level-security.mdx
```diff
@@ -149,332 +210,81 @@ create policy "Team members can update team details if they belong to the team."
...
-on profiles for select
-to authenticated, anon
-using (
-  true
+create policy "Restrict updates."
+on profiles
+as restrictive
+to authenticated
+for update using (
+  using (auth.jwt()->>'aal' = 'aal2')
```
kangmingtay, line 284:
```suggestion
create policy "Restrict updates."
on profiles
as restrictive
for update
to authenticated using (
  auth.jwt()->>'aal' = 'aal2'
```

apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -0,0 +1,378 @@
...
+
+You can do:
+
+```sql
+create policy "rls_test_select" on test_table
+to authenticated
+using ( (select auth.uid()) = user_id );
+```
+
+This method works well for JWT functions like `auth.uid()` and `auth.jwt()` as well as `security definer` Functions. Wrapping the function causes an initPlan to be run by the Postgres optimizer, which allows it to "cache" the results per-statement, rather than calling the function on each row.
```
olirice, line 217: cool!

apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -0,0 +1,378 @@
...
+```
+
+{/* prettier-ignore */}
+<Admonition type="caution">
+Security-definer functions should never be created in a schema in the "Exposed schemas" inside your [API settings](https://supabase.com/dashboard/project/_/settings/api)`. 
+</Admonition>
+
+### Minimize joins
+
+You can often rewrite your Policies to avoid joins between the source and the target table. Instead, try to organize your Policy to fetch all the relevant data from the target table into an array or set, then you can use an `IN` or `ANY` operation in your filter.
```
olirice, line 308:
```suggestion
You can often rewrite your Policies to avoid joins between the source and the target table. Instead, try to organize your policy to fetch all the relevant data from the target table into an array or set, then you can use an `IN` or `ANY` operation in your filter.
```

apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -0,0 +1,378 @@
...
+
+For example, this is an example of a slow Policy which joins the source `test_table` to the target `team_user`:
```
olirice, line 310:
```suggestion
For example, this is an example of a slow policy which joins the source `test_table` to the target `team_user`:
```

apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -0,0 +1,378 @@
...
+    where user_id = auth.uid() -- no join
+  )
+);
+```
+
+In this case you can also consider [using a `security definer` function](#use-security-definer-functions) to bypass RLS on the join table:
+
+{/* prettier-ignore */}
+<Admonition type="note">
+If the `in` list gets to be over 10K items, then extra analysis and a different approach may be needed.
```
olirice, line 342: 10k seems high here. Just wanted to confirm that's right

olirice (approved):

apps/docs/pages/guides/auth/auth-mfa.mdx
```diff
@@ -240,18 +218,14 @@ export function EnrollMFA({
...
-For example, call `supabase.auth.mfa.unenroll({factorId: "d30fd651-184e-4748-a928-0a4b9be1d429"})` to unenroll a factor with ID `d30fd651-184e-4748-a928-0a4b9be1d429`.
+When a user unenrolls a factor, call `supabase.auth.mfa.unenroll()` with the ID of the factor. For example, call `supabase.auth.mfa.unenroll({factorId: "d30fd651-184e-4748-a928-0a4b9be1d429"})` to unenroll a factor with ID `d30fd651-184e-4748-a928-0a4b9be1d429`.
 
 
 #### Example: React
 
-Below is an example that creates a new `UnenrollMFA` component that illustrates
-the important pieces of the MFA enrollment flow. Note that users can only unenroll a factor after completing the enrollment flow and obtaining
-an `aal2` JWT claim.Here are some points of note:
+Below is an example that creates a new `UnenrollMFA` component that illustrates the important pieces of the MFA enrollment flow. Note that users can only unenroll a factor after completing the enrollment flow and obtaining an `aal2` JWT claim.Here are some points of note:
```
imor, line 228:
```suggestion
Below is an example that creates a new `UnenrollMFA` component that illustrates the important pieces of the MFA enrollment flow. Note that users can only unenroll a factor after completing the enrollment flow and obtaining an `aal2` JWT claim. Here are some points of note:
```

apps/docs/pages/guides/auth/auth-mfa.mdx
```diff
@@ -468,12 +431,9 @@ function AuthMFA() {
 
 ### Enforce rules for MFA logins
 
-Adding MFA to your app's UI does not in-and-of-itself offer a higher level of
-security to your users. You also need to enforce the MFA rules in your
-application's database, APIs and server-side rendering.
+Adding MFA to your app's UI does not in-and-of-itself offer a higher level of security to your users. You also need to enforce the MFA rules in your application's database, APIs and server-side rendering.
```
imor, line 434:
```suggestion
Adding MFA to your app's UI does not in-and-of-itself offer a higher level of security to your users. You also need to enforce the MFA rules in your application's database, APIs, and server-side rendering.
```

apps/docs/pages/guides/auth/row-level-security.mdx
```diff
@@ -39,47 +40,123 @@ from todos
...
 
-Here are some examples to show you the power of PostgreSQL's RLS.
+<Admonition type="note">
+Initializing a client with a Service Key will not override RLS if a user token is set. If a user is signed in with a client, Supabase will adhere to the RLS policy of the user.
+</Admonition>
 
-### Allow read access
+### Always enable RLS on public tables
+
+You should _always_ enable RLS on tables created in a public schemas. This is considered "default safe". Unfortunately this is not enabled by default on Postgres, so you will need to keep this in mind - especially if you are using the SQL Editor or database migrations. RLS is already enabled by default if you create a table using the Table Editor. If you want to allow public access to a table, just add a Policy with `true`:
```
imor, line 114:
```suggestion
You should _always_ enable RLS on tables created in a public schema. This is considered "default safe". Unfortunately this is not enabled by default on Postgres, so you will need to keep this in mind - especially if you are using the SQL Editor or database migrations. RLS is already enabled by default if you create a table using the Table Editor. If you want to allow public access to a table, just add a Policy with `true`:
```

apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -0,0 +1,378 @@
...
+```sql
+select *
+from todos
+where auth.uid() = todos.user_id;
+-- Policy is implicitly added.
+```
+
+## Enabling Row Level Security
+
+You can enable RLS for any table using the `enable row level security` command:
```
imor, line 41:
```suggestion
You can enable RLS for any table using the `enable row level security` clause:
```

PR #16729, opened 2023-08-23
[WIP]: Improve join the community section

Isaiah-Hamilton:

This pull request hopefully improves the join the community section.

Before the Tweet Cards are on a sweeper in a straight line making it look not very full.

CleanShot 2023-08-22 at 23 25 44@2x

## What is the new behavior?

Now the Tweet Cards layout is on an uneven grid making it look there is a lot more tweet cards straight from the start.

CleanShot 2023-08-22 at 23 25 24@2x

## Additional context

This is a POC pull request so the code could use some work, and some ideas still need working on.


fsansalvadore (commented):
Oh, I was just noticing that it suddenly requires a lot of scrolling on mobile...
Wdys if "show more" adds 15 at a time until all are revealed? @Isaiah-Hamilton

Isaiah-Hamilton:
> Oh, I was just noticing that it suddenly requires a lot of scrolling on mobile...
Wdys if "show more" adds 15 at a time until all are revealed? @Isaiah-Hamilton

This was my original idea but was having some trouble on how to implementing it. Any suggestions?

apps/www/components/Sections/TwitterSocialProof.tsx
```diff
@@ -1,31 +1,35 @@
...
+      ...Tweets.slice(prevTweets.length, prevTweets.length + 10),
+    ])
+
+    if (tweets.length >= Tweets.length) {
+      setShowButton(false)
+    }
+
+    const container = document.querySelector('.tweet-container')
+    // @ts-ignore
+    container.style.height = `${container.clientHeight + 400}px`
```
fsansalvadore, line 27: Not sure we need this. The container should adjust based on the content.

apps/www/components/Sections/TwitterSocialProof.tsx
```diff
@@ -1,31 +1,35 @@
...
+
+    const button = document.querySelector('.show-button')
+    // @ts-ignore
+    button.style.marginBottom = `${parseInt(button.style.marginBottom) + 20}px`
```
fsansalvadore, line 31: This also shouldn't be needed.

apps/www/components/Sections/TwitterSocialProof.tsx
```diff
@@ -53,70 +57,38 @@ function TwitterSocialProof() {
...
-              <div ref={nextRef} className="p mr-4 cursor-pointer">
-                <IconArrowRight />
-              </div>
+      <div className="lg:-mx-10 xl:-mx-18 mt-6">
+        <div
+          className={`tweet-container columns-1 sm:columns-2 lg:columns-3 xl:columns-4 gap-4 overflow-hidden relative h-[1000px]`}
+        >
+          {showButton && (
+            <div
+              className={`absolute bottom-0 left-0 z-10 w-full h-[25%] bg-gradient-to-t from-[#1c1c1c] via-[#1c1c1c]`}
```
fsansalvadore, line 66: You can simply add a margin-bottom: 20px here or use the `bottom` property, something like bottom-4 instead of bottom-0?

fsansalvadore (changes requested):

fsansalvadore:
@Isaiah-Hamilton btw can you move the `TweetCard.tsx` component to the global `ui` package so we can also use it in docs?

You can add it in this folder: `/packages/ui/src/components`
And add the export here: `/packages/ui/index.tsx`

Thanks!

apps/www/components/Sections/TwitterSocialProof.tsx
Isaiah-Hamilton: Sadly the container doesn't adjust on its own. What happens when your remove the code above:

https://github.com/supabase/supabase/assets/70828596/327a8939-e2fe-491e-b3fb-04cf6daa7081

apps/www/components/Sections/TwitterSocialProof.tsx
Isaiah-Hamilton: Same thing as above

apps/www/components/Sections/TwitterSocialProof.tsx
Isaiah-Hamilton: Yeah, it just depending how much of the gradient you want to hide the tweet cards.

apps/www/components/Sections/TwitterSocialProof.tsx
fsansalvadore: @Isaiah-Hamilton took a shot at this, the problem was actually that the container had a fixed height. Now it actually grows with the content. 👍

fsansalvadore (approved):

PR #16809, opened 2023-08-25
Store monaco assets locally and load from there

joshenlim:


alaister (changes requested):
Getting stuck in the loading state on `/dashboard`, which I think is due to these new files not having the same base path. Does it work if you move `public/monaco-editor` to `public/dashboard/monaco-editor`?

Screenshot 2023-08-28 at 3 21 14 pm
Screenshot 2023-08-28 at 3 21 37 pm

joshenlim:
@alaister great catch! thanks man - had to add basePath to wherever we're loading the monaco assets due to the /dashboard suffix

alaister (approved):
LGTM!

PR #16810, opened 2023-08-25
Restore enterprise pricing col

fsansalvadore:

Restore Enterprise pricing column.

## Before

Screenshot 2023-08-25 at 14 50 15

## After

Screenshot 2023-08-25 at 14 50 02

### DIFF
--- a/apps/www/pages/pricing/index.tsx
+++ b/apps/www/pages/pricing/index.tsx
@@ -104,7 +104,7 @@ export default function IndexPage() {
   }) => {
     return (
       <div className="mt-8 px-4 mobile-header">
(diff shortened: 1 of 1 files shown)


kevcodez:
Screenshot 2023-08-25 at 18 24 09

Looks a bit too compact on this resolution IMO (shortly before it goes into two columns).

kevcodez:
Screenshot 2023-08-24 at 15 52 56

Quite a lot of padding for the Enterprise row, if it's just a single column

kevcodez (approved):
Just two nits, feel free to tackle or leave them be

fsansalvadore:
@kevcodez yeah the "compact" version is actually intentional because otherwise the text would be much more squashed and it looked worse.

For the Enterprise spacing, I agree it could be better but it's preferable to keep the bullet points section aligned with the other columns.

PR #16811, opened 2023-08-25
feat: add test otp configuration UI

hf:
Adds a UI to configure test OTPs (predetermined phone number and OTP combinations) which are useful when developing mobile apps that need to be submitted to the Apple and Google stores for review.

Preview of the UI design:

image


kdewald:
Awesome feature, thanks for getting it in!

joshenlim:
@hf should we clear `SMS_TEST_OTP_VALID_UNTIL` if the user clears the `SMS_TEST_OTP` field? at the moment because the former is hidden when we clear `SMS_TEST_OTP` so we run into an error `Both SMS_TEST_OTP and SMS_TEST_OTP_VALID_UNTIL need to be provided` but the only way to clear `SMS_TEST_OTP_VALID_UNTIL` is by putting something in `SMS_TEST_OTP` then clearing `SMS_TEST_OTP_VALID_UNTIL`

studio/stores/authConfig/schema/AuthProviders/AuthProvidersFormValidation.tsx
```diff
@@ -243,6 +243,21 @@ const PROVIDER_PHONE = {
...
+      type: 'string',
+      title: 'Test Phone Numbers and OTPs',
+      description:
+        'Register phone number and OTP combinations for testing as a comma separated list of <phone number>=<otp> pairs. Example: `18005550123=789012`',
+    },
+    SMS_TEST_OTP_VALID_UNTIL: {
+      type: 'datetime',
+      title: 'Test OTPs Valid Until',
+      description:
+        "Test phone number and OTP combinations won't be active past this date and time.",
```
joshenlim, line 256: can we also include the timezone that we're showing in the input field/date picker?

`Test phone number and OTP combinations won't be active past this date and time. The date and time shown in the field is formatted to your local time zone (+XXXX)`

can use dayjs to generate the TZ - something like `dayjs().format('ZZ')`

kdewald:
Glad this is making progress! :D

joshenlim:
@hf found an issue here if i clear the test phone numbers and otps field
image

studio/components/interfaces/Auth/AuthProvidersForm/ProviderForm.tsx
```diff
@@ -55,23 +55,17 @@ const ProviderForm = ({ config, provider }: ProviderFormProps) => {
...
-    // When the key is a 'double negative' key, we must reverse the boolean before the payload can be sent
+    // Format payload for the following checks:
+    // 1. Convert all empty string values to null
+    // 2. When the key is a 'double negative' key, we must reverse the boolean before the payload can be sent
     Object.keys(values).map((x: string) => {
-      if (doubleNegativeKeys.includes(x)) {
-        payload[x] = !values[x]
-      }
+      if (doubleNegativeKeys.includes(x)) payload[x] = !values[x]
+      if (payload[x] === '') payload[x] = null
```
joshenlim, line 63: @hf just FYI im adding this as the fix from the previous comment i mentioned
noticed that get auth config returns empty properties as null so figured this should be okay

joshenlim (approved):

azlekov:
Hey, I'm following this feature and I saw I have access to the menus in the supabase studio. I decided to try to configure it:

image

Unfortunately it breaks the GoTrue service and cannot start:

image

Is this a know issue?

kdewald:
I'm seeing the same problem on my end as well. Is there anything we can do to help?

hf:
Let us check what the issue is...

hf:
Sorry everyone not sure how this fell through, the `key=value` setting in GoTrue should be `key:value`. I'm making a fix immediately. The UI still will use `=`.

azlekov:
Now setting test numbers is okay, but when try to verify a test phone number by entering test otp, it gives me: `Token has expired or is invalid`. I remember that locally this also is not working as the supabase client is using older version of gotrue. Is this the same case for production?

PR #16813, opened 2023-08-25
feat(studio): Add MFA support for the dashboard

ivasilov:

This PR enables MFA support for the dashboard. The user can now setup 2 TOTP apps and use them to login to the dashboard.

No MFA is present currently on `/dashboard`.

## What is the new behavior?

The user will have to provide TOTP codes upon login if he setup his account to have MFA.


studio/components/interfaces/Account/AddNewFactorModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+import { useMfaUnenrollMutation } from 'data/profile/mfa-unenroll-mutation'
+import { useStore } from 'hooks'
+import Image from 'next/image'
+import { Dispatch, SetStateAction, useEffect, useState } from 'react'
+import { Input, Modal } from 'ui'
+
+const AddNewFactorModal = ({ onClose }: { onClose: () => void }) => {
+  const [verificationComplete, setVerificationComplete] = useState(false)
+  // Generate a name with a number between 0 and 1000
+  const [name, setName] = useState(`App ${Math.floor(Math.random() * 1000)}`)
```
ivasilov, line 15: This generates a `App 451` type of name. There was a mention of generating names using 2 nouns, do we have that type of generator anywhere else?

studio/components/interfaces/SignIn/SignInSSOForm.tsx
```diff
@@ -38,7 +38,7 @@ const SignInSSOForm = () => {
           process.env.NEXT_PUBLIC_VERCEL_ENV === 'preview'
             ? location.origin
             : process.env.NEXT_PUBLIC_SITE_URL
-        }${BASE_PATH}${getReturnToPath()}`,
+        }${BASE_PATH}${getReturnToPath('/')}`,
```
ivasilov, line 41: This is to force the user to be redirected to the sign in page, where he would be checked if he has MFA.

studio/components/interfaces/SignIn/SignInWithGitHub.tsx
```diff
@@ -24,7 +24,7 @@ const SignInWithGitHub = () => {
             process.env.NEXT_PUBLIC_VERCEL_ENV === 'preview'
               ? location.origin
               : process.env.NEXT_PUBLIC_SITE_URL
-          }${BASE_PATH}${getReturnToPath()}`,
+          }${BASE_PATH}${getReturnToPath('/')}`,
```
ivasilov, line 27: This is to force the user to be redirected to the sign in page, where he would be checked if he has MFA.

studio/components/layouts/AccountLayout/AccountLayout.tsx
```diff
@@ -78,6 +78,13 @@ const AccountLayout = ({ children, title, breadcrumbs }: PropsWithChildren<Accou
                 href: `/account/me`,
                 key: `/account/me`,
               },
+              {
```
ivasilov, line 81: This will be behind feature flag.

studio/components/layouts/SignInLayout/SignInLayout.tsx
```diff
@@ -77,20 +88,12 @@ const SignInLayout = ({
 
   return (
     <>
-      {IS_PLATFORM && (
-        <Head>
-          <script
-            dangerouslySetInnerHTML={{
-              __html: `window._getReturnToPath = ${getReturnToPath.toString()};if (localStorage.getItem('${STORAGE_KEY}') && !(new URLSearchParams(location.search).has('next'))) {location.replace('${
```
ivasilov, line 84: This code no longer works because the storage key is set when there's `aal1` level for the user. If the user has MFA (and requires `aal2` to fully function), he'll be ping-ponged between `/dashboard` and `/sign-in`.

studio/pages/sign-in.tsx
```diff
@@ -1,37 +1,19 @@
...
-        router.replace('/projects')
-      }
-    })
-
-    return () => {
-      subscription.unsubscribe()
-    }
-  }, [])
+  if (!IS_PLATFORM) {
+    // on selfhosted instance just redirect to projects page
```
ivasilov, line 14: No need for all this code happening in `useEffect`, just replaced it with a simple `if`.

studio/components/layouts/AccountLayout/AccountLayout.tsx
alaister: Sounds correct to me!
Small thing, but would you mind also adding this 'Security' link to the V2 navigation?

studio/components/layouts/SignInLayout/SignInLayout.tsx
```diff
@@ -42,24 +43,34 @@ const SignInLayout = ({
...
-        data: { session },
-      } = await auth.getSession()
-
-      if (session) {
-        await queryClient.resetQueries()
-
-        await pushNext()
-      }
-    })()
+      auth.getSession().then(async ({ data: { session } }) => {
```
alaister, line 52: Let's make @hf happy and use `getAccessToken()` here instead of `auth.getSession()`

studio/components/interfaces/SignIn/SignInMfaForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const pushNext = usePushNext()
+  const queryClient = useQueryClient()
+  const { data: factors, isSuccess: factorsSuccess } = useMfaListFactorsQuery()
+  const { mutateAsync: mfaChallengeAndVerify, isLoading } = useMfaChallengeAndVerifyMutation()
+  const [selectedFactor, setSelectedFactor] = useState<Factor | null>(null)
+
+  // if the user wanders into this page and he has no MFA setup, send the user to the next screen
+  if (factorsSuccess && factors.totp.length === 0) {
+    queryClient.resetQueries().then(() => pushNext())
+  }
```
alaister, line 27: Should this be in a `useEffect()`?

alaister (commented):
Looking great! Everything seems to be working, so only a couple of code comments:

studio/pages/sign-in-mfa.tsx
```diff
@@ -0,0 +1,25 @@
+import SignInMfaForm from 'components/interfaces/SignIn/SignInMfaForm'
+import { SignInLayout } from 'components/layouts'
+import { NextPageWithLayout } from 'types'
+
+const SignInMfaPage: NextPageWithLayout = () => {
+  return (
+    <>
```
joshenlim, line 7: nit: unnecessary fragment

studio/components/layouts/AccountLayout/AccountLayout.tsx
joshenlim: ^ yeah this one's okay - IS_PLATFORM is just what we use to discern between self-hosted and hosted (its not a feature flag that we use through configcat)

should be present despite having mfa set up or not

joshenlim:
@ivasilov im gonna jump on your PR to make some style changes if you don't mind 🙏

studio/pages/account/security.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+Security.getLayout = (page) => (
+  <AccountLayout title="Security" breadcrumbs={[{ key: 'security', label: 'Security' }]}>
+    {page}
+  </AccountLayout>
+)
+
+export default Security
+
+const ProfileCard = () => {
```
joshenlim, line 79: could we shift all these components into `components/interfaces/Account/Security`?

studio/pages/account/security.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+  return (
+    <div className="1xl:px-28 mx-auto flex flex-col px-5 pt-6 pb-14 lg:px-16 xl:px-24 2xl:px-32">
+      <div className="flex items-center justify-between">
+        <FormHeader
+          title="Multi-factor authentication"
+          description="Add an additional layer of security to your account by requiring more than just a password to sign in."
+        />
+      </div>
+      <Accordion
```
joshenlim, line 37: let's use `Collapsible` instead - refer to ProviderForm.tsx
the styling should be the same as that from the auth providers page

studio/pages/account/security.tsx
```diff
@@ -0,0 +1,246 @@
...
+  Button,
+  IconAlertCircle,
+  IconAlertTriangle,
+  IconSmartphone,
+  IconTrash,
+  Modal,
+} from 'ui'
+
+const Security: NextPageWithLayout = () => {
+  const { data } = useMfaListFactorsQuery()
```
joshenlim, line 27: handle isLoading, isSuccess, isError states

studio/pages/account/security.tsx
```diff
@@ -0,0 +1,246 @@
...
+          header={
+            <div className="flex flex-row justify-between items-center p-2 w-full">
+              <div className="flex flex-row gap-2 items-center p-2">
+                <IconSmartphone />
+                <span className="text-lg">Authenticator app</span>
+              </div>
+
+              {data ? (
+                <>
+                  {data.totp.length === 0 && <Badge color="red">0 apps configured</Badge>}
```
joshenlim, line 54: Could be simplified to `{data.totp.length} apps configured` ?

studio/pages/account/security.tsx
joshenlim: also nit: ProfileCard sounds odd as the component name here haha
maybe TOTPFactors?

studio/pages/account/security.tsx
```diff
@@ -0,0 +1,246 @@
...
+    <>
+      <section className="px-6 pt-4">
+        <div>
+          <span className="text-sm prose color-slate-1000">
+            Generate one-time passwords via authenticator apps like 1Password, Authy, etc. as a
+            second factor to verify your identity when prompted during sign-in.
+          </span>
+        </div>
+        {isLoading && (
+          <div className="py-2">
```
joshenlim, line 94: <GenericSkeletonLoader />

joshenlim: also is the py-2 required? otherwise lets just wrap isLoading, isError, isSuccess with the py-n instead

joshenlim: mainly just to keep things consistent cause below we've got py-4 for the error state

studio/pages/account/security.tsx
```diff
@@ -0,0 +1,246 @@
+import { AddNewFactorModal } from 'components/interfaces/Account'
+import { AccountLayout } from 'components/layouts'
+import AlertError from 'components/ui/AlertError'
+import ConfirmationModal from 'components/ui/ConfirmationModal'
+import { FormHeader } from 'components/ui/Forms'
+import ShimmeringLoader from 'components/ui/ShimmeringLoader'
+import { useMfaListFactorsQuery } from 'data/profile/mfa-list-factors-query'
+import { useMfaUnenrollMutation } from 'data/profile/mfa-unenroll-mutation'
+import { useStore } from 'hooks'
+import { useState } from 'react'
```
joshenlim, line 10: let's order the import statements as well

external imports
-empty line-
internal imports

studio/pages/account/security.tsx
```diff
@@ -0,0 +1,246 @@
...
+                  <div key={factor.id} className="flex flex-row justify-between py-2">
+                    <div className="flex flex-col">
+                      <span className="text-sm text-white">
+                        Name: {factor.friendly_name ?? 'No name provided'}
+                      </span>
+                      <span className="text-sm text-scale-900">Factor ID: {factor.id}</span>
+                    </div>
+                    <div className="flex items-center gap-2">
+                      <span className="text-sm text-scale-900">
+                        Created at {new Date(factor.updated_at).toLocaleString()}
```
joshenlim, line 130: use dayjs to format, we should include the timezone `'DD MMM YYYY, HH:mm:ss (ZZ)'`

which on that note - i think we should really standardize the way we format timestamps on the dashboard

studio/components/interfaces/SignIn/SignInMfaForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+    <Form
+      validateOnBlur
+      id="sign-in-mfa-form"
+      initialValues={{ code: '' }}
+      validationSchema={signInSchema}
+      onSubmit={onSignIn}
+    >
+      {() => (
+        <div className="flex flex-col gap-4">
+          <Input
```
joshenlim, line 77: can we enable autofocus on this input field

joshenlim:
one observation - if say im logged out, and I'm trying to access a project route (e.g supabase.com/dashboard/project/ref/sql/new), but because i'm logged out, i get bumped to the sign in page

notice that the URL on the sign in page would be something like
`/dashboard/sign-in?returnTo=%2Fdashboard%2Fproject%2Feewtenzudjonfysxnjco` -> there's a returnTo parameter

we need to do the same thing on the sign-in-mfa route after verifying MFA, the user should be redirected to /project/ref/sql/new immediately

joshenlim:
another observation - the support form should be accesible despite MFA not verified yet (at least aa1 status)

studio/components/interfaces/Account/AddNewFactorModal.tsx
joshenlim: we don't have, but i think that's okay

studio/data/profile/mfa-challenge-and-verify-mutation.ts
```diff
@@ -0,0 +1,48 @@
...
+      const { refreshFactors, ...params } = vars
+      return mfaChallengeAndVerify(params)
+    },
+    {
+      async onSuccess(data, variables, context) {
+        // when a MFA is added, the aaLevel is bumped up
+        const refreshFactors = variables.refreshFactors ?? true
+
+        await Promise.all([
+          ...(refreshFactors ? [queryClient.invalidateQueries(profileKeys.mfaFactors())] : []),
```
joshenlim, line 39: cc @ivasilov just fyi i added this change here - mainly so when signing in we don't need to refresh this (just saves us one call)

ivasilov: @joshenlim unfortunately, this addition is made irrelevant by this line https://github.com/supabase/supabase/blob/15818116b9c5a1ceb7d924a02139f458b89ec3c1/studio/components/interfaces/SignIn/SignInMfaForm.tsx#L78. All queries gets reset (and they need to be refetched) which causes the input field to clear. It bugged me too so I had to go down the rabbit hole.

joshenlim: ahh i see hmmmm... i think we might still be able to keep this in

am thinking what could've been happening is like after challengeAndVerify succeeds, it invalidates mfafactors + aa levels, and then re-invalidates those 2 again because of the resetQueries

so am thinking its possibly helping to reduce duplicate requests?

joshenlim (approved):

PR #16819, opened 2023-08-25
Only allow numbers in input field

saltcod:
Currently, this input type is string. 
For some reason, the browser autofills my email address into this field. 
Having to manually remove my email from this field multiple times a day is slowly driving me crazy. 
Changing the input type to number solves this issue. 

CleanShot 2023-08-25 at 13 13 30@2x


kevcodez:
Don't think number input works, Versions start with a number between 12-19, a dot and additional characters, i.e. 15.2 or 15.2.0-3, so not a regular number

PR #16820, opened 2023-08-25
Fix duplicate key

saltcod:
"Issues with Apis" and "Abuse report" have the same value:  "Problem" 
So both get selected when one of them is selected.  

CleanShot 2023-08-25 at 13 22 08@2x


studio/components/interfaces/Support/Support.constants.ts
```diff
@@ -36,7 +36,7 @@ export const CATEGORY_OPTIONS = [
     query: undefined,
   },
   {
-    value: 'Problem',
+    value: 'Abuse',
```
joshenlim, line 39: before we get this in, just need to double check on API's end whether this is a valid category

joshenlim: yeah these are the available support categories so this change will send the support ticket to feedback instead

image

joshenlim: while you're on this PR @saltcod would you mind just changing the values to using `SupportCategories` from shared-types? will help us prevent any sort of deviance between FE and API

but yeah these comments are separate to the issue that you're fixing in this PR, may need a different solution

joshenlim (changes requested):
Left some comments but TLDR, 'Abuse' as a category is not supported and will not be funnelled into the support flow

studio/components/interfaces/Support/SupportForm.tsx
```diff
@@ -282,6 +282,7 @@ const SupportForm = ({ setSentCategory }: SupportFormProps) => {
               subject: subject ?? initialValues.subject,
               category: selectedCategoryFromUrl?.value ?? initialValues.category,
               message: message ?? initialValues.message,
+              severity: initialValues.severity,
```
saltcod, line 285: Issue: 
https://supabase.slack.com/archives/C0161K73J1J/p1693412541497319?thread_ts=1693404721.979579&cid=C0161K73J1J

saltcod:
Depends on
https://github.com/supabase/shared-types/pull/46

joshenlim:
lgtm! tested every category on the preview - works :)

joshenlim (approved):

PR #16864, opened 2023-08-26
Update kotlin docs

jan-tennert:

Docs update

Old docs

## What is the new behavior?

Updated docs

## Additional context

There are new modules for supabase-kt with documentation on their Github README, this could also be put here, but I'm not sure how I'd add custom pages but this should be enough for now


kevcodez:
@jan-tennert Apologies for the delay, could you please run prettier again on `apps/docs/docs/ref/kotlin/introduction.mdx`?

kevcodez (approved):

PR #16887, opened 2023-08-27
chore: add invoice check for project pause state

pamelachia:

add checks on project pause state so users shouldn't be able to resume projects if they have outstanding payments
Screenshot 2023-08-28 at 6 24 07 AM

### DIFF
--- a/studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
+++ b/studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
@@ -12,8 +12,9 @@ import { useProjectRestoreMutation } from 'data/projects/project-restore-mutatio
 import { setProjectStatus } from 'data/projects/projects-query'
 import { useProjectSubscriptionV2Query } from 'data/subscriptions/project-subscription-v2-query'
 import { useCheckPermissions, useSelectedOrganization, useStore } from 'hooks'
-import { PROJECT_STATUS } from 'lib/constants'
+import { IS_PLATFORM, PROJECT_STATUS } from 'lib/constants'
 import { useProjectContext } from './ProjectContext'
+import { useOverdueInvoicesQuery } from 'data/invoices/invoices-overdue-query'
 
 export interface ProjectPausedStateProps {
   product?: string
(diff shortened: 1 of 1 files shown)


studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
```diff
@@ -95,56 +98,64 @@ const ProjectPausedState = ({ product }: ProjectPausedStateProps) => {
...
-                    Unpaused projects count towards compute usage. For every hour your instance is
-                    active, we'll bill you based on the instance size of your project. See{' '}
-                    <Link href="https://www.notion.so/supabase/Organization-Level-Billing-707638e35c92489995dc3ac991a324d1">
-                      <a target="_blank" rel="noreferrer" className="underline">
-                        Compute Instance Usage Billing
-                      </a>
-                    </Link>{' '}
-                    for more details.
+                    You have outstanding invoices that need to be paid before you’re able to resume
+                    this project. Please contact support if you need more help.
```
kevcodez, line 121: Should have a CTA to get to the outstanding invoices (/org/<slug>/invoices)

contact support should also be a link to to the support/new form

studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
```diff
@@ -29,6 +30,8 @@ const ProjectPausedState = ({ product }: ProjectPausedStateProps) => {
   const { data: subscription } = useProjectSubscriptionV2Query({ projectRef: ref })
   const isFreePlan = subscription?.plan?.id === 'free'
   const billedViaOrg = Boolean(selectedOrganization?.subscription_id)
+  const { data: allOverdueInvoices } = useOverdueInvoicesQuery({ enabled: IS_PLATFORM })
+  const haveOverdueInvoices: boolean = (allOverdueInvoices?.length ?? 0) > 1
```
kevcodez, line 34: The overdue invoices endpoint doesn't work here, as it covers ALL projects a user is a member of. It's used on the front page where all orgs are displayed.

For this use-case, we need to only check for the currently selected org

kevcodez (changes requested):

kevcodez:
Check(s) definitely need to be implemented on the backend, too. Otherwise people can just unpause through the API with outstanding invoices

studio/components/layouts/ProjectLayout/ProjectPausedState.tsx
kevcodez: We could also remove the check from the frontend entirely and move it to the backend, then just react to the appropriate error and display a warning. Then we can control this logic on the backend, without duplicating it here

pamelachia: Got it on just implementing another endpoint in the backend.
For users without org-level billing, I need to check on a project level for now. Do I check for the `stripe_customer_id` related to the org only? How do I find on a project level?

kevcodez: All projects in an org point to a single Stripe customer, so it's okay to not let a user unpause ANY project if the related org has open invoices (check works for both org and non org level billing).

There is not necessarily a need to implement a new endpoint at all. See 

> We could also remove the check from the frontend entirely and move it to the backend, then just react to the appropriate error and display a warning. Then we can control this logic on the backend, without duplicating it here

This would not require any new endpoint. When the user presses unpause, it would show an error about outstanding invoices and could then display the CTA. It's a bit delayed then, but reduces code complexity and avoids spreading the logic in multiple places

PR #16900, opened 2023-08-28
chore: update scheduled backups view

dragarcia:
## Main changes
* Remove Download button for backups.
    * Will follow up with another PR to clean up dangling components.
* Use a more accurate variable in the form of `pitr_enabled` to indicate whether a project is subscribed to PITR or not.
* Change sorting of scheduled backups by `inserted_at` instead of `id`.
* Update request body sent to the restore endpoint.


darora (commented):
Any way to make this independent of the backend change?

joshenlim (approved):

PR #16904, opened 2023-08-28
fix(studio): invalid sql in table definitions

rmrt1n:
Fixed invalid SQL in table definitions when columns are named as reserved identifiers in postgres, mentioned in [#13869](https://github.com/supabase/supabase/issues/13869)

Currently, columns with the same names as reserved keywords (e.g. desc, and, or) are not quoted, unlike columns with capital letters in them are (e.g. productId).

## What is the new behavior?
Now these columns are quoted just like camelCase ones. The reserved keywords are fetched using the `pg_get_keywords()` function.
### Old
image

### New
image

## Additional context
I couldn't reproduce a related issue: `If the definition is copied and pasted into the SQL Editor to create a table any camelCase columns are converted to lowercase`. Also, the `pg_get_tabledef` is really large, so I'm not sure if I missed some parts to change. I'd love to get some feedback on this!

### DIFF
--- a/studio/data/database/database-query-constants.ts
+++ b/studio/data/database/database-query-constants.ts
@@ -298,8 +298,8 @@ export const CREATE_PG_GET_TABLEDEF_SQL = minify(
             END IF;
             
             --v17 put double-quotes around case-sensitive column names
-            SELECT COUNT(*) INTO v_cnt FROM information_schema.columns t WHERE EXISTS (SELECT REGEXP_MATCHES(s.column_name, '([A-Z]+)','g') FROM information_schema.columns s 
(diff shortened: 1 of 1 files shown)


alaister:
Hey @rmrt1n,
Thanks for your PR!

We use https://github.com/MichaelDBA/pg_get_tabledef to get the table definitions, so would you mind opening up this PR there? That way we can stay consistent 😄

Thanks again!

rmrt1n:
Hi @alaister, just wanted to let you know that the fix has been implemented there!

alaister:
Awesome thanks for letting me know – update PR here: https://github.com/supabase/supabase/pull/16967

PR #16961, opened 2023-08-29
Skip unnecessary calls if user is not logged in

joshenlim:


studio/components/interfaces/App/RouteValidationWrapper.tsx
```diff
@@ -16,6 +16,9 @@ const RouteValidationWrapper = ({ children }: PropsWithChildren<{}>) => {
   const { ref, slug } = useParams()
   const navLayoutV2 = useFlag('navigationLayoutV2')
 
+  const { session } = useAuth()
+  const isLoggedIn = Boolean(session)
```
alaister, line 20: You can use `useIsLoggedIn()` from `lib/auth` here :)

alaister (commented):
Looks good! There's a handy hook already for logged in status:

studio/components/interfaces/App/RouteValidationWrapper.tsx
joshenlim: oooh okay gotcha - i'll update accordingly 🙏

joshenlim: just pushed :)

studio/components/ui/PageTelemetry.tsx
```diff
@@ -59,14 +61,16 @@ const PageTelemetry = ({ children }: PropsWithChildren<{}>) => {
...
 
-      post(`${API_URL}/telemetry/pageview`, {
-        ...(ref && { projectRef: ref }),
-        ...(selectedOrganization && { orgSlug: selectedOrganization.slug }),
-        referrer: referrer,
-        title: document.title,
-        path: router.route,
-        location: router.asPath,
-      })
+      if (isLoggedIn) {
```
alaister, line 64: Just to confirm, we only want to register page views for logged in users? (aka we don't care about the views on the sign in pages?)

joshenlim: @pamelachia - best to check with you about this. the only reason im guarding the pageview call is cause it currently 401s on the sign in pages

alaister: oh if it's 401 already, I assume it's fine!

alaister (approved):
LGTM!

Issue #15359, opened 2023-06-26
Main Site pricing page: nested button in anchor [a11y]


GrahamTheDevRel:
## Describe the bug

On the pricing page there are `<button>` elements nested within `<a>` elements.

This is a big accessibility issue as it will cause confusion for screen reader users around expected behaviour as a button indicates an action on the same page, an anchor indicates a change of page and they will encounter a link, then a button, both with the same text.

It also means that each "button" is now 2 tab stops for keyboard users instead of 1, which can be annoying.

It can also mean that on voice navigation (with something like Dragon Naturally Speaking), commands may not work as expected.

## Expected behaviour

As the "buttons" in question lead to another page, these should just be anchors (`<a>`) to avoid confusion. 

Then transfer all the styling from the `<button>` onto the `<a>`.

## Additional context

This appears to be a site wide issue. Therefore there are numerous instances on this page, not just in the pricing table.

Note for beginner devs: the `<button>` element appears to be a custom element on the site, so you will have to bear that in mind if you decide to tackle this issue as you will need to grab all the styles from that element.

Arindam200:
I would like to work on this issue!

GrahamTheDevRel:
@MildTomato WIthout seeing outputted HTML I always struggle, but does this not have the same issue but in reverse?

The issue is nested interactive elements, on my understanding the above fixes the handlers, but won't fix the accessibility tree / other issues?

I would have expected the fix to be more like:

`<a href="https://..." class="all the button classes via @apply or something?">hello there</a>` and nothing else?

I may not be understanding though as I am vanilla HTML, CSS so apologies if I misunderstood! 💗

MildTomato:
The `asChild` prop is a bit odd to get your head around, but perhaps best explained [here](https://www.radix-ui.com/docs/primitives/guides/composition) on radix docs.


The following:
```
<Button asChild className="button-class">
  <a href="https://...">hello there</a>
</Button>
```

would output this:

```
<a href="https://..." className="button-class">hello there</a>
```

GrahamTheDevRel:
@MildTomato that makes perfect sense. Also the `asChild` prop is clear, I just wasn't familiar and thanks for my "Today I Learned (TIL)" moment! Looks like the perfect solve.

MildTomato:
@GrahamTheDevRel I can guarantee it took me a lot longer for me to understand that one 😂

PR #16979, opened 2023-08-29
Improve site a11y

fsansalvadore:

A11y and keyboard navigation UX and styling:

- Improve focus on tab on main pages / components
- Remove double focus on Buttons inside Links
- More semantic html tags

Fixes:
- #15359 

https://github.com/supabase/supabase/assets/25671831/aa74b6f1-5bee-4193-b4a9-002c0d4b640d

https://github.com/supabase/supabase/assets/25671831/744cba09-e7c2-4d0d-96c0-bd509a9ab682

### DIFF
--- a/apps/www/components/BuiltWithSupabase/index.tsx
+++ b/apps/www/components/BuiltWithSupabase/index.tsx
@@ -16,22 +16,18 @@ const BuiltExamples = () => {
         <p className="p">There are many example apps and starter projects to get you started.</p>
         <div className="flex justify-center gap-2 py-4">
           <Link href="/docs/guides/examples" as="/docs/guides/examples" passHref>
-            <a tabIndex={-1}>
-              <Button type="default" size="small" className="h-full">
-                View all examples
-              </Button>
-            </a>
+            <Button type="default" size="small" className="h-full" asChild>
+              <a>View all examples</a>
+            </Button>
           </Link>
           <Link
             href="https://github.com/supabase/supabase/tree/master/examples"
             as="https://github.com/supabase/supabase/tree/master/examples"
             passHref
           >
-            <a tabIndex={-1}>
-              <Button type="default" icon={<IconGitHub />} size="small">
-                Official GitHub library
-              </Button>
-            </a>
+            <Button type="default" icon={<IconGitHub />} size="small" asChild>
+              <a>Official GitHub library</a>
+            </Button>
           </Link>
         </div>
       </div>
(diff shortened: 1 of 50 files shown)


apps/www/components/BuiltWithSupabase/index.tsx
```diff
@@ -16,7 +16,7 @@ const BuiltExamples = () => {
         <p className="p">There are many example apps and starter projects to get you started.</p>
         <div className="flex justify-center gap-2 py-4">
           <Link href="/docs/guides/examples" as="/docs/guides/examples" passHref>
-            <a>
+            <a tabIndex={-1}>
```
MildTomato, line 19: you can do 

```
<Link pashref>
  <Button aschild>
    <a>

**Review by MildTomato — CHANGES_REQUESTED** [review]

File: apps/www/styles/index.css [hunk 2]
```diff
@@ -8,6 +8,33 @@
 @import './../../../packages/ui/build/css/themes/dark.css';
 @import './../../../packages/ui/build/css/themes/light.css';
 
+@layer components {
+  .focus-base {
```
MildTomato, line 12: Could put these into the tailwind config maybe.
Don't have to do it in this PR

MildTomato (commented):
.

apps/www/components/BuiltWithSupabase/index.tsx
fsansalvadore: Oh, that's how it's composed... I was trying the Button outside of the Link tag and it wasn't working 😁
Will update thanks 👍

apps/www/styles/index.css
fsansalvadore: That would be ideal, but I couldn't figure out a way to use raw tailwind classes to create a new class 🤔

Like:

```js
extend: {
  focus: {
    base: '@apply focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-scale-900 ring-0 transition-all focus-visible:ring-offset-transparent',
  }
}
```

Any idea?

MildTomato: Need to use a utility for a composition of tailwind.

Could use matchUtilities
https://tailwindcss.com/docs/plugins#dynamic-utilities

Or maybe make a utility just to reset all the focus stuff, so all you need then is something like `focus-reset focus-visible:ring-scale-900`

apps/www/components/ExampleCard.tsx
```diff
@@ -47,7 +47,7 @@ function ExampleCard(props: any) {
         >
           <Link href={props.repo_url} as={props.repo_url} passHref>
             <a
-              className="text-scale-1100 hover:text-scale-1200 flex flex-row items-center text-sm"
+              className="text-scale-1100 hover:text-scale-1200 flex flex-row items-center text-sm focus-base focus-rounded-sm"
```
MildTomato, line 50: Can we remove all the scale stuff as well?

`text`, `text-light` etc

MildTomato (changes requested):
Maybe we are regressing on the tailwind classes

PR #16993, opened 2023-08-29
docs: update storage & auth docs

kangmingtay:
* Improve access control guide for storage
* Update linkedin oauth docs to reflect the API change
* Use admonitions where necessary for callouts

### DIFF
--- a/apps/docs/pages/guides/auth/social-login/auth-linkedin.mdx
+++ b/apps/docs/pages/guides/auth/social-login/auth-linkedin.mdx
@@ -12,7 +12,7 @@ To enable LinkedIn Auth for your project, you need to set up a LinkedIn OAuth ap
 
 <Admonition type="caution" label="update">
 
-Linkedin has changed some of their APIs for OAuth which won't work with the current Supabase implementation. See [here](https://learn.microsoft.com/en-us/linkedin/shared/authentication/authorization-code-flow?context=linkedin%2Fcontext&tabs=HTTPS1) for more information.
+Linkedin has updated their [OAuth APIs](https://learn.microsoft.com/en-us/linkedin/shared/authentication/authorization-code-flow?context=linkedin%2Fcontext&tabs=HTTPS1), which has broken the current Supabase implementation. We're actively tracking [this issue](https://github.com/supabase/gotrue/issues/1216) and you can subscribe to it for future updates.
 
 </Admonition>
 
--- a/apps/docs/pages/guides/auth/social-login/auth-linkedin.mdx
+++ b/apps/docs/pages/guides/auth/social-login/auth-linkedin.mdx
@@ -10,6 +10,12 @@ To enable LinkedIn Auth for your project, you need to set up a LinkedIn OAuth ap
 
 ## Overview
 
+<Admonition type="caution" label="update">
+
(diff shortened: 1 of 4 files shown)


apps/docs/pages/guides/storage/access-control.mdx
```diff
@@ -34,7 +34,27 @@ To allow read access to files, the RLS policy must allow users to `SELECT` the `
...
+For public buckets, you can access the assets directly without a token or an Authorisation header. The [getPublicUrl](/docs/reference/javascript/storage-from-getpublicurl) helper method returns the full public URL for an asset. This calls the `/object/public/` API endpoint internally. While no authorization is required for accessing objects in a public bucket, proper access control is required for other operations like uploading, deleting objects from public buckets as well. An easy way to get started would be to create RLS policies for `SELECT`, `INSERT`, `UPDATE`, `DELETE` operations and tightening the policies until it meets your security requirements. For example, one can start with the following `INSERT` policy:
+
+```sql
+create policy "policy_name"
```
inian, line 40: I don't think we should be recommending this even as a starting point. This will give users access to all buckets, even ones which are created in the future which is dangerous. 

Probably we can move this part to [this section](https://supabase.com/docs/guides/storage/access-control#policies) instead and link to that? Since this is relevant for both public and private buckets.

kangmingtay: updated the copy and moved it to the policies section

apps/docs/pages/guides/auth/social-login/auth-linkedin.mdx
```diff
@@ -10,6 +10,12 @@ To enable LinkedIn Auth for your project, you need to set up a LinkedIn OAuth ap
 
 ## Overview
 
+<Admonition type="caution" label="update">
```
kiwicopple, line 13:
```suggestion
<Admonition type="caution" label="LinkedIn">
```

apps/docs/pages/guides/auth/social-login/auth-linkedin.mdx
```diff
@@ -10,6 +10,12 @@ To enable LinkedIn Auth for your project, you need to set up a LinkedIn OAuth ap
...
+
+Linkedin has changed some of their APIs for OAuth which won't work with the current Supabase implementation. See [here](https://learn.microsoft.com/en-us/linkedin/shared/authentication/authorization-code-flow?context=linkedin%2Fcontext&tabs=HTTPS1) for more information.
```
kiwicopple, line 15:
```suggestion
LinkedIn has updated their OAuth APIs, which has broken the implementation for the current Supabase implementation. See [here](https://learn.microsoft.com/en-us/linkedin/shared/authentication/authorization-code-flow?context=linkedin%2Fcontext&tabs=HTTPS1) for more information.
```

kiwicopple: we also need to share what we are doing about this. Will we deprecate it? Are we looking into it? How can developers track our progress?

kiwicopple (approved):

apps/docs/pages/guides/auth/social-login/auth-linkedin.mdx
kangmingtay: keeping this as update because it looks weird if the label as "linkedin"
image

PR #17033, opened 2023-08-30
Add JWT init script

Lan-Hekary:

As we discussed in #16681 passing JWT info to PostgreREST only applies them to rest transactions not the entire DB.
This PR apply them to the entire DB through an init script.

## Discussion needed
The only downside that this script runs only one time if there is no database data directory.
meaning that this script won't run for existing databases and if the user changed the secret in the .env file it won't be updated until this script in manually executed.
Any Solution for this ?

Also should we revert the #16681 changes as this PR effectively replaces it ?


Lan-Hekary:
Hello @soedirgo,
Please take a look and advise me on the best course of action so that I can edit the PR accordingly.
Thanks.

soedirgo:
Thanks again @Lan-Hekary!

> if the user changed the secret in the .env file it won't be updated until this script in manually executed

The easiest thing would be to add a comment warning about it in the .env - this is already a problem for `POSTGRES_PASSWORD` etc.

> Also should we revert the https://github.com/supabase/supabase/pull/16681

Sounds good to me :+1:

soedirgo (approved):

PR #17038, opened 2023-08-30
Fix the SupportForm crash

ivasilov:
The support page was broken in https://github.com/supabase/supabase/pull/17011. Two undefined values were passed and the form didn't expect them so it crashed.


alaister (commented):
I think it's best practice for no values to be `undefined` in a react form, so maybe it's worth defaulting them all to an empty string? Something like:
Screenshot 2023-08-31 at 12 30 10 am

alaister (approved):

alaister (approved):

PR #17045, opened 2023-08-30
chore: add tests dispatch on prod deploy

egor-romanov:

- chore: add tests dispatch event on prod deploys


.github/workflows/smoke.yml
```diff
@@ -0,0 +1,27 @@
+# We are using Playwright but you may use your E2E framework of choice
+name: Smoke Tests
+
+on:
+  deployment_status:
+  workflow_dispatch:
+
+jobs:
+  trigger_tests:
+    if: github.event_name == 'workflow_dispatch' || github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success' && github.event.deployment_status.environment == 'Production – studio'
```
MildTomato, line 10: `github.event.deployment_status.environment == 'Production – studio` 

that seems odd. sure that works?

egor-romanov: We should test when the build is actually deployed, which is controlled by this event

This is according to docs/guides though, we have to try this out to test if it works like this

PR #17066, opened 2023-08-31
Fix support form not passing site url and additioanl redirect urls

joshenlim:


ivasilov (approved):
Tested, works correctly.

ivasilov (approved):

Issue #17029, opened 2023-08-30
Dashboard keep showing loading for a long time when import data with wrong format


saltcod:
### Discussed in https://github.com/orgs/supabase/discussions/16948

<div type='discussions-op-text'>

<sup>Originally posted by **hieuwu** April 11, 2023</sup>
### Discussed in https://github.com/orgs/supabase/discussions/13528

<div type='discussions-op-text'>

<sup>Originally posted by **hieuwu** April  8, 2023</sup>
# ⚙️ Summary
Currently, when I import data from CSV file, I notice that if we set the value of a column on Supabase is float4 and the data of this column is a string, the platform will stop working without any response.
For example, my column is `price` with type is `float4`. I import a data set with a row containing value at price column is "this is a string". The platform keep showing loading message forever

<img width="720" alt="Screen Shot 2023-04-08 at 14 33 58" src="https://user-images.githubusercontent.com/43868345/230711338-f1a51f0c-4460-426d-a938-7d775b6bcea0.png">

# 💡 Suggestion
It would be nice if we have some mechanism to verify the format of data for number before performing any operation.

I imported the file below
[product.csv](https://github.com/supabase/supabase/files/11182812/product.csv)
</div></div>

Hallidayo:
@saltcod - On importing the example spread sheet the studio now gives feedback to the user about issues about the data in the spreadsheet. I seem to get warnings on both `Import` and `Paste` and I get really good feedback for changes. Seems okay to close?

saltcod:
thanks @Hallidayo !

PR #17072, opened 2023-08-31
feat(studio): check column types in imported spreadsheet

rmrt1n:
Spreadsheet importer now checks the column types of the imported data with the table's schema. Mentioned in [#17029](https://github.com/supabase/supabase/issues/17029).

Currently no type check is done.

## What is the new behavior?
Now the column types are checked. The type checking performed isn't strict, as some types can be casted to another, e.g. a column with a float type can still accept int types, a column with a text type can still accept numeric text (which would be inferred as int/float by inferColumnType).
### Old
image

### New
image

## Additional context


saltcod (approved):

saltcod:
checked preview, thanks @rmrt1n !

PR #17087, opened 2023-08-31
fix: $ in front of column name causes issue with definition tab

### LINKED ISSUE #17013: $ in front of column name causes issue with definition tab
**Issue:** $ symbol in front of a column name breaks the UI

image

**Steps to reproduce:**
1. Create a new table with multiple columns
2. Make sure one of the column contains $ symbol in front of the name. (In my case $age - Check image attached below)
3. Go to table definition and the UI breaks as in image attached before.

image

Bhavyawahie:

Fixes [#17013](https://github.com/supabase/supabase/issues/17013). The given PR brings about **client-side validation** on column's `name` field for a given table's definition made using the `Table Editor.`  

The current behavior allows for naming columns of a given table without any validation for special characters `(case in point being column names beginning with'$')` which cause a syntax error with the supabase-meta container running the postgres-meta service.

> 

###  **Replicating the error**

> ezgif com-video-to-gif (11)

###  **Error from Postgres-meta container's logs**

> ```shell
> "request":{"method":"POST","url":"/query","pg":"db","opt":""}}
> {"level":"error","time":"2023-08-31T11:31:36.849Z","pid":19,"hostname":"a26f486d93cb","reqId":"req-2st","error":{"message":"syntax error at or near \"$\""}," \"$\""}
> ``` 

## What is the new behavior?

The changes in validation allow for users to not create a table until the said column name within that table does not begin with a '$'. Additionally, any previously created column names are also validated on any subsequent updation to them.

### **Table Creation**

> ezgif com-video-to-gif (13)

### **Table Updation**

> ezgif com-video-to-gif (12)

### **Accessing Table Definition**

> ezgif com-video-to-gif (14)

This also solves the problem of `Application error` that was earlier caused by column names beginning with '$' on accessing the Table Definition tab inside the Table Editor.

### DIFF
--- a/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/ColumnManagement.tsx
+++ b/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/ColumnManagement.tsx
@@ -209,7 +209,7 @@ const ColumnManagement = ({
                       ].join(' ')}
                     >
                       <span className="text-xs text-scale-1200">
-                        Recommended to use lowercase and use an underscore to separate words e.g.
+                        Recommended to use lowercase (must not begin with $)  and use an underscore to separate words e.g.
                         column_name
                       </span>
                     </div>
--- a/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/ColumnManagement.tsx
+++ b/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/ColumnManagement.tsx
@@ -209,7 +209,7 @@ const ColumnManagement = ({
                       ].join(' ')}
                     >
                       <span className="text-xs text-scale-1200">
-                        Recommended to use lowercase (must not begin with $)  and use an underscore to separate words e.g.
+                        Recommended to use only lowercase letters, with an underscore to separate words e.g.
                         column_name
                       </span>
                     </div>
--- a/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
+++ b/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
@@ -177,6 +177,9 @@ export const validateFields = (field: ColumnField) => {
   if (field.name.length === 0) {
     errors['name'] = `Please assign a name for your column`
   }
+  if (field.name.startsWith('$')) {
+    errors['name'] = `Column name must not begin with $`
+  }
   if (field.format.length === 0) {
     errors['format'] = `Please select a type for your column`
   }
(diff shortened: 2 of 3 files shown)


Bhavyawahie:
@alaister have a look!

studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/ColumnManagement.tsx
```diff
@@ -209,7 +209,7 @@ const ColumnManagement = ({
                       ].join(' ')}
                     >
                       <span className="text-xs text-scale-1200">
-                        Recommended to use lowercase and use an underscore to separate words e.g.
+                        Recommended to use lowercase (must not begin with $)  and use an underscore to separate words e.g.
```
kiwicopple, line 212:
```suggestion
                        Recommended to use only lowercase letters, with an underscore to separate words e.g.
```

alaister (changes requested):
Hey @Bhavyawahie,

Thanks for the PR!

While not recommended, it is supported in postgres to name columns beginning with `$`. For example, we also allow columns with spaces in the name. Additionally, this doesn't seem to fix the definitions tab error for people who already have columns named with `$`.

With this in mind, could you remove this new validation and fix the bug in the definitions tab? I'm happy to help out with some pointers if it isn't clear where the error might be (I have a feeling it might be in our SQL formatted library 😕)

Thanks!

Bhavyawahie:
Thanks for the feedback @alaister. Happy to work as instructed although I feel this validation check is also necessary at this point until the said changes in the definitions tab are proposed/made. 

The SQL editor also disallows to name/rename column names with `column_name` beginning with $ sign or any special character for that matter `(/^[a-zA-Z_][a-zA-Z0-9_]*$/)` thus may cause irregularities in the UI as far as SQL column nomenclature rules across the platform is concerned.
 
> 
image

alaister:
Postgres (SQL) is doing the validation, which is causing that error to happen in the SQL editor. If you wrap the column name containing the `$` or spaces with double quotes (`"`) it will work. So do you think you can make the changes to the definitions tab here in this PR? 🙏🏻

Bhavyawahie:
Sure, given if you could point me to where the error might be 😅🙏

rmrt1n:
This is `sql-formatter`'s parsing error caused by `pg_get_tabledef` returning unquoted `$column_name`. Updating `pg_get_tabledef` to quote column/table names starting with `$` should fix the issue (like below):
image

However, column names with spaces in them, as shown above, are still broken. Handling quoted identifiers is tricky. Currently, `pg_get_tabledef` quotes columns that are: 1. contains a capital letter, and 2. is a reserved keyword. A quick workaround is to quote every column in the table.

A better(?) approach might look like this:
1. quote reserved keywords
2. quote all columns that doesn't follow postgres's valid identifer syntax: `/^[a-z_][a-zA-Z0-9_$]$/`. This should include columns that are camelCase, $dollar_sign, and "contains spaces".

This should probably fix most parse issues related to table/column names. Unfortunately this doesn't take into account identifiers that contain non-latin characters. Any thoughts @alaister ?

alaister:
Hi,

Unfortunately, this has got a little stale, so I'm going to go ahead and close it.

I think this issue should be solved on the https://github.com/MichaelDBA/pg_get_tabledef side, so maybe you can consider opening up an issue there? Thanks 🙏🏻

PR #17088, opened 2023-09-01
fixed mobile documentation site menu

AbhijeetBiswalCodes:

https://github.com/supabase/supabase/assets/112644558/39fe779d-7821-4f85-8375-f00bbca62cab 

https://github.com/supabase/supabase/assets/112644558/3def0fe8-b972-48a3-b631-c55b20927232

#what is the current behaviour? 

currently when the mobile menu of docs opened and when we scroll down in the menu the scroll gets out of the screen and the right side of the screen is also scrollable.

#what is the new behaviour? 
the scroll in the menu is fixed now its working perfectly.


fsansalvadore (approved):
That seems to do the trick, thanks @AbhijeetBiswalCodes 👍

PR #17093, opened 2023-09-01
docs: add sign in with token hash

J0:

We recently allowed verification in with token hash [as well as guides for email SSR](https://supabase.com/docs/guides/auth/server-side/email-based-auth-with-pkce-flow-for-ssr)

This is the corresponding docs update.


spec/supabase_js_v2.yml
```diff
@@ -415,6 +415,7 @@ functions:
     notes: |
       - The `verifyOtp` method takes in different verification types. If a phone number is used, the type can either be `sms` or `phone_change`. If an email address is used, the type can be one of the following: `email`, `recovery`, `invite` or `email_change` (`signup` and `magiclink` types are deprecated).
       - The verification type used should be determined based on the corresponding auth method called before `verifyOtp` to sign up / sign-in a user.
+      - It is also possible to use the `TokenHash` found in the [email templates](/docs/guides/auth/auth-email-templates) as a sign in method
```
J0, line 418: We should probably link to the workaround from here

kiwicopple (approved):

PR #17095, opened 2023-09-01
Fix the bucket url to handle trailing whitespace

### LINKED ISSUE #16641: The storage UI allows a trailing space for bucket name, but this does not work.
# Bug report

- [x ] I confirm this is a bug with Supabase, not with my own application. 
- [x ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

If you accidentally enter a trailing space in a new bucket name (spaces are supposedly not allowed according to the UI message) the bucket is created but cannot be accessed at all from the UI.  Because it is a space it is impossible to see this in the UI.  It cost a user many hours tracking this down even with help in Discord.  You can not access or delete the bucket.  Finally had them try do delete with SQL and their bucket name with trailing space and the bucket was deleted.

The UI should not allow invalid names for buckets.

## Steps to reproduce

Create a bucket with the name `cat-videos ` (note the trailing space)
Try to do anything with it in the UI after.

Note the UI also seems to take `cat videos` with the middle space and it works even though space is not a valid character according to:
image

## Expected behavior

The UI should not allow invalid bucket names.

ivasilov:
This PR fixes 2 issues:
- Adding new buckets with trailing whitespace (at start or end) are not possible anymore. Errors will be shown in the `Create new bucket modal`.
- Existing buckets with spaces in their names will be possible to be clicked and shown.


alaister (changes requested):
This works great for existing buckets, but I maybe it needs another `encodeURIComponent()` call after creating a new bucket? The redirect doesn't seem to including trailing spaces.

alaister:
Additionally, we should still consider the changes in https://github.com/supabase/supabase/pull/16869 to prevent future issues

joshenlim:
looks great! i do think we should nonetheless also consider this comment from Gary as well
https://github.com/supabase/supabase/issues/16641#issuecomment-1686643678

We should technically be following S3's guide here since it should be the same. Could we add some validations to prevent spaces in bucket names in CreateBucketModal?

joshenlim (approved):

alaister (approved):

nwaliaez:
```
const regex = /^[a-z0-9.-]+$/
  if (!regex.test(id)) {
    return res.status(400).json({ error: { message: 'Bucket name invalid' } })
  }
```
Above code is removed now I can add spaces in bucket name is that fine ? Also able to create bucket with name sdfq_-**(

PR #17096, opened 2023-09-01
Show warning when column types don't match

### LINKED ISSUE #14351: Table UI changes an int8/uuid column fk link to match the foreign tables column type, but then errors.
# Bug report

## To Reproduce

Use table UI to select a uuid/int column to be a foreign key link.
image
Select a column of the opposite type in the table dropdowns and the UI changes the column type to match.
image
Save the changes and get error.
image

Postgres can't change a bigint column to uuid or the opposite.  But the UI implies it is acceptable by changing the orginial column type.  This is causing confusion for users.   Note: this is not just for tables with data.

## Expected behavior
Either don't change column type at all or error up from with the selection.

## System information

Supabase hosted dashboard.

ivasilov:
This PR adds a warning and disables the `Save` button when the column types don't match during foreign column selection.
Fixes https://github.com/supabase/supabase/issues/14351


saltcod:
Looks good!  

Couple of things: 
If you don't name the column first, you get blank fields:
screenshot-2023-09-01-at-14 35 19

Let's use a consistent "is of type" or "has type" — I think "is of type" is prob preferred? 

Last sentence there maybe: "can't be referenced until they are of the same type"

ivasilov:
> Looks good!
> 
> Couple of things: If you don't name the column first, you get blank fields: screenshot-2023-09-01-at-14 35 19
> 
> Let's use a consistent "is of type" or "has type" — I think "is of type" is prob preferred?
> 
> Last sentence there maybe: "can't be referenced until they are of the same type"

Screenshots of the four variants:
Screenshot 2023-09-04 at 17 00 37
Screenshot 2023-09-04 at 17 00 44
Screenshot 2023-09-04 at 17 00 25
Screenshot 2023-09-04 at 17 00 51

wdyt?

saltcod (approved):

PR #17113, opened 2023-09-02
Fix tenant URL in Azure settings

aar2dee2:
Fixes auth issues reported in https://github.com/supabase/gotrue/pull/54

See comment: https://github.com/supabase/gotrue/pull/54#issuecomment-1702991208

docs update

Setting up Azure provider as per docs results in errors reported [here](https://github.com/supabase/gotrue/pull/54)

## What is the new behavior?

Use common tenant URL instead of custom tenant ID generated in the Azure application

## Additional context


apps/docs/pages/guides/auth/social-login/auth-azure.mdx
```diff
@@ -57,10 +57,7 @@ This will serve as the `client_secret` when you make API calls to authenticate t
 
 This will allow your users to use your custom Azure login page when logging in.
 
-- Select the Directory (Tenant) ID value.
-- The Azure Tenant URL should look like this: `https://login.microsoftonline.com/<tenant-id>`
-
-![Obtain the tenant url](/docs/img/guides/auth-azure/azure-tenant-url.png)
+Set the tenant URL in your Supabase Azure Provider settings to `https://login.microsoftonline.com/common`
```
kangmingtay, line 60: this isn't necessary because the tenant URL defaults to use the `/common` endpoint - https://github.com/supabase/gotrue/blob/25f2dcbc97bac18266f1d3583614656182154f85/internal/api/provider/azure.go#L12

apps/docs/pages/guides/auth/social-login/auth-azure.mdx
```diff
@@ -77,7 +74,7 @@ async function signInWithAzure() {
   const { data, error } = await supabase.auth.signInWithOAuth({
     provider: 'azure',
     options: {
-      scopes: 'email',
+      scopes: 'email', // this is important
```
kangmingtay, line 77: i don't think this change is necessary, the admonition above already calls it out

kangmingtay (changes requested):

apps/docs/pages/guides/auth/social-login/auth-azure.mdx
kangmingtay: yup that sounds good! we can also mention that it defaults to use the `/common` endpoint if it's not set

aar2dee2: great, I'll update it.

Is there a specific scenario in which the user should use the url with `tenant-id` and not `/common`?

kangmingtay: sflr, yes, there's a difference between using the tenant-id and common endpoint as mentioned in azure's docs: https://learn.microsoft.com/en-us/azure/active-directory/develop/howto-convert-app-to-be-multi-tenant#update-your-code-to-send-requests-to-common

aar2dee2: okay, so my understanding is `/<your-tenant-id>` is used when you want only users from a specific Azure organization to be able to sign in? 
For every other use case, we use `/common`

kangmingtay: yup that's right, in most cases you should be using the /common endpoint unless you want to restrict azure sign-ins to the users added to your tenant

aar2dee2: thanks. I've updated the docs.

kangmingtay: this is great, thanks for helping to update this

kangmingtay (approved):

kangmingtay:
@aar2dee2 can you please run prettier on the changes you made so it passes the CI?

aar2dee2:
> @aar2dee2 can you please run prettier on the changes you made so it passes the CI?

@kangmingtay 
done!

gregnr (approved):

Issue #16777, opened 2023-08-24
[Self Hosting] dependency failed to start: container supabase-vector is unhealthy


Andres6936:
# Bug report

<!-- 

⚠️ We receive a lot of bug reports which have already been solved or discussed. If you are looking for help, please try these first:

- Docs: https://docs.supabase.com
- Discussions: https://github.com/supabase/supabase/discussions
- Discord: https://discord.supabase.com

Before opening a bug report, please verify the following: 

-->

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Self Hosting fail to start when docker compose is used.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to Gidpot machine'…'
2. Choice the repository of supabase
3. Open the project of supabase
4. Navigate to docker directory
5. Copy the .env.example to .env
6. Run docker compose pull
7. Run docker compose up

## Expected behavior

Self Hosting start without problems

## Screenshots

```shellsession
gitpod /workspace/supabase/docker (master) $ docker compose up
WARN[0000] The "MFA_ENABLED" variable is not set. Defaulting to a blank string. 
[+] Running 13/12
 ✔ Network docker_default                    Created                                                                                                                                                                                                                              0.9s 
 ✔ Container supabase-vector                 Created                                                                                                                                                                                                                              0.9s 
 ✔ Container supabase-imgproxy               Created                                                                                                                                                                                                                              0.9s 
 ✔ Container supabase-db                     Created                                                                                                                                                                                                                              0.0s 
 ✔ Container supabase-analytics              Created                                                                                                                                                                                                                              0.1s 
 ✔ Container supabase-meta                   Created                                                                                                                                                                                                                              0.1s 
 ✔ Container supabase-auth                   Created                                                                                                                                                                                                                              0.1s 
 ✔ Container realtime-dev.supabase-realtime  Created                                                                                                                                                                                                                              0.1s 
 ✔ Container supabase-kong                   Created                                                                                                                                                                                                                              0.1s 
 ✔ Container supabase-rest                   Created                                                                                                                                                                                                                              0.1s 
 ✔ Container supabase-edge-functions         Created                                                                                                                                                                                                                              0.1s 
[...]

PR #17122, opened 2023-09-02
chore: Move logging into Docker logs only

filipecabaco:

Change the Vector ingestion to use docker logs directly from the docker socket instead of syslog via an open port


kiwicopple:
possible fix for #16777

kiwicopple (approved):

kiwicopple:
fixes #16777

PR #17132, opened 2023-09-03
Update getting-started.mdx because of dangerous default

activenode:
I somehow cannot relate why the default to stop containers would be `--no-backup` . Many people copy paste scripts from the official docs.

Docs Update


activenode:
thanks

Issue #14164, opened 2023-05-05
Error when trying to upload files to Storage thru the Studio [Self-hosted]


kaibeckers:
# Studio cannot upload files to Storage [Self-hosted]

- [x] I confirm this is a bug with Supabase, not with my application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Uploading a file to a storage bucket in the Studio will not work. An "ERR_CONN_REFUSED" error will appear in the console. This error isn't forwarded to the frontend GUI aswell: the user won't see the upload has failed.

Creating folders doesn't work either: they appear, but once you refresh, they disappear. 

 All tables in the 'storage' schema can be accessed, as evident by the fact I can create a bucket & there are migrations, except for the 'objects' tables.

## Workaround

As of right now, using the Supabase SDK (for NodeJS) works fine, here's an example;

```js
const filePath = './source/images/adecal';
const fileData = fs.readFileSync(filePath);

supabase.storage
  .from('abucket')
  .upload('adecal.png', fileData, {
    contentType: 'image/png'
  })
```

## To Reproduce

1. Create a VPS that is running on Ubuntu 22.04, with Docker & Docker-Compose installed
2. Follow the self-hosting guide: clone repo, set all ENV's, etc.
3. Enter the Studio
4. Create a bucket
5. Try to upload a file or create a folder

## Expected behavior

The image is uploaded and there are no errors in the console.

The folder will stay and not disappear once refreshed.

## Screenshots
<img width="358" alt="Schermafbeelding 2023-04-27 om 16 19 06" src="https://user-images.githubusercontent.com/53176455/234891288-79492ce7-b36f-48a5-95b3-6f7660e7d432.png">
<img width="921" alt="Schermafbeelding 2023-04-27 om 16 19 20" src="https://user-images.githubusercontent.com/53176455/234891315-5e9f2eb6-12ff-4554-9a27-ae115ff5f6a4.png">

https://user-images.githubusercontent.com/53176455/234891615-1b8b8dfb-a7d7-412f-8c76-e6c831e91592.mov

## System information

- OS: Ubuntu 22.04
- Docker Version: v23.0.4, build f480fb1
- D-Compose Version: v2.17.3

- Chrome
- macOS 12.6.5

## Additional context

Other features like Authentication are broken as well: fetching errors (?)

[...]

PR #17145, opened 2023-09-04
Don't fetch custom domains when selfhosting

ivasilov:
This PR adds a check when fetching custom domains. The query should only be made when the site is in platform mode.

This PR fixes an issue described in [this comment](https://github.com/supabase/supabase/issues/14164#issuecomment-1699716878).


alaister (approved):
LGTM!

PR #17151, opened 2023-09-04
fix(www): className on Img components

gregnr:
For `` components, some classes in the `className` prop are getting overwritten by styles in the `style` attribute (maybe coming from `react-medium-image-zoom`?). Specifically in blog posts we are unable to show/hide images based on light/dark mode (via tailwind `className`).

Eg.
```jsx
<div>
  <Img
    alt="my light mode image"
    className="dark:hidden"
    src="/images/blog/my-image--light.png"
  />
  <Img
    alt="my dark mode image"
    className="hidden dark:block"
    src="/images/blog/my-image--dark.png"
  />
</div>
```

This change applies the `` `className` prop to the top level `<figure />` instead of the Next.js `<Image />`.

### DIFF
--- a/apps/www/lib/mdx/mdxComponents.tsx
+++ b/apps/www/lib/mdx/mdxComponents.tsx
@@ -12,6 +12,7 @@ import {
   CollapsibleTrigger_Shadcn_,
   CollapsibleContent_Shadcn_,
   IconTriangle,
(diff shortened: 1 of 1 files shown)


fsansalvadore (approved):
thanks @gregnr  👍

PR #17193, opened 2023-09-06
Only show first 3 roles in each policy if theres more than 3 roles applied

joshenlim:
image


alaister (approved):
Nice! LGTM!

alaister:
Possible improvement: hover the `+ 3 more roles` thing for a tooltip that lists the extra roles

alaister (approved):
double approve 😎

PR #17198, opened 2023-09-06
Enforce CNAME record check prior to activating custom domain

joshenlim:


studio/components/interfaces/Settings/General/CustomDomainConfig/CustomDomainConfig.utils.ts
```diff
@@ -0,0 +1,9 @@
+// [Joshen] Should tally with https://github.com/supabase/cli/blob/63790a1bd43bee06f82c4f510e709925526a4daa/internal/utils/api.go#L98
+export const verifyCNAME = async (domain: string): Promise<boolean> => {
```
ivasilov, line 2: small nit: you should wrap this function into a `react-query` and turn it into a hook. You're using the same abstractions (`isActivated`, `onSuccess`, `onError`) already.

joshenlim: Ohh good point actually - will do that

studio/components/interfaces/Settings/General/CustomDomainConfig/CustomDomainConfig.utils.ts
```diff
@@ -0,0 +1,9 @@
...
+  const res = await fetch(`https://1.1.1.1/dns-query?name=${domain}&type=CNAME`, {
+    method: 'GET',
+    headers: { accept: 'application/dns-json' },
+  })
+  const verification = await res.json()
+  return verification.Status === 0
```
alaister, line 8: Is this check sufficient? For example I have NOT added a cname record for `supabase-v2-fly.alaister.dev`, but I still get the following response:
```json
{
	"Status": 0,
	"TC": false,
	"RD": true,
	"RA": true,
	"AD": true,
	"CD": false,
	"Question": [
		{
			"name": "supabase-v2-fly.alaister.dev",
			"type": 5
		}
	],
	"Authority": [
		{
			"name": "alaister.dev",
			"type": 6,
			"TTL": 1739,
			"data": "fay.ns.cloudflare.com. dns.cloudflare.com. 2319479414 10000 2400 604800 1800"
		}
	]
}
```

alaister (approved):
LGTM!

PR #17213, opened 2023-09-06
Add loading state to the MFA page

ivasilov:
This PR adds a loading state to the MFA page. Previously, every sign-in redirected to MFA page where the user was tested whether he has added any TOTP factors. This showed a minor flicker where the text says "Enter your TOTP code" which was confusing to users who didn't have enabled MFA.


alaister (approved):
LGTM!

PR #17216, opened 2023-09-06
Add better empty state if no org

saltcod:
A better empty state if user has no org and no projects. 

This PR creates this state under these ^ conditions:
screenshot-2023-09-06-at-20 37 48

**To test:** 
— create a new org & a new project
— delete the project 
— delete the org
— see this new empty state

### DIFF
--- a/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
+++ b/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
@@ -24,7 +24,7 @@ export interface ProjectListProps {
 }
 
 const ProjectList = ({ rewriteHref }: ProjectListProps) => {
-  const { data: organizations } = useOrganizationsQuery()
+  const { data: organizations, isSuccess } = useOrganizationsQuery()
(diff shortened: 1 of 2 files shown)


studio/components/interfaces/Home/ProjectList/ProjectList.tsx
```diff
@@ -40,7 +40,7 @@ const ProjectList = ({ rewriteHref }: ProjectListProps) => {
   const projectsByOrg = groupBy(allProjects, 'organization_id')
   const isLoadingPermissions = IS_PLATFORM ? _isLoadingPermissions : false
 
-  return (
+  return organizations && organizations?.length > 0 ? (
```
alaister, line 43: Can we use `isSuccess` from the query here so we don't show `<NoProjectsState />` while it's loading? It might be okay because the loading state could be handled further up, but nice for consistency either way 😄

kevcodez (approved):
Tested and it looks good 👍

PR #17225, opened 2023-09-07
fix: RLS docs, add better explanation of `using` and `with check` expressions

### LINKED ISSUE #17133: Potential mistake for sections under `RLS` -> `Creating Policies` documentation
# Improve documentation

## Link 

https://supabase.com/docs/guides/database/postgres/row-level-security#creating-policies

## Describe the problem

The wording and the example actions are confusing and inconsistent.

## Describe the improvement

Under `SELECT policies`, it correctly says: `You can specify select policies with the using clause.` along with the correct example.

Under `INSERT policies`, it says `You can specify select policies with the with check clause.`, when I believe it should say `You can specify insert policies...`. Furthermore, the example below it doesn't use the `with check` clause but uses the `using` clause:
```sql
-- 1. Create table
create table profiles (
  id uuid primary key,
  user_id references auth.users,
  avatar_url text
);

-- 2. Enable RLS
alter table profiles enable row level security;

-- 3. Create Policy
create policy "Users can create a profile."
on profiles for insert
using ( auth.id() = user_id );      -- the actual Policy
```

Under the `UPDATE policies` it says `You can specify select policies with the using clause.` when I believe it should say `... update policies ...`. I also believe it should  mention the use of the `with check` clause with a relevant example.

Same for the `DELETE policies` subsection.

## Additional context

I'm relatively new to Supabase, so I'm not sure of my assumptions here. I'm willing to send a PR with the fix if someone can confirm they are correct.

kishanhitk:

 docs update

Fixes: #17133, #17299
Some of the RLS examples are not correct, and they need a better explanation, for things like `using` and `with check`.
## What is the new behavior?

Correct RLS examples with more details.

## Additional context


gregwolanski (approved):

apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -84,9 +84,13 @@ on profiles
 for select using ( auth.uid() = user_id );
 ```
 
+[comment]: <> (TODO: Improve the RLS docs)
+
```
charislam, line 88:
```suggestion
```

Let's not leave comments in here for readability, if there are outstanding issues to address we can make an issue

apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -129,12 +136,14 @@ alter table profiles enable row level security;
 create policy "Users can update their own profile."
 on profiles for update
 to authenticated                    -- the Postgres Role (recommended)
-using ( auth.id() = user_id );      -- the actual Policy
+using ( auth.id() = user_id )       -- Policy to check the current user's right to update the row.
+with check ( auth.id() = user_id ); -- Policy to ensure the resultant row still complies with the ownership condition.
```
kangmingtay, line 140:
```suggestion
with check ( auth.id() = user_id ); -- checks if the new row complies with the policy expression
```

apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -129,12 +136,14 @@ alter table profiles enable row level security;
 create policy "Users can update their own profile."
 on profiles for update
 to authenticated                    -- the Postgres Role (recommended)
-using ( auth.id() = user_id );      -- the actual Policy
+using ( auth.id() = user_id )       -- Policy to check the current user's right to update the row.
```
kangmingtay, line 139:
```suggestion
using ( auth.id() = user_id )       -- checks if the existing row complies with the policy expression
```

apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -105,15 +109,18 @@ alter table profiles enable row level security;
...
+with check ( auth.id() = user_id ); -- the actual Policy
 ```
 
 ### UPDATE Policies
 
-You can specify select policies with the `using` clause.
+You can specify update policies by combining both the `using` and `with check` expressions.
 
-Let's say you have a table called `profiles` in the public schema and you only want users to be able to update their own profile:
+The `using` represents the condition that must be true for the update to be allowed, and `with check` ensures that the updates made adhere to the policy constraints.
```
kangmingtay, line 119:
```suggestion
The `using` clause represents the condition that must be true for the update to be allowed, and `with check` clause ensures that the updates made adhere to the policy constraints.
```

charislam (approved):

PR #17245, opened 2023-09-07
Fix/eu vat

saltcod:
Fixes issue with eu_vat taxIDs not saving correctly


kevcodez (approved):
Awesome, thanks for tackling!

PR #17259, opened 2023-09-08
Support search storage in column view

joshenlim:


alaister (approved):
Works great!
Possible improvement: show some loading state while searching (columns view)

PR #17260, opened 2023-09-08
Improvement of Postgresql RLS docs

### LINKED ISSUE #17133: Potential mistake for sections under `RLS` -> `Creating Policies` documentation

torm89:

Docs update

Fixes: https://github.com/supabase/supabase/issues/17133

## What is the new behavior?

Docs is better.

## Additional context

...


apps/docs/pages/guides/database/postgres/row-level-security.mdx
```diff
@@ -105,12 +105,12 @@ alter table profiles enable row level security;
 create policy "Users can create a profile."
 on profiles for insert
 to authenticated                    -- the Postgres Role (recommended)
-using ( auth.id() = user_id );      -- the actual Policy
+with check ( auth.id() = user_id );      -- the actual Policy
```
fernandolujan, line 108: I'm not sure that `auth.id()` is actually valid here

I ran into this issue while walking through the docs. Should be `auth.uid()` instead

fernandolujan (changes requested):

gregnr (changes requested):

apps/docs/pages/guides/database/postgres/row-level-security.mdx
torm89: Done.

gregnr (approved):
Nice work! 🎉

PR #17263, opened 2023-09-08
feat: local query performance report

Ziinc:
This PR adds local query performance rendering, and is the first step to enabling reports on cli/self-hosted.

It also adds some test coverage for database reports.


Ziinc:
closing for now.

PR #17282, opened 2023-09-08
chore: remove new label from reports menu

Ziinc:
Small tweak to remove the NEW label from reports menu items

Screenshot 2023-09-09 at 2 56 20 AM


Ziinc:
merging into storage reports PR

PR #17288, opened 2023-09-08
chore: add storage cache miss template to logs explorer

Ziinc:
This PR adds in storage cache hit template to the logs explorer, as a supplement to debugging low cache rate.


alaister (approved):
LGTM!

The indentation seems to be one too far in – not sure if this is something to actually be concerned about though 😅
Screenshot 2023-09-14 at 4 53 29 pm

PR #17291, opened 2023-09-09
Fix default value of boolean column not respecting

### LINKED ISSUE #17284: Default value of boolean column not respected
Setting true or false for a boolean column isn't respected. Always resets to null.

To reproduce: 
1. create a new table
2. create a new column, boolean as the type 
3. enter true/false/TRUE/FALSE as the default
4. enter a new row, find that the field is pre-filled as NULL not true/false. 
screenshot-2023-09-08-at-16 25 42

samueldenzil:

Bug fix

Changed `field.value` to `field.defaultValue` because when the user clicks on the save button, we change the defaultValue inside the "Add new column to `database_name`" page.

Setting true or false for a boolean column isn't respected. Always resets to null.

Fixes #17284 

## What is the new behavior?
Works with true/false/TRUE/FALSE

[Screencast.webm](https://github.com/supabase/supabase/assets/71846487/f2386beb-cfd2-4f1b-a322-87dd04f68c12)

### DIFF
--- a/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/InputField.tsx
+++ b/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/InputField.tsx
@@ -202,7 +202,7 @@ const InputField = ({
       ...(field.isNullable ? [{ value: 'null', label: 'NULL' }] : []),
     ]
 
-    const defaultValue = field.value === null ? 'null' : field.value
+    const defaultValue = field.defaultValue === null ? 'null' : field.defaultValue
 
     return (
       <Listbox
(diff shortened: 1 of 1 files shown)


studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/InputField.tsx
```diff
@@ -202,7 +202,7 @@ const InputField = ({
       ...(field.isNullable ? [{ value: 'null', label: 'NULL' }] : []),
     ]
 
-    const defaultValue = field.value === null ? 'null' : field.value
+    const defaultValue = field.defaultValue === null ? 'null' : field.defaultValue
```
saltcod, line 205:
```suggestion
const defaultValue = field.defaultValue !== null ? field.defaultValue : 'null'
```

ivasilov (approved):

PR #17293, opened 2023-09-09
Recommend Docker Desktop alternatives

psteinroe:

docs update

no information on how to use docker 

## What is the new behavior?

since OrbStack has been a life-changer for me as a avid cli user, and I keep recommending it to people, it might be a good addition to the docs. not sure if supabase wants to recommend other software here, and, if so, whether this is the right place for it though.

### DIFF
--- a/apps/docs/pages/guides/cli/local-development.mdx
+++ b/apps/docs/pages/guides/cli/local-development.mdx
@@ -62,7 +62,19 @@ git init
 supabase init
 ```
 
-Make sure Docker is running. On Mac, we recommend [OrbStack](https://orbstack.dev) to run docker on your machine. On other operating systems, install [Docker Desktop](https://www.docker.com/products/docker-desktop/). The [start](/docs/reference/cli/usage#supabase-start) command uses Docker to start the Supabase [services](/docs/guides/getting-started/architecture).
+Make sure Docker is running.
+
+<Admonition type="note">
+
+Docker is available via
+- [Docker Desktop](https://docs.docker.com/get-docker/) (macOS, Windows, Linux)
+- [Rancher Desktop](https://rancherdesktop.io/) (macOS, Windows, Linux)
+- [OrbStack](https://orbstack.dev/) (macOS)
+- [colima](https://github.com/abiosoft/colima) (macOS)
+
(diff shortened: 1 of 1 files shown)


gregnr:
Thanks @psteinroe! OrbStack is great, but I'd prefer to stay unopinionated in the docs.

I think there is still an opportunity for improvement though. Perhaps we add a callout (`<Admonition>`) after that line that directs users to places they can download Docker for development:
- [Docker Desktop](https://docs.docker.com/get-docker/) (macOS, Windows, Linux)
- [Rancher Desktop](https://rancherdesktop.io/) (macOS, Windows, Linux)
- [OrbStack](https://orbstack.dev/) (macOS)
- [colima](https://github.com/abiosoft/colima) (macOS)

psteinroe:
> Thanks @psteinroe! OrbStack is great, but I'd prefer to stay unopinionated in the docs.
> 
> I think there is still an opportunity for improvement though. Perhaps we add a callout (`<Admonition>`) after that line that directs users to places they can download Docker for development:
> 
> * [Docker Desktop](https://docs.docker.com/get-docker/) (macOS, Windows, Linux)
> * [Rancher Desktop](https://rancherdesktop.io/) (macOS, Windows, Linux)
> * [OrbStack](https://orbstack.dev/) (macOS)
> * [colima](https://github.com/abiosoft/colima) (macOS)

sorry the the delay! made the changes now.

apps/docs/pages/guides/cli/local-development.mdx
```diff
@@ -62,7 +62,19 @@ git init
 supabase init
 ```
 
-Make sure Docker is running. The [start](/docs/reference/cli/usage#supabase-start) command uses Docker to start the Supabase [services](/docs/guides/getting-started/architecture).
+Make sure Docker is running.
+
+<Admonition type="note">
+
+Docker is available via
```
gregnr, line 69:
```suggestion
There are a number of different projects available to download Docker from:
```

psteinroe:
@gregnr format is fixed, I think its ready now. thanks for your wording improvements!

awalias (approved):

PR #16097, opened 2023-07-28
Chore: self hosting improvements

kiwicopple:
## What kind of change does this PR introduce?

- [x] Hides the Dashboard behind Kong
- [x] Adds Basic Auth to the Dashboard (through Kong)
- [x] Removes the caddy instructions - much simpler just with Kong
- [x] Adds more instructions to the self-hosting guide
- [x] Verify all services working as expected
- [x] Add instructions for viewing the logs 

Docs Preview: https://docs-git-chore-self-hosting-improvements-supabase.vercel.app/docs/guides/self-hosting/docker

- Partially addresses: 
  - #12963
- Fixes:
  - #15553
  - #14212 
  - #13064
  - #12560
  - #10319
  - #16153
  - #16122
  - #14222
  

As part of this PR I have added a video demonstration to YT which should hopefully make the steps as clear as possible:

https://youtu.be/FqiQKRKsfZE

PR #17306, opened 2023-09-10
Update docker-compose.yml

xaddison:
logflare run cmd not able to find run.sh and issues with parsing EOF method - workaround

Bug fix

analytics container will not start 

[#16097](https://github.com/supabase/supabase/pull/16097)

## What is the new behavior?

container is running successfully supabase/analytics (other dependent containers are also starting now, still an issue with supabase/auth) 

## Additional context

This was blocking most of the other containers from starting via docker-compose up as they depend_on this container. 

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -337,12 +337,16 @@ services:
     ports:
       - 4000:4000
     entrypoint: |
-                  sh -c `cat <<'EOF' > run.sh && sh run.sh
-                  ./logflare eval Logflare.Release.migrate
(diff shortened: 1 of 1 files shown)


filipecabaco:
Hi @xaddison 

I tested locally and everything worked as intended with the old code. Can you please check if you have anything in `docker/volumes/db/data`? If so and if you can try to remove that entire folder as that retains the docker volumes for the database.

The description of the issue could be related with some old state from previous docker versions but want to double check.

Ziinc (changes requested):

PR #17319, opened 2023-09-11
Use custom domain if active in API docs

joshenlim:


studio/components/to-be-cleaned/Docs/Snippets.ts
```diff
@@ -1,5 +1,5 @@
 const snippets = {
```
alaister, line 1: Nice move to TS! Should this file now be moved out of `to-be-cleaned` or is a bigger refactor planned?

alaister (approved):
LGTM!

studio/components/to-be-cleaned/Docs/Snippets.ts
joshenlim: yeah we can probably move this file out of to-be-cleaned, i'm gonna look into converting more JS files to TS in the Docs folder first before i move them

PR #17333, opened 2023-09-11
Fix/pooler settings

saltcod:


studio/components/interfaces/Settings/Database/ConnectionPooling.tsx
```diff
@@ -140,7 +140,12 @@ export const PgbouncerConfig = ({ projectRef, bouncerInfo, connectionInfo }: Con
 
   const canUpdateConnectionPoolingConfiguration = useCheckPermissions(
     PermissionAction.UPDATE,
-    'projects'
+    'projects',
+    {
+      resource: {
+        project_id: projectRef,
```
kevcodez, line 146: Shouldn't this be the projects id rather than the ref? Looking at other usages of useCheckPermissions it looks like we're passing the id (numeric) instead of the ref

joshenlim: this should be the project id

joshenlim (approved):

PR #17341, opened 2023-09-11
SQL Editor: CLI commands to import snippet as migration/seed

gregnr:
### 🔍 Preview
https://studio-staging-git-feat-sql-editor-cli-import-supabase.vercel.app/dashboard/project/_/sql/new

 
Adds new button "Download as migration file" to SQL editor in the snippet context menu:

 screenshot-2023-11-16-at-14 49 37

Clicking the button brings up a modal with pre-generated commands to import the current snippet as a migration, seed, or SQL file:
screenshot-2023-11-16-at-14 49 57
screenshot-2023-11-16-at-14 49 49


sweatybridge (approved):

studio/components/interfaces/SQLEditor/UtilityPanel/UtilityActions.tsx
```diff
@@ -23,10 +26,31 @@ const UtilityActions = ({
...
+      )}
       <SavingIndicator id={id} />
+      {IS_PLATFORM && snippet && (
+        <Button
+          type="text"
+          size="tiny"
+          icon={<IconDownload size="tiny" />}
+          onClick={() => setIsDownloadSnippetModalOpen(true)}
+        >
+          Download as migration file
```
joshenlim, line 51: i'd opt to put "Download as migration file" behind a tooltip to prevent clutter here

same considerations for table editor header - lets keep things neat and clean while we still can

joshenlim: Can i also check if this is a heavily requested feature? Alternatively we put this in the snippet's dropdown in the left side nav

joshenlim:
This is just future consideration - but we could consider leveraging the new API side panel for this too
have a section for SQL editor
clicking "Download as migration" opens the API side panel with that section open

joshenlim:
was also looking at the thread - i was thinking the same thing with Terry RE tabs, could look something like this
image

language selection as a dropdown on the side
main consideration is that users are probably looking for one thing at a time, its a bit of a spread showing all three types at once

studio/components/interfaces/SQLEditor/UtilityPanel/UtilityActions.tsx
gregnr: Good points @joshenlim - how about we put it under the snippet's left side dropdown for now and we can make it more visible in the future if there's lots of usage

saltcod:
Hey guys – made some updates here. Sorry for the delay.

1. moved the button to the context menu for ui simplicity 
2. changed the layout to a tab, avoiding tab-ception via the toggle to the right 

no major changes otherwise
screenshot-2023-11-13-at-11 25 19

screenshot-2023-11-13-at-11 25 13

apps/studio/components/layouts/SQLEditorLayout/QueryItem.tsx
```diff
@@ -155,6 +158,8 @@ const QueryItemActions = observer(({ tabInfo, activeId }: QueryItemActionsProps)
     setDeleteModalOpen(true)
   }
 
+  const onClickDownloadMigration = () => {}
```
joshenlim, line 161: @gregnr missing implementation here?

apps/studio/components/interfaces/SQLEditor/DownloadSnippetModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+              }
+            />
+          </div>
+
+          <Tabs type="underlined">
+            <Tabs.Panel id="migration" label="Migration">
+              <div className="flex flex-col gap-2 my-2 w-full">
+                <h2 className="text-lg">Download as migration</h2>
+                <p className="text-sm text-scale-1000">
+                  Use the snippet in a new migration named{' '}
```
joshenlim, line 70: just wondering - should we also just state where to run this command in? help the user with less thinking

also, am not too sure what the outcome of this command is but is `use` the right verb here? should it be `download`? "Download the snippet into a new migration"?

apps/studio/components/interfaces/SQLEditor/DownloadSnippetModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+                    {generateMigrationCliCommand(id, migrationName, true)}
+                  </CodeBlock>
+                </pre>
+              )}
+            </Tabs.Panel>
+            <Tabs.Panel id="seed" label="Seed file">
+              <div className="flex flex-col gap-2 my-2 w-full">
+                <h2 className="text-lg">Download as seed file</h2>
+                <p className="text-sm text-scale-1000">
+                  Alternatively if your query consists of sample data, append the snippet to the end
```
joshenlim, line 97: am just wondering if this should make the outcome clearer

```suggestion
                  Alternatively if your query consists of sample data, run the following command to append the snippet to the end
```

joshenlim: I don't think we should add "Alternatively" here either - we should make each tab mutually exclusive i think language wise, although I get what we're trying to do here where the user starts from the first tab, then goes to the second

apps/studio/components/interfaces/SQLEditor/DownloadSnippetModal.tsx
saltcod: 1. changed 'use' to 'download'. I think that's right. 
2. added links at the bottom here. Played with adding helper text above the tabs there as well, but felt like it was getting in the way. 
Do you think this is enough? The "learn more about" should convey:  "this is for local development", which may or may not be pertinent to you. Does it convey that well enough to you? 

screenshot-2023-11-20-at-10 24 34

saltcod: screenshot-2023-11-20-at-10 34 10

apps/studio/components/interfaces/SQLEditor/DownloadSnippetModal.tsx
```diff
@@ -141,7 +120,43 @@ const DownloadSnippetModal = ({ id, ...props }: DownloadSnippetModalProps) => {
                 </CodeBlock>
               )}
             </Tabs.Panel>
+            <p className="text-xs text-lighter mt-4 text-right mr-6 mb-4">
+              * Run from your local <code>/supabase</code> directory
```
gregnr, line 124: This should be project root right? Otherwise we should update each command to remove the `./supabase` prefix.

joshenlim (approved):

PR #17349, opened 2023-09-12
Edge functions use custom domain if configured

joshenlim:


alaister (approved):
LGTM!

Possible improvement: could also show the custom domain here:
Screenshot 2023-09-12 at 5 34 13 pm

PR #17354, opened 2023-09-12
Ensure settings infrastructure page is accessible irregardless or project

joshenlim:


alaister (approved):
Works great!

Something I noticed was the default state is error and not loading on this page:
Screenshot 2023-09-12 at 9 23 55 pm
(don't have to fix here – it's just a flash after a hard refresh)

kevcodez (approved):

PR #17356, opened 2023-09-12
Provide convenient unit selection for custom expiry in storage

joshenlim:
image


alaister (approved):
LGTM!

Seems a bit strange though to have seconds and then jump straight to days – minutes and hours seem like a useful option?

joshenlim:
hahaha actually fair that hmm, lemme relook at the UX in a bit. Maybe seconds shouldn't even be an option - no one would practically wanna count in seconds for an expiry

joshenlim:
okay switched to days, weeks, months, years and the helper text below will show when the url will expire

PR #17383, opened 2023-09-13
fix/ correct hipaa description in pricing page

pamelachia:

Ensure it is clearer in the pricing page that HIPAA is a paid add on.


darora (approved):
lgtm but pending FE approval

PR #17387, opened 2023-09-13
Docs: update supabase_js_v2.yml to add example on using `!inner`

bk7312:

Minor update to the `supabase_js_v2.yml` documentation

In the `Filtering through foreign tables` example under `Fetch data` in the `Database` section, there was a note to use the `!inner` hint to filter out the parent table rows, but no example or info on how to use it. 

If a user finds the note on using the `!inner` hint, it may not be obvious how to use it or where to find more info about it.

inner

## What is the new behavior?

Added an example on how to use the `!inner` hint and made a reference to the `Filter Foreign Tables` example under `Using filters`.


steve-chavez:
FYI, similar change on https://github.com/supabase/supabase/pull/17633

PR #17391, opened 2023-09-13
added small gaps in between the framework icons on the docs get started page

AbhijeetBiswal:
#17491
current behaviour : there is no gaps between the framework icons on pc screen 

current_behaviour 

new behaviour : added small gaps in between the framework icons now its look perfect 

new_behaviour


JSaretin:
a lot nicer :+1:

fsansalvadore:
Thanks @AbhijeetBiswal but this was fixed in this PR: #17540

PR #17394, opened 2023-09-13
feat: use a persisted auth debug log if `supabase.dashboard.auth.debug.persist` is true

hf:
Bumps `@supabase/gotrue-js` to the latest version which allows a custom debug function to be set. If the local storage key `supabase.dashboard.auth.debug.persist` is set to `true`, then a `auth-debug-log` indexed DB will be opened and all debug messages will be persisted in the `events` object store. This can be used to inspect the behavior of multi-tab and multi-window scenarios.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14121,17 +14121,17 @@
       }
     },
     "node_modules/@supabase/gotrue-js": {
-      "version": "2.48.0",
-      "resolved": "https://registry.npmjs.org/@supabase/gotrue-js/-/gotrue-js-2.48.0.tgz",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -68,5 +68,8 @@
   ],
   "volta": {
     "node": "18.17.1"
+  },
+  "dependencies": {
+    "@supabase/gotrue-js": "^2.52.0"
```
hf, line 73: This is probably a bad idea.

ivasilov (approved):

PR #17399, opened 2023-09-13
Fix/cardbutton classes

saltcod:
We removed the px-6 class from the CardButton component here: 

https://github.com/supabase/supabase/commit/17d4445ce70cd2a4870d12fa1b88ca90de707fc5#diff-bae01e5bc89aa196bcefbcb38b044c9de7499680f467bd6dfae1523b2efb7476L50

But we still need this, conditionally on some instances. 

This PR passes className down to the CardButton component to handle this. 
Since we were using someting similar (containerHeightClassName) I changed it to just use className.

**Places to check:** 

/dashboard/project/_/sql/quickstarts
/dashboard/project/_/logs/explorer/templates
/dashboard/org/_/general


alaister (approved):
LGTM!

PR #17422, opened 2023-09-14
Fix project cards not showing paused state

joshenlim:


alaister (approved):
LGTM!

styling q: does the hover state feel 'correct' to you? I feel like it's strange the whole card doesn't react. wdyt?
Screenshot 2023-09-14 at 10 16 54 pm

joshenlim:
> does the hover state feel 'correct' to you? I feel like it's strange the whole card doesn't react. wdyt?

hmmm yeah you're right really - we'll need to look into this separately

PR #17458, opened 2023-09-15
Improve support for viewing data in table editor for readonly entities

joshenlim:


alaister (changes requested):
Almost hit approve but found one bug: I can no longer edit inline JSONB columns on tables:
Screenshot 2023-09-15 at 10 57 54

Looking 🔥 otherwise

joshenlim:
@alaister great catch 🙏 just pushed a fix!

alaister (approved):
LGTM!

PR #17460, opened 2023-09-15
Add realtime toggle button in table editor

joshenlim:
image
image


studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -75,62 +86,134 @@ const GridHeaderActions = ({
...
+        buttonLabel={`${isRealtimeEnabled ? 'Disable' : 'Enable'} realtime`}
+        buttonLoadingLabel={`${isRealtimeEnabled ? 'Disabling' : 'Enabling'} realtime`}
+        onSelectCancel={() => setShowEnableRealtime(false)}
+        onSelectConfirm={() => toggleRealtime()}
+      >
+        <Modal.Content className="py-4 space-y-2">
+          <p className="text-sm">
+            Any changes on this table will {isRealtimeEnabled && 'no longer '}be broadcasted to
+            authorized subscribers once realtime has been{' '}
+            {isRealtimeEnabled ? 'disabled' : 'enabled'}.
```
alaister, line 203: active voice 😉
(prettier required)
```suggestion
            Once realtime has been {isRealtimeEnabled ? 'disabled' : 'enabled'}, the table will {isRealtimeEnabled && 'no longer '}broadcast any changes to authorized subscribers.
```

alaister (approved):
LGTM!

one suggestion:

saltcod:
🤔  adds a bit of extra clutter to an already busy area. 
Ideas: 
— for this one specifically — wonder would just the icon + tooltip be enough? 
— more broadly, wonder if we could go with icons + tooltips for all functionality in that area — RLS / realtime / api / definition view

joshenlim:
> wonder would just the icon + tooltip be enough?

my concern would be that user's will find it a hassle to figure out what the button does without having to hover over it to see the tooltip (just like the side nav, where there's a handful of users requested for either texts with the icons or an expandable side nav). in particular, these functionalities don't necessarily have icons which are instinctive to them (e.g Zap for realtime vs lock for auth, the latter is more natural perhaps)

PR #17528, opened 2023-09-18
fix:#17512

### LINKED ISSUE #17512: Layout Shift on Supabase homepage Login Button gets annoying
# Bug report

## Describe the bug

Not so much a bug but an improvement request: 
When you're already logged in to supabase, if you head to supabase.com you'll get a flash of unauthenticated content for about 1s (In Australia at least) which is just enough time to click on the "Login" problem is, this often is the perfect time for it to switch to authenticated and you accidentally click on the "Star us on GitHub" button 

Nice feature to get more stars, but super annoying otherwise. Annoying enough for me to open a bug report 

## To Reproduce

1. Go to supabase.com
2. Click on login 
3. Sucked in you hit the GitHub link by accident 
4. ...
5. Swearing practice 

## Expected behavior

Just move the auth'd interface to align with the un-authed interface so buttons don't move 

## Screenshots

NA

## System information

any browser

yashnirmal:
Adding a new `useIsUserLoading` hook in `/packages/common/auth.tsx`
Screenshot 2023-09-18 at 9 49 57 PM
urns false)
and adding a condition in `appa/www/components/Nav/index.tsx`, to conditionally render the singin/up button when the user is loaded (i.e. `useIsUserLoading` hook ret

Issue fix #17512 
the button and links on the right side of the navigation were getting shifted after `user` state was changing from null to not null

## What is the new behavior?
the button and links on the right side of the navigation are rendered once the `user` state is fully loaded

### DIFF
--- a/apps/www/components/Nav/index.tsx
+++ b/apps/www/components/Nav/index.tsx
@@ -13,7 +13,7 @@ import Developers from '~/components/Nav/Developers'
 
 import ScrollProgress from '~/components/ScrollProgress'
 
-import { useIsLoggedIn, useTheme } from 'common'
+import { useIsLoggedIn, useIsUserLoading, useTheme } from 'common'
 import TextLink from '../TextLink'
 import Image from 'next/image'
 import * as supabaseLogoWordmarkDark from 'common/assets/images/supabase-logo-wordmark--dark.png'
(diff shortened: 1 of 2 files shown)


fsansalvadore (approved):
Ideally we would also want some loading state UI, but we'd need to provide designs for that. 
So good enough for now.

Thanks @yashnirmal

PR #17542, opened 2023-09-19
Filter favourited items from list of snippets

saltcod:
Filters favourited snippets from your personal list so they're not duplicated. 
Also hides the "no queries found" message if you have favourited queries to avoid that shock.

screenshot-2023-09-18-at-21 51 56

screenshot-2023-09-18-at-21 51 08


alaister (approved):
LGTM!

PR #17553, opened 2023-09-19
feat: disable features

alaister:

feature

## What is the new behaviour?

Features can be disabled as needed


joshenlim:
probably need to add billing:all check to `<UpgradeToPro>` as well?

studio/__mocks__/hooks.js
```diff
@@ -8,6 +8,8 @@ export const useParams = jest.fn().mockImplementation(() => ({ ref: '123' }))
 
 export const useLocalStorage = jest.fn().mockImplementation(() => [[], jest.fn()])
 
+export const useIsFeatureEnabled = jest.fn().mockImplementation(() => [])
```
kevcodez, line 11: Should we actually check the arg and return `true` if a string is passed in an `{}` if an array is passed in? Never returns a boolean does it?

studio/data/projects/project-restore-mutation.ts
```diff
@@ -11,7 +11,6 @@ export type ProjectRestoreVariables = {
 export async function restoreProject({ ref }: ProjectRestoreVariables) {
   const { data, error } = await post('/platform/projects/{ref}/restore', {
     params: { path: { ref } },
```
kevcodez, line 13: Just double-checking this doesn't cause any issues - sometimes fetch without a body can cause weird issues

studio/hooks/misc/useIsFeatureEnabled.ts
```diff
@@ -0,0 +1,43 @@
+import { Feature } from 'data/profile/types'
+import { useProfile } from 'lib/profile'
+
+function checkFeature(feature: Feature, features?: Feature[]) {
+  return !features?.includes(feature) ?? true
+}
+
+type SnakeToCamelCase<S extends string> = S extends `${infer First}_${infer Rest}`
+  ? `${First}${SnakeToCamelCase<Capitalize<Rest>>}`
```
kevcodez, line 9: Proud of ya son

ivasilov (approved):

studio/__mocks__/hooks.js
alaister: True – good catch – I've updated!

joshenlim (approved):

PR #17566, opened 2023-09-19
Fix: minimist import in generating embedding script

gregnr:
The `minimist` import in the `apps/docs/scripts/search/generate-embeddings.ts` script is causing [embedding generation CI](https://github.com/supabase/supabase/actions/runs/6232522738/job/16915864789?pr=17546) to fail.

Problem was that
```ts
import * as minimist from 'minimist'
```
returns an object that is one level too high containing `minimist` as the default export, despite the fact that `minimist` has no default export. The reason why `minimist` is getting a default export is because we have `esModuleInterop: true` (for good reason).

Fixed by removing `* as`:
```ts
import minimist from 'minimist'
```

### DIFF
--- a/apps/docs/generator/index.ts
+++ b/apps/docs/generator/index.ts
@@ -10,6 +10,7 @@ const main = (command: string[], options: any) => {
 }
 
 // Run everything
+/* @ts-ignore - required for Next.js build */
 const argv = minimist(process.argv.slice(2))
 main(argv['_'], argv)
 
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -84,8 +84,8 @@
     "react-markdown": "^8.0.3",
     "rehype-slug": "^5.1.0",
     "remark": "^14.0.2",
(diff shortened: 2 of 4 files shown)


apps/docs/generator/index.ts
```diff
@@ -10,6 +10,7 @@ const main = (command: string[], options: any) => {
 }
 
 // Run everything
+/* @ts-ignore - required for Next.js build */
 const argv = minimist(process.argv.slice(2))
```
gregnr, line 14: FYI `apps/docs/generator` has its own `tsconfig.json` without `esModuleInterop`, so the above `minimist` import needs to remain as
```ts
import * as minimist from 'minimist'
```

Needed to add this `@ts-ignore` to get the Next.js project to build - it must be ignoring this `tsconfig.json`. Proper solution is to convert `apps/docs/generator` to use `esModuleInterop`, but probably not worth it because we are likely going to remove `generator` soon.

fsansalvadore (approved):

PR #17569, opened 2023-09-19
docs: adds import statement to main.dart

EvgeniiaVak:

Documentation improvement for getting started with flutter tutorial.

There is no import statement in the docs, so I needed to try and find it. Android Studio also offered `'package:supabase/supabase.dart'` (which is wrong) as the first choice for auto-completion.


alaister:
Hi @EvgeniiaVak,

Thanks for the PR, and sorry for the delay!

It looks like in the meantime the docs have been updated to include the correct import: https://supabase.com/docs/guides/getting-started/quickstarts/flutter

Thanks again!

PR #17578, opened 2023-09-20
Debug tool for internal staff

joshenlim:


alaister (approved):

alaister (changes requested):

studio/pages/_app.tsx
```diff
@@ -61,6 +62,10 @@ dayjs.extend(timezone)
 dayjs.extend(relativeTime)
 dart(Prism)
 
+if (!IS_PLATFORM && process.env.NEXT_PUBLIC_ENVIRONMENT === 'staging') {
```
kangmingtay, line 65: maybe we can add a disclaimer here to make it clear that this is only for our staging environment and we do not use it in prod to track customer data?

joshenlim: added

alaister (approved):

PR #17636, opened 2023-09-22
Chore/protected schemas in database pages

joshenlim:


studio/components/interfaces/Database/Functions/CreateFunction.tsx
```diff
@@ -757,6 +723,13 @@ const InputDefinition = observer(({}) => {
...
 
+  const { project } = useProjectContext()
+  const { data } = useSchemasQuery({
+    projectRef: project?.ref,
+    connectionString: project?.connectionString,
+  })
+  const schemas = data?.filter((schema) => !EXCLUDED_SCHEMAS.includes(schema.name)) ?? []
+
   return (
     <Listbox
```
ivasilov, line 734: This Listbox for schemas looks like it's duplicated on a lot of places. We could add a component which fetches the schemas, renders loading state while fetching and show schemas in a listbox. Just an idea, we'll handle it in a separate PR.

ivasilov (approved):

studio/components/interfaces/Database/Functions/CreateFunction.tsx
joshenlim: ahh yes good point actually - something SchemaSelection component

PR #17638, opened 2023-09-22
Add option to download table as CSV in table dropdown

joshenlim:
image

### DIFF
--- a/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
+++ b/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
@@ -1,12 +1,28 @@
+import saveAs from 'file-saver'
+import Papa from 'papaparse'
 import clsx from 'clsx'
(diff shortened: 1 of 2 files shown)


ivasilov:
I'm not exactly sure if this is a good idea since it will hang for big tables. The proper way would to prepare the CSV from API side but that will take longer to implement. I guess we can merge it and see how it goes.

ivasilov (approved):

joshenlim:
> I'm not exactly sure if this is a good idea since it will hang for big tables. The proper way would to prepare the CSV from API side but that will take longer to implement. I guess we can merge it and see how it goes.

good point actually - we probably should check the DB size if we're gonna download the whole table (right now theres 2 ways to do it, either here or select all rows via the checkboxes and download), perhaps set a max limit - there's also a browser memory constraint that we're limited by and i can imagine the browser would stall if the user's download a 5GB table for example

PR #17657, opened 2023-09-22
Chore/refactor pooler form

saltcod:
— Refactor pooler connection panel to use common Form component
— add min 10 as the default pool size
— add a cmdk menu item for pooler


ivasilov (changes requested):
Few things:
- If I leave `Max client connections` or `Default pool size` empty I cannot save, but there's no error label to explain why I cannot save
- The loading screen doesn't include all fields Screenshot 2023-09-29 at 13 19 21

PR #17666, opened 2023-09-23
feat: add editing support for `citext` data type in table editor

### LINKED ISSUE #17451: Editing of citext field in studio broken
image

when attempting to edit a field via the studio table editor, it gives a select control with NULL as the only option

kishanhitk:

feature

Columns of `citext` are not editable, as the current logic assumes it to be an enum.

image

Fixes: #17451
## What is the new behavior?
Columns of `citext` are treated as text columns, and are editable as text.

image


ivasilov (changes requested):
hey @kishanhitk, instead of expanding `isTextColumn`, can you define a new function `isCiTextColumn` and add in the `if-else` chain beneath `isTextColumn`? I think that approach is clearer.

kishanhitk:
Thanks @ivasilov,
I have made the suggested change, and also added `citext` as one of the possible types in `ColumnType`.
Does it look good now?

ivasilov (approved):

ivasilov:
@kishanhitk Thank you for your contribution. Merging.

Issue #15741, opened 2023-07-12
Add Update, Upsert, and Delete examples to Python reference docs


AdamGEmerson:
# Improve documentation

## Link 

Add a link to the page which needs improvement (if relevant)
[Python Docs](https://supabase.com/docs/reference/python/select)

## Describe the problem

The Python documentation only has examples for Fetch and Insert in the "Database" section, i'd like to improve the Python documentation by bringing more inline with the JS docs and adding example sections for Update, Upsert, and Delete.

This has been requested in the issues of the supabase-py repo [here](https://github.com/supabase-community/supabase-py/issues/486).

## Describe the improvement

I believe this should be accomplished by editing the[ python yml file](https://github.com/supabase/supabase/blob/master/spec/supabase_py_v2.yml ) to include the additional API endpoints.

## Additional context

I see that there has been some movement on the python docs with #15053, but I think that is only expanding on the INSERT documentation, not adding additional endpoints for UPDATE, UPSERT, etc.

**If this is something that anyone on the team deems worthwhile then I would love to take a crack at it.**

Thanks!

wwzeng1:
Hi! I’m one of the founders of Sweep, a github app that solves issues(like writing documentation) by writing pull requests. This looks like a good issue for Sweep https://github.com/sweepai/sweep to try. We have onboarding [instructions here](https://github.com/sweepai/sweep/blob/main/README.md#-getting-started), I’m also happy to help you onboard directly :)

gregnr:
@AdamGEmerson that would be great! Feel free to open a PR and we'll take a look!

AdamGEmerson:
@gregnr Great! I'll see what I can whip up.

Regarding implementation, I've noticed that the [existing Python INSERT](https://supabase.com/docs/reference/python/insert) documentation is pulling in the parameter data from the `postgrest-js` typedocs, which isn't precisely a one to one match with the `postrgest-py` library.  

I think we should probably emulate the Flutter (and other community library) docs and remove the `postgrest-js` references from the Python docs.

That said, in lieu of the rich typing information afforded by typedocs, I could provide a direct link to the `postgrest-py` endpoint in the descriptions of the supabase-py endpoints.

Something like this: 
![Python INSERT w/ postgrest link](https://github.com/supabase/supabase/assets/35545129/3b1a6e0e-f453-469b-8628-07fd8c1a8a29)

Which would link to: 
[...]

PR #17676, opened 2023-09-24
Docs: Add Database documentation for python library

AdamGEmerson:

Python docs additions for the following methods:

### Database
- `update()`
- `upsert()`
- `delete()`

### Filters
- applying filters
- chaining filters
- conditional chaining
- filter by value within json columns
- filter foreign tables
- `eq()`
- `neq()`
- `gt()`
- `gte()`
- `lt()`
- `lte()`
- `like()`
- `ilike()`
- `is_()`
- `in_()`
- `contains()`
- `contained_by()`
- `match()`
- `not_`
- `filter()`

### Modifiers
- `order()`
- `limit()`
- `single()`
- `maybe_single()`

---
As I was going through the documentation, I found a few issues in the JS docs that I corrected as well:
- Added missing response object for `like()` filter in js docs.
- Fixed invalid query example on js docs for `contains()` (query and response inconsistent).

Missing documentation for most of the database methods, per the discussion here:  #15741 

## What is the new behavior?

The Python documentation is now much closer to parity with the JS library.  I tried to keep the examples as similar as possible to the existing database examples for each method in the JS library, although some methods behave differently (`not_`). 

### Some caveats

RPC has issues in the python library currently, so I have omitted it for now.

The python library does not appear to have filters equivalent to:
- `rangeGt()`
- `rangeGte()`
- `rangeLt()`
- `rangeLte()`
- `rangeAdjacent()`
- `overlaps()`
- `textSearch()`

The python library does not appear to have equivalent modifiers to:
- `range()`
- `abortSignal()`
- `csv()`
- `returns()`


silentworks:
Hi @AdamGEmerson, thanks for this PR. There are some line breaks showing up in the preview which would be great to get removed. You can see the preview here https://docs-git-fork-adamgemerson-master-supabase.vercel.app/docs/reference/python/select?example=filtering-through-foreign-tables.

We can merge this PR in and then improve on those after if you wish, just so this PR doesn't get stale or outdated.

silentworks (approved):
Looks good minus the small issue I mentioned in my previous comment.

AdamGEmerson:
@silentworks Just generated a new PR to address the aforementioned line breaks: #18011 

Thanks!

PR #17679, opened 2023-09-24
Fix: Resolve null.length TypeError in WrapperRow.tsx

monicakh:

Bug fix

The issue is caused by an attempt to access the 'length' property of a null value in the `WrapperRow.tsx` file at line 117 when we don't have any foreign tables available:
image

## What is the new behavior?

image


alaister (approved):
LGTM! 🔥

PR #17688, opened 2023-09-25
Added set duration to some toasts

### LINKED ISSUE #17687: Toasts don't have durations set
# Bug report

## Describe the bug

It looks like majority of the toasts that appear do not have durations set for them, so they remain visible until closed by the user. Normally, there is some sort of timeout for these toasts.

## To Reproduce

An example of where this happens is here:

https://github.com/supabase/supabase/assets/59986120/8f658fe0-984e-4847-8147-38cc985276c7

## Expected behavior

I think it would be better to have a set timeout on all of the toasts. I saw some which had a 4s duration set, but majority don't have a duration set.

faizanraso:

This PR resolves https://github.com/supabase/supabase/issues/17687

Currently, majority of the toasts that appear to do not have a set duration, so they appear on the screen until a user dismisses them, a different page is navigated to etc.

## What is the new behavior?

I added durations of 4s to some of the error toasts.

## Additional context

Could a member of the supabase team please advise if this is something they would like? I can continue and set timeouts for the remainder of the toasts as well.

### DIFF
--- a/studio/components/grid/components/header/Header.tsx
+++ b/studio/components/grid/components/header/Header.tsx
@@ -366,6 +366,7 @@ const RowHeader = ({ table, sorts, filters }: RowHeaderProps) => {
       ui.setNotification({
         category: 'error',
         message: `Sorry! We're unable to support exporting of CSV for row counts larger than ${MAX_EXPORT_ROW_COUNT.toLocaleString()} at the moment.`,
+        duration: 4000
       })
       return setIsExporting(false)
     }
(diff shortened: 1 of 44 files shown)


saltcod:
Hey @faizanraso thanks for this, but we've intentionally set these error toasts to not expire so they don't just disappear before you can read them. The others, we do have a 4s expiry on.

faizanraso:
> Hey @faizanraso thanks for this, but we've intentionally set these error toasts to not expire so they don't just disappear before you can read them. The others, we do have a 4s expiry on. 

Thanks for letting me know! @saltcod

PR #17693, opened 2023-09-25
Chore/service status indications

joshenlim:
image


studio/components/interfaces/Home/ServiceStatus.tsx
```diff
@@ -0,0 +1,181 @@
...
+      isSuccess: isSuccessAuth,
+    },
+    {
+      name: 'Database',
+      isLoading: isLoadingPostgres,
+      isError: isErrorPostgres,
+      isSuccess: isSuccessPostgres,
+    },
+    {
+      name: 'Postgrest',
```
steve-chavez, line 103: IMO it's clearer this way:
```suggestion
      name: 'PostgREST',
```

steve-chavez: Alternatively just "REST" or "REST API"

studio/data/service-status/edge-functions-status-query.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export type EdgeFunctionServiceStatusVariables = {
+  projectRef?: string
+}
+
+export async function getEdgeFunctionServiceStatus(
+  { projectRef }: EdgeFunctionServiceStatusVariables,
+  signal?: AbortSignal
+) {
+  const res = await get(`https://obuldanrptloktxcffvn.supabase.co/functions/v1/health-check`, {
```
alaister, line 15: Should this be a static URL? Or should `projectRef` be interpolated?

joshenlim: static URL! but good point - shouldn't need project ref here
imma keep it in the query key though, although we'd be pinging the same endpoint, but its to have RQ check once more each time the user switches projects

this check for edge functions is kinda like a v1 until we're able to make it more fine grained

alaister (approved):
LGTM!

I think it might also be good to include an action for the user to take when their services are unhealthy

PR #17717, opened 2023-09-26
Added redirect from /documentation to /docs

AdamNReese:
Redirects from /documentation to /docs on the website.

/documentation path shows a 404 not found message

## What is the new behavior?
/documentation redirects to the documentation at /docs

## Additional context
This is my first PR for the project. I want to learn the codebase and make more useful contributions.


kiwicopple:
thanks for the contribution @AdamNReese 

I think we shouldn't have any links to `/documentation` so this redirect isn't necessary. If we had it, then this would be a great contribution. Thanks!

Issue #10977, opened 2022-12-14
The file system does not support extended attributes on windows server


pckimlong:
# Bug report

## Describe the bug

I got this error when trying to host supabase on windows server 2019 docker

`The file system does not support extended attributes or has the feature disabled`

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Start Supabase docker locally
2. Open studio at localhost:3000
4. Create bucket in storage UI
5. Upload a file

## Expected behavior

Upload work

## System information

- OS:  Windows Server 2019

## Additional context

I saw there others has the same problem but they are on MacOS, but I am Windows OS

pckimlong:
I try do it on docker windows 11, Everything work find. Seem like only docker on MacOS and Windows Server has this issue

JiangXin-1215:
same problem,os windows 11

the log is

2023-02-09 09: 07: 09 {
    "level": "info",
    "time": "2023-02-09T01:07:09.742Z",
    "pid": 18,
    "hostname": "a04d50663e28",
    "reqId": "req-7",
    "tenantId": "stub",
    "project": "stub",
    "req": {
        "method": "GET",
        "url": "/bucket",
        "headers": {
            "host": "storage:5000",
            "x_forwarded_proto": "http",
            "x_real_ip": "172.18.0.1",
            "x_client_info": "supabase-js/2.1.1",
            "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36",
            "accept": "*/*",
            "referer": "http://localhost:3000/"
        },
        "hostname": "storage:5000",
        "remoteAddress": "172.18.0.5",
        "remotePort": 50318
    },
    "res": {
        "statusCode": 200
    },
    "responseTime": 7.564552999800071,
    "msg": "GET | 200 | 172.18.0.5 | req-7 | /bucket | Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36"
}
2023-02-09 09: 07: 09 {
    "level": "info",
    "time": "2023-02-09T01:07:09.807Z",
    "pid": 18,
    "hostname": "a04d50663e28",
    "reqId": "req-8",
    "tenantId": "stub",
    "project": "stub",
    "results": [],
    "msg": "results"
}
[...]

PR #17779, opened 2023-09-28
fix: use storage-api with the S3 adapter for better compatibility

fenos:

Feature / Fix

Currently, Storage-API uses the file-system adapter when using docker-compose.
As this solution works just fine, some of our users are experiencing errors when uploading files because of the limitation of the Docker File sharing implementation, only `VirtioFS` works reliably on MacOS.

Issue: https://github.com/supabase/supabase/issues/10977

## What is the new behavior?

We are now defaulting storage-api to use a local minio container and the S3 adapter, for better compatibility across systems.


inian (approved):

docker/docker-compose.yml
```diff
@@ -226,12 +257,18 @@ services:
       PGRST_JWT_SECRET: ${JWT_SECRET}
       DATABASE_URL: postgres://supabase_storage_admin:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
       FILE_SIZE_LIMIT: 52428800
-      STORAGE_BACKEND: file
+      STORAGE_BACKEND: s3
+      GLOBAL_S3_BUCKET: stub
+      GLOBAL_S3_ENDPOINT: http://minio:9000
+      GLOBAL_S3_PROTOCOL: http
+      GLOBAL_S3_FORCE_PATH_STYLE: true
```
sweatybridge, line 264: Does storage fail gracefully if user didn't create a s3 bucket? If not, we probably need to update the self-hosted guide together with these changes.

In addition, we need to document the upgrade path for existing self-hosted users with locally mounted volumes. Seems like quite a bit of work to get this through.

sweatybridge (approved):

ethicnology:
In supabase process, how long it takes for an approved PR to be merged ?

sweatybridge:
Apologies for the delay. Since this is a breaking change for existing users of filesystem backed storage, I've moved the s3 services to a separate compose file.

PR #17797, opened 2023-09-29
Update with-kotlin.mdx to fix sign in onclick

ByteSizedDesign:
The Onclick function in the sign in page was calling an non existent onSignIn() function from the model. I've altered the file to call the correct onLogin() function in the ViewModel instead.

Corrects example code within the documentation. 

Code calls a non existent function causing an error.

## What is the new behavior?

Code calls the correct onLogin() function in the ViewModel.

## Additional context


alaister:
Hi @Chemilmakhlouta,

Thanks for your PR!

It looks like the underlying issue has been resolved (https://supabase.com/docs/guides/getting-started/tutorials/with-kotlin), so I'm going to go ahead and close this as well.

PR #17824, opened 2023-09-30
Add disabled cursor style

mozi47:
Add cursor: not-allowed style to the theme toggle when it's disabled on the home and launch week pages

image


saltcod:
Apologies for the long delay on this one. Going to mark it as closed as this element is no longer being rendered on these pages. Thanks for the contribution, sorry we weren't able to get it in!

PR #17827, opened 2023-09-30
Update image-search-openai-clip.mdx to fix syntax error 

ankerbachryhl:

# Doc update

The docs used .upsert(vectors=[]) instead of of .upsert(records=[])

When using vectors in .upsert instead of records you get: "TypeError: Collection.upsert() got an unexpected keyword argument 'vectors'"

## What is the new behavior?

.upsert works as expected and add records in vector db.

### DIFF
--- a/apps/docs/pages/guides/ai/examples/image-search-openai-clip.mdx
+++ b/apps/docs/pages/guides/ai/examples/image-search-openai-clip.mdx
@@ -97,7 +97,7 @@ def seed():
 
     # add records to the *images* collection
     images.upsert(
-        vectors=[
+        records=[
             (
                 "one.jpg",        # the vector's identifier
                 img_emb1,          # the vector. list or np.array


apps/docs/pages/guides/ai/examples/image-search-openai-clip.mdx
```diff
@@ -97,7 +97,7 @@ def seed():
 
     # add records to the *images* collection
     images.upsert(
-        vectors=[
+        records=[
```
itsmdasifraza, line 100: Not needed!

ankerbachryhl: but .upsert in python vecs package doesn't accept vectors= as arguments only records?

alaister:
Hi @ankerbachryhl,

Thanks for your PR, and sorry for the delay!

It looks like the underlying issue has been resolved (https://supabase.com/docs/guides/ai/examples/image-search-openai-clip), so I'm going to go ahead and close this as well.

PR #17880, opened 2023-10-03
Feat/api side panel

joshenlim:
screenshot-2023-10-16-at-15 08 04


studio/components/interfaces/App/FeaturePreview/APISidePanelPreview.tsx
```diff
@@ -0,0 +1,48 @@
+import { BASE_PATH } from 'lib/constants'
+
+const APISidePanelPreview = () => {
+  return (
+    <div className="space-y-2">
+      <p className="text-foreground-light text-sm">
+        Get building with your client application faster with your Project's API documentation now
```
joshenlim, line 7: @saltcod wondering if you might have some thoughts about the copy writing here - feels like these are a bit wordy but i'm not sure how to concisely describe the changes haha

saltcod: Added a bunch of notes to notion! I think my biggest takeaway is we should cut a ton of this to the bone! API panel and this feature preview panel alike.

saltcod: Maybe an even more condense version of the top copy here and X all of the bottom stuff (or more likely, move the top stuff to the bottom) 

https://share.cleanshot.com/nsV61Qb1

(+ add a "Give Feedback" button to link to the Discussion we'll create, as mentioned)

joshenlim: > Added a bunch of notes to notion! I think my biggest takeaway is we should cut a ton of this to the bone! API panel and this feature preview panel alike.

gotcha on this! will go through em and see what can be actioned on immediately :)

joshenlim: > X all of the bottom stuff

i'm thinking about this though - should we? github doesnt have this section but i'm thinking it might be worth pointing out what are the UI changes once they've enabled so there won't be confusion as to things moving/changing

we could probably leave out _what_ the feature preview does though which is what most of the text is about, but i'm thinking we keep things like "GraphQL interface will shift to database page", API docs button shifts to bottom of left nav bar

joshenlim: so maybe just like this
image

studio/components/interfaces/ProjectAPIDocs/ProjectAPIDocs.tsx
```diff
@@ -0,0 +1,219 @@
...
+import { BASE_PATH } from 'lib/constants'
+
+const ProjectAPIDocs = () => {
+  const { ref } = useParams()
+  const snap = useAppStateSnapshot()
+
+  const { activeDocsSection } = snap
+  useEffect(() => {
+    console.log({ activeDocsSection })
+  }, [activeDocsSection])
```
alaister: Cheeky

studio/components/interfaces/ProjectAPIDocs/ProjectAPIDocs.tsx
```diff
@@ -0,0 +1,219 @@
...
+import FirstLevelNav from './FirstLevelNav'
+import SecondLevelNav from './SecondLevelNav'
+import { useCustomDomainsQuery } from 'data/custom-domains/custom-domains-query'
+import { BASE_PATH } from 'lib/constants'
+
+const ProjectAPIDocs = () => {
+  const { ref } = useParams()
+  const snap = useAppStateSnapshot()
+
+  const { activeDocsSection } = snap
```
alaister: I'm not sure if destructing is ideal for valtio because it doesn't always work as expected with js proxies. I'd stay on the safe side here and just always do `snap.activeDocsSection`

joshenlim: removed this one - realised its not needed

alaister (approved):
LGTM!

PR #17898, opened 2023-10-03
Update auth-azure.mdx

abs:
"Below" is better than "above" here

Docs update

image

## What is the new behavior?

"Above" becomes "below"


Isaiah-Hamilton (approved):

kevcodez:
Looks like this has already been fixed, so closing this one

PR #17906, opened 2023-10-04
Redirect to sign in if assurance level check fails on MFA page

joshenlim:


studio/pages/sign-in-mfa.tsx
```diff
@@ -31,8 +32,9 @@ const SignInMfaPage: NextPageWithLayout = () => {
           const { data, error } = await auth.mfa.getAuthenticatorAssuranceLevel()
           if (error) {
             // if there was a problem signing in via the url, don't redirect
+            toast.error(`Failed to retrieve assurance level: ${error.message}`)
```
kangmingtay, line 35: the only reason it will fail to retrieve the assurance level is if the session is missing (https://github.com/supabase/gotrue-js/blob/45b6e3e656f842c60ebe2709f59bad4eb6f5811e/src/GoTrueClient.ts#L2292) - we should just return an error saying that the user is not logged in and ask them to retry

alaister, line 35: Worth adding some action for the user to take here? Like:
```suggestion
            toast.error(`Failed to retrieve assurance level: ${error.message}. Please try signing in again`)
```

alaister: lol @kangmingtay already beat me to it

alaister (approved):
LGTM!

PR #17937, opened 2023-10-04
Update reference to cookies in nextjs examples

silentworks:

Docs update for Next.js cookie setup

The build process for Next.js fails when it cannot detect that a cookie is being used in a server route.

## What is the new behavior?

The build process for Next.js is successful because the cookies function is in the scope of the bundler, so it knows that server route is dynamic.

## Additional context


apps/docs/pages/guides/auth/auth-helpers/nextjs.mdx
```diff
@@ -171,7 +171,8 @@ export async function GET(request) {
   const code = requestUrl.searchParams.get('code')
 
   if (code) {
-    const supabase = createRouteHandlerClient({ cookies })
+    const cookieStore = () => cookies()
+    const supabase = createRouteHandlerClient({ cookies: () => cookieStore })
```
gregnr, line 175: Nice. Confirming this should be a double callback?

ie:
```ts
createRouteHandlerClient({ cookies: () => () => cookies() })
```

silentworks: Good catch, that is indeed incorrect. This is what find and replace does after replacing one before running it.

silentworks: Oh wait, they are all wrong.

gregnr (approved):

gregnr:
@silentworks should the `with-supabase` Next.js example template be updated?

https://github.com/vercel/next.js/blob/3c326ea5e822bd97aaa08313bbda246124a37fd3/examples/with-supabase/app/page.tsx#L43

PR #17943, opened 2023-10-05
Fix introduction padding

Isaiah-Hamilton:

fix issue with introduction padding not being able to see title of introduction section.

CleanShot 2023-10-04 at 21 05 15

## What is the new behavior?

CleanShot 2023-10-04 at 21 05 45

## Additional context

I don't know if there is a way to target the desired div like that in tailwindcss. We could also add more padding so it doesn't look weird super close to the navbar.

### DIFF
--- a/apps/docs/layouts/ref/RefSubLayout.tsx
+++ b/apps/docs/layouts/ref/RefSubLayout.tsx
@@ -52,7 +52,7 @@ interface ISectionExamples {}
 
 const RefSubLayout: FC<RefSubLayoutType> & RefSubLayoutSubComponents = (props) => {
   return (
-    <div className="flex flex-col w-full divide-y px-5 max-w-7xl mx-auto py-16">
+    <div className="flex flex-col w-full divide-y px-5 max-w-7xl mx-auto pb-16 introduction-padding">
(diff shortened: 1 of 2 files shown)


apps/docs/layouts/ref/RefSubLayout.tsx
```diff
@@ -52,7 +52,7 @@ interface ISectionExamples {}
 
 const RefSubLayout: FC<RefSubLayoutType> & RefSubLayoutSubComponents = (props) => {
   return (
-    <div className="flex flex-col w-full divide-y px-5 max-w-7xl mx-auto py-16">
+    <div className="flex flex-col w-full divide-y px-5 max-w-7xl mx-auto pb-16 introduction-padding">
```
MildTomato, line 55: This seems a bit hacky.

Could we not investigate why the viewport is not fully at the top instead ?

MildTomato (changes requested):

alaister:
Hey Isaiah,

It looks like this has got a little stale, so I'm going to close it. Please feel free to fix any merge conflicts and open up a new PR if the changes are still relevant.

PR #17949, opened 2023-10-05
docs: google auth docs improvements.

thorwebdev:

Use tabs for different integration paths.


dshukertjr (approved):

dshukertjr (changes requested):
Actually, do you think it's better to avoid tabs inside tabs? Maybe for the `web` tab, we can just have js based web development.

Pretty much all the frameworks have web support these days such as Flutter-web, or React-native-web, so it might get messy if we try to separate all the frameworks in side another tab in the `web` tab.
Screenshot 2023-10-05 at 17 18 29

thorwebdev:
Yah, this was a recent contribution from our Kotlin library maintainer. But yes, you're right, it doesn't really make sense here, will take it out.

dshukertjr (approved):
LGTM!

PR #17955, opened 2023-10-05
chore: Build the docker image on schedule

ivasilov:
This PR changes the action for the Docker image for `studio` to be built every Monday at 04:00 UTC (Singapore noon).


.github/workflows/publish_image.yml
```diff
@@ -1,11 +1,9 @@
 name: Publish to Image Registry
 
 on:
-  push:
-    branches:
-      - 'studio'
-    tags:
```
kevcodez, line 7: Do we no longer create tags? Should we still run it on tag creation?

kevcodez (approved):

.github/workflows/publish_image.yml
sweatybridge: Yes I think we should keep these so that our [release announcements](https://github.com/supabase/supabase/releases) are accompanied by docker hub image tag.

ivasilov: Fixed.

sweatybridge (approved):

PR #17960, opened 2023-10-05
fix: Refetch tables when enabling/disabling RLS

ivasilov:


studio/components/interfaces/Auth/Policies/Policies.tsx
```diff
@@ -76,6 +81,8 @@ const Policies = ({ tables, hasTables, isLocked }: PoliciesProps) => {
         category: 'error',
         message: `Failed to toggle RLS: ${res.error.message}`,
       })
+    } else {
+      queryClient.invalidateQueries(tableKeys.list(ref, snap.selectedSchemaName))
```
alaister, line 85: improved loading state:
```suggestion
      await queryClient.invalidateQueries(tableKeys.list(ref, snap.selectedSchemaName))
```

Issue #17860, opened 2023-10-02
table-editor(timestamp): Editing year not accepting keyboard input


tautf:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When trying to edit a row of a table with the table editor that is of type timestamp, your not able to input the year for the timestamp.
Day and month are proberly accepted. Year is not. Only the last character is replaced always. Typing 2024 will result in 1902 -> 1900 -> 1902 -> 1904.
Using to up and down arrows your able to move to different years.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

See video:
https://github.com/supabase/supabase/assets/18403881/ee3b18b4-817a-433a-8039-34fb53d669e7

## Expected behavior

Input taken as for day and month.

## Screenshots

See video

## System information

- OS: [e.g. macOS, Windows] macOS Ventura (latest)
- Browser (if applies) [e.g. chrome, safari] Brave 1.58.131
- Version of supabase-js: [e.g. 6.0.2] -
- Version of Node.js: [e.g. 10.10.0] latest 18 LTS

shishiro26:
I would like to work on this issue !

saltcod:
Hey @shishiro26 — we don't have an assignment process. Feel free to jump in, thanks!

shishiro26:
Oh, I didn't know that. I will start working on it right now. @saltcod

shishiro26:
Hey @saltcod, I just have a doubt: the createdAt value should be set as non-editable, right? It is a value that is set only once and not every time.

shishiro26:
Hey @saltcod, could you please check my Pull request

shishiro26:
Hey @saltcod i am facing trouble in installing the dependencies can u please help me in his I couldn't run npm install in my windows.

saltcod:
Are you running an old version of Node by chance? Not familiar with Windows development I'm afraid.

tranhoangvuit:
That issue pertains to [dayjs](https://github.com/iamkun/dayjs/issues/2591); they already have a [PR](https://github.com/iamkun/dayjs/pull/2551) for it. Let's wait for them to release a newer version, and then we will update on our side as well. In the meantime, let's use a UI (date picker) for it.
[...]

PR #17989, opened 2023-10-06
fixed the bug in the editor table

shishiro26:

Bug fix.

When trying to edit a row of a table with the table editor that is of type timestamp, the user is not able to input the year for the timestamp. Day and month are properly accepted. Year is not. Only the last character is replaced always. Typing 2024 will result in 1902 -> 1900 -> 1902 -> 1904. Using the up and down arrows, the user is able to move to different years. Also, if the user tries to edit the month and types zero, the whole input gets cleared.

## What is the new behavior?

The PR updates the `BaseEditor` component to handle the issues related to editing timestamps. It ensures that the user can input the year, and the input for day and month works correctly without clearing the input when typing zero.

#17860


saltcod:
@shishiro26 looks like you need to run prettier & there's also a failed test in the runner. Could you run the tests locally? likely something needs adjusting with this change.

shishiro26:
Ok @saltcod

saltcod:
aah sorry @shishiro26 – you just do npm run build. That will output the errors for you.

PR #17739, opened 2023-09-26
fix: api reports styling

Ziinc:
This PR adds in the following changes:
- reduced x,b padding for routes tables
- add collapsible toggle
- add "Open in Logs Explorer" button

https://github.com/supabase/supabase/assets/22714384/9ea49cbe-e216-4c9c-91d2-d33250b7cf9d

PR #18023, opened 2023-10-08
feat: rework database health report disk space

kevcodez:
Everything related to database size on the reports page is currently super confusing. We mix multiple concepts (database space / disk size) and throw in usage, limits and project vs org-based billing. Users are pretty confused. On top of that, we were comparing bytes with GB, which lead to something like this:

image

As this is the **database report** page, we don't need to mix in the subscription, plan usage etc.

This PR aims to improve the situation by providing more meaningful insights about database space:
* Show top 5 largest objects in database
* Show used database space
* Link to database space docs
* Move panel to the bottom (CPU/RAM/Swap is more important!)
* Refactor to use ReportWidget
* Adds large objects template to SQL templates

Screenshot 2023-10-18 at 10 38 46

Screenshot 2023-10-18 at 10 38 22

Database space docs can also be improved greatly (show more ways to find big indexes or tables), go more in-depth, this is out of scope for this PR.

### DIFF
--- a/studio/components/interfaces/Reports/ReportWidget.tsx
+++ b/studio/components/interfaces/Reports/ReportWidget.tsx
@@ -3,7 +3,7 @@ import Panel from 'components/ui/Panel'
 import { NextRouter, useRouter } from 'next/router'
 import { Button, IconExternalLink, IconHelpCircle, Loading } from 'ui'
 import { LogsEndpointParams } from '../Settings/Logs'
-import { BaseReportParams } from './Reports.types'
+import { BaseReportParams, ReportQueryType } from './Reports.types'
 
 export interface ReportWidgetProps<T = any> {
   data: T[]
@@ -17,7 +17,9 @@ export interface ReportWidgetProps<T = any> {
   appendProps?: Partial<ReportWidgetRendererProps>
   // omitting params will hide the "View in logs explorer" button
   params?: BaseReportParams | LogsEndpointParams
+  queryType?: ReportQueryType
   isLoading: boolean
+  resolvedSql?: string
 }
 
 export interface ReportWidgetRendererProps<T = any> extends ReportWidgetProps<T> {
@@ -77,13 +79,25 @@ const ReportWidget: React.FC<ReportWidgetProps> = (props) => {
                   icon={<IconExternalLink strokeWidth={1.5} />}
                   className="px-1"
                   onClick={() => {
+                    const isDbQueryType = props.queryType === 'db'
+
+                    const pathname = isDbQueryType
+                      ? `/project/${projectRef}/sql/new`
+                      : `/project/${projectRef}/logs/explorer`
+
+                    const query: Record<string, string | undefined> = {}
+
+                    if (isDbQueryType) {
+                      query.content = props.resolvedSql
+                    } else {
+                      query.q = props.params?.sql
+                      query.its = props.params!.iso_timestamp_start
+                      query.ite = props.params!.iso_timestamp_end
+                    }
+
                     router.push({
-                      pathname: `/project/${projectRef}/logs/explorer`,
-                      query: {
-                        q: props.params?.sql,
-                        its: props.params!.iso_timestamp_start,
-                        ite: props.params!.iso_timestamp_end,
-                      },
+                      pathname,
+                      query,
                     })
                   }}
                 />
@@ -97,7 +111,9 @@ const ReportWidget: React.FC<ReportWidgetProps> = (props) => {
                       'border border-scale-200',
                     ].join(' ')}
                   >
-                    <span className="text-xs text-foreground">Open in Logs Explorer</span>
+                    <span className="text-xs text-foreground">
+                      {props.queryType === 'db' ? 'Open in SQL Editor' : 'Open in Logs Explorer'}
+                    </span>
                   </div>
                 </Tooltip.Content>
               </Tooltip.Portal>
--- a/studio/components/interfaces/Reports/Reports.constants.ts
+++ b/studio/components/interfaces/Reports/Reports.constants.ts
@@ -355,6 +355,29 @@ select
       },
     },
   },
+  [Presets.DATABASE]: {
+    title: 'database',
+    queries: {
+      largeObjects: {
+        queryType: 'db',
+        sql: (_) => `SELECT 
+        SCHEMA_NAME,
+        relname,
+        table_size
+      FROM
+        (SELECT 
+          pg_catalog.pg_namespace.nspname AS SCHEMA_NAME,
+          relname,
+          pg_relation_size(pg_catalog.pg_class.oid) AS table_size
+        FROM pg_catalog.pg_class
+        JOIN pg_catalog.pg_namespace ON relnamespace = pg_catalog.pg_namespace.oid
+        ) t
+      WHERE SCHEMA_NAME NOT LIKE 'pg_%'
+      ORDER BY table_size DESC
+      LIMIT 5;`,
+      },
+    },
+  },
 }
 
 export const DATETIME_FORMAT = 'MMM D, ha'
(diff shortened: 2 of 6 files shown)


studio/pages/project/[ref]/reports/database.tsx
```diff
@@ -285,6 +133,55 @@ const DatabaseUsage = observer(() => {
               </div>
             </Panel.Content>
           </Panel>
+
+          <Panel
+            title={
+              <div className="flex justify-between w-full">
+                <h2>Database size</h2>
+                <span className="text-lg tracking-wide">{formatBytes(databaseSizeBytes)}</span>
```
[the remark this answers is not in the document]
Comment (Ziinc) [reply studio/pages/project/[ref]/reports/database.tsx:141]: having the value so far from the label without a line to guide the eye reduces scanning speed.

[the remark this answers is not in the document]
Comment (Ziinc) [reply studio/pages/project/[ref]/reports/database.tsx:141]: Screenshot 2023-10-09 at 8 17 17 PM
somehting like this for example.

studio/pages/project/[ref]/reports/database.tsx
```diff
@@ -285,6 +133,55 @@ const DatabaseUsage = observer(() => {
...
+          >
+            <Panel.Content>
+              <div className="space-y-1">
+                <span className="text-md">Largest Objects</span>
+
+                <Loading active={!loadedLargestObjects}>
+                  <div className="space-y-3 mt-4">
+                    {largestObjects?.result?.map((object) => (
+                      <div key={`${object.schema_name}.${object.relname}`}>
+                        <SparkBar
```
Comment (Ziinc) [inline studio/pages/project/[ref]/reports/database.tsx:153]: we can display this in a 2 column table.  Don't need to show 10 items as not all will be interesting and will take up unnecessary vertical height. Showing top 3 with an expand for 10 will be enough. Can also link to the scripts query page to make it more usable, like how the API reports is styled [here](https://github.com/supabase/supabase/pull/17739) in this PR.

Ziinc (changes requested):

studio/pages/project/[ref]/reports/database.tsx
Comment (Ziinc) [reply studio/pages/project/[ref]/reports/database.tsx:153]: we can remove use of the sparkbar and instead show it as a % in brackets next to the value. In this situation it would be better to show a proportion visualization, such as segmented horizontal bar chart.

Comment (kevcodez) [reply studio/pages/project/[ref]/reports/database.tsx:153]: Screenshot 2023-10-09 at 15 40 41

Was inspired by Vercel for the current styling (even though Vercel puts the text in the bar)

Comment (kevcodez) [reply studio/pages/project/[ref]/reports/database.tsx:153]: Do we have a proportion visualization anywhere already?

Comment (kevcodez) [reply studio/pages/project/[ref]/reports/database.tsx:153]: Can just add the percentage in brackets for now

loong:
Love this!

Would be cool if you could click on a button and it'll open the query for this in the sql editor

kevcodez:
@Ziinc it's still WIP but I'm trying to convert this entire page to use ReportWidget - this is a bit tricky since a lot of the current components are just working with Log queries. I.e. ReportFilterBar hardcodes filters that only work with Logflare HTTP

Still open
* Replace Metric Charts (memory, swap, ...) with ReportWidget, needs extension of the reporting logic to not only query logflare/user db but also the API endpoint that is currently used, this is not supported at all yet
* Clean up
* Ensure filtering works
* Allow SQL editor page to accept a query parameter to prefill the SQL

studio/hooks/analytics/useDbQuery.tsx
```diff
@@ -27,6 +27,8 @@ const useDbQuery = (
 
   const resolvedSql = typeof sql === 'function' ? sql([]) : sql
 
+  params.sql = resolvedSql
```
kevcodez, line 30: Probably not 100% correct, but `params` was missing the SQL. Only worked for useLogsQuery but not for useDbQuery.

studio/pages/project/[ref]/reports/database.tsx
```diff
@@ -285,8 +135,127 @@ const DatabaseUsage = observer(() => {
...
+  const { ref: projectRef } = useParams()
+
+  const queryHooks = queriesFactory<keyof typeof PRESET_CONFIG.database.queries>(
+    PRESET_CONFIG.database.queries,
+    projectRef ?? 'default'
+  )
+  const largeObjects = queryHooks.largeObjects()
+  const activeHooks = [largeObjects]
+
+  const handleRefresh = async () => {
```
Comment (Ziinc) [inline studio/pages/project/[ref]/reports/database.tsx:231]: can remove, as it is a static query without a refresh button

studio/pages/project/[ref]/reports/database.tsx
```diff
@@ -285,8 +135,127 @@ const DatabaseUsage = observer(() => {
...
+    },
+    errors: {
+      largeObjects: largeObjects.error,
+    },
+    params: {
+      largeObjects: largeObjects.params,
+    },
+    mergeParams: handleSetParams,
+    isLoading,
+    refresh: handleRefresh,
```
Comment (Ziinc) [inline studio/pages/project/[ref]/reports/database.tsx:259]: can remove

studio/pages/project/[ref]/reports/database.tsx
```diff
@@ -285,8 +135,127 @@ const DatabaseUsage = observer(() => {
...
+  const handleRefresh = async () => {
+    activeHooks.forEach((hook) => hook.runQuery())
+  }
+  const handleSetParams = (params: Partial<BaseReportParams>) => {
+    activeHooks.forEach((hook) => {
+      hook.setParams?.((prev) => ({ ...prev, ...params }))
+    })
+  }
+  useEffect(() => {
+    if (largeObjects.changeQuery) {
```
Comment (Ziinc) [inline studio/pages/project/[ref]/reports/database.tsx:240]: i think this can be removed as the sql should be resolved within the hook and there wouldn't be any changes to the sql after initial load.

studio/components/interfaces/Reports/Reports.constants.ts
```diff
@@ -358,6 +358,34 @@ select
...
+          relname,
+          pg_relation_size(pg_catalog.pg_class.oid) AS table_size
+        FROM pg_catalog.pg_class
+        JOIN pg_catalog.pg_namespace ON relnamespace = pg_catalog.pg_namespace.oid
+        ) t
+      WHERE SCHEMA_NAME NOT LIKE 'pg_%'
+      ORDER BY table_size DESC
+      LIMIT 5;`,
+      },
+      databaseSpace: {
```
Ziinc, line 382: this looks like it isn't used

studio/components/interfaces/Reports/ReportWidget.tsx
```diff
@@ -77,13 +78,24 @@ const ReportWidget: React.FC<ReportWidgetProps> = (props) => {
...
                   className="px-1"
                   onClick={() => {
+                    const pathname =
+                      props.queryType === 'db'
+                        ? `/project/${projectRef}/sql/new`
+                        : `/project/${projectRef}/logs/explorer`
+
+                    const query: Record<string, string | undefined> = {}
+
+                    if (props.queryType !== 'db') {
```
Ziinc, line 88: would be clearer to match the condition at line 82

studio/components/interfaces/Reports/ReportWidget.tsx
```diff
@@ -77,13 +78,24 @@ const ReportWidget: React.FC<ReportWidgetProps> = (props) => {
                   icon={<IconExternalLink strokeWidth={1.5} />}
```
Ziinc, line 78: i do think that we shouldn't tie this button too much to opening it in editor/explorer, as it reduces possible use cases (like opening in a main report, or opening in usage report). Can refactor in the future.

Ziinc (commented):
overall looks good, added comments on the useDatabaseReport hook for the areas that are not in use. As conversion on the prom queries to ReportWidget isn't a priority at the moment, would suggest removing the unused handlers for now to keep the code cleaner, even if they need to be added back in the future.

Ziinc (approved):

alaister (approved):

PR #18024, opened 2023-10-08
fix: misleading CTA / current billing cycle check

kevcodez:
* Correctly tell the user to upgrade plan OR disable spend cap, depending on their subscription
* Pass in subscription end date to correct check if the current billing cycle is selected (previously, we just added 1 month in the date range picker leading to false negatives)

### DIFF
--- a/studio/components/interfaces/Organization/Usage/Usage.tsx
+++ b/studio/components/interfaces/Organization/Usage/Usage.tsx
@@ -63,6 +63,8 @@ const Usage = () => {
     )
   }, [dateRange, subscription])
 
+  console.log({ currentBillingCycleSelected, dateRange, subscription })
+
   const startDate = useMemo(() => {
     // If end date is in future, set end date to now
     if (!dateRange?.period_start?.date) {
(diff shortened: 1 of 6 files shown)


studio/components/interfaces/Organization/Usage/Usage.tsx
```diff
@@ -63,6 +63,8 @@ const Usage = () => {
     )
   }, [dateRange, subscription])
 
+  console.log({ currentBillingCycleSelected, dateRange, subscription })
```
alaister, line 66: Cheeky console.log

alaister (changes requested):
Looks good, but there's a cheeky console.log!

studio/components/interfaces/Organization/Usage/Usage.tsx
```diff
@@ -63,6 +63,8 @@ const Usage = () => {
...
+
```
joshenlim, line 67:
```suggestion
```

joshenlim, line 67:
```suggestion
```

alaister (approved):

PR #18044, opened 2023-10-09
chore: Remove ProjectContent mobx store

ivasilov:
This PR removes `ProjectContent` store which was used to store recent and saved logs.


alaister (approved):
LGTM! 🔥

PR #18052, opened 2023-10-09
Revert Button change

fsansalvadore:
Revert Button component to previous state.

Note: the occurrences of `type={undefined}` fix a bug that the Button change was supposed to fix, where the buttons with asChild lose styling in Safari.

### DIFF
--- a/packages/ui/src/components/Button/Button.tsx
+++ b/packages/ui/src/components/Button/Button.tsx
@@ -162,13 +162,12 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       icon,
       iconRight,
       iconLeft,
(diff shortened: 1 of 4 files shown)


MildTomato (approved):

Steellgold (approved):

packages/ui/src/components/Button/Button.tsx
```diff
@@ -162,13 +162,12 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       icon,
       iconRight,
       iconLeft,
-      htmlType,
+      htmlType = 'button',
```
fredjens, line 165: This should be "submit"

PR #18076, opened 2023-10-10
feat: auto-redirect to base path if set

kevcodez:
When going to `/` and a base path is set to i.e. `/dashboard`, next will automatically redirect to it.

### DIFF
--- a/studio/next.config.js
+++ b/studio/next.config.js
@@ -177,6 +177,16 @@ const nextConfig = {
         destination: '/project/:ref/sql/new',
         permanent: true,
       },
+      ...(process.env.NEXT_PUBLIC_BASE_PATH?.length
+        ? [
+            {
+              source: '/',
+              destination: process.env.NEXT_PUBLIC_BASE_PATH,
+              basePath: false,
+              permanent: true,
+            },
+          ]
+        : []),
     ]
   },
   async headers() {


studio/next.config.js
```diff
@@ -177,6 +177,16 @@ const nextConfig = {
...
+      ...(process.env.NEXT_PUBLIC_BASE_PATH?.length
+        ? [
+            {
+              source: '/',
+              destination: process.env.NEXT_PUBLIC_BASE_PATH,
+              basePath: false,
+              permanent: true,
+            },
+          ]
+        : []),
```
kamilogorek, line 189: You can safely spread it without fallback as well, but not sure what's preferred style in codebases yet :)

```suggestion
      ...(process.env.NEXT_PUBLIC_BASE_PATH?.length && [
        {
          source: '/',
          destination: process.env.NEXT_PUBLIC_BASE_PATH,
          basePath: false,
          permanent: true,
        }
      ]),
```

ivasilov (approved):

studio/next.config.js
kamilogorek: What the heck is next doing there again o_O Spreading of _any_ value inside the object is fine by the JS spec, as it's syntax sugar for `Object.assign`. Oh well ¯\\_(ツ)_/¯

kevcodez: It's a spread inside an array though - maybe that's why?

kevcodez: Also looking forward for some juicy syntax and TS tips in the future from you! :D

kamilogorek: Oh, it IS an array. I thought that redirects is object. Then yes, it only works for objects implementing `Symbol.iterator` and primitives don't do that. Too many brackets and my brain failed 🥲

PR #18093, opened 2023-10-10
chore: Update Broadcast documentation

filipecabaco:

Updates JS library reference docs with more information regarding on how to use Broadcast from Realtime


apps/docs/components/reference/RefFunctionSection.tsx
```diff
@@ -23,7 +23,7 @@ const RefFunctionSection: React.FC<IRefFunctionSection> = (props) => {
   const tsDefinition =
     hasTsRef && props.typeSpec ? extractTsDocNode(hasTsRef, props.typeSpec) : null
   const parameters = hasTsRef && tsDefinition ? generateParameters(tsDefinition) : ''
-  const shortText = hasTsRef && tsDefinition ? tsDefinition.signatures[0].comment.shortText : ''
+  const shortText = hasTsRef && tsDefinition ? tsDefinition.signatures[0].comment?.shortText : ''
```
filipecabaco, line 26: Why is this failing?

w3b6x9 (approved):

spec/common-client-libs-sections.json
```diff
@@ -785,6 +785,13 @@
         "slug": "getchannels",
         "product": "realtime",
         "type": "function"
+      },
+      {
+        "id": "broadcast-message",
+        "title": "Broadcast a message",
+        "slug": "broadcastMessage",
```
gregnr, line 792: Can we stick to all lowercase for consistency?
```suggestion
        "slug": "broadcastmessage",
```

Alternatively `kebab-case` (which is ideal), but it appears we don't use that everywhere, particularly not in realtime.

gregnr (approved):

PR #18095, opened 2023-10-11
Docs: Update local dev guide to use correct db sync command

oscargws:
Docs show CLI command that no longer exists

Docs update

Docs currently instruct users to use `supabase db pull` which isn't a valid command in order to receive their existing supabase db schema when linking their local environment. It appears the correct usage is to use `supabase db remote commit`

## What is the new behavior?

Use `supabase db remote commit`


sweatybridge:
Hello, we added `db pull` command in newer versions of cli. The `db remote` commands will be deprecated going forward. Could you check the version you are running locally?

PR #18143, opened 2023-10-12
Chore: Mandatory Services Field in Support Form

monicakh:

Chore fix 

 This change will make the "Services" field mandatory in the support form. 

## What is the new behavior?

Users will be required to select a service when submitting a support request. This change ensures that support requests are categorized correctly.

Screenshot 2023-10-12 at 3 07 34 PM

### DIFF
--- a/studio/components/ui/MultiSelect/index.tsx
+++ b/studio/components/ui/MultiSelect/index.tsx
@@ -141,29 +141,45 @@ export default function MultiSelect({
       >
         <Popover_Shadcn_ open={open} onOpenChange={setOpen} modal={false}>
           <PopoverTrigger_Shadcn_ asChild>
-              <div
-                className={[
-                  'flex w-full flex-wrap items-start gap-1.5 p-1.5 cursor-pointer',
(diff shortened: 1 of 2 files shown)


studio/components/ui/MultiSelect/index.tsx
```diff
@@ -146,9 +145,41 @@ export default function MultiSelect({
               className={[
                 'flex w-full flex-wrap items-start gap-1.5 p-1.5 cursor-pointer',
                 `${selectedOptions.length === 0 ? 'h-9' : ''}`,
+                `${
```
ivasilov, line 148: There's no need for this. Switch the `[].join(' ')` for `cn([...])` with the removed classes and it will work.

ivasilov (changes requested):
The approach taken for showing errors on the Services MultiSelect runs in parallel and will cause issues in the future even if it's working at the moment. You should try and incorporate the error into `Formik` errors object.

To do this, set `errors.services` to an error if there's an error. Then, you can get this `errors` object in line 220, which you can pass it to the Services Multiselect in the `error` prop. There's no need to save the error in a separate React state.

PR #18148, opened 2023-10-12
Chore/changelog v2

saltcod:
Switches the source of https://supabase.com/changelog
**from:**  https://github.com/supabase/supabase/releases
**to**:  https://github.com/orgs/supabase/discussions/categories/changelog

— grab from discussions graphql endpoint instead of from releases endpoint in the restapi
— blends the old releases posts with the new discussions posts into one stream
— paginate the responses (not totally necessary now but if we add several changelogs a week (hopefully we will), we'll need this eventually) 

— 

Before merging, need to de-duplicate the content between these two sources. 
Will do this after PR approval, and before PR merge.


MildTomato:
missing a env var ?

Screenshot 2023-11-08 at 7 46 53 PM

alaister:
@saltcod currently getting a 500 error when trying to visit https://zone-www-dot-com-git-chore-changelog-v2-supabase.vercel.app/changelog

saltcod:
thanks @alaister — a Discussion had an  tag that needs to be self-closing . 
Updated to accommodate these.

apps/www/pages/changelog.tsx
```diff
@@ -6,40 +6,219 @@ import CTABanner from '~/components/CTABanner'
...
+              id
+              publishedAt
+              createdAt
+              url
+              title
+              # labels(first: 10) {
+                # nodes {
+                  # name
+                # }
+              # }
```
alaister, line 118: Any reason this still needs to be here if it's commented out?

alaister (approved):
LGTM! 🔥

apps/www/pages/changelog.tsx
saltcod: aah right right was testing out a thing with that, will remove

PR #18156, opened 2023-10-12
Update user-create-mutation.ts to not omit Cookie Data

activenode:
Avoid stripping away cookie data and hence avoid having problems with Cookie Based Authentication Proxies

I'd assume it's a Bug as it unnecessarily strips away Cookies which other requests don't.

https://github.com/supabase/supabase/blame/072222523afe612d575c2ff7e73e514692f545af/studio/data/auth/user-create-mutation.ts#L50

Currently credentials are stripped with `credentials: 'omit'` 
Ref: https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials

That means that in any Proxy Setup which uses Authentication the Auth system cannot read the credentials and the request fails for no good reason.

## What is the new behavior?

Credentials are not stripped, the default behaviour of passing credentials to the same site, which for obvious reasons is safe, is the new default.

## Additional context

Was actually reported by one of the users (`colin`) in the YouTube video here https://www.youtube.com/watch?v=wyUr_U6Cma4


activenode:
Can anyone have a look on this one? It's a really simple fix. @burggraf do you know who is responsible for this type of issue?

alaister (approved):
Hey @activenode,

My apologies for the delay here – thanks for the PR!

PR #18176, opened 2023-10-13
fix: content param doesn't work in the SQL editor

ivasilov:


kevcodez (approved):
Tested it and works fine

joshenlim (approved):

PR #18186, opened 2023-10-13
feat(hooks): update default edge function url

gregnr:
Currently when you create a webhook through the dashboard, it uses the old subdomain-based functions URL in the format:

`https://<project-ref>.functions.supabase.co/<function-name>`

This updates the URL format to be path-based:

`https://<project-ref>.supabase.co/functions/v1/<function-name>`

while also providing backwards compatibility with old webhooks that use the original URL format.


laktek (approved):
LGTM

joshenlim (approved):

gregnr:
@joshenlim yes best to use the path based `/functions/v1/<function>` URL going forward. Do you know where else we reference the old subdomain-based functions url (`<ref>.functions.supabase.co/<function>`)?

joshenlim:
oh actually i dont think anywhere else uses the old URL - just did a check through the repo, so we all good! 😄

PR #18202, opened 2023-10-14
Update auth-google.mdx to fix incorrect variable name

benank:

Fixes a typo in the docs for sign in with Google on flutter.

It uses the wrong variable name.

## What is the new behavior?

It uses the right variable name.


kiwicopple (approved):

kiwicopple:
after resolving conflicts, it looks like this was solved in another PR :)

thanks for the contribution @benank !

Issue #18184, opened 2023-10-13
Code examples in api panel should use const, not let


saltcod:
We use `const` everywhere, we should use it here instead of `let`.
Applies to all these code examples in the api docs side panel.

![screenshot-2023-10-13-at-16 17 59](https://github.com/supabase/supabase/assets/105593/4dc4771a-26ef-4989-b6c1-a9c2d0ab1fa6)

shishiro26:
Hey @saltcod can you check my pr .

IbraheemHaseeb7:
is this done?

Mohammed-Yasin-Mulla:
I will be picking this up please assign it to me
I am working on this

saltcod:
addressed, thanks all!

PR #18210, opened 2023-10-15
Replace let with const in .mdx file

Mohammed-Yasin-Mulla:
Bug Fix:  Replace let with const in the docs
Issue Link : https://github.com/supabase/supabase/issues/18184#issue-1942469529


fsansalvadore (approved):
Thanks @Mohammed-Yasin-Mulla 🚀

PR #18214, opened 2023-10-15
Docs update: Improved the clarity of the deep-link set up

NathanBrodin:
In the initial documentation, it says we can choose any value we would like for the scheme, but it needs to be the exact same as the applicationId to work. (I got stuck on that error for a long time).

docs update

Initial document can lead to errors and miscomprehensions.

## What is the new behavior?

Now, it's clearer for the user what value he can choose.


dshukertjr:
Hi @NathanBrodin . I believe you actually can choose literally any scheme for deep links as long as its value is unique across all apps. Because of this reason, people usually go with the application ID, but I believe application ID matching the deep link scheme is not a requirement. Do you have any sources saying that it's a requirement?

NathanBrodin:
After more research from my side, The scheme for deep links does not need to match the applicationId. My error was probably from somewhere else.
Thanks for your answer, I will close this pull request!

PR #18221, opened 2023-10-16
fix: broken firefox scroll top behavior

benmann:
👀 See https://github.com/orgs/supabase/discussions/18154

Simple `<Link>` elements don't scroll to the top of the page after a transition propery – Firefox only

## What is the new behavior?

Simple `<Link>` elements scroll to the top of the page after a transition propery in Firefox

## Additional context

There might be other ways to solve this, if we wanna keep the transition at all costs (see discssion)


alaister:
Hi @benmann,

Thanks for your PR!

It looks like the underlying issue has been resolved, so I'm going to go ahead and close this as well.

PR #18236, opened 2023-10-16
Add custom icons to Icon library

fsansalvadore:

Add custom icons to Icon library so they are easier to use with Intellisense and all:
- IconBriefcase2
- IconChangelog
- IconDocumentation
- IconIntegrations
- IconLinkedinSolid
- IconLifeBuoy2
- IconPartners
- IconYoutubeSolid
- IconYCombinator

And use them in "Developers" nav dropdown menu and in footer.

Screenshot 2023-10-16 at 15 25 48

Screenshot 2023-10-16 at 15 24 56


apps/www/components/Nav/DevelopersDropdown.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { ReactNode } from 'react'
```
gregnr, line 1: Still needed?

gregnr (approved):

apps/www/components/Nav/DevelopersDropdown.tsx
fsansalvadore: Ah, no. Thanks for catching it @gregnr 👍

saltcod (approved):

PR #18268, opened 2023-10-17
fix project wild card route not parsing query param properly

joshenlim:


studio/pages/project/_/[[...routeSlug]].tsx
```diff
@@ -49,9 +49,10 @@ const GenericProjectPage: NextPage = () => {
       }
 
       const slugPath = slug.reduce((a, b) => `${a}/${b}`, '').slice(1)
+
       return [
         `/project/${projectRef}/${slugPath}`,
-        queryString ?? `?${queryString}`,
+        queryString.length > 0 ? `?${queryString}` : undefined,
```
Comment (kamilogorek) [inline studio/pages/project/_/[[...routeSlug]].tsx:55]: There's also same issue on line 46.

This whole logic can be rewritten into (starting on line 36):

```js
const queryString = new URLSearchParams(queryParams as Record<string, string>);

// and then in both places

return [`/project/${projectRef}`, queryString.size ? `?${queryString}` : undefined, hash ?? `#${hash}`]
```

(not sure if `hash` also needs fixing here.

kamilogorek (changes requested):

studio/pages/project/_/[[...routeSlug]].tsx
Comment (joshenlim) [reply studio/pages/project/_/[[...routeSlug]].tsx:55]: ahhh yes! good catch 🙏

studio/pages/project/_/[[...routeSlug]].tsx
```diff
@@ -33,25 +33,29 @@ const Header = () => {
...
 
   const urlRewriterFactory = (slug: string | string[] | undefined) => {
     return (projectRef: string) => {
       const hash = location.hash
 
       if (!Array.isArray(slug)) {
-        return [`/project/${projectRef}`, queryString ?? `?${queryString}`, hash ?? `#${hash}`]
+        return [
+          `/project/${projectRef}`,
+          // @ts-ignore [Joshen] https://github.com/microsoft/TypeScript/issues/54466
```
Comment (kamilogorek) [inline studio/pages/project/_/[[...routeSlug]].tsx:45]: Oh, sad :(

Looking at it again, this is same query for all urls, so you can also basically revert to updated initial version with:
```js
const queryString = Object.keys(queryParams).length
    ? `?${new URLSearchParams(queryParams as Record<string, string>)}`
    : undefined;
```

and use `queryString` as value directly, without any checks there. I'm fine with any.

You can also add:
`const hash = location.hash ? `#${location.hash}` : undefined` (or '' if desired).

and make it look pretty with `return [`/project/${projectRef}`, queryString, hash]`

kamilogorek (approved):

alaister (approved):

PR #18298, opened 2023-10-18
misc: Add project name to docker-compose.yml

kamilogorek:
This makes any project listing display a nice stack of containers under `supabase` namespace instead of default `docker` name. Useful when running multiple containers on the same machine.

Note that this can be also achieved with `docker compose up -p supabase` or `--project-name`.

image


LB22:
I did the same plus another thing in this PR https://github.com/supabase/supabase/pull/17967

PR #18301, opened 2023-10-18
feat: add UI to configure skip nonce check for Google

hf:
Adds the UI option to skip nonce checks for Google ID tokens. Useful in React Native for iOS.

Preview:

image

### DIFF
--- a/studio/stores/authConfig/schema/AuthProviders/AuthProvidersFormValidation.tsx
+++ b/studio/stores/authConfig/schema/AuthProviders/AuthProvidersFormValidation.tsx
@@ -850,6 +850,12 @@ const EXTERNAL_PROVIDER_GOOGLE = {
         'Comma separated list of client IDs of Android apps, One Tap or Chrome extensions that are allowed to log in to your project.',
       type: 'string',
     },
+    EXTERNAL_GOOGLE_SKIP_NONCE_CHECK: {
+      title: 'Skip nonce checks',
+      description:
+        "Allows ID tokens with any nonce to be accepted, which is less secure. Useful in situations where you don't have access to the nonce used to issue the ID token, such with iOS.",
+      type: 'boolean',
+    },
   },
   validationSchema: object().shape({
     EXTERNAL_GOOGLE_ENABLED: boolean().required(),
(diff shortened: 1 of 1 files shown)


studio/stores/authConfig/schema/AuthProviders/AuthProvidersFormValidation.tsx
```diff
@@ -847,6 +847,12 @@ const EXTERNAL_PROVIDER_GOOGLE = {
         'Comma separated list of client IDs of Android apps, One Tap or Chrome extensions that are allowed to log in to your project.',
       type: 'string',
     },
+    EXTERNAL_GOOGLE_SKIP_NONCE_CHECK: {
```
alaister, line 850: Does this need to be added to the `validationSchema` below?

alaister (commented):
Is this supported in the backend yet?
Screenshot 2023-10-18 at 21 05 30

studio/stores/authConfig/schema/AuthProviders/AuthProvidersFormValidation.tsx
```diff
@@ -847,6 +847,12 @@ const EXTERNAL_PROVIDER_GOOGLE = {
...
+      title: 'Skip nonce checks',
+      description:
+        "Allows ID tokens with any nonce to be accepted, which is less secure. Useful in situations where you don't have access to the nonce used to issue the ID token, such with React Native on iOS.",
```
dshukertjr, line 853: I believe this applies to all Google login attempts on iOS, and not just react native, so we could probably say something like this:
```suggestion
        "Allows ID tokens with any nonce to be accepted, which is less secure. Useful in situations where you don't have access to the nonce used to issue the ID token, such with iOS.",
```

kangmingtay:
@alaister it's not supported on the backend yet and is blocked on https://github.com/supabase/infrastructure/pull/15273#pullrequestreview-1685518677

Schotsl:
I'd love to see this merged!

iykazrji:
+ 1 

Please! 🙏🏾

hf:
@alaister @dshukertjr addressed comments.

alaister (approved):
LGTM!

Mykyta-Chernenko:
@alaister, @hf 
I use flutter,   supabase_flutter: ^1.10.25 and sign_in_with_apple: ^5.0.0 
I enabled skip nonce check 
image

and I call it like this
```
var user = supabase.auth.signInWithIdToken(
                    provider: Provider.google,
                    idToken: idToken,
                    accessToken: accessToken,
                  );
```

However, i still get "AuthException(message: Passed nonce and nonce in id_token should either both exist or not., statusCode: 400)"                

Is there something I need to update/upgrade for skip nonce check to work?

PR #18302, opened 2023-10-18
fixes quotes in search result (using cmdk) crashes the application

### LINKED ISSUE #18276: Quotes in search result (using `cmdk`) crashes the application
# Bug report

## Describe the bug

If a result to your search query (not the __query__ itself, as reported in #16506) contains quotes (don’t think it matters if it’s single or double), then the app crashes. The screencast shows the `DOMException` that causes the app to crash, but I have also provided additional context below.

## To Reproduce

See the screencast below:

https://github.com/supabase/supabase/assets/14703164/ce56629d-a626-41d9-9ceb-fc54f7490881

## Expected behavior

I should be able to hover over the result and click on it.

## Screenshots

N/A

## System information

- OS: N/A
- Browser: Brave
- Version of supabase-js: N/A
- Version of Node.js: N/A

## Additional context

I reported the error on pacocoursey/cmdk#189 because I __believe__ the issue exists in the library itself, but because it is causing the app to crash, I thought it would be useful to report it here as well. I know you guys use a forked version of `cmdk`, so maybe the fix can be applied to the forked version directly. LMK if you need more info.

YugBhanushali:

If the search results contain quotes, then the app does not crash when the user hovers over the results, and during click on the result the user is redirected to the corresponding documents.

Currently, when the user hovers over a search result containing quotes, it leads to an app crash.

https://github.com/supabase/supabase/assets/95876621/bc27a727-f2d0-44c7-b43d-42e54f862515

## What is the new behavior?

User is able to hover and is able to click on the search results containing quotes and gets redirected to corresponding docs.

https://github.com/supabase/supabase/assets/95876621/1c3f3025-9fd7-47e1-9596-946b93917226

## Additional context

Fixes #18276


packages/ui/src/components/Command/DocsSearch.tsx
```diff
@@ -205,7 +210,7 @@ const DocsSearch = () => {
                         openLink(page.type, formatSectionUrl(page, section))
                       }}
                       key={`${page.meta.title}__${section.heading}-item-index-${i}`}
-                      value={`${page.meta.title}__${section.heading}-item-index-${i}`}
+                      value={`${removeDoubleQuotes(page.meta.title)}__${section.heading ? removeDoubleQuotes(section.heading) : ""}-item-index-${i}`}
```
fsansalvadore, line 213:
```suggestion
                      value={`${removeDoubleQuotes(page.meta.title)}__${removeDoubleQuotes(section.heading ?? '')}-item-index-${i}`}
```

fsansalvadore: you could shorthand it like this maybe? :)

YugBhanushali: Will update it

fsansalvadore: If you update that part then we can merge. Thanks, @YugBhanushali 👍

fsansalvadore: Nice, thanks!

fsansalvadore (approved):
Awesome, thanks @YugBhanushali!

PR #18311, opened 2023-10-18
feat: partner billing plan upgrades

kevcodez:
Initial support for partner billing to allow upgrades without credit card enforcement + slight copy adjustment.

This is just to unblock further testing, proper disabling of features+more changes are coming in separate PRs.


studio/components/interfaces/Organization/BillingSettingsV2/Subscription/PlanUpdateSidePanel.tsx
```diff
@@ -73,6 +73,9 @@ const PlanUpdateSidePanel = () => {
     }
   )
 
+  const billingViaPartner = subscription?.billing_via_partner === true
```
joshenlim, line 76: im guessing this requires an API change as well? but still safe to go through i guess with the `?`

joshenlim (approved):

PR #18314, opened 2023-10-18
Migrate to styling tokens

fsansalvadore:

- Migrate styling to tokens system in:
  - www
  - docs
  - studio
  - packages
- Add GUI to test and update tokens in sandbox mode
  - Shows a GUI in dev and preview env if you append `#theme-sandbox` to the preview url (or locally)
  - Doesn't show up in prod
- Consolidated InteractiveShimmerCard into Panel component


apps/www/components/Blog/BlogListItem.tsx
```diff
@@ -30,7 +30,7 @@ const BlogListItem = ({ post }: Props) => {
           <div className="flex flex-col space-y-6">
             <div className="flex flex-col space-y-3">
               <div
-                className={`border-scale-300 relative mb-4 h-60 w-full overflow-auto rounded-lg border shadow-sm`}
+                className={`border-border relative mb-4 h-60 w-full overflow-auto rounded-lg border shadow-sm`}
```
MildTomato, line 33: border-default ?

MildTomato, line 33: can just be `border`

are you thinking find and replace will be hard?

apps/www/components/Card.tsx
```diff
@@ -56,21 +56,21 @@ const Card = (props: Props) => {
       target="_blank"
       className="relative flex transform flex-col overflow-hidden rounded-md shadow-lg transition hover:-translate-y-1 hover:shadow-2xl"
     >
-      <div className="dark:bg-dark-800 flex-shrink-0">
-        <img className="h-64 w-full object-cover" src={imgUrl} alt={title} />
+      <div className="flex-shrink-0">
+        <img className="h-64 w-full text-light object-cover" src={imgUrl} alt={title} />
```
MildTomato, line 60: `text-foreground-light`

apps/www/components/Card.tsx
```diff
@@ -56,21 +56,21 @@ const Card = (props: Props) => {
...
-      <div className="dark:bg-scale-300 flex flex-1 flex-col justify-between bg-white p-8">
+      <div className="bg-surface-100 flex flex-1 flex-col justify-between p-8">
         <div className="flex-1">
           {/* {icons && (
             <div className="-mt-12 mb-6 flex -space-x-3">
               {iconMarkup}
             </div>
           )} */}
-          <p className="font-base dark:text-dark-300 text-sm capitalize text-gray-600">{type}</p>
+          <p className="font-base text-lighter text-sm capitalize">{type}</p>
```
MildTomato, line 69: `text-foreground-lighter`

apps/www/components/Card.tsx
MildTomato: thought we were following the 'foreground-x' convention ?

apps/www/components/Blog/BlogListItem.tsx
fsansalvadore: But let's say you only want `border-t`, you would have to write `border-t border-border`. If you write `border-t border`, the second "border" would set the border on all 4 sides.

fsansalvadore: So we could just probably have the shorthand `border` when it's all 4 borders and we know it's the default color.

apps/www/components/Card.tsx
fsansalvadore: You're right, it's probably better to keep it more explicit to reduce confusion. I'll update it 👍

fsansalvadore: On the other hand, doing a "Find all" in vscode reveals that there are already several occurrences where `text-light` and `text-lighter` have been used across docs/studio/www. This might be a symptom that actually omitting the `foreground` part when used on text might be good UX. Or developers tend to prefer writing that shorthand for velocity. Wdyt? @MildTomato

MildTomato: I reckon the thing that caused the issue was just the default.. so `text`, so maybe we just deal with that and just use `text-default` so its easy to find, and then we still get to use shorthand everywhere else.

I also picked up on others using the shorthand 😂 might as well lean into it.

MildTomato: I do personally like shorthand :D

fsansalvadore: Ah, for just `text` I leaned on the `text-foreground` shorthand 😅 It felt more clear to what it was referring to.

apps/www/components/Blog/BlogListItem.tsx
fsansalvadore: But it could be actually `border-default` instead of `border-border`. Much better...

fsansalvadore:
> can just be border

@MildTomato not sure why my comments on this don't show in the Conversation tab but only do in the "Files changed" tab...

I'll repeat here: `border` actually sets the 4 borders (as the tw class does). Let's say you would want `border-t border-transparent` and it be not transparent on hover, you'd need to do `border-t border-transparent hover:border-border`.

But in fact it could actually be `border-default` instead of border-border. Much better... So it would be `border-t border-transparent hover:border-default`.

apps/www/components/Blog/BlogListItem.tsx
MildTomato: I think we should stay away from double barrel class names all together, `border-border` will give me nightmares

apps/www/components/Card.tsx
MildTomato: shall we lean hard into one or the other?

mixing -foreground with shorthand can make it more confusing.

apps/www/components/Blog/BlogListItem.tsx
fsansalvadore: > border-border is the same as border right?
Not exactly. `border` is [this tailwind utility](https://tailwindcss.com/docs/border-width#all-sides) to set all sides.

fsansalvadore: I agree `border-border` is very ugly. But `border-default` should be a good solution.

apps/www/components/Card.tsx
fsansalvadore: `text-default` doesn't seem to work though.

fsansalvadore: Nevermind, I keep forgetting it's `text-DEFAULT`.

fsansalvadore: Though I think I'd rather prefer to keep `text-foreground-light` etc not shorthand, but have `text-foreground` instead of `text-default`. For the sake of consistency, clarity and legibility.

apps/www/components/Blog/BlogListItem.tsx
fsansalvadore: Hm for some reason `border-DEFAULT` doesn't work though...

fsansalvadore: These don't produce any output:

Screenshot 2023-10-23 at 18 37 31

This one does:

Screenshot 2023-10-23 at 18 37 10

apps/www/components/Card.tsx
MildTomato: > text-default doesn't seem to work though. (nor does text-DEFAULT)

we could make it work. probably an overshight.

let's go with what you prefer though.

apps/docs/components/reference/ApiOperationSection.tsx
```diff
@@ -96,7 +96,7 @@ const ApiOperationSection = (props) => {
           >
             {operation.responseList.map((response: any) => (
               <Tabs.Panel id={response.responseCode} label={response.responseCode}>
-                <p className="text-scale-1000">{response.description}</p>
+                <p className="text-background0">{response.description}</p>
```
ivasilov, line 99: Misclick probably?

studio/components/ui/WarningBanner.tsx
```diff
@@ -4,14 +4,12 @@ const WarningBanner = ({ title, description }: { title: string; description?: st
   return (
     <div className="rounded bg-yellow-900 px-6 py-3">
       <div className="flex flex-row gap-4">
-        <div className="mt-1 text-yellow-1100 dark:text-yellow-800">
+        <div className="mt-1 text-yellow-900">
```
ivasilov, line 7: Maybe use `text-warning` here?

apps/docs/components/reference/ApiOperationSection.tsx
fsansalvadore: nice catch!

studio/components/ui/WarningBanner.tsx
MildTomato: yea switch to warning :D

MildTomato:
> 
> @MildTomato not sure why my comments on this don't show in the Conversation tab but only do in the "Files changed" tab...
> 
> I'll repeat here: just border actually sets the 4 borders (as the tw class does). Let's say you would want border-t border-transparent and it be not transparent on hover, you'd need to do border-t border-transparent hover:border-border.
> 
> But in fact it could actually be border-default instead of border-border. Much better... So it would be border-t border-transparent hover:border-default.
> 

Fixed it in a PR to support default classes

https://github.com/supabase/supabase/pull/18688

studio/components/ui/WarningBanner.tsx
fsansalvadore: ok updated 👍

fsansalvadore:
@MildTomato all `border-border`s are now `border-default`. ✅
If you can double-check that everything looks good, let's merge on Monday?

MildTomato (approved):

MildTomato (changes requested):
package-lock has been changed

apps/docs/package.json
```diff
@@ -57,6 +57,7 @@
     "@supabase/supabase-js": "^2.38.2",
     "common": "*",
     "config": "*",
+    "dat.gui": "^0.7.9",
```
MildTomato, line 60: ohh i guess this is why package-lock changed

MildTomato: @fsansalvadore if this is fine, then good to merge

fsansalvadore: yeah it's the GUI for the sandbox mode

MildTomato: nice :D

MildTomato (approved):

PR #18366, opened 2023-10-20
Chore/table editor header simplification

saltcod:
1. RLS off warning 
screenshot-2023-12-19-at-16 47 31

2. RLS off warning, dismissed 

screenshot-2023-12-19-at-16 48 01

3. Enable RLS from table editor
screenshot-2023-12-19-at-16 49 30

screenshot-2023-12-19-at-16 49 48

4. Data | Definition toggle moved to the bottom
screenshot-2023-11-06-at-21 13 33

 
6. RLS enabled, no policies: 
 
screenshot-2023-12-19-at-16 51 32

7. RLS enabled, with policies:
 
screenshot-2023-12-19-at-16 51 07


studio/components/grid/components/footer/Footer.tsx
```diff
@@ -1,14 +1,63 @@
...
+    } else {
+      setUrlState({ view })
+    }
+  }
+
   return (
-    <div className="sb-grid-footer">
-      <div className="sb-grid-footer__inner">
-        <Pagination isLoading={isLoading} />
+    <div className="bg-scale-100 px-2 w-full">
```
MildTomato, line 40: rogue div.

just use 1?

studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -137,29 +137,58 @@ const GridHeaderActions = ({ table, openAPIDocsPanel, refreshDocs }: GridHeaderA
             </Tooltip.Portal>
           </Tooltip.Root>
         )}
+        <Tooltip.Root delayDuration={0}>
+          <Tooltip.Trigger className="w-full">
+            {(table.rls_enabled || showRLSWarning) && (
+              <Link href={`/project/${projectRef}/auth/policies?search=${table.id}`}>
```
MildTomato, line 143:
```
<Link passHref href={} {...}>
  <Button asChild>
    <a>
```

studio/components/grid/components/footer/Footer.tsx
```diff
@@ -1,14 +1,63 @@
...
+      <div className="flex h-10 items-center">
+        {selectedView === 'data' && <Pagination isLoading={isLoading} />}
+
+        {(isTableSelected || isViewSelected) && (
+          <>
+            {/* {canEditViaTableEditor && (
```
MildTomato, line 46: comments

studio/components/grid/components/header/RLSBannerWarning.tsx
```diff
@@ -46,10 +46,10 @@ export default function RLSBannerWarning() {
...
       {!isAcknowledged && !rlsEnabled && isPublicTable ? (
         <div>
-          <div className="text-center bg-amber-500 text-amber-1100 dark:text-amber-900 text-xs py-2.5 flex items-center justify-center relative">
+          <div className="text-center text-xs py-2.5 flex items-center justify-center relative">
             <IconAlertCircle size={16} strokeWidth={2} />
-            <span className="uppercase font-bold ml-2">Warning</span>: You are allowing anonymous
-            access to your table.{' '}
+            <span className="font-bold mx-2">Table is public</span> You are allowing anonymous
+            access to your data.{' '}
             <Link href={`/project/${projectRef}/auth/policies?search=${tableID}`}>
```
MildTomato, line 53: i know you didn't touch this line, but maybe throw a passHref in

MildTomato (changes requested):

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -132,28 +169,70 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
...
+            <>
+              {table.rls_enabled ? (
+                <div className="flex items-center gap-1">
+                  <Link passHref href={`/project/${projectRef}/auth/policies?search=${table.id}`}>
+                    <Button
+                      type={table.rls_enabled ? 'default' : 'warning'}
+                      className="group"
+                      icon={
+                        policies.length > 0 ? (
+                          <span>
```
MildTomato, line 183: Span in a span

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -132,28 +169,70 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
...
-          </Button>
-        )}
+                <Popover_Shadcn_ open={open} onOpenChange={() => setOpen(!open)} modal={false}>
+                  <PopoverTrigger_Shadcn_ asChild>
+                    <Button type="warning" icon={<IconLock size={14} />}>
+                      RLS Disabled
+                    </Button>
+                  </PopoverTrigger_Shadcn_>
+                  <PopoverContent_Shadcn_ className="w-48 min-w-[395px]">
+                    <div className="grid gap-2 text-sm px-2 py-4">
```
MildTomato, line 207: This div isn't needed. Move classes to content component

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -132,28 +169,70 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
...
+                      <div className="grid gap-2 mt-1">
+                        <p>
+                          You can restrict and control who can read, write and update data in this
+                          table using Row Level Security.
+                        </p>
+                        <p>
+                          With RLS enabled, anonymous users will not be able to read/write data in
+                          the table.
+                        </p>
+                        <p className="mt-2">
```
MildTomato, line 220: Button doesn't need to be in a p

apps/studio/components/grid/components/footer/Footer.tsx
```diff
@@ -1,14 +1,55 @@
...
+    } else {
+      setUrlState({ view })
+    }
+  }
+
   return (
-    <div className="sb-grid-footer">
-      <div className="sb-grid-footer__inner">
-        <Pagination isLoading={isLoading} />
+    <div className="px-2 w-full">
```
MildTomato, line 37: I'm viewing this on my phone, but this div looks like a superfluous div

MildTomato: Move classes to next div and remove the extra

apps/studio/components/grid/components/header/Header.tsx
```diff
@@ -117,7 +116,6 @@ const DefaultHeader = ({
   return (
     <div className="flex items-center gap-4">
       <div className="flex items-center gap-2">
```
MildTomato, line 118: Should be using correct markup here. Like role="menu"

Can deal with in a follow up PR

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -105,6 +120,28 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
...
+
+  const onToggleRLS = async () => {
+    const payload = {
+      id: table.id,
+      rls_enabled: !table.rls_enabled,
+    }
+
+    const res: any = await meta.tables.update(payload.id, payload)
+    if (res.error) {
+      return ui.setNotification({
```
MildTomato, line 135: General comment - I think we are now using error toast messages wrongly as a default. It was meant to be a fallback rather than a consistent pattern. 

Going forward I think we should be aiming towards inline error messages that are actually in the UI you're interacting with.

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -132,28 +169,70 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
...
         )}
-
-        {(table.rls_enabled || showRLSWarning) && (
-          <Button
-            asChild
-            type={table.rls_enabled ? 'link' : 'warning'}
-            icon={
-              table.rls_enabled ? (
-                <IconLock strokeWidth={2} size={14} />
+        <div>
```
MildTomato, line 172: Possible random div. Hard to tell on phone though.

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -132,28 +169,70 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
...
+                            <span className="text-right text-xs rounded-xl px-2 py-0.5 bg-surface-200 dark:bg-surface-100 text-brand-1100">
+                              {policies.length}
+                            </span>
+                          </span>
+                        ) : (
+                          <IconPlusCircle size={12} />
+                        )
+                      }
+                    >
+                      <span className="flex-1 text-left">
```
MildTomato, line 193: Is this needed ? Classes might work in button component

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -132,28 +169,70 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
...
-        )}
+                <Popover_Shadcn_ open={open} onOpenChange={() => setOpen(!open)} modal={false}>
+                  <PopoverTrigger_Shadcn_ asChild>
+                    <Button type="warning" icon={<IconLock size={14} />}>
+                      RLS Disabled
+                    </Button>
+                  </PopoverTrigger_Shadcn_>
+                  <PopoverContent_Shadcn_ className="w-48 min-w-[395px]">
+                    <div className="grid gap-2 text-sm px-2 py-4">
+                      <h3 className="flex items-center gap-2">
```
MildTomato, line 208: These classes don't do anything ?

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -204,6 +283,17 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
...
         </Modal.Content>
       </ConfirmationModal>
+
+      <ConfirmModal
+        danger={table.rls_enabled}
+        visible={rlsConfirmModalOpen}
+        title="Confirm to enable Row Level Security"
+        description="Are you sure you want to enable Row Level Security for this table?"
+        buttonLabel="Confirm"
+        buttonLoadingLabel="Saving"
```
MildTomato, line 293: "Updating" ?

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -204,6 +283,17 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
           )}
         </Modal.Content>
       </ConfirmationModal>
+
+      <ConfirmModal
+        danger={table.rls_enabled}
+        visible={rlsConfirmModalOpen}
+        title="Confirm to enable Row Level Security"
+        description="Are you sure you want to enable Row Level Security for this table?"
+        buttonLabel="Confirm"
```
MildTomato, line 292: Could be more affirmative? 
"Enable RLS"

MildTomato (commented):
I reviewed this on my phone so some of the comments might be bad. Haven't looked at the preview yet.

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
MildTomato: Might be because the button stretches.

In which case, change to use flex rather than grid tailwind.

Or use div rather than p, makes more semantic sense

apps/studio/components/grid/components/header/Header.tsx
MildTomato: https://www.w3.org/WAI/tutorials/menus/application-menus/

Probably this is the right implantation for this. Although this is a bit of a guess

MildTomato: Worth mentioning the <Trigger> components for things like Popover already have aria labels for things like "haspopup", but you'll need to probably set role="menuitem" on them

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
saltcod: annoyingly addressed with !h-[28px] !py-0

ivasilov (commented):
I did a quick test and found if you go to the definition for a table you can't go back via the same widget (two-option button?).

joshenlim:
Some observations (apologies in advanced if i missed any decisions from the thread! its a long one, might have missed something)

### RLS warning banner
- Did we decide to remove this one? I created a new table with RLS off but i don't see the banner

### Data/Definition toggle
- Can we retain the same styling between grid and definition view? Seems like theres a small padding difference + no border top for the latter
image
image

- Also can we keep the data/definition toggle bottom bar rendered while switching between both views? noticed that the bottom bar disappears while the definition is loading

### Auth policy button after enabling RLS
- Should we just put 0 Auth policies here instead of the AddCircleIcon button? IMO the icon feels weird in the context of a warning (i'd think it might be better in a default or primary button)
image

- This is a consequence i think from the increasing the RQ stale time but enabling RLS through the table editor, doesnt update the auth/policies page if you navigate to that from the table editor after enabling RLS. Missing some query invalidation

- Did we decide on using the language "Auth policies" over "RLS policies"? I remember we had a convo during one of the syncs but I can't recall on the decision. Just worried we might be confusing users as "policies" are a database feature, not an "Auth" product feature (and there's also the separate conversation of whether RLS should be under database or Auth but that one's for next time) Not to mention our docs reference policies to RLS, rather than Auth

joshenlim:
@saltcod footer just slightlyyyyyy off

https://github.com/supabase/supabase/assets/19742402/4f461157-17a0-4b86-9cf9-1d724b329458

saltcod:
@joshenlim sorry, meant to mention  — honestly can't figure that out! I tried everything yesterday and just couldn't see how that was happening

saltcod:
**Changes**: 
— "Auth policy" is now:   “Add RLS policy”
— If RLS enabled and no policies, the button is default

Here's the current states: 
screenshot-2024-02-12-at-13 55 48


@joshenlim on this one:
> consequence i think from the increasing the RQ stale time but enabling RLS through the table editor, doesnt update the auth/policies page if you navigate to that from the table editor after enabling RLS
This is quite weird. The change does happen, but there's a 1-2 second delay for me locally. Oddly, on master, I get the same delay. Have you seen issues with stale time like this before? This is weird, it seems like the queries are being invalidated properly, but the ui is taking 1-2s to update. 

Extra weird that these are using the same function, which definitely does the invalidation correctly, or I'm missing something silly
screenshot-2024-02-12-at-14 01 00

https://github.com/supabase/supabase/blob/master/apps/studio/data/tables/table-update-mutation.ts#L59

joshenlim:
>Extra weird that these are using the same function, which definitely does the invalidation correctly, or I'm missing something silly

Okay i see whats happening - both are using the same mutation, but both pages are using different queries
the table editor uses the `useTable` hook which innately uses the `useTableQuery`, where policies page uses the `useTablesQuery`

Hence why although both are using the same mutation, only useTableQuery is being invalidated

I'll jump on the branch

joshenlim:
Adding an alert in the auth policies page here RE RLS enabled but no policies, same language as GridHeaderActions
image

joshenlim:
> footer just slightlyyyyyy off

Just pushed a fix for this too 🙏

joshenlim (approved):

joshenlim:
looks good to me otherwise :)

MildTomato:
> "Failed to toggle RLS"

I think we could strive for better error messages. like:

"Failed to enable RLS for `countries` table"

MildTomato (approved):
You don't need my approvals for these 😄 

but since you asked, here it is 🚢

PR #18414, opened 2023-10-23
Fix: prettier newline split within JSX tags

gregnr:
## Problem
`prettier` will wrap long text in markdown so that it fits within a max line length (ie. 100). This works great for regular markdown since it will concatenate these lines back together upon render, however the MDX spec doesn't support this due to nuances when parsing JSX syntax.

The problem is that `prettier` still wraps long text in MDX anyway in certain situations, and this results in explicit newlines in the final render when we don't want them. Specifically this happens when you add text directly inside a JSX tag:
```markdown
<Admonition type="note">
  If you get stuck while working through this guide, refer to the [full example on
  GitHub](https://github.com/supabase/supabase/tree/master/examples/user-management/expo-user-management).
</Admonition>
```
image

Notice how `prettier` wrapped the text between "full example on" and "GitHub", which resulted in an explicit newline in the final render.

## Solution
Stop `prettier` from wrapping text within a JSX tag by adding a newline character at the beginning and end of the JSX tag:
```markdown
<Admonition type="note">

If you get stuck while working through this guide, refer to the [full example on GitHub](https://github.com/supabase/supabase/tree/master/examples/user-management/expo-user-management).

</Admonition>
```
image


kamilogorek (approved):
What kind of magic is this?! 👀

kangmingtay (approved):

kangmingtay:
is it possible to specify some sort of config in prettier to ignore formatting all text wrapped in `<Admonition></Admonition>` ?

kamilogorek:
Unfortunately this is actual recommendation from docusaurus to just add those lines yourself - https://docusaurus.io/docs/next/markdown-features/admonitions#usage-with-prettier

There's nothing for `eslint` itself, and we'd probably need to integrate `remark` on top of what we have now. Not worth it imo.

gregnr:
@kangmingtay unfortunately not (that I'm aware of). Prettier is [intentionally opinionated](https://prettier.io/docs/en/option-philosophy) and doesn't give too many options, so you're more or less limited to these:
https://prettier.io/docs/en/options.html

PR #18420, opened 2023-10-24
Chore/update network restrictions

joshenlim:


studio/components/interfaces/Settings/Database/BannedIPs.tsx
```diff
@@ -64,7 +64,7 @@ const BannedIPs = () => {
     <div id="banned-ips">
       <div className="flex items-center justify-between">
         <FormHeader
-          title="Banned IPs"
+          title="Network Bans"
           description="We monitor unsuccessful logins and block IPs for security"
```
MildTomato, line 68:
```suggestion
          List of IP addresses that are temporarily blocked if their traffic pattern looks abusive.
```

studio/components/interfaces/Settings/Database/NetworkRestrictions/AddRestrictionModal.tsx
```diff
@@ -34,7 +34,9 @@ const AddRestrictionModal = ({
 
     // Validate CIDR block size
     const isOutOfCidrSizeRange = cidrBlockSize < 0 || cidrBlockSize > 32
-    if (isOutOfCidrSizeRange) errors.cidrBlockSize = 'Size has to be between 0 to 32'
+    if (cidrBlockSize.length === 0 || isOutOfCidrSizeRange) {
```
alaister, line 37: I think 0 may be valid value actually – maybe check with infra?

joshenlim: yeap 0 is a valid value - the additional check here just verifies if the input field is empty

joshenlim: so validation will say "no" if input field is empty, or a number below 0 or higher than 32 is given

alaister: ah my bad! nice

alaister (approved):
LGTM!

PR #18431, opened 2023-10-24
Chore/upgrade next 13 fix table editor

joshenlim:


alaister (commented):
Small nit: this doesn't look full-width if there are limited columns:
Screenshot 2023-10-25 at 17 13 25

alaister (approved):
but otherwise, 🔥🔥🔥

saltcod:
I'm pretty regularly able to cause this tab to crash when I use the side panel
https://studio-staging-git-chore-upgrade-next-13-fix-ta-656062-supabase.vercel.app/dashboard/project/_/editor/

Creating a new column or a new table opens the side panel and every 3rd or 4th time, it will cause the tab to crash. 

happening to anyone else? 

screenshot-2023-10-25-at-11 17 25

joshenlim:
> Small nit: this doesn't look full-width if there are limited columns:

yeah this one is tough to fix properly for now... react data grid moved to CSS grids so i can't force it to expand the entire width now. in fact at best it can only span to all the columns

which means if your table has little columns, it'll look squashed
flip side, if your table has many many columns, it'll look huge

still trying to figure out a better solution

PR #18436, opened 2023-10-24
docs: remove Realtime client-side throttling

w3b6x9:

docs update

## What is the new behavior?

removes any mention of Realtime client-side throttling

## Additional context

accompanying pr: https://github.com/supabase/realtime-js/pull/260


filipecabaco (approved):

charislam (approved):
Can you also delete the MDX file? The page will still be built as long as the file exists, even if we don't link to it

PR #18470, opened 2023-10-25
ref: Fetch all GitHub repositories for GitHub integration config page

kamilogorek:
Depends on https://github.com/supabase/infrastructure/pull/15346 and requires recreating openapi types.


alaister:
Please let me know once https://github.com/supabase/infrastructure/pull/15346 has been deployed to prod and we'll get this out ASAP 💪🏻

alaister (approved):
Didn't think I'd be approving a PR with `while (true)` in it in this day and age, but that is a good use of it 😂

PR #18479, opened 2023-10-25
docs: Fix SvelteKit Auth-Helper let-const error

### LINKED ISSUE #18393: Tutorial results in error: You are assigning to a const
Fairly new to Svelte and some of its syntax, but was following the [linked tutorial](https://supabase.com/docs/guides/auth/auth-helpers/sveltekit?language=js#:~:text=const%20%7B%20supabase%2C%20session%20%7D%20%3D%20data) and it gives an error in VSCode: "You are assigning to a const"

The particular piece giving the error is:

```
const { supabase, session } = data
$: ({ supabase, session } = data)
```

Looks to have been [a commit](https://github.com/supabase/supabase/commit/b09243b628c4f47c98fff0d89c08f561865ae56b) changing it from "let" to "const", and when I switch it back, it seems to work. Whilst it would appear that commit was changing a lot of other let references to const, would like to understand if that is the correct thing to do in this scenario, or if it should be handled in another way?

Creative-Difficulty:

docs

The tutorial throws an error as described in #18393 

## What is the new behavior?

The error is fixed

## Additional context
Closes #18393 


silentworks:
There are no file changes here in this PR.

PR #18484, opened 2023-10-26
Fix: Add force-dynamic to quickstart

dijonmusters:

docs update

Quickstart fails when deployed as Next.js tries to statically generate page

## What is the new behavior?

Quickstart does not fail when deployed as we explicitly mark the route as `dynamic`

## Additional context

https://www.reddit.com/r/nextjs/comments/177o5l1/deployment_errors_cookie_related_supabase_nextjs/


alaister (approved):

dijonmusters:
@silentworks pointed out that calling the `cookies` function from the route flags to Next.js that this is a dynamic route, therefore, shouldn't need `export const dynamic = 'force-dynamic'`

PR #18508, opened 2023-10-26
Chore/Table editor menu design iteration

monicakh:

Iterating on the design of the table editor menu (Work in Progress)


studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
```diff
@@ -224,11 +200,35 @@ const TableEditorMenu = () => {
...
-                      <p>Tables</p>
-                      {totalCount !== undefined && (
-                        <p style={{ fontVariantNumeric: 'tabular-nums' }}>({totalCount})</p>
-                      )}
+                      {/* Table search input */}
+                      <div className="mb-2 block relative">
+                        <div className={`relative table-editor-search border-none'}`}>
+                          {isSearchExpanded ? (
+                            <div className="flex items-center">
+                              <Input
```
MildTomato, line 208: Probably want to always have an Input.

so rather than the SearchIcon being replaced with an Input, its actually just an Input with CSS classes changing on it.

MildTomato: initially I figured you'd do something like:

<Input className="w-8 !border-transparent focus:w-full focus:border"/>

although that won't help for when the input becomes unfocused, but has a value.

so maybe 

<Input className={cn(
"w-8 !border-transparent",
searchText && "!w-full !border"
)}/>

MildTomato: Feel free to use `Input_Shadcn_` component, might be easier. will be easier to control.

```jsx

import {Input_Shadcn_} from 'ui'
import {Search} from 'lucide-react'

<Input_Shadcn_ className={cn(
  "w-8 !border-transparent",
  searchText && "!w-full !border"
)}>
  <Search/>
</Input_Shadcn_>
```

monicakh: It doesnt look like we can include the `<Search/>` icon inside `Input_Shadcn_` component, I get a blank page.

PR #18509, opened 2023-10-26
[wip] docs: new auth feature about user sessions

hf:
Docs about a new feature for User Sessions. 

Don't merge until feature has been released!

### DIFF
--- a/apps/docs/pages/guides/auth/sessions.mdx
+++ b/apps/docs/pages/guides/auth/sessions.mdx
@@ -10,7 +10,7 @@ Supabase Auth provides fine-grained control over your user's sessions.
 
 A session is created when a user signs in. It is represented by the Supabase Auth access token in the form of a JWT, and a refresh token which is a unique string.
 
-Access tokens are designed to be short lived, usually between 5 minutes and 1 hour. You can exchange a refresh token only once to get a new access and refresh token pair.
+Access tokens are designed to be short lived, usually between 5 minutes and 1 hour. Refresh tokens only can be used once and never expire. You can exchange a refresh token to get a new access and refresh token pair.
 
 This process is called **refreshing the session.** 
 
@@ -19,10 +19,10 @@ By default, there's no limit to the number of sessions a user can have on Supaba
 - By restricting the number of sessions a user can have.
 - By controlling the lifetime of a session.
 
-A session can terminate, depending on configuration, when:
+A session can be revoked, depending on configuration, when:
 
 - The user clicks sign out.
-- The user changes their password or performs a security sensitive action.
+- The user changes their password or performs a security sensitive action such as enabling multi-factor authentication (MFA).
 - It times out due to inactivity.
 - It reaches its maximum lifetime.
 - A new session for the same user was started.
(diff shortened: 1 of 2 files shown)


apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+<Admonition type="note">
+
+This feature is only available on Pro plans and up.
+
+</Admonition>
+
+Supabase Auth can be configured to limit the lifetime of a user's session. By default, all sessions are active until the user clicks log out or performs some other action that terminates a session.
+
+In some applications, it's useful or required from a security point of view to ensure that users authenticate often, or that sessions are not left active on devices for too long.
```
kangmingtay, line 40:
```suggestion
This is useful for more security sensitive applications to ensure that users are re-authenticated often so that sessions are not left active on devices for too long.
```

apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+There are two ways in which you can limit the lifetime of a session:
+
+- By enabling timeboxed sessions which terminate after a fixed amount of time.
+- By enabling inactivity timeout, such that if a session remains unrefreshed for too long it will be terminated.
```
kangmingtay, line 45: For the second point, do you mean that the session will only be revoked if there is no user activity detected after a certain time or do you mean it will be revoked if it hasn't been refreshed within the timeout? 

```suggestion
- By enabling time-boxed sessions which are revoked after a fixed amount of time.
- By enabling timeout after inactivity, such that a session will be revoked if no user activity has been detected.
```

apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+To make sure that users are required to re-authenticate periodically, you can select the _Enforce timeboxed sessions_ option in the Auth configuration screen for your project.
+
+To make sure that sessions expire after a period of inactivity, you can select and configure the _Terminate inactive sessions_ option.
```
kangmingtay, line 49:
```suggestion
To make sure that sessions expire after a period of inactivity, you can select and configure the _Revoke inactive sessions_ option.
```

apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+Sessions are not proactively terminated when you change these settings, but rather the check is enforced whenever a session is refreshed next.
```
kangmingtay, line 51:
```suggestion
Sessions are not proactively revoked when you change these settings, but rather the check is enforced whenever a session is refreshed next.
```

apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## Restricting the number of sessions per user
+
+<Admonition type="note">
+
+This feature is only available on Pro plans and up.
+
+</Admonition>
```
kangmingtay, line 59: do we need this admonition here as well? if both of these features are only available on Pro plans and up, then we should move it to the top

apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+Supabase Auth can be configured to restrict the number of sessions per user.
+
+When your application's requirements demand that a user does not have more than one active session at once, you can enforce this by selecting the _Single session per user_ option in the Auth configuration page for your project.
+
+Sessions are not proactively removed when you toggle this setting. Instead, on the next session refresh the remaining of the user's sessions will be terminated.
```
kangmingtay, line 65: Reworded this section but i think we can actually proactively remove the session - will elaborate on it more in our internal chat

```suggestion
Sessions are not proactively removed when you toggle this setting. Instead, the current session will only be revoked on the next refresh. This is due to the property of JWTs - a JWT cannot be revoked and is valid for the duration of it's expiry unless the JWT secret is rolled.
```

apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+Enforcing only a single session per user can sometimes be too strict, especially if you're building an app for multiple devices. In those cases it's often necessary to enforce a single session per device per user -- one session on iOS, one on Android, one on web, etc.
+
+Supabase Auth lets you do this by assigning a **tag** to a session on sign in.
+
+You can configure the allowed tags in the Auth configuration page for your project. You can choose as many or as few tags as you want depending on your use case.
+
+For example, if you're building an app for iOS, Android, Desktop and Web, natural choices for tags could be: `ios`, `android`, `desktop` and `web`.
+
+Finally, you can configure the session tag when creating the client library. This example shows how you can do this with the `web` tag. If you do not specify the tag in the client, the first tag in the list will generally be used.
```
kangmingtay, line 75: Reminder to add a link to the JS reference when it's ready :)

apps/docs/pages/guides/auth/sessions.mdx
hf: Yeah. When we roll out with the session management APIs we can think about it differently I guess.

apps/docs/pages/guides/auth/sessions.mdx
hf: Actually I wouldn't prefer to explain this again, as it's explained at the start of the document.

J0 (dismissed):

kangmingtay:
@J0 @hf let's not approve this PR until we have shipped everything else in case it gets accidentally merged 😬 can we also move it to a draft too?

PR #18526, opened 2023-10-27
studio: fixed dropping enum types

### LINKED ISSUE #18468: In Studio, dropping an enum type omits the schema
# Bug report
When dropping an enum type from Studio, the schema is not included in the drop statement. If two types have the same name in different schemes the `search_path` will determine which one gets deleted.

You can reproduce this issue by mirroring a builtin type with a user defined type of the same name. For example, if we define an enum type named `public.bit` (there is a builtin type named `pg_catalog.bit`) it is not possible to delete it from the Studio UI since pg_catalog is first on the `search_path`

```json
{
    "error": "ERROR:  2BP01: cannot drop type bit because it is required by the database system\n",
    "length": 135,
    "name": "error",
    "severity": "ERROR",
    "code": "2BP01",
    "file": "dependency.c",
    "line": "536",
    "routine": "findDependentObjects",
    "message": "cannot drop type bit because it is required by the database system",
    "formattedError": "ERROR:  2BP01: cannot drop type bit because it is required by the database system\n"
}
```

Screenshot 2023-10-25 at 7 04 09 AM

Screenshot 2023-10-25 at 7 08 59 AM

EskinderAbera:

Bug fix.

https://github.com/supabase/supabase/issues/18468

## What is the new behavior?

anyone can delete enum types they created in schema.
Screenshot (195)
Screenshot (196)

closes #18468


alaister (approved):
Thanks, @EskinderAbera!

PR #18541, opened 2023-10-27
docs: update auth guides 

kangmingtay:
* Clean up email, magiclink, social login guides
* Update workos guide and display it in the list of social providers supported
* Update JS spec 
* Link auth rate limits page to docs

### DIFF
--- a/apps/docs/pages/guides/auth/auth-magic-link.mdx
+++ b/apps/docs/pages/guides/auth/auth-magic-link.mdx
@@ -19,7 +19,7 @@ Set up Magic Link logins for your Supabase application.
 
 <Admonition type="caution">
 
-When making use of Magic Links with [the PKCE Flow](/blog/supabase-auth-sso-pkce#introducing-pkce) do bear in mind that links can only be used in the same browser that they are sent from. Consequently, a magic link sent from Chrome on Desktop will be invalid if used on a Mobile Device.
+When using Magic Links in [the PKCE Flow](/blog/supabase-auth-sso-pkce#introducing-pkce), keep in mind that links can only be opened from the same browser that they are sent from (as defined in the PKCE spec). Consequently, a magic link sent from Chrome on Desktop will be invalid if used on a Mobile Device.
 
 </Admonition>
 
--- a/apps/docs/pages/guides/auth/auth-magic-link.mdx
+++ b/apps/docs/pages/guides/auth/auth-magic-link.mdx
@@ -3,28 +3,23 @@ import Layout from '~/layouts/DefaultGuideLayout'
 export const meta = {
   id: 'auth-magic-link',
   title: 'Login With Magic Link',
-  description: 'Use Supabase to authenticate and authorize your users using magic links.',
+  description: 'Use Supabase to authenticate your users using magic links.',
 }
 
-Magic links are a form of passwordless logins where users click on a link sent to their email address to log in to their accounts.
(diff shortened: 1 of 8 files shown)


apps/docs/pages/guides/auth/auth-magic-link.mdx
```diff
@@ -3,28 +3,23 @@ import Layout from '~/layouts/DefaultGuideLayout'
...
-1. For [Site URL](https://supabase.com/dashboard/project/_/auth/url-configuration), enter the final (hosted) URL of your app.
-1. For [Auth Providers](https://supabase.com/dashboard/project/_/auth/providers), **enable email provider**.
-
-## Add login code to your client app
+## Sign in with magic link
 
 <Admonition type="caution">
 
-When making use of Magic Links with [the PKCE Flow](https://supabase.com/blog/supabase-auth-sso-pkce#introducing-pkce) do bear in mind that links can only be used in the same browser that they are sent from. Consequently, a magic link sent from Chrome on Desktop will be invalid if used on a Mobile Device.
+When making use of Magic Links with [the PKCE Flow](/blog/supabase-auth-sso-pkce#introducing-pkce) do bear in mind that links can only be used in the same browser that they are sent from. Consequently, a magic link sent from Chrome on Desktop will be invalid if used on a Mobile Device.
```
gregnr, line 22: Would it be possible to explain the reason behind this? Something like:

```suggestion
When using Magic Links in [the PKCE Flow](/blog/supabase-auth-sso-pkce#introducing-pkce), keep in mind that links can only be opened from the same browser that they are sent from (as defined in the PKCE spec). Consequently, a magic link sent from Chrome on Desktop will be invalid if used on a Mobile Device.
```

 I can see a lot of users getting frustrated as to why we would do this when (I assume) its part of the PKCE spec.

apps/docs/pages/guides/auth/social-login.mdx
```diff
@@ -47,9 +46,13 @@ Supabase supports a suite of social providers. Follow these guides to configure
   
 ## Provider Tokens
   
-Once the OAuth flow completes, Supabase Auth will sign your user in. You will receive a copy of the provider token used in the OAuth flow in case you need to use it further. For example, you can use the Google provider token to access Google APIs on behalf of your user.
+You can use the provider token and provider refresh token returned to make API calls to the OAuth provider. For example, you can use the Google provider token to access Google APIs on behalf of your user. 
+
+Supabase Auth does not manage refreshing the provider token for the user. Your application will need to use the provider refresh token to obtain a new provider token. If no provider refresh token is returned, then it could mean one of the following:
+* The OAuth provider does not return a refresh token
+* Additional scopes need to be specified in order for the OAuth provider to return a refresh token.
```
gregnr, line 53: For consistency:

```suggestion
Supabase Auth does not manage refreshing the provider token for the user. Your application will need to use the provider refresh token to obtain a new provider token. If no provider refresh token is returned, then it could mean one of the following:
- The OAuth provider does not return a refresh token
- Additional scopes need to be specified in order for the OAuth provider to return a refresh token.
```

gregnr (approved):

apps/docs/pages/guides/auth/auth-magic-link.mdx
kangmingtay: yeah we should have a separate page to explain what PKCE is instead of linking the blogpost, will add it in a FLUP PR

PR #18586, opened 2023-10-30
feat: fly terms of service screen

phamhieu:

https://supabase.slack.com/archives/C04KMCPT3MG/p1698638295979429?thread_ts=1697514659.955059&cid=C04KMCPT3MG

Screenshot 2023-10-30 at 5 17 24 PM


studio/pages/sign-in-fly-tos.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    width={120}
+                  />
+                </a>
+              </Link>
+            </div>
+          </div>
+        </nav>
+      </div>
+      <div className="flex w-[320px] flex-col items-center justify-center space-y-3">
+        <h4 className="text-lg">Fly Single Sign On</h4>
```
inian, line 44: Not relevant to our users that this is SSO
```suggestion
        <h4 className="text-lg">Continue to Supabase Dashboard</h4>
```

studio/pages/sign-in-fly-tos.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </nav>
+      </div>
+      <div className="flex w-[320px] flex-col items-center justify-center space-y-3">
+        <h4 className="text-lg">Continue to Supabase Dashboard</h4>
+      </div>
+      <div className="flex items-center space-x-4">
+        {/* require extension_id */}
+        <Button onClick={onSignInWithFly} disabled={!extension_id && !organization_id}>
+          Login with Fly.io
+        </Button>
```
joshenlim, line 52: can we add some UI indication if neither extension id nor organization id are missing from the query params in the URL? maybe something like a text below the button at least

<p className="text-red-900 text-sm">Unable to login as neither extension ID nor organization ID are provided</p>

studio/pages/sign-in-fly-tos.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const SignInFlyTos = () => {
+  const router = useRouter()
+  const {
+    query: { extension_id, organization_id },
+  } = router
+  const { resolvedTheme } = useTheme()
+
+  const onSignInWithFly = async () => {
+    window.location.href = extension_id
```
joshenlim, line 16: must it be a window.location.href? will a `<Link />` work?

phamhieu: as we need to redirect to a external url (management api) to trigger sso flow. I'm not sure how to do it with <Link />

joshenlim: `<Link />` also works with ext url

so something like `<Link href={`${API_URL}/tos/fly?extension_id=${extension_id}`}>...</Link>`
similar to how we link to docs site in the dashboard - but just need to double check though that it works. I'm okay with window.location.href otherwise if Link fit the use case here

joshenlim (approved):

studio/pages/sign-in-fly-tos.tsx
phamhieu: I will stick with Button for now cos it's easier to manage enable/disable with button.

PR #18617, opened 2023-10-31
feat(docs): Add guide for regional edge function invocation

laktek:

Adds a new guide for regional edge function invocation.

### DIFF
--- a/apps/docs/pages/guides/functions/regional-invocation.mdx
+++ b/apps/docs/pages/guides/functions/regional-invocation.mdx
@@ -19,7 +19,7 @@ If you prefer to do this, you can specify the region when invoking the function.
 
 ### How to set the region via HTTP headers
 
-You can pass the region via `x-region` HTTP header, when calling a function.
+Use the `x-region` HTTP header when calling an Edge Function to determine where the Function should be executed.
 
 ```bash
 curl --request POST 'https://<project_ref>.supabase.co/functions/v1/hello-world' \
--- a/apps/docs/pages/guides/functions/regional-invocation.mdx
+++ b/apps/docs/pages/guides/functions/regional-invocation.mdx
@@ -8,7 +8,13 @@ export const meta = {
 
 By default, Edge Functions are executed in the region closest to the user making the request. In most cases, this helps to reduce network latency and provide faster responses to the user.
 
-However, if your function does lot of DB operations (eg: adding a bunch of records) or storage operations (eg: uploading files), invoking the function in the same region as your DB may provide better performance.
+However, if your function does lot of database or storage operations, invoking the function in the same region as your DB may provide better performance.
+
+Some situations where this might be helpful include:
+
+- Bulk adding and editing records in your database
+- Uploading files
+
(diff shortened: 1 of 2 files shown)


mansueli (approved):

apps/docs/pages/guides/functions/regional-invocation.mdx
```diff
@@ -0,0 +1,68 @@
+import Layout from '~/layouts/DefaultGuideLayout'
+
+export const meta = {
+  id: 'function-regional-invocation',
+  title: 'Regional Invocation',
+  description: 'How to execute an Edge Functions in a particular region.',
+}
+
+By default, Edge Functions are executed in the region closest to the user making the request. In most cases, this helps to reduce network latency and provide faster responses to the user.
```
kiwicopple, line 9:
```suggestion
By default, Edge Functions are executed in the region closest to the user making the request. This helps to reduce network latency and provide faster responses to the user.
```

The "helps" implies that it's circumstantial so we can remove the qualifier

apps/docs/pages/guides/functions/regional-invocation.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+However, if your function does lot of DB operations (eg: adding a bunch of records) or storage operations (eg: uploading files), invoking the function in the same region as your DB may provide better performance.
```
kiwicopple, line 11:
```suggestion
However, if your function does lot of database or storage operations, invoking the function in the same region as your DB may provide better performance.

Some situations where this might be helpful include:

- Bulk adding and editing records in your database
- Uploading files
```

I'm writing for clippy here. We should answer questions like "when would I want to use regional deployments"

apps/docs/pages/guides/functions/regional-invocation.mdx
```diff
@@ -0,0 +1,68 @@
...
+If you prefer to do this, you can specify the region when invoking the function.
+
+### How to set the region via HTTP headers
+
+You can pass the region via `x-region` HTTP header, when calling a function.
```
kiwicopple, line 22:
```suggestion
Use the `x-region` HTTP header when calling an Edge Function to determine where the Function should be executed.
```

apps/docs/pages/guides/functions/regional-invocation.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+```bash
+curl --request POST 'https://<project_ref>.supabase.co/functions/v1/hello-world' \
+  --header 'Authorization: Bearer ANON_KEY' \
+  --header 'Content-Type: application/json' \
+  --header 'x-region: eu-west-3' \
+  --data '{ "name":"Functions" }'
+```
+
+or using one of the [client libraries](/docs#reference-documentation), e.g. using [supabase-js](/docs/reference/javascript/functions-invoke):
```
kiwicopple, line 32:
```suggestion
You can also specify the region using the [client libraries](/docs#reference-documentation), such as [supabase-js](/docs/reference/javascript/functions-invoke):
```

gregnr (approved):

PR #18619, opened 2023-10-31
fix: missed exclamation mark

m1nsuppp:

An exclamation mark was added because it was missing in the markdown displaying the image.

current

## What is the new behavior?

new behavior

## Additional context


fsansalvadore (approved):
👍👍👍

PR #18623, opened 2023-10-31
Update the functions documentation for python lib

silentworks:

docs update

## What is the new behavior?

## Additional context

### DIFF
--- a/spec/supabase_py_v2.yml
+++ b/spec/supabase_py_v2.yml
@@ -1795,16 +1795,54 @@ functions:
         name: Basic invocation
         description:
         code: |
+          ```python
+          resp = supabase.functions.invoke(
+            "hello-world",
+            invoke_options={
+              "body": { "foo": "bar" }
+            },
+          )
           ```
-          func = supabase.functions()
-          @asyncio.coroutine
-          async def test_func(loop):
-            resp = await func.invoke("hello-world",invoke_options={'body':{}})
-            return resp
+      - id: error-handling
+        name: Error handling
+        description: |
+          A `FunctionsHttpError` error is returned if your function throws an error, `FunctionsRelayError` if the Supabase Relay has an error processing your function.
+        isSpotlight: true
+        code: |
+          ```python
+          from supafunc.errors import FunctionsRelayError, FunctionsHttpError
 
-          loop = asyncio.get_event_loop()
-          resp = loop.run_until_complete(test_func(loop))
-          loop.close()
+          try:
+            resp = supabase.functions.invoke(
+                "hello-world",
+                invoke_options={
+                    "body": {"foo": "bar"},
+                    "headers": {"my-custom-header": "my-custom-header-value"},
+                },
+            )
+          except FunctionsHttpError as exception:
+            err = exception.to_dict()
+            print(f"Function returned an error {err.get("message")}")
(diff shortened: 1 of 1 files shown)


spec/supabase_py_v2.yml
```diff
@@ -1795,16 +1795,54 @@ functions:
...
-            resp = await func.invoke("hello-world",invoke_options={'body':{}})
-            return resp
-
-          loop = asyncio.get_event_loop()
-          resp = loop.run_until_complete(test_func(loop))
-          loop.close()
+      - id: error-handling
+        name: Error handling
+        description: |
+          A `FunctionsHttpError` error is returned if your function throws an error, `FunctionsRelayError` if the Supabase Relay has an error processing your function.
```
kiwicopple, line 1809:
```suggestion
          Returns one of the following errors:
          
          - `FunctionsHttpError`: if your function throws an error
          - `FunctionsRelayError`: if the Supabase Relay encounters an error processing your function
```

spec/supabase_py_v2.yml
```diff
@@ -1795,16 +1795,54 @@ functions:
...
+        isSpotlight: true
+        code: |
+          ```python
+          from supafunc.errors import FunctionsRelayError, FunctionsHttpError
```
kiwicopple, line 1813: is `supafunc` something that we can change? Ideally we don't use things like `supaXX`. better if it can be `supabase-functions` - no abbreviations

spec/supabase_py_v2.yml
```diff
@@ -1795,16 +1795,54 @@ functions:
...
+            err = exception.to_dict()
+            print(f"Function returned an error {err.get("message")}")
+          except FunctionsRelayError as exception:
+            err = exception.to_dict()
+            print(f"Relay error: {err.get("message")}")
+          ```
+      - id: passing-custom-headers
+        name: Passing custom headers
+        description: |
+          You can pass custom headers to your function. Note: supabase-py automatically passes the `Authorization` header with the signed in user's JWT.
```
kiwicopple, line 1833:
```suggestion
          The library accepts custom headers via the `headers` option. 
          
          Note: `supabase-py` automatically populates the `Authorization` header if there is a signed in user.
```

kiwicopple (approved):

spec/supabase_py_v2.yml
silentworks: This is because the python package name is just that, I think `functions` was taken or even `functions-py` was taken too at the time of the packages being created. We could let the `supabase-py` package export these errors too but it would mean we flood that namespace with errors from every package.

```py
from supabase.errors import FunctionsRelayError, FunctionsHttpError
```

silentworks: @J0 how hard would it be to get the package renamed to `supabase-functions` on PyPi?

J0: > I think functions was taken or even functions-py was taken too at the time when the package was created. 
Yup, this is accurate - a poll was done on Slack and `supafunc` had the highest votes.

There's no concept of rename I think. We'll have to publish to the new namespace from here and deprecate the old namespace slowly.I've [created a new package](https://pypi.org/manage/project/supabase-functions/collaboration/) and sent out invites to both.

PR #18628, opened 2023-10-31
fix: Fix a render error in the foreign key selector

ivasilov:
The `SelectorGrid` wasn't properly migrated when we switched to using `react-select-grid`. The error was caused by different type definitions between the old and the new data-grid libs. 

To replicate:
1. Add a table with `json` column.
2. Add a row with some json content in the column (for example, `[{ "example": 2}]`)
3. Add another table and have the first table as a foreign reference
4. Try to add a reference by clicking the empty cell in the Table editor. Foreign Key selector should open and crash.

The bug was reported via support ticket.


alaister (approved):
LGTM!

joshenlim (approved):

PR #18640, opened 2023-11-01
chore: prefetch project

alaister:
Unknowns:

- Is each button knowing what it has to preload the best approach? Can we invert control here?
- Can the API handle the extra strain of so many requests?
- Is adding a default `staleTime` of 1 minute to all queries okay? Are we invalidating everything everywhere?


alaister:
Closing this for now – will revisit down the road!

PR #18641, opened 2023-11-01
feat: Add automatic login flow support

kamilogorek:
Adds support for CLI automated login flow, which uses the browser session to authenticate and generate a new token, which is used in the cli `login` command.

RFC: https://www.notion.so/supabase/RFC-Automatic-browser-based-CLI-login-flow-fc4f91fc7e7f4f168d2da6192841dcec?d=d8c9bcbe826c4665a89f9240d0854a36

CLI changes: https://github.com/supabase/cli/pull/1630
API changes: https://github.com/supabase/infrastructure/pull/15529

### DIFF
--- a/studio/data/cli/login.ts
+++ b/studio/data/cli/login.ts
@@ -1,0 +1,21 @@
+import { post } from 'data/fetchers'
+
+export async function createCliLoginSession(
+  sessionId: string,
+  publicKey: string,
+  tokenName?: string
+) {
+  if (!sessionId) {
+    throw new Error('sessionId is required')
+  }
+
+  const { data } = await post(`/platform/cli/login`, {
+    body: {
+      session_id: sessionId,
+      public_key: publicKey,
+      token_name: tokenName,
+    },
+  })
+
+  return data
+}
--- a/studio/data/cli/login.ts
+++ b/studio/data/cli/login.ts
@@ -9,13 +9,15 @@ export async function createCliLoginSession(
     throw new Error('sessionId is required')
   }
 
-  const { data } = await post(`/platform/cli/login`, {
+  const { data, error } = await post(`/platform/cli/login`, {
     body: {
       session_id: sessionId,
       public_key: publicKey,
       token_name: tokenName,
     },
   })
+
+  if (error) throw error
 
   return data
 }
--- a/studio/pages/cli/login.tsx
+++ b/studio/pages/cli/login.tsx
@@ -1,0 +1,54 @@
+import { NextPage } from 'next'
+import APIAuthorizationLayout from 'components/layouts/APIAuthorizationLayout'
+import { createCliLoginSession } from 'data/cli/login'
+import { useParams } from 'common'
+import { useRouter } from 'next/router'
+import { withAuth } from 'hooks'
+import Connecting from 'components/ui/Loading/Loading'
+import { useEffect, useState } from 'react'
+
+const CliLoginPage: NextPage = () => {
+  const router = useRouter()
+  const { session_id, public_key, token_name, success } = useParams()
+  const [isSuccessfulLogin, setSuccessfulLogin] = useState(false)
+
+  useEffect(() => {
+    if (!router.isReady) {
+      return
+    }
+
+    if (success) {
+      setSuccessfulLogin(true)
+      return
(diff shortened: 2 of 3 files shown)


studio/pages/cli/login.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+      if (session) {
+        router.push(`/cli/login?success=true`)
+      } else {
+        router.push(`/404`)
+      }
+    }
+
+    createSession()
+  }, [router, router.isReady, session_id, public_key, token_name, success])
```
kamilogorek, line 41: image

studio/data/cli/login.ts
```diff
@@ -0,0 +1,21 @@
...
+export async function createCliLoginSession(
+  sessionId: string,
+  publicKey: string,
+  tokenName?: string
+) {
+  if (!sessionId) {
+    throw new Error('sessionId is required')
+  }
+
+  const { data } = await post(`/platform/cli/login`, {
```
joshenlim, line 12: needs an error handler?

const { data, error } = await post(...)

if (error) throw error

studio/pages/cli/login.tsx
```diff
@@ -0,0 +1,58 @@
...
+      return
+    }
+
+    async function createSession() {
+      if (!session_id || !public_key) {
+        router.push('/404')
+        return
+      }
+
+      const session = await createCliLoginSession(session_id, public_key, token_name)
```
joshenlim, line 31: should wrap this in a try catch and handle accordingly? since createCliLoginSession could throw an error

catch could just show a toast.error(...)

studio/pages/cli/login.tsx
```diff
@@ -0,0 +1,58 @@
...
+        router.push(`/404`)
+      }
+    }
+
+    createSession()
+  }, [router, router.isReady, session_id, public_key, token_name, success])
+
+  return (
+    <APIAuthorizationLayout>
+      <div>
```
joshenlim, line 45: do we need this extra div?

studio/data/cli/login.ts
joshenlim: could we make this an RQ mutation actually? that way we get some helpful handlers out of the box like isSuccess, isError etc etc

can refer to branch-create-mutation for example

studio/pages/cli/login.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { useParams } from 'common'
+import { useRouter } from 'next/router'
+import { withAuth } from 'hooks'
+import Connecting from 'components/ui/Loading/Loading'
+import { useEffect, useState } from 'react'
+
+const CliLoginPage: NextPage = () => {
+  const router = useRouter()
+  const { session_id, public_key, token_name, success } = useParams()
+  const [isSuccessfulLogin, setSuccessfulLogin] = useState(false)
```
joshenlim, line 13: we won't need this then once createCliLoginSession becomes an RQ mutation

studio/pages/cli/login.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    createSession()
+  }, [router, router.isReady, session_id, public_key, token_name, success])
+
+  return (
+    <APIAuthorizationLayout>
+      <div>
+        <div className={`flex items-center justify-center h-full`}>
+          {isSuccessfulLogin ? (
```
joshenlim, line 47: once we change this to RQ mutation, be sure to handle the 3 states

```
{isLoading && <... />}
{isError && <... />}
{isSuccess && <... />}
```

studio/data/cli/login.ts
kamilogorek: I was thinking about this, but because this is just a one-off call, that will instantly redirect us on either success or failure, it felt like an overkill.

studio/pages/cli/login.tsx
kamilogorek: See my comments above. State is fully controlled by the current url, so mutations here just feels redundant imo

joshenlim: ahh okay i see what you mean 👍

joshenlim (approved):

PR #18648, opened 2023-11-01
output html in error

saltcod:
dangerously output html from schema dropdown error

screenshot-2023-11-01-at-15 42 40

### DIFF
--- a/studio/components/ui/SchemaSelector.tsx
+++ b/studio/components/ui/SchemaSelector.tsx
@@ -79,7 +79,12 @@ const SchemaSelector = ({
             Failed to load schemas
           </AlertTitle_Shadcn_>
           <AlertDescription_Shadcn_ className="text-xs mb-2">
-            Error: {schemasError?.message}
+            <div
+              className="text-wrap"
+              dangerouslySetInnerHTML={{
+                __html: schemasError?.message,
+              }}
+            />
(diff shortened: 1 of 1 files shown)


studio/components/ui/SchemaSelector.tsx
```diff
@@ -79,7 +79,12 @@ const SchemaSelector = ({
             Failed to load schemas
           </AlertTitle_Shadcn_>
           <AlertDescription_Shadcn_ className="text-xs mb-2">
-            Error: {schemasError?.message}
+            <div
+              className="text-wrap"
+              dangerouslySetInnerHTML={{
+                __html: schemasError?.message,
```
joshenlim, line 85: @saltcod this one is tricky as not all error messages are html
based on observation from whenever this happens its usually on 503s from the API that only HTML is returned
other wise its usually a normal text message

joshenlim: thinking about this i'd rather we handle it at the react query level tbh
if error.code === 503 and error.message.includes('...') then return a human readable string

it also enforces a consistent UI for errors, otherwise this will get rendered as an h1

encima: @joshenlim is there an open issue to add this UI handling for all HTML responses?

PR #18659, opened 2023-11-02
chore: cleanup deprecated project usage

kevcodez:
- Replaces usage of deprecated project usage endpoint
- Removes a bunch of unused prefetch calls
- Removes some unused API routes


studio/data/api.d.ts
```diff
@@ -52,6 +52,10 @@ export interface paths {
      */
     get: operations['ProjectsResourceWarningsController_getProjectsResourceWarnings']
   }
+  '/platform/tos/fly': {
```
kevcodez, line 55: Just regenerated this using the npm command

studio/data/analytics/project-log-requests-count-query.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export async function getProjectLogRequestsCountStats(
+  { projectRef }: ProjectLogRequestsCountVariables,
+  signal?: AbortSignal
+) {
+  if (!projectRef) {
+    throw new Error('projectRef is required')
+  }
+
+  const response = await get<ProjectLogRequestsCountResponse>(
```
joshenlim, line 25: let's use the new get method from data/fetchers
they're typed with the API specs

joshenlim: also just to double check if this endpoint is already available on prod?

studio/components/interfaces/Settings/Database/DiskSizeConfiguration.tsx
```diff
@@ -62,7 +60,7 @@ const DiskSizeConfiguration = ({ disabled = false }: DiskSizeConfigurationProps)
     updateProjectUsage({ projectRef, volumeSize })
   }
 
-  const currentDiskSize = projectUsage?.disk_volume_size_gb ?? 0
+  const currentDiskSize = project?.volumeSizeGb ?? 0
```
joshenlim, line 63: is the volumeSizeGb change already updated on prod?

kevcodez: This was already present before actually

studio/data/analytics/project-log-requests-count-query.ts
kevcodez: It's on prod yes, will use the new data fetcher

joshenlim (approved):

PR #18675, opened 2023-11-02
chore: cleanup older toggles

kevcodez:
All these toggles are enabled on prod for all users and have been for a bit.

- resourceExhaustionWarnings
- auditLogs
- mfaSetup
- integrationsV2
- oauthApps
- storageReport
- sharedSnippets
- transferProject (we still have a kill switch)

Idk why the OrganizationLayout is suddenly having issues with conditional rendering


studio/components/layouts/AccountLayout/AccountLayout.tsx
```diff
@@ -91,7 +89,7 @@ const AccountLayout = ({ children, title, breadcrumbs }: PropsWithChildren<Accou
               },
               // show the MFA page only if the feature flag is set or the user has already MFA setup.
               // He should be able to edit/revoke his MFA even if MFA feature flag is disabled.
-              ...(mfaSetup || (factors?.all || []).length > 0
+              ...((factors?.all || []).length > 0
```
ivasilov, line 92: Remove the conditional and the comment.

kevcodez: Don't we still only want to show these if the user has setup MFA?

kevcodez: ah nvm

ivasilov (approved):

PR #18679, opened 2023-11-02
Move TCE info to pg_sodium extension

olirice:
- Removes our dedicated TCE documentation (replaced with a "redirect" for anyone who stored the link)
- Moves the relevant parts of the TCE doc to pgsodium extension docs

Moving forward TCE will be a SQL-only feature rather than one integrated with Studio so re-pointing these docs to the project source better communicates that difference.


steve-chavez (approved):

apps/docs/pages/guides/database/column-encryption.mdx
```diff
@@ -9,279 +9,7 @@ export const meta = {
...
-```
-
-The new label indicates which column is to be associated with the secret, and that's it! Your `account_id` and secret are now protected under the same authentication signature as the secret itself.
-
-## Resources
-
-- [Supabase Vault](/docs/guides/database/vault)
-- Read more about Supabase Vault in the [blog post](https://supabase.com/blog/vault-now-in-beta)
-- [Supabase Vault on GitHub](https://github.com/supabase/vault)
+These docs have moved. See [pg_sodium](/docs/guides/database/extensions/pgsodium)
```
kevcodez, line 12: Is there a reason not to use an automated redirect through next config?

olirice: didn't know how. fixed

alaister (approved):

PR #18683, opened 2023-11-02
chore: use node parseArgs instead of minimist

kevcodez:
Node 18 comes with a built-in args parser - no need for another dependency for these simple use-cases


apps/docs/generator/index.ts
```diff
@@ -3,28 +3,42 @@ import CliGenerator from './cli'
 import ConfigGenerator from './config'
 import SdkGenerator from './sdk'
 import LegacyGenerator from './legacy'
-import * as minimist from 'minimist'
+import { parseArgs } from 'node:util'
+const assert = require('node:assert')
```
gregnr, line 7: Any reason this needed to be `require` over `import`?

spec/parser/tsdoc.ts
```diff
@@ -1,28 +1,25 @@
 import { writeToDisk } from './helpers'
 import stringify from 'json-stringify-safe'
+import { parseArgs } from 'node:util'
+import fs from 'fs'
+const assert = require('node:assert')
```
gregnr, line 5: Here as well

gregnr (approved):
Supa nice 👍

apps/docs/generator/index.ts
kevcodez: There were some weird type errors + execution issues, but I now found a working way with es6 imports

PR #18687, opened 2023-11-03
Remove TCE support for table editor

joshenlim:
Transparent column encryption can still be done through the SQL, we're just removing it from the table editor as it's been causing quite a lot of issues due to its ease of use through the GUI


alaister (approved):
LGTM!

PR #18692, opened 2023-11-03
updated createClient() function for latest nextjs app otherwise it won't work with it

saleem-sadiq:

Documentation Update

The current documentation for the `createClient` function does not mention the required parameter, causing confusion for users.

## What is the new behavior?

The new documentation includes information about the required parameter for the `createClient` function. Specifically, it now states that `createClient` should be called with `createClient(cookies())`, and provides guidance on importing the `cookies` function from 'next/headers' for proper usage.

## Additional context

This PR addresses the documentation issue and enhances the clarity of the `createClient` function. Users will now have the correct information for its usage, which should help prevent misunderstandings and errors in the future.


dijonmusters (approved):
Great catch! Thanks @saleem-sadiq! 🙌

Issue #18310, opened 2023-10-18
Edit page link on wrappers page leads to 404


kamilogorek:
On https://supabase.com/docs/guides/database/extensions/wrappers/airtable and any other wrapper page "Edit this page on GitHub" leads to 404, as those pages are pulled from a separate wrappers docs page.

Ref: https://github.com/supabase/supabase/blob/master/apps/docs/pages/guides/database/extensions/wrappers/%5Bslug%5D.tsx

yousseefben:
Hello, I tried to fix it in a better way, feel free to review it, thanks :)

charislam:
Closing as fixed, thanks @yousseefben!

PR #18700, opened 2023-11-03
🐞 fix: edit page link on wrapper broken 404

yousseefben:

Bug Fix.

Edit page link on wrappers page leads to 404

## What is the new behavior?

Actually the edit GitHub page link works correctly.

## Additional context

[issue: 18310](https://github.com/supabase/supabase/issues/18310)


apps/docs/layouts/guides/index.tsx
```diff
@@ -23,13 +23,14 @@ interface Props {
...
   }
+  editLink?: string
   children: any
   toc?: any
   currentPage?: string
   hideToc?: boolean
 }
 
-const Layout: FC<Props> = (props) => {
+const Layout: FC<Props> = (props, test) => {
```
charislam, line 33:
```suggestion
const Layout: FC<Props> = (props) => {
```

looks like this was kept by mistake?

charislam (approved):
Looks great! Thanks so much for this ❤️ 

Mind taking a quick look at the conflicts and then we should be good?

charislam (approved):

PR #18701, opened 2023-11-03
Fix google nonce check

joshenlim:
There were 2 issues here i think
- Projects would have `EXTERNAL_GOOGLE_SKIP_NONCE_CHECK` as NULL to begin with
- The Toggle in the provider form, when initially toggled to true, would somehow set the value of `EXTERNAL_GOOGLE_SKIP_NONCE_CHECK` as `[true]` instead of `true`
- Also users would then be unable to toggle it off as the UI component doesnt know how to read `[true]`

Fix here is to update `generateInitialValues`
- Properties that are boolean should be initialized to be `false` if `null`, and not empty strings, this might have been why
- Also just adding the required check in the form validation, i think the above is the main fix but this is just a why not


hf (approved):
Interesting!

ivasilov (approved):

PR #18716, opened 2023-11-04
Update sveltekit ssr guide

silentworks:

Docs update

Cookie chunking doesn't work in the browser client since it runs on both client and server.

## What is the new behavior?

Cookie chunking now works on browser client when you import the chunking function and perform the operation inside of the `cookies.get` method.

## Additional context


dijonmusters (approved):
LGTM 🙌

PR #18717, opened 2023-11-04
fix: Delay between Hover

### LINKED ISSUE #18709: Database name tooltip
Want to add the database name in a <Tooltip> here — it's using the name property now, which is good, but there's a delay between hover and this appearing. Tooltip doesn't have this delay. 

screenshot-2023-11-03-at-15 46 15

Should probably work like the table editor — only long table names show the tooltip, not the short ones.
screenshot-2023-11-03-at-15 48 00

samueldenzil:

Fixes: #18709 

After:
Screenshot from 2023-11-04 09-53-24


alaister (approved):
Thanks, @samueldenzil!

PR #18727, opened 2023-11-04
Next.js Secrets Management

yaralahruthik:
Next.js by default allows .env to be checked into version control. It would be better for the instructions to include the .env.local instructions.

Docs Update

## Additional context

Next.js by default allows .env to be checked into version control. It would be better for the instructions to include the .env.local instructions. I had to hack around to come at this solution.


yaralahruthik:
There seems to be more to this! I'll figure out the issues!

PR #18735, opened 2023-11-05
Clarify SUPABASE_PROJECT_ID env is not auto-detected be CLI

martypdx:

Doc update with clarification about CLI and env var `SUPABASE_PROJECT_ID`

While the code example shows the correct use of `--project-ref`, the explanatory text does not mention this difference between token and db password versus the project id. This may lead to time lost when improving the infrastructure and ci build and trying to consolidate this env var.

## What is the new behavior?

Added a sentence that explains exactly what to expect

## Additional context

[Issue on cli repo](https://github.com/supabase/cli/issues/1627) when I thought this was a bug


sweatybridge:
CLI beta release has added support for SUPABASE_PROJECT_ID env var so I think we can close this.

PR #18745, opened 2023-11-06
fixed: hover delay on database name tooltip

AbhijeetBiswal:
fixed issue - #18709


alaister:
Hi @AbhijeetBiswal,

Thanks for your contribution. However, it looks like this has already been solved in #18717, so I'm going to go ahead and close this.

In future, you might like to check the issue to see if there are already open PRs to avoid doing unnecessary work.

Thanks again 🙌🏻

PR #18749, opened 2023-11-06
Chore/make api docs permanent

joshenlim:
Make the new API side panel documentation that was behind a toggle-able feature preview as the de-facto, which will replace the existing API docs page of the dashboard.

Also added an empty state for the feature preview modal
image


alaister (approved):
LGTM!

GaryAustin1:
Already have one user looking for the Bash option.  This feature was used quite a bit by users doing direct REST calls and as a reference by me when they were having issues formatting their URLs. It was a simple way to get them going versus point them to PostgREST.org to figure out the url format.  It also included the full URL to the API and headers which PostgREST.org does not.  

Does this still exist somewhere ?

PR #18793, opened 2023-11-07
fix: Move dat.gui dependency to the common package

ivasilov:
Remove dat.gui from docs and studio since it's not used directly. Add dat.gui as a dependency to `common` package.


fsansalvadore (approved):
Good call 👍

saltcod (approved):

PR #18838, opened 2023-11-08
Update llamaindex.ipynb doc_hash to hash

thinkjrs:
Update `doc_hash` to `hash` since running this as-is provided an `AttributeError`. 

Here's what the Python built-in `vars(documents[0])` gives me (truncated):

```python
{'id_': 'd94b1287-b3ef-4aba-ac6b-7a4e353b0327',
 'embedding': None,
 'metadata': {},
 'excluded_embed_metadata_keys': [],
 'excluded_llm_metadata_keys': [],
 'relationships': {},
 'hash': '4c702b4df575421e1d1af4b1fd50511b226e0c9863dbfffeccb8b689b8448f35',...}
```

Docs example update/bugfix.

Running the Llamaindex example spits out an `AttributeError` in cell 4 for the non-extant `doc_hash` attribute, which has been changed to `hash`, as you can see in the above output from `vars` which I ran on the `Document` object returned in the `documents` array within the notebook.

## What is the new behavior?

It is the expected output as before: the string hash is printed.

## Additional context

This merge is not consequential to the library's functionality in any way but allows the example notebook to execute without error.


gregnr (approved):
Thanks @thinkjrs!

Just tried this out again and noticed the `SimpleWebPageReader` import has also changed on the `llama_index` side - updated to fix that as well.

PR #18844, opened 2023-11-09
Ensure that database triggers, functions and indexes page and coherent with the protected schema UX

joshenlim:
The dashboard always had a guard rail to prevent users from managing things via the dashboard GUI in protected schemas, by showing this warning where appropriate
image

For more context on what protected schemas are:
image

But we realised that triggers + functions were still showing the "create" button while viewing a protected schema, despite not being able to create triggers/functions in protected schemas - this is no change in behaviour, just making the UX consistent.

However, indexes were missing this protected schemas check, and as such this PR adds the protected schemas check to the indexes page. _**This means that the following behaviour will change on the dashboard**_
- Users will no longer be able to create/update/delete indexes through the database/indexes page, while in a protected schema.
- The alternative is do so the relevant operations via the SQL editor instead (as per the same as all other database/xxx pages)

### DIFF
--- a/studio/components/interfaces/Database/Functions/FunctionsList/FunctionsList.tsx
+++ b/studio/components/interfaces/Database/Functions/FunctionsList/FunctionsList.tsx
@@ -102,32 +102,34 @@ const FunctionsList = ({
               />
             </div>
 
-            <Tooltip.Root delayDuration={0}>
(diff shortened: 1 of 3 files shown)


alaister (approved):
LGTM!

PR #18859, opened 2023-11-09
wrap tab lists to prevent overflow

charislam:

**Bug fix**

Quick fix to add a wrapping option that makes tabs visible if there are many of them.

Long-term fix would be to detect overflow and switch to a dropdown in that case. Or use dropdown for smaller screens only and let it wrap on larger screens (since it will be rarer for wrapping to trigger there).

Tabs are cut off on the right when there are many of them. There is no way to scroll right on desktop because scrollbar width is set to 0. Also not super obvious on mobile that there is room to scroll.

image

## What is the new behavior?

image

## Additional context

### DIFF
--- a/packages/ui/src/components/Tabs/Tabs.tsx
+++ b/packages/ui/src/components/Tabs/Tabs.tsx
@@ -16,6 +16,7 @@ interface TabsProps {
   onChange?: any
   onClick?: any
   scrollable?: boolean
+  wrappable?: boolean
   addOnBefore?: React.ReactNode
   addOnAfter?: React.ReactNode
   listClassNames?: string
@@ -35,6 +36,7 @@ const Tabs: React.FC<PropsWithChildren<TabsProps>> & TabsSubComponents = ({
   onChange,
   onClick,
   scrollable,
+  wrappable,
   addOnBefore,
   addOnAfter,
   listClassNames,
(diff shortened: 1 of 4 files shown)


MildTomato:
Should we just update the component Tabs styling instead so it can support wrap / no wrap ?

charislam:
Yeah sure! @MildTomato is this what you meant?

packages/ui/src/components/Tabs/Tabs.tsx
```diff
@@ -89,7 +91,10 @@ const Tabs: React.FC<PropsWithChildren<TabsProps>> & TabsSubComponents = ({
 
   const listClasses = [__styles[type].list]
   if (scrollable) listClasses.push(__styles.scrollable)
+  if (wrappable) listClasses.push(__styles.wrappable)
   if (listClassNames) listClasses.push(listClassNames)
+  console.log(wrappable)
+  console.log(listClasses)
```
charislam, line 97:
```suggestion
```

saltcod:
wondering if I'm missing something on this one. 

This is the original page with the issue
https://docs-git-fix-tabs-scroll-supabase.vercel.app/docs/guides/auth/server-side/creating-a-client?environment=get-server-side-props&framework=nextjs

Am I looking at the wrong thing? I don't see the wrapping going on

screenshot-2023-11-14-at-21 13 58


screenshot-2023-11-14-at-21 14 35

MildTomato (approved):
Lgtm

charislam:
@saltcod I made the wrappability opt-in only (you need to set `wrappable`) since this is a shared UI component, and I didn't want to break styling assumptions in other apps.

I just tweaked the PR to make `wrappable` default on the docs site only, and now it works as expected:

image

PR #18865, opened 2023-11-10
Chore/revert changes to api docs

joshenlim:
This addresses some issues with the recent change in making our API side panel feature preview replace the original API docs page ([ref](https://github.com/supabase/supabase/pull/18749)). The change led to a couple of confusions such as:
- Inability to find how to change languages for API docs
- New API docs side panel doesn't support URLs
- GraphiQL interface was out of place having been placed under the "Database" section

We've decided that we'll bring back the old API docs pages for now until the above issues are addressed / we find a better solution. The new API docs side panel will be back under the feature preview section which you can still toggle through the left side navigation bar here:
image
image

The API docs button in the Table Editor will still use the new API side panel, just only the "Entities" part of the docs. We've also updated the language selector on this part in hopes to improve the visibility for changing languages for code snippets
image


alaister (approved):
LGTM!

PR #18883, opened 2023-11-11
Fix the footer links in the supabase.com

shishiro26:
## What kind of change does this PR introduce?
#18882 

Bug fix, feature, docs update, ...

## What is the current behavior?
Many of the links in the footer of https://supabase.com/ aren't rendering as links

Please link any relevant issues here.

## What is the new behavior?

In this the links are being rendered as the links rather than the div 's 

Feel free to include screenshots if it includes visual changes.
image

## Additional context

Add any other context or screenshots.

PR #18885, opened 2023-11-11
Footer not rendering as link Fix

navArxus:
It fix the link not rendering as a link 

The link are currenly appear as a text cursor 
#18883 
## What is the new behavior?
The link appear as a cursor pointer 

## Additional context
It  is the fix of #18883 in which the footer link appear as a cursor text now it will appear as a cursor pointer 


alaister (approved):
Thanks, @nav302004!

Please note I've changed the approach used, but the result is still the same.

PR #18905, opened 2023-11-13
docs: rename `client` to `supabase` within Kotlin docs

dshukertjr:

Renames the `client` variable name to `supabase` to match other client libraries' docs.


thorwebdev (approved):
LGTM, thanks

PR #18915, opened 2023-11-13
chore: Move the `studio` app to `apps/studio`

ivasilov:
This PR moves the `studio` app to be in the `apps` folder.


.prettierignore
```diff
@@ -12,10 +12,4 @@ examples/slack-clone/nextjs-slack-clone/full-schema.sql
...
 apps/docs/pages/guides/auth/*.mdx
 apps/docs/pages/guides/integrations/*.mdx
-studio/public
-studio/__mocks__
-studio/.next
-studio/.storybook
-studio/.swc
-studio/.turbo
-studio/node
+apps/studio/public
```
kevcodez, line 15: Why are we keeping the public folder but removing all the other stuff?

kevcodez: As you removed the file inside studio, should we add .swc, .turbo etc again?

ivasilov: You're right, added `apps/**/.turbo` here. The other folders are probably a leftover from the previous git repo.

kevcodez (approved):

alaister (approved):

PR #18974, opened 2023-11-15
Chore/branch management fixes

joshenlim:
- Fix delete branch modal not showing when under Pull requests or All branches tab
- Sort preview branches by last updated at
- Show last updated at instead of created at for each branch
- Add max width to branch name and add title property


studio/components/interfaces/BranchManagement/Overview.tsx
```diff
@@ -1,4 +1,4 @@
-import { useParams } from 'common'
+import { noop } from 'lodash'
```
alaister, line 1:
```suggestion
import { noop } from 'lib/void'
```
reduce deps = make Kevin happy? 🤷

alaister (approved):
LGTM!

PR #18977, opened 2023-11-15
feat: add friendly name to spec

J0:

- Corresponding docs update for: https://github.com/supabase/gotrue/pull/1277


apps/www/pages/contact/mfa.tsx
hf, line 1: File is deleted?

J0: Yup that is intended I think we no longer need the early access page as it's available to all. Will wait for website team to confirm if it's ok to delete though

J0:
Actually, the Auth deploy needs to go out before this can be merged. Reverting to draft

apps/www/pages/contact/mfa.tsx
J0: Decided to do in another PR as it's not relevant to this one

charislam (approved):

PR #18979, opened 2023-11-15
Prevent client from crashing when new params are added to resource warnings endpoint

joshenlim:
As the title suggests ^


alaister (approved):
LGTM!

PR #18980, opened 2023-11-15
Update query-optimization.mdx

ohrrkan:
Looks like you redirect on an unofficial and invalid certificat Postgres website.

By the way looking at the Blame, l will probably suggest to investigate why this was change ?intentionally? 3 months ago...


kevcodez (approved):

kevcodez:
Thanks! This was accidentally replaced when we changed the spelling, must have slipped through.

PR #18991, opened 2023-11-15
Update image-transformations.mdx

kkukshtel:
Add in a note about how storage can now support quality transformations per Storage V3

https://supabase.com/blog/storage-v3-resumable-uploads#more-image-transformations-options

Update docs to reflect image optimization transforms for storage v3

No documentation outside of blog post and digging into the API reference

## What is the new behavior?

Adds docs that show usage of new parameter

## Additional context

No sure if the Kotlin code is right, and happy to have the docs re-written or whatever. Just wanted to call this out because it was really hard to find this existed and I was halfway down the road of implementing Sharp for a function until I accidentally stumbled across this release.

### DIFF
--- a/apps/docs/pages/guides/storage/serving/image-transformations.mdx
+++ b/apps/docs/pages/guides/storage/serving/image-transformations.mdx
@@ -225,7 +225,48 @@ const MyImage = (props) => {
 
 ## Transformation options
 
-We currently support a few transformation options focusing on resizing and cropping images.
+We currently support a few transformation options focusing on optimizing, resizing, and cropping images.
+
+### Optimizing
+
+You can set the quality of the returned image by passing a value from 20 to 100 (with 100 being the highest quality) to the `quality` parameter. This parameter defaults to 80.
+
+Example:
+
+<Tabs
+  scrollable
+  size="small"
+  type="underlined"
+  defaultActiveId="js"
+  queryGroup="language"
+>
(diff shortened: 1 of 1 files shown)


charislam:
Thanks for the PR, @kkukshtel! This is definitely a feature we want people to know about.

Have you seen [this section](https://github.com/supabase/supabase/blob/c2ffb23bb43784e6856777e6cca4bc5d56a796f2/apps/docs/pages/guides/storage/serving/image-transformations.mdx#L108-L113) further up the page? I'm thinking that it might be better to include the optimization here, rather than creating a new subsection for it

kkukshtel:
Yeah I wasn't sure where exactly to put it, mostly because I wasn't sure what to consider as the "basic" transform (and also ambiguity about what would happen if you just passed `{}` as the transform parameter. Similarly, that section doesn't mention the format parameter. IMO the _right_ way to do it would be to rewrite the page to make it more about the transforms specifically (sort of what I was doing) instead of trying to broadly present the idea of a transform and then some parameters. I think right now the page feels like a declarative overview of parameters, so if it's that what it's meant to be it could be leaned more into (or broken out into a separate page?)

charislam (approved):

charislam:
Fair enough! In the interests of moving things along, I'll merge this and make a note to revisit the page as a whole later!

PR #19002, opened 2023-11-16
Add org slug as copyable input to org settings

joshenlim:
Add org slug as copyable input to org settings
image


alaister (approved):
LGTM!

PR #19004, opened 2023-11-16
Add storage to service status

joshenlim:
- Add storage to service status
- Also scaffold errors in `services`, was thinking we should perhaps float the error in a tooltip eventually
image

### DIFF
--- a/apps/studio/components/interfaces/Home/ServiceStatus.tsx
+++ b/apps/studio/components/interfaces/Home/ServiceStatus.tsx
@@ -110,50 +110,54 @@ const ServiceStatus = () => {
   const allServicesOperational = services.every((service) => service.isSuccess)
 
   return (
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Home/ServiceStatus.tsx
```diff
@@ -40,23 +43,61 @@ const ServiceStatus = () => {
...
+          },
+        ]
+      : []),
+    {
+      name: 'Storage',
+      error: storageStatus?.error,
+      docsUrl: undefined,
+      isLoading,
+      isSuccess: storageStatus?.healthy,
+    },
```
alaister, line 95: Should this be conditional by wrapping in `storageEnabled`?

joshenlim: oh yes good point!

alaister (approved):
LGTM!

alaister (approved):

PR #19012, opened 2023-11-16
Chore/refactor organization details query to organization members

joshenlim:
- Refactor organization details query to organization members + use new get method from data/fetchers
- Proactively added support for showing if org member has enabled mfa

This is related to this PR: https://github.com/supabase/infrastructure/pull/15725
The new column "MFA enabled" should show once this PR goes in


alaister (approved):
LGTM!

PR #19018, opened 2023-11-16
Add note on disk downsizing when updating

encima:

Docs update

Currently, this functionality is not documented

## What is the new behavior?

Adding a note to the documentation to explain "right-sizing" behaviour when updating a project.


charislam (approved):

apps/docs/pages/guides/platform/database-size.mdx
```diff
@@ -71,6 +71,12 @@ Due to provider restrictions, disk expansions can occur only once every six hour
 
 </Admonition>
 
+<Admonition type="info">
```
darora, line 74: Looks fine, though perhaps we should also add a note to the caveats section of the upgrades doc?

darora (approved):

PR #19040, opened 2023-11-17
chore: cleanup studio dependencies

kevcodez:
- Get rid of unused dependencies
- Move dev dependencies to dev dependencies


apps/studio/package.json
```diff
@@ -52,8 +49,6 @@
     "file-saver": "^2.0.5",
     "framer-motion": "^6.5.1",
     "generate-password": "^1.7.0",
-    "graphql": "^16.8.1",
-    "graphql-ws": "^5.14.1",
```
alaister, line 56: These might still be needed for GraphiQL

kevcodez: I've successfully triggered some graphl requests, what else should I test?

These might also be transitive dependencies from graphiql

kevcodez: Already a transitive dep, shouldn't be a need to define it separately

alaister: Noice!

alaister (approved):
🔥

(if only the red lines were green 😮‍💨)

PR #19049, opened 2023-11-17
Move setProject out of withAuth so self-hosted still gets project ref #19046

saltcod:
Project ref was set via `setProject` in `withStore`

Since self-hosted setup doesn't use withAuth, this was no longer getting set. 
Not having a project ref ("default") causes lots of issues — creating a table, viewing logs, etc.

This PR moves that `setProject` call to the rootStore — **this is likely not the best place**. But is a temp work around if you're blocked. 

— 

props K-dog for the help.

### DIFF
--- a/apps/studio/hooks/misc/useStore.tsx
+++ b/apps/studio/hooks/misc/useStore.tsx
@@ -26,7 +26,6 @@ export const StoreProvider = ({ children, rootStore }: PropsWithChildren<StorePr
   const selectedProject = useSelectedProject()
   useEffect(() => {
     if (selectedProject) {
-      console.log({ selectedProject })
       rootStore.setProject(selectedProject)
     }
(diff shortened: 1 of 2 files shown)


apps/studio/hooks/misc/useStore.tsx
```diff
@@ -22,6 +23,14 @@ interface StoreProvider {
 export const StoreProvider = ({ children, rootStore }: PropsWithChildren<StoreProvider>) => {
   const { ui } = rootStore
 
+  const selectedProject = useSelectedProject()
+  useEffect(() => {
+    if (selectedProject) {
+      console.log({ selectedProject })
```
alaister, line 29:
```suggestion
```

alaister (approved):
LGTM!

PR #19055, opened 2023-11-17
Chore/add platform checks

saltcod:
Stop resourceWarnings from loading in self-hosted


apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
```diff
@@ -37,7 +37,7 @@ const ProjectList = ({ rewriteHref }: ProjectListProps) => {
     isError: isErrorPermissions,
     error: permissionsError,
   } = usePermissionsQuery()
-  const { data: resourceWarnings } = useResourceWarningsQuery()
+  const { data: resourceWarnings } = useResourceWarningsQuery({ enabled: IS_PLATFORM })
```
joshenlim, line 40: lets put this IS_PLATFORM check inside the react query instead? then its just one place instead of many

kevcodez (approved):
Just the cleanup from the imports

PR #19067, opened 2023-11-18
docs(swift): rename `authStateChanges` and remove `File` from upload method

grdsdev:

Docs update

## What is the new behavior?

- `onAuthStateChange()` method was replaced by a `authStateChanges` property.
- Storage's `upload` and `update` methods accept a `Data` file instead of the `File` type.


thorwebdev (approved):
LGTM, thanks

PR #19075, opened 2023-11-19
docs: Improve docs for joins and TypeScript

thorwebdev:

- Improve docs for joins and TypeScript

### DIFF
--- a/apps/docs/docs/ref/javascript/typescript-support.mdx
+++ b/apps/docs/docs/ref/javascript/typescript-support.mdx
@@ -228,11 +228,9 @@ const { data } = await supabase.from('countries').select().returns<MyType>()
     );
     type CountriesWithCities = DbResultOk<typeof countriesWithCitiesQuery>;
 
-    const {
-      data,
-      error,
-    }: { data: CountriesWithCities | null; error: DbResultErr | null } =
-      await countriesWithCitiesQuery;
+    const { data, error } = await countriesWithCitiesQuery;
+    if (error) throw error;
+    const countriesWithCities: CountriesWithCities = data;
     ```
 
   </RefSubLayout.Examples>
--- a/apps/docs/docs/ref/javascript/typescript-support.mdx
+++ b/apps/docs/docs/ref/javascript/typescript-support.mdx
@@ -9,7 +9,7 @@ With TypeScript, `supabase-js` detects things like `not null` constraints and [g
 
 `supabase-js` also detects relationships between tables. A foreign table with one-to-many relationship is typed as `T[]`. Likewise, a foreign table with many-to-one relationship is typed as `T | null`.
 
-## Generating types
+## Generating TypeScript Types
 
 <RefSubLayout.EducationRow>
   <RefSubLayout.Details>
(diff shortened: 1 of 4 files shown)


kiwicopple (approved):

apps/docs/docs/ref/javascript/typescript-support.mdx
```diff
@@ -181,8 +219,20 @@ You can use the following helper types to make the TypeScript support easier to
...
+      cities ( id, name )
+    `
+    );
+    type CountriesWithCities = DbResultOk<typeof countriesWithCitiesQuery>;
+
+    const {
+      data,
+      error,
+    }: { data: CountriesWithCities | null; error: DbResultErr | null } =
+      await countriesWithCitiesQuery;
```
soedirgo, line 235: We should let this be inferred - `{ data: CountriesWithCities | null; error: DbResultErr | null }` needs more effort to satisfy type checks since it can include invalid states like `{ data: CountriesWithCities; error: DbResultErr }`. E.g. instead of doing:

```ts
if (error) throw error;
// data: DbResultOk
// ...do things with data...
```

You'd need:
```ts
if (error) throw error;
// data: DbResultOk | null
if (!data) throw new Error('...');
// data: DbResultOk
// ...do things with data...
```

apps/docs/pages/guides/api/joins-and-nesting.mdx
```diff
@@ -89,6 +89,51 @@ const { data, error } = await supabase.from('countries').select(`
...
+  id,
+  name,
+  cities ( id, name )
+`
+)
+type CountriesWithCities = DbResultOk<typeof countriesWithCitiesQuery>
+
+const { data, error }: { data: CountriesWithCities | null; error: DbResultErr | null } =
+  await countriesWithCitiesQuery
+```
```
soedirgo, line 135: Ditto

apps/docs/pages/guides/api/rest/generating-types.mdx
```diff
@@ -2,7 +2,7 @@ import Layout from '~/layouts/DefaultGuideLayout'
 
 export const meta = {
   id: 'generating-types',
-  title: 'Generating Types',
+  title: 'Generating TypeScript Types',
```
soedirgo, line 5: Should we just link to the reference page? There's quite a bit of duplicated content here

apps/docs/docs/ref/javascript/typescript-support.mdx
```diff
@@ -147,12 +163,17 @@ You can use the following helper types to make the TypeScript support easier to
   </RefSubLayout.Details>
   <RefSubLayout.Examples>
     ```ts ./database.types.ts
```
MildTomato, line 165: double check this works.
MDX parser is a bit funny when there are no empty lines at start and end of component children

apps/docs/pages/guides/api/rest/generating-types.mdx
soedirgo: We can keep the generating & GHA part here, but link to the reference on actually using the types?

thorwebdev: cc @RamiroND @MildTomato regarding SEO considerations here.

PR #19095, opened 2023-11-20
Show email that support will reply to after submitting

joshenlim:
I also fixed the copy in the screen shot below - missing "reach out to _you_ at"

image


alaister (approved):
LGTM!

PR #19124, opened 2023-11-21
fix: check if preview_branch_refs exists

Hallidayo:

Bug Fix - Local Dev

Running the development version of Supabase if you go to the main projects dashboard this error is appearing:

Screenshot 2023-11-19 at 17 35 35

## What is the new behavior?

Added a check for the variable and now the page loads without error.

## Additional context

@MildTomato


apps/studio/components/interfaces/Home/ProjectList/ProjectCard.tsx
```diff
@@ -25,7 +25,7 @@ const ProjectCard = ({
   const { name, ref: projectRef } = project
   const desc = `${project.cloud_provider} | ${project.region}`
 
-  const isBranchingEnabled = project.preview_branch_refs.length > 0
+  const isBranchingEnabled = project.preview_branch_refs && project.preview_branch_refs.length > 0
```
MildTomato, line 28: @joshenlim i suppose this only makes sense if default project is also counted in the preview_branch_refs array

joshenlim: yeah you're right - the api route for default project probably doesnt have this param hence why it crashed here

joshenlim, line 28: i'd opt to do it this way tbh

```suggestion
  const isBranchingEnabled = (project?.preview_branch_refs ?? []).length > 0
```

Hallidayo: Hi @joshenlim - Shall I create a commit for this?

Hallidayo:
Hi @MildTomato & @joshenlim - I've just updated the change to use @joshenlim line.

saltcod:
@Hallidayo a quick conflict there & we should be g2g!

alaister:
Hey @Hallidayo,

I'm gonna go ahead and close this as I believe it's been fixed in another PR! Thanks!

PR #19126, opened 2023-11-21
Allow access to db backups page while project is restoring, but disable restore button if project is restoring

joshenlim:


alaister (approved):
LGTM!

Issue #18046, opened 2023-10-09
update vector column documentation to leverage indexes


anngbaum:
# Improve documentation

## Link 

The code suggested for querying a vector embedding ([here](https://supabase.com/docs/guides/ai/vector-columns#querying-a-vector--embedding)) includes the following:

```
  select
    documents.id,
    documents.content,
    1 - (documents.embedding <=> query_embedding) as similarity
  from documents
  where 1 - (documents.embedding <=> query_embedding) > match_threshold
  order by similarity desc
  limit match_count;
```
However, in our testing, `order by similarity` as a calculated sort does not seem to use the index of the vectors. We dramatically improved our query time **by an order of magnitude** simply by switching to use:

`order by (documents.embedding <=> query_embedding) asc`

## Describe the problem

The documentation is misleading because this will cause an inefficient query that will not take advantage of the index for the vector column.

## Describe the improvement

Guide users to user `order by (documents.embedding <=> query_embedding) asc`

## Additional context

I'll add some comments below of the testing I performed to confirm this was the case.

anngbaum:
we have a table called `recipe_embeddings` that just has `recipe_id` and `embedding` columns.

with a predetermined `search_vector`, I ran `EXPLAIN ANALYZE` for the following:

```
SELECT * FROM (
WITH search_vector AS (
  SELECT search_query_vectors.search_vector FROM search_query_vectors WHERE search_query_vectors.search_query = 'chicken' LIMIT 1
)
SELECT
  recipe_vectors.recipe_id,
  1 - (recipe_vectors.embedding <=> (SELECT * FROM search_vector)) as similarity
FROM recipe_vectors
ORDER BY similarity DESC
LIMIT 600
) AS found_recipes
```
and got the following results:

```
| QUERY PLAN                                                                                                                                                    |
| ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Limit  (cost=16538.85..16607.85 rows=600 width=24) (actual time=6721.532..6722.006 rows=600 loops=1)                                                          |
|   InitPlan 1 (returns $0)                                                                                                                                     |
|     ->  Limit  (cost=0.28..2.50 rows=1 width=18) (actual time=1.185..1.186 rows=1 loops=1)                                                                    |
|           ->  Index Scan using search_query_vectors_pkey on search_query_vectors  (cost=0.28..2.50 rows=1 width=18) (actual time=1.184..1.184 rows=1 loops=1) |
[...]

PR #19133, opened 2023-11-21
docs: update vector sorting to ensure index gets used

anngbaum:

The current guidance leads to a query that does not leverage the index successfully, resulting in poor performance if implemented exactly as mentioned. Discussed in more detail in #18046 

Example query that does not leverage the index.

## What is the new behavior?

A better example query.

## Additional context

#18046


gregnr (approved):
Amazing, thanks @anngbaum!

PR #19162, opened 2023-11-23
feat/notifications v2

joshenlim:
Example:

image


apps/studio/components/layouts/ProjectLayout/LayoutHeader/NotificationsPopoverV2/NotificationRow.tsx
```diff
@@ -0,0 +1,196 @@
...
+  listRef,
+  item: notification,
+  setRowHeight,
+  getProject,
+  getOrganization,
+  onArchiveNotification,
+}: NotificationRowProps) => {
+  const ref = useRef<HTMLDivElement>(null)
+  const { status, priority } = notification
+  // @ts-ignore
```
alaister, line 34:
```suggestion
```
Still needed?

apps/studio/components/layouts/ProjectLayout/LayoutHeader/NotificationsPopoverV2/NotificationsPopover.tsx
```diff
@@ -0,0 +1,265 @@
...
+    updateNotifications({ ids, status: 'seen' })
+  }
+
+  return (
+    <Popover_Shadcn_
+      modal={false}
+      open={open}
+      onOpenChange={(open) => {
+        setOpen(open)
+        if (!open) markNotificationsRead()
```
alaister, line 86: For making notifications as read, what do you think of having each notification manage it's own read state? So instead of making them all read when you open the popover you do something like this on each row:
```
import { useInView } from 'react-intersection-observer'

const { ref, inView } = useInView()

useEffect(() => {
  if (inView) {
    markRead()
  }
}, [inView])

<div ref={ref}>
```

That way you can't accidentally mark notifications as read if they're out of view.

alaister (approved):
LGTM!

Couple of small nits:

apps/studio/components/layouts/ProjectLayout/LayoutHeader/NotificationsPopoverV2/NotificationsPopover.tsx
```diff
@@ -0,0 +1,265 @@
...
+                />
+              ) : (
+                <div className="flex flex-col gap-y-4 items-center flex-grow justify-center">
+                  <IconInbox size={32} className="text-foreground-light" />
+                  <div className="flex flex-col gap-y-1">
+                    <p className="text-foreground-light text-sm w-64 text-center">
+                      {activeTab === 'archived'
+                        ? `No archived ${
+                            selectedFilter !== undefined ? `${selectedFilter} ` : ''
+                          }notifications`
```
loong, line 246: Copy is a bit weird when "All Notifications" filter is on and it becomes "No archived all notifications". Maybe just "No archived notifications" would be better.
Screenshot 2024-01-04 16 21 01@2x

joshenlim: oh whoops good catch this!

apps/studio/components/layouts/ProjectLayout/LayoutHeader/NotificationsPopoverV2/NotificationsPopover.tsx
joshenlim: great point this - and a better behaviour, i'll look into it 🙏

PR #19168, opened 2023-11-23
Remove combineChunks from SvelteKit browser client cookie getter

briankintz:

Bug fix in docs example code for SvelteKit client initialization for SSR.

Sign out via browser client doesn't work due to doubled cookie chunk combining.

See https://github.com/supabase/auth-helpers/issues/692 

## What is the new behavior?

Since chunking is handled by @supabase/ssr, change the example code to simply return the requested cookie.


charislam:
@silentworks, do you mind taking a look at this one? I think you added `combineChunks` to the example a few weeks ago to solve a bug, what is the correct solution here?

AlbertMarashi:
Err can we just make sure that removing this code doesn't cause sign in to break again, because I think that's why it was added in the first place

silentworks:
@AlbertMarashi is correct about removing this, it will break sign in. We will need to investigate this more in order find a solution that works in both instance. @briankintz I would suggest signing out from the server client instead of the browser client as this works without any issues when I tested this.

AlbertMarashi:
@silentworks this fix seemed to have fixed signout for me, and sign-in is still working too. 

https://github.com/supabase/auth-helpers/issues/692#issuecomment-1822767903 

not sure what needs to happen

PaperPrototype:
> @AlbertMarashi is correct about removing this, it will break sign in. We will need to investigate this more in order find a solution that works in both instance. @briankintz I would suggest signing out from the server client instead of the browser client as this works without any issues when I tested this.

How exactly would you signout from the server?

briankintz:
> @AlbertMarashi is correct about removing this, it will break sign in. We will need to investigate this more in order find a solution that works in both instance. @briankintz I would suggest signing out from the server client instead of the browser client as this works without any issues when I tested this.

@silentworks I'm sure you know the codebase better than I do, but I can't see how this is a problem. The signIn methods from `gotrue-js` all use `get/set/removeItem`, which in the SSR package all handle chunking internally and expect the cookie `get/set/remove` methods to only operate on a single given chunk. Adding a second layer of chunking doesn't make sense to me and prevents chunked cookies from being deleted by [causing `deleteChunks` to return early](https://github.com/supabase/auth-helpers/blob/main/packages/ssr/src/utils/chunker.ts#L74).

silentworks (approved):
Thank you for this PR.

PR #19171, opened 2023-11-23
fix: Pass searchParams during SignInWithSSO rendering

kamilogorek:
Replaces https://github.com/supabase/supabase/pull/19170 to work with SSR.

### DIFF
--- a/apps/studio/components/interfaces/SignIn/SignInWithSSO.tsx
+++ b/apps/studio/components/interfaces/SignIn/SignInWithSSO.tsx
@@ -1,10 +1,17 @@
 import Link from 'next/link'
 import { Button, IconLock } from 'ui'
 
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/SignIn/SignInWithSSO.tsx
```diff
@@ -1,10 +1,13 @@
 import Link from 'next/link'
 import { Button, IconLock } from 'ui'
 
-const SignInWithSSO = () => {
+const SignInWithSSO = ({ searchParams }: { searchParams?: string}) => {
```
kamilogorek, line 4:
```suggestion
const SignInWithSSO = ({ searchParams }: { searchParams?: string }) => {
```

ivasilov (approved):

PR #19178, opened 2023-11-23
chore: fixed issue when delete sorted column cause crash on table editor

### LINKED ISSUE #18955: Deleting a sorted column crashes the Table Editor
If a table is sorted by column, and you delete that column, the table will crash.

To reproduce: 

1. create a table in the Table Editor
2. sort by a column, doesn't matter which one 
3. open a new tab (important), go to the SQL Editor 
4. delete that column via sql:  `ALTER TABLE your_table DROP COLUMN your_column`
5. go back to the original tab and hit refresh, see error since the sort column no longer exists

screenshot-2023-11-14-at-11 21 19

tranhoangvuit:

fixes: https://github.com/supabase/supabase/issues/18955

- When the column we filtered with gets deleted, it will cause an issue: `column \"name1\" does not exist` 
- With the SQL editor, we currently do not support saving queries in local development.
- If there is something wrong with the result of the query, it is simply ignored, which causes an exception.

## What is the new behavior?

Fixed those issues.

## Additional context

Should we refactor the error handling? now we have to two layers for getting error.

### DIFF
--- a/apps/studio/components/grid/SupabaseGrid.tsx
+++ b/apps/studio/components/grid/SupabaseGrid.tsx
@@ -84,7 +84,7 @@ const SupabaseGridLayout = (props: SupabaseGridProps) => {
     {
       keepPreviousData: true,
       retryDelay: (retryAttempt, error: any) => {
-        if (error && error.error.message?.includes('does not exist')) {
+        if (error && error.message?.includes('does not exist')) {
           setParams((prevParams) => {
             return {
               ...prevParams,
--- a/apps/studio/components/grid/SupabaseGrid.tsx
+++ b/apps/studio/components/grid/SupabaseGrid.tsx
@@ -69,7 +69,14 @@ const SupabaseGridLayout = (props: SupabaseGridProps) => {
   const roleImpersonationState = useRoleImpersonationStateSnapshot()
 
   const { project } = useProjectContext()
-  const { data, error, isSuccess, isError, isLoading, isRefetching } = useTableRowsQuery(
+  const {
+    data,
+    error: rqError,
+    isSuccess,
+    isError,
+    isLoading,
+    isRefetching,
+  } = useTableRowsQuery(
     {
       queryKey: [props.table.schema, props.table.name],
       projectRef: project?.ref,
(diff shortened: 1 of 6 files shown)


apps/studio/components/grid/SupabaseGrid.tsx
```diff
@@ -87,6 +87,16 @@ const SupabaseGridLayout = forwardRef<SupabaseGridRef, SupabaseGridProps>(
       }
     )
 
+    if (error?.error?.message.includes('does not exist')) {
```
tranhoangvuit, line 90: I'm not sure about this one because we have an 'error' within 'error', and it's of 'any' type. Is there a better way to handle it?
Full error message here: `column "name" does not exist`

apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityTabResults.tsx
```diff
@@ -110,7 +110,7 @@ const UtilityTabResults = ({ id, isExecuting }: UtilityTabResultsProps) => {
                   </pre>
                 ))
               ) : (
-                <p className="font-mono text-sm">{result.error.error}</p>
+                <p className="font-mono text-sm">{result.error.error?.message}</p>
```
tranhoangvuit, line 113: Recently, if we type something wrong in a query, instead of showing an error, it throws an exception because our response from pg-meta has changed; it is now an object.

kiwicopple:
looks like there is a check failing here:

image

saltcod:
quick thought here for an easier win @tranhoangvuit — think it's possible to check if a column exists before rendering the table?

tranhoangvuit:
> quick thought here for an easier win @tranhoangvuit — think it's possible to check if a column exists before rendering the table?

Thank you for your suggestion. However, keeping the filter with an empty value and the table continuously loading may confuse the user ?

apps/studio/components/grid/SupabaseGrid.tsx
tranhoangvuit: I think we can add `?` before includes to prevent it. but it will be better if we handle in `table-rows-query` part ?

ivasilov:
Hey, I was solving a similar issue in https://github.com/supabase/supabase/pull/19833, can you check it out and fix your PR to handle that issue? I get an error screen when trying to access the table editor with a non-existent column to sort by.

tranhoangvuit:
> Hey, I was solving a similar issue in https://github.com/supabase/supabase/pull/19833, can you check it out and fix your PR to handle that issue? I get an error screen when trying to access the table editor with a non-existent column to sort by.

Thank you for the nice suggestion; that approach will help handle errors more quickly. However, the way to check for errors is different. I logged the error but cannot find error.routine === 'errorMissingColumn'. This is an error on my side.

image

We formatted the error here: https://github.com/supabase/supabase/blob/master/apps/studio/data/sql/execute-sql-query.ts#L25

alaister (approved):

ivasilov (approved):

PR #19188, opened 2023-11-24
Fix button padding with icon

joshenlim:
For some reason `cn` was stripping away the `pl-10` class that would be part of `inputClasses` if an icon is provided

### DIFF
--- a/packages/ui/src/components/Input/Input.tsx
+++ b/packages/ui/src/components/Input/Input.tsx
@@ -142,6 +142,8 @@ function Input({
   if (disabled) inputClasses.push(__styles.disabled)
   if (inputClassName) inputClasses.push(inputClassName)
 
+  console.log(cn(inputClasses))
+
   return (
     <FormLayout
       label={label}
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Input/Input.tsx
```diff
@@ -142,6 +142,8 @@ function Input({
   if (disabled) inputClasses.push(__styles.disabled)
   if (inputClassName) inputClasses.push(inputClassName)
 
+  console.log(cn(inputClasses))
+
```
alaister, line 146:
```suggestion
```

MildTomato:
hang on, shall we just reorder the classes instead. so cn works ?

packages/ui/src/components/Input/Input.tsx
```diff
@@ -171,7 +171,7 @@ function Input({
           ref={inputRef}
           type={type}
           value={reveal && hidden ? HIDDEN_PLACEHOLDER : value}
-          className={cn(inputClasses)}
+          className={inputClasses.join(' ')}
```
MildTomato, line 174: if (error) inputClasses.push(__styles.variants.error)
  if (!error) inputClasses.push(__styles.variants.standard)
  if (size) inputClasses.push(__styles.size[size])
  if (disabled) inputClasses.push(__styles.disabled)
  if (inputClassName) inputClasses.push(inputClassName)
  if (icon) inputClasses.push(__styles.with_icon)
  
  try moving this to bottom ?

alaister (approved):

PR #19191, opened 2023-11-24
Chore/db design input component

MildTomato:

- textarea now used for input
- moved components to app directory
- input component now reusable
- `/new` route removed

## What is the new behavior?

## Additional context


apps/database-new/app/NewThreadInput.tsx
```diff
@@ -0,0 +1,81 @@
+'use client'
```
MildTomato, line 1: this can be moved to server probably

MildTomato: can do it later

saltcod (approved):

Issue #19218, opened 2023-11-25
Sveltekit example uses `const` for a variable that is being reassigned


probablyVincentIp:
# Improve documentation

## Link 

https://supabase.com/docs/guides/getting-started/tutorials/with-sveltekit

## Describe the problem

The documentation initializes destructured variables as `const` but reassigns in the next line

## Describe the improvement

a `const` is not reassignable (should be `let`)

## Additional context

n/a

PR #19220, opened 2023-11-25
fixed: Sveltekit example uses const for a variable that is being reas…

AbhijeetBiswal:
fixed issue :- #19218


kevcodez:
Thanks https://github.com/supabase/supabase/pull/19217 was slightly faster ;)

PR #19221, opened 2023-11-25
feat: sidebar show and close

Hallidayo:

Feature - Supabase Studio

When working on a big table and other screens it would be nice to expand the width of the page so that you can see more.

## What is the new behavior?

I've added show and hide functionality just to give more width to the main panel:

https://github.com/supabase/supabase/assets/22655069/51039e6e-d5fb-499d-a204-001941d40e7d

This example is in auth but the change would happen on the majority of pages e.g. storage and table edtior.

### DIFF
--- a/apps/studio/components/layouts/ProjectLayout/ProductMenuBar.tsx
+++ b/apps/studio/components/layouts/ProjectLayout/ProductMenuBar.tsx
@@ -1,25 +1,41 @@
-import { PropsWithChildren } from 'react'
+import { ArrowLeft, ArrowRight } from 'lucide-react'
+import { PropsWithChildren, useState } from 'react'
 
 interface ProductMenuBarProps {
   title: string
 }
 
 const ProductMenuBar = ({ title, children }: PropsWithChildren<ProductMenuBarProps>) => {
+  const [open, setOpen] = useState(true)
   return (
     <div
       className={[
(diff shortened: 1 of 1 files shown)


kiwicopple:
my main gripe with this is that there is an annoying bar with nothing in it. 

image

I like the idea of this, but i think we can imrpove the DX. I don't know if we need to provide a function to "collapse" it, more that we could add some simple responsiveness (check out the OpenAI dashboard for an example). Thoughts?

Isaiah-Hamilton:
Something like [this](https://github.com/supabase/supabase/pull/7387#issue-1281678922) could work but fully open and fully closed + what @kiwicopple suggested with the OpenAI dashboard as an example

Hallidayo:
@kiwicopple & @Isaiah-Hamilton - I've had a play around and came up with this which is similar to OpenAI:


https://github.com/supabase/supabase/assets/22655069/5da33032-0185-4d89-8ef6-4c762ddfb40b

saltcod:
Hi @Hallidayo!   Thanks for this — a clever idea. We've actually taken some inspiration on a recent global nav iteration. We're looking at collapsing/able sidebars, among lots of other things. I'm going to mark this one as closed, but appreciate the thoughtful PR!

PR #19244, opened 2023-11-27
docs: add schame name for wrappers extension creation

burmecia:

This PR is to add schema name `extensions` when create wrappers extension.

Without specifying schema name, wrappers extension will create the stats collection table `wrappers_fdw_stats` in `public` schema which will cause RLS warning.

## What is the new behavior?

Wrappers extension should be created in `extensions` schema.

## Additional context

N/A

### DIFF
--- a/apps/docs/pages/guides/database/extensions/wrappers/overview.mdx
+++ b/apps/docs/pages/guides/database/extensions/wrappers/overview.mdx
@@ -47,7 +47,7 @@ from
 {/* prettier-ignore */}
 ```sql
 -- Enable the "wrappers" extension
-create extension wrappers with schema extensions;;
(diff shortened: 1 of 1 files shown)


kamilogorek (approved):

apps/docs/pages/guides/database/extensions/wrappers/overview.mdx
```diff
@@ -47,7 +47,7 @@ from
 {/* prettier-ignore */}
 ```sql
 -- Enable the "wrappers" extension
-create extension wrappers;
+create extension wrappers with schema extensions;;
```
kamilogorek, line 50:
```suggestion
create extension wrappers with schema extensions;
```

PR #19247, opened 2023-11-27
Update CODEOWNERS

inian:
the intention before was to let anyone approve apps/docs/pages but this led to an increase in notifications for everyone. 

removing this codeowner rule should have the same effect.


alaister (approved):

loong (approved):
+100

PR #19253, opened 2023-11-27
chore: Bump the OpenAI lib version to v4 in the docs app

ivasilov:
This PR bumps the `openai` lib to v4 and migrates all its usage in the `docs` app. All things should work as before.

### DIFF
--- a/apps/docs/scripts/search/generate-embeddings.ts
+++ b/apps/docs/scripts/search/generate-embeddings.ts
@@ -1,11 +1,9 @@
 import { createClient } from '@supabase/supabase-js'
 import dotenv from 'dotenv'
-import 'openai'
-import { Configuration, OpenAIApi } from 'openai'
-import { inspect } from 'util'
+import { parseArgs } from 'node:util'
+import { OpenAI } from 'openai'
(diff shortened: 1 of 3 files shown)


apps/docs/scripts/search/generate-embeddings.ts
```diff
@@ -193,19 +191,14 @@ async function generateEmbeddings() {
...
-          const openai = new OpenAIApi(configuration)
+          const openai = new OpenAI({ apiKey: process.env.OPENAI_KEY })
 
-          const embeddingResponse = await openai.createEmbedding({
+          const embeddingResponse = await openai.embeddings.create({
             model: 'text-embedding-ada-002',
             input,
           })
 
-          if (embeddingResponse.status !== 200) {
```
ivasilov, line 204: This error will still be handled by the catch block on line 220 (although without `inspect`). Should I add `inspect` there/

gregnr: Should be okay now without the `inspect` 👍

gregnr (approved):

charislam (approved):

PR #19264, opened 2023-11-28
Assign keys to each side panel editor on the table editor to ensure that input fields have the latest data

joshenlim:
There's an issue in the Table Editor when you're editing rows in the side panel, specifically for column types that are rendering the `Listbox` component, whereby the data rendered in that input field is stale (from the previous row that you opened). This was caused by the `Listbox` component not re-rendering correctly when the `value` passed to it has changed

We're assigning keys to each editor in the side panel to ensure that the components in it are always re-rendered whenever the selected row/column/table changes - will make sure we don't run into stale data issues there.


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
```diff
@@ -513,10 +513,34 @@ const SidePanelEditor = ({
     }
   }
 
+  // To force a re-rendering of the input fields in each side panel so that
+  // all the form fields are having the latest state
```
MildTomato, line 517: i still think this is a hack.

fine for now to fix the issue for time being.

alaister (approved):

PR #19269, opened 2023-11-28
docs: remove dart v0 and add dart v2

dshukertjr:

- Remove Dart v0 docs. This probably should have been removed when the Dart SDK became v1.
- Adds Dart v2 docs


spec/supabase_dart_v2.yml
```diff
@@ -0,0 +1,3024 @@
...
+            ''');
+          ```
+      - id: query-the-same-foreign-table-multiple-times
+        name: Query the same foreign table multiple times
+        description: |
+          Sometimes you will need to query the same foreign table twice.
+          In this case, you can use the name of the joined column to identify
+          which join you intend to use. For convenience, you can also give an
+          alias for each column. For example, if we had a shop of products,
+          and we wanted to get the supplier and the purchaser at the same time
```
charislam, line 886: I think the example description doesn't match the code? example is about product but code is about messages

spec/supabase_dart_v2.yml
```diff
@@ -0,0 +1,3024 @@
...
+          ```dart
+          final List<FileObject> objects = await supabase
+            .storage
+            .from('avatars')
+            .list();
+          ```
+  - id: using-modifiers
+    title: Using Modifiers
+    description: |
+      Filters work on the row level—they allow you to return rows that
```
charislam, line 1756:
```suggestion
      Filters work on the row level. That is, they allow you to return rows that
```

spec/supabase_dart_v2.yml
```diff
@@ -0,0 +1,3024 @@
...
+            ```
+  - id: or
+    title: or()
+    description: |
+      Finds all rows satisfying at least one of the filters.
+    notes: |
+      - `.or()` expects you to use the raw [PostgREST syntax](https://postgrest.org/en/stable/api.html#horizontal-filtering-rows) for the filter names and values.
+
+        ```dart
+        .or('id.in.(6,7),arraycol.cs.{"a","b"}')  // Use Postgres list () and 'in' for in_ filter. Array {} and 'cs' for contains.
```
charislam, line 2015:
```suggestion
        .or('id.in.(6,7),arraycol.cs.{"a","b"}')  // Use Postgres list () and 'in' instead of `inFilter`. Array {} and 'cs' for contains.
```

if I remember the above correctly this has been renamed right?

spec/supabase_dart_v2.yml
```diff
@@ -0,0 +1,3024 @@
...
+    description: |
+      Finds all rows which doesn't satisfy the filter.
+    notes: |
+      - `.not()` expects you to use the raw [PostgREST syntax](https://postgrest.org/en/stable/api.html#horizontal-filtering-rows) for the filter names and values.
+
+        ```dart
+        .not('name','eq','Paris')
+        .not('arraycol','cs','{"a","b"}') // Use Postgres array {} for array column and 'cs' for contains.
+        .not('rangecol','cs','(1,2]') // Use Postgres range syntax for range column.
+        .not('id','in','(6,7)')  // Use Postgres list () and 'in' for in_ filter.
```
charislam, line 2051:
```suggestion
        .not('id','in','(6,7)')  // Use Postgres list () and 'in' instead of `inFilter`.
```

spec/supabase_dart_v2.yml
```diff
@@ -0,0 +1,3024 @@
...
+            .delete()
+            .not('name', 'eq', 'Paris');
+          ```
+      - id: with-rpc
+        name: With `rpc()`
+        code: |
+          ```dart
+          // Only valid if the Stored Procedure returns a table type.
+          final data = await supabase
+            .rpc('echo_all_cities)
```
charislam, line 2089:
```suggestion
            .rpc('echo_all_cities')
```

spec/supabase_dart_v2.yml
```diff
@@ -0,0 +1,3024 @@
...
+        code: |
+          ```dart
+          // Only valid if the Stored Procedure returns a table type.
+          final data = await supabase
+            .rpc('echo_all_cities')
+            .ilike('name', '%la%');
+          ```
+
+  - id: is
+    title: is_()
```
charislam, line 2483: I thought this was changed to `isFilter`, am I wrong?

spec/supabase_dart_v2.yml
```diff
@@ -0,0 +1,3024 @@
...
+        code: |
+          ```dart
+          // Only valid if the Stored Procedure returns a table type.
+          final data = await supabase
+            .rpc('echo_all_cities')
+            .is_('name', null);
+          ```
+
+  - id: in
+    title: in_()
```
charislam, line 2528: I thought this was changed to `inFilter`, am I wrong?

charislam (approved):
Nice! What a monster PR 🚀 

A few suggestions for polish

One thing you might want to look at is formatting for code blocks within the upgrade guide. Some of the examples are a bit hard to read due to the indentation and line break patterns, I tried to fix some of the more egregious ones

spec/supabase_dart_v2.yml
dshukertjr: Great catch!

spec/supabase_dart_v2.yml
dshukertjr: Sorry, not sure how this happened. Will fix it.

spec/supabase_dart_v2.yml
```diff
@@ -0,0 +1,3024 @@
...
+          ```dart
+          final res = await supabase.auth.mfa.challengeAndVerify(
+            factorId: '34e770dd-9ff9-416c-87fa-43b31d7ef225',
+            code: '123456',
+          );
+          ```
+  - id: mfa-unenroll
+    title: 'mfa.unenroll()'
+    notes: |
+      Unenroll removes a MFA factor. A user has to have an aal2 authenticator level in order to unenroll a verified factor.
```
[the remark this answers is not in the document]
dshukertjr: Added a link to the [MFA guide](https://supabase.com/docs/guides/auth/auth-mfa#overview) at the MFA overview secdion.

spec/supabase_dart_v2.yml
dshukertjr: Thanks for catching this. I have updated the code to the correct one.

apps/docs/docs/ref/dart/v1/installing.mdx
```diff
@@ -0,0 +1,42 @@
+---
```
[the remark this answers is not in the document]
dshukertjr: Not sure what the cause was then, but it seems to show up now!
https://docs-git-docs-flutter-v2-supabase.vercel.app/docs/reference/dart/installing?framework=flutter

PR #19277, opened 2023-11-28
fix: reset payment method on org creation error

kevcodez:
Users can get into a deadlock when the payment method creation + Stripe customer succeeds, but the initial charge for the plan fails.

PR ensures we reset the payment method upon org creation errors to avoid deadlocks. I've added URL params, as the page refreshes on payment method reset (from Stripe, not from us)

### DIFF
--- a/apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
+++ b/apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
@@ -3,7 +3,7 @@ import type { PaymentMethod } from '@stripe/stripe-js'
 import { useQueryClient } from '@tanstack/react-query'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import {
   Button,
   IconEdit2,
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
```diff
@@ -58,11 +58,11 @@ const NewOrgForm = ({ onPaymentMethodReset }: NewOrgFormProps) => {
   const stripe = useStripe()
   const elements = useElements()
 
-  const { plan } = useParams()
+  const { plan, name, kind, size, spend_cap } = useParams()
 
-  const [orgName, setOrgName] = useState('')
-  const [orgKind, setOrgKind] = useState(ORG_KIND_DEFAULT)
-  const [orgSize, setOrgSize] = useState(ORG_SIZE_DEFAULT)
```
MildTomato, line 65: I only glanced at this, but if things are in params shouldn't we just use them, and not store them in useState as well ?

apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
```diff
@@ -58,11 +58,11 @@ const NewOrgForm = ({ onPaymentMethodReset }: NewOrgFormProps) => {
...
+  const [orgName, setOrgName] = useState(name || '')
+  const [orgKind, setOrgKind] = useState(kind || ORG_KIND_DEFAULT)
```
joshenlim, line 64: i just realised - shouldn't `ORG_KIND_DEFAULT` be `ORG_KIND_TYPES.PERSONAL` instead of `"PERSONAL"`?
Below where you do `Object.entries` on `ORG_KIND_TYPES` the values are the sentence cased values instead of the keys

joshenlim, line 64: low prio since i doubt many will run into this - but we should add validations against `kind` and `size`, if an invalid value is provided in the URL the listbox will not render the option. can probably default to XXX_DEFAULT

joshenlim (approved):

PR #19304, opened 2023-11-29
feat: Add streaming responses to the new RLS Editor

ivasilov:


apps/studio/app/api/ai/sql/suggest/route.ts
```diff
@@ -0,0 +1,132 @@
...
+          headers: { Allow: 'POST' },
+        }
+      )
+  }
+}
+
+async function handlePost(req: NextRequest) {
+  const openai = new OpenAI({ apiKey: openAiKey })
+
+  let body = await (req.json() as Promise<{
```
joshenlim, line 49:
```suggestion
  const body = await (req.json() as Promise<{
```

apps/studio/app/api/ai/sql/suggest/route.ts
```diff
@@ -0,0 +1,132 @@
...
+    messages: { content: string; role: 'user' | 'assistant' }[]
+    entityDefinitions: string[]
+    policyDefinition: string
+  }>)
+
+  let { messages, entityDefinitions, policyDefinition } = body
```
joshenlim, line 55:
```suggestion
  const { messages, entityDefinitions, policyDefinition } = body
```

apps/studio/app/api/ai/sql/suggest/route.ts
```diff
@@ -0,0 +1,132 @@
...
+        - You can add short explanations to your messages.
+        - The result should be a valid markdown. The SQL code should be wrapped in \`\`\`.
+        - Always use "auth.uid()" instead of "current_user".
+        - Only use "WITH CHECK" on INSERT or UPDATE policies.
+        - The policy name should be short text explaining the policy, enclosed in double quotes.
+        - Always make sure that every \`\`\` has a corresponding ending tag \`\`\`.
+        - Always put explanations as separate text. Don't use inline SQL comments. 
+        
+        The output should look like this: 
+        "CREATE POLICY user_policy ON users FOR INSERT USING (user_name = current_user) WITH (true);"
```
joshenlim, line 78: we'll need to dynamically render this i suppose if policyDefinition is provided

apps/studio/app/api/ai/sql/suggest/route.ts
```diff
@@ -0,0 +1,132 @@
...
+export const runtime = 'edge'
+
+const openAiKey = process.env.OPENAI_KEY
+
+/**
+ * This route has to be with an edge runtime to support streaming. This route is not using the
+ * apiWrapper because it's not working with edge API routes (different params) but the route doesn't
+ * require authentication.
+ */
+export async function POST(request: NextRequest) {
```
joshenlim, line 15: just for consistency

```suggestion
export async function handler(request: NextRequest) {
```

apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/index.tsx
```diff
@@ -97,23 +80,26 @@ export const AIPolicyEditorPanel = memo(function ({
...
+      entityDefinitions: isOptedInToAI ? entityDefinitions : undefined,
+      policyDefinition:
+        selectedPolicy !== undefined ? generatePolicyDefinition(selectedPolicy) : undefined,
     },
   })
 
+  const messages = useMemo(() => {
+    const merged = [...debugThread, ...chatMessages.map((m) => ({ ...m, isDebug: false }))]
+
+    return sortBy(merged, (m) => m.createdAt)
```
joshenlim, line 100: alaister ran into some scenarios where by the createdAt from the user and assistant were the exact same timestamps, leading to messages rendering in the wrong order sometimes

e.g 
user
assistant
assistant
user

our fix previously was to also sort on role if the created at is the same - but do you reckon we'll run into the same issue with streaming?

joshenlim (approved):

apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/index.tsx
ivasilov: Most probably, adding that trick here also.

apps/studio/app/api/ai/sql/suggest/route.ts
ivasilov: It has to be POST because this is a `/app` route handler.

PR #19310, opened 2023-11-29
feat: add postgreSQL Hooks docs

J0:

Adds documentation for PostgreSQL hooks. Should only be merged after hooks go live on prod.


apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,598 @@
...
+
+### Timeouts
+
+Hook timeouts are controlled by a `local` statement timeout scoped to the Postgres Function Hook. Ensure that your hooks complete within **2 seconds** to avoid any errors
+
+### Testing the Hook
+
+We strongly recommend testing the hook prior to going into production via [PgTAP](https://supabase.com/docs/guides/database/extensions/pgtap) or similar. You can enable the extension under Database > Extensions > PgTap
+
+Here's an example of what a test could look like for an MFA Verification test.
```
J0, line 531:
```suggestion
Here's an example of what a test could look like for an MFA Verification hook.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,598 @@
...
+	from authenticated, anon;
+```
+
+### Timeouts
+
+Hook timeouts are controlled by a `local` statement timeout scoped to the Postgres Function Hook. Ensure that your hooks complete within **2 seconds** to avoid any errors
+
+### Testing the Hook
+
+We strongly recommend testing the hook prior to going into production via [PgTAP](https://supabase.com/docs/guides/database/extensions/pgtap) or similar. You can enable the extension under Database > Extensions > PgTap
```
J0, line 529:
```suggestion
We strongly recommend testing the hook prior to going into production via [PgTAP](https://supabase.com/docs/guides/database/extensions/pgtap) or similar. You can enable the extension under `Database > Extensions > PgTap`
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,598 @@
...
+    end
+
+
+
+    subgraph Auth Hook
+    ValidateInput ---->CallHook[Execute PostgreSQL Function]
+
+    CallHook ---> ValidateOutput((Validate Outputs))
+    end`}
+    />
```
hf, line 56: Create a task for design team to make this prettier.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,598 @@
...
+
+
+
+## Use Cases
```
hf, line 60: What I would suggest here is to:

- Split the "spec" from the "example" -- have one section explaining each hook's description, inputs and outputs (maybe a table would work? try it)
- In the examples section use tabs so the page does not get huge / busy.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+<div className="border-b pb-3">
+### Example: Add Custom Claims to Token
+<Accordion.Item
+        header={<span className="text-foreground font-bold">Details </span>}
+        id={`add-custom-claims-to-token`}
+      >
+
+****Scenario****
+
+You wish to assign roles to users. In particular, you wish to assign users with a `@supabase.com` email account a role of admin so that they can perform restricted actions
```
hf, line 491: Maybe be more direct in the scenario. You can word it something like:

Often you find yourself needing to assign different permissions to different users. One way of doing this is by adding a custom `admin` boolean property in the `user_metadata` section of the JWT, which can then be picked up by RLS policies or your backend APIs.

...

Then, instead of using the email-domain approach, which I don't think is very useful in general, maybe you can fetch from a table `public.admins` which has a `user_id` column referencing if the user is an admin or not.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
hf: Further comments on this I think a structure similar to this one conveys the message better:

1. Page title and intro text
2. What are hooks? (Technical explanation, table of available hooks)
3. Creating and managing hooks
4. Available hooks
    1. Custom access token
        1. Inputs, outputs, best practices
        2. Examples (Tabs)
    3. MFA verification attempts
        1. Inputs, outputs, best practices
        2. Examples (Tabs)
    4. Password verification attempts
        1. Inputs, outputs, best practices
        2. Examples (Tabs)
5. Frequently asked questions

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+    - `password`
+    - `otp`
+    - `totp`
+    - `recovery`
+    - `invite`
+    - `sso/saml`
+    - `magiclink`
+    - `email/signup`
+    - `email_change`
+    - `token_refresh`
```
hf, line 210: These look too long, maybe add them `as`, `a`, `list`?

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+- `claims` A json containing the updated claims after the hook has been run
+
+****************On Error****************
+
+Return this if your hook processed the input with some error.
+
+- `error` An object that contains information about the error.
+    - `http_code` A number indicating the HTTP code to be returned. If not set, the code is HTTP 500 Internal Server Error.
+    - `message` A message to be returned in the HTTP response. Must be set.
+- In general all other fields attached to the returned JSON are ignored.
```
hf, line 229: You can move this duplicate error explanation to the technical overview or the section about writing hooks.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+
+
+
+<Admonition type="warn">
+Claims returned must conform to our specification. While you are free to add claims, you can only remove the `user_metadata` and app_metadata claims as we rely on the rest  of the claims for proper functioning of the service.
+</Admonition>
```
hf, line 235: This shouldn't be an admonition. It should be explained properly at the top of the hook's description. Please add the list of claims to the User Sessions page and link to them here.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+  user_id uuid not null,
+  factor_id uuid not null,
+  last_failed_at timestamp not null default now(),
+  primary key (user_id, factor_id)
+);
+```
+
+Then you should create a hook which will read and enter information into this table. Example:
+
+```sql
```
hf, line 284: For all SQL code snippets, try to use `  ` tabs to cut down on visual noise on the page.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+
+```jsx
+{
+  "error": {
+    "http_code": 429,
+    "message": "You can only verify a factor once very 10 seconds."
+  }
+}
+```
+## Password Verification Attempts Hook
```
hf, line 131: Add a warning here that explains the following:

Because this hook runs on _unauthenticated_ requests, this function can be called many times by just about anyone. Be careful when using this hook to prevent users from entering your app, as you can accidentally create a vulnerability where attackers can use this feature from preventing legitimate users to enter.

General recommendations:

- Always check if the password is valid first, before taking any additional action. This is a sure signal that the user is legitimate.
- Instead of blocking further password verification attempts, you can use this function to trigger an email or other notification to the user or administrator in your system.

---

In that direction, can you adjust the example to reflect these recommendations?

kangmingtay:
Identical headers cause the navigation panel on the right to pick it up weirdly 
image

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+
+    CallHook ---> ValidateOutput((Validate Outputs))
+    end`}
+    />
+
+
+
+## Use Cases
+
+Based on user feedback, we decided on a few use cases to illustrate how a hook. These use cases are not exhaustive and you can opt to [write your own hook](#writing-your-own-hook) if you have a use case which doesn’t fit any of the examples below.
```
kangmingtay, line 62:
```suggestion
Based on user feedback, we decided on a few use cases to illustrate how a hook works. These use cases are not exhaustive and you can opt to [write your own hook](#writing-your-own-hook) if you have a use case which doesn’t fit any of the examples below.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+- Ask users to wait a set time between verification attempts, to limit the number of attempts that can be done in a period of time.
+- Force sign out users if they have too many invalid verification attempts.
+- Count, rate limit, ban or do other actions.
+
+### Inputs
+
+Supabase Auth will supply the following fields to your hook:
+
+- `factor_id` Unique identifier for the MFA factor the user is verifying. Correlate this to the `auth.mfa_factors` table.
+- `user_id` Unique identifier for the user. Correlate this with `auth.users`
```
kangmingtay, line 79:
```suggestion
- `user_id` Unique identifier for the user.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+- `valid` Whether the verification attempt was valid, in the case for TOTP this means that the six digit code was correct (true) or incorrect (false). The hook will generally receive true once, and false many times for the same user and factor.
```
kangmingtay, line 80:
```suggestion
- `valid` Whether the verification attempt was valid. For TOTP, this means that the six digit code was correct (true) or incorrect (false). The hook will generally receive true once, and false many times for the same user and factor.
```

kangmingtay: > The hook will generally receive true once, and false many times for the same user and factor.

this statement also sounds confusing to me

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+{
+	"factor_id": "6eab6a69-7766-48bf-95d8-bd8f606894db",
+	"user_id": "3919cb6e-4215-4478-a960-6d3454326cec",
+	"valid": true,
+}
+```
+
+### Outputs
+
+In exchange your hook should return back a JSON object with these fields set:
```
kangmingtay, line 94:
```suggestion
In exchange your hook should return a JSON object with these fields set:
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+{
+  "error": {
+    "http_code": 429,
+    "message": "You can only verify a factor once very 10 seconds."
+  }
+}
+```
+## Password Verification Attempts Hook
+
+As with the MFA Verification Hooks, you may wish to increase security beyond the requirements of the default password implementation. You may need to do so meet compliance requirements set out by your client or you might be working in a field where data is sensitive.
```
kangmingtay, line 133:
```suggestion
As with the MFA Verification Hooks, you may wish to increase security beyond the requirements of the default password implementation in order to fulfil security or compliance requirements.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+  "error": {
+    "http_code": 429,
+    "message": "You can only verify a factor once very 10 seconds."
+  }
+}
+```
+
+## Custom Access Token Hook
+
+You may wish to add additional claims based on the authentication method used or add additional metadata to select users. The custom access token hook runs before a token is issued and allows you to add claims.
```
kangmingtay, line 193:
```suggestion
The custom access token hook runs before a token is issued and allows you to add additional claims based on the authentication method used or add additional metadata to select users.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
kangmingtay: we should also call out the claims that can't be removed (prevented by us) like the `role` , `exp`, etc...

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+Based on user feedback, we decided on a few use cases to illustrate how a hook works. These use cases are not exhaustive and you can opt to [write your own hook](#writing-your-own-hook) if you have a use case which doesn’t fit any of the examples below.
+
+## MFA Verification Attempt Hook
+
+Depending on your security or compliance requirements, you may need to enforce stricter restrictions beyond the default [Supabase MFA implementation](https://supabase.com/docs/guides/auth/auth-mfa).
+
+For such scenarios, the MFA Verification Attempt hook can allow you to add additional checks to:
+
+- Ask users to wait a set time between verification attempts, to limit the number of attempts that can be done in a period of time.
+- Force sign out users if they have too many invalid verification attempts.
```
charislam, line 71:
```suggestion
- Sign out users who have too many invalid verification attempts.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+- Count, rate limit, ban or do other actions.
```
charislam, line 72:
```suggestion
- Count, rate limit, or ban signins.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+	"user_id": "3919cb6e-4215-4478-a960-6d3454326cec",
+	"valid": true,
+}
+```
+
+### Outputs
+
+In exchange your hook should return a JSON object with these fields set:
+
+********************************On Success********************************
```
charislam, line 96: I'd prefer to use actual headers here, an `h4` would work well

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+
+Return these only if your hook processed the input without errors.
```
charislam, line 98:
```suggestion
Return this if your hook processed the input without errors.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+```json
+{
+   "decision": "reject",
+   "message" : "You have exceeded maximum number of MFA attempts."
+}
+```
+
+****************On Error****************
+
+Return this if your hook processed the input with some error.
```
charislam, line 114: I'm not entirely clear on what should be a rejection versus an error. I thought perhaps an error would be a server error (application's fault) while a rejection is a rule violation (user's fault), but this example of verifying a factor too often sounds more like a user error, in line with the too many invalid attempts above

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+    - `message` A message to be returned in the HTTP response. Must be set.
+- In general all other fields attached to the returned JSON are ignored.
+
+Example output:
+
+```jsx
+{
+  "error": {
+    "http_code": 429,
+    "message": "You can only verify a factor once very 10 seconds."
```
charislam, line 127:
```suggestion
    "message": "You can only verify a factor once every 10 seconds."
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+
+## Custom Access Token Hook
+
+The custom access token hook runs before a token is issued and allows you to add additional claims based on the authentication method used or add additional metadata to select users.
+
+### Inputs
+
+- `user_id`Unique identifier for the user attempting to sign in. Correlate this to the `auth.users` table.
+- `claims` Claims which are attached to the access token.
+- `authentication_method` the authentication method used to request for the access token. Possible values include:
```
charislam, line 199:
```suggestion
- `authentication_method` the authentication method used to request the access token. Possible values include:
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+
+First you need to create a table that will record the last time a user had an incorrect password verification attempt.
+
+You can do this by creating a table such as:
+
+``sql
+create table public.password_failed_verification_attempts (
+	user_id uuid not null,
+  last_failed_at timestamp not null default now(),
+  primary key (user_id, factor_id)
```
charislam, line 393:
```suggestion
  primary key (user_id, factor_id)
```

no `factor_id` in this one right?

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+        do update
+          set last_failed_at = now();
+
+		-- finally let Supabase Auth do the default behavior for a failed attempt
+		return jsonb_build_object('decision', 'continue');
+	end;
+	$$;
+```
+
+Supabase Auth will invoke this function with the `supabase_auth_admin` role which does not have permissions to modify the `public` schema. So you should grant it appropriate permissions:
```
charislam, line 448: these explanations about granting and revoking permissions would also be good candidates for moving into the explainer up top

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+
+select * from finish();
+rollback;
+````
+
+
+
+## Connecting the Hook to Supabase Auth
+
+Head over to `Authentication > Hooks (Beta)` and select the appropriate PostgreSQL function from the dropdown menu.
```
charislam, line 665:
```suggestion
In the dashboard, navigate to `Authentication > Hooks (Beta)` and select the appropriate PostgreSQL function from the dropdown menu.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
J0: ah yes, good catch. Thanks!

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,669 @@
...
+
+    CallHook ---> ValidateOutput((Validate Outputs))
+    end`}
+    />
+
+
+
+## Use Cases
+
+Based on user feedback, we decided on a few use cases to illustrate how a hook works. These use cases are not exhaustive and you can opt to [write your own hook](#writing-your-own-hook) if you have a use case which doesn’t fit any of the examples below.
```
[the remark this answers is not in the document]
J0: Switched to describing points instead of hooks and section about writing a hook before the section with examples

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+revoke execute
+	on function public.custom_access_token_hook
+	from authenticated, anon;
+```
+
+<Admonition type="warning">
+Avoid adding a `security definer` tag to the function to avoid permission errors.
+</Admonition>
+
+At time of writing, Postgres functions have an implicit tag of `security invoker` and will run with the permissions of the user that calls it. In the case of Auth Hooks, `supabase_auth_admin`  is the invoker and it will run with the permissions of `supabase_auth_admin`.
```
charislam, line 127:
```suggestion
At time of writing, Postgres functions have an implicit tag of `security invoker` and run with the permissions of the calling user. In the case of Auth Hooks, `supabase_auth_admin`  is the invoker and hook functions run with the permissions of `supabase_auth_admin`.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+The `security definer` tag specifies that the hook is to be executed with the privileges of the user that owns it.  In the case of functions created via the Supabase Dashboard this is `postgres` which does not have permissions over the Auth schema. Introducing a `security definer` tag could prevent Supabase Auth from having the appropriate permissions to access tables or execute functions.
+
+
+### Timeouts
+
+Hook timeouts are controlled by a `local` statement timeout scoped to the Postgres Function Hook. Ensure that your hooks complete within **2 seconds** to avoid any errors
```
charislam, line 134:
```suggestion
Hook timeouts are controlled by a `local` statement timeout scoped to the Postgres Function Hook. Ensure that your hooks complete within **2 seconds** to avoid any errors.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+### Testing the Hook
+
+We strongly recommend testing the hook prior to going into production via [PgTAP](https://supabase.com/docs/guides/database/extensions/pgtap) or similar. You can enable the extension under `Database > Extensions > PgTAP`
```
charislam, line 138:
```suggestion
We strongly recommend testing the hook prior to going into production via [PgTAP](/docs/guides/database/extensions/pgtap) or similar. You can enable the extension under `Database > Extensions > PgTAP`
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+Here's an example of what a `PgTAP` test could look like for an MFA Verification hook.
```
charislam, line 140:
```suggestion
Here's an example of a `PgTAP` test for an MFA Verification hook:
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+```sql
+begin;
+select plan (1);
+
+-- table to store test results
+create temporary table test_results (
+  name text primary key
+  , result boolean
+  , debug text
```
charislam, line 150:
```suggestion
  name text primary key,
  result boolean,
  debug text
```

we style all other sql blocks in docs with trailing commas so let's just that for consistency

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+create or replace function tests.record_result(name text, result boolean, debug text DEFAULT NULL)
+  returns void language plpgsql as $$
+begin
+  insert into test_results(name, result, debug) values (name, result, debug);
+end;
+$$;
+
+do $$
+  declare
+     -- Arrange and setup any variables used in the test
```
charislam, line 163:
```suggestion
     -- Arrange and set up any variables used in the test
```

`set up` is the verb, `setup` is the noun

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+    _result jsonb;
+  begin
+    set search_path to 'public';
+
+    insert into auth.users(id) values (_user1);
+
+    -- Act and run the function
+    perform tests.record_result(
+      'Valid event result'
+      , (select hook_mfa_verification_attempt(_valid_user1_event)) ->> 'decision' = 'continue'
```
charislam, line 177:
```suggestion
      'Valid event result',
      (select hook_mfa_verification_attempt(_valid_user1_event)) ->> 'decision' = 'continue'
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+  end;
+$$;
+
+set search_path to 'extensions';
+
+-- Assert that the desired result is received
+select set_eq(
+  $$ select name, result, debug from test_results $$,
+  $$ select * from (values
+    , ('Valid event result', true, null)
```
charislam, line 190:
```suggestion
  $$ select * from (values,
    ('Valid event result', true, null)
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+);
+
+select * from finish();
+rollback;
+````
+
+
+### Connecting the Hook
+
+In the dashboard, navigate to `Authentication > Hooks (Beta)` and select the appropriate PostgreSQL function from the dropdown menu.
```
charislam, line 202: dashboard link here?

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+<Tabs
+  scrollable
+  size="small"
+  type="underlined"
+  defaultActiveId="js"
+  queryGroup="language"
+>
+<TabPanel id="add-admin-role" label="Limit Failed MFA Verification Attempts">
+
+Imagine that your company requires that a user can only input an incorrect MFA Verification code every 2 seconds and not more than that.
```
charislam, line 257:
```suggestion
Imagine that your company requires that a user can input an incorrect MFA Verification code no more than once every 2 seconds.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+#### Database Setup
+
+First you need to create a table that will record the last time a user had an incorrect MFA verification attempt for a factor.
```
charislam, line 261:
```suggestion
Dreate a table to record the last time a user had an incorrect MFA verification attempt for a factor.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+You can do this by creating a table such as:
+
```
charislam, line 264:
```suggestion
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+```sql
+create table public.mfa_failed_verification_attempts (
+  user_id uuid not null,
+  factor_id uuid not null,
+  last_failed_at timestamp not null default now(),
+  primary key (user_id, factor_id)
+);
+```
+
+Then you should create a hook which will read and enter information into this table. Example:
```
charislam, line 274:
```suggestion
Create a hook to read and write information to this table. For example:
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+	from authenticated, anon;
+```
+</TabPanel>
+</Tabs>
+
+## Password Verification Attempts Hook
+
+Imagine that you wish to increase security beyond the requirements of the default password implementation in order to fulfill security or compliance requirements.
+
+As the hook runs on unauthenticated requests, malicious users can abuse the hook by calling it multiple times.  Pay extra care when using the hook as you can unintentionally block legitimate users from accessing your application.
```
charislam, line 352:
```suggestion
As this hook runs on unauthenticated requests, malicious users can abuse the hook by calling it multiple times.  Pay extra care when using the hook as you can unintentionally block legitimate users from accessing your application.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+- `decision` A string containing the decision whether to allow authentication to move forward. Use `reject` to completely reject the verification attempt and log the user out of all active sessions. Use `continue` to use the default Supabase Auth behavior.
+- `message` The message to show the user if the decision was `reject`.
+- `should_logout_user` Whether to logout a user if a `reject` decision is issued. Has no effect when a `continue` decision is issued.
+
+Example output:
+
+```json
+{
+   "decision": "reject",
+   "message" : "You have exceeded maximum number of MFA attempts.",
```
charislam, line 381: this is for password not MFA right?

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+Example output:
+
+```json
+{
+   "decision": "reject",
+   "message" : "You have exceeded maximum number of MFA attempts."
+}
+```
+### Example: Limit Failed MFA Verification Attempts
+<Tabs
```
charislam, line 248: Right now the tabs look like this so there's not much purpose to having the tab:

image

I think what you want to do is have one global `Tabs` section, then have three `TabPanels` within it corresponding to the three different hooks

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+  scrollable
+  size="small"
+  type="underlined"
+  defaultActiveId="js"
+  queryGroup="language"
+>
+<TabPanel id="add-admin-role" label="Limit Failed Password Verification Attempts">
+
+As part of new security measures within the company, users can only input an incorrect password every 10 seconds and not more than that. You are charged with impelementing a hook to enforce this.
```
charislam, line 397:
```suggestion
As part of new security measures within the company, users can only input an incorrect password every 10 seconds and not more than that. You want to write a hook to enforce this.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+#### Database Setup
+
+First you need to create a table that will record the last time a user had an incorrect password verification attempt.
```
charislam, line 401:
```suggestion
Create a table to record each user's last incorrect password verification attempt.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+You can do this by creating a table such as:
+
```
charislam, line 404:
```suggestion
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+```sql
+create table public.password_failed_verification_attempts (
+	user_id uuid not null,
+  last_failed_at timestamp not null default now(),
+  primary key (user_id)
```
charislam, line 409:
```suggestion
create table public.password_failed_verification_attempts (
    user_id uuid not null,
    last_failed_at timestamp not null default now(),
    primary key (user_id)
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+);
+```
+
+Then you should create a hook which will read and enter information into this table. Example:
```
charislam, line 413:
```suggestion
Create a hook which to read and write information to this table. For example:
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+create function public.hook_password_verification_attempt(event jsonb)
+returns jsonb
+language plpgsql
+as $$
+	declare
+		last_failed_at timestamp;
+	begin
+		if event->'valid' is true
+		then
+			-- code is valid, accept it
```
charislam, line 425:
```suggestion
			-- password is valid, accept it
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+</TabPanel>
+</Tabs>
+
+## Custom Access Token Hook
+
+The custom access token hook runs before a token is issued and allows you to add additional claims based on the authentication method used or add additional metadata to select users.
+
+Claims returned must conform to our specification. Supabase Auth will check for these claims after the hook is run and return an error if they are not present.
+
+Here's a listing of all fields currently available on an access token:
```
charislam, line 490:
```suggestion
These are the fields currently available on an access token:
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+| `user_metadata` | No                |
+| `role`          | Yes               |
+| `aal`           | Yes               |
+| `amr`           | No                |
+| `session_id`    | No                |
+
+
+### Inputs
+
+- `user_id`Unique identifier for the user attempting to sign in. Correlate this to the `auth.users` table.
```
charislam, line 513:
```suggestion
- `user_id` Unique identifier for the user attempting to sign in. Correlate this to the `auth.users` table.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+<Tabs
+  scrollable
+  size="small"
+  type="underlined"
+  defaultActiveId="js"
+  queryGroup="language"
+>
+<TabPanel id="add-admin-role" label="Add Admin Role">
+
+Your company would like to allow users with a `@supabase.com` email account to perform restricted actions. To support this, users who are registered as admins with the company should be accorded the `admin` role.
```
charislam, line 537: from what I can see the function doesn't currently check if the users are admins right? it just automatically makes everybody with a `@supabase.com` email an admin?

charislam (approved):

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
hf: I'm not sure this is a very useful section. I'd maybe include it as an example for one of the hooks?

apps/docs/public/img/guides/auth-postgres-hooks/hooks-ui.png
hf, line 1: This is an old image. Let's remove it -- I'm not a fan of including screenshots of config screen as they can often evolve and will be forgotten.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,614 @@
...
+```
+### Example: Limit Failed MFA Verification Attempts
+<Tabs
+  scrollable
+  size="small"
+  type="underlined"
+  defaultActiveId="js"
+  queryGroup="language"
+>
+<TabPanel id="add-admin-role" label="Limit Failed MFA Verification Attempts">
```
hf, line 255: Can you make sure these IDs are correct, they end up in the URL.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
J0: Yes, thanks!

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
J0: That could work too - I've kept the tabs gone ahead and added a draft example for each extensibility point but let me know if another style of organizing the content is preferred

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,804 @@
...
+
+## Custom Access Token Extensibility Point
+
+The custom access token hook runs before a token is issued and allows you to add additional claims based on the authentication method used or add additional metadata to select users.
+
+Claims returned must conform to our specification. Supabase Auth will check for these claims after the hook is run and return an error if they are not present.
+
+These are the fields currently available on an access token:
+
+| Field           | Is Required Claim |
```
kangmingtay, line 638:
```suggestion
| Field           | Required |
```

kangmingtay (commented):
image

We also opt to make this more concise by changing the present participle terms to their verbs. So it'll be:

Create and manage hooks
  - Write a Hook
  - Handle Hook errors
  - Assign Hook permissions
  - Timeouts
  - Connect the Hook

Thanks for summarising the changes too!

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
+import Layout from '~/layouts/DefaultGuideLayout'
+import {CodeBlock, Admonition, Accordion} from 'ui'
+import Image from 'next/image'
+
+export const meta = {
+  id: 'auth-postgres-hooks',
+  title: 'Auth Hooks (Postgres)',
```
hf, line 7:
```suggestion
  title: 'Auth Hooks',
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+  description: 'Use Supabase Postgres Functions to customize your authentication flow',
+}
+
+<Admonition type="info">
+Hooks are in beta. They will be available for all projects in early 2024.
+</Admonition>
+
+
+Supabase allows you to use PostgreSQL functions to alter the default Supabase Auth flow. Developers can use hooks to add custom behavior that's not supported natively.
```
hf, line 16:
```suggestion
Supabase allows you to use Postgres functions to alter the default Supabase Auth flow. Developers can use hooks to add custom behavior that's not supported natively.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+- Add additional custom claims to your JWT
+
+
+You can use hooks at specific points along an Auth flow to perform custom behavior via Postgres Functions. These are known as **extensibility points**. The following extensibility points are available:
+
+| Extensibility Point              | Available on Plan       |
+|----------------------------------|-------------------------|
+| MFA Verification Attempt         | Teams and Enterprise    |
+| Password Verification Attempt    | Teams and Enterprise    |
+| Custom Access Token              | Free                    |
```
hf, line 32:
```suggestion
| Extensibility Point              | Available on Plan       |
|----------------------------------|-------------------------|
| Custom Access Token              | Free, Pro                |
| MFA Verification Attempt         | Teams and Enterprise    |
| Password Verification Attempt    | Teams and Enterprise    |
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+declare
+  -- Add your variables here
+begin
+  -- Add your logic here
+  return response
+end;
+$$ language plpgsql;
+```
+
+Note that the JSONB operator `?` can't be used when writing a hook, because it is a reserved character in the ORM used by Supabase Auth.
```
hf, line 63: Huh? Where did this come from? It does not follow from anything in the doc so far.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+
+<Image
+  alt="Auth Hooks Flowchart"
+  width={'1000'}
+  height={'400'}
+  src="/docs/img/guides/auth-postgres-hooks/hooks-flowchart.png"/>
+
+
+## Create a Hook
+### Write a Hook
```
hf, line 47: I think this section is a bit too simplistic. 

Yes, you can mention the template (and less prominently), but the main goal of it should be to explain:

- What does an Auth hook function look like?
- That you can / cannot use JSON/JSONB.
- Link to Postgres docs about JSON/JSONB operators.
- Link to Postgres docs about functions and languages.
- Explain the roles and grants needed for a function.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+  return response
+end;
+$$ language plpgsql;
+```
+
+Note that the JSONB operator `?` can't be used when writing a hook, because it is a reserved character in the ORM used by Supabase Auth.
+
+### Handle Hook errors
+
+Unless otherwise specified, your hook should return errors with the following fields:
```
hf, line 67: Avoid language like "unless" ... 

This is it, this is the spec, it's talking about concrete auth hooks.

You should also add some sentences here to talk about what does an "error" mean in hook terms. It's a runtime error that modifies the behavior of Supabase Auth; not a developer error such as division by zero.

```suggestion
To return an error from a hook, return a JSON object with the following properties:
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+
+- `error` An object that contains information about the error.
+    - `http_code` A number indicating the HTTP code to be returned. If not set, the code is HTTP 500 Internal Server Error.
+    - `message` A message to be returned in the HTTP response. Required.
+
+You should return an error if an input in the `event` payload has been processed incorrectly. This error could result from a server error (for example, an unexpected error with the data) or a user error (for example, the user has exceeded the number of available attempts).
+
+In general, Supabase Auth ignores all fields other than the `error` field attached to the returned JSON.
```
hf, line 75:
```suggestion
When returning an error object, Supabase Auth ignores all other fields you may have attached.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+
+Example output:
+```jsx
```
hf, line 78:
```suggestion
```json
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+{
+  "error": {
+    "http_code": 429,
+    "message": "You can only verify a factor once every 10 seconds."
+  }
+}
+```
+
+### Assign Hook permissions
```
hf, line 87:
```suggestion
### Hook grants and roles
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+revoke execute
+  on function public.custom_access_token_hook
+  from authenticated, anon;
+```
+
+<Admonition type="caution">
+Avoid adding a `security definer` tag to the function to avoid permission errors.
+</Admonition>
+
+At time of writing, Postgres functions have an implicit tag of `security invoker` and run with the permissions of the calling user. In the case of Auth Hooks, `supabase_auth_admin`  is the invoker and hook functions run with the permissions of `supabase_auth_admin`.
```
hf, line 111: Avoid language like "at time of writing." It hasn't changed in years, not gonna change ever.

```suggestion
Postgres functions have an implicit tag of `security invoker` and run with the permissions of the calling user. In the case of Auth Hooks, `supabase_auth_admin`  is the invoker and hook functions run with the permissions of `supabase_auth_admin`.
```

This looks like a great place to reference Postgres docs.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+
+The `security definer` tag executes a function with the privileges of the user who owns it.  In the case of functions created via the Supabase Dashboard, this user is `postgres`, which does not have permissions for the Auth schema. Using `security definer` tag would prevent Supabase Auth from having the appropriate permissions to access tables or execute functions.
```
hf, line 113: Maybe just link to Postgres docs. `security definer` with the correct grants can be easier to work with.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+
+
+### Timeouts
+
+Ensure that your hooks complete within **2 seconds** to avoid any errors.
+
+
+
```
hf, line 121: In general, just use one new line between text.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+### Connect a Hook
+
+In the dashboard, navigate to [`Authentication > Hooks (Beta)`](/dashboard/project/_/auth/hooks) and select the appropriate PostgreSQL function from the dropdown menu.
+
+
+## MFA Verification Attempt Extensibility Point
```
hf, line 127: Headings should be sentence case:

```suggestion
## Hook: MFA verification attempt
```

Also in general, I don't think I like the `extensibility point` language a lot. It's long and hard to pronounce, why not just use `hook` everywhere?

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+```
+**Examples**
+<Tabs
+  scrollable
+  size="small"
+  type="underlined"
+  defaultActiveId="js"
+  queryGroup="language"
+>
+<TabPanel id="limit-failed-mfa-verification-attempts" label="Limit Failed MFA Verification Attempts">
```
hf, line 176: Use sentence case here too:

```suggestion
<TabPanel id="limit-failed-mfa-verification-attempts" label="Limit failed MFA verification attempts">
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
J0: Perhaps we can just link to https://supabase.com/docs/guides/database/functions#security-definer-vs-invoker ?

I think we already do a decent job of explaining this there

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+
+Imagine that your company requires that a user can input an incorrect MFA Verification code no more than once every 2 seconds.
```
hf, line 178: Maybe use general language, like specifying the requirement more than "imagining."

```suggestion
Natively users can try to enter incorrect MFA verification codes without limit. In some cases, you may be required to limit this behavior to allow users to enter a verification code only once in 2 seconds.
```

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+
+#### Database Setup
```
hf, line 180: Everything is database setup with hooks... split it up better? Maybe no headings at all?

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+revoke all
+  on table public.mfa_failed_verification_attempts
+  from authenticated, anon;
+```
+</TabPanel>
+<TabPanel id="send-email-after-failed-attempts" label="Send Email After Failed Verification Attempts">
+
+Enable the `pg_net` extension under [`Database > Extensions`](/dashboard/project/_/database/extensions)
+
+Create a table to record the last time a user had an incorrect MFA verification attempt for a factor.
```
hf, line 268: Let's not just add examples for examples' sake. Try to explain the utility of this example. Why would someone want to use this?

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
J0: The main driver towards using the `extensibility point` or `execution point` was to have a way to distinguish between the hook (specific implementation) vs the location at which the hook is called. 

For instance, one might not necessarily use the `Custom Access Token Hook` to mint access tokens even though that is the main use case we anticipate. In this case the `points` terminology provides a way to distinguish between the set of possible cases (`Point`)  vs the specific implementation (`Custom Access Token Hook`)

Open to suggestions if there is a simpler way of doing so though.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
```diff
@@ -0,0 +1,851 @@
...
+  from authenticated, anon;
+
+revoke all
+  on table public.mfa_failed_verification_attempts
+  from authenticated, anon;
+```
+</TabPanel>
+<TabPanel id="send-email-after-failed-attempts" label="Send Email After Failed Verification Attempts">
+
+Enable the `pg_net` extension under [`Database > Extensions`](/dashboard/project/_/database/extensions)
```
hf, line 266: Maybe add some introduction as to what the example will be doing. Why is pg_net necessary, what service is going to be used to send the email. Adjust the title of the tab accordingly.

apps/docs/pages/guides/auth/auth-postgres-hooks.mdx
J0: Sounds good, will include. The initial intention was that by using a template one could sidestep concerns like:

> What does an Auth hook function look like?
> That you can / cannot use JSON/JSONB.

Since the dev can edit the template and not have to worry about what it looks like.

> Link to Postgres docs about JSON/JSONB operators.
> Link to Postgres docs about functions and languages.

Thanks, it's a good suggestion. Let's add that.

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,720 @@
...
+revoke execute
+  on function public.custom_access_token_hook
+  from authenticated, anon;
+```
+
+</TabPanel>
+<TabPanel id="add-metadata-claim-plv8" label="Add claim via plv8">
+You can add custom claims to the access token. In this example, we fetch the level of an existing user from the `profiles` table and assign new users a default level of zero.
+
+For convenience, we enable the `plv8` extension so that we can run JavaScript within our query.
```
J0, line 630:
```suggestion
We enable the `plv8` extension to run JavaScript within our query.
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,719 @@
...
+Create a `profiles` table to hold information:
+
+```sql
+create table profiles (
+  user_id uuid not null primary key references auth.users (id),
+  level integer,
+  is_admin boolean not null default false
+);
+```
+
```
J0, line 641:
```suggestion
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,718 @@
...
+  last_attempt_at timestamp not null default now(),
+  attempt_successful boolean not null,
+  primary key (user_id, attempt_id)
+);
+```
+
+Next, store the API key of our email API provider:
+
+```sql
+select vault.create_secret('my_api_key', 'my_api_key_name', 'my_email_provider_api_key');
```
J0, line 422:
```suggestion
select vault.create_secret('my_api_key', 'my_api_key_name', 'description_of_my_api_key');
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,718 @@
...
+```
+
```
J0, line 424:
```suggestion
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+  width={'1000'}
+  height={'400'}
+  src="/docs/img/guides/auth-hooks/hooks-flowchart.png"
+/>
+
+## Create a Hook
+
+### What is a Hook
+
+A hook is a [Postgres Function](https://www.postgresql.org/docs/current/sql-createfunction.html) that takes in a `jsonb` object and return a `jsonb` object. You can transform the `jsonb` input in the body of the hook via native operators such as [JSON operators](https://www.postgresql.org/docs/16/functions-json.html) or extensions such as [plv8](https://supabase.com/docs/guides/database/extensions/plv8) which allow you to use JavaScript within Postgres.
```
J0, line 47:
```suggestion
A hook is a [Postgres Function](https://www.postgresql.org/docs/current/sql-createfunction.html) that takes in a `jsonb` object and return a `jsonb` object. You can transform the `jsonb` input in the body of the hook via native operators such as [JSON operators](https://www.postgresql.org/docs/16/functions-json.html) or extensions such as [plv8](https://supabase.com/docs/guides/database/extensions/plv8).
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+
+<Admonition type="caution">
+  The `?` JSONB operator is not supported as it conflicts with our ORM.
+</Admonition>
+
+A hook typically has the following signature:
```
J0, line 53:
```suggestion
Here is an example hook signature:
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+
+Also revoke permissions from the `authenticated` and `anon` roles to ensure the function is not accessible by Supabase Serverless APIs.
+
+```sql
+revoke execute
+  on function public.custom_access_token_hook
+  from authenticated, anon;
+```
+
+For security, we recommend against the use the `security definer` tag. The `security definer` tag specifies that the function is to be executed with the privileges of the user that owns it. When a function is created via the Supabase dashboard with the tag, it will have the extensive permissions of the `postgres` role which could allow for undesirable actions to occur.
```
J0, line 91:
```suggestion
For security, we recommend against the use the `security definer` tag. The `security definer` tag specifies that the function is to be executed with the privileges of the user that owns it. When a function is created via the Supabase dashboard with the tag, it will have the extensive permissions of the `postgres` role which make it easier for undesirable actions to occur.
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+```json
+{
+  "error": {
+    "http_code": 429,
+    "message": "You can only verify a factor once every 10 seconds."
+  }
+}
+```
+
+When an error is returned, the error is propagated upwards and translated into a HTTP error which is then returned to your application. Supabase Auth will only take into account the error and disregard the rest of the payload.
```
J0, line 124:
```suggestion
When an error is returned, the error is propagated from the hook to Supabase Auth and translated into a HTTP error which is returned to your application. Supabase Auth will only take into account the error and disregard the rest of the payload.
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+  // Update the level in the claims
+  event.claims.level = user_level;
+
+  return event;
+$$;
+```
+
+</TabPanel>
+<TabPanel id="restrict-access-to-sso-users" label="Restrict access to SSO users">
+You can restrict access to internal applications with a hook. For example, you require that employees log in via [SAML Single Sign On (SSO)](/docs/guides/auth/sso/auth-sso-saml). You can exempt some employees from the policy via an allowlist.
```
J0, line 671:
```suggestion
You can restrict access to internal applications with a hook. For example, you can require that employees log in via [SAML Single Sign On (SSO)](/docs/guides/auth/sso/auth-sso-saml). You can exempt select employees from the policy via an allowlist.
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+
+## Create a Hook
+
+### What is a Hook
+
+A hook is a [Postgres Function](https://www.postgresql.org/docs/current/sql-createfunction.html) that takes in a `jsonb` object and return a `jsonb` object. You can transform the `jsonb` input in the body of the hook via native operators such as [JSON operators](https://www.postgresql.org/docs/16/functions-json.html) or extensions such as [plv8](https://supabase.com/docs/guides/database/extensions/plv8).
+
+<Admonition type="caution">
+  The `?` JSONB operator is not supported as it conflicts with our ORM.
+</Admonition>
```
hf, line 51:
```suggestion
<Admonition type="caution">
  If you're using the Supabase SQL Editor, there's an issue when using the `?` (*Does the string exist as a top-level key within the JSON value?*) operator. Use a direct connection to the database if you need to use it when defining a function.
</Admonition>
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+  width={'1000'}
+  height={'400'}
+  src="/docs/img/guides/auth-hooks/hooks-flowchart.png"
+/>
+
+## Create a Hook
+
+### What is a Hook
+
+A hook is a [Postgres Function](https://www.postgresql.org/docs/current/sql-createfunction.html) that takes in a `jsonb` object and return a `jsonb` object. You can transform the `jsonb` input in the body of the hook via native operators such as [JSON operators](https://www.postgresql.org/docs/16/functions-json.html) or extensions such as [plv8](https://supabase.com/docs/guides/database/extensions/plv8).
```
hf, line 47:
```suggestion
A hook is a [Postgres Function](https://www.postgresql.org/docs/current/sql-createfunction.html) with a single argument -- the `event` of type [JSONB]() -- and which returns a JSONB object.

This function is invoked by Supabase Auth at its point in the flow by providing an event object. The object returned by the function gives instructions on how Supabase Auth should continue processing.

To access properties of the `event` argument, you can use the [JSON operators and functions](https://www.postgresql.org/docs/current/functions-json.html). 

There are no restrictions as to what language can be used to write Auth Hooks. If [PL/pgSQL](https://www.postgresql.org/docs/current/plpgsql.html) is too difficult consider using the [plv8](/docs/guides/database/extensions/plv8) extension which lets you use JavaScript to define functions.
```

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+```json
+{
+  "decision": "reject",
+  "message": "You have exceeded maximum number of MFA attempts."
+}
+```
+
+**Examples**
+
+<Tabs
```
hf, line 178: For all tabs in the doc, can you check how we can have the first example be active and not hidden?

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+
+## Hook: Custom Access Token
+
+The custom access token hook runs before a token is issued and allows you to add additional claims based on the authentication method used.
+
+Claims returned must conform to our specification. Supabase Auth will check for these claims after the hook is run and return an error if they are not present.
+
+These are the fields currently available on an access token:
+
+| Field           | Required |
```
hf, line 527: This table renders really terribly. Maybe you can convert these into two lists:

Example:

- Required claims: `aud`, ...
- Optional claims: `jti`, ...

apps/docs/pages/guides/auth/auth-hooks.mdx
```diff
@@ -0,0 +1,717 @@
...
+| `nbf`           | No       |
+| `sub`           | Yes      |
+| `email`         | Yes      |
+| `phone`         | Yes      |
+| `app_metadata`  | No       |
+| `user_metadata` | No       |
+| `role`          | Yes      |
+| `aal`           | Yes      |
+| `amr`           | No       |
+| `session_id`    | No       |
```
hf, line 543: This should be a required claim.

apps/docs/public/img/guides/auth-hooks/hooks-flowchart.png
hf, line 1: Maybe remove the diagram until design team creates a good one?

hf (approved):
Feel free to merge after addressing these last comments.

PR #19324, opened 2023-11-30
Chore/enabling branching fixes

joshenlim:
- Close branching modal when navigating to other pages (via the CTAs from enabling PITR notice)
- Remove gate of enabling PITR to enable branching, swapped for a strong recommendation notice instead
- Remove ?panel query when closing add on panels in project addons (useEffect to re-open the panels were getting triggered randomly)
- Chucked a fix in here to fix SSL button not working in settings/database


apps/studio/components/interfaces/Settings/Addons/ComputeInstanceSidePanel.tsx
```diff
@@ -73,7 +73,10 @@ const ComputeInstanceSidePanel = () => {
 
   const snap = useSubscriptionPageStateSnapshot()
   const visible = snap.panelKey === 'computeInstance'
-  const onClose = () => snap.setPanelKey(undefined)
+  const onClose = () => {
+    router.push(router.asPath.split('?')[0], undefined, { shallow: true })
```
ivasilov, line 77: Did you try using `router.pathname`? I think it doesn't include query params.

MildTomato:
```js
router.push(`/projects/${router.query.ref}/`)
```

MildTomato: use `asPath`, includes the params.
https://nextjs.org/docs/pages/api-reference/functions/use-router#router-object

MildTomato (approved):

PR #19329, opened 2023-11-30
Customer Story: Shotgun

RamiroND:


apps/www/_customers/shotgun
```diff
@@ -0,0 +1,80 @@
+---
+name: Shotgun
+title: 'Supabase migration drives shotgun to an 83% reduction in data infrastructure costs'
```
kiwicopple, line 3:
```suggestion
title: 'Supabase migration drives an 83% reduction in data infrastructure costs'
```

if you don't want to remove, it should at least be capatilized

apps/www/_customers/shotgun
```diff
@@ -0,0 +1,80 @@
...
+# Use meta_title to add a custom meta title. Otherwise it defaults to '{name} | Supabase Customer Stories':
+meta_title: 'Supabase migration drives shotgun to an 83% reduction in data infrastructure costs'
```
kiwicopple, line 5:
```suggestion
meta_title: 'Supabase migration delivers an 83% reduction in data infrastructure costs for Shotgun'
```

fsansalvadore (approved):

PR #19332, opened 2023-11-30
Chore/remove react query

saltcod:


MildTomato:
I'm gonna move the thread fetch stuff out of actions since its not really an action

MildTomato (approved):

Issue #17981, opened 2023-10-06
Application error: a client-side exception has occurred - happens whenever csv data is uploaded into a table and another table is selected 


i-pip:
# Bug report

- [ x] I confirm this is a bug with Supabase, not with my own application. 
- [ x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Uploading CSV data into any table and navigation to a new table in the dashboard results in a Client-side exception. The exception is a result of two React DnD backends being present at the same time. 

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to the Supabase Dashboard
2. Click on any empty table
3. upload csv data
4. Click on any other table
5. See error

## Screenshots

![Supabase App Error](https://github.com/supabase/supabase/assets/15890505/3280e757-cc4d-4e0c-b0fc-07ccdd70a1e1)

## System information

- OS: macOS
- Browser (if applies) chrome, safari, edge

jha-adrs:
I tried to replicate it in Chrome, in Windows, but could not reproduce the issue, probably due to Safari

botxboom:
Hi @i-pip , I have raised a PR https://github.com/supabase/supabase/pull/19340 for this issue.

PR #19340, opened 2023-11-30
fix: drag and drop causing page crash

botxboom:

Issue: #17981 
This bug is related to drag and drop feature. When user tries to load data into the table from csv file by dragging and switches table, the page crashes.
This is a bug fix.

Currently when user tries to load data into the table by using drag and drop. after loading the data, If user clicks on another table. The page crashes.

https://github.com/supabase/supabase/assets/28553853/4e93fdae-eca7-457c-9b27-00b2a0b04638

## What is the new behavior?

The new fix will remove the error causing page crash. 
This bug is resolved by providing the individual context to DnDProvider componenets.

https://github.com/supabase/supabase/assets/28553853/e12156d1-9065-40be-b936-f411fc0693bc

## Additional context


alaister (approved):
Hey @botxboom,

Thanks for this PR! I can confirm that it solves the issue.

I'm just wondering why you've chosen to use `window` as the context as opposed to something else?

botxboom:
> Hey @botxboom,
> 
> Thanks for this PR! I can confirm that it solves the issue.
> 
> I'm just wondering why you've chosen to use `window` as the context as opposed to something else?

Hi @alaister , 
I was checking the `DnDProvider` from `react-dnd`. I believe it uses global (singleton) context. As we have used it in two of our components, we need to provide/specify context for each provider or else it will pick the global context and cause `cannot have two HTML5Backend` error.

ref: react-dnd (when no context is given).
`function getGlobalContext() {
    return typeof global !== 'undefined' ? global : window;
}`

PR #19374, opened 2023-12-01
fix layout and layout shift

saltcod:
fix issue with the <main> tag

screenshot-2024-02-19-at-16 30 53


apps/database-new/components/MainTagWrapper.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+const MainTagWrapper = ({ children }: MainTagWrapperProps) => {
+  const pathname = usePathname()
+  const isFullHeight = pathname.startsWith('/thread') || pathname === '/'
+
+  return (
+    <main
+      role="main"
+      className={cn('w-full flex flex-col grow', isFullHeight && 'h-[calc(100vh-115px)]')}
```
MildTomato, line 16: > 115px

magic number. we don't need to do this do we ?

MildTomato: Can just have a outer container of h-screen, and then child elements just fill the space

PR #19380, opened 2023-12-01
db design now saving messages to supabase db

MildTomato:

## What is the new behavior?

## Additional context


MildTomato:
need to get migration from remote db project

apps/database-new/app/actions.ts
```diff
@@ -168,6 +168,20 @@ export async function createThread(prevState: any, formData: FormData) {
       console.error(error)
     }
 
+    // insert into supabase
+    try {
+      const { error } = await supabase.from('messages_user').insert({
```
saltcod, line 173: is there a migration file for this?
I've got the rls policies migration that I just remembered need to be run as well

can we add local ones rather than generate one from remote project?

MildTomato: > can we add local ones rather than generate one from remote project?

not sure what you mean - it's the same thing.
I'll db pull all the changes and you'll still be able to run locally.

saltcod: yeah it's fine. it just pulls in a lot of cruft into that migration. Fine though

MildTomato: I have an idea on how to do forking when I was doing some of this.
It started to make more sense after I started saving things to supabase

saltcod (approved):

PR #19398, opened 2023-12-03
Update with-svelte.mdx

shinriyo:

docs fix.

You can't go to the login page.

## What is the new behavior?

You can get to the login page.

## Additional context

It does not use an FQDN for the redirect.
It should be http://localhost:5173/

FQDN does mean "Fully" Qualified Domain Name.

I wested my time on the bug document. So, I fix.


charislam (approved):

shinriyo:
Thank you for your merge!

PR #19404, opened 2023-12-04
Chore/rls ai assistant updates 041223

joshenlim:
- Send policy definition if editing a policy to ai assistant when editing
- Only send entity definitions if opted into ai sql

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/AIPolicyEditorPanel.utils.ts
+++ b/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/AIPolicyEditorPanel.utils.ts
@@ -74,3 +74,13 @@ COMMIT;
 `.trim()
   }
 }
+
+export const generatePolicyDefinition = (policy: PostgresPolicy) => {
+  return `
+CREATE POLICY "${policy.name}" on "${policy.schema}"."${policy.table}"
+AS ${policy.action} FOR ${policy.command}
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/AIPolicyEditorPanel.utils.ts
```diff
@@ -74,3 +74,13 @@ COMMIT;
 `.trim()
   }
 }
+
+export const generatePolicyDefinition = (policy: PostgresPolicy) => {
+  return `
+CREATE POLICY "${policy.name}" on "${policy.schema}"."${policy.table}"
+AS ${policy.action} FOR ${policy.command}
+TO ${policy.roles.join(', ')}
+USING (${policy.definition})${policy.check === null ? ';' : ''}
```
alaister, line 83: Do policies always have a using definition? For example `INSERT` expressions only have `WITH CHECK` and no `USING`

joshenlim: ahhh good point - lemme update

joshenlim:
Also added the collapsible policy details in here - pardon the extra hacky height rendering, i can't find a clean solution for this at the moment unless anyone might have some ideas

alaister (approved):
LGTM!

PR #19405, opened 2023-12-04
feat: explicit gotrue-js package dependency, version 2.58.0

hf:
gotrue-js was removed from being an explicit dependency. This is useful for us to independently bump the version. This also bumps the version to 2.58.0 which fixes a bug with `SIGNED_OUT` not being called.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9998,8 +9998,9 @@
       }
     },
     "node_modules/@supabase/gotrue-js": {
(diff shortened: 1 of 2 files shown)


alaister (approved):
LGTM!

Is there any risk of the wrong gotrue-js version getting installed if a different supabase-js version is installed in future without gotrue-js also being bumped?

PR #19411, opened 2023-12-04
docs: add single session per user docs

hf:
Adds the single session per user docs to the User Sessions page.

### DIFF
--- a/apps/docs/pages/guides/auth/sessions.mdx
+++ b/apps/docs/pages/guides/auth/sessions.mdx
@@ -74,7 +74,7 @@ To make sure that users are required to re-authenticate periodically, you can se
 
 To make sure that sessions expire after a period of inactivity, you can set a positive duration for the _Inactivity timeout_ option in the [Auth settings](/dashboard/project/_/settings/auth).
 
-You can also configure to have only one active session per user per device or browser. When this is enabled, the session from the most recent sign in will remain active, while the rest are terminated. Enable this via the _Single session per user_ option in the [Auth settings](/dashboard/project/_/settings/auth).
+You can also enforce only one active session per user per device or browser. When this is enabled, the session from the most recent sign in will remain active, while the rest are terminated. Enable this via the _Single session per user_ option in the [Auth settings](/dashboard/project/_/settings/auth).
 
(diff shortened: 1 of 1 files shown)


kangmingtay (approved):

J0 (approved):

apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -26,6 +26,7 @@ A session terminates, depending on configuration, when:
 - The user changes their password or performs a security sensitive action.
 - It times out due to inactivity.
 - It reaches its maximum lifetime.
+- A user signed in on another device.
```
charislam, line 29:
```suggestion
- A user signs in on another device.
```

I'd suggest putting the condition here (`when "single user per session" is configured` or something like that), as it's pretty easy to misread this as the default behavior, which might not be what many people want. (Might apply to the other items as well)

apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -52,7 +53,7 @@ Upon sign out, all refresh tokens and potentially other database objects related
 
 Be aware that access tokens of revoked sessions remain valid until their expiry time, encoded in the `exp` claim.
 
-## Controlling the lifetime of a session
+## Controlling the lifetime and number of sessions a user can have
```
charislam, line 56:
```suggestion
## Limiting session lifetime and number of allowed sessions per user
```

apps/docs/pages/guides/auth/sessions.mdx
```diff
@@ -73,6 +74,8 @@ To make sure that users are required to re-authenticate periodically, you can se
 
 To make sure that sessions expire after a period of inactivity, you can set a positive duration for the _Inactivity timeout_ option in the [Auth settings](/dashboard/project/_/settings/auth).
 
+You can also configure to have only one active session per user per device or browser. When this is enabled, the session from the most recent sign in will remain active, while the rest are terminated. Enable this via the _Single session per user_ option in the [Auth settings](/dashboard/project/_/settings/auth).
```
charislam, line 77:
```suggestion
You can also enforce only one active session per user per device or browser. When this is enabled, the session from the most recent sign in will remain active, while the rest are terminated. Enable this via the _Single session per user_ option in the [Auth settings](/dashboard/project/_/settings/auth).
```

charislam (approved):
minor suggestions to make this a bit clearer

apps/docs/pages/guides/auth/sessions.mdx
hf: It has a sentence above it that says "depending on configuration"

PR #19421, opened 2023-12-04
fix: implement force mounting for cmdk menu items

charislam:

**Bug fix:** We want some command menu items to be force-mounted (that is, they should be mounted regardless of what the current search value is. Examples include all search results from docs, and docs search and AI responses when within the docs site.

Force mounting isn't implemented in a published release of `cmdk` (https://github.com/pacocoursey/cmdk/issues/164), but we can hack it in the interim by manipulating the item value under the hood.

When you try to search for a string with whitespace, the command scoring algorithm removes `Docs search` from the suggested commands. `AI` persists due to some bizarre interaction between the `cmdk` value derivation and the `Experimental badge`.

image

## What is the new behavior?

The `Docs search` command item is always mounted.

image

## Additional context


charislam:
closing and reopening to try to fix preview branch error

MildTomato (approved):
not QA'd any previews.
looks good though

PR #19430, opened 2023-12-05
Chore/realtime inspector fixes 051223

joshenlim:
- Fix broken URLs in empty state for realtime inspector
- Limit realtime inspector filter incoming db changes message to just the public schema
  - This is because any other schemas won't work out of the box due to missing permissions
- Hide copy to clipboard button for realtime message panel timestamp
- Add placeholder value for sending test message
- Fix sending test message getting sent as string instead of JSON
- Change channel name input to a form
- Add join channel empty state UI
- Update misc copy writing based on feedback
- Shift toggles for broadcast, presence and db changes to filter message popover, instead of within the channel name popover

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/AIPolicyChat.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/AIPolicyChat.tsx
@@ -2,7 +2,7 @@ import { zodResolver } from '@hookform/resolvers/zod'
 import { compact, last, sortBy } from 'lodash'
 import { Loader2 } from 'lucide-react'
 import OpenAI from 'openai'
-import { useEffect, useMemo, useRef, useState } from 'react'
+import { useEffect, useMemo, useRef } from 'react'
 import { useForm } from 'react-hook-form'
 import {
   AiIcon,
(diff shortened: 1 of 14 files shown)


ivasilov:
Few minor UI issues:
1. I can't just listen to postgres changes, i have to join a channel.
Screenshot 2023-12-05 at 11 18 28

2. The intro page with the div that looks like input is misleading, I suggest we change it. It mislead both me and Alaister (100% of people who saw the PR so far :D ) 
Screenshot 2023-12-05 at 11 14 44

3. There's some issue with the switches, turning one off turns the other on.
https://github.com/supabase/supabase/assets/568291/46416f12-383d-4944-bbe0-8c4a859681ca

4. When clicking `join channel`, should it automatically start listening? Otherwise, there's two clicks and the `join channel` is not needed IMO. Maybe remove the `join channel` button since it does nothing. If you want to change the channel while listening, you leave the channel (and stop listening), change the channel and start listening again.
5. It should be `joined channel` instead of `joining channel`, right?
Screenshot 2023-12-05 at 11 25 28
6. Different wordings for the same thing
Screenshot 2023-12-05 at 11 30 35
Screenshot 2023-12-05 at 11 30 38

joshenlim:
> 2. The intro page with the div that looks like input is misleading

Yeah good point, i'm on the fence about using "UI screenshots" as images within the dashboard given the confusion it may cause, will opt to remove it for now

> 3. There's some issue with the switches, turning one off turns the other on.

Good catch!

> 4. When clicking join channel, should it automatically start listening?

That's true - feels unnecessary that extra click

> 5. It should be joined channel instead of joining channel, right?

Yup! Good catch

> 6. Different wordings for the same thing

Good catch!

ivasilov (approved):

PR #19449, opened 2023-12-05
chore: adjust logic for new project screen

kevcodez:
While there are surely ways to avoid this whole usage vs no-usage situation, we need to make sure that customers launching new projects are absolutely guaranteed to see the new project screen with instructions first. We had people reach out that it was difficult to find the project keys etc - because they had 2 requests (maybe through dashboard) and we showed them the usage sections rather than the onboarding.

PR ensures we show customers the onboarding on the project dashboard if
* Project is no older than 48h
* Project has <25 requests in last week


alaister (approved):
LGTM!

PR #19510, opened 2023-12-07
chore: move role impersonation state into a context

alaister:

chore/Bug fix

Role impersonation state is not reset between projects.

## What is the new behaviour?

Each project now has its own distinct state.

## Additional context

Due to Next.js `getLayout` running all the time (😡), the role impersonation state also gets reset between page transitions, meaning it will no longer be persisted _within_ a project. This may not be a bad thing, though, as it may help prevent confusion.

Using Valtio with Context: https://valtio.pmnd.rs/docs/how-tos/how-to-use-with-context


apps/studio/state/role-impersonation-state.tsx
```diff
@@ -0,0 +1,52 @@
+import { PropsWithChildren, createContext, useCallback, useContext } from 'react'
```
joshenlim, line 1: can be a ts file right? no need for tsx?

alaister: It's got JSX in it so it does need to be .tsx unfortunately

joshenlim (approved):

PR #19551, opened 2023-12-08
fix: make consistent recommendations about filters

### LINKED ISSUE #19506: Inconsistent recommendations around adding filters and RLS
# Improve documentation

## Link 
https://supabase.com/docs/guides/auth/managing-user-data#private-access
Here it says
>The nice thing about this pattern? We can now query this table via the API and we don't need to include data filters in our API queries - the Policies will handle that for us:

https://supabase.com/docs/guides/database/postgres/row-level-security#add-filters-to-every-query
While here you recommend
>Policies are "implicit where clauses", so it's common to run select statements without any filters. This is a bad pattern for performance. Instead of doing this (JS client example):

## Describe the problem
The docs should be consistent with proposed best practices and not encourage new users to use RLS policies as data filters.

## Describe the improvement
Don't suggest we don't need to include data filters when we have RLS policies.

charislam:
The docs have inconsistent recommendations about combining filters with RLS. Add a note to make this consistent across pages: filters aren't needed for restricting user access to rows with RLS enabled, but they might help with performance.

Resolves #19506

### DIFF
--- a/apps/docs/pages/guides/auth/managing-user-data.mdx
+++ b/apps/docs/pages/guides/auth/managing-user-data.mdx
@@ -129,7 +129,7 @@ val data = supabase.postgrest["profiles"].select(Columns.list("id", "username", 
 
 <Admonition type="tip">
 
-Security is handled at the database level by the RLS policy, so you don't _need_ a filter on `id` to protect user data. Depending on your table size and query, you might still want to add a filter for performance. Postgres can use the filter to construct a more efficient query.
+Security is handled at the database level by the RLS policy. The policy restricts the returned rows, so you don't _need_ a filter on `id`. Depending on your table size and query, you might still want to add a filter for performance. Postgres can use the filter to construct a more efficient query.
 
 </Admonition>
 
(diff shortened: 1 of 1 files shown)


kangmingtay (approved):

apps/docs/pages/guides/auth/managing-user-data.mdx
```diff
@@ -127,6 +127,12 @@ val data = supabase.postgrest["profiles"].select(Columns.list("id", "username",
 </TabPanel>
 </Tabs>
 
+<Admonition type="tip">
+
+Security is handled at the database level by the RLS policy, so you don't _need_ a filter on `id` to protect user data. Depending on your table size and query, you might still want to add a filter for performance. Postgres can use the filter to construct a more efficient query.
```
charislam, line 132:
```suggestion
Security is handled at the database level by the RLS policy. The policy restricts the returned rows, so you don't _need_ a filter on `id`. Depending on your table size and query, you might still want to add a filter for performance. Postgres can use the filter to construct a more efficient query.
```

PR #19580, opened 2023-12-11
Chore/column editor show types from extensions schema

joshenlim:
Show postgres enumerated types from the `extensions` schema on the table and column side panel editor (for the Table Editor and Database pages)

Database extensions that are installed through the dashboard on the database/extensions page are, most of the time, installed by default in the extensions schema (it's the default dropdown option) unless the extension has a schema that it's required to be in, or the user changes it to be installed in another schema. 

If the installed extension (e.g vector) has enumerated types, the Table Editor then can access those types for users to assign them to columns, without having the user to install them in another schema.


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.tsx
```diff
@@ -60,8 +61,13 @@ const TableEditor = ({
...
 
   const realtimeEnabled = useIsFeatureEnabled('realtime:all')
-
-  const enumTypes = meta.types.list((type: PostgresType) => !EXCLUDED_SCHEMAS.includes(type.schema))
+  const { data: types } = usePostgresTypesQuery({
+    projectRef: project?.ref,
+    connectionString: project?.connectionString,
+  })
+  const enumTypes = (types ?? []).filter(
+    (type) => !EXCLUDED_SCHEMAS.filter((x) => x !== 'extensions').includes(type.schema)
```
alaister, line 69: small perf nit: could move the `EXCLUDED_SCHEMAS.filter((x) => x !== 'extensions')` part out of this `.filter()` into its own const (maybe something like `EXCLUDED_SCHEMAS_WITHOUT_EXTENSIONS`) above the component. That way it we can avoid some nested loops :)

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.tsx
```diff
@@ -44,6 +44,8 @@ import {
 import ColumnForeignKey from './ColumnForeignKey'
 import ColumnType from './ColumnType'
 import HeaderTitle from './HeaderTitle'
+import { toJS } from 'mobx'
```
alaister, line 47: In use?

alaister (commented):
🔥

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.tsx
joshenlim: gotcha!

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.tsx
joshenlim: nope! good catch 🙏

alaister (approved):
LGTM!

PR #19657, opened 2023-12-13
Update denyUrls for sentry client config

joshenlim:


kevcodez (approved):
It looks good, maybe @kamilogorek can have another look too

alaister (approved):

kamilogorek:
LGTM, although I'd not reuse Sentry's list. It has to go through every regexp match for every captured exception now. I'd rather keep our list targeted with actual reference to why it's configured.

joshenlim:
ahh okay - lemme just remove all those that i added and just keep the coinbase one then 🙏

PR #19672, opened 2023-12-13
feat: display api urls and keys instead of placeholders in docs

charislam:
**feat:** adds a component to display API endpoints and anon keys in docs if user is authenticated
makes the getting started UX smoother so users don't have to jump back and forth between docs and trying to find their keys in dashboard

Not logged in:
image

Logged in:
image

**testing:** in order for sign in to work, everything has to be proxied via the www preview site.

sign in here: https://zone-www-dot-com-git-feat-docs-api-keys-supabase.vercel.app/dashboard
check docs here: https://zone-www-dot-com-git-feat-docs-api-keys-supabase.vercel.app/docs

**out of scope:** it would be very cool to display these directly in the code blocks. I looked into it and it would require quite a bit of work forking and modifying `code-hike`, so out of scope for now


apps/docs/components/Navigation/NavigationMenu/TopNavBar.tsx
```diff
@@ -23,37 +23,6 @@ const TopNavBar: FC = () => {
   const libraryMeta = REFERENCES?.[library] ?? undefined
   const versions = libraryMeta?.versions ?? []
 
-  const version = versions.includes(pathSegments[pathSegments.indexOf(library) + 1])
```
charislam, line 26: obsolete, unused variables

apps/docs/components/Navigation/NavigationMenu/TopNavBar.tsx
```diff
@@ -118,6 +87,11 @@ const TopNavBar: FC = () => {
               </a>
             </Button>
           )}
+          {process.env.NEXT_PUBLIC_DEV_AUTH_PAGE === 'true' && (
```
charislam, line 90: I set this on Vercel Development/Preview. Add to local dev if you need to sign in locally

kangmingtay: you can probably create a feature flag on config cat to toggle this behaviour to only appear on local dev / staging for the team

apps/docs/lib/userAuth.ts
```diff
@@ -0,0 +1,58 @@
...
+let currentSession: Session | null
+
+auth.onAuthStateChange((event, session) => {
+  currentSession = session
+
+  if (event === 'SIGNED_OUT') {
+    Object.keys(LOCAL_STORAGE_KEYS).forEach((key) => {
+      remove('local', LOCAL_STORAGE_KEYS[key])
+    })
+  }
```
kangmingtay, line 17: i think the client library automatically clears the session from local storage on sign out

reference: https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L1493

apps/docs/components/Navigation/NavigationMenu/TopNavBar.tsx
charislam: but does config cat know who you are?

the problem is that you're anonymous on docs staging, so how would you get authed so configcat can identify you?

apps/docs/pages/__dev-secret-auth.tsx
```diff
@@ -0,0 +1,63 @@
...
+          same domain.
+        </p>
+        <div className="flex flex-col gap-2 max-w-sm">
+          <Button_Shadcn_ onClick={() => signIn({ provider: 'github' })}>
+            {' '}
+            Sign in with GitHub
+          </Button_Shadcn_>
+          <Button_Shadcn_ onClick={() => signIn({ provider: 'google' })}>
+            Sign in with Google
+          </Button_Shadcn_>
```
kangmingtay, line 57: we don't offer google as a login option so this button should be removed

apps/docs/lib/userAuth.ts
charislam: good to know! 👍🏼 I still need to clear out the data that the frontend saves though

apps/docs/components/Navigation/NavigationMenu/TopNavBar.tsx
charislam, line 90: Needed to test auth on local dev. The proxy magic that makes signin via dashboard work on prod and (with some manual trickery) on preview doesn't work on local dev.

I tried proxying everything through www served on localhost and ended up with some mad infinite reloading from Next.js Fast Refresh. Couldn't for the life of me figure out why after banging my head against it for hours

apps/docs/components/ProjectConfigVariables.tsx
```diff
@@ -0,0 +1,455 @@
...
+            <ChevronsUpDown className="ml-2 h-4 w-4 shrink-0 opacity-50" />
+          </Button>
+        </PopoverTrigger>
+      </div>
+      <PopoverContent className="z-[999999] p-0" side="bottom">
+        <Command>
+          <CommandInput placeholder="Search project..." className="border-none ring-0" />
+          <CommandEmpty>No project found.</CommandEmpty>
+          <CommandGroup>
+            {instances.map((instance) => (
```
joshenlim, line 186: We probably need to add a scrollarea here
image

can refer to https://github.com/supabase/supabase/blob/1fb755bfc715446de9600ce84084f1b3290b2829/apps/studio/components/layouts/AppLayout/ProjectDropdown.tsx#L138

apps/docs/components/ProjectConfigVariables.tsx
```diff
@@ -0,0 +1,455 @@
...
+  parentStateSummary,
+  instances,
+  className,
+}: {
+  parentStateSummary: ProjectConfigVariablesState
+  instances: InstanceData[]
+  className?: string
+}) {
+  const { selectedInstanceKey, setSelectedInstanceKey } = useSnapshot(instanceStore)
+  const selectedInstance = getInstanceDataFromKey(instances, selectedInstanceKey)
```
joshenlim, line 160: it does look like we're showing branch instances here (I'm guessing those with 3 breadcrumbs) - is that what we want?

image

apps/docs/components/ProjectConfigVariables.tsx
```diff
@@ -0,0 +1,455 @@
...
+                onSelect={(selectedValue: string) => {
+                  const newSelectedInstance = fromPrettyString(instances, selectedValue)
+                  setSelectedInstanceKey(toInstanceKey(newSelectedInstance))
+                  setOpen(false)
+                }}
+              >
+                <Check
+                  className={cn(
+                    'mr-2 h-4 w-4',
+                    instance === selectedInstance ? 'opacity-100' : 'opacity-0'
```
joshenlim, line 198: selected state probably should be based on project ref to account for identical project names?

image

apps/docs/components/ProjectConfigVariables.tsx
joshenlim: yeah thinking twice - this makes sense 👍

joshenlim (approved):

PR #19708, opened 2023-12-13
fix: update basic auth settings form

kangmingtay:
* Nothing too fancy, just changes the radio buttons to form fields in the basic auth config page so that it is consistent with the rest of the auth configs
* Tested locally that the configs are propagated to the db correctly

image

image

## What is the new behavior?
image

image

## Additional context
* Would be nice if we could add icons for hcaptcha and turnstile :)


alaister (approved):
LGTM!

PR #19716, opened 2023-12-14
fix: time until hackathon end count down clock

dshukertjr:

Currently, the count down on the launch week landing page is set to count down the time until the LWX ends, but the hackathon actually goes for another two days as it says so in the hackathon blog post:

>The submission deadline is 11:59 pm Sunday midnight PT 17th December 2023

Screenshot 2023-12-14 at 14 49 33

This PR fixes the count down clock.


fsansalvadore (approved):
Good call!

PR #19728, opened 2023-12-14
chore: change current to all local for migrations in db reset command

tobiastornros:

Updates docs for db reset command in cli. Make it clear that all local migrations will be applied. At least for me "current" isn't that clear that all migrations available locally can be applied.

### DIFF
--- a/spec/cli_v1_commands.yaml
+++ b/spec/cli_v1_commands.yaml
@@ -2700,11 +2700,11 @@ commands:
         default_value: ''
       - id: linked
         name: --linked
-        description: Resets the linked project to all local migrations.
+        description: Resets the linked project with local migrations.
         default_value: 'false'
       - id: local
         name: --local
-        description: Resets the local database to all local migrations.
+        description: Resets the local database with local migrations.
         default_value: 'true'
       - id: version
         name: --version <string>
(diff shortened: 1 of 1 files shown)


spec/cli_v1_commands.yaml
```diff
@@ -2700,11 +2700,11 @@ commands:
         default_value: ''
       - id: linked
         name: --linked
-        description: Resets the linked project to current migrations.
+        description: Resets the linked project to all local migrations.
```
kiwicopple, line 2703:
```suggestion
        description: Resets the linked project with local migrations.
```

spec/cli_v1_commands.yaml
```diff
@@ -2700,11 +2700,11 @@ commands:
...
         default_value: 'false'
       - id: local
         name: --local
-        description: Resets the local database to current migrations.
+        description: Resets the local database to all local migrations.
```
kiwicopple, line 2707:
```suggestion
        description: Resets the local database with local migrations.
```

kiwicopple (approved):

spec/cli_v1_commands.yaml
tobiastornros: Awesoooome!

PR #19732, opened 2023-12-14
Include Next.js middleware matcher

### LINKED ISSUE #18285: Improve Supabase Auth with the Next.js App Router Docs
# Improve documentation

## Link 

Check out: [Supabase Auth with the Next.js App Router](https://supabase.com/docs/guides/auth/auth-helpers/nextjs)

## Describe the problem
So, after using Supabase in production, some pages felt like they took ages to load. And after some digging it turns out, that the middleware was being called more than necessary. 

Take a product list page, for example. Next.js pre-fetches all the links, this causes the middleware to be called multiple times, for no reason, sometimes leading to a 400 error.

I did a test on the "with-supabase" template and here's what I found:

| Before fix | After fix |
| image  | image  |
| Middleware is called 9 times per page load.  |  The middleware is called once per. page load   |

I referred to the [Next.js Middleware Docs](https://nextjs.org/docs/app/building-your-application/routing/middleware) and found that we can use a matcher to decide when the middleware runs. And i feel like this should be part of the Supabase guide, since It substantially enhances performance, and there's no logical reason to invoke the middleware for resources like favicons, images, and CSS files.

## Describe the improvement
```ts filename="middleware.ts"
import { createMiddlewareClient } from "@supabase/auth-helpers-nextjs";
import { NextResponse } from "next/server";

import type { NextRequest } from "next/server";

export async function middleware(req: NextRequest) {
  const res = NextResponse.next();

  // Create a Supabase client configured to use cookies
  const supabase = createMiddlewareClient({ req, res });

  // Refresh session if expired - required for Server Components
  // https://supabase.com/docs/guides/auth/auth-helpers/nextjs#managing-session-with-middleware
  await supabase.auth.getSession();

  return res;
}

// Ensure the middleware is only called for relevant paths.
export const config = {
  matcher: [
    /*
     * Match all request paths except for the ones starting with:
     * - _next/static (static files)
     * - _next/image (image optimization files)
     * - favicon.ico (favicon file)
     */
    '/((?!_next/static|_next/image|favicon.ico).*)',
  ],
}

```

I hope this helps others as much as it helped me :v:

## Additional context

No

jleguina:

Modify docs to add a matcher to the Next.js middleware to avoid unnecessary renders.

Fixes #18285 

## What is the new behavior?

N/A


apps/docs/pages/guides/auth/auth-helpers/nextjs.mdx
```diff
@@ -52,7 +52,7 @@ When using the Supabase client on the server, you must perform extra steps to en
 
 Next.js Server Components allow you to read a cookie but not write back to it. Middleware on the other hand allow you to both read and write to cookies.
 
-Next.js [Middleware](https://nextjs.org/docs/app/building-your-application/routing/middleware) runs immediately before each route is rendered. We'll use Middleware to refresh the user's session before loading Server Component routes.
+Next.js [Middleware](https://nextjs.org/docs/app/building-your-application/routing/middleware) runs immediately before each route is rendered. To avoid unnecessary renders, we include a matcher to decide when the middleware should run. You can read more on matching paths in the Next.js [documentation](https://nextjs.org/docs/app/building-your-application/routing/middleware#matching-paths). We'll use Middleware to refresh the user's session before loading Server Component routes.
```
thorwebdev, line 55:
```suggestion
Next.js [Middleware](https://nextjs.org/docs/app/building-your-application/routing/middleware) runs immediately before each route is rendered. To avoid unnecessary execution, we include a matcher config to decide when the middleware should run. You can read more on matching paths in the Next.js [documentation](https://nextjs.org/docs/app/building-your-application/routing/middleware#matching-paths). We'll use Middleware to refresh the user's session before loading Server Component routes.
```

thorwebdev (approved):
LGTM, thanks 💚

PR #19743, opened 2023-12-14
chore: Migrate Connection Pooling form to use `react-hook-form`

ivasilov:


alaister (commented):
Looks good! After saving the dirty form state seems to remain though:
Screenshot 2023-12-15 at 14 46 00
(save button doesn't disable again)

ivasilov:
> Looks good! After saving the dirty form state seems to remain though: Screenshot 2023-12-15 at 14 46 00 (save button doesn't disable again)

Good catch, fixed.

saltcod (approved):

PR #19752, opened 2023-12-14
Add documentation for Swift v2 and update user management example to use Swift v2

grdsdev:

docs update

## What is the new behavior?

- Create a new doc spec for Swift v2.
- Update swift-user-management example for using the new Swift v2 library.


thorwebdev (approved):
LGTM, thanks 💚

PR #19813, opened 2023-12-16
fixed typo

JennaHuff:
Fixed typo in string "Give users access to only their own a top level folder named as uid"


alaister (approved):
Thanks, @JennaHuff!

PR #19819, opened 2023-12-17
[#19341] Bug Fix - Renaming Table does not rename primary key constraint

### LINKED ISSUE #19341: Unable to update table PK via dashboard after renaming table
# Bug report

## Describe the bug

Opening this issue because I ran into it again today #5549

After renaming a table, the _pkey constraint uses the old table name. This causes a problem if you try to change the primary key field.

For example, a table originally named "reporttype_report" was renamed to "report_reporttype".

When trying to change the primary key, It gives an error "constraint "report_reporttype_pkey" of relation "report_reporttype" does not exist" because the constraint is still named "reporttype_report_pkey"

## To Reproduce

1. create a table
2. rename it in the admin console
3. add a new column
4. try to add the new column to the primary key

## Expected behavior
Should succeed. Instead you have to go find the constraint name and manually ALTER the table.

## System information
OS: macOS
Browser: Chrome

Prakhar-FF13:

fixes #19341 

Renaming a table does not rename the primary key which causes issues like changing primary key to not work.

## What is the new behavior?

The issue above has been fixed. Renaming the table and changing the primary key now works.

## Additional context

None

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
@@ -79,7 +79,7 @@ const SidePanelEditor = ({
   })
   const { mutateAsync: createPublication } = useDatabasePublicationCreateMutation()
   const { mutateAsync: updatePublication } = useDatabasePublicationUpdateMutation({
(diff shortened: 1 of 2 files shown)


saltcod:
fixed in https://github.com/supabase/supabase/pull/21285

PR #19833, opened 2023-12-18
fix: Remove sorting if the column doesn't exist

ivasilov:
This PR fixes an issue where if you sort by a column, share the link and then delete the column, the link no longer works. 

If you delete the column while sorted, it automatically removes the sorting so this bug only happens with a shared or bookmarked link.


ivasilov:
Closing this PR in favor of https://github.com/supabase/supabase/pull/19178

PR #19841, opened 2023-12-18
Feat/sortby on tables

saltcod:
Add option to set custom sort property on tables

### DIFF
--- a/apps/studio/data/tables/tables-query.ts
+++ b/apps/studio/data/tables/tables-query.ts
@@ -11,12 +11,13 @@ export type TablesVariables = {
   projectRef?: string
   connectionString?: string
   schema?: string
+  sortByProperty?: string
 }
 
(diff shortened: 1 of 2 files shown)


apps/studio/data/tables/tables-query.ts
```diff
@@ -26,23 +24,34 @@ export async function getTables(
...
+    params: {
+      header: {
+        'x-connection-encrypted': connectionString!,
+      },
+      path: {
+        ref: projectRef,
+      },
+      query: queryParams as any,
+    },
+    headers: Object.fromEntries(headers),
```
joshenlim, line 42: probably no longer need Object.fromEntries i think
can just be `headers: headers`, or shorthand just `headers`

joshenlim (approved):
lgtm! just tiny comment

PR #19882, opened 2023-12-19
fix cmdk links

charislam:
fix docs links in cmdk still being broken in dashboard

### DIFF
--- a/packages/ui/src/components/Command/DocsSearch.tsx
+++ b/packages/ui/src/components/Command/DocsSearch.tsx
@@ -227,8 +227,13 @@ const DocsSearch = () => {
     switch (pageType) {
       case PageType.Markdown:
       case PageType.Reference:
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Command/DocsSearch.tsx
```diff
@@ -227,8 +227,13 @@ const DocsSearch = () => {
     switch (pageType) {
       case PageType.Markdown:
       case PageType.Reference:
-        const docsLink = site === 'docs' ? link : `https://supabase.com/docs${link}`
-        return router.push(docsLink)
+        if (site === 'docs') {
+          return router.push(link)
+        } else if (site === 'website') {
+          return router.push(`/docs${link}`)
```
gregnr, line 233: Curious why this logic doesn't work for `studio`?

charislam: yep, exactly!

saltcod (approved):

PR #19892, opened 2023-12-20
chore: use data/fetchers for custom domains

alaister:

chore

Seeing this annoying error in the console constantly.
Screenshot 2023-12-20 at 14 49 06

## What is the new behaviour?

Custom domains fetches using the `data/fetchers` methods because it needed updating, and the above error is fixed. Also some wording has been improved.

### DIFF
--- a/apps/studio/data/custom-domains/custom-domains-query.ts
+++ b/apps/studio/data/custom-domains/custom-domains-query.ts
@@ -2,6 +2,7 @@ import { useQuery, UseQueryOptions } from '@tanstack/react-query'
 
 import { get } from 'data/fetchers'
 import { IS_PLATFORM } from 'lib/constants'
+import { ResponseError } from 'types'
 import { customDomainKeys } from './keys'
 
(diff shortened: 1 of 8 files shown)


apps/studio/data/custom-domains/custom-domains-query.ts
```diff
@@ -105,9 +92,9 @@ export async function getCustomDomains(
       } as const
     }
 
-    const isNoHostnameConfiguredError =
-      (response.error as any)?.code === 400 &&
-      (response.error as any)?.message?.includes('custom hostname configuration')
+    const isNoHostnameConfiguredError = (error as any)?.message?.includes(
```
joshenlim, line 95: probably type CustomDomainsError as ResponseError on line 113, then won't need the `any` here

alaister: Good shout!

joshenlim (approved):

PR #19902, opened 2023-12-20
ref-docs: update `supabase.auth.onAuthStateChange()` docs

hf:
Adds more context about the `supabase.auth.onAuthStateChange()` method, especially around events and best practices.

Moves it high up, as this should be done early on by developers.

### DIFF
--- a/spec/supabase_js_v2.yml
+++ b/spec/supabase_js_v2.yml
@@ -192,10 +192,9 @@ functions:
       - The auth methods can be accessed via the `supabase.auth` namespace.
       - By default, the supabase client sets `persistSession` to true and attempts to store the session in local storage. When using the supabase client in an environment that doesn't support local storage, you might notice the following warning message being logged:
 
-      > No storage option exists to persist the session, which may result in unexpected behavior when using auth.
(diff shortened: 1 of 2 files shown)


spec/supabase_js_v2.yml
```diff
@@ -284,6 +285,203 @@ functions:
             }
           )
           ```
+
+  - id: on-auth-state-change
+    title: 'onAuthStateChange()'
+    $ref: '@supabase/gotrue-js.GoTrueClient.onAuthStateChange'
+    notes: |
+      - Lets you subscribe to important events ocurring on the user's session.
```
J0, line 293:
```suggestion
      - Subscribes to important events occurring on the user's session.
```

spec/supabase_js_v2.yml
```diff
@@ -284,6 +285,203 @@ functions:
...
+      - Should be used on the frontend/client. It is less useful on the server.
```
J0, line 294:
```suggestion
      - Use on the frontend/client. It is less useful on the server.
```

J0 (approved):

PR #19906, opened 2023-12-20
Add HOSTNAME to studio docker service

Obeyed:
NextJS behavior changed and healthcheck will fail. See https://github.com/vercel/next.js/pull/44627

Bug fix.

When running with eg. Docker Swarm the healthcheck fails because the containerized app is not reachable with the localhost hostname but whatever is set in the HOSTNAME environment for the container.

Example output from logs on startup:

```
supabase-studio   |   ▲ Next.js 13.5.6                                                                                                                                                     
supabase-studio   |   - Local:        http://146ce5541833:3000                                                                                                                             
supabase-studio   |   - Network:      http://192.168.48.3:3000                                                                                                                             
supabase-studio   |                                                                                                                                                                        
supabase-studio   |  ✓ Ready in 1425ms
```

## What is the new behavior?

Example output when using HOSTNAME environment variable:

```
supabase-studio   |   ▲ Next.js 13.5.6
supabase-studio   |   - Local:        http://localhost:3000
supabase-studio   |   - Network:      http://0.0.0.0:3000
supabase-studio   | 
supabase-studio   |  ✓ Ready in 457ms
```


sweatybridge (approved):
I believe this relies on port 3000 being exposed on host as well. Lgtm

sweatybridge:
Closing due to staleness. The health checks are now defined explicitly in compose file to avoid making assumptions about hostname.

PR #19909, opened 2023-12-20
chore: stop polyfilling node modules

kevcodez:
Nextjs has polyfills for crypto to enable browser support, blowing up our main JS bundle

This behaviour can be disabled by setting an experimental flag. It reduces our main bundle size by 25%

This is also the default webpack behaviour since 5.0, nextjs still has it enabled as default, for legacy reasons I suppose

- Replaced generate-password with generate-password-browser for a browser-compatible alternative
- Replaced md5 library with built-in base64 (it's a fallback anyway, shouldn't matter much in this case)
- Fix product menu duplicate key issues on reports page
- Fixed storage reports item being included even when storage is not enabled


ivasilov (approved):

apps/studio/components/layouts/ReportsLayout/ReportsLayout.tsx
```diff
@@ -20,7 +20,8 @@ const ReportsLayout = ({ title, children }: PropsWithChildren<ReportsLayoutProps
 
   const menuItems = [
     {
-      title: '',
+      title: 'Custom reports',
+      key: 'custom-reports',
```
alaister, line 24:
```suggestion
      key: 'reports',
```
Not sure if this is the fix, but this menu item is missing the active state

alaister (approved):
LGTM! One small thing:

alaister (approved):
LGTM! One small thing:

PR #19923, opened 2023-12-20
Chore/rewrite roles form with shadcn

saltcod:


apps/studio/components/interfaces/Database/Roles/RoleRow.tsx
```diff
@@ -200,7 +201,7 @@ const RoleRow = ({ role, disabled = false, onSelectDelete }: RoleRowProps) => {
                               </Tooltip.Content>
                             </Tooltip.Portal>
                           </Tooltip.Root>
-                        )
+                        ) as any)
```
saltcod, line 204: @ivasilov curious about what to do here. 
I typed role permissions which caused this to cascade. 

afterLabel is typed as string now, but as you can see, it is not that.

apps/studio/components/interfaces/Database/Roles/CreateRolePanel.tsx
```diff
@@ -50,113 +76,124 @@ const CreateRolePanel = ({ visible, onClose }: CreateRolePanelProps) => {
...
+                </FormLabel_Shadcn_>
+                <FormControl_Shadcn_ className="col-span-8">
+                  <Input_Shadcn_ {...field} className="w-full" />
+                </FormControl_Shadcn_>
+                <FormMessage_Shadcn_ className="col-start-5 col-span-8" />
+              </FormItem_Shadcn_>
+            )}
+          />
+          <div className="grid gap-2 mt-4 md:grid md:grid-cols-12">
+            <div className="col-span-4">
```
MildTomato, line 125: these tailwind classes could be constants (or maybe even components) so we can reuse them elsewhere.

or maybe we make custom classes in which we can control the size / col width of these layouts from 1 place.

apps/studio/components/interfaces/Database/Roles/RoleRow.tsx
MildTomato: afterLabel is meant to be a string (currently) so no wonder this doesn't work.

We're going to deprecate these components anyway, and then you won't have props like `afterLabel` anymore.

MildTomato: https://github.com/supabase/supabase/blob/master/packages/ui/src/components/Toggle/Toggle.tsx/#L14-L15

quick fix: maybe change these to `string | React.ReactNode`

apps/studio/components/interfaces/Database/Roles/CreateRolePanel.tsx
ivasilov: I was thinking the exact same thing (classes as string constants). I'm still not fully convinced on having some classes set as default on FormLabel, FormItem, FormControl, FormMessage components.

saltcod: same same same 
I say we try and tackle this early next year. Adding to the 2024 list from yesterday.

ivasilov (approved):

PR #19940, opened 2023-12-21
chore: improve auth screens loading

alaister:

chore


apps/studio/components/interfaces/Auth/Users/UsersList.tsx
```diff
@@ -36,59 +37,64 @@ const UsersList = ({
...
+        <>
+          {isLoading && (
+            <>
+              <UsersListItemSkeleton />
+              <UsersListItemSkeleton />
+              <UsersListItemSkeleton />
+              <UsersListItemSkeleton />
+            </>
+          )}
+          {isSuccess && users.length == 0 && (
```
joshenlim, line 63:
```suggestion
          {isSuccess && users.length === 0 && (
```

apps/studio/components/interfaces/Auth/Users/UsersList.tsx
```diff
@@ -36,59 +37,64 @@ const UsersList = ({
...
+                  </p>
+                </div>
               </Table.td>
             </Table.tr>
-          </>
-        }
-      />
-    </Loading>
+          )}
+          {isSuccess &&
```
joshenlim, line 80: while we're refactoring this, probably add isError handler? with `<AlertError/>`

joshenlim (approved):

apps/studio/components/interfaces/Auth/Users/UsersList.tsx
alaister: Good shout! I've added it purely based on `error` since `isError` would be passed as a prop so TS wouldn't be any extra help

PR #19956, opened 2023-12-21
openai.mdx updated deno function to use latest openai package

karim-attia:

docs update

gpt-3.5-turbo reply works but uses outdated openai package

## What is the new behavior?

gpt-3.5-turbo reply works using code recommended by Open AI on https://github.com/openai/openai-node

## Additional context

Improves code readability and beginner friendliness. In addition, it's possible to use GPT-4 and GPT-4 Turbo by just changing the model with this code snippet.


charislam (approved):

charislam:
Thanks so much for this, it's great! You'll just need to run a quick `npm run format` on this so it will pass our merge checks

PR #19977, opened 2023-12-22
Chore/table editor sidepanel revamp

saltcod:
ui updates to the table editor sidebar

screenshot-2023-12-22-at-13 06 32


saltcod:
@joshenlim's feedback: 
1. missing "more" icon in the side menu for each table here (even when selected)
views don't have actions on prod at the moment, so no more menu there

2. Insert button also hidden on prod when a view is selected https://share.cleanshot.com/wFJcrQtn

3.  search styling — fixed
4. search jumping / padding — fixed
5. empty state — I implemented the design but realized that we're using the original pattern 13 other times, so returned it to where it was for consistency. You'll already be focused in that <input> when you see the empty state, so pretty easy to just delete / backspace   https://share.cleanshot.com/mW0qpHRS

MildTomato (approved):

Issue #19992, opened 2023-12-24
Swift Supabase | Upload/Replace file documentation


Lightskyz:
# Improve documentation

## Link 

- https://supabase.com/docs/reference/swift/storage-from-upload
- https://supabase.com/docs/reference/swift/storage-from-update

## Describe the problem

- Error in the parameter name, it's not `fileOptions` but `options`
- Moreover, when we upload the file in the bucket, the default content-type is `text/plain` so I suggest to add this option in the documentation especially that we're uploading a `avatar.png` in the example and it store as `text/plain` instead of `image/png` 

### Upload File
```swift
let fileName = "avatar1.png"

try await supabase.storage
  .from("avatars")
  .upload(
    path: "public/\(fileName)",
    file: fileData,
    options: FileOptions(
      cacheControl: "3600",
      contentType: "image/png",
      upsert: false
    )
  )
```

### Update File

```swift
let fileName = "avatar1.png"

try await supabase.storage
  .from("avatars")
  .update(
    path: "public/\(fileName)",
    file: fileData,
    options: FileOptions(
      cacheControl: "3600",
      contentType: "image/png",
      upsert: true
    )
  )
```

## Additional context

Currently in the documentation:
### Upload File
![Upload File](https://github.com/supabase/supabase/assets/9927423/6ca84030-9849-46f2-b51e-5933a0aca37c)

### Update File
![Update File](https://github.com/supabase/supabase/assets/9927423/61659167-7edd-47c5-b7ac-9b51dc0d0f44)

theahmadchand:
I want to fix this issue

ZainAli104:
Is this issue fixed or not yet?

theahmadchand:
> Is this issue fixed or not yet?

I created a [PR](https://github.com/supabase/supabase/pull/19995) for that

Hallidayo:
Hi everyone, As a pull request has been merged for this issue I'm going to close. Please feel feel to re-open or a create a new issue if this is still happening.

PR #19995, opened 2023-12-24
docs: :memo: Upload/Replace file documentation #19992

theahmadchand:

Docs update

https://github.com/supabase/supabase/issues/19992

- Error in the parameter name, it's not fileOptions but options
- Moreover, when we upload the file in the bucket, the default content-type is text/plain so I suggest to add this option in the documentation especially that we're uploading a avatar.png in the example and it store as text/plain instead of image/png

**Upload a file**

image

**Replace an existing file**

image

## What is the new behavior?

 
- Updated the parameter name from fileOptions to options
- Added `contentType: "image/png"` in the options

**Upload a file**

image

**Replace an existing file**

image


charislam (approved):

charislam:
Thanks! LGTM, you'll just need to take a look at the conflicts

PR #19998, opened 2023-12-25
Update Kotlin tutorials to be aligned with supabase-kt v2.0.0

hieuwu:
Update Kotlin tutorials document after upgrading Supabase to 2.0.0 and update the code accordingly
- Replace old instance `GoTrue` with `Auth`
- Update filter query
- Bump Supabase version

Currently the tutorials is outdated after the release of Supabase Kt 2.0.0

## What is the new behavior?
Update for Supbabase 2.0.0


hieuwu:
Hi @jan-tennert pleae take a look when you have time

apps/docs/pages/guides/getting-started/tutorials/with-kotlin.mdx
```diff
@@ -268,16 +268,20 @@ class ProductRepositoryImpl @Inject constructor(
...
     }
 
     override suspend fun deleteProduct(id: String) {
         return withContext(Dispatchers.IO) {
-            postgrest.from("products").delete {
-                eq("id", id)
+            postgrest["products"].delete {
+                filter {
+                    eq("id", id)
+                }
```
jan-tennert, line 284: So the docs now only mention the `Supabase#from` / `Postgrest#from` method to make postgrest requests rather than `Postgrest#get` (the operator function). Maybe it would be less confusing if we are also using them here?

jan-tennert (approved):
Other than that, thanks!

apps/docs/pages/guides/getting-started/tutorials/with-kotlin.mdx
hieuwu: That makes sense. Just updated

jan-tennert (approved):

charislam (approved):

PR #19999, opened 2023-12-25
Data Type in create column side panel should reset for new column

### LINKED ISSUE #19109: Create column side panel doesn't reset
When you create a column, save it an then create another column right away. 
The data type from the first column is "prefilled" the second time, but isn't actually selected.
If you don't explicitly select it again, it'll fail.  

On success, the side panel should reset to a clean slate.

https://share.cleanshot.com/0gXC77Xr
https://share.cleanshot.com/vfWhQCJP

PaperBoardOfficial:

Bug fix

This PR fixes https://github.com/supabase/supabase/issues/19109

## What is the new behavior?

RCA: The issue was that the 'selected' state was being persisted even after closing the column editor side panel. This was happening because whenever we reopen the side panel, the old value of 'selected' get picked up.

Now, I passed the 'isNewRecord' variable from ColumnEditor to ColumnType to ListBox. There are two cases:
Case 1 : When isNewRecord is true (i.e we are creating a new column), we open the column editor, the old value of 'selected' isn't picked up.
Case 2: When isNewRecord is false (i.e we are editing existing column), the old value of 'selected' is picked up.

## Additional context

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.tsx
@@ -284,6 +284,7 @@ const ColumnEditor = ({
             error={errors.format}
             disabled={columnFields?.foreignKey !== undefined}
             onOptionSelect={(format: string) => onUpdateField({ format, defaultValue: null })}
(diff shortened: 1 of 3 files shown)


alaister:
Hi @learner-mohit,

Thanks for your PR! However, it looks like this has already been fixed.

Feel free to re-open this PR if you can still replicate.

Thanks again 🙌

PR #20004, opened 2023-12-25
Update auth-email-otp.mdx - Clarification of Documentation? Question

christopherreay:

question about the meaning of the DOCS, and the behaviour of one of the options


charislam:
Thanks for the questions, @christopherreay! It'll be easier to get an answer if you open this as a [discussion](https://github.com/supabase/supabase/discussions) rather than a PR.

Closing as we don't want to add these questions into the code base, but feel free to copy them into a discussion!

PR #20008, opened 2023-12-25
fix: docker.mdx supabase-analytics instead of supabase-db

ethicnology:

The documentation says to destroy the analytics container but the command shown destroy the database container

Following the actual documentation, user destroy his database container

## What is the new behavior?

It destroys the analytics container

## Additional context

Self-hosting


ethicnology:
I'm not even sure the previous command are correct, when I followed this part of the documentation I've lost the access to the database by deleting supabase-db

ivasilov (approved):

ethicnology:
@ivasilov 

I wonder if we should have change one of the previous line before:
```
docker exec -it $(docker ps | grep supabase-db | awk '{print $1}') psql -U supabase_admin --password
```
to supabase-analytics

ivasilov:
> @ivasilov
> 
> I wonder if we should have change one of the previous line before:
> 
> ```
> docker exec -it $(docker ps | grep supabase-db | awk '{print $1}') psql -U supabase_admin --password
> ```
> 
> to supabase-analytics

Naah, that line is ok.

PR #20012, opened 2023-12-26
docs: fix api route example

hsuanxyz:


jordienr:
Not sure what this is trying to fix/achieve, please write a better description and we can revisit 🙏

PR #20019, opened 2023-12-26
fix: Remove datatable story

ivasilov:
This PR removes the `datatable` story because we don't have a `datatable` component. If we import it in the future, we can copy over the story from https://github.com/shadcn-ui/ui/pull/11.


alaister (approved):
LGTM!

PR #20055, opened 2023-12-29
proposed changes for issue #20043

### LINKED ISSUE #20043: Impossible to start multiple instances of the docker compose stack due to hardcoded ports
# Bug report

## Describe the bug

The [docker-compose.yml](https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml#L336) currently specifically hardcodes port 4000 of the analytics container to be exposed to 4000. This prevents the docker stack from beeing started on systems where port 4000 is already in use by another application or by another instance of supabase.

## To Reproduce

Follow the [documentation](https://supabase.com/docs/guides/self-hosting/docker) for starting a local docker instance on a system where port 4000 is already in use.

## Expected behavior

This should be configurable to another port by an environment variable.

## System information

Independent of the system it is run on.

shashankshet:

Bug fix #20043 

Before port 4000 was hard coded, due to which multiple instance couldn't be started

## What is the new behavior?
proposed change checks for env variable ANALYTICS_PORT, if specified it will use the port value which was configured in ANALYTICS_PORT else it will use port 4000 by default

please let me know if any changes are required.


Wursteintopf:
Thanks for the Pull Request, this change would be highly appreciated!

PR #20084, opened 2023-12-31
feat: remove sentry sign in events

hf:
This event is no longer useful as random logouts appear to have been almost (if not not completely) solved.


alaister (approved):
LGTM!

PR #20100, opened 2024-01-02
Fix view policies from table editor

joshenlim:
Fix viewing policies from table editor showing 0 results if table is not from public schema


alaister (approved):
LGTM!

PR #20124, opened 2024-01-03
Add validations whenever using new URL

joshenlim:


apps/studio/localStores/storageExplorer/StorageExplorerStore.ts
```diff
@@ -389,20 +389,23 @@ class StorageExplorerStore {
...
-      // Also save it to cache
-      const fileCache = {
-        id: file.id,
-        url: fileUrl,
-        expiresIn: DEFAULT_EXPIRY,
-        fetchedAt: Date.now(),
-      }
-      this.addFileToPreviewCache(fileCache)
+      if (signedUrl) {
+        const formattedUrl = new URL(signedUrl)
```
alaister, line 393: For this one instead of checking that there is a `signedUrl` you could also wrap the `new URL()` call in a `try/catch` and then just return `''` if it fails? wdyt?

joshenlim: ahh good point

alaister (approved):
LGTM!

PR #20127, opened 2024-01-03
Cleanup sentry filters

kamilogorek:
All appropriate comments are next to their related code.

### DIFF
--- a/apps/studio/sentry.client.config.ts
+++ b/apps/studio/sentry.client.config.ts
@@ -20,6 +20,7 @@ Sentry.init({
     }),
   ],
   ignoreErrors: [
+    // Used exclusively in Monaco Editor.
     'ResizeObserver',
     // [Joshen] We currently use stripe-js for customers to save their credit card data
     // I'm unable to reproduce this error on local, staging nor prod across chrome, safari or firefox
@@ -33,14 +34,9 @@ Sentry.init({
     'The provided `href` (/org/[slug]/team) value is missing query values (slug)',
     'The provided `href` (/org/[slug]/billing) value is missing query values (slug)',
     'The provided `href` (/org/[slug]/invoices) value is missing query values (slug)',
-    // [Joshen] Seems to be from hcaptcha
-    "undefined is not an object (evaluating 'n.chat.setReady')",
-    "undefined is not an object (evaluating 'i.chat.setReady')",
   ],
-  denyUrls: [
-    // [Joshen] We're getting a huge influx of events from this particular source
-    'injectCoinbaseWalletDappProvider.js',
-  ],
+  // Allow for exception captures originating only from our own code.
+  allowUrls: ['app:///_next/'],
 })
 
 // Replace dynamic query param with a template text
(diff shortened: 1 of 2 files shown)


apps/studio/sentry.client.config.ts
```diff
@@ -39,15 +34,9 @@ Sentry.init({
     'The provided `href` (/org/[slug]/team) value is missing query values (slug)',
     'The provided `href` (/org/[slug]/billing) value is missing query values (slug)',
     'The provided `href` (/org/[slug]/invoices) value is missing query values (slug)',
-    // [Joshen] Seems to be from hcaptcha
```
kamilogorek, line 42: Those will be filtered via `allowUrls`

apps/studio/sentry.client.config.ts
```diff
@@ -39,15 +34,9 @@ Sentry.init({
...
-    "undefined is not an object (evaluating 'n.chat.setReady')",
-    "undefined is not an object (evaluating 'i.chat.setReady')",
   ],
-  denyUrls: [
```
kamilogorek, line 46: This as well

apps/studio/sentry.client.config.ts
```diff
@@ -75,64 +64,3 @@ function standardiseRouterUrl(url: string) {
 
   return finalUrl
 }
-
```
kamilogorek, line 78: Also code related to implemented Chrome fix.

apps/studio/sentry.client.config.ts
```diff
@@ -39,15 +34,9 @@ Sentry.init({
...
-    "undefined is not an object (evaluating 'n.chat.setReady')",
-    "undefined is not an object (evaluating 'i.chat.setReady')",
   ],
-  denyUrls: [
-    // [Joshen] We're getting a huge influx of events from this particular source
-    'injectCoinbaseWalletDappProvider.js',
-  ],
-  beforeSend: (event) => filterConsoleErrors(event),
+  // Allow for exception captures originating only from our own code.
+  allowUrls: ['app:///_next/'],
```
kamilogorek, line 39: All errors that are thrown from our own code will *always* start with this prefix, as it's enforced by `@sentry/nextjs` SDK itself.

kamilogorek: It can be expanded further once we have a smooth integration and all our own issues are resolved.

apps/studio/sentry.client.config.ts
```diff
@@ -20,17 +20,12 @@ Sentry.init({
     }),
   ],
   ignoreErrors: [
+    // Used exclusively in Monaco Editor.
     'ResizeObserver',
-    'Non-Error exception captured',
```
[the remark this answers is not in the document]
kamilogorek: Note that it's fine to filter _some_ of them, but it should be done granularly via `beforeSend` and it's `hint.originalException` property, not here.

saltcod (approved):

kamilogorek:
We should also remove filters from https://supabase.sentry.io/settings/projects/supabase-app/filters/data-filters/ as they are set in the SDK (and we want to actually remove those listed there).

joshenlim (approved):

PR #20130, opened 2024-01-03
fix: Fix the drag resizing in the GraphiQL

ivasilov:
When resizing the three panels (query, response and docs) in GraphiQL it's possible to over-expand the docs panel and squish the response panel. Also, when collapsing the docs panel by dragging, disable the docs icon.


alaister (approved):
LGTM!

PR #20138, opened 2024-01-03
Chore/GitHub gist sentry

saltcod:
screenshot-2024-01-03-at-14 33 04


apps/studio/sentry.client.config.ts
```diff
@@ -42,6 +42,8 @@ Sentry.init({
     // [Joshen] Seems to be from hcaptcha
     "undefined is not an object (evaluating 'n.chat.setReady')",
     "undefined is not an object (evaluating 'i.chat.setReady')",
+    // [Terry] When users paste in an embedded Github Gist
+    'Unexpected "❤ by GitHu" at line',
```
kamilogorek, line 46:
```suggestion
    // Error thrown by `sql-formatter` lexer when given invalid input
    // Original format: new Error(`Parse error: Unexpected "${text}" at line ${line} column ${col}`)
    /^Parse error: Unexpected ".+" at line \d+ column \d+$/,
```

kevcodez (approved):

PR #20157, opened 2024-01-04
fix: Don't fetch columns when fetching tables by default

ivasilov:
This PR changes the default behavior when fetching multiple tables. The columns are not included but they can be provided by adding a flag.

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ForeignKeySelector/ForeignKeySelector.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ForeignKeySelector/ForeignKeySelector.tsx
@@ -2,12 +2,12 @@ import type { PostgresColumn, PostgresSchema, PostgresTable } from '@supabase/po
 import { find, get, isEmpty, sortBy } from 'lodash'
 import { useEffect, useState } from 'react'
 
-import { Dictionary } from 'types'
 import { useProjectContext } from 'components/layouts/ProjectLayout/ProjectContext'
 import InformationBox from 'components/ui/InformationBox'
 import { FOREIGN_KEY_CASCADE_ACTION } from 'data/database/database-query-constants'
 import { useSchemasQuery } from 'data/database/schemas-query'
 import { useTablesQuery } from 'data/tables/tables-query'
+import { Dictionary } from 'types'
 import {
   AlertDescription_Shadcn_,
   AlertTitle_Shadcn_,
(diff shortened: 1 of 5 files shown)


apps/studio/stores/pgmeta/MetaStore.ts
```diff
@@ -426,7 +426,8 @@ export default class MetaStore implements IMetaStore {
     const connectionString = project?.connectionString
     const tables = await queryClient.fetchQuery({
       queryKey: tableKeys.list(projectRef, 'public'),
```
alaister, line 428: I'm not sure if this needs changing too?
```suggestion
      queryKey: tableKeys.list(projectRef, sourceTableSchema),
```

apps/studio/data/tables/tables-query.ts
```diff
@@ -51,19 +64,19 @@ export async function getTables(
   if (Array.isArray(data) && sortByProperty) {
     return sortBy(data, (t) => t[sortByProperty]) as PostgresTable[]
   }
-  return data as PostgresTable[]
+  return data as Omit<PostgresTable, 'columns'>[]
```
alaister, line 67: Columns will be included if `include_columns` is true though right? So wondering if using the `Omit` is correct?

apps/studio/data/tables/keys.ts
```diff
@@ -1,6 +1,10 @@
 export const tableKeys = {
-  list: (projectRef: string | undefined, schema: string | undefined) =>
-    ['projects', projectRef, 'tables', { schema }] as const,
+  list: (
+    projectRef: string | undefined,
+    schema: string | undefined,
+    includeColumns?: boolean | undefined
+  ) =>
+    ['projects', projectRef, 'tables', { schema }, { includeColumns: !!includeColumns }] as const,
```
alaister, line 7: Any reason you want another object in the key? I've always thought of the final object as kind of the "options" object.
```suggestion
    ['projects', projectRef, 'tables', { schema, includeColumns: !!includeColumns }] as const,
```

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ForeignKeySelector/ForeignKeySelector.tsx
```diff
@@ -257,7 +259,7 @@ const ForeignKeySelector = ({
               ---
             </Listbox.Option>
             {/* @ts-ignore */}
```
alaister, line 261: Wondering if this `@ts-ignore` can be removed while we're here?
```suggestion
```

alaister (commented):
Looking good! Just a couple of questions:

apps/studio/data/tables/tables-query.ts
ivasilov: It returns no columns by default, so this type would be correct in most cases. When using `include_columns`, we would have to cast the return type ([example](https://github.com/supabase/supabase/pull/20157/files#diff-99d7ec12a697697fb966a710290a9603843572d25db8531694eccfce4090487fR67)). 

I don't have a strong opinion on this, just trying to make the default case more type-safe.

alaister (approved):
LGTM!

apps/studio/data/tables/tables-query.ts
alaister: I think casting is a viable solution for now – I don't think it's worth the time investment to get something "slightly" nicer 👍

PR #20161, opened 2024-01-04
Upgrade octokit

fsansalvadore:

Remove 'octokit' dependency from www.
Import from "@oktokit/core" instead.
Upgrades all octokit versions

Fixes https://github.com/supabase/supabase/security/dependabot/1601 and https://github.com/supabase/supabase/security/dependabot/1602


apps/www/package.json
```diff
@@ -37,7 +37,6 @@
     "next-mdx-remote": "^4.4.1",
     "next-seo": "^5.14.1",
     "next-themes": "^0.2.1",
-    "octokit": "^2.0.22",
```
alaister, line 40: Does `@octokit/core` need to be added to `package.json` in place of this?

alaister (approved):

PR #20162, opened 2024-01-04
docs: release auth hooks

J0:

We remove the notice that Auth Hooks in beta as we will be unflagging Auth Hooks on the dashboard.

### DIFF
--- a/apps/docs/pages/guides/auth/auth-hooks.mdx
+++ b/apps/docs/pages/guides/auth/auth-hooks.mdx
@@ -8,10 +8,6 @@ export const meta = {
   description: 'Use Supabase Postgres Functions to customize your authentication flow',
 }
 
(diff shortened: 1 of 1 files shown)


charislam (approved):

thibault-chausson:
Maybe a bug...
Because I am using the free version on SupaBase and the online [tutorial](https://supabase.com/docs/guides/auth/auth-hooks?language=add-admin-role) states that we can add hooks to customize the token, but it does not work when I try to add them. Attached is a screenshot of the tutorial and my Supabase page for editing the hooks.
Thank you for your help :pray::pray:!

## Tutorial 
Capture d’écran 2024-01-09 à 19 23 32

## My dashboard 
Capture d’écran 2024-01-09 à 19 22 51

PR #20168, opened 2024-01-04
chore: Send a PR to supabase/cli on publish

ivasilov:


.github/workflows/publish_image.yml
```diff
@@ -140,3 +140,58 @@ jobs:
...
+      service_to_update: 'studio'
+      cli_repo_owner: supabase
+      cli_repo: cli
+      cli_repo_main_branch: 'main'
+      cli_repo_pr_title: 'New Studio Version'
+      github_user: 'Ivan Vasilov'
+      github_user_email: 'vasilov.ivan@gmail.com'
+
+    steps:
+      - name: Setup Git config
```
kevcodez, line 159: I think this is a lot of implicit information and I'd rather have the CLI repo handle the majority of this, in case they change something.

What I'd prefer, is triggering a custom event and then have a github action in the CLI repo that reacts to this event.

Example step to dispatch an event


```

 - name: Dispatch
        uses: peter-evans/repository-dispatch@v2.1.0
        with:
          # This action creates repository_dispatch events. The default GITHUB_TOKEN token can only be used if you are dispatching
          # the same repository that the workflow is executing in. To dispatch to a remote repository you must create a Personal
          # Access Token (PAT) with the repo scope and store it as a secret. If you will be dispatching to a public repository
          # then you can use the more limited public_repo scope.
          token: ${{ secrets.NPM_TOKEN_XYZ }}
          repository: ${{ matrix.repository }}
          event-type: studio-released
          client-payload: '{"version": "${{ steps.last_release.outputs.tag_name }}", "message": "${{ github.event.head_commit.message }}"}'
```

kevcodez: The CLI could then react to it with a new action

```
  repository_dispatch:
    types: [studio-released]

```

Ideally, CLI can then use their regular package manager

.github/workflows/publish_image.yml
```diff
@@ -140,3 +140,58 @@ jobs:
...
+        run: |
+          cd '${{ env.cli_repo }}'
+          git checkout -b '${{ env.service_to_update }}/${{ needs.settings.outputs.image_version }}'
+
+      - name: Update file
+        run: |
+          cd '${{ env.cli_repo }}'
+          sed -i 's/${{ env.cli_repo_owner }}\/${{ env.service_to_update }}:.*/${{ env.cli_repo_owner }}\/${{ env.service_to_update }}:${{ needs.settings.outputs.image_version }}"/' internal/utils/misc.go
+
+      - name: Commit changes
```
kevcodez, line 181: You could also combine these three steps into a single simpler one using a GH action

Example

```
- name: Create Pull Request
        uses: peter-evans/create-pull-request@v4.2.0
        with:
          add-paths: package*.json
          branch: studio-${{ github.event.client_payload.version }}
          body: Bump the import package to version ${{ github.event.client_payload.version }}.
          base: main
          title: '${{ github.event.client_payload.message }}'
          reviewers: ...
          committer: ...
          author: ...
```

ivasilov: I've looked into this action, it looks like it only supports PRs towards its own repo. Even if it supports other repos, the job still has to check the repo, make changes and then use this action to commit and push (which is easiest part of the job).

kevcodez: The idea would be that supabase only emits an event and the cli repo would then do the changes (as the cli repo knows it's structure) and create a PR on it's own repo.

Trigger would be the dispatched event (as outlined in the other comment)

.github/workflows/publish_image.yml
```diff
@@ -134,3 +134,59 @@ jobs:
...
+    needs:
+      - settings
+      # - publish
+    runs-on: ubuntu-latest
+    env:
+      service_to_update: 'studio'
+      cli_repo_owner: supabase
+      cli_repo: cli
+      cli_repo_main_branch: 'develop'
+      cli_repo_pr_title: 'fix: Bump studio to the latest image version'
```
kevcodez, line 148:
```suggestion
      cli_repo_pr_title: 'chore: Bump studio to the latest image version'
```

dependency upgrades should be chore not fix (or doesnt `chore` properly trigger a release)?

.github/workflows/publish_image.yml
```diff
@@ -134,3 +134,59 @@ jobs:
     with:
       version: ${{ needs.settings.outputs.image_version }}
     secrets: inherit
+
+  update-remote-repo:
+    needs:
+      - settings
```
kevcodez, line 140: Is that enough permissions to do a PR?

.github/workflows/publish_image.yml
```diff
@@ -134,3 +134,59 @@ jobs:
...
+      github_user_email: 'vasilov.ivan@gmail.com'
+
+    steps:
+      - name: Setup Git config
+        run: |
+          git config --global user.name '${{ env.github_user }}'
+          git config --global user.email '${{ env.github_user_email }}'
+
+      - name: Checkout remote repository
+        uses: actions/checkout@v2
```
kevcodez, line 159:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/publish_image.yml
```diff
@@ -134,3 +134,59 @@ jobs:
...
+      - name: Update file
+        run: |
+          cd '${{ env.cli_repo }}'
+          sed -i 's/${{ env.cli_repo_owner }}\/${{ env.service_to_update }}:.*/${{ env.cli_repo_owner }}\/${{ env.service_to_update }}:${{ needs.settings.outputs.image_version }}"/' pkg/config/constants.go
+
+      - name: Commit changes
+        run: |
+          cd '${{ env.cli_repo }}'
+          git add .
+          git commit -m "Update ${{ env.service_to_update }} version to ${{ needs.settings.outputs.image_version }}"
```
kevcodez, line 180:
```suggestion
          git commit -m "chore: Update ${{ env.service_to_update }} version to ${{ needs.settings.outputs.image_version }}"
```

Semantic versioning prefix

.github/workflows/publish_image.yml
```diff
@@ -134,3 +134,59 @@ jobs:
...
+
+      - name: Push changes
+        run: |
+          cd '${{ env.cli_repo }}'
```
kevcodez, line 184: Do you need `'` here?

.github/workflows/publish_image.yml
```diff
@@ -134,3 +134,59 @@ jobs:
...
+          git push origin '${{ env.service_to_update }}/${{ needs.settings.outputs.image_version }}'
+
+      - name: Create Pull Request
+        run: |
+          cd '${{ env.cli_repo }}'
+          gh pr create --base '${{ env.cli_repo_main_branch }}' --title "${{ env.cli_repo_pr_title }} ${{ needs.settings.outputs.image_version }}" --body "Updates service ${{ env.service_to_update }} new version to ${{ needs.settings.outputs.image_version }}"
+        env:
+          GH_TOKEN: ${{ secrets.USER_ACCESS_TOKEN }}
```
kevcodez, line 192: This shouldn't run on a user access token, but with the GH permissions ideally to avoid expiring user keys

kevcodez: This might just work without the env var

.github/workflows/publish_image.yml
kevcodez: but for the sake of not refactoring the action, no hard feelings, can leave it as-is

.github/workflows/publish_image.yml
ivasilov: According to [these docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-github-cli-in-workflows), you have to use a token to run `github cli`. Can you show me the docs in how we can do this without the token (or with a different token)?

kevcodez: I think there is a default token provided (with the proper scope and permissions depending on `permissions`) when using `secrets.GITHUB_TOKEN`

kevcodez:
Closing as stale

PR #20173, opened 2024-01-04
ci: trigger deploy of site with shared origin [ci shared auth]

charislam:
add a trigger to generate a set of preview deploys for studio/docs/www that are proxied together via www's URL
makes it possible to test authenticated features in preview for docs/www, which otherwise have no signin mechanism

### DIFF
--- a/apps/www/components/Nav/index.tsx
+++ b/apps/www/components/Nav/index.tsx
@@ -164,10 +164,10 @@ const Nav = () => {
                     ) : (
                       <>
                         <Button type="default" className="hidden lg:block" asChild>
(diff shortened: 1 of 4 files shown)


apps/www/components/Nav/index.tsx
```diff
@@ -164,10 +164,10 @@ const Nav = () => {
                     ) : (
                       <>
                         <Button type="default" className="hidden lg:block" asChild>
-                          <Link href="https://supabase.com/dashboard">Sign in</Link>
+                          <Link href="/dashboard">Sign in</Link>
```
charislam, line 167: not 100% tied to this but it's a nice convenience thing so you don't have to manually navigate to log in on the preview origin

PR #20189, opened 2024-01-05
chore: require spend cap to be disabled when scaling disk

kevcodez:
When users scale beyond 8GB of disk (included in Pro plan), the spend cap needs to be disabled to scale seamlessly.


ecktoteckto (approved):
lgtm

PR #20205, opened 2024-01-05
Fix bug: verify email links don't work if opened in a different browser

scosman:
Currently following this guide, you get a 500 error if clicking there "verify email" link in a browser other than the one you submitted the create account request from. 

We should verify in the callback, but then redirect the user to login (as opposed to error), as this isn't a real error path, and user action is needed to recover.

Bug fix for the sample code in the guide.

Long discussions on issue and best UX here:

 - https://github.com/supabase/auth-helpers/issues/545
 - https://github.com/supabase/auth-helpers/issues/545#issuecomment-1666960329

## What is the new behavior?

Detects possible issue (500) and redirects user to login to resolve the issue.

## Additional context

NA

### DIFF
--- a/apps/docs/pages/guides/getting-started/tutorials/with-sveltekit.mdx
+++ b/apps/docs/pages/guides/getting-started/tutorials/with-sveltekit.mdx
@@ -267,12 +267,23 @@ In the following code snippet, we perform the following steps:
 ```js title=src/routes/auth/callback/+server.js
 // src/routes/auth/callback/+server.js
 import { redirect } from '@sveltejs/kit'
+import { isAuthApiError } from '@supabase/supabase-js'
 
 export const GET = async ({ url, locals: { supabase } }) => {
   const code = url.searchParams.get('code')
 
   if (code) {
-    await supabase.auth.exchangeCodeForSession(code)
+    try {
+      await supabase.auth.exchangeCodeForSession(code)
+    } catch (error) {
+       // If opened in another browser without PCKE cookie, need to redirect user to login.
+       if (isAuthApiError(error)) {
+         // optionally add "Email verified! Please login." message to your login UI using the query param.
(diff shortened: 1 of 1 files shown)


apps/docs/pages/guides/getting-started/tutorials/with-sveltekit.mdx
```diff
@@ -267,12 +267,23 @@ In the following code snippet, we perform the following steps:
...
 export const GET = async ({ url, locals: { supabase } }) => {
   const code = url.searchParams.get('code')
 
   if (code) {
-    await supabase.auth.exchangeCodeForSession(code)
+    try {
+      await supabase.auth.exchangeCodeForSession(code)
+    } catch (error) {
+       // If opened in another browser without PCKE cookie, need to redirect user to login.
+       if (isAuthApiError(error)) {
```
charislam, line 280: @kangmingtay, is this the only possible reason `exchangeCodeForSession` can throw an auth API error? also the email wouldn't be verified because there's no code verifier sent in the token exchange right?

scosman: From my testing: the email is verified, but the uncaught auth error breaks the UI flow. 

Would be great to target the exact error case here if possible!

scosman: @charislam for additional context, the official auth guide is redirecting on error as well: https://supabase.com/docs/guides/auth/server-side/email-based-auth-with-pkce-flow-for-ssr?framework=sveltekit

It's less specific, not checking that it's an auth error.

Hallidayo:
Hi @scosman - We’re doing a bit of housekeeping and plan to close out PRs that have been inactive for more than 3 months.

PR #20212, opened 2024-01-06
chore: sign impersonation JWT with browser crypto

kevcodez:
jwt-simple was published 5 years ago and is not compatible with browsers. Thus, Next.js polyfills all the Node.js crypto APIs blowing up our bundle.

This PR aims to get rid of the jwt-simple library and use built-in crypto APIs from the browser and removes the polyfill from Next.js.

Tested on staging by creating multiple users, setting up an RLS policy and then impersonating the user

### DIFF
--- a/apps/studio/lib/role-impersonation.ts
+++ b/apps/studio/lib/role-impersonation.ts
@@ -1,5 +1,3 @@
-import jwt from 'jwt-simple'
-
 import { User } from 'data/auth/users-query'
 import { uuidv4 } from './helpers'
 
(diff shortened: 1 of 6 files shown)


apps/studio/lib/role-impersonation.ts
```diff
@@ -114,12 +112,63 @@ export function wrapWithRoleImpersonation(
...
-) {
+): Promise<string> {
   const claims = getPostgrestClaims(projectRef, role)
 
-  return jwt.encode(claims, jwtSecret, 'HS256')
+  return createToken(claims, jwtSecret)
+}
+
+async function createToken(payload: object, key: string) {
+  var header = { typ: 'JWT', alg: 'HS256' }
```
joshenlim, line 126: use `const`s around here? since they aren't getting replaced
header, segments, footer

apps/studio/lib/role-impersonation.ts
```diff
@@ -114,12 +112,63 @@ export function wrapWithRoleImpersonation(
...
+      },
+      {
+        name: 'HMAC',
+        hash: { name: 'SHA-256' },
+      },
+      false,
+      ['sign', 'verify']
+    )
+    .then((key) => {
+      var jsonString = JSON.stringify(data)
```
joshenlim, line 157: likewise `const`s around here, and in the then block below

also quick question - any error handling (try catches) we need to do here?

apps/studio/lib/role-impersonation.ts
kevcodez: Using const now

apps/studio/lib/role-impersonation.ts
kevcodez: Using const now

At least added a console error log - it worked fine on my end, and canIUse support looks good for these functions

joshenlim (approved):

PR #20250, opened 2024-01-08
Remove trailing comma

### LINKED ISSUE #17794: Ending JSON Data With a Comma Leads to an Error
## Describe the bug

When writing JSON data, sometimes a comma is left at the end of the last value in the object, like this:

{
  "id": 3,
  "title": "Test Title 1"**,** <- (here)
}

Most code editors are able to detect this and remove this comma when saving. When using the JSON editor in supabase, it throws an error and does not allow the user to save.

https://github.com/supabase/supabase/assets/59986120/29b6e901-7124-4d60-ada7-e75d581340ff

## Expected behavior

The editor should detect the trailing comma, remove it, and save the updated data.

faizanraso:

This PR resolves https://github.com/supabase/supabase/issues/17794

Currently, when using the JSON editor, leaving a trailing comma after the last value in the object will throw an error.

## What is the new behavior?

The trailing comma is removed, and the data is saved. The video below demonstrates this update. A trailing comma is added to the JSON data and is submitted (saved). When the data is reopened, the trailing comma is not there, and no error is thrown.

https://github.com/supabase/supabase/assets/59986120/3e34a160-e0c1-4555-b34b-ff241b2b2f9e


apps/studio/components/grid/components/editor/JsonEditor.tsx
```diff
@@ -36,6 +36,9 @@ export const JsonEditor = <TRow, TSummaryRow = unknown>({
   }, [])
 
   const saveChanges = useCallback((newValue: string | null) => {
+    if (newValue && newValue.slice(-3, -2) === ',') {
```
kevcodez, line 39: We should not only check for the character on the last element, we should remove trailing commas in all places - we can probably skip the check and just check for `newValue` to be truthy and apply the regex replace. This will ensure _all_ occurrences of trailing commas are properly replaced

We also need to ensure that we don't replace false positives, some unit tests for that would be great, as any false positives/wrong replacements would be very destructive

Additionally, a comment would be great, as it's hard to just parse a regex on the fly without some context why/what's happening

kevcodez (changes requested):

apps/studio/components/grid/components/editor/JsonEditor.tsx
faizanraso: I'm not sure I full understand what you mean by trailing commas in all places. Currently, the JSON editor does not except an entry/value which contains a comma after the last item in the object. For example:

```
{
   "a": "one",
   "b": "two"**,**
}
```

The JSON editor does not accept the above entry as there is comma after `"b": "two"`. The purpose of this PR is to detect if there is a comma after the last item in the object (which is why the `newValue.slice(-3, -2) === ','` check is performed) and replaces it using regex.

If you could elaborate a little bit more on your suggestions, I would love to make the changes!

kevcodez: A valid JSON is also an array with multiple objects, i.e.

```
[
{
   "fruit1": "apple",
   "fruit2": "banana",
},
{
   "fruit1": "apple",
   "fruit2": "banana"
}
]
```

If the trailing comma is on another element (not just the last one), this should also be recognized and fixed

faizanraso: @kevcodez Thanks for the explanation. I've updated my code accordingly. 

The regular expression now searches for any comma that immediately precedes a closing curly brace, "}" or a closing square bracket, "]", which indicates that it is a trailing comma. I also noticed that the SideEditor JSON Editor returns is also not able to handle trailing commas, so I utilized the same regular expression to check for, and remove any trailing commas when saving on that editor. 

I also added comments to both of the editor files to explain the logic.

joshenlim:
hey @faizanraso - would you mind extracting your regex into its own function like `removeJSONTrailingComma` (probably can be in lib/helpers since it looks like can be used in multiple places, not just the table editor) and write some unit tests for it in `tests/unit`?

this would help us eyeball any edge cases we need to check for and reduce manual testing + prevent regression in the future if any

faizanraso:
> hey @faizanraso - would you mind extracting your regex into its own function like `removeJSONTrailingComma` (probably can be in lib/helpers since it looks like can be used in multiple places, not just the table editor) and write some unit tests for it in `tests/unit`?
> 
> this would help us eyeball any edge cases we need to check for and reduce manual testing + prevent regression in the future if any

Hi @joshenlim, thanks for the feedback! I'm currently working on this. I created a new branch for this issue as I was having some issues with commiting on this one. The new can be found here: https://github.com/supabase/supabase/pull/20781

faizanraso:
> > hey @faizanraso - would you mind extracting your regex into its own function like `removeJSONTrailingComma` (probably can be in lib/helpers since it looks like can be used in multiple places, not just the table editor) and write some unit tests for it in `tests/unit`?
> > this would help us eyeball any edge cases we need to check for and reduce manual testing + prevent regression in the future if any
> 
> Hi @joshenlim, thanks for the feedback! I'm currently working on this. I created a new branch for this issue as I was having some issues with commiting on this one. The new can be found here: #20781

Hi @joshenlim , the PR is all ready and can be found here: #20781. I'm not too sure if the test cases added are sufficient, so please let me know if you want me to make any changes.

PR #20261, opened 2024-01-09
Feature: allow users to copy the selected cell value

jordienr:

When a user clicks a cell in the database grid/table, they will now be able to ⌘ + V (or control + V) to copy the value

- Copying with keyboard is not allowed
- Pressing any key will open the context menu to edit the value, this has been respected

## What is the new behavior?

- ⌘+C/Control+C will now copy the cell value into your clipboard

CleanShot 2024-01-09 at 10 27 19

## Additional context


jordienr:
> you mean "⌘+**C**/Control+**C**" ?

yup, my bad 🥴

apps/studio/components/grid/components/grid/Grid.tsx
```diff
@@ -97,7 +99,28 @@ export const Grid = memo(
         })
       }
 
+      const [selectedCell, setSelectedCell] = useState<{ rowIdx: number; row: any; column: any }>()
```
alaister, line 102: I'm wondering if we could get away with using `useRef()` here instead of `useState()`? Does this state need to be part of the React lifecycle for anything? Swapping it would reduce unnecessary rerenders.

alaister (approved):
This is great! The amount of times I've thought I'd copied something without it working before.

I've left one possible performance improvement comment, but otherwise, LGTM! 🔥

apps/studio/components/grid/components/grid/Grid.tsx
```diff
@@ -97,7 +99,28 @@ export const Grid = memo(
...
+        if (!text) return
+        copyToClipboard(text)
+      }
+
+      useKeyboardShortcuts({
+        'Command+c': (event: KeyboardEvent) => {
+          event.stopPropagation()
+          copyCellValue()
+        },
+        'Control+c': (event: KeyboardEvent) => {
```
joshenlim, line 116: this seems like a duplicate?

apps/studio/components/grid/components/grid/Grid.tsx
jordienr: good catch, prob better to avoid useState in this one. thanks!

PR #20262, opened 2024-01-09
Offer select closest region CTA in project creation

joshenlim:
image


apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
```diff
@@ -0,0 +1,62 @@
...
+  selectedRegion,
+  onSelectRegion,
+}: RegionSelectorProps) => {
+  const router = useRouter()
+  const availableRegions = getAvailableRegions(PROVIDERS[cloudProvider].id)
+  const { data: region, isLoading, isSuccess } = useDefaultRegionQuery({ cloudProvider })
+
+  useEffect(() => {
+    if (isSuccess && region) onSelectRegion(region)
+  }, [isSuccess])
```
alaister, line 26: Shouldn't actually make a difference since `isSuccess` and `region` and linked, but better to be explicit
```suggestion
  }, [isSuccess, region])
```

alaister (approved):
LGTM!

PR #20263, opened 2024-01-09
chore: Update billing related text on add on upgrade pages

ecktoteckto:

Update of explanation text.

At the moment texts on the add-on upgrade pages do not take into account whether billing is done via Supabase or a partner.

## What is the new behavior?

Texts are customized depending on billing system.

### DIFF
--- a/apps/studio/components/interfaces/Settings/Addons/ComputeInstanceSidePanel.tsx
+++ b/apps/studio/components/interfaces/Settings/Addons/ComputeInstanceSidePanel.tsx
@@ -383,38 +383,22 @@ const ComputeInstanceSidePanel = () => {
               </p>
             )}
 
-            {hasChanges &&
-              (selectedCategory !== 'micro' && selectedCompute?.price_interval === 'monthly' ? (
-                // Monthly payment with project-level subscription
-                <p className="text-sm text-foreground-light">
-                  Upon clicking confirm, the amount of{' '}
-                  <span className="text-foreground">
-                    ${selectedCompute?.price.toLocaleString()}
-                  </span>{' '}
-                  will be added to your monthly invoice. Any previous compute addon is prorated and
-                  you're immediately charged for the remaining days of your billing cycle. The addon
-                  is prepaid per month and in case of a downgrade, you get credits for the remaining
-                  time.
-                </p>
-              ) : selectedCategory !== 'micro' ? (
-                // Hourly usage-billing with org-based subscription
-                <p className="text-sm text-foreground-light">
-                  There are no immediate charges when changing compute. Compute Hours are a
-                  usage-based item and you're billed at the end of your billing cycle based on your
-                  compute usage. Read more about{' '}
-                  <Link
-                    href="https://supabase.com/docs/guides/platform/org-based-billing#usage-based-billing-for-compute"
-                    target="_blank"
-                    rel="noreferrer"
-                    className="underline"
-                  >
-                    Compute Billing
-                  </Link>
-                  .
-                </p>
-              ) : (
-                <></>
-              ))}
+            {hasChanges && selectedCategory !== 'micro' && (
+              <p className="text-sm text-foreground-light">
+                There are no immediate charges when changing compute. Compute Hours are a
+                usage-based item and you're billed at the end of your billing cycle based on your
+                compute usage. Read more about{' '}
+                <Link
+                  href="https://supabase.com/docs/guides/platform/org-based-billing#usage-based-billing-for-compute"
+                  target="_blank"
+                  rel="noreferrer"
+                  className="underline"
+                >
+                  Compute Billing
+                </Link>
+                .
+              </p>
+            )}
 
             {hasChanges && !blockMicroDowngradeDueToPitr && (
               <Alert
@@ -453,7 +437,7 @@ const ComputeInstanceSidePanel = () => {
                   <IconAlertTriangle className="h-4 w-4" />
                   <AlertDescription_Shadcn_>
                     You have a scheduled subscription change that will be canceled if you change
-                    your PITR add on.
+                    your Optimized Compute add on.
                   </AlertDescription_Shadcn_>
                 </Alert_Shadcn_>
               )}
--- a/apps/studio/components/interfaces/Settings/Addons/CustomDomainSidePanel.tsx
+++ b/apps/studio/components/interfaces/Settings/Addons/CustomDomainSidePanel.tsx
@@ -246,8 +246,9 @@ const CustomDomainSidePanel = () => {
               (selectedCustomDomain?.price ?? 0) < (subscriptionCDOption?.variant.price ?? 0) ? (
                 subscription?.billing_via_partner === false && (
                   <p className="text-sm text-foreground-light">
-                    Upon clicking confirm, the amount of that's unused during the current billing
-                    cycle will be returned as credits that can be used for subsequent billing cycles
+                    Upon clicking confirm, the add-on is removed immediately and any unused time in
+                    the current billing cycle is added as prorated credits to your organization and
+                    used in subsequent billing cycles
                   </p>
                 )
               ) : (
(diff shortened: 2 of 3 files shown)


apps/studio/components/interfaces/Settings/Addons/CustomDomainSidePanel.tsx
```diff
@@ -244,19 +244,31 @@ const CustomDomainSidePanel = () => {
...
                   <span className="text-foreground">
                     ${selectedCustomDomain?.price.toLocaleString()}
                   </span>{' '}
-                  will be added to your monthly invoice. You're immediately charged for the
-                  remaining days of your billing cycle. The addon is prepaid per month and in case
-                  of a downgrade, you get credits for the remaining time.
+                  will be added to your monthly invoice.{' '}
+                  {subscription?.billing_via_partner ? (
+                    <>
+                      For the current billing cycle you'll be charged a pro-rated amount at the end
```
ecktoteckto, line 262: I haven't included the info what the billing cycle looks like. Wanted to avoid an information overload, Fly's billing cycle might change in the future and it's documented here https://supabase.com/docs/guides/platform/fly-postgres#pricing

apps/studio/components/interfaces/Settings/Addons/CustomDomainSidePanel.tsx
```diff
@@ -244,19 +244,31 @@ const CustomDomainSidePanel = () => {
             <>
               {selectedOption === 'cd_none' ||
               (selectedCustomDomain?.price ?? 0) < (subscriptionCDOption?.variant.price ?? 0) ? (
-                <p className="text-sm text-foreground-light">
-                  Upon clicking confirm, the amount of that's unused during the current billing
-                  cycle will be returned as credits that can be used for subsequent billing cycles
-                </p>
+                !subscription?.billing_via_partner && (
```
ecktoteckto, line 247: Is this correct?

apps/studio/components/interfaces/Settings/Addons/CustomDomainSidePanel.tsx
```diff
@@ -244,19 +244,31 @@ const CustomDomainSidePanel = () => {
...
+                  {subscription?.billing_via_partner ? (
+                    <>
+                      For the current billing cycle you'll be charged a pro-rated amount at the end
+                      of the cycle.{' '}
+                    </>
+                  ) : (
+                    <>
+                      The addon is prepaid per month and in case of a downgrade, you get credits for
+                      the remaining time. For the current billing cycle you're immediately charged a
+                      pro-rated amount for the remaining days.
```
ecktoteckto, line 269: Moved "The addon is prepaid..." up higher. This way it's about future billing cycles first, and then about the current billing cycle.

apps/studio/components/interfaces/Settings/Addons/PITRSidePanel.tsx
```diff
@@ -341,19 +342,42 @@ const PITRSidePanel = () => {
...
+              {subscription?.billing_via_partner &&
+                subscription.scheduled_plan_change?.target_plan !== undefined && (
+                  <Alert_Shadcn_ variant={'warning'} className="mb-2">
+                    <IconAlertTriangle className="h-4 w-4" />
+                    <AlertDescription_Shadcn_>
+                      You have a scheduled subscription change that will be canceled if you change
+                      your PITR add on.
+                    </AlertDescription_Shadcn_>
+                  </Alert_Shadcn_>
+                )}
```
ecktoteckto, line 380: Am I right that this is missing right now?

apps/studio/components/interfaces/Settings/Addons/ComputeInstanceSidePanel.tsx
```diff
@@ -383,38 +383,22 @@ const ComputeInstanceSidePanel = () => {
               </p>
             )}
 
-            {hasChanges &&
-              (selectedCategory !== 'micro' && selectedCompute?.price_interval === 'monthly' ? (
```
ecktoteckto, line 387: Is outdated, right?

apps/studio/components/interfaces/Settings/Addons/ComputeInstanceSidePanel.tsx
```diff
@@ -383,38 +383,22 @@ const ComputeInstanceSidePanel = () => {
...
-                <p className="text-sm text-foreground-light">
-                  Upon clicking confirm, the amount of{' '}
-                  <span className="text-foreground">
-                    ${selectedCompute?.price.toLocaleString()}
-                  </span>{' '}
-                  will be added to your monthly invoice. Any previous compute addon is prorated and
-                  you're immediately charged for the remaining days of your billing cycle. The addon
-                  is prepaid per month and in case of a downgrade, you get credits for the remaining
-                  time.
-                </p>
```
ecktoteckto, line 398: Is obsolete, right?

kevcodez: Yeah - monthly price_interval was only relevant during migration to org-based billing. Compute is only billed hourly

apps/studio/components/interfaces/Settings/Addons/CustomDomainSidePanel.tsx
kevcodez: When using optional chaining, it's best to explicitly check for false or true, otherwise you'll get a false negative when the data hasn't loaded yet, i.e.

!subscription?.billing_via_partner will not render if value is false, null or undefined. During loading the data, it will be undefined, so it will show in that case, too. Best to check like

```
subscription?.billing_via_partner === false
```

apps/studio/components/interfaces/Settings/Addons/PITRSidePanel.tsx
```diff
@@ -341,19 +342,42 @@ const PITRSidePanel = () => {
             <>
               {selectedOption === 'pitr_0' ||
               (selectedPitr?.price ?? 0) < (subscriptionPitr?.variant.price ?? 0) ? (
-                <p className="text-sm text-foreground-light">
-                  Upon clicking confirm, the amount of that's unused during the current billing
-                  cycle will be returned as credits that can be used for subsequent billing cycles
-                </p>
+                subscription?.billing_via_partner === false && (
+                  <p className="text-sm text-foreground-light">
+                    Upon clicking confirm, the amount of that's unused during the current billing
```
kevcodez, line 347: > Upon clicking confirm, the amount of that's unused during the current billing cycle will be returned as credits that can be used for subsequent billing cycles

Sounds a bit odd, would change to

> Upon clicking confirm, the add-on is removed immediately and any unused time in the current billing cycle is added as prorated credits to your organization and used in subsequent billing cycles

apps/studio/components/interfaces/Settings/Addons/PITRSidePanel.tsx
kevcodez: Yup, good call!

apps/studio/components/interfaces/Settings/Addons/PITRSidePanel.tsx
kevcodez: Let's do the same copy adjustment in the custom domain addon panel

apps/studio/components/interfaces/Settings/Addons/CustomDomainSidePanel.tsx
ecktoteckto: Oh yea, thanks! Changed it!

ecktoteckto: But the condition itself is correct, right?

kevcodez: Yep

kevcodez (approved):

apps/studio/components/interfaces/Settings/Addons/CustomDomainSidePanel.tsx
```diff
@@ -244,19 +244,32 @@ const CustomDomainSidePanel = () => {
...
               (selectedCustomDomain?.price ?? 0) < (subscriptionCDOption?.variant.price ?? 0) ? (
-                <p className="text-sm text-foreground-light">
-                  Upon clicking confirm, the amount of that's unused during the current billing
-                  cycle will be returned as credits that can be used for subsequent billing cycles
-                </p>
+                subscription?.billing_via_partner === false && (
+                  <p className="text-sm text-foreground-light">
+                    Upon clicking confirm, the add-on is removed immediately and any unused time in
+                    the current billing cycle is added as prorated credits to your organization and
+                    used in subsequent billing cycles
```
kevcodez, line 251: Could use a "." at the end

ecktoteckto: Done (also in PITRSidePanenl).

PR #20284, opened 2024-01-09
Feature: Mobile view for tables

wra-sol:
Feature - Mobile View for Table editor
Currently, view is rigid and non-responsive. Unable to even preview details on mobile.

Modified table view and header to be viewable on mobile.
Added 'More Settings' on Header
Reorganized Header actions to accommodate new responsive styling.
Modified ProductMenuBar to hide when table is selected.

Settings with expanded more options Table view with minimized settings


saltcod:
Hi @wra-sol thanks so much for this. We're in the process of doing a more comprehensive look at the Dashboard sidebars and the nav in general. I'm going to mark this one as closed in advance of that one. We appreciate the contribution.

PR #20287, opened 2024-01-09
docs: add info about React Native `AppState` and Supabase Auth

hf:
It wasn't explained anywhere that folks should call start/stopAutoRefresh in React Native or other non-browser environments.

### DIFF
--- a/apps/docs/pages/guides/getting-started/tutorials/with-expo-react-native.mdx
+++ b/apps/docs/pages/guides/getting-started/tutorials/with-expo-react-native.mdx
@@ -169,9 +169,21 @@ Users would be able to sign in with their email and password.
 
 ```tsx components/Auth.tsx
 import React, { useState } from 'react'
-import { Alert, StyleSheet, View } from 'react-native'
+import { Alert, StyleSheet, View, AppState } from 'react-native'
 import { supabase } from '../lib/supabase'
 import { Button, Input } from 'react-native-elements'
+
+// Tells Supabase Auth to continuously refresh the session automatically if
+// the app is in the foreground. When this is added, you will continue to receive
+// `onAuthStateChange` events with the `TOKEN_REFRESHED` or `SIGNED_OUT` event
+// if the user's session is terminated. This should only be registered once.
+AppState.addEventListener('change', (state) => {
+  if (state === 'active') {
+    supabase.auth.startAutoRefresh()
+  } else {
+    supabase.auth.stopAutoRefresh()
+  }
+})
 
 export default function Auth() {
   const [email, setEmail] = useState('')
--- a/apps/docs/spec/supabase_js_v2.yml
+++ b/apps/docs/spec/supabase_js_v2.yml
@@ -749,6 +749,56 @@ functions:
           ```js
           const { data, error } = await supabase.auth.getSession()
           ```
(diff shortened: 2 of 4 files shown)


apps/docs/pages/guides/auth/quickstarts/react.mdx
```diff
@@ -78,13 +78,23 @@ export const meta = {
 
       ```js src/index.js
         import './index.css'
+        import { AppState } from 'react-native'
```
charislam, line 81: might be a bit confusing to mix native and browser since this isn't a native-only tutorial

what do you think about adding a step below this for the changes here, so we can explicitly mark it as native-only?

apps/docs/spec/supabase_js_v2.yml
```diff
@@ -747,6 +747,54 @@ functions:
...
+          import { AppState } from 'react-native'
+
+          // make sure you register this only once!
+          AppState.addEventListener('change', (state) => {
+            if (state === 'active') {
+              supabase.auth.startAutoRefresh()
+            } else {
+              supabase.auth.stopAutoRefresh()
+            }
+          })
```
charislam, line 773:
```suggestion
          })
          ```
```

apps/docs/spec/supabase_js_v2.yml
```diff
@@ -747,6 +747,54 @@ functions:
...
```
charislam, line 797:
```suggestion
          })
          ```
```

charislam (approved):

apps/docs/pages/guides/getting-started/tutorials/with-expo-react-native.mdx
```diff
@@ -169,10 +169,22 @@ Users would be able to sign in with their email and password.
...
 import React, { useState } from 'react'
-import { Alert, StyleSheet, View } from 'react-native'
+import { Alert, StyleSheet, View, AppState } from 'react-native'
 import { supabase } from '../lib/supabase'
 import { Button, Input } from 'react-native-elements'
 
+// Tells Supabase Auth to continuously refresh the session in the background if
+// the app is focused. When this is added, you will continue to receive
+// `onAuthStateChange` events with the `TOKEN_REFRESHED` or `SIGNED_OUT` event
+// if the user's session is terminated. This should only be registered once.
```
dshukertjr, line 179: The term `background` might confuse the reader as it's often used for mobile app life cycle. 
```suggestion
// Tells Supabase Auth to continuously refresh the session automatically if
// the app is in the foreground. When this is added, you will continue to receive
// `onAuthStateChange` events with the `TOKEN_REFRESHED` or `SIGNED_OUT` event
// if the user's session is terminated. This should only be registered once.
```

Also, not sure if the following line is adding any new information here. 

> When this is added, you will continue to receive`onAuthStateChange` events with the `TOKEN_REFRESHED` or `SIGNED_OUT` event

apps/docs/pages/guides/auth/quickstarts/react.mdx
dshukertjr: I'm guessing this edit was meant to be fore this page instead?
https://supabase.com/docs/guides/auth/quickstarts/react-native

PR #20305, opened 2024-01-10
Chore/notifications v2 design fixes

joshenlim:


saltcod:
Looks great! 

Some p minor visual feedback around spacing and alignment: 
1. Something about the square alert icon + the inbox icon here doesn't feel quite right. Tricky I recognize. Would a bright red dot suffice here? 

screenshot-2024-01-10-at-14 10 44

could also do an always-on background here in such cases, or turn the inbox red maybe 
screenshot-2024-01-10-at-14 12 59


2. The Notion notifications panel feels much more comfortable to me — prob because I'm very used to it by now. The spacing here feels right. Wonder can we adjust ours slightly to match.

screenshot-2024-01-10-at-14 18 47

3. Could we show the archive button only on hover like Notion does? That's prob contentious so just a suggestion. The counter would be that people might not know they can archive them, but might be safe to assume devs know about these sorts of interfaces by now well enough to make this assumption.
4. Re: the double icon, could move the warning to the left side which would help with the spacing issue mentioned above. Hiding the Archive button until hover would also clean it up quite a bit 

screenshot-2024-01-10-at-14 26 11


All suggestions!

joshenlim:
> Something about the square alert icon + the inbox icon here doesn't feel quite right. Tricky I recognize. Would a bright red dot suffice here?

This only shows up if you've got a critical notification that needs attention - there'll also be one for when there's a warning notification (in amber), but otherwise it'll just be a number showing the unread count

> The Notion notifications panel feels much more comfortable to me — prob because I'm very used to it by now. The spacing here feels right. Wonder can we adjust ours slightly to match.

Notion's is slightly different cause there's a user icon flushed to the left - but otherwise the padding is quite similar actually

image

> Could we show the archive button only on hover like Notion does? That's prob contentious so just a suggestion. The counter would be that people might not know they can archive them, but might be safe to assume devs know about these sorts of interfaces by now well enough to make this assumption.

Great catch! I forgot to change that back when i was fixing some stuff 😅

> Re: the double icon, could move the warning to the left side which would help with the spacing issue mentioned above. Hiding the Archive button until hover would also clean it up quite a bit

Will hide the archive button till hover

apps/studio/components/layouts/ProjectLayout/LayoutHeader/NotificationsPopoverV2/NotificationsFilter.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const NotificationsFilter = () => {
+  const [open, setOpen] = useState(false)
+  return (
+    <Popover_Shadcn_ modal={false} open={open} onOpenChange={setOpen}>
+      <PopoverTrigger_Shadcn_ asChild>
+        <Button type="text" icon={<Settings2Icon size={14} />} className="px-1" />
+      </PopoverTrigger_Shadcn_>
+      <PopoverContent_Shadcn_ className="p-0 w-52" side="bottom" align="end">
+        Hello
```
alaister, line 13: Hello?

alaister (approved):
LGTM!

PR #20316, opened 2024-01-10
Fix octokit mobile bug

fsansalvadore:

Fixes same problem as #19963.
Adding `octokit` back in since downgrading `@octokit/core` alongside other `@octokit` deps would be more painful.

Will remove `octokit` dependency in the near future once this bug is fixed.

For some reason mobile breaks with latest `octokit` or latest `@octokit/core` versions.


fsansalvadore:
@alaister I couldn't figure out where this problem comes from exactly and how to work around it with latest `@octokit/core` or latest `octokit` deps. 

But if you try navigating the homepage (or any www page) from a mobile device, navbar/footer/js based components don't work properly. If you navigate from this PRs preview they work.

MildTomato (approved):

PR #20335, opened 2024-01-11
Support emptying bucket

joshenlim:
image

### DIFF
--- a/apps/studio/components/ui/ConfirmationModal.tsx
+++ b/apps/studio/components/ui/ConfirmationModal.tsx
@@ -67,7 +67,7 @@ const ConfirmationModal = ({
             disabled={loading || buttonDisabled}
             onClick={onConfirm}
           >
-            {loading ? buttonLoadingLabel : buttonLabel}
+            {loading ? buttonLoadingLabel || buttonLabel : buttonLabel}
           </Button>
         </div>
       }
--- a/apps/studio/components/ui/ConfirmationModal.tsx
+++ b/apps/studio/components/ui/ConfirmationModal.tsx
@@ -67,7 +67,7 @@ const ConfirmationModal = ({
             disabled={loading || buttonDisabled}
             onClick={onConfirm}
           >
-            {loading ? buttonLoadingLabel || buttonLabel : buttonLabel}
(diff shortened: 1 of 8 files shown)


apps/studio/components/ui/ConfirmationModal.tsx
```diff
@@ -67,7 +67,7 @@ const ConfirmationModal = ({
             disabled={loading || buttonDisabled}
             onClick={onConfirm}
           >
-            {loading ? buttonLoadingLabel : buttonLabel}
+            {loading ? buttonLoadingLabel || buttonLabel : buttonLabel}
```
alaister, line 70: Please confirm that this is actually the same thing (not entirely necessary, but who doesn't like a bit of boolean logic?
```suggestion
            {(loading && buttonLoadingLabel) || buttonLabel}
```

alaister (changes requested):
Looks good! Would you also mind adding an API route so this feature also works on self-hosted?

apps/studio/components/ui/ConfirmationModal.tsx
joshenlim: ^ oh i added this so that if its loading, but we dont have buttonLoadingLabel, use buttonLabel as the fallback

alaister: I *think* that's what my one should do too, but since it's clearly more confusing (😅), let's go with the ternary one

joshenlim: ohhhh actually yeah yours is fine tbh - removes the duplicate of calling buttonLabel too haha

joshenlim:
> Looks good! Would you also mind adding an API route so this feature also works on self-hosted?

btw the endpoint is there already [here](https://github.com/supabase/supabase/blob/436bdb10aeae3d1c10a145f859fe151b45474b51/apps/studio/pages/api/storage/%5Bref%5D/buckets/%5Bid%5D/empty.ts) cause delete bucket uses the empty functionality too haha

alaister (approved):
LGTM!

PR #20339, opened 2024-01-11
Removed deprecation dates for Supavisor from docs and pushed users to GH

KevinBrolly:
Better not to have specific dates in the docs (which have now changed) and encourage users to look at the more up-to-date GH discussion.


MildTomato (approved):

MildTomato:
@charislam just tagging you so you know this change was made

PR #20343, opened 2024-01-11
Feat: add log explorer query update and deletion

jordienr:
- Add log explorer query deletion
- Fix table rows alignment (vertical)
- Fix longer descriptions breaking table layout
- Add option to update the query title and description

There's no way to delete or update saved queries in the log explorer

## What is the new behavior?

There's a new dropdown with the option to update or delete a query

## Additional context

CleanShot 2024-01-11 at 14 52 11


joshenlim:
Some nits:
- Could we make the delete query a blocking action? currently the modal closes immediately when we hit delete - the delete button should be in a disabled + loading state while deleting, then close after the deletion is successful (dont close if error)
- Let's use IconMoreVertical instead since most places use that. We should also change the Auth/users page to use IconMoreVertical
- Shift IconMoreVertical button to the right side of the Run button (more consistent with the other pages)
- Unrelated to this PR, but for saving a new logs query, could we also disable the "Save" button while its saving? currently its only in a loading state

alaister (approved):
LGTM!

PR #20359, opened 2024-01-11
chore: change clip-path attribute to clipPath on all SVGs

clarencepenz:

Clear terminal warnings `clip-path`

`clip-path` warnings in terminal

## What is the new behavior?

`clip-path` warnings cleared by changing it to `clipPath`


saltcod:
This is great @clarencepenz and apologies for the delay. Any chance you could resolve the conflicts on these?

PR #20368, opened 2024-01-12
chore: disable truncation on delete button only

alaister:
Follow up to https://github.com/supabase/supabase/pull/20349

- Moves the truncation warning to the delete button so you can still export CSVs
- Removes SqlRowService since it's no longer used 🎉


apps/studio/components/grid/components/header/Header.tsx
```diff
@@ -305,9 +308,25 @@ const RowHeader = ({ table, sorts, filters }: RowHeaderProps) => {
       return setIsExporting(false)
     }
 
+    if (!project) {
+      ui.setNotification({
```
joshenlim, line 312: toast.error

joshenlim (approved):

apps/studio/components/grid/components/header/Header.tsx
```diff
@@ -336,6 +351,12 @@ const RowHeader = ({ table, sorts, filters }: RowHeaderProps) => {
...
   const { selectedRows, editable, allRowsSelected } = state
 
+  useSubscribeToImpersonatedRole(() => {
+    if (allRowsSelected || selectedRows.size > 0) {
+      deselectRows()
+    }
+  })
+
   return (
     <div className="flex items-center gap-6">
```
joshenlim, line 361:
```suggestion
    <div className="flex items-center gap-4">
```

PR #20429, opened 2024-01-15
Chore/update database connection string settings

joshenlim:
Changes are mainly to improve communication regarding the [PGBouncer and IPv4 deprecation notice](https://github.com/orgs/supabase/discussions/17817), and also encourage behaviour to connect to the DB via Supavisor connection pooler instead of direct connections.

## Connection Pooling settings will be shifted to the top of database settings
image
To encourage connection via the pooler - the first "connection string" users see in that page is that of the pooler. Connection string for directly connecting to the database is still available below. Also added links to documentation for the connection pooler.

## Added indication if project is on Supavisor or PGBouncer for connection pooler
image
Added a more visual and obvious badge instead of having to inspect the connection pooler URL to identify the connection pooler

## Added connection string format for Supavisor
image
To inform users that they can connect to their DB via Supavisor with different users / to different databases. The dynamic parameters also have tooltips to describe what they are

## Added PGBouncer and IPv4 deprecation notice in Connection String section
image
In line with the deprecation notice, and so that users still have some information on this after dismissing the deprecation notice banner

## Added small note regarding the pooler mode - to use Session if using prepared statements in database
image


apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
```diff
@@ -65,162 +96,111 @@ const DatabaseConnectionString = () => {
...
+                <AlertDescription_Shadcn_ className="space-y-3">
+                  <p>
+                    We strongly recommend using{' '}
+                    <a
+                      href="https://github.com/supabase/supavisor"
+                      target="_blank"
+                      rel="noreferrer"
+                      className="text-foreground underline underline-offset-[4px] decoration-brand-500 hover:decoration-foreground"
+                    >
+                      Supavisor
```
inian, line 165: instead of linking to Supavisor, we should say "we strongly recommend using connection pooling to connect to the database" and when they click on connection pooling it can take them to the previous section.

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseSettings.tsx
```diff
@@ -97,150 +97,145 @@ const DatabaseSettings = () => {
...
+              </Button>
+            </div>
+          </Alert_Shadcn_>
+        )}
+
+        <Panel
+          title={
+            <div className="w-full flex items-center justify-between">
+              <h5 key="panel-title" className="mb-0">
+                Connect to your database directly
```
inian, line 154: I think we should combine the `Connect to your database directly` and `Connection string` panels. It should be clear to the user that there are 2 ways of connecting to the db - pooler and direct. That way we can add the `resolves to ipv4` and the deprecation notice to the direct connection panel. Right now if user just sees the `Connect to your database directly` panel, they may not see the deprecation notice below or think it may not apply to them.

apps/studio/components/interfaces/Settings/Database/ConnectionPooling.tsx
```diff
@@ -120,15 +67,26 @@ const FormSchema = z.object({
...
-
-const MainConfig = ({ projectRef, poolingInfo }: MainConfigProps) => {
+export const ConnectionPooling = () => {
   const { ui } = useStore()
+  const { ref: projectRef } = useParams()
   const { project } = useProjectContext()
 
+  const { data: addons } = useProjectAddonsQuery({ projectRef })
+  const computeInstance = addons?.selected_addons.find((addon) => addon.type === 'compute_instance')
+  computeInstance?.variant.identifier
```
alaister, line 77: this just hanging out?
```suggestion
```

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
```diff
@@ -1,28 +1,57 @@
...
+export const DatabaseConnectionString = () => {
   const router = useRouter()
+  const { project: projectDetails } = useProjectContext()
   const { ref: projectRef, connectionString } = useParams()
   const telemetryProps = useTelemetryProps()
   const state = useDatabaseSelectorStateSnapshot()
 
-  const readReplicasEnabled = useFlag('readReplicas')
+  // [Joshen] TODO this needs to be obtained from BE as 26th Jan is when we'll start - projects will be affected at different rates
+  const resolvesToIpV6 = false // Number(new Date()) > Number(dayjs.utc('01-26-2024', 'MM-DD-YYYY').toDate())
```
alaister, line 49: Just confirming this should be hardcoded to `false` for now?

alaister (approved):
Code-wise, LGTM!

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
```diff
@@ -65,162 +96,111 @@ const DatabaseConnectionString = () => {
...
+                    We strongly recommend using{' '}
+                    <a
+                      href="https://github.com/supabase/supavisor"
+                      target="_blank"
+                      rel="noreferrer"
+                      className="text-foreground underline underline-offset-[4px] decoration-brand-500 hover:decoration-foreground"
+                    >
+                      Supavisor
+                    </a>{' '}
+                    to connect to your database. You'll only need to change the connection string
```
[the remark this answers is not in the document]
joshenlim: yup - we havent had a hard rule for shorthands yet, although open to not have them

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseSettings.tsx
joshenlim: image

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
joshenlim: yup - i'm checking with BE whether there'll be some sort of project level flag, but we can update this logic in a separate PR so the warnings can go through first

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
joshenlim: image

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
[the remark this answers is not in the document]
inian: cc @charislam if we should have something in our guidelines here.

inian (approved):

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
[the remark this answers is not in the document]
charislam: I like shorthand, it sounds friendlier and less intimidating
We don't have anything written officially in our style guide (can add it) but my general opinion is: prefer shorthand, unless it's a strong warning (like a dangerous action), in which case it sounds stronger if you spell it out: `this doesn't prevent you from` vs. `this does not prevent you from...`

PR #20447, opened 2024-01-15
chore: migrate tabs component to new NavMenu component

jordienr:

~~Swaps the tabs component in /settings/vault to use the shadcn component~~

This started as a refactor to use the Shadcn tabs component but since most of the tabs in the dashboard require navigating to different urls it would require us to add a router.push in the onChange event which isn't great for accessibility. After some feedback from Jonny and trying to adapt the shadcn component I've done a few things:

- I've decided to create a new simpler component that looks like the tabs component, but is just a list of links. 

- I've added the correct styles to the Shadcn Tabs component in case we want to use it for client-side tabs somewhere in the dashboard. 

- I've also used the new component in `/settings/vault/secrets`.

---

## Before and after. 

Here's the old use of Tabs
```jsx
// secrets.tsx page
<Tabs
  size="small"
  type="underlined"
  activeId="keys"
  onChange={(id: any) => {
    if (id === 'secrets') router.push(`/project/${ref}/settings/vault/secrets`)
  }}
  >
    <Tabs.Panel id="secrets" label="Secrets Management" />
    <Tabs.Panel id="keys" label="Encryption Keys">
      <EncryptionKeysManagement />
    </Tabs.Panel>
</Tabs>
```
With new component
```jsx
<NavMenu className="mb-4" label={'Vault menu'}>
  <NavMenuItem active={props.activeTab === 'secrets'}>
    <Link href={`/project/${props.projRef}/settings/vault/secrets`}>Secrets Management</Link>
  </NavMenuItem>
  <NavMenuItem active={props.activeTab === 'keys'}>
    <Link href={`/project/${props.projRef}/settings/vault/keys`}>Encryption Keys</Link>
  </NavMenuItem>
</NavMenu>
```
---

## End result
CleanShot 2024-01-17 at 17 53 18


saltcod:
some light mode love required here :^)

screenshot-2024-01-15-at-17 00 26

apps/studio/pages/project/[ref]/settings/vault/secrets.tsx
```diff
@@ -35,19 +35,22 @@ const VaultSettingsSecrets: NextPageWithLayout = () => {
...
-        <Tabs
-          size="small"
-          type="underlined"
-          activeId="secrets"
-          onChange={(id: any) => {
-            if (id === 'keys') router.push(`/project/${ref}/settings/vault/keys`)
+        <Tabs_Shadcn_
+          defaultValue="secrets"
+          value={router.query.tab as string}
+          onValueChange={(e) => {
```
Comment (saltcod) [inline apps/studio/pages/project/[ref]/settings/vault/secrets.tsx:41]: https://ui.shadcn.com/docs/components/tabs
the tabs here are arrow-key-accessible.

I realized that the Vault page doesn't do this b/c of the router.push. Wonder if router.replace would work / makes sense here?

ivasilov (dismissed):
There's a slight flicker when switching the tabs, I'm guessing because you haven't migrated the other page https://github.com/supabase/supabase/blob/688f2b6ac9db3d7c7de290dd1a34d25aecfb1029/apps/studio/pages/project/%5Bref%5D/settings/vault/keys.tsx which also should use the new tabs component.

packages/ui/src/components/NavMenu/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const NavMenu = ({
+  children,
+  className,
+  label,
+}: PropsWithChildren<{
+  className?: string
+  label: string
+}>) => {
+  return (
+    <nav aria-label={label} dir="ltr" role="menu" className={cn('flex border-b', className)}>
```
MildTomato, line 13: you could do


```tsx
interface NavProps extends HTMLAttributes<HTMLDivElement> {
  // Add any specific props for <nav> element here
  asChild?: boolean;
}

const NavMenu: React.ForwardRefRenderFunction<HTMLDivElement, NavProps> = (
  { ...props },
  ref) => {
return (
    <nav dir="ltr" role="menu" {...props} className={cn('flex border-b', props.className)}>
//..
```

so then all the props are available..

```tsx
<NavMenu ariaLabel={"nav menu"} dir={"override-dir"} role={"override-role"} className={"bg-red"}>
//..
```

MildTomato: {...props} is in that weird order so it doesn't overwrite the className

ivasilov (approved):

PR #20453, opened 2024-01-15
feat: add retry mechanism for webhooks

mansueli:

This PR allows people to use retry on their webhooks in the database, so it is more robust/ more fault tolerant.

image

## What is the new behavior?

## Additional context


apps/studio/components/interfaces/Database/Hooks/HTTPRequestFields.tsx
```diff
@@ -123,6 +123,14 @@ const HTTPRequestFields = ({
             type="number"
             actions={<p className="text-foreground-light pr-2">ms</p>}
           />
+          <Input
+            id="max_retries"
+            name="max_retries"
+            label="Number of retries"
+            labelOptional="Between 0 and 5"
+            type="number"
```
laktek, line 131: Better to use `type="range"` and set the min/max? (I know there's a check on submission) https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range

mansueli: I think that would work only if @supabase/frontend has a better component for it that displays also the current value. As I think it should be consistent for both the retries and the timeout picker. 

Seemed messy/less clear when I tried it:
image


Code used in the screenshot:
````html
          <Input
            id="timeout_ms"
            name="timeout_ms"
            label="Timeout"
            labelOptional="Between 1000ms to 5000ms"
            type="range"
            min="1000"
            max="5000"
            step="100"
            actions={<p className="text-foreground-light pr-2"></p>}
          />
          <Input
            id="max_retries"
            name="max_retries"
            label="Number of retries"
            labelOptional="Between 0 to 5"
            min="0"
            max="5"
            step="1"
            type="range"
            actions={<p className="text-foreground-light pr-2"></p>}
          />
````

File: docker/volumes/db/webhooks.sql [hunk 2]
```diff
@@ -205,4 +205,116 @@ BEGIN;
   ALTER function supabase_functions.http_request() SET search_path = supabase_functions;
   REVOKE ALL ON FUNCTION supabase_functions.http_request() FROM PUBLIC;
   GRANT EXECUTE ON FUNCTION supabase_functions.http_request() TO postgres, anon, authenticated, service_role;
+  INSERT INTO supabase_functions.migrations (version) VALUES ('20240115163000_add_retry_to_http_request');
+  CREATE OR REPLACE FUNCTION supabase_functions.http_request()
```
sweatybridge, line 209: It would be nice to update this function in place on line 28. This file is only run on fresh start of database so we don't need to worry about backwards compatibility with existing migrations.

docker/volumes/db/webhooks.sql
```diff
@@ -25,81 +25,115 @@ BEGIN;
...
-        RAISE EXCEPTION 'url argument is missing';
-      END IF;
+        DECLARE
+          local_request_id bigint;
+          payload jsonb;
+          url text := TG_ARGV[0]::text;
+          method text := TG_ARGV[1]::text;
+          headers jsonb DEFAULT '{}'::jsonb;
+          params jsonb DEFAULT '{}'::jsonb;
+          timeout_ms integer DEFAULT 1000;
```
steve-chavez, line 41: This is a bad default, even with 2000 (current default on pg_net) we're having trouble and I'll be changing it on https://github.com/supabase/pg_net/issues/86#issuecomment-1898428584.

I suggest leaving the default to pg_net and maybe recommend it to be higher than 5 seconds.

mansueli: I was keeping the same default that was already there. Happy to change it.

steve-chavez: How about no DEFAULT? Leaving it as NULL should let `pg_net` pick its default.

mansueli: That is okay with me.

docker/volumes/db/webhooks.sql
```diff
@@ -25,81 +25,113 @@ BEGIN;
...
-        (TG_RELID, TG_NAME, request_id);
-
-      RETURN NEW;
-    END
-  $function$;
+              IF local_request_id IS NOT NULL THEN
+                SELECT (response).status_code::integer
+                  INTO status_code
+                  FROM net._http_collect_response(local_request_id);
+                IF status_code < 400 THEN
```
sweatybridge, line 113:
```suggestion
                IF status_code < 500 THEN
```

I think it's common to retry only 5xx errors. Because 4xx indicates a problem with client side request that typically can't be fixed by retry.

Perhaps we can also make it configurable with an array of retry-able status codes as argument.

sweatybridge (approved):
Just some nitpicks.

joshenlim:
Do not merge yet - needs a corresponding change in the BE API to support the new retry param

PR #20466, opened 2024-01-16
fix: table editor not refresh when change schema 

### LINKED ISSUE #20437: Table editor not refreshed when switching schema
# Bug report

## Describe the bug

When I'm switching schemas in the table editor the table doesn't get refreshed and still shows the table from the previously selected schema. This leads to confusion when trying to analyse some data or finding an issue but you are looking on the wrong table, especially if they look very similar. 

## To Reproduce

Steps to reproduce the behaviour:

1. Go to the table editor. Which should show all tables of 'public' schema
2. Select a table so that it shows the data in the editor.
3. Select a different schema from drop down
4. It still shows the table of public schema in the editor

## Expected behavior

The editor should be cleared and prompt the 'Select a table from the navigation panel on the left to view its data, or create a new one.' message when switching schema.

tranhoangvuit:

fixes: https://github.com/supabase/supabase/issues/20437

When we change the schema in the left menu of the table editor, it does not reset the selected table in the grid view, which can cause confusion for the user.

## What is the new behavior?

When we change the schema in the left menu of the table editor, it will reset selected table.

## Additional context

### DIFF
--- a/apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
+++ b/apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
@@ -34,8 +34,10 @@ import {
 } from 'ui'
 import { useProjectContext } from '../ProjectLayout/ProjectContext'
 import EntityListItem from './EntityListItem'
+import { useRouter } from 'next/router'
 
 const TableEditorMenu = () => {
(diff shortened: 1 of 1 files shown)


alaister (approved):
LGTM! Thanks, @tranhoangvuit!

PR #20491, opened 2024-01-16
PP events

fsansalvadore:


apps/www/pages/pricing/index.tsx
```diff
@@ -53,6 +56,11 @@ export default function IndexPage() {
     }
   }, [asPath])
 
+  const telemetryProps = useTelemetryProps()
+  const sendTelemetryEvent = async (event: TelemetryEvent) => {
```
MildTomato, line 60: maybe this can be more global ?

apps/www/pages/pricing/index.tsx
```diff
@@ -292,7 +300,15 @@ export default function IndexPage() {
                           )}
                         </div>
                         <a href={plan.href}>
```
MildTomato, line 302: Probably should be like this.
otherwise it will get focussed twice when keyboard tabbing.

```js
<Button asChild>
  <a>
  //
```

MildTomato (approved):
left some comments. approved though :D

apps/www/pages/pricing/index.tsx
fsansalvadore: I'll address this when fixing events ✌️

PR #20498, opened 2024-01-16
Improve Billing Breakdown section

ecktoteckto:

Improve Billing Breakdown section in Studio

- Compute Credits are not shown
- Confusion about what "Projected Costs" means
- Lack of tooltips that describe things better

## What is the new behavior?
- "Current Costs" added right above "Projected Costs". These are the costs accumulated from the beginning of the billing cycle up to now
- Compute Credits are shown now and placed at the end of the fee items instead of in between
- Tooltips added for "Projected Costs", "Current Costs", "Compute Credits" and "Compute Hours"

new-billing-breakdown

## Additional info

There's still an issue with the calculation of the Projected Costs. Will be tackled with a different PR.

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
@@ -198,6 +198,34 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
             <tfoot>
               <tr>
                 <td className="py-4 text-sm font-medium">
+                  <span className="mr-2">Current Costs</span>
+                  <Tooltip.Root delayDuration={0}>
+                    <Tooltip.Trigger>
+                      <IconInfo size={12} strokeWidth={2} />
+                    </Tooltip.Trigger>
+                    <Tooltip.Portal>
+                      <Tooltip.Content side="bottom">
+                        <Tooltip.Arrow className="radix-tooltip-arrow" />
+                        <div
+                          className={[
+                            'rounded bg-alternative py-1 px-2 leading-none shadow',
+                            'border border-background',
+                          ].join(' ')}
+                        >
+                          <span className="text-xs text-foreground">
+                            Costs accumulated from the beginning of the billing cycle up to now.
+                          </span>
+                        </div>
+                      </Tooltip.Content>
+                    </Tooltip.Portal>
+                  </Tooltip.Root>
+                </td>
+                <td className="py-4 text-sm text-right font-medium" colSpan={3}>
+                  ${upcomingInvoice?.amount_total ?? '-'}
+                </td>
+              </tr>
+              <tr>
+                <td className="py-4 text-sm font-medium">
                   <span className="mr-2">Projected Costs</span>
                   <Tooltip.Root delayDuration={0}>
                     <Tooltip.Trigger>
--- a/apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
@@ -178,16 +178,17 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
                               billingMetricUnit(fee.usage_metric) &&
                               ` (${billingMetricUnit(fee.usage_metric)})`}
                           </span>
-                          {feeTooltipData.map(
-                            (tooltipData) =>
-                              fee.usage_metric?.startsWith(tooltipData.identifier) && (
-                                <InvoiceTooltip
-                                  text={tooltipData.text}
-                                  linkRef={tooltipData.linkRef}
-                                  key={tooltipData.identifier}
-                                />
-                              )
-                          )}
+                          {(() => {
+                            const matchingTooltipData = feeTooltipData.find((it) =>
+                              fee.usage_metric?.startsWith(it.identifier)
+                            )
+                            return matchingTooltipData ? (
+                              <InvoiceTooltip
+                                text={matchingTooltipData.text}
+                                linkRef={matchingTooltipData.linkRef}
+                              ></InvoiceTooltip>
+                            ) : null
+                          })()}
                         </td>
                         <td className="py-2 pr-4 text-sm text-right tabular-nums max-w-[100px]">
                           {fee.usage_original
--- a/apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
@@ -3,14 +3,35 @@ import clsx from 'clsx'
 import AlertError from 'components/ui/AlertError'
 import ShimmeringLoader from 'components/ui/ShimmeringLoader'
 import { useOrgUpcomingInvoiceQuery } from 'data/invoices/org-invoice-upcoming-query'
-import { useMemo, useState } from 'react'
+import React, { useMemo, useState } from 'react'
 import { Button, Collapsible, IconChevronRight, IconInfo } from 'ui'
 import * as Tooltip from '@radix-ui/react-tooltip'
 import { billingMetricUnit, formatUsage } from '../helpers'
+import Link from 'next/link'
 
 export interface UpcomingInvoiceProps {
   slug?: string
 }
+
+interface TooltipData {
+  identifier: string
+  text: string
+  link?: {
+    href: string
+    text: string
+  }
+}
+
+const tooltips: TooltipData[] = [
+  {
+    identifier: 'COMPUTE',
+    text: 'Every project is a dedicated server and database. For every hour your project is active, it incurs compute costs based on the instance size of your project. Paused projects do not incur compute costs.',
+    link: {
+      href: 'https://supabase.com/docs/guides/platform/org-based-billing#usage-based-billing-for-compute',
+      text: 'Compute Hours',
+    },
+  },
+]
 
 const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
   const {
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
```diff
@@ -23,20 +39,26 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
...
+  const computeCredits = useMemo(() => {
+    return (upcomingInvoice?.lines || []).find((item) =>
+      item.description.startsWith('Compute Credits')
+    )
+  }, [upcomingInvoice])
+
   const fixedFees = useMemo(() => {
     return (upcomingInvoice?.lines || [])
-      .filter((item) => !item.breakdown)
+      .filter((item) => item !== computeCredits && (!item.breakdown || !item.breakdown.length))
```
kevcodez, line 50: Now that you are separating out the credits credits anyway, do we still need to adjust the latter statement

`(!item.breakdown || !item.breakdown.length)`

and do we still need the API change?

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
```diff
@@ -23,20 +39,26 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
...
         // Prorations should be below regular usage fees
         return Number(a.proration) - Number(b.proration)
       })
-  }, [upcomingInvoice])
+  }, [upcomingInvoice, computeCredits])
 
   const feesWithBreakdown = useMemo(() => {
     return (upcomingInvoice?.lines || [])
-      .filter((item) => item.breakdown?.length)
+      .filter((item) => item !== computeCredits && item.breakdown?.length)
```
kevcodez, line 59: Compute credits will never have a breakdown, so the first expression is not needed

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
```diff
@@ -146,12 +172,23 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
...
                           </span>
+                          {(() => {
+                            const matchingTooltipData = feeTooltipData.find((it) =>
+                              fee.usage_metric?.startsWith(it.identifier)
+                            )
+                            return matchingTooltipData ? (
+                              <InvoiceTooltip
+                                text={matchingTooltipData.text}
+                                linkRef={matchingTooltipData.linkRef}
+                              ></InvoiceTooltip>
```
kevcodez, line 189: Unless the component has children, we usually go for self-closing tags so <InvoiceTooltip .... />

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
```diff
@@ -195,33 +232,39 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
                 </Collapsible>
               ))}
 
+            {computeCredits && (
```
kevcodez, line 235: Thinking about it, or rather looking at it, it does look a little odd that after all these broken down metrics we have one last item without a breakdown. I don't have too hard feelings on this, but wdyt?

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
```diff
@@ -195,33 +232,39 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
...
+                <tr className="border-b">
+                  <td className="py-2 text-sm max-w-[200px]">
+                    <span className="mr-2">{computeCredits.description}</span>
+                    <InvoiceTooltip
+                      text="Paid plans come with $10 in Compute Credits to cover one Starter instance or parts of any other instance. Compute Credits are given to you not only for the first month but for every month while you are on a paid plan."
+                      linkRef="https://supabase.com/docs/guides/platform/org-based-billing#compute-credits"
+                    />
+                  </td>
+                  <td className="py-2 text-sm text-right" colSpan={3}>
+                    ${computeCredits.amount}
```
kevcodez, line 246: Is `amount` properly formatted with thousand separators and currency?

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
```diff
@@ -195,33 +232,39 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
                 </Collapsible>
               ))}
 
+            {computeCredits && (
+              <tbody>
+                <tr className="border-b">
+                  <td className="py-2 text-sm max-w-[200px]">
+                    <span className="mr-2">{computeCredits.description}</span>
+                    <InvoiceTooltip
+                      text="Paid plans come with $10 in Compute Credits to cover one Starter instance or parts of any other instance. Compute Credits are given to you not only for the first month but for every month while you are on a paid plan."
```
kevcodez, line 241:
```suggestion
                      text="Paid plans come with $10 in Compute Credits to cover one Starter instance or parts of any other instance. Compute Credits are given to you every month and do not stack up while you are on a paid plan."
```

> not only for the first month but for every month

sounds a bit repeated and there's easier ways to put it, no hard feelings though

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
```diff
@@ -233,4 +276,40 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
...
+  return (
+    <Tooltip.Root delayDuration={0}>
+      <Tooltip.Trigger>
+        <IconInfo size={12} strokeWidth={2} />
+      </Tooltip.Trigger>
+      <Tooltip.Portal>
+        <Tooltip.Content side="bottom">
+          <Tooltip.Arrow className="radix-tooltip-arrow" />
+          <div
+            className={[
```
kevcodez, line 289: You could also just put it into one className

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
```diff
@@ -233,4 +276,40 @@ const UpcomingInvoice = ({ slug }: UpcomingInvoiceProps) => {
...
+            ].join(' ')}
+          >
+            <span className="text-xs text-foreground">
+              <p>
+                {text}{' '}
+                {linkRef && (
+                  <Link
+                    href={linkRef}
+                    target="_blank"
+                    className="transition text-brand hover:text-brand-600"
```
kevcodez, line 301: Could also do a simple `underline` here

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
ecktoteckto: - removed the second part of the condition
-  will delete the API branch

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
ecktoteckto: Good catch. Wasn't the case.
Using Javascript's `Intl.NumberFormat`now brings one change though. The minus sign for a negative amount is positioned before the dollar sign now instead of after the dollar sign. But that seems to be the standard way of doing it (and that's also how Stripe does it).
Once you say "OK" to this I'll use this helper function for other pages (e.g. usage breadown) as well.

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
ecktoteckto: See my comment in Slack.

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
```diff
@@ -312,4 +309,16 @@ const InvoiceTooltip = ({ text, linkRef }: { text: string; linkRef?: string }) =
   )
 }
 
+const formatToCurrency = (amount: number | undefined | null): string | null => {
+  if (amount === undefined || amount === null) {
+    return null
+  } else {
+    return Intl.NumberFormat('en-US', {
```
kevcodez, line 316: Initializing Intl.NumberFormat is quite expensive, given we re-use it quite a lot + all the re-rendering, let's move

```ts
Intl.NumberFormat('en-US', {
      style: 'currency',
      currency: 'USD',
      minimumFractionDigits: 0,
    })
```

to a constant outside of this function

kevcodez: Could also extract this NumberFormat + the formatToCurrency function to some utils, as we can re-use that nicely on other pages

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
kevcodez: Makes sense, commented below regarding extracting the function and initializing

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/UpcomingInvoice.tsx
ecktoteckto: Done

kevcodez (approved):

PR #20511, opened 2024-01-17
docs: Add `signInWithSSO` to Flutter docs.

dshukertjr:
- Adds `signInWithSSO()` method to the Flutter docs
- Adds Dart code sample to setting up SSO guide.

### DIFF
--- a/apps/docs/spec/supabase_dart_v2.yml
+++ b/apps/docs/spec/supabase_dart_v2.yml
@@ -319,34 +319,18 @@ functions:
         isSpotlight: true
         code: |
           ```dart
-            // You can extract the user's email domain and use it to trigger the
-            // authentication flow with the correct identity provider.
-
-            const { data, error } = await supabase.auth.signInWithSSO({
-              domain: 'company.com'
-            })
-
-            if (data?.url) {
-              // redirect the user to the identity provider's authentication flow
-              window.location.href = data.url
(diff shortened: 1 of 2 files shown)


apps/docs/spec/supabase_dart_v2.yml
```diff
@@ -305,6 +305,33 @@ functions:
...
           final String? oAuthToken = session?.providerToken;
           ```
+  - id: sign-in-with-sso
+    title: 'signInWithSSO()'
+    notes: |
+      - Before you can call this method you need to [establish a connection](/docs/guides/auth/sso/auth-sso-saml#managing-saml-20-connections) to an identity provider. Use the [CLI commands](/docs/reference/cli/supabase-sso) to do this.
+      - If you've associated an email domain to the identity provider, you can use the `domain` property to start a sign-in flow.
+      - In case you need to use a different way to start the authentication flow with an identity provider, you can use the `providerId` property. For example:
+          - Mapping specific user email addresses with an identity provider.
+          - Using different hints to identity the identity provider to be used by the user, like a company-specific page, IP address or other tracking information.
```
charislam, line 315:
```suggestion
          - Using different hints to identify the correct identity provider, like a company-specific page, IP address or other tracking information.
```

charislam (approved):

PR #20526, opened 2024-01-17
chore: update pg_upgrade docs around incompatible extension versions; remove plv8 object exception

pcnc:
* Due to more complex upgrade paths for TimescaleDB and plv8, upgrading from versions prior to TimescaleDB 2.9.1 and plv8 3.1.5 is not currently supported
* Previous limitation of upgrading projects with plv8 enabled has been removed

### DIFF
--- a/apps/docs/pages/guides/platform/migrating-and-upgrading-projects.mdx
+++ b/apps/docs/pages/guides/platform/migrating-and-upgrading-projects.mdx
@@ -74,28 +74,14 @@ When upgrading, the Supabase platform will "right-size" your disk based on the c
 pg_upgrade does not support upgrading of databases containing reg\* data types referencing system OIDs.
 If you have created any objects that depend on the following extensions, you will need to recreate them after the upgrade.
 
-Extensions registering system OID-specific reg\* data types:
+#### Extensions
 
-- `plv8`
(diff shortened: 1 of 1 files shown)


apps/docs/pages/guides/platform/migrating-and-upgrading-projects.mdx
```diff
@@ -74,28 +74,13 @@ When upgrading, the Supabase platform will "right-size" your disk based on the c
 pg_upgrade does not support upgrading of databases containing reg\* data types referencing system OIDs.
 If you have created any objects that depend on the following extensions, you will need to recreate them after the upgrade.
 
-Extensions registering system OID-specific reg\* data types:
+#### Extensions
 
-- `plv8`
-- `plcoffee`
-- `plls`
+pg_upgrade does not currently support upgrading of databases using extensions older than the following versions:
```
pcnc, line 79: Alternative wording, though it might prove confusing for users skimming the docs, leading them to believe that projects using those specific versions can't be upgraded, rather than the reverse

```suggestion
pg_upgrade does not currently support upgrading of databases using extensions prior to following versions:
```

charislam (approved):

PR #20556, opened 2024-01-18
fix: failed to retrieve postgres version in setting

### LINKED ISSUE #20402: "Failed to retrieve Postgres version  Error: <!DOCTYPE html>" when running locally as Studio settings tries to call Management API endpoints
# Bug report

- [+] I confirm this is a bug with Supabase, not with my own application. 
- [+] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

After deploying Supabase via docker-compose, the admin UI works, but PostgreSQL DB is semi-broken. Several problems occur, maybe all due to lack of correct PostgreSQL deployment, maybe due to something else:
1. Navigating to http://service-url.xxx:8000/project/default/settings/infrastructure - pages load but has this error in "Service Versions" block: Failed to retrieve Postgres version Error: <!DOCTYPE html><html lang="en"><head><meta charSet="utf-8"/><title>Supabase</title><meta name="viewport" content="initial-scale=1.0, width=device-width"/> ... and a lot of HTML code afterwards (note: github truncates DOCTYPE html blocks, check the screenshot below to see the actual error)
2. Navigating to http://service-url.xxx:8000/project/default/settings/addons - similar error: Failed to retrieve project addons also with the same DOCTYPE html code in it as in point 1.
3. http://service-url.xxx:8000/project/default/settings/database - Connection Pooling - Failed to retrieve pooling configuration. Try refreshing your browser, but if the issue persists, please reach out to us via support.
4. supabase-rest and supabase-edge-functions containers are up, but they doesn't show "healthy" status (maybe it's ok)
5. supabase-studio shows "unhealthy status", but in the "docker compose logs studio" I don't see any problems. Says ready in 431 ms, Local & Network addresses looks unknown to me, but I guess that's expected as they are internal docker addresses?
6. supabase-analytics contains a lot of this error: [error] Cluster size is 1 but expected 3
7. Weirdness with PostgreSQL:

I can connect to PSQL with `psql -h 127.0.0.1 -p 5432 -d postgres -U postgres:
psql (16.1, server 15.1 (Ubuntu 15.1-1.pgdg20.04+1))

`And I see 2 databases there - pg_default and pg_global. But when I'm trying to connect to either of them, I'm getting connection to server at "127.0.0.1", port 5432 failed: FATAL:  database "pg_default" does not exist

I have tried deploying into 2 different servers

## To Reproduce

1. Used a default docker-deployment procedure from help.
2. Adjusted the keys, passwords, logins, and deployment URL addresses in .env, nothing else.
3. Run `docker-compose up -d`
4. Seen the errors mentioned above.

## Expected behavior

Working Supabase and PSQL.

## Screenshots

image

## System information

- Local OS: Garuda Linux
- Deployment OSes: 
- 1. Linux 6.6.5-arch1-1 (Arch Linux rolling)
- 2. Linux 6.6.2-arch1-1 (Arch Linux rolling)
- Version of supabase-js: supabase/studio:20240101-8e4a094
- Version of Node.js: v21.4.0

tranhoangvuit:

fixes: https://github.com/supabase/supabase/issues/20402

When we access `project/default/settings/infrastructure`, it returns an error page instead of the current version.

## What is the new behavior?

We have created an API to return the current app version. It's hardcoded for now, but we might find a better way to handle this in the future.

## Additional context


apps/studio/pages/api/constants.ts
KevinBrolly, line 1: @tranhoangvuit I am not sure we want these as constants, would we not have to then update them each time we done a new release so that self-hosted stays up to date? Also, you list goTrue and postgres versions as constants but I cannot see where they are actively used. I admit also I am not familiar with self-hosted supabase as much as the platform, so I could be mistaken here.

tranhoangvuit: Thank @KevinBrolly for your review.
I'm considering whether we should also show it on a self-hosted setup. Could you let me know how we handle it on Supabase Cloud, as shown in this image?

image

Regarding the PostgreSQL version, I think we can call the metadata [here ](https://github.com/supabase/postgres-meta/blob/master/src/server/routes/config.ts) to display the version. However, since GoTrue doesn't support this, maybe we shouldn't display the version here or only do so for self-hosted setups."

tranhoangvuit: Thank @KevinBrolly, sorry for late update, I changed to use redirect to default page instead of showing version of postgres in self-host.

ivasilov (approved):

ivasilov (approved):

PR #20559, opened 2024-01-18
docs: add server-side auth guide for nextjs

charislam:


apps/docs/pages/guides/auth/server-side/nextjs.mdx
```diff
@@ -0,0 +1,947 @@
...
+
+Add a matcher so the middleware doesn't run on routes that don't access Supabase.
+
+<Admonition type="danger">
+
+The middleware uses `supabase.auth.getSession()`.
+
+For performance, this function doesn't revalidate the user's Auth token except when the cookie's token expires.
+
+If you extend your middleware to process user data, replace `getSession` with `getUser` to ensure that user data is always revalidated.
```
hf, line 192: We need to be much clearer and more paranoid in wording about this, as folks won't get how serious it can be. Especially since Next.js and this whole SSR is brand new and people are not used to thinking about cookies as they've been out of fashion for 15 years now.

```suggestion
Be careful when protecting pages. Your user's session is stored in Cookies, which means that anyone can spoof values in this header. Avoid using `supabase.auth.getSession()` to protect pages and instead use `supabase.auth.getUser()`.

Do not blindly trust the return values of `supabase.auth.getSession()` inside server components, as the returned object could be controlled by an attacker. It's safe to trust `supabase.auth.getUser()` because it sends a request to Supabase Auth which validates the access token.
```

charislam: Thanks, good point. I've reworded it a little 👇🏼 , then added it to the Server Component step as well, and changed the middleware to use `getUser` just in case

```
Be careful when protecting pages. The server gets the user session from the cookies, which can be spoofed by anyone.

Always use `supabase.auth.getUser()` to protect pages and user data.

_Never_ trust `supabase.auth.getSession()` inside Server Components. It isn't guaranteed to revalidate the Auth token.

It's safe to trust `getUser()` because it sends a request to the Supabase Auth server every time to revalidate the Auth token.
```

saltcod: this section was great btw — I did not know this

saltcod (approved):

kangmingtay:
@charislam we might need to add a template error page in one of the steps because currently, if the user runs into an error, they'll see a "404 page not found" error rather than the actual error

charislam:
> @charislam we might need to add a template error page in one of the steps because currently, if the user runs into an error, they'll see a "404 page not found" error rather than the actual error

on second thought, I'm going to add a short comment about "make an error page, this currently 404s" (worded a bit better). not being lazy here 😛 I want to balance being comprehensive with not feeling overwhelming with how much you have to do, and I think creating an error page should be relatively self-explanatory 🤞🏼

PR #20571, opened 2024-01-19
feat: limit sign-in password length to 72

hf:
Anything over 72 is not acceptable.

### DIFF
--- a/apps/studio/components/interfaces/SignIn/SignInForm.tsx
+++ b/apps/studio/components/interfaces/SignIn/SignInForm.tsx
@@ -16,7 +16,7 @@ const signInSchema = object({
   email: string().email('Must be a valid email').required('Email is required'),
   password: string()
     .required('Password is required')
-    .max(72, 'Good job with the password, but it should be 72 chars or less'),
+    .max(72, 'Password must be at most 72 characters'),
 })
 
 const SignInForm = () => {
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/SignIn/SignInForm.tsx
```diff
@@ -14,7 +14,9 @@ import { Button, Form, Input } from 'ui'
 
 const signInSchema = object({
   email: string().email('Must be a valid email').required('Email is required'),
-  password: string().required('Password is required'),
+  password: string()
+    .required('Password is required')
+    .max(72, 'Password must be at most 72 characters'),
```
alaister, line 19: @hf Should we be validating here on sign in? Will gotrue choke if it receives a string longer than 72 characters?

alaister, line 19: @hf Should we be validating here on sign in? Will gotrue choke if it receives a string longer than 72 characters?

joshenlim, line 19:
```suggestion
    .max(72, 'Password cannot exceed 72 characters'),
```

apps/studio/lib/schemas.ts
```diff
@@ -5,5 +5,10 @@ import YupPassword from 'yup-password'
 YupPassword(yup)
 
 export const passwordSchema = yup.object({
-  password: yup.string().password().required().label('Password'),
+  password: yup
+    .string()
+    .password()
+    .required()
+    .max(72, 'Password must be at most 72 characters')
```
joshenlim, line 12:
```suggestion
    .max(72, 'Password cannot exceed 72 characters')
```

apps/studio/components/interfaces/SignIn/SignInForm.tsx
hf: Hmm good point this will block people from signing in.

alaister (approved):
LGTM!

PR #20553, opened 2024-01-18
docs: Add caution callout to Apple secret generation form

xjialu:
## What kind of change does this PR introduce?

Docs Update: Added temporary message before Safari secret generation is fixed

PR #20578, opened 2024-01-19
docs: mention that apple key generator does not work on safari

hf:
For some reason it doesn't work and I'm not able to quickly see why.


charislam (approved):

charislam:
Also addressed in https://github.com/supabase/supabase/pull/20553 which is waiting for response from the contributor
If you merge first, do me a favor and close the other one? 🙇🏼

PR #20580, opened 2024-01-19
fix: add schema name to url query params

jordienr:

Adds the schema name as a query param to the button that redirects to the policies page.

it always sends you to the "public" schema

## What is the new behavior?

It will load the correct schema when it redirects.

## BEFORE:
CleanShot 2024-01-19 at 16 27 26

## AFTER
CleanShot 2024-01-19 at 16 30 11


monicakh (approved):
LGTM!

PR #20631, opened 2024-01-22
feat: branching statuses first pass

alaister:

feature

## What is the new behaviour?

Screenshot 2024-01-23 at 01 10 21


apps/studio/components/interfaces/BranchManagement/BranchStatusBadge.tsx
```diff
@@ -0,0 +1,45 @@
...
+  if (status === 'ACTIVE_HEALTHY') {
+    return null
+  }
+
+  const isUnhealthy = UNHEALTHY_STATUES.includes(status)
+  const isWaiting = WAITING_STATUSES.includes(status)
+
+  return (
+    <Badge color={isUnhealthy ? 'red' : 'slate'} className="flex items-center gap-1.5">
+      {isUnhealthy && <AlertTriangleIcon size={12} />}
```
MildTomato, line 38: use heroicon ?

MildTomato: Screenshot 2024-01-22 at 10 15 33 PM

MildTomato: We already use this in Docs for callouts

alaister: do you has link pls?

apps/studio/components/interfaces/BranchManagement/BranchStatusBadge.tsx
```diff
@@ -0,0 +1,45 @@
...
+type Status = BranchData['status']
+
+export interface BranchStatusBadgeProps {
+  status: Status
+}
+
+const UNHEALTHY_STATUES: Status[] = ['ACTIVE_UNHEALTHY', 'INIT_FAILED', 'UNKNOWN']
+const WAITING_STATUSES: Status[] = ['COMING_UP', 'GOING_DOWN', 'PAUSING', 'RESTORING', 'UPGRADING']
+
+const STATUS_TO_LABEL: Record<Status, string> = {
```
joshenlim, line 14: can we use PROJECT_STATUS actually? seems like there's quite an overlap

alaister: There will be more branching related statuses added here in future

apps/studio/components/interfaces/BranchManagement/BranchStatusBadge.tsx
joshenlim: i created one in NotificationsPopover.constants.ts - maybe lets just bring that out into components/ui 
either that or we bring it into the UI library (which sounds better imo)

unless docs is already using a component from the ui library (but last i checked i saw that it was just SVG)

apps/studio/components/interfaces/BranchManagement/BranchStatusBadge.tsx
MildTomato: resolving

apps/studio/data/branches/branch-query.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export async function getBranch({ id }: BranchVariables, signal?: AbortSignal) {
+  if (!id) throw new Error('id is required')
+
+  const { data, error } = await get(`/v1/branches/{branch_id}`, {
+    params: { path: { branch_id: id } },
+    signal,
+  })
+  if (error) throw error
```
joshenlim, line 19: use the new handleError from data/fetchers

apps/studio/components/interfaces/BranchManagement/BranchStatusBadge.tsx
joshenlim, line 38: I've created a `<WarningIcon>` - can use in place

joshenlim:
Possible to show axis legends based on the selected columns? The x axis probably should be dynamic based on the column type (if possible) - e.g shouldnt be a date if the values are numbers

image

joshenlim (approved):

PR #20638, opened 2024-01-22
docs(Auth): update signup behavior for existing confirmed user

842u:
According to https://github.com/supabase/supabase-js/issues/947, i think it's worth to mentioning about this behavior in docs to clarify when signup returns error and when returns obfuscated user.


charislam (approved):

GaryAustin1:
@charislam 
Are we sure this documentation change is correct?
image
Seems if you have email confirm set it does not matter what phone confirm is to return the fake user.
At least one user was trying by the new docs and gets what to me is the expected result of fake user with email confirm set.

PR #20642, opened 2024-01-22
Customize price plan info for partner Fly

ecktoteckto:

Customizes price plan info depending on whether it's a Fly subscription or not

The side panel for plan updates doesn't take into account the limited feature set for Fly subscriptions

## What is the new behavior?

The price plan info shows the actual feature set and number of free orgs for Fly subscriptions
plan-update-sidepanel-fly

### DIFF
--- a/packages/shared-data/plans.ts
+++ b/packages/shared-data/plans.ts
@@ -7,6 +7,7 @@ export interface PricingInformation {
   priceLabel?: string
   priceMonthly: number | string
   warning?: string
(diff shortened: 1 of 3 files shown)


packages/shared-data/plans.ts
ecktoteckto, line 1: Went for a pragmatic approach here. Probably no need to optimize for different partners yet.

kevcodez (approved):

PR #20648, opened 2024-01-23
fix: close command menu after navigating

charislam:


gregnr:
Hey, confirming we want to close the menu after clicking an external (new tab) like like GH discussions?

charislam:
> Hey, confirming we want to close the menu after clicking an external (new tab) like like GH discussions?

yeah I think it's kind of weird behavior if it _doesn't_ close, since choosing where you want to go should be the end of your interaction with the search

I guess you might want to open several pages at once, but I think it's better to keep the UX as consistent as possible, so I'm going to err on the side of closing

gregnr (approved):
@charislam makes sense, LGTM!

PR #20658, opened 2024-01-23
Chore/update connection string settings kaizen

joshenlim:


ivasilov (commented):
Screenshot 2024-01-23 at 11 50 35
This wording should be probably be `switch to...`. Connect to a different user sounds off. Otherwise looks ok.

ivasilov (approved):

PR #20709, opened 2024-01-24
fix: include integrations in our search indexes

charislam:
Include integrations content in our search indexes so the pages are findable by FTS and embedding search.

They are ignored by RAG for AI answer generation because of problems we've had in the past with the returned answers.


gregnr (approved):
Nice! 🙌

PR #20728, opened 2024-01-25
Use Dropdown component instead of Command for pooler mode selection

joshenlim:


apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
```diff
@@ -66,98 +63,67 @@ export const UsePoolerCheckbox = ({
...
-                              </Command_Shadcn_>
-                              <Separator />
-                              <p
-                                className="px-3 py-2 text-xs text-foreground-light hover:text-foreground cursor-pointer transition"
+                          <DropdownMenu>
+                            <DropdownMenuTrigger asChild>
+                              <Button
+                                type="outline"
+                                className="py-0.5 pr-1.5"
+                                iconRight={<IconChevronDown strokeWidth={1.5} />}
```
MildTomato, line 71: strokeWidth 1

we don't use anything but 1 for icons

apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
```diff
@@ -66,98 +63,67 @@ export const UsePoolerCheckbox = ({
...
+                              >
+                                <span className="capitalize">Mode: {poolingMode}</span>
+                              </Button>
+                            </DropdownMenuTrigger>
+                            <DropdownMenuContent side="bottom" align="start">
+                              <DropdownMenuItem
+                                key="transaction"
+                                className="text-foreground"
```
MildTomato, line 79: text-foreground

isn't this handled ?

apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
```diff
@@ -66,98 +63,67 @@ export const UsePoolerCheckbox = ({
...
+                                onClick={() => onSelectPoolingMode('transaction')}
+                              >
+                                <Tooltip.Root delayDuration={0}>
+                                  <Tooltip.Trigger asChild>
+                                    <span className="w-full">Transaction mode</span>
+                                  </Tooltip.Trigger>
+                                  {data.pool_mode === 'session' && (
+                                    <Tooltip.Portal>
+                                      <Tooltip.Content side="right">
+                                        <Tooltip.Arrow className="radix-tooltip-arrow" />
```
MildTomato, line 89: Think we should get tooltip in UI lib, we keep repeating the same code everywhere

apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
```diff
@@ -66,98 +63,67 @@ export const UsePoolerCheckbox = ({
...
+                                        <div
+                                          className={[
+                                            'rounded bg-alternative py-1 px-2 leading-none shadow',
+                                            'border border-background w-[380px]',
+                                          ].join(' ')}
+                                        >
+                                          <span className="text-xs text-foreground">
```
MildTomato, line 96: text-light

apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
```diff
@@ -66,98 +63,67 @@ export const UsePoolerCheckbox = ({
...
+                              >
+                                <Tooltip.Root delayDuration={0}>
+                                  <Tooltip.Trigger asChild>
+                                    <span className="w-full">Transaction mode</span>
+                                  </Tooltip.Trigger>
+                                  {data.pool_mode === 'session' && (
+                                    <Tooltip.Portal>
+                                      <Tooltip.Content side="right">
+                                        <Tooltip.Arrow className="radix-tooltip-arrow" />
+                                        <div
```
MildTomato, line 90: this div is random, shouldn't all these classes be in the `Tooltip.Content` ?

joshenlim: this isn't a UI library component so its unstyled - although yes we should move it to the UI lib as you mentioned in the other comment

apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
joshenlim: let's do this when we move the tooltip to the UI library - all the tooltips across the dashboard is using text-foreground

apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
joshenlim: i see that we've already got the shadcn ui tooltip - we'll slowly replace all the tooltips with that in subsequent PRs

apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
joshenlim: by default its text-foreground-light which looks a bit weird to me, but i can remove this so we dont have custom CSS going on

image

apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
joshenlim: just a pref though but doesnt strokeWidth 1 feel too thin?

apps/studio/components/interfaces/Settings/Database/UsePoolerCheckbox.tsx
MildTomato: yea but tooltip.content takes css classes. don't need to add more divs in it

MildTomato: if you've done this for every other one it doesn't matter

joshenlim: ^ ahh yeah - been doing this for every other one, no worries will be resolved when we move all to shadcn's tooltips
i'll do that next week

alaister (approved):
LGTM!

PR #20758, opened 2024-01-26
docs: unquote current setting for visual consistnecy

J0:

docs change

## Additional context


J0:
Actually probably clearer as is

PR #20765, opened 2024-01-26
feat: revamped instance size selection

kevcodez:
Screenshot 2024-02-01 at 16 48 42

Immediately pick an instance size without having to click through categories. Also some prep work for upcoming changes.

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/DowngradeModal.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/DowngradeModal.tsx
@@ -1,3 +1,4 @@
+import { ProjectInfo } from 'data/projects/projects-query'
 import { OrgSubscription, ProjectAddon } from 'data/subscriptions/types'
 import { PricingInformation } from 'shared-data'
 import { Alert, IconAlertOctagon, IconMinusCircle, IconPauseCircle, Modal } from 'ui'
@@ -8,6 +9,7 @@ export interface DowngradeModalProps {
   subscription?: OrgSubscription
   onClose: () => void
   onConfirm: () => void
+  projects: ProjectInfo[]
 }
 
 const ProjectDowngradeListItem = ({ projectAddon }: { projectAddon: ProjectAddon }) => {
@@ -39,7 +41,26 @@ const DowngradeModal = ({
   subscription,
   onClose,
   onConfirm,
+  projects,
 }: DowngradeModalProps) => {
+  // Filter out the micro addon as we're dealing with that separately
+  const previousProjectAddons =
+    subscription?.project_addons.flatMap((projectAddons) => {
+      const addons = projectAddons.addons.filter((it) => it.variant.identifier !== 'ci_micro')
+      if (!addons.length) {
+        return []
+      } else {
+        return {
+          ...projectAddons,
+          // Overwrite addons, filtered out the micro addon
+          addons,
+        }
+      }
+    }) || []
+
+  const hasInstancesOnMicro = projects.some((project) => project.infra_compute_size === 'micro')
+  const downgradingToNano = subscription?.nano_enabled === true
+
   return (
     <Modal
       size="large"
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/Organization/BillingSettings/Subscription/DowngradeModal.tsx
```diff
@@ -39,7 +41,26 @@ const DowngradeModal = ({
...
+        return {
+          ...projectAddons,
+          // Overwrite addons, filtered out the micro addon
+          addons,
+        }
+      }
+    }) || []
+
+  const hasInstancesOnMicro = projects.some((project) => project.infra_compute_size === 'micro')
+  const downgradingToNano = subscription?.nano_enabled === true
```
ivasilov, line 62: What does this flag mean? Is `nano_enabled` on the current subscription or the subscription you're downgrading to?

ivasilov: Also, tried to test this flow but `nano_enabled` doesn't exist yet on the Subscription API.

kevcodez: This is forwards compatible, the flag will be exposed in an upcoming API PR - if the flag is missing, this will be false, which is correct

kevcodez: With `nano_enabled` when downgrading from paid -> free and having micro instances, it will downgrade the instances to nano

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
```diff
@@ -52,6 +53,11 @@ const PlanUpdateSidePanel = () => {
     PermissionAction.BILLING_WRITE,
     'stripe.subscriptions'
   )
+  const { data: allProjects } = useProjectsQuery()
```
kevcodez, line 56: We should really have API support for filtering by org slug

apps/studio/components/interfaces/Settings/Addons/ComputeInstanceSidePanel.tsx
```diff
@@ -123,32 +101,79 @@ const ComputeInstanceSidePanel = () => {
...
+        // @ts-ignore API types it as Record<string, never>
+        meta: {
+          cpu_cores: 2,
+          cpu_dedicated: false,
+          memory_gb: 1,
+          baseline_disk_io_mbs: 87,
+          max_disk_io_mbs: 2085,
+          connections_direct: 60,
+          connections_pooler: 200,
+        } as ProjectAddonVariantMeta,
```
thebengeu, line 149: Looks like there are 3 instances of these, or a subset thereof, should there be a constant?

apps/studio/components/interfaces/Settings/Addons/ComputeInstanceSidePanel.tsx
```diff
@@ -123,32 +101,79 @@ const ComputeInstanceSidePanel = () => {
...
+      })
+    }
+
+    return computeOptions
+  }, [availableAddons, hasMicroOptionFromApi])
+
+  console.log({ availableOptions })
```
thebengeu, line 156: Remove?

apps/studio/components/interfaces/Settings/Addons/Addons.tsx
```diff
@@ -155,11 +183,23 @@ const Addons = () => {
...
       {isSuccess && (
         <>
+          {selectedProject?.infra_compute_size === 'nano' && subscription?.plan.id !== 'free' && (
+            <ScaffoldContainer className="mt-4">
+              <Alert_Shadcn_ variant="default">
+                <IconInfo strokeWidth={2} />
+                <AlertTitle_Shadcn_>Free compute upgrade to Micro</AlertTitle_Shadcn_>
+                <AlertDescription_Shadcn_>
+                  Paid Plans include a free upgrade to Micro compute. Your project is ready to
+                  upgrade for no additional charges.
```
thebengeu, line 193: Since technically it's "$10 of Compute Credits to cover one Micro instance or parts of any other [Compute Add-On](https://supabase.com/docs/guides/platform/compute-add-ons)", as per https://supabase.com/docs/guides/platform/org-based-billing#compute-credits, should the wording make it clearer that that there are other options to use the $10 of Compute Credits besides upgrading to Micro compute?

thebengeu (changes requested):

apps/studio/components/interfaces/Settings/Addons/Addons.tsx
kevcodez: nano and micro are both $10 a month, so it's up to the user to do the free upgrade here - we're just trying to advertise the free upgrade here. Were you thinking about adding something along the lines of "

Paid Plans include a free upgrade to Micro compute. Your project is ready to upgrade for no additional charges. Paid plans come with $10 in Compute Credits to cover one Micro instance or parts of any other [Compute Add-on](https://...)?

ivasilov (approved):

apps/studio/components/interfaces/Settings/Addons/ComputeInstanceSidePanel.tsx
thebengeu: I may be missing something, why not `meta: INSTANCE_MICRO_SPECS`, since in the end it's still typed as `Record<string, never>`? Relatedly, it may make sense for `ProjectAddonVariantMeta` to extend `Partial<InstanceSpecs>`, since any instance specs could be considered to be metadata for compute addons, but I understand why we may want to keep them decoupled too.

thebengeu (approved):

PR #20774, opened 2024-01-26
feat: Move some components into `ui-patterns`

ivasilov:
Migrated bunch of components to a new `ui-patterns` package. It should be used for more complex components, which depend on big libraries (`reactflow` for example) or are built using multiple `ui` components.

All components should be directly imported (`ui-packages/ThemeToggle` for example). Few other components (`Command`, `Markdown`, `Codeblock`) are left in `ui` because of collision [with this PR](https://github.com/supabase/supabase/pull/20903). They'll be migrated later.


saltcod:
weird that this link doesn't work on this branch 
screenshot-2024-02-02-at-13 59 02

saltcod (approved):

PR #20777, opened 2024-01-26
refactor: move auth guides into content directory

charislam:
Begin the process of moving our MDX files into their own `content` directory.

Fixed a few minor bugs re: ToC and tabs not rerendering consistently on page navigation. (The ToC thing wasn't a problem before the refactor, the tabs thing is a problem on prod.)

Moved MDX files can't import their own components, so everything they require needs to be back in the `component` prop for `mdx-remote`'s `serializer`. Cleaned this up a bit and lazy-loaded heavy/rare stuff. Also, the `component` prop doesn't take arbitrary objects (only actual components), so imported data has to be wrapped in a component.

Analyzed total bundle size and first load data transfer to make sure this isn't causing performance regression:

| | Total bundle JS size | First load data transfer (home page) | First load data transfer (Auth redirect URLs page) |
| - | - | - | - |
| Before | 4.07 MB gzipped | 882 kB | 757 kB |
| After | 3.47 MB gzipped | 852 kB | 723 kB |


saltcod (approved):

apps/docs/content/guides/auth/column-level-security.mdx
```diff
@@ -151,8 +149,4 @@ authenticated;
 ## Considerations when using column-level privileges
 - If you turn off a column privilege you won't be able to use that column at all.
 - All operations (insert, update, delete) as well as using `select *` will fail.
-
-
-export const Page = ({ children }) => <Layout meta={meta} children={children} />
-
-export default Page
+```
```
charislam, line 152:
```suggestion
```

PR #20766, opened 2024-01-26
docs: fix code snippets for expo tutorial

NekodRider:
## What kind of change does this PR introduce?

Fix #20764 

## What is the new behavior?
1. No more errors from TypeScript compiler.
2. Replace `react-native-document-picker` with `expo-document-picker`.

PR #20796, opened 2024-01-27
docs: React Native tutorial avatar implementation rework

bdotsamir:

This PR aims to fix the current implementation of how the developer uploads avatars to Supabase storage when bootstrapping their React Native project with Expo.

Currently, the developer utilizes the package `react-native-document-picker`. While this is not *technically* incorrect- the user *could* select an image from their filesystem- it's not one that provides the most optimal UX.

## What is the new behavior?

This PR replaces the `react-native-document-picker` package with `expo-image-picker`, which provides an easier access to the end user's photos to upload as an avatar for their account. I believe it is safe to assume that the developer is using expo for this project, since this page explicitly instructs the developer to bootstrap their React Native project with Expo. It also, since this is not a drop-in replacement, changes the way that the `uploadAvatar()` function is implemented to account for the differences between the packages, such as cancellation detection. Check the diff for more.

## Additional context

There were a couple different methods that I tried here, namely fetching the image returned by expo by its URI and converting the image to a Blob format. This did not appear to work, as [Blobs are not natively supported by React Native](https://github.com/react-native-community/discussions-and-proposals/issues/109). I ended up using array buffers and uploading the image data that way, as you can see with this PR.

Also, and this could be a point of discussion, is there a particular reason as to why these functions are just regular ol' `const`s and not utilizing `React.useCallback()`?

### DIFF
--- a/apps/docs/pages/guides/getting-started/tutorials/with-expo-react-native.mdx
+++ b/apps/docs/pages/guides/getting-started/tutorials/with-expo-react-native.mdx
@@ -451,10 +451,10 @@ photos and videos.
 
 ### Additional dependency installation
 
-You will need a file picker that works on the environment you will build the project for, we will use react-native-document-picker in this example.
+You will need an image picker that works on the environment you will build the project for, we will use `expo-image-picker` in this example.
 
 ```bash
-expo install react-native-document-picker
+expo install expo-image-picker
 ```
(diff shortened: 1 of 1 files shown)


bdotsamir:
Closed, merged into #20766. Thanks neko!

PR #20830, opened 2024-01-29
feat: Show a badge if the pgbouncer hasn't been removed.

ivasilov:


apps/studio/pages/api/database/[ref]/pg-bouncer.ts
```diff
@@ -0,0 +1,33 @@
...
+      res.setHeader('Allow', ['GET'])
+      res.status(405).json({ data: null, error: { message: `Method ${method} Not Allowed` } })
+  }
+}
+
+const handleGetAll = async (req: NextApiRequest, res: NextApiResponse) => {
+  const host = req.query.host
+
+  try {
+    const url = `http://${host}:6543`
```
Comment (inian) [inline apps/studio/pages/api/database/[ref]/pg-bouncer.ts:22]: We need to do some validation here to make sure we arent pinging arbitrary hosts and only the pgbouncer hosts belonging to the user/project, else it could lead to SSRF vuln. Given the project ref, we can construct the host directly on the api instead of relying on a separate host query parameter

inian (approved):

alaister (approved):
LGTM!

PR #20840, opened 2024-01-29
Reset state and set enable form actions for add role sidebar.

SantoshPisini:

Bug fix

Resolves #20839 

## What is the new behavior?

Using the form action buttons are not disabled & the value is being reset.

## Additional context

`hasChanges={true}` is set to `true` as to enable form action buttons, and if the user clicks on 'Save' with an empty value, form validation is triggered, hence I feel this is valid as the user will be blocked until they enter a valid value.

[create_role_sidebar_fix.webm](https://github.com/supabase/supabase/assets/39261271/c42fd77f-3bd2-43e8-9ac3-e236a7bfe4dd)


SantoshPisini:
@saltcod Please review :)

SantoshPisini:
@encima Is there any tag that I need to add for this PR to review?

encima:
nope, you have done the hard work, we just need to review it!

apps/studio/components/interfaces/Database/Roles/CreateRolePanel.tsx
```diff
@@ -87,8 +87,8 @@ const CreateRolePanel = ({ visible, onClose }: CreateRolePanelProps) => {
           <FormActions
             form={formId}
             isSubmitting={isCreating}
-            hasChanges={form.formState.isDirty}
-            handleReset={() => form.reset()}
+            hasChanges={true}
```
joshenlim, line 90: quick question - this doesnt look right - it should still be `form.formState.isDirty` right?

SantoshPisini: `hasChanges={true}` is set to `true ` as to enable form action buttons, and if the user clicks on 'Save' with an empty value, form validation is triggered, hence I feel this is valid as the user will be blocked until they enter a valid value.

joshenlim (approved):

apps/studio/components/interfaces/Database/Roles/CreateRolePanel.tsx
```diff
@@ -71,7 +71,7 @@ const CreateRolePanel = ({ visible, onClose }: CreateRolePanelProps) => {
 
   const handleClose = () => {
     onClose()
-    form.reset()
+    form.reset(initialValues)
```
joshenlim, line 74: this isn't necessary i think

```suggestion
    form.reset()
```

joshenlim: sorry! was wrong here - reverting the change

PR #20844, opened 2024-01-29
fix: get tokenizer working again for Supabase AI

charislam:
AI responses suddenly broke because of WASM compatibility issues

`@dqbd/tiktoken` has now been split into two packages, one of which is `js-tiktoken`, which doesn't require WASM and should work better on Edge environments
switching to that seems to fix things


gregnr (approved):
Nice, the new JS-only lib should make things much simpler 👍

PR #20851, opened 2024-01-30
Feat/highlight new rls assistant

saltcod:
screenshot-2024-01-29-at-21 13 12

Highlight the new RLS Assistant from the old one


apps/studio/components/interfaces/Auth/Policies/PolicySelection.tsx
```diff
@@ -58,6 +72,33 @@ const PolicySelection = ({
             onClick={onViewEditor}
           />
         </div>
+        <div className="bg-surface-200 rounded-md p-4 flex items-center  gap-2 text-sm">
```
MildTomato, line 75: we don't have a component for this?

apps/studio/components/interfaces/Auth/Policies/PolicySelection.tsx
```diff
@@ -58,6 +72,33 @@ const PolicySelection = ({
...
+          <Info size={16} strokeWidth={2} />
```
MildTomato, line 76: We use hero icon for Info, Warning and Destructive icons

joshenlim: jfyi icons are here - https://github.com/supabase/supabase/blob/feat/highlight-new-rls-assistant/packages/ui/src/components/Admonition/index.tsx#L21

they aren't exported - will bring them out

apps/studio/components/interfaces/Auth/Policies/PolicySelection.tsx
joshenlim: opted for alert shadcn component instead

alaister (approved):
LGTM!

PR #20865, opened 2024-01-30
chore: update themes and add dark-new theme set

MildTomato:

- new theme support added.
- new theme called 'deep-dark' added

## This feature is feature flagged

- under `enableExperimentalTheme`
- if someone chooses `"deep-dark"` theme and then uses this app without these new theme settings, all the styling will break.

## What is the new behavior?

- background and border colors replaced so there is consistency between themes.
- updated theme picker
- big radio button component added to `./ui` 

**Two types of picker have been implemented.**

### Single Theme selection

This is the same as before, but a new 'deep-dark' theme has been added
also, fancy new radio buttons 🤷 

Screenshot 2024-02-12 at 7 46 01 AM

this setting is in https://github.com/supabase/supabase/blob/chore/start-adding-new-dark-theme/apps/studio/components/interfaces/Account/Preferences/ThemeSettings.tsx

### Resolved theme selection

This is picking wether you want a permanent theme, or to use system.
but when using system, the theme resolves to the preferred dark or light theme you prefer.
(this is how GitHub works)

This is waiting on [upstreaming some changes](https://github.com/pacocoursey/next-themes/pull/251) to npm:next-themes.
Although I have (@MildTomato) already forked and released an updated version to support this if we want to go this route

these settings are here:
https://github.com/supabase/supabase/blob/chore/start-adding-new-dark-theme/apps/studio/components/interfaces/Account/Preferences/ThemeSettingsWithPreferredTheme.tsx
we can replace the one above with this when ready if we want to.

Screenshot 2024-02-12 at 7 41 17 AM
Screenshot 2024-02-12 at 7 41 22 AM

### new [big radio button](https://github.com/supabase/supabase/blob/chore/start-adding-new-dark-theme/packages/ui/src/components/shadcn/ui/radio-group.tsx/#L44-L112) in `./ui`

- `children` is where the image is. accepts anything.
- the radio indicator can be hidden with `showIndicator` set to `false`
- the whole this is a RadioGroupPrimitive.Item which you can pass anything to, for example, maybe setting the content padding with `"p-0"` in className
- you can use `"group-data-[state=checked]:"` to set anything when active
- you can use `"group-hover:"` to set anything for hover
- wrap these in `<form>` if you want the <input> to be inserted as well (see radix docs)

Screenshot 2024-02-12 at 8 12 07 AM


MildTomato:
- [ ] There is an issue with active states on icon side bar

ivasilov (approved):

ivasilov:
The SQL editor looks broken. The RLS editor looks like it's using the old dark.
Screenshot 2024-02-12 at 09 31 26
Screenshot 2024-02-12 at 09 32 18

MildTomato:
@ivasilov 

> The SQL editor looks broken.

This seems like the current behavior. It flashes white before setting the dark mode.

apps/www/styles/index.css
```diff
@@ -42,7 +43,7 @@ html {
 }
 
 body {
-  @apply bg-background;
+  @apply bg-background text-foreground;
```
fsansalvadore, line 46: @MildTomato I just added this `text-foreground` to the body selector to fix some inheritance styling issues.

Screenshot 2024-02-14 at 16 35 05

fsansalvadore:
@MildTomato fyi you can still pick the `deep-dark` theme via the command menu.

fsansalvadore (changes requested):
@MildTomato you can still pick the deep-dark theme via the command menu.
Prob need to filter the incoming themes how you did in the ThemeToggle.

MildTomato:
> @MildTomato you can still pick the deep-dark theme via the command menu.
Prob need to filter the incoming themes how you did in the ThemeToggle.

I know it's a bit dumb manually filtering all of these areas but I didn't want to have to wait on a feature flag call. we can remove all the filtering later.

fsansalvadore (approved):
Looks good ✌️

PR #20880, opened 2024-01-30
Update configuration.mdx | added more examples

TheOtherBrian1:
Added more examples to the change timeout section. Updated examples to use the authenticated role instead of authenticator role(gotrue server)

docs update

Does not show users how to quickly update global timeout. Also provides an example that tells people to update the timeout for the authenticator role (gotrue) instead of the authenticated role(registered users)
## What is the new behavior?

Provides 2 new examples:
- how to inspect timeouts for roles
- how to update global timeout

Modified one example to use the authenticated role instead of authenticator.

## Additional context

I noticed a few users were getting confused and creating tickets when their authenticated role did not update (they were updating the authenticator role by mistake). This is supposed to prevent that issue.

Note, there is a bug with the CLI that prevents users from updating their global timeouts. I recommend withholding from approving this pull request until it is fixed.

### DIFF
--- a/apps/docs/pages/guides/database/postgres/configuration.mdx
+++ b/apps/docs/pages/guides/database/postgres/configuration.mdx
@@ -21,6 +21,7 @@ FROM pg_roles
 WHERE rolname IN (
   'anon',
   'authenticated',
+  'authenticator',
   'postgres',
   'service_role'
 );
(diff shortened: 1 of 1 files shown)


apps/docs/pages/guides/database/postgres/configuration.mdx
```diff
@@ -12,14 +12,47 @@ Postgres provides a set of sensible defaults for you database size. In some case
...
 The timeout values were picked as a reasonable default for the majority of use-cases, but can be modified using the [`alter role`](https://www.postgresql.org/docs/current/sql-alterrole.html) statement:
 
+<Admonition type="note" label="Self hosting">
+When updating either the `anon` or `authenticated` role, set the `authenticator` role to the greater timeout value between the two.
+</Admonition>
+
 ```sql
--- For API users
+alter role authenticated set statement_timeout = '15s';
 alter role authenticator set statement_timeout = '15s';
```
mansueli, line 56: Small nit but because of the precedence, I would put authenticator first as it takes precedence.

Also maybe keep a comment but put lines making it explicit that you would need to set them e.g

```
-- For API users, you need to set authenticator and then the authenticated value: 
alter role authenticator set statement_timeout = '16s';
alter role authenticated set statement_timeout = '15s';
```

PR #20886, opened 2024-01-30
Fix some things in edge function unit testing docs

carlosdp:
There's a couple things in these docs that don't match up with what is actually written, some imports in the code that aren't used, and an out-of-date Deno CLI command structure being used in this doc.

This just updates it to be more accurate.


charislam (approved):

joshenlim:
The changes in this PR seems to have already been applied to the latest `docs/content/guides/functions/unit-test.mdx` file, will hence be closing this PR

PR #20907, opened 2024-01-31
feat: query performance improvements

jordienr:

Improvements to the Query Performance page 

Cannot sort or search queries in query performance table.

## What is the new behavior?

- You can search by query or role
- You can sort the results
- You can expand the query to read it inside the table 
- Search, sort and preset state is persisted in the URL in case you want to share it with someone in your team
- The page will only request the data it needs (right now, it requests the 3 presets on first load)
- Adding more filters, pagination should be easier with the new `useQueryPerformanceQuery`

## Additional context

CleanShot 2024-02-02 at 17 23 02@2x


saltcod:
we should maybe truncate the first column as well? 
screenshot-2024-02-02-at-13 46 38

screenshot-2024-02-02-at-13 47 13

super minor, but a few rows of skeleton might work better here?
screenshot-2024-02-02-at-13 50 43

The table can only be sorted by latency, which seems to indicate that that's really important. But the word "latency" doesn't show up anywhere else on the page. Leaves with a feeling of "ummmm what's latency? Is it just total time?"

screenshot-2024-02-02-at-13 51 27

Could just be a toggle with "Latency (or Total Time) + an arrow up/down" ? 
screenshot-2024-02-02-at-13 56 39

saltcod:
also, shame were not yet using shadcn data table component, which has column visibility
https://ui.shadcn.com/docs/components/data-table
(not suggesting you rewrite this!)

saltcod:
Nice! 

Couple of tiny things: 

1. Can we align these? 
screenshot-2024-02-05-at-14 36 23

2. Since that code is always sql, can we add <simplecodeblock> or something to format it? 
3. I think we should do a standard arrow up/down icon thing on this. Not the end of the world, but I think it'll clean it & simplify it  
screenshot-2024-02-05-at-14 37 32

apps/studio/components/interfaces/Reports/ReportQueryPerformanceTableRow.tsx
```diff
@@ -0,0 +1,77 @@
...
+    </div>
+  )
+}
+
+const ReportQueryPerformanceTableRow = ({ sql, colSpan, children }: Props) => {
+  const [expanded, setExpanded] = React.useState(false)
+  const { resolvedTheme } = useTheme()
+  const isDarkTheme = resolvedTheme?.includes('dark')!
+  SyntaxHighlighter.registerLanguage('sql', sqlTheme)
+  const monokaiTheme = monokaiCustomTheme(isDarkTheme)
```
ivasilov, line 39: Is this variable some sort of global which is used for Monaco?

apps/studio/components/interfaces/Reports/ReportQueryPerformanceTableRow.tsx
```diff
@@ -0,0 +1,77 @@
...
+  sql: string
+  colSpan: number
+  children: React.ReactNode
+}
+
+const QueryActions = ({ sql, className }: { sql: string; className: string }) => {
+  if (sql.includes('insufficient privilege')) return null
+
+  return (
+    <div className={[className, 'flex justify-end items-center'].join(' ')}>
```
ivasilov, line 21: Use `cn` here.

packages/ui/package.json
```diff
@@ -100,6 +100,7 @@
     "@storybook/addon-styling": "^1.3.7",
     "@storybook/react": "^7.5.3",
     "@storybook/react-webpack5": "^7.5.3",
+    "@tailwindcss/line-clamp": "^0.4.4",
```
ivasilov, line 103: This should be added as a dependency to `packages/config`, not to `ui`.

ivasilov (approved):

PR #20932, opened 2024-02-01
feat: Enable Upgrade to IPv4 button

ivasilov:


apps/studio/components/interfaces/Settings/Addons/Addons.tsx
```diff
@@ -101,6 +103,10 @@ const Addons = () => {
 
   const meta = computeInstance?.variant?.meta as ProjectAddonVariantMeta | undefined
 
+  const canEnableIPv4 = ['concurrent-ipv6', 'ipv6'].includes(
```
inian, line 106: can we restrict to ipv6 only for now?

inian (approved):

joshenlim (approved):

PR #20957, opened 2024-02-02
fix: error with quotes in search results

charislam:
Search results with double quotes in the title need to be escaped or cmdk's attempt to query for selectors will throw an error for invalid selector.

```
DOMException: Failed to execute 'querySelector' on 'Element': '[cmdk-item=""][data-value="__forcemount__--realtime "concurrent peak connections" quota-item-index-0"]' is not a valid selector.
```

Also add an error boundary so errors in command menu don't crash the entire site.

image


jordienr (approved):
lgtm!

PR #20978, opened 2024-02-03
fix: mobile search width

### LINKED ISSUE #17437: width of search card in docs needs to fixed on the mobile view
# Bug report

## Describe the bug

when we write a long query on the search card in the docs on the mobile view the width of the card increases and goes out of the screen.

https://github.com/supabase/supabase/assets/144774379/055a6303-b7ab-4779-8530-b04073820e3b

## Expected behavior

The width of the search card should not go out of the screen when we write a long query in the input.

Hallidayo:

Supabase Docs

On a mobile the search window expands the full width

## What is the new behavior?

The modal now has a max width:

Screenshot 2024-02-03 at 18 41 42

## Additional context

Closes #17437

### DIFF
--- a/packages/ui/src/components/Command/Command.utils.tsx
+++ b/packages/ui/src/components/Command/Command.utils.tsx
@@ -69,6 +69,7 @@ export const CommandDialog = ({ children, onKeyDown, page, ...props }: CommandDi
         '!border-overlay/90',
         'transition ease-out',
         'place-self-start mx-auto top-24',
+        'max-w-[80%] xs:max-w-[50%]',
         animateBounce ? 'scale-[101.5%]' : 'scale-100'
       )}
     >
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Command/Command.utils.tsx
```diff
@@ -69,6 +69,7 @@ export const CommandDialog = ({ children, onKeyDown, page, ...props }: CommandDi
         '!border-overlay/90',
         'transition ease-out',
         'place-self-start mx-auto top-24',
+        'max-w-[80%] xs:max-w-[50%]',
```
fsansalvadore, line 72: Can we have the max-width be same width as the content below? 
So something like `max-w-[calc(100vw-40px)]` (if that's the correct padding, need to double check)

Hallidayo: @fsansalvadore - Just pushed a commit for the above, upped from `40px` to `60px`

charislam (approved):

packages/ui/src/components/Command/Command.utils.tsx
charislam: could do something like a `max-w-[min(100vw-40px,600px)]`?

sorry if that's dodgy tailwind, wrote it off the top of my head

PR #21012, opened 2024-02-05
chore: Fix octokit vulnerability

ivasilov:


fsansalvadore:
The problem with this for me was that mobile breaks for some reason.

image

Still didn't quite figure out why. @ivasilov

fsansalvadore (approved):

ivasilov:
We discussed it with Francesco and decided to merge this to see if it causes issues for anyone else.

PR #21036, opened 2024-02-06
docs: Updated Flutter reference docs for filters to be more aligned with js docs

dshukertjr:

The overall theme of this PR is aligning the Flutter reference docs's filter section with the JS reference docs. 

- Better example for `contains()`, `containedBy()` and `overlaps()` on the Flutter docs that is more aligned with the JS docs. 
- Add data sources to the filters on Flutter reference docs.
- Remove `with update`, `with delete` and `with rpc` examples as they are redundant, and JS docs don't have them. 
- A small typo fix on js reference docs.

### DIFF
--- a/apps/docs/spec/supabase_dart_v2.yml
+++ b/apps/docs/spec/supabase_dart_v2.yml
@@ -1920,7 +1920,7 @@ functions:
     description: |
       Filters allow you to only return rows that match certain conditions.
 
-      Filters can be used on `select()`, `update()`, and `delete()` queries.
+      Filters can be used on `select()`, `update()`, `upsert()`, and `delete()` queries.
 
       If a Database function returns a table response, you can also apply filters.
     examples:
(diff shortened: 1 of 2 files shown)


apps/docs/spec/supabase_dart_v2.yml
```diff
@@ -2512,84 +2527,85 @@ functions:
...
+              countries (id, name)
+            values
+              (1, 'Afghanistan'),
+              (2, 'Albania'),
+              (3, 'Algeria');
+            ```
+          response: |
+            ```json
+            {
+              "data": [
```
charislam, line 2553: the other sample responses directly return the `data` array without the wrapping object, which one is correct?

charislam (approved):

apps/docs/spec/supabase_dart_v2.yml
dshukertjr: Thanks for catching this one. Directly returning the array is the correct one.

PR #21038, opened 2024-02-06
fix: always set project details

alaister:

Bug fix

### DIFF
--- a/apps/studio/stores/RootStore.ts
+++ b/apps/studio/stores/RootStore.ts
@@ -14,15 +14,11 @@ export interface IRootStore {
   ui: IUiStore
   meta: IMetaStore
 
-  selectedProjectRef?: string
-
   setProject: (project: Project) => void
 }
 export class RootStore implements IRootStore {
   ui: IUiStore
   meta: IMetaStore
-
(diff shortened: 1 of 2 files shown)


apps/studio/stores/RootStore.ts
```diff
@@ -35,16 +31,8 @@ export class RootStore implements IRootStore {
    * This method will also trigger project detail loading when it's not available
    */
   setProject(project: Project) {
-    if (this.selectedProjectRef === project.ref) return
```
alaister, line 38: If the connection string changed but the project ref doesn't, we skip the update. This is the real fix.

apps/studio/stores/RootStore.ts
```diff
@@ -35,16 +31,8 @@ export class RootStore implements IRootStore {
...
-
-    // reset ui projectRef in case of switching projects
-    // this will show the loading screen instead of showing the previous project
-    this.ui.setProjectRef(undefined)
```
alaister, line 42: This line is unnecessary since we only call the `setProject()` method with an actual project.

joshenlim (approved):

PR #21055, opened 2024-02-06
Make the logs explorer look & feel very similar to sql editor

jordienr:

- update look & feel of log explorer to match SQL editor

- we have 2 different UXs for similar things

## What is the new behavior?

they look very similar

## BEFORE
CleanShot 2024-02-06 at 17 23 53@2x

## AFTER
CleanShot 2024-02-06 at 17 23 19@2x


kiwicopple (approved):

NxPKG (approved):

joshenlim:
Some comments to make it slightly more similar to the SQL editor
- "Results will be shown below"
  - Could we make this a dropdown like the SQL editor with just "Results", and we shift Download + Copy as markdown CTAs there
  - Save query + Run -> could we shift these 2 buttons down to the row where "Results" is? Field reference can stay up top
  - There seems to be an empty data grid when there's no results yet, can remove 
image

alaister (approved):

PR #21057, opened 2024-02-06
rm unused line-clamp tailwind plugin

jordienr:


ivasilov (approved):

joshenlim (approved):

joshenlim (changes requested):
Oh whoops looks like docs and www might be using this? i see a couple of files in www like [this one](https://github.com/supabase/supabase/blob/master/apps/www/components/Nav/MenuItem.tsx#L78) using line-clamp-1 class name

jordienr:
> Oh whoops looks like docs and www might be using this? i see a couple of files in www like [this one](https://github.com/supabase/supabase/blob/master/apps/www/components/Nav/MenuItem.tsx?rgh-link-date=2024-02-07T02%3A35%3A43Z#L78) using line-clamp-1 class name

Just uninstalling because I installed it in a previous PR and it turns out it's no longer necessary (since tailwind 3.3).
Sorry should have added more context to the PR description 😅.  

> Warning As of Tailwind CSS v3.3 the line-clamp utilities are now included in the framework by default and this plugin is no longer required.

joshenlim (approved):

PR #21078, opened 2024-02-07
Support for composite foreign keys in table editor

joshenlim:
# Support for composite foreign keys in table editor
The changes in this PR addresses some existing limitations with the current UI for managing foreign keys across tables by providing and more "correct" way of doing so.

### Managing foreign keys will be done at the table level
Previously, the table editor would allow assign foreign keys at the column level, but that assumes a 1:1 relation between columns and hence did not support the concept of composite foreign keys. Not to mention that it wasn't easy to see what foreign keys there were on the table itself.

We've thus shifted the management of foreign keys into the table side panel editor instead:

This way, you can manage all foreign keys across all columns on the table in one place, rather than going into each column individually to do so.

We've also swapped out the existing foreign key management in the column side panel editor with a CTA to do so from the table instead:
image

If the column does belong to some foreign keys, we'll still show them here nonetheless:
image

And composite foreign keys can be done as such:
image


alaister (approved):
Wow, what a great PR! I couldn't find any bugs!

The only small nit would be swapping the order of "Foreign Keys" and "Data Type" here:
Screenshot 2024-02-08 at 16 04 53

PR #21082, opened 2024-02-07
Fix deleting a query thats not saved throwing an error

joshenlim:


apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
```diff
@@ -607,7 +607,7 @@ const SQLEditor = () => {
                   <AiIconAnimation loading={isAiLoading} />
                 </motion.div>
 
-                <AnimatePresence initial={false} exitBeforeEnter>
+                <AnimatePresence initial={false} mode="wait">
```
joshenlim, line 610: seems like exitBeforeEnter is deprecated with the new version of framer-motion
i ran into a client error to replace it with mode wait locally

jordienr (approved):
tested and it works!

PR #21085, opened 2024-02-07
This commit adds a redirect rule to the 'redirect.js' file

jazzstack:
This commit adds a redirect rule to the 'redirect.js' file to permanently redirect requests from '/docs/guides/storage/uploads' to '/docs/guides/storage/uploads/standard-uploads'. This will ensure that users accessing the old URL are redirected to the appropriate location.


Jazz0P:
Thank you for letting me know. 
I guess I may have missed it (that it was already present) when editing the redirect.js.

Issue #20883, opened 2024-01-30
Log Explorer -  `Insert Source` should insert source at cursor or in FROM clause


robpetty:
# Bug report

I noticed a small UI issue in the Logs Explorer feature, specifically related to Insert source.

## Describe the bug

- Navigate to Logs->Query
- Change Insert source.

Expected to insert source where cursor is currently or, alternatively, replace the source in the `from` clause

<img width="545" alt="image" src="https://github.com/supabase/supabase/assets/1473437/0bf9f1cc-eb21-40fe-9beb-26aa582971a6">

Actual:

Appends to end of text regardless of cursor position

<img width="496" alt="image" src="https://github.com/supabase/supabase/assets/1473437/9a55cb78-a926-4ab6-9a35-e527815b244d">

## To Reproduce

1. Navigate to Logs Explorer and see default query
2. Select new source from `Insert Source`

## Expected behavior

Query is correct.

## Screenshots

See above.

## System information

- OS: macOS
- Chrome Version 121.0.6167.85 (Official Build) (arm64)

encima:
Hi @robpetty,

Thanks for reporting. This is a bug but, I think, not in the way you reported.
- `Insert source` - purely adds the source selected (although the bug is that is could/should replace the current source in the `from` clause)
- `Templates` - This is the button you are looking for and this creates correct queries

I will update this issue to to reflect this

robpetty:
@encima thanks for the explanation.  Makes sense.

santosh2345:
I want to solve this issue and contribute.

encima:
@santosh2345 Excellent! Feel free to open a PR and mention this issue in the description

joshenlim:
Resolved by https://github.com/supabase/supabase/pull/21114

PR #21114, opened 2024-02-08
fix: insert source after from command in log explorer

faisalsaifii:

Bug fix for #20883 

#20883 Inserting a source would add it at the end of the query

## What is the new behavior?

Now when we insert a source, it gets added after the FROM command


jordienr (approved):

jordienr:
Good one thanks! we will be improving that feature so that its easier to use.

faisalsaifii:
Thanks @jordienr, let me know if any requirements arise up for this. Will be happy to help.

PR #21116, opened 2024-02-08
Feat: CRUD custom reports

jordienr:

- Allows users to create, delete, update and read many custom reports

You can only create one custom report

## What is the new behavior?

Users can now create more than one custom.

## Additional context

CleanShot 2024-02-08 at 15 41 52@2x
CleanShot 2024-02-08 at 15 42 03@2x


kiwicopple (approved):
lfg 🔥

saltcod:
Awesome @jordienr !

Couple of small things:

When the first report is created, the query isn't getting revalidated here I think? 
screenshot-2024-02-08-at-12 11 40

This takes a second —a <loading> spinner would be good there onSubmit
screenshot-2024-02-08-at-12 12 56

joshenlim:
Could we update the "Create report" CTA when we land on the reports page and there's no custom reports to open the modal that "Create new custom report" in the side menu does? so that the behaviour is consistent

joshenlim:
Also some QOL improvements at the top of my mind

given that we support descriptions in the reports, lets add it below the title too
image

and also add an edit button beside the title, on hover of the title, to edit the title/description (by opening the edit report modal)

joshenlim:
there seems to be some lag when going between custom reports - are the contents not cached in react query?

saltcod (approved):

PR #21122, opened 2024-02-08
docs: add local development section to auth hooks

J0:

Arises from two tickets about local development. Additionally we use app_metadata instead of user_metadata for storing claims (my bad). More internal context [here](https://supabase.slack.com/archives/C02FHG9QQAF/p1707387686521819?thread_ts=1707289169.664599&cid=C02FHG9QQAF)

[Sample `config.toml` when `supabase init` is run](https://github.com/supabase/cli/blob/861262644489ff5bf57fb1124d1e3d6030ffb01a/internal/utils/templates/init_config.toml#L113)

### DIFF
--- a/apps/docs/content/guides/auth/auth-hooks.mdx
+++ b/apps/docs/content/guides/auth/auth-hooks.mdx
@@ -712,4 +712,4 @@ uri = "pg-functions://postgres/public/custom_access_token_hook"
 ...
 ```
 
-Save your Auth Hook as a migration in order to version the Auth Hook and share it with other team members. Run [`supabase migrations new`](https://supabase.com/docs/reference/cli/supabase-migration-new) to create a migration.
+Save your Auth Hook as a migration in order to version the Auth Hook and share it with other team members. Run [`supabase migration new`](https://supabase.com/docs/reference/cli/supabase-migration-new) to create a migration.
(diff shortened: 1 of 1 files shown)


apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -695,3 +695,21 @@ revoke execute
...
+...
+
+[auth.hook.custom_access_token]
+enabled = true
+uri = "pg-functions://postgres/public/custom_access_token_hook"
+
+...
+```
+
+Save your Auth Hook as a migration in order to version the Auth Hook and share it with other team members. Run [`supabase migrations new`](https://supabase.com/docs/reference/cli/supabase-migration-new) to create a migration.
```
charislam, line 715:
```suggestion
Save your Auth Hook as a migration in order to version the Auth Hook and share it with other team members. Run [`supabase migration new`](https://supabase.com/docs/reference/cli/supabase-migration-new) to create a migration.
```

charislam (approved):

apps/docs/content/guides/auth/auth-hooks.mdx
J0: Thanks for catching this!

PR #21131, opened 2024-02-09
feat: add docs feedback widget

charislam:
Add a docs feedback widget to allow upvoting and downvoting pages.

Votes (completely anonymized, barely more than a counter) go to a database in the main `supabase-com` project. If the user has accepted telemetry, the votes also go to Logflare with a bit more info. ~~Post-vote, users can leave a comment, which also goes to a database in the same project.~~

~~If we get a lot of good feedback, the next step would be to hook the backend up to our regular feedback pipelines so it auto-populates in the Feedback DB in Notion. Not doing that right now to speed up this first iteration, because there's no good pre-existing way to do it. (I tried the `platform/feedback/send` endpoint, but it requires user authentication in order to post to it, and we want to accept anonymous feedback.)~~

Revised plan: Feedback is now open to logged-in users only, and it goes to the `platform/feedback/send` endpoint. There is a warning in the modal that the feedback is not anonymous.

image
image
image


saltcod:
a couple of ultra-minor design and UX things: 

1. I love the initial state, but I think the subsequent state could do with like 15mins of Marijana's time  
screenshot-2024-03-08-at-12 56 48


2. can't esc out of the modal, even when its empty. 

screenshot-2024-03-08-at-12 57 37

this is def "make-it-pop" level feedback, nothing blocking.  

This is really great — wish we'd had this years ago!

saltcod (approved):

PR #21136, opened 2024-02-09
Update Account.jsx

lvnilesh:

Bug fix

Column 'avatarUrl' of relation 'profiles' does not exist

## What is the new behavior?

profile image upload succeeds

## Additional context

### DIFF
--- a/examples/user-management/react-user-management/src/Account.jsx
+++ b/examples/user-management/react-user-management/src/Account.jsx
@@ -43,7 +43,7 @@ export default function Account({ session }) {
       id: user.id,
       username,
       website,
-      avatarUrl,
+      avatar_url,
       updated_at: new Date(),
     }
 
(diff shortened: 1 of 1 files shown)


examples/user-management/react-user-management/src/Account.jsx
```diff
@@ -43,7 +43,7 @@ export default function Account({ session }) {
       id: user.id,
       username,
       website,
-      avatarUrl,
+      avatar_url,
```
charislam, line 46:
```suggestion
      avatar_url: avatarUrl,
```

charislam: Thanks! Appreciate it :)

Haven't tested, but looks like you should need this to get it working?

charislam (approved):

charislam (approved):

Issue #21059, opened 2024-02-06
IPv4 to IPv6 Migration + New Supavisor pooler issue and Overflow error


Kinfe123:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

IPv4 to IPv6 Migration + New Supavisor pooler issue and Overflow error

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

I dont know much about the error but it happens lately after the migration and new pooler . it might not be propagated well ... but it keeps me sync while working with ORMs and stuff like from local machine as well production .. may be if is a location issue due to edge propagation or alg on the new pooler , i am from Ethiopia.

## Expected behavior

I was expecting seeing the schemas along with tables with that schema.

## Screenshots
![Screenshot from 2024-02-06 17-48-23](https://github.com/supabase/supabase/assets/65047246/d42ab3d2-bd55-4384-93c9-b7e776e610c4)

## System information

- OS: Fedora
- Browser: Chrome

encima:
Hello @Kinfe123 
Can you confirm your network has support for ipv6? You can check this in the [announcement](https://github.com/orgs/supabase/discussions/17817) (running `curl -6 https://ipconfig.co/ip`)

If it does not, you can use an application like [Cloudflare Warp](https://cloudflarewarp.com/) to enable it.

KevinBrolly:
Hey @Kinfe123 We have been seeing some reports of ECONNREFUSED errors when some users have certain characters in their password when they have an older project/on some versions of PG12 and PG13.

Could you try and reset your DB password and see if that resolves the issue? Or upgrade to the latest release on the platform (if you are not there already).

KevinBrolly:
Hey @Kinfe123 Just to follow up the team let me know this should be fixed now for all users - since this issue was made 2 days ago I am assuming you are fine now. If not just re-open and let us know.

Kinfe123:
i did everything that day .. i am the one who decide on using PG versions

PR #21138, opened 2024-02-09
fix: alert description overflow

Hallidayo:

Supabase Studio > Table Editor (Connection Refused)

If a long description is in the `<Alert />` then the text will overflow.

## What is the new behavior?

Text does now not overflow:

Screenshot 2024-02-09 at 08 43 46

## Additional context

Issue #21059 mentions this.


ivasilov (changes requested):
The screenshot you posted is breaking normal words like `browser`. Can you try using `break-words` class and see if it's better?

Hallidayo:
> The screenshot you posted is breaking normal words like `browser`. Can you try using `break-words` class and see if it's better?

Hi @ivasilov - `break-words` looks much better, just pushed a commit.

ivasilov (approved):

PR #21197, opened 2024-02-12
clarify realtime pricing

inian:
users werent sure of pricing of broadcast and presence. confirming the pricing is the same across all three


kevcodez (approved):
Also adjusted in Notion

PR #21208, opened 2024-02-13
Fix md5 connection string

joshenlim:


apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseSettings.utils.ts
```diff
@@ -14,27 +14,25 @@ export const getConnectionStrings = (
     pgVersion?: string
   }
 ) => {
+  const isMd5 = poolingInfo.connectionString.includes('?options=reference')
```
alaister, line 17: Just to confirm that `options=reference` always appears as the first option in the query string? If it can appear elsewhere maybe it's better to remove the leading `?`

alaister (approved):
LGTM! One small thing:

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseSettings.utils.ts
joshenlim: good point - it always appears as the first option, but removing the leading ? is probably better

PR #21217, opened 2024-02-13
Fix policies page locked schema notice

### LINKED ISSUE #19367: Locked status prevents user from seeing if RLS is enabled/disabled in the GUI
https://github.com/supabase/supabase/blob/5fed762d1fcb7ed060439f0fb4c85f2cf3b5bdd6/apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/PolicyTableRowHeader.tsx#L37-L47

The ternary operator here and in `Policies/index.tsx` will hide the current state of RLS policies (on/off) on a table, if that table is "locked" in the gui.

It's not _really_ locked, as we were able to disable the RLS policies for `storage.objects` in SQL.

Would you all accept a PR to show _both_ of these chips? It would help folks see the status of RLS enabled/disabled on locked tables, if they inadvertently disabled the policies.

joshenlim:
Resolves https://github.com/supabase/supabase/issues/19367

But also adds this notice that we show on the table editor / database pages about locked schemas

image


alaister (approved):
LGTM!

PR #21285, opened 2024-02-15
Fix table editor unable to update PK after renaming table

### LINKED ISSUE #19341: Unable to update table PK via dashboard after renaming table

joshenlim:
Resolves https://github.com/supabase/supabase/issues/19341

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
@@ -33,6 +33,7 @@ import { ForeignKey } from './ForeignKeySelector/ForeignKeySelector.types'
 import { ColumnField, CreateColumnPayload, UpdateColumnPayload } from './SidePanelEditor.types'
 import { checkIfRelationChanged } from './TableEditor/ForeignKeysManagement/ForeignKeysManagement.utils'
 import { ImportContent } from './TableEditor/TableEditor.types'
+import { Constraint } from 'data/database/constraints-query'
 
 const BATCH_SIZE = 1000
 const CHUNK_SIZE = 1024 * 1024 * 0.1 // 0.1MB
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
```diff
@@ -266,11 +268,13 @@ export const createColumn = async ({
   connectionString,
   payload,
   selectedTable,
+  primaryKey,
 }: {
   projectRef: string
   connectionString: string | undefined
   payload: CreateColumnPayload
   selectedTable: PostgresTable
+  primaryKey?: Constraint
```
ivasilov, line 277: Why pass in the primary key when it's available on `selectedTable.primary_keys`?

apps/studio/data/database/constraints-query.ts
```diff
@@ -0,0 +1,69 @@
...
+    {
+      projectRef,
+      connectionString,
+      sql: getTableConstraints({ schema, table }),
+      queryKey: ['table-constraints'],
+    },
+    {
+      enabled: typeof schema !== 'undefined' && typeof table !== 'undefined',
+      select(data) {
+        return ((data as any)?.result ?? []).map((constraint: Constraint) => constraint)
```
alaister, line 64: Does this do the same thing? The map isn't doing anything apart from adding types
```suggestion
        return ((data as any)?.result ?? []) as Constraint[]
```

alaister (approved):
LGTM!

apps/studio/data/database/constraints-query.ts
joshenlim: not sure why but TS is complaining 😕 

image

alaister: hmm maybe the whole map can just be removed entirely then?

joshenlim: unless we rewrite it like this?

```export const useTableConstraintsQuery = (
  { projectRef, connectionString, schema, table }: TableConstraintsVariables,
  options: UseQueryOptions<ExecuteSqlData, TableConstraintsError, Constraint[]> = {}
) => {
  return useExecuteSqlQuery(
    {
      projectRef,
      connectionString,
      sql: getTableConstraints({ schema, table }),
      queryKey: ['table-constraints'],
    },
    {
      enabled: typeof schema !== 'undefined' && typeof table !== 'undefined',
      select(data) {
        return ((data as any)?.result ?? []) as Constraint[]
      },
      ...options,
    }
  )
}```

joshenlim: > hmm maybe the whole map can just be removed entirely then?

oh 🤦 that worked hahaha

PR #21363, opened 2024-02-19
Fix settings/infrastructure compute instance data

jordienr:

- fixes the info being rendered in settings/infrastructure

A user may have a compute instance addon but its always showing 2 CPU cores

## What is the new behavior?

it should render the correct info based on the compute instance 

## Additional context
If you want to test this:
- checkout to master
- create project
- add any addon over 2 CPU cores (XL or over)
- go to settings/infrastructure - it doesn't show the correct CPU cores
- checkout to this branch
- should show the correct CPU cores

### DIFF
--- a/apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureActivity.tsx
+++ b/apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureActivity.tsx
@@ -50,10 +50,18 @@ const InfrastructureActivity = () => {
   const selectedAddons = addons?.selected_addons ?? []
 
   const { computeInstance } = getAddons(selectedAddons)
-  const currentComputeInstanceSpecs =
-    computeInstance?.variant?.meta ?? project?.infra_compute_size === 'nano'
-      ? INSTANCE_NANO_SPECS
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureActivity.tsx
```diff
@@ -50,10 +50,18 @@ const InfrastructureActivity = () => {
   const selectedAddons = addons?.selected_addons ?? []
 
   const { computeInstance } = getAddons(selectedAddons)
-  const currentComputeInstanceSpecs =
-    computeInstance?.variant?.meta ?? project?.infra_compute_size === 'nano'
-      ? INSTANCE_NANO_SPECS
-      : INSTANCE_MICRO_SPECS
+
+  function getCurrentComputeInstanceSpecs() {
```
kevcodez, line 54: Should we rather do useMemo?

jordienr: No need to IMO, this function is not expensive.

kevcodez (approved):

PR #21365, opened 2024-02-19
update events naming structure

fsansalvadore:

Update telemetry naming convention.


awalias (approved):
I haven't personally been inspecting any of these events so the change is fine with me!

PR #21367, opened 2024-02-19
Update humans.txt

kiwicopple:


apps/docs/public/humans.txt
```diff
@@ -57,7 +54,6 @@ Pavel Borisov
 Qiao Han
 Raminder Singh
 Ramiro Nuñez Dosio
-Rodrigo Martins Mansueli
```
mansueli, line 60: Please don't remove me 😅

mansueli (approved):

apps/docs/public/humans.txt
loong: brutal

PR #21388, opened 2024-02-20
feat: Conversational AI for the SQL Editor

ivasilov:


jordienr:
when the convo gets too long it doesn't scroll, the whole screengrows instead
CleanShot 2024-03-04 at 12 07 49@2x

jordienr (changes requested):
just needs scroll in assistant drawer, the rest LGTM! 🚀

jordienr (approved):

saltcod:
sorry if I missed it somewhere, but wanted to note this: 
screenshot-2024-03-04-at-21 47 18

joshenlim:
one observation - "inserting" a code snippet from the AI has the experimental note, but not for "replacing code"
think the latter should also have the experimental note
image

joshenlim:
could we also look into consolidating the components between this and the RLS assistant ones too where possible? we can do it in a separate PR

joshenlim:
another though - wondering if the assistant should be contextual to the snippet?
each snippet should have its own trace of assistant chat?

PR #21412, opened 2024-02-20
fix: Add limitation to Postgres Changes docs regarding Delete events ignoring filtering

filipecabaco:

Add limitation to Postgres Changes docs regarding Delete events ignoring filtering

### DIFF
--- a/apps/docs/content/guides/realtime/postgres-changes.mdx
+++ b/apps/docs/content/guides/realtime/postgres-changes.mdx
@@ -86,7 +86,7 @@ In this example we'll set up a database table, secure it with Row Level Security
   </StepHikeCompact.Step>
 
   <StepHikeCompact.Step step={3}>
-  
+
     <StepHikeCompact.Details title="Enable Postgres replication">
 
       Go to your project's [Replication settings](https://supabase.com/dashboard/project/_/database/replication), and under `supabase_realtime`, toggle on the tables you want to listen to.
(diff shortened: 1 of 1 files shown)


apps/docs/content/guides/realtime/postgres-changes.mdx
```diff
@@ -1347,6 +1347,10 @@ myChannel.updateAuth("fresh-token")
 
 ## Limitations
 
+### Delete events are not filterable
+
+When tracking Delete events with Postgres Changes you are not able to apply Filters on the received records. This is due to a limitation with the way we are pulling changes from Postgres where filters can't be applied. We're working on solutions to this limitation.
```
charislam, line 1352:
```suggestion
You can't filter Delete events when tracking Postgres Changes. This limitation is due to the way changes are pulled from Postgres.
```

charislam (approved):
non-blocking suggestion for concision/more neutral tone

apps/docs/content/guides/realtime/postgres-changes.mdx
w3b6x9, line 1352:
```suggestion
You can't filter `DELETE` events when tracking Postgres Changes. This limitation is due to the current way Realtime polls for changes from Postgres.
```

@filipecabaco you can also mention this [issue](https://github.com/supabase/walrus/issues/53) so folks can track it. @charislam would love your thoughts on doing so.

charislam: I'm not a huge fan of linking issues in docs because it feels less polished, which can make the product seem like a work in progress. I recognize that's a very subjective feeling, though, so if you think it would really help, I'm not going to block it 😄

filipecabaco: I will keep the changes that @charislam suggested and lets keep the issue out for now and lets see the feedback

PR #21416, opened 2024-02-21
fix: local studio doesn't have organization

charislam:


apps/studio/components/interfaces/Auth/BasicAuthSettingsForm/BasicAuthSettingsForm.tsx
```diff
@@ -89,9 +89,12 @@ const BasicAuthSettingsForm = () => {
...
 
   const organization = useSelectedOrganization()
-  const { data: subscription, isSuccess: isSuccessSubscription } = useOrgSubscriptionQuery({
-    orgSlug: organization!.slug,
-  })
+  const { data: subscription, isSuccess: isSuccessSubscription } = useOrgSubscriptionQuery(
+    {
+      orgSlug: organization?.slug,
+    },
+    { enabled: !!organization }
```
kevcodez, line 96: What about `enabled: IS_PLATFORM`?

Underlying RQ won't/shouldn't fire when org slug is undefined either way

kevcodez: We can actually move the IS_PLATFORM check to the RQ itself, but would suggest doing that in a separate PR (can look into it eventually)

apps/studio/components/interfaces/Auth/BasicAuthSettingsForm/BasicAuthSettingsForm.tsx
```diff
@@ -244,7 +247,7 @@ const BasicAuthSettingsForm = () => {
                     }}
                     formValues={values}
                   />
-                  {isProPlanAndUp ? (
+                  {isProPlanAndUp || !IS_PLATFORM ? (
```
kevcodez, line 250: Shouldn't isProPlanAndUp be false all the time anyway given `isSuccessSubscription` should never be true? 🤔

apps/studio/components/interfaces/Auth/BasicAuthSettingsForm/BasicAuthSettingsForm.tsx
charislam: Good idea 👍🏼

apps/studio/components/interfaces/Auth/BasicAuthSettingsForm/BasicAuthSettingsForm.tsx
charislam: Yeah it's false, but we actually need this branch to run, because we don't want to prompt for a Pro upgrade on local. I added a new variable `promptProPlanUpgrade` to make that clearer.

kevcodez (approved):

PR #21432, opened 2024-02-21
Update with-svelte.mdx

kitchenchem224:

bug fix

Unable to call `supabase.auth.signOut()`

## What is the new behavior?

call `supabase.auth.signOut()` on click event `on:click`

## Additional context

[docs page here](https://supabase.com/docs/guides/getting-started/tutorials/with-svelte?database-method=dashboard)


charislam (approved):

charislam:
ahh thanks for the heads-up. I don't think the formatter recognizes Svelte, and it certainly won't format it properly as an `html` snippet, but `tsx` seems to be a reasonable hack to get it working

PR #21443, opened 2024-02-22
Update error handling for table editor RQs

joshenlim:


apps/studio/data/fetchers.ts
```diff
@@ -148,3 +149,7 @@ export const options: typeof _options = async (url, init) => {
     headers,
   })
 }
+
+export const handleError = (error: ResponseError) => {
+  throw new Error(error.message)
```
ivasilov, line 154: I'm wondering if we should do `if (error instanceof Error) throw error else throw new Error`. An error might have other fields which we're stripping in this approach.

joshenlim: so just throw the error directly if its already an Error object?

hmm, this will be used specifically for the data/fetchers methods though - you got a use case in mind that will be negatively affected by this approach?

> an error might have other fields which we're stripping in this approach

currently the errors would have the id and status code iirc - but FE will only surface the message (which i think should be the only thing that gets surfaced, at least for now)

ivasilov (approved):

PR #21466, opened 2024-02-22
Update next.js setup docs email template

ryanhalliday:
The email templates need to have the type parameter. 

Docs

I was getting 401 `AuthApiError: Email link is invalid or has expired` until I changed this for my `magiclink`

## What is the new behavior?

Tell the user everything they need to add

## Additional context

The wording here is a bit clumsy but it's hard to convey succinctly. A variable for the email type would be kinda cool.


charislam:
Hmm so you changed the example to `signInWithEmail` somewhere?

It's good to document this so we should definitely find somewhere for it to go, but the reason I'm hesitant to put it here is that it's a Getting Started guide. Which means the reader should have to make a minimum number of decisions, and we should just have a single path that works perfectly if they follow the instructions as written.

So if they copy/pasted everything else exactly, they shouldn't have to make a decision on what the `type` is here.

And if they didn't, then the information on how to do that change should be elsewhere. (Which I think is acceptable because it's a change from the default.)

Of course, if that code doesn't work with an exact copy/paste then I misunderstood and we should fix that 😄

ryanhalliday:
Yeah I did change the example to that, but with the incorrect type param for each email it will potentially also break recovery/email change etc. 
I followed through the tutorial, then made that change after completing it, and didn't immediately know why it was broken.

I would say best to not edit the other templates rather than break them in a way that the user doesn't realise? 

Really a dynamic template variable in the emails would be so much cooler.

charislam:
Oh, I get what you're saying. Yeah, agreed 💯. Going to make a slight edit to your PR and put it in! Thanks so much for bringing this up!

apps/docs/content/guides/auth/server-side/nextjs.mdx
```diff
@@ -400,7 +400,7 @@ If you have email confirmation turned on (the default), a new user will receive
 
 Change the email template to support a server-side authentication flow.
 
-Go to the [Auth templates](https://supabase.com/dashboard/project/_/auth/templates) page in your dashboard. In each email template, change `{{ .ConfirmationURL }}` to `{{ .SiteURL }}/auth/confirm?token_hash={{ .TokenHash }}&type=signup`.
+Go to the [Auth templates](https://supabase.com/dashboard/project/_/auth/templates) page in your dashboard. In each email template, change `{{ .ConfirmationURL }}` to `{{ .SiteURL }}/auth/confirm?token_hash={{ .TokenHash }}&type=signup|invite|magiclink|recovery|email_change|email` and ensure you remove the extra `&type=` options to match the email template. Eg `{{ .SiteURL }}/auth/confirm?token_hash={{ .TokenHash }}&type=signup`
```
charislam, line 403:
```suggestion
Go to the [Auth templates](https://supabase.com/dashboard/project/_/auth/templates) page in your dashboard. In the `Confirm signup` template, change `{{ .ConfirmationURL }}` to `{{ .SiteURL }}/auth/confirm?token_hash={{ .TokenHash }}&type=signup`.
```

charislam (approved):

PR #21469, opened 2024-02-22
fix: statically build the right nav on each page

charislam:
**Before:** All pages have the Home nav menu in static HTML, which is blown away and replaced by the proper nav menu upon hydration. This leads to jankiness when the page first loads and an unpleasant flash of the wrong nav menu (especially obvious on the JavaScript ref page, which takes a long time to process and rerender the nav).

**Now:** All pages have their correct nav menu in static HTML.


apps/docs/layouts/MainSkeleton.tsx
charislam, line 1: Most of this is extracted out of `./SiteLayout.tsx` below, just added the `menuId` prop to thread through a static value for that.

kiwicopple (approved):

PR #21547, opened 2024-02-26
Chore/read replica fixes 260224

joshenlim:
- Staging: only allow apse1 as selection for replica
- Fix modal stuck in loading state if dropping replica fails
- Add warnings to prevent deploying RR if cloud provider is not AWS, or PG version is not at least 15
- Add warnings on add ons side panels for downgrading compute and pitr if read replicas are present
- Add links to change compute instance from settings/infrastructure
- Add option to remove all replicas in one action from settings/infrastructure


alaister (approved):
LGTM!

Issue #21391, opened 2024-02-20
RLS docs encourage using a vulnerable RLS policy


aantn:
The docs make it clear that the `user_metadata` inside JWTs is not to be trusted. See [here](https://supabase.com/docs/guides/auth/row-level-security#authjwt:~:text=user_metadata%20%2D%20can%20be%20updated%20by%20the%20authenticated%20user%20using%20the%20supabase.auth.update()%20function.%20It%20is%20not%20a%20good%20place%20to%20store%20authorization%20data.) and [here](https://github.com/supabase-community/supabase-custom-claims?tab=readme-ov-file#whats-the-difference-between-authusersraw_app_meta_data-and-authusersraw_user_meta_data).

However, the RLS docs explicitly recommend relying on `user_metadata` inside JWTs. See [here](https://supabase.com/docs/guides/auth/sso/auth-sso-saml#row-level-security:~:text=auth.jwt()%23%3E%3E%27%7Buser_metadata%2Ciss%7D%27%20Returns%20the%20identity%20provider%27s%20SAML%202.0%20EntityID)

To add insult to injury, when logging in with SSO, the user's actual custom_claims and groups are written to auth.users.raw_user_meta_data which is certainly not secure. (It can be overriden easily.) I assume that should be ignored entirely.

I have two questions:

1. Is the recommended RLS policy indeed vulnerable to privilege escalation. Or are the docs wrong and if so, which page should be trusted?
2. What is the secure way to check a user's SSO claims and/or groups? Is it via `auth.jwt()`? Is it via the `auth.identities` table as [implied here](https://supabase.com/docs/guides/auth/sso/auth-sso-saml#row-level-security:~:text=Accessing%20the%20stored%20attributes)? Or some other manner?

aantn:
I've been testing and I'm nearly certain now that the [RLS docs (for SSO)](https://supabase.com/docs/guides/auth/sso/auth-sso-saml#row-level-security) are wrong and dangerous.

E.g. a malicious user can bypass the suggested RLS policy by changing their ISS in javascript:

```
await this.client.auth.update({ data: { iss: "https://lets.be.someone.else.today.com" }});
```

That gets reflected in `auth.jwt` in SQL meaning the docs suggestion that you can rely on the ISS value is wrong. Likewise, you cannot rely on custom claims or groups as they appear in the JWT.

aantn:
[...]

PR #21561, opened 2024-02-26
docs: warn about `user_metadata` used in RLS

hf:
Adds a warning to the Row Level Security page that using `user_metadata` can be insecure.

Fixes:
- https://github.com/supabase/supabase/issues/21391


apps/docs/content/guides/auth/row-level-security.mdx
```diff
@@ -73,6 +73,12 @@ Returns the ID of the user making the request.
 
 ### `auth.jwt()`
 
+<Admonition type="caution">
+
+Not all information present in the JWT should be used in RLS policies. For instance, creating an RLS policy that relies on the `user_metadata` claim can create security issues in your application as this information can be modified by authenticated end-users.
```
charislam, line 78:
```suggestion
Not all information present in the JWT should be used in RLS policies. For instance, creating an RLS policy that relies on the `user_metadata` claim can create security issues in your application as this information can be modified by authenticated end users.
```

charislam (approved):

PR #21577, opened 2024-02-26
feat: archive badge added for examples

### LINKED ISSUE #21521: Several example projects are archived
As seen [here](https://github.com/supabase/supabase/blob/79d806ad22624baf086df7385827dc73e9a38316/apps/www/data/Examples.ts#L197) and in a few other spots in that file, there's several example projects that are linked but that are in an [archived example repo](https://github.com/supabase/examples-archive), and probably shouldn't be used for reference until they're updated!

Hallidayo:

Supabase Website - [Auth](https://supabase.com/auth)
Supabase Studio - New Project/Examples

Their are a couple of examples which have archived repos which the user does not know until they go to the example/repo

## What is the new behavior?

A badge has been added to show that the example has an archived expo:

Screenshot 2024-02-26 at 20 38 08
Screenshot 2024-02-26 at 20 22 47

## Additional context

Closes #21521


MildTomato:
Should we just remove these, maybe as a good rule of thumb we shouldn't be surfacing outdated/archived repos to Supabase users.

what do others think ?

charislam:
+1 for removing them

saltcod:
+1 for removing!

Hallidayo:
Hi @MildTomato, @charislam, @saltcod  - I've just removed the archived repo examples.

saltcod (approved):

PR #21590, opened 2024-02-27
Fix view definition query

joshenlim:


alaister (approved):

apps/studio/data/database/view-definition-query.ts
```diff
@@ -7,10 +7,8 @@ type GetViewDefinition = {
 }
 
 export const getViewDefinitionQuery = ({ schema, name }: GetViewDefinition) => {
```
alaister, line 9:
```suggestion
export const getViewDefinitionQuery = ({ schema = 'public', name }: GetViewDefinition) => {
```
now that the `.filter(Boolean)` has been removed, we should always ensure there's a schema

joshenlim: does the check at the bottom suffice though? query is enabled only if both schema and name provided

alaister: oh true. If we're always providing a schema then that's fine

PR #21617, opened 2024-02-27
fix: typos in example sql of blog post

nicholasdly:

Fixes to minor typos found in the provided SQL snippets of the [Matryoshka embeddings](https://supabase.com/blog/matryoshka-embeddings) blog post.

Attempting to create the defined `match_documents_adaptive` PostgreSQL function raises a syntax error due to the incorrect language specification (line 390).

Even after that typo is fixed, attempting to create the function again raises another syntax error due to an invalid comma (line 395).

## What is the new behavior?

The `match_documents_adaptive` PostgreSQL function can now be created exactly as described without error.

## Additional context

While experimenting with the topic discussed in the blog post, a coworker and I ran into the described errors attempting to run the PostgreSQL code that would create the `match_documents_adaptive` function in our Azure PostgreSQL database.


gregnr (approved):
Thanks for catching this @nicholasdly!

Since the function is a simple query that doesn't require `plpgsql`, I changed the language back to `sql` and removed the `begin` and `end` statements that shouldn't have been there in the first place.

Thanks again for the correction 🚀

PR #21629, opened 2024-02-28
Add separate flag to disable compute size changes

joshenlim:


apps/studio/components/interfaces/Settings/Addons/Addons.tsx
```diff
@@ -249,14 +250,15 @@ const Addons = () => {
                       </p>
                     )}
                     <ProjectUpdateDisabledTooltip
-                      projectUpdateDisabled={projectUpdateDisabled}
+                      projectUpdateDisabled={computeSizeChangesDisabled}
```
alaister, line 253: We don't want to disable this when `projectUpdateDisabled` is enabled as well?
```suggestion
                      projectUpdateDisabled={computeSizeChangesDisabled || projectUpdateDisabled}
```

joshenlim: so disabling compute sizes will be its own flag, separate from the flag that controls "project creation and update"

joshenlim: hmm actually good point...

joshenlim: yup okay i'll add that in! thanks man! 🙏 🙏

alaister (approved):
LGTM!

PR #21658, opened 2024-02-29
fix: update rate limit from 4 to 3

J0:

Current rate limit is 3


apps/docs/content/guides/platform/going-into-prod.mdx
```diff
@@ -62,7 +62,7 @@ After developing your project and deciding it's Production Ready, you should run
 
 | Endpoint                                         | Path                                                           | Limited By               | Rate Limit                                                                                                                                 |
 | ------------------------------------------------ | -------------------------------------------------------------- | ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------ |
-| All endpoints that send emails                   | `/auth/v1/signup` `/auth/v1/recover` `/auth/v1/user`[^1]       | Sum of combined requests | Defaults to 30 emails per hour. As of 14th July 2023, this has been updated to 4 emails per hour. Is customizable with custom SMTP set up. |
+| All endpoints that send emails                   | `/auth/v1/signup` `/auth/v1/recover` `/auth/v1/user`[^1]       | Sum of combined requests | Defaults to 30 emails per hour. As of 29 Feb 2024, this has been updated to 3 emails per hour. Is customizable with custom SMTP set up. |
```
w3b6x9, line 65: @J0 can we append instead to show a bit of history since there was some displeasure back when the decrease down to 4 happened? https://github.com/orgs/supabase/discussions/15896#discussioncomment-6848291

can we also align the date with this change https://github.com/supabase/supabase/commit/fcb4938d0f521434b1322248da1a88f0b20a41a0 back in Oct 2023?

```suggestion
| All endpoints that send emails                   | `/auth/v1/signup` `/auth/v1/recover` `/auth/v1/user`[^1]       | Sum of combined requests | Defaults to 30 emails per hour. As of 14th July 2023, this has been updated to 4 emails per hour. As of 21 Oct 2023, this has been updated to 3 emails per hour. You can only change this with your own custom SMTP setup. |
```

J0: Sure that works, I think the latest email rate limit is [currentl 2](https://github.com/supabase/infrastructure/pull/17131/files) so we might need to add an additional line.

Let me check if the change is being applied and get back

joshenlim (approved):

PR #21683, opened 2024-03-01
Chore/foreign key management in column editor

### LINKED ISSUE #21512: UX for adding single-column FKs got worse
# Bug report

## Describe the bug

hello supabase team!

I submitted feedback on the UI but wanted to get others' opinions as well. Just noticed this update to move column-level FKs to the table-level:
https://github.com/supabase/supabase/pull/21078

I'm not sure if others are experiencing the same pain, but my previous flow for adding a FK column was this:
1. create a column, select the reference column it points to

so easy! it was amazing how supabase could automate so much of the tediousness that would be required for manually writing DDL. but now... I noticed that since this moved to the table-level, it is a huge pain to do this:
1. first, look up the data type of the target column to reference to, make a note of that data type for later
2. create a source table column with that data type
3. edit the table-level FKs to set up the relationship

If you get to 3 and realize it was the wrong data type, you have to effectively start everything over, that's a lot of frustration and clicking around for what used to be basically one field. Not only that, but when creating the source column it tells you to edit the table to access FKs, with a link to take you there!, which is particularly aggravating given that the source column does not exist yet!! So you click that link to edit the table-level FKs and simply cannot, you have to back out, AND your progress on creating the new column is GONE.

This UX was one of the worst evolutions I've seen. I get the motivations were to support single- and composite- column FKs in a single place of the UI, but the price is heavy for the single-column FK workflow. I hope we can find some way to get back to the previous level of usability, as one of the primary draws to supabase for me is the ease of maintaining an RDBMS, and adding non-composite FKs is a way more common maintenance activity for me than creating a composite one.

PS.
I'd also like to say how much I appreciate this project, I'm starting a project and supabase has helped me get running a lot faster than if I had done other hosting. really hoping I can help the community as a whole with this feedback, if sensible.

## To Reproduce

Go to Table Editor, try to create a single-column FK.

## Expected behavior

Go to Table Editor, when adding a column make it easy to FK it to another column without worrying about data type.

## System information

- OS: Linux
- Browser (if applies): vivaldi

joshenlim:
Support the new foreign key flow in the Column side panel editor

Resolves https://github.com/supabase/supabase/issues/21512

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
@@ -159,16 +159,20 @@ const GridHeaderActions = ({ table, isViewSelected, isTableSelected }: GridHeade
             <div className="flex items-center gap-1">
               {policies.length < 1 ? (
                 <Tooltip.Root delayDuration={0}>
-                  <Tooltip.Trigger className="w-full">
-                    <Link passHref href={`/project/${projectRef}/auth/policies?search=${table.id}`}>
-                      <Button
-                        type="default"
-                        className="group !h-[28px] !py-0"
-                        icon={<IconPlusCircle size={12} />}
+                  <Tooltip.Trigger asChild className="w-full">
+                    <Button
+                      asChild
+                      type="default"
+                      className="group !h-[28px] !py-0"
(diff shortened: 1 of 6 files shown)


alaister (approved):
LGTM!

PR #21685, opened 2024-03-01
fix logo on page refresh

fsansalvadore:

The navbar logo isn't in sync with theme when refreshing the page.
This was previously fixed by checking mount/unmount which was removed because it made the logo (and the whole navbar) flash on load.
This PR fixes it definitely.

https://github.com/supabase/supabase/assets/25671831/50230e0c-1b1c-46b9-a810-fa742354a8f7

## What is the new behavior?

https://github.com/supabase/supabase/assets/25671831/ab461356-2607-43f3-a4e7-9e77781d1e92


MildTomato:
> This PR fixes it definitely

famous last words

MildTomato (approved):

PR #21697, opened 2024-03-01
Chore/store last schema used

saltcod:
stores the last-selected schema in localStorage, so when you go back to the Table Editor, that same schema is selected. 
Currently, every time you visit the table editor, it selects the 'public' schema

The <SchemaSelector> is used 11 times. We should decide on a case-by-case where else we should do this.

screenshot-2024-03-01-at-11 25 34


apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
```diff
@@ -122,10 +128,10 @@ const TableEditorMenu = () => {
         <div className="flex flex-col gap-1">
           <SchemaSelector
             className="mx-4 h-7"
-            selectedSchemaName={snap.selectedSchemaName}
+            selectedSchemaName={selectedSchema}
             onSelectSchema={(name: string) => {
               setSearchText('')
-              snap.setSelectedSchemaName(name)
```
joshenlim, line 128: i haven't tested this locally but be careful with this - there are a few other components using snap.selectedSchemaName

if we're removing updating the schema name in the valtio store i think it'll break something
e.g ForeignKeySelector.tsx L301
e.g TableGridEditor L325 etc etc

joshenlim: yup can confirm this breaks a few stuff - TableEditor L181
if you try go to another schema, create a table there, it will create the table in the public schema instead (since its the default on pg-meta's end)

joshenlim (changes requested):
This change will cause a few breaking changes

There are several UI components that are using `selectedSchemaName` from the valtio store, and since we're removing updating that variable from SchemaSelector in TableEditorMenu, it's causing some inconsistencies

PR #21714, opened 2024-03-02
feat: full screen side panel button

Hallidayo:

Supabase Studio > Side Panel

Currently the side panel just shows on a small side of the screen.

## What is the new behavior?

Users now have the ability to full size the panel:

https://github.com/supabase/supabase/assets/22655069/873da0a5-2ffe-4cc3-8a36-6d9f32105568

## Additional context

Discussed in #21438


saltcod:
Hey @Hallidayo!  This is quite clever! Sorry for the delay on reply here, but we're in the middle of an iteration on this panel. We're hoping to get it out in the next couple of weeks. Going to mark this one as closed, but offer an appreciative handshake just the same :^)

Issue #6267, opened 2022-04-02
Edge Functions require CORS headers in order to be called with function.js


GaryAustin1:
# Bug report
Using function.js invoke to call a Edge Function generates CORS errors.

## To Reproduce

Invoke the sample hello function using Supabase.js function.invoke() call.

## Expected behavior
It appears a critical part of the process has been left out in generating functions in the examples.
As a quick test I modified the example return to add Access Control Headers
```
 return new Response(
    JSON.stringify(data),
    { 
      headers: {
        "Content-Type": "application/json",
        "Access-Control-Allow-Origin": "*",
        "Access-Control-Allow-Methods": "POST",
        "Access-Control-Expose-Headers": "Content-Length, X-JSON",
        "Access-Control-Allow-Headers": "apikey, X-Client-Info, Content-Type, Authorization, Accept, Accept-Language, X-Authorization",
      }
      },
  );
```
With this set of headers the Edge Function worked.
EDIT it worked with fetch... to work with invoke() I had to add apikey and X-Client-Info

I have not parsed the exact headers, but just did a quick copy from the web searching for CORS errors.

There might be a better way.

## Screenshots
Before the headers:
![image](https://user-images.githubusercontent.com/54564956/161389424-83620af5-d53e-49de-91d6-a9be8967a237.png)
After the headers:
![image](https://user-images.githubusercontent.com/54564956/161389554-e861745e-fda0-4c72-abac-23a8a1e8592a.png)

## System information
Latest Edge Functions Examples
Latest supabase.js

GaryAustin1:
The above was tested using javascript fetch as I'm having another issue with the body format on functions.invoke()...
Going back to invoke and still getting CORS error.. I suspect some more header info needs to be added to the above...

GaryAustin1:
After a bit of trial and error.  You have to handle the preflight request (OPTIONS) and the POST request as two different calls for browser calls with function.js.

This is the code I ended up with:
```
import { serve } from "https://deno.land/std@0.131.0/http/server.ts";

console.log("Hello from Functions!");

serve(async (req) => {
        console.log(req.method)
        if (req.method === 'OPTIONS') {
            return new Response(
                'ok',
                {
                    headers: {
                        "Access-Control-Allow-Origin": "*",
                        "Access-Control-Allow-Methods": "POST",
                        "Access-Control-Expose-Headers": "Content-Length, X-JSON",
[...]

PR #21723, opened 2024-03-04
Alternate setup (proxy) cors.mdx

ajayvignesh01:

Docs update: The existing approach in the docs to invoke edge functions from the browser involves handling CORS preflight requests. This however counts as two function invocation in billing as there is always an OPTION request before the actual request.

The PR proposes adding an alternate approach to the docs showing how to use a simple proxy to route requests through the same origin and avoid CORS checks, therefore only invoking the function once per request.  The proposed approach mentions that hosting providers usually offer a built-in proxy solution, and goes on to show how to setup one for any app hosted on Vercel by configuring the vercel.json file.

Handling CORS preflight requests, which results in two function invocations per request.

## What is the new behavior?

Add another method for users to choose from to handle sending requests from the browser.

## Additional context

### DIFF
--- a/apps/docs/content/guides/functions/cors.mdx
+++ b/apps/docs/content/guides/functions/cors.mdx
@@ -50,3 +50,59 @@ Deno.serve(async (req) => {
   }
 })
 ```
+
+### Alternate setup (Proxy)
+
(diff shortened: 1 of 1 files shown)


inian:
Closing this PR as per this comment https://github.com/supabase/supabase/issues/6267#issuecomment-1975943846

Issue #21740, opened 2024-03-04
Format numbers in statistics charts


saltcod:
The numbers on these charts are hard to parse.
We'd like to add formatting to make them more scannable

Eg: 9913 becomes 9,913
.toLocaleString should probably do it

![screenshot-2024-03-04-at-10 37 30](https://github.com/supabase/supabase/assets/105593/fed450ae-3aac-4b18-b493-09c94717e7b3)

thefool76:
hey there! I would like to work on it. can you please assign it ?

uchechiukpa:
If this has not been assigned yet, I would like to work on it.

saltcod:
Got a PR ^^ reviewing rn! thanks all

PR #21742, opened 2024-03-04
Format numbers in statistics charts

eastcoasting:
Addresses [Format numbers in statistics charts #21740](https://github.com/supabase/supabase/issues/21740#issue-2166945581)

Using `Intl.NumberFormat`, optionally can use the config parameters to set more compact formats for larger numbers (K, M, B).

### DIFF
--- a/apps/studio/components/interfaces/Home/ProjectUsage.tsx
+++ b/apps/studio/components/interfaces/Home/ProjectUsage.tsx
@@ -71,6 +71,7 @@ const ProjectUsage = () => {
     endDateLocal.toISOString()
   )
   const datetimeFormat = selectedInterval.format || 'MMM D, ha'
+  const numberFormat = new Intl.NumberFormat()
 
   const handleBarClick = (
     value: UsageApiCounts,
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Home/ProjectUsage.tsx
```diff
@@ -129,7 +129,7 @@ const ProjectUsage = () => {
                 yAxisKey="total_rest_requests"
                 onBarClick={(v: unknown) => handleBarClick(v as UsageApiCounts, 'rest')}
                 customDateFormat={datetimeFormat}
-                highlightedValue={sumBy(charts, 'total_rest_requests')}
+                highlightedValue={Intl.NumberFormat("en-US").format(sumBy(charts, 'total_rest_requests'))}
```
nikodraca, line 132: I'm not sure if Supabase supports localization, if so this should probably be set to the user's locale and not hardcoded to `en-us`

If not, disregard 😄

saltcod: no localization (yet!)

nikodraca: Gotcha 👍 it still might be worthwhile moving this somewhere central like a config file (if there's a pattern for this already) so that you don't have to hunt down every instance of this when there is localization, but that's more of a nit 🙂

eastcoasting: One option is just removing the locale?
image

nikodraca: that makes sense to me: [The implementation's default locale will be used](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument) 🙂

saltcod (approved):

saltcod:
@eastcoasting what happened there?

eastcoasting:
Sorry I closed on accident by deleting my fork. Reopened [here](https://github.com/supabase/supabase/pull/21750).

PR #21744, opened 2024-03-04
perf: lazyload feature preview modal

kevcodez:
The feature preview modal brings in some heavy dependencies such as react-markdown. There is no need to render this in a blocking fashion.

PR aims to load the feature preview modal lazily. Shaves ~16% off the main bundle 🚀 

Tested locally, but please test, too.


alaister (approved):
LGTM!

PR #21745, opened 2024-03-04
perf: load hcaptcha on demand

kevcodez:
Instead of loading hcaptcha on *every* page and blocking initial render, we can also only load it on the few necessary pages/components.

Shaves 9% off the main bundle. Tested locally, but please test again 🙏


alaister (approved):
LGTM!

PR #21748, opened 2024-03-04
fix: Cannot access list projects on self-host

tranhoangvuit:
Currently, when self-hosting, accessing `/projects` results in an error. This PR addresses and resolves that issue

It shows an error:

Screenshot 2024-03-04 224432

## What is the new behavior?
We can access to projects page normally.

image

### DIFF
--- a/apps/studio/components/interfaces/Home/ProjectList/ProjectCard.tsx
+++ b/apps/studio/components/interfaces/Home/ProjectList/ProjectCard.tsx
@@ -26,7 +26,7 @@ const ProjectCard = ({
   const { name, ref: projectRef } = project
   const desc = `${project.cloud_provider} | ${project.region}`
 
-  const isBranchingEnabled = project.preview_branch_refs.length > 0
+  const isBranchingEnabled = project.preview_branch_refs?.length > 0
   const isGithubIntegrated = githubIntegration !== undefined
   const isVercelIntegrated = vercelIntegration !== undefined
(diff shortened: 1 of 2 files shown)


tranhoangvuit:
Hi @ivasilov, thank you for your feedback. We can still access this page via the URL when someone switches from our cloud to local or self-hosted environments. I want to prevent this page from throwing an exception. Perhaps it would be better to just redirect it to the project details if someone accesses this page?

ivasilov:
> Hi @ivasilov, thank you for your feedback. We can still access this page via the URL when someone switches from our cloud to local or self-hosted environments. I want to prevent this page from throwing an exception. Perhaps it would be better to just redirect it to the project details if someone accesses this page?

Yeah, let's just add a redirect.

apps/studio/components/interfaces/Home/ProjectList/ProjectCard.tsx
```diff
@@ -26,7 +26,7 @@ const ProjectCard = ({
   const { name, ref: projectRef } = project
   const desc = `${project.cloud_provider} | ${project.region}`
 
-  const isBranchingEnabled = project.preview_branch_refs.length > 0
+  const isBranchingEnabled = project.preview_branch_refs?.length > 0
```
tranhoangvuit, line 29: [note]: I just want to prevent errors from being thrown before redirecting to a new page

ivasilov (approved):

PR #21752, opened 2024-03-04
perf: improve prism-react-renderer setup

kevcodez:
- Upgrade to latest major version 2.x
- Only load it in the one component that actually uses the Renderer
- Removed unnecessary "parse-numeric-range" dependency for dead code path

Main bundle size decreases by 16%


alaister (approved):
LGTM!

PR #21756, opened 2024-03-04
Chore/exclude chars

saltcod:
don't allow "@" ";" or "/"


apps/studio/pages/new/[slug].tsx
```diff
@@ -539,6 +540,7 @@ const Wizard: NextPageWithLayout = () => {
                 )}
 
                 <Panel.Content className="border-b border-panel-border-interior-light [[data-theme*=dark]_&]:border-panel-border-interior-dark">
+                  {dbPass}
```
Comment (charislam) [inline apps/studio/pages/new/[slug].tsx:543]:
```suggestion
```

pretty sure this was unintentional, right?

charislam (approved):

MildTomato (changes requested):
Blocking as it makes passwords more insecure.

Do we have rationale for this?

---

Edit - found internal docs. Closed PR while discussion ongoing.

PR #21806, opened 2024-03-06
Hybrid RLS Assistant

joshenlim:
This change aims to streamline the existing RLS policy user flow by
- Not hiding templates on a separate "step"
- Showing the SQL that will be triggered from creating/updating RLS policies

Eventually we'll plan to see how we can integrate the AI assistant into this workflow as well

Before:
image

After:
image


alaister (commented):
Small one but this dropdown looks a little busted in both dark and light mode:

Screenshot 2024-03-25 at 10 56 37

joshenlim:
> Small one but this dropdown looks a little busted in both dark and light mode:

@alaister i'll probably fix this in another PR - this component needs fixing in general, but is used in several places
it's functionality is okay though right?

alaister (approved):
LGTM!

alaister (approved):
Double approved!

PR #21814, opened 2024-03-06
fix: update terminology around anonymous vs public/unauthenticated

charislam:


apps/docs/content/guides/auth/row-level-security.mdx
```diff
@@ -40,11 +40,11 @@ where auth.uid() = todos.user_id;
 -- Policy is implicitly added.
 ```
 
-## Authenticated and anonymous roles
+## Authenticated and unauthenticated roles
 
 Supabase Auth maps every request to one of the roles:
```
kangmingtay, line 45:
```suggestion
Supabase Auth maps every request to one of the postgres roles:
```

apps/docs/content/guides/database/postgres/roles.mdx
```diff
@@ -86,7 +86,7 @@ The default Postgres role. This has admin privileges.
 
 ### `anon`
 
-For "anonymous access." This is the role which the API (PostgREST) will use when a user _is not_ logged in.
+For unauthenticated, public access. This is the role which the API (PostgREST) will use when a user _is not_ logged in.
```
kangmingtay, line 89: maybe we can also add [below](https://github.com/supabase/supabase/pull/21814/files#diff-5556b95c3fe109bc51da5574574956f62c3b9303b87a13727ed098af1fa1b2ecR98) that the `authenticated` role is used by an anonymous user too?

kangmingtay (approved):

hf (approved):

apps/docs/content/guides/database/postgres/roles.mdx
charislam: @kangmingtay, can you put that in your PR? I want to keep this one all pre-release of anonymous signin, and yours all post-release of anonymous signin, if that makes sense

PR #21827, opened 2024-03-07
docs: Adding JS example of multiple foreign keys to the same table

masaok:

docs update

Missing example for multiple foreign keys to same table

## What is the new behavior?

Added the example (but only in JS for now)

## Additional context

ChatGPT4 gave me the answer, and it worked.

https://chat.openai.com/share/85cdc4a0-fac6-45c9-a775-3ed14b5116b3

Gemini was close, but wrong:

https://g.co/gemini/share/529e4d698ee5


apps/docs/content/guides/database/joins-and-nesting.mdx
```diff
@@ -260,3 +260,60 @@ GET https://[REF].supabase.co/rest/v1/teams?select=id,team_name,users(id,name)
...
+  "user_id" int references users,
+  "scan_id_start" int references scans, -- clocking in
+  "scan_id_end" int references scans, -- clocking out
+  "attendance_status" text
+);
+```
+
+In this case, you need to explicitly define the join because the joining column on `shifts` is ambiguous.
+
+If we wanted to fetch all the `shifts` with `scan_id_start` related to a specific `scan`:
```
charislam, line 294:
```suggestion
To fetch all the `shifts` with `scan_id_start` related to a specific `scan`, use the following syntax:
```

charislam (approved):

PR #21855, opened 2024-03-07
fix: persist pgsodium encryption key between restarts

ghost:

Bug fix for https://github.com/supabase/vault/issues/27

Currently, the docker-compose example does not persist the `pgsodium` encryption key that is used by supabase vault (a new, random key is generated on each `docker compose down`/`docker compose up` combination). 

## What is the new behavior?

**WARN** needs: https://github.com/supabase/postgres/pull/901

The `pgsodium` key `/etc/postgresql-custom/pgsodium/pgsodium_root.key` is cached in `/volumes/db/pgsodium`

## Additional context

NEEDS: https://github.com/supabase/postgres/pull/901

~~See https://github.com/supabase/vault/issues/27#issuecomment-1984187441 for the `:Z` flag requirement.~~

### DIFF
--- a/apps/docs/content/guides/api/api-keys.mdx
+++ b/apps/docs/content/guides/api/api-keys.mdx
@@ -20,10 +20,10 @@ The keys are both long-lived JWTs. If you decode these keys, you will see that t
 
 ## The `anon` key
 
-The `anon` key has very few privileges. You can use it in your [RLS policies](/docs/guides/auth/row-level-security) for "anonymous" access. For example, this policy will allow access to the `profiles` table:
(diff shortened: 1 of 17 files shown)


sweatybridge (approved):

nbarrow-inspire-labs:
closing due to this not fixing the issue on clean installs

nbarrow-inspire-labs:
@sweatybridge hold off on this, noticed a bug. Will update momentarily.

nbarrow-inspire-labs:
@sweatybridge and others, we would need something like https://github.com/supabase/postgres/pull/901 before this would work (it would need modifications as I did here)

nbarrow-inspire-labs:
Okay I think this would work, as updated, but it needs https://github.com/supabase/postgres/pull/901 first, or something of the like.

nbarrow-inspire-labs:
Thanks for getting to this @sweatybridge , your fix was a lot more elegant than mine.

PR #21955, opened 2024-03-11
Docs: Hybrid search

gregnr:
Adds docs for hybrid search (combining keyword search with semantic search).

Also adds [`remark-math`](https://github.com/remarkjs/remark-math) and [`rehype-katex`](https://github.com/remarkjs/remark-math/tree/main/packages/rehype-katex) markdown plugins in order to render math equations ([used](https://docs-git-docs-hybrid-search-supabase.vercel.app/docs/guides/ai/hybrid-search#smoothing-constant-k) in this guide).

### Preview
https://docs-git-docs-hybrid-search-supabase.vercel.app/docs/guides/ai/hybrid-search


apps/docs/content/guides/ai/hybrid-search.mdx
```diff
@@ -0,0 +1,228 @@
...
+language sql
+as $$
+with full_text as (
+  select
+    id,
+    row_number() over() as rank_ix
+  from
+    documents
+  where
+    fts @@ websearch_to_tsquery(query_text)
```
gregnr, line 104: Changed `query_text` to a `text` type so that it can easily be called from the Supabase client libs (not possible to pass in a `tsvector` directly). The consequence is we have to choose a wrapping function - I'm choosing `websearch_to_tsquery(query_text)` here as the safest default for web apps.

@olirice Any performance concerns calling `websearch_to_tsquery` here and below in the `order by` (ie. is it cached)?

apps/docs/content/guides/ai/hybrid-search.mdx
```diff
@@ -0,0 +1,228 @@
...
+  order by
+    -- Note: this is not indexable but will only rank matches of the where clause
+    -- which shouldn't be too big
+    ts_rank_cd(fts, websearch_to_tsquery(query_text)) desc
+  limit match_count * 2
```
gregnr, line 109: I chose a limit of 2x `match_count`. Let me know if you think this should default to something else.

apps/docs/content/guides/ai/hybrid-search.mdx
```diff
@@ -0,0 +1,228 @@
...
+
+This method ensures that items that are ranked high in multiple lists are given a high rank in the final list. It also ensures that items that are ranked high in only a few lists but low in others are not given a high rank in the final list. Placing the rank in the denominator when calculating score (ie. reciprocal rank) helps penalize the low ranking records.
+
+### Smoothing constant `k`
+
+To prevent extremely high scores for items that are ranked first (since we're dividing by the rank), a `k` constant is often added to the denominator to smooth the score:
+
+$\dfrac{1}{k+rank}$
```
gregnr, line 43: Here is a [KaTeX](https://katex.org/) example (supported via the new markdown plugins). Fyi, Notion uses KaTeX as well for their math equations, so these can be copy-pasted into guides now.

apps/docs/content/guides/ai/hybrid-search.mdx
```diff
@@ -0,0 +1,228 @@
...
+from
+  full_text
+  full outer join semantic
+    on full_text.id = semantic.id
+  join documents
+    on coalesce(full_text.id, semantic.id) = documents.id
+order by
+  coalesce(1.0 / (rrf_k + full_text.rank_ix), 0.0) * full_text_weight +
+  coalesce(1.0 / (rrf_k + semantic.rank_ix), 0.0) * semantic_weight
+  desc
```
gregnr, line 132: I added optional weight arguments to the function (defaults to 1 for each). They allow you to modify how much weight semantic search gets vs fts.

apps/docs/content/guides/ai/hybrid-search.mdx
olirice: What you have currently is best

Here are the 3 options I tried
Screenshot 2024-03-12 at 11 02 49 AM
Screenshot 2024-03-12 at 11 02 09 AM
Screenshot 2024-03-12 at 11 01 15 AM

apps/docs/content/guides/ai/hybrid-search.mdx
```diff
@@ -0,0 +1,232 @@
...
+id: 'ai-hybrid-search'
+title: 'Hybrid search'
+description: 'Combine keyword search with semantic search to get both direct and contextual results.'
+subtitle: 'Combine keyword search with semantic search.'
+sidebar_label: 'Hybrid search'
+---
+
+Hybrid search combines [full text search](/docs/guides/database/full-text-search) (searching by keyword) with [semantic search](/docs/guides/ai/concepts#what-are-embeddings) (searching by meaning) in order to identify results that are both directly and contextually relevant to the user's query.
+
+In Postgres, “search” just means filtering records in a table using a search query.
```
gregnr, line 11: @charislam how critical are these Vale errors? It seems like the issues it's picking up are actually better the other way around (it changes meaning subtly). Do you have any strong opinions on these suggestions?

charislam: Errors will be blocking one day, use your judgment for warnings and suggestions. In this case, I actually agree with Vale:

- If you're providing a definition for `search` in Postgres, then the `just` serves no purpose
- If you're saying `just` to mean "in comparison to some other, more complex definition of search," it's clearer to make the explicit comparison

charislam: But I'm going to remove the `just` error actually, there are other use cases for `just` so it can get annoying

apps/docs/content/guides/ai/hybrid-search.mdx
```diff
@@ -0,0 +1,232 @@
...
+
+In RRF, we iterate over each record and assign a score (noting that each record could exist in one or both lists). The score is calculated as 1 divided by that record's rank in each list, summed together between both lists. For example, if a record with an ID of `123` was ranked third in the keyword search and ninth in semantic search, it would receive a score of $$\dfrac{1}{3} + \dfrac{1}{9} = 0.444$$. If the record was found in only one list and not the other, it would receive a score of 0 for the other list. The records are then sorted by this score to create the final list. The items with the highest scores are ranked first, and lowest scores ranked last.
+
+This method ensures that items that are ranked high in multiple lists are given a high rank in the final list. It also ensures that items that are ranked high in only a few lists but low in others are not given a high rank in the final list. Placing the rank in the denominator when calculating score (ie. reciprocal rank) helps penalize the low ranking records.
```
charislam, line 37:
```suggestion
This method ensures that items that are ranked high in multiple lists are given a high rank in the final list. It also ensures that items that are ranked high in only a few lists but low in others are not given a high rank in the final list. Placing the rank in the denominator when calculating score helps penalize the low ranking records.
```

charislam (approved):

apps/docs/content/guides/ai/hybrid-search.mdx
gregnr: Got it. The point of the sentence was to help demystify what we actually me by "search" in the context of Postgres - mainly for developers who have less experience on the SQL side (search can be a bit of a black box on other platforms). If there's a concern with the word `just` I think I'll just remove the entire sentence.

PR #21958, opened 2024-03-12
feat: add transition-safe-* tailwind utilities

charislam:
Add Tailwind utilities for transitions that respect `prefers-reduced-motion`.


MildTomato:
can we not use [this](https://tailwindcss.com/docs/transition-property#prefers-reduced-motion)?

MildTomato (approved):

PR #21961, opened 2024-03-12
Chore/misc fixes 120324

joshenlim:


apps/studio/components/interfaces/Home/Connect/ConnectTabContent.tsx
```diff
@@ -40,10 +40,9 @@ const ConnectTabContentNew = ({ projectKeys, filePath }: ConnectContentTabProps)
           usePoolerConnection: false,
         })
       : { uri: '' }
-  const connectionStringPooler =
-    filePath !== 'drizzle'
-      ? connectionStringsPooler.uri.replace('6543', '5432')
-      : connectionStringsPooler.uri
+  const connectionStringPooler = !['drizzle', 'prisma'].includes(filePath)
```
kevcodez, line 43: I think it would be cleaner if we'd path

connectionStringPoolerTransaction
connectionStringPoolerSession

rather than doing port number replacements (unless I am completely aware of how the different ports works, I won't have any idea of why we're doing random port replacements and what they mean

apps/studio/components/interfaces/Home/Connect/content/prisma/content.tsx
```diff
@@ -23,7 +23,7 @@ const ContentFile = ({ connectionStringPooler, connectionStringDirect }: Content
 DATABASE_URL="${connectionStringPooler}"
 
 # Direct connection to the database. Used for migrations.
-DIRECT_URL="${connectionStringDirect}"
+DIRECT_URL="${connectionStringPooler.replace('6543', '5432')}"
```
kevcodez, line 26: Same here, it would be better if we'd pass in variables that contain the pooler mode and then we just pick the one we want here, instead of random port replacements (who knows what they mean)

kevcodez (approved):

PR #21989, opened 2024-03-13
fix: FormItemLayout width fix, `<Sheet/>` renamed with prop name reverted, started new form story

MildTomato:

- `<FormItemLayout/>` layout prop width fix
- `<Sheet/>` renamed without suffix
- started new Stories for form (WIP)

## What is the new behavior?

## Additional context

### DIFF
--- a/packages/ui-patterns/form/FormLayout2.tsx
+++ b/packages/ui-patterns/form/FormLayout2.tsx
@@ -1,0 +1,434 @@
+import { zodResolver } from '@hookform/resolvers/zod'
+import { Box, FileWarning } from 'lucide-react'
+import { useForm } from 'react-hook-form'
+import {
+  Badge_Shadcn_,
+  Button,
+  Checkbox_Shadcn_,
+  Form_Shadcn_,
+  FormControl_Shadcn_,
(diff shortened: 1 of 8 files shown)


packages/ui-patterns/form/FormLayout2.tsx
```diff
@@ -0,0 +1,434 @@
+import { zodResolver } from '@hookform/resolvers/zod'
```
ivasilov, line 1: I'm guessing this file was committed by mistake or it should be a story?

packages/ui-patterns/form/Page2.stories.tsx
```diff
@@ -0,0 +1,37 @@
+// import { within, userEvent, expect } from '@storybook/test'
```
ivasilov, line 1: Is this also committed by mistake?

packages/ui-patterns/form/FormLayout2.tsx
MildTomato: it is a story. its imported into a story elsewhere.
confusing i know

ivasilov (approved):

PR #21990, opened 2024-03-13
feat: store sql editor charts

jordienr:

when a user changes a chart in one of their queries the state will be stored so that they don't have to set it again every time.


joshenlim:
Found an issue sorry: if you start from a new query - you can’t change the chart axes. only after you refresh, then it works
https://github.com/supabase/supabase/assets/19742402/b857d87e-9e4a-418c-889a-042bbc06fed6

joshenlim (approved):

PR #22005, opened 2024-03-13
ref: Add GitHub configuration/revoke links for better UX

kamilogorek:
On organization "Integrations" tab, users will now see this nice footer, that will point them exactly where they need to go in order to revoke/configure authorization/app.

We do not render this info on a specific project's page.

image

(fixed missing dot from the end of the screenshot)


alaister (approved):
LGTM!

PR #22052, opened 2024-03-15
docs: hybrid-search example for full_text subquery ordered correctly

### LINKED ISSUE #22027: hybrid-search example for full_text subquery doesn't order correctly
# Improve documentation

The subquery to find fts results indexes in a non-deterministic order...

query in question [here](https://github.com/supabase/supabase/blob/6e8538f95756153176dcf786c3c2387093b50826/apps/docs/content/guides/ai/hybrid-search.mdx#L96C1-L107C35)
```
  select
    id,
    row_number() over() as rank_ix
  from
    documents
  where
    fts @@ websearch_to_tsquery(query_text)
  order by
    -- Note: this is not indexable but will only rank matches of the where clause
    -- which shouldn't be too big
    ts_rank_cd(fts, websearch_to_tsquery(query_text)) desc
  limit least(match_count, 30) * 2
```

## Describe the problem
in my db, I found that the `order by` was not being performed before the row_number was being calculated. moving the `order by` into the `row_number` `over()` clause fixed this problem! see screenshots below.

broken example:
Screenshot 2024-03-14 at 11 59 38 AM

fixed:
Screenshot 2024-03-14 at 12 00 21 PM

## Describe the improvement
update the documentation to:

```
  select
    id,
    -- Note: this is not indexable but will only rank matches of the where clause
    -- which shouldn't be too big
    row_number() over(order by ts_rank_cd(fts, websearch_to_tsquery(query_text)) desc) as rank_ix
  from
    documents
  where
    fts @@ websearch_to_tsquery(query_text)  
  limit least(match_count, 30) * 2
```

## Additional context

For whatever reason, I didn't have the same problem with the semantic_search! Might be worth correcting that as well just to be safe.

ikunal-04:

closes #22027 
Docs updated

## What is the new behavior?

## Additional context

### DIFF
--- a/apps/docs/content/guides/ai/hybrid-search.mdx
+++ b/apps/docs/content/guides/ai/hybrid-search.mdx
@@ -95,23 +95,23 @@ as $$
 with full_text as (
   select
     id,
-    -- Note: this is not indexable but will only rank matches of the where clause
+    -- Note: ts_rank_cd is not indexable but will only rank matches of the where clause
     -- which shouldn't be too big
     row_number() over(order by ts_rank_cd(fts, websearch_to_tsquery(query_text)) desc) as rank_ix
   from
     documents
   where
-    fts @@ websearch_to_tsquery(query_text)  
+    fts @@ websearch_to_tsquery(query_text)
+  order by rank_ix
   limit least(match_count, 30) * 2
 ),
 semantic as (
   select
     id,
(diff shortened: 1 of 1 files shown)


gregnr (approved):
Thanks for the contribution @ikunal-04! I've pushed one more commit to this branch to update the `row_number()` ordering on the semantic search CTE as well.

PR #22060, opened 2024-03-16
fix notification button

jordienr:
## before
CleanShot 2024-03-18 at 18 06 38@2x

## after
CleanShot 2024-03-18 at 18 07 27@2x


joshenlim:
Ahh its actually meant to be overlapping, with the warning or critical icon overlapping the inbox icon

jordienr:
> Ahh its actually meant to be overlapping, with the warning or critical icon overlapping the inbox icon

got it, it was just missing `relative` then. I swapped the icons since they're deprecated too.

joshenlim (approved):

Issue #21226, opened 2024-02-13
app.settings.jwt_secret cleared on project pause/resume


rkistner:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When pausing and unpausing a free project, `app.settings.jwt_secret` is cleared. It is only recovered by resetting the secret.

## To Reproduce

1. Create new project - `SELECT current_setting('app.settings.jwt_secret')` works.
2. Pause and unpause project - `SELECT current_setting('app.settings.jwt_secret')` gives error `unrecognized configuration parameter "app.settings.jwt_secret"`.
3. Restart the project - still the same error.
4. Generate a new JWT secret - `SELECT current_setting('app.settings.jwt_secret')` works again.
5. Restart project - still works.

`show app.settings.jwt_secret` is exactly the same.

## Screenshot

![image](https://github.com/supabase/supabase/assets/94081/bcd7fb7e-892b-457a-85d1-0bdfd4f80015)

## Expected behavior

`SELECT current_setting('app.settings.jwt_secret')`  should always work and return the current secret.

## System information

N/A - reproduce directly on the Supabase dashboard.

## Additional context

The functionality is important for validating auth tokens - for example in Edge Functions or in other integrated software. We specifically use this for PowerSync to get seamless authentication with Supabase tokens.

estubmo:
Are you self hosting by any chance?  And have you changed the name of the database?

I believe this is the issue: 
https://github.com/supabase/supabase/blob/6b3420a5fd2fefc56655e327283f1f03ebccb07e/docker/volumes/db/jwt.sql#L4
The database name isn't being set depending on the POSTGRES_DB env var.

rkistner:
In my case it's the hosted service with the default `postgres` database, so not the same issue.

kobiebotha:
Ran into this today too

hysuki:
I was gain same error after restoring the database on supabase cloud.

celikomer:
We are also having the same issue after restarting the DB on Supabase Cloud.

rkistner:
The PR introducing the feature appears to explain the issue, at least for the self-hosted version:

https://github.com/supabase/supabase/pull/17033

> The only downside that this script runs only one time if there is no database data directory.
[...]

PR #22074, opened 2024-03-17
fix: jwt.sql script not running on correct db if non-default name is set

estubmo:

This should fix setting `app.settings.jwt_secret` and `app.settings.jwt_exp` when `POSTGRES_DB` is defined as anything else than  `postgres` for docker compose setups.

During a setup with docker compose the container `supabase-db` will [copy the script to a  docker entrypoint volume](https://github.com/estubmo/supabase/blob/6b3420a5fd2fefc56655e327283f1f03ebccb07e/docker/docker-compose.yml#L378) (Where the container will find and initialize scripts), set settings `app.settings.jwt_secret` and `app.settings.jwt_exp` in the database called `postgres`.

https://github.com/supabase/supabase/issues/21226

## What is the new behavior?

During a setup with docker compose the container `supabase-db` will still copy the 99-jwt.sql to the docker entrypoint volume, but now it will set settings `app.settings.jwt_secret` and `app.settings.jwt_exp` to the **database defined by the `POSTGRES_DB` env variable.**


Hallidayo:
Hi @estubmo - We’re doing a bit of housekeeping and plan to close out PRs that have been inactive for more than 3 months.

PR #22076, opened 2024-03-17
fix: stripe wrapper not working as expected in self hosted

### LINKED ISSUE #20891: Bug Report: Stripe Wrapper Not Working as Expected in Self-Hosted Supabase
Whenstripe.accounts table

Error: SQL identifier cannot be null or undefined
This error prevents me from viewing or interacting with the stripe.accounts table data through the Supabase Studio UI.

Steps to Reproduce
Navigate to Supabase Studio in a web browser.
Log into my self-hosted Supabase instance.
Go to the "Table Editor" and select the stripe.accounts table.
The error message appears, and the table data fails to load.
Expected Behavior
I expect to be able to view and interact with the stripe.accounts table data without encountering client-side exceptions.

Docker Configuration
Here is the relevant output of docker ps, showing the Supabase-related containers in my setup:
Docker Version Number:
supabase/storage-api:v0.46.5
supabase/gotrue:v2.139.1
supabase/postgres-meta:v0.75.0
kong:2.8.1
supabase/studio:20240122-7845baf
supabase/realtime:v2.25.60
supabase/logflare:1.4.0
supabase/postgres:15.1.0.147
darthsim/imgproxy:v3.8.0
timberio/vector:0.28.1
`

tranhoangvuit:

fixes: https://github.com/supabase/supabase/issues/20891

Currently, when we add a new Stripe wrapper, it will throw an exception when you access that table in the Table Editor. This is because, in the self-hosted version, we do not yet support querying a foreign table with a specific ID.

## What is the new behavior?

Update the foreign table endpoint to support querying with a specific ID

image

## Additional context

### DIFF
--- a/apps/studio/pages/api/pg-meta/[ref]/foreign-tables.ts
+++ b/apps/studio/pages/api/pg-meta/[ref]/foreign-tables.ts
@@ -13,6 +13,7 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
 
   switch (method) {
     case 'GET':
+      if (req.query.id) return handleGetOne(req, res)
       return handleGetAll(req, res)
     default:
       res.setHeader('Allow', ['GET'])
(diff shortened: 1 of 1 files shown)


alaister (approved):
LGTM!

PR #22096, opened 2024-03-18
feat: email provider aws notify alert

### LINKED ISSUE #22054: Additional Information to help setup with AWS SES
# Improve documentation

## Link 

https://supabase.com/docs/guides/auth/auth-smtp

## Describe the problem

AWS SES has some additional requirements that can be confusing to those used to services like Resend or Sendgrid, such as:
- being placed in a testing sandbox initially and needing to request to leave
- requiring a verified domain

## Describe the improvement

We can add this information as a callout when using SES, with links to AWS documentation, to catch these issues early if people run into them

## Additional context

Add any other context or screenshots that help clarify your question.

Hallidayo:

Supabase Studio & Docs

If users are using AWS SES for an smtp provider some extra config may be needed and this is currently not highlighted to the user in the docs or studio

## What is the new behavior?

If AWS is used in the provider host name then an alert appears:

https://github.com/supabase/supabase/assets/22655069/3c41afe2-6600-4bcb-b540-86f5e9b171e0

## Additional context

Closes #22054


saltcod:
Thanks @Hallidayo — do you know what the extra config is? 

Wondering if: 
—  its short, should we add it to this actual Alert?
— it's long, we add it to somewhere in the docs and link to that instead?  

I'd prefer to not send users out to AWS documentation without some more info first.

Hallidayo:
@saltcod, @encima  - Just added a warning into the docs as well:

Screenshot 2024-03-24 at 16 29 23


What do you guys think? Is it better to just have the warning in the docs rather than the studio or both in place?

PR #22097, opened 2024-03-18
feat: alert for white space on field names

### LINKED ISSUE #22089: Row names are saved with whitespaces. There is no trimming.
Saving a table with a whitespace in a row name is allowed. This causes a lot of confusion while trying to use that table and row, not realizing I accidentally put a space in  the name,

Is this intentional? If not, this looks like a bug.

Only tried via the UI.

Hallidayo:

Supabase Studio - Table Editor

From the table editor you can add/update column names with an starting or ending white space which can cause issues when querying.

## What is the new behavior?

An error now appears which does not allow the user to do this.

https://github.com/supabase/supabase/assets/22655069/0d91b799-d242-451e-9b5c-4b5454831482

## Additional context

Closes #22089

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.utils.ts
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.utils.ts
@@ -21,6 +21,9 @@ export const validateFields = (field: TableField) => {
   if (some(field.columns, (column: ColumnField) => column.name.length === 0)) {
     errors['columns'] = 'Ensure that all your columns are named'
   }
+  if (some(field.columns, (column: ColumnField) => /^\s|\s$/.test(column.name))) {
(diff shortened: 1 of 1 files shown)


encima (approved):
The check makes sense. Is there a reason you chose not to automatically strip the whitespace and alert the user on the action that was taken?

encima (changes requested):
On reflection, it is probably better to fix and alert rather than alert as I cannot think of a use case where leading or trailing whitespace in tables would be wanted

onurusluca:
@encima That was my thought also. Never heard of a case where you'd need whitespace on a row name. It will just be headache to know where is the problem because API says "Row `application_logs ` doesn't exist.

joshenlim:
yeap agreed here that white spaces should just be automatically trimmed - we do that in several parts of the dashboard too like table names, bucket names etc

joshenlim:
Closing in favour of https://github.com/supabase/supabase/pull/22170

PR #22104, opened 2024-03-19
Chore/fix database tables add foreign key not working

joshenlim:


apps/studio/components/interfaces/TableGridEditor/DeleteConfirmationDialogs.tsx
```diff
@@ -152,14 +148,15 @@ const DeleteConfirmationDialogs = ({
...
+    if (project === undefined) return
 
     const selectedColumnToDelete = snap.confirmationDialog.column
     if (selectedColumnToDelete === undefined) return
 
     deleteColumn({
       id: selectedColumnToDelete.id,
       cascade: isDeleteWithCascade,
-      projectRef: projectRef!,
+      projectRef: project?.ref,
```
alaister, line 159: This so doesn't matter but I can't unsee... 🙈

because there's `if (project === undefined) return` above, this can be this right?:

```suggestion
      projectRef: project.ref,
```

alaister (approved):
LGTM!

PR #22119, opened 2024-03-20
Fix consistency for RLS in table editor

### LINKED ISSUE #22110: Dashboard allows enabling RLS on auth schema but not disabling
# Bug report

## Describe the bug

The dashboard lets me enable RLS for tables in the auth schema, but doesn't let me add policies or turn RLS off without going through the SQL Editor.

## To Reproduce

1. Go to Table Editor,  switch to auth and click Enable RLS

2. See lack of Disable RLS/New Policy buttons on Policies page.

3. Enable RLS button turns into Add New Policy button which just directs to the Policies page. 

## Expected behavior
1. Not possible to enable RLS on auth if it's not actually allowed through the dashboard.
2. Better UI would be for the Enable RLS button to turn into a Disable RLS button. Additional button appears to Add New Policy if enabled.

## Screenshots
Screenshot 2024-03-19 081414

Screenshot 2024-03-19 081448

Screenshot 2024-03-19 081507 

## System information
N/A

joshenlim:
Resolves https://github.com/supabase/supabase/issues/22110

We have guard rails in place across the dashboard to prevent editing system schemas via the dashboard GUI - places with these guard rails have this notice as such:
image

We previously added an indication in the Table Editor to show the RLS status of the table, and it also served as a convenient way to enable RLS from there without having to go to the auth policies page. However, we didn't have the guardrails in place here, so this PR fixes that

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
@@ -18,6 +18,7 @@ import { useDatabasePublicationUpdateMutation } from 'data/database-publications
 import { useTableUpdateMutation } from 'data/tables/table-update-mutation'
 import { useCheckPermissions, useIsFeatureEnabled } from 'hooks'
 import { RoleImpersonationPopover } from '../RoleImpersonationSelector'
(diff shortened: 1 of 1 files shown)


alaister (approved):
LGTM!

saltcod (approved):

PR #22123, opened 2024-03-20
Fix applying table privileges changes to wrong table if there are multiple tables with same name in different schema

### LINKED ISSUE #21244: Column Level Security bug when table `public.users` references `auth.users`
# Bug report

## Describe the bug

After changing the column level privileges of different columns of a table on supabase UI and clicking on `Apply Changes`, nothing happens.

https://github.com/supabase/supabase/assets/23220309/0a1e0439-c4e6-4195-b3d1-648b46445f4e

## To Reproduce

1. Create a `public.users` table

These are the columns I created:

image

Foreign key:

image

## Expected behavior

I expected that Column Level Security would work on my `public.users` table. It works only if I rename it to `public.profiles`. Maybe the issue lies because I have two tables named `users` on both `auth` and `public`  schemas. But I wasn't expecting to have any problem with that. Changing `public.users` to `public.profiles` solved the problem.

## System information

- OS: Arch Linux 
- Browser (if applies): chrome

joshenlim:
Resolves https://github.com/supabase/supabase/issues/21244


alaister (approved):
LGTM!

PR #22154, opened 2024-03-21
change clip-path attribute to clipPath on all SVGs

clarencepenz:

Clear terminal warnings `clip-path`

`clip-path` warnings in terminal

## What is the new behavior?

`clip-path` warnings cleared by changing it to `clipPath`


clarencepenz:
Hey @saltcod made the changes on a new PR and [closed](https://github.com/supabase/supabase/pull/20359) the old one.

fsansalvadore:
@clarencepenz thanks for the PR!

Updating clip-rule isn't necessary for svgs that are used as static images.
So all the svgs that live in the `public` and static files folders.

Can you please narrow down which svgs are actually causing the warnings in the terminal and show a before and after screenshot of the warnings?

PR #22170, opened 2024-03-22
Add trims to table and column names

### LINKED ISSUE #22089: Row names are saved with whitespaces. There is no trimming.

joshenlim:
Resolves https://github.com/supabase/supabase/issues/22089


encima (approved):

alaister (approved):
LGTM!

PR #22184, opened 2024-03-23
update functions doc for dart v2.0

flikkr:

YES

Doc update

Doc mentions `ResponseType` in v2.0 dart library. However is does not exist anymore and now we have `HttpMethod`.

## What is the new behavior?

Remove `ResponseType` and introduce `HttpMethod`. 

## Additional context

N/A

### DIFF
--- a/apps/docs/spec/supabase_dart_v2.yml
+++ b/apps/docs/spec/supabase_dart_v2.yml
@@ -1068,17 +1068,17 @@ functions:
           final res = await supabase.functions.invoke('hello', body: {'foo': 'baa'});
           final data = res.data;
           ```
-      - id: specifying-response-type
-        name: Specifying response type.
+      - id: specifying-method-type
+        name: Specifying method type.
         description: |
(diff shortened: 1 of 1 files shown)


apps/docs/spec/supabase_dart_v2.yml
```diff
@@ -1068,17 +1068,17 @@ functions:
           final res = await supabase.functions.invoke('hello', body: {'foo': 'baa'});
           final data = res.data;
           ```
-      - id: specifying-response-type
-        name: Specifying response type.
+      - id: specifying-method-type
+        name: Specifying method type.
         description: |
-          By default, `invoke()` will parse the response as JSON. You can parse the response in the following formats: `json`, `blob`, `text`, and `arrayBuffer`.
+          By default, `invoke()` will make a POST request. You can parse the response in the following formats: `get`, `post`, `put`, `delete`, and `patch`.
```
charislam, line 1074: @dshukertjr, could you take a look please? The statement that you can 'parse' a result as an HTTP method sounds a bit odd to me...

PR #22208, opened 2024-03-25
Update docs with replication role creation

encima:

YES

Docs update for logical replication

Current docs restrict replication to `postgres` user

## What is the new behavior?

Changes now allow the `postgres` user to create roles with the replication privilege so that a privileged role does not need to be used

## Additional context

This may benefit from expanding further to cover permissions but I opted not to as replication is possible but not officially supported / covered


encima:
@supabase/docs

charislam (approved):

PR #22227, opened 2024-03-26
Add validations for new RLS editor

joshenlim:


apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/index.tsx
```diff
@@ -308,6 +309,14 @@ export const AIPolicyEditorPanel = memo(function ({
     const using = editorOneRef.current?.getValue().trim() ?? undefined
     const check = editorTwoRef.current?.getValue().trim() ?? undefined
 
+    if (command === 'insert' && (check === undefined || check.length === 0)) {
+      return setFieldError('Please provide a SQL expression for the WITH CHECK statement')
+    } else if (command !== 'insert' && (using === undefined || using.length === 0)) {
+      return setFieldError('Please provide a SQL expression for the USING statement')
```
alaister, line 315: I think it's possible to create an `update` policy with a `with check` only and no `using):

```
create policy "Enable update for users based on email" on "public"."other" as PERMISSIVE
for UPDATE
  to public
with
  check (auth.jwt () ->> 'email' = email);
```

alaister: but this may not be an issue if we don't want to allow this from the ui

joshenlim: ahh hmm okay let me fix this in another PR - i'll need to change the behaviour of the editor then
this was also the behaviour of the old UI so i think its not too bad

alaister (approved):

PR #22263, opened 2024-03-26
Preload RLS form with selected table.

### LINKED ISSUE #22218: cannot add rls in the dashboard
delimiter errors
even when copying straight from the examples

jordienr:
## Issue

- If you go to any table that's not the first one and click "Add RLS policy"
- You get redirected to the policies view
- When you click "Create new policy" the table in the form is not the one you come from.

This should fix it

Resolves https://github.com/supabase/supabase/issues/22218

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
@@ -1,5 +1,6 @@
 import { useEffect } from 'react'
 
+import { useParams } from 'common'
 import { useProjectContext } from 'components/layouts/ProjectLayout/ProjectContext'
 import { useDatabaseRolesQuery } from 'data/database-roles/database-roles-query'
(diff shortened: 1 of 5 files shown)


apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
```diff
@@ -30,6 +31,8 @@ interface PolicyDetailsV2Props {
 export const PolicyDetailsV2 = ({ isEditing, form, onUpdateCommand }: PolicyDetailsV2Props) => {
   const { project } = useProjectContext()
   const snap = useTableEditorStateSnapshot()
+  const router = useRouter()
+  const searchTableId = router.query.search as string
```
joshenlim, line 35: `const { search } = useParams()`

joshenlim: also - realised this will be an odd behaviour if i cleared the UI search input, but it still defaults to the table in the URL search param. will update

joshenlim (approved):

PR #22283, opened 2024-03-27
Fix 'Fix with assistant' for RLS editor

### LINKED ISSUE #22225: RLS editor shows USING clause as a single line
# Bug report

## Describe the bug

When updating a RLS policy the USING clause editor is displayed as a single line, regardless of the line count of clauses or maximum line length in clauses.

## To Reproduce

1. Create and save a RLS policy with USING clause longer than 80 characters and has subquery: ex) `((true AND true AND true AND true AND true AND true AND true AND true AND true AND true) OR (1 IN ( SELECT 1 FROM ( SELECT 1 AS "?column?") f)))`
2. Edit the RLS policy from step 1
3. See error

## Expected behavior

USING clause have more visible height (ex: 8 rows at maximum)

## Screenshots

image

1. The wrapper element for editor have height `28px` and should be something like `18px * 3 + 0.25rem * 2`, where height per line is 18px, 3 is line count 2 + wrapped line 1, and padding-top 0.25rem + padding-bottom 0.25rem. Orange CSS rule is current (2024-03-26) implementation and green one is proof of concept CSS rule. 

2. Line number of the closing parenthesis is `8` and should be `9` (USING clause starts at line number 7 and the clause has line count 2)

## System information

- OS: Windows 10
- Browser (if applies): Chrome 122
- Version of supabase-js: N/A
- Version of Node.js: N/A

## Additional context

1. USING clauses are auto-formatted.
2. Some line in USING clauses can be over 80 characters and cause line wrap
3. Line breaks will be inserted by auto-formatting when RLS policy has subqueries

joshenlim:
Resolves https://github.com/supabase/supabase/issues/22225

Just FYI i'm opting to _not_ auto format the query from the user, and instead provide a way to format the query by right clicking into the editor

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/QueryError.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/QueryError.tsx
@@ -18,6 +18,7 @@ import { useSelectedOrganization } from 'hooks'
 import Telemetry from 'lib/telemetry'
 import { useTelemetryProps } from 'common'
 import { useRouter } from 'next/router'
+import { useIsRLSAIAssistantEnabled } from 'components/interfaces/App/FeaturePreview/FeaturePreviewContext'
 
 const QueryError = ({
   error,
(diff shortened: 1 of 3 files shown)


alaister (approved):
LGTM!

PR #22307, opened 2024-03-28
feat: initial version of @supabase/pg-meta

soedirgo:

YES

feature

We currently route dashboard SQL queries through supabase/postgres-meta, which:
- can't batch queries
- has no pooling
- is an extra hop

## What is the new behavior?

This is an initial effort to move pg-meta into a pure-JS SQL& Zod schema builder, which allows for batching queries and eventually route queries through Supavisor, eliminating one hop for dashboard queries.

The API for the `@supabase/pg-meta` package is not fixed - we can change it into an interface that's more convenient to use for the dashboard.

### DIFF
--- a/packages/pg-meta/package.json
+++ b/packages/pg-meta/package.json
@@ -1,0 +1,16 @@
+{
+  "name": "@supabase/pg-meta",
+  "version": "0.0.0",
+  "license": "MIT",
+  "author": "Supabase",
+  "main": "./src/index.ts",
+  "repository": "supabase/supabase",
+  "dependencies": {
+    "pg-format": "^1.0.4",
+    "zod": "^3.22.4"
+  },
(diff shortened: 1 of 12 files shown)


packages/pg-meta/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "author": "Supabase",
+  "main": "./src/index.ts",
+  "repository": "supabase/supabase",
+  "dependencies": {
+    "pg-format": "^1.0.4",
+    "zod": "^3.22.4"
+  },
+  "devDependencies": {
+    "@types/pg-format": "^1.0.1",
+    "typescript": "^5.2.2"
```
kevcodez, line 14: We use TS 5.4.x across the code base

alaister (approved):
LGTM!

PR #22330, opened 2024-03-29
Read replicas support enabling physical backups

joshenlim:
If physical backups are not enabled yet:
image

While enabling physical backups: (Purely browser state, will revert to the above if user refreshes browser prior to completion of enabling physical backups since the process is async)
Screenshot 2024-03-29 at 15 21 47

Also add missing warning for compute size:
image


apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureConfiguration/DeployNewReplicaPanel.tsx
```diff
@@ -39,12 +41,33 @@ const DeployNewReplicaPanel = ({
...
   const project = useSelectedProject()
   const org = useSelectedOrganization()
+  const [refetchInterval, setRefetchInterval] = useState<number | boolean>(false)
 
   const { data } = useReadReplicasQuery({ projectRef })
-  const { data: backups } = useBackupsQuery({ projectRef })
+  const { data: backups } = useBackupsQuery(
+    { projectRef },
+    {
+      refetchInterval: refetchInterval as any,
```
alaister, line 50:
```suggestion
      refetchInterval,
```

apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureConfiguration/DeployNewReplicaPanel.tsx
```diff
@@ -39,12 +41,33 @@ const DeployNewReplicaPanel = ({
   const { ref: projectRef } = useParams()
   const project = useSelectedProject()
   const org = useSelectedOrganization()
+  const [refetchInterval, setRefetchInterval] = useState<number | boolean>(false)
```
alaister, line 44:
```suggestion
  const [refetchInterval, setRefetchInterval] = useState<number | false>(false)
```

alaister (approved):
LGTM! One small TS thing:

PR #22381, opened 2024-04-01
fix: better handling of getSession for sveltekit

charislam:

YES

Enhancement

Docs are not entirely clear on best practices for using `getSession` for SvelteKit

## What is the new behavior?

Docs demonstrate best practices for security and ensuring you never depend on a unchecked access token or user object

## Additional context


mansueli (approved):

probablykasper:
@charislam How come this is the recommended way (adding latency as a workaround for a security hole), instead of having the JWT be validated internally in `getSession` with a public key (which could be in the anon key)?

PR #22405, opened 2024-04-02
Small error in the docs. Probably due to the switch away from auth-helpers-nextjs

lostintime101:

YES

Simple docs tutorial update.


lostintime101:
> seems to me that this should work given Next.js's default aliasing 🤔
> 
> this is the way it's done in the `with-supabase` template, and that worked for me out of the box when I tested yesterday. Is it not working for you?

I have to change it to get it to work. I'm using NextJS 13.4.4

charislam:
hmm can't seem to replicate (and would imagine either none of the `@` aliases work, or all of them work). will leave this open for now in case someone can reproduce it

PR #22410, opened 2024-04-02
fix(web): add cancel button onClick event in ConfirmationModal component

xiaodong2008:

YES

This PR fix a bug of UI in the web(www)

When the user presses the cancel button, there is no any response.

## What is the new behavior?

After add a correct event callback, it works normally.


dy-xiaodong2022:
This bug was fixed in another pull request #22417 today.

PR #22436, opened 2024-04-03
Fix sveltekit hooks.server.ts example

thibaultdalban:
The `supabase` variable doesn't exist in this context, but an instance is accessible through the `event.locals`.

YES

docs update

Invalid example

## What is the new behavior?

Fix the example


charislam (approved):

Tihomir971:
@thibaultdalban In +layout.ts we need to return "user" also. Error from typescript.
- Do we need last const {
		data: { session }
	} = await supabase.auth.getSession(); ?
We have session in data

PR #22440, opened 2024-04-03
add anon users filter

jordienr:

YES

adds a filter for anon users to the user list table


joshenlim (dismissed):
Can we shift the users filtering logic to API instead? So that we just have one place for the filtering logic, otherwise we'll have dupe logic like pagination and all that

joshenlim:
pending on infra PR [here](https://github.com/supabase/infrastructure/pull/17668)

alaister (approved):

PR #22500, opened 2024-04-05
doc: Add Realtime Authorization Demo

filipecabaco:

Add the Realtime Authorization Demo


examples/realtime/nextjs-authorization-demo/README.md
```diff
@@ -0,0 +1,161 @@
...
+);
+```
+
+## Coding concerns
+### Need to create channels
+At the moment if you are using Realtime channels you would just connect and go. If you want your channel to respect the RLS rules setup above you will need to preemptively create the channel that will be used before connecting.
+
+This can be done with a quick API call:
+```ts
+await supabase.realtime.createChannel(name)
```
w3b6x9, line 154: I thought we were using `next` version? Is it already in supabase-js? Did you just locally link realtime-js next or something?

w3b6x9 (commented):
i've only just begun but thought i'd publish some comments now

examples/realtime/nextjs-authorization-demo/README.md
```diff
@@ -0,0 +1,161 @@
...
+    FROM rooms_users ru
+    WHERE ru.user_id = auth.uid()
+    AND ru.name = realtime.channel_name()
+  )
+);
+```
+
+## Coding concerns
+### Need to create channels
+At the moment if you are using Realtime channels you would just connect and go. If you want your channel to respect the RLS rules setup above you will need to preemptively create the channel that will be used before connecting.
```
w3b6x9, line 150: We should specifically mention Broadcast and Presence and not just channel b/c a channel can use Postgres Changes and that can respect RLS rules.

Referencing "you" here is confusing b/c who's supposed to create it? The developer? Their users? I think in the demo it's the `authenticated` user.

examples/realtime/nextjs-authorization-demo/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+This can be done with a quick API call:
+```ts
+await supabase.realtime.createChannel(name)
+```
+You can check this code at the [create-room-modal.tsx](components/create-room-modal.tsx) component.
+
+### Connecting to broadcast is essentially the same
+The biggest difference from the previous way of work was really the requirement of creating the channel, other than that you can connect as usual to your Realtime channel which might fail to connect if it does not respect the RLS rules set by you.
```
w3b6x9, line 159: Again, should reference Broadcast and Presence specifically.

Should highlight somewhere in this README that there's two ways of creating channels for Broadcast and Presence.

One through client.channel() and the new way from client.createChannel(). It can be quite confusing.

examples/realtime/nextjs-authorization-demo/README.md
```diff
@@ -0,0 +1,161 @@
...
+grant execute on function insert_user () to supabase_auth_admin;
+```
+
+### RLS Rules
+We do have a lot of rules setup so we will detail them with schema.table in the section below.
+
+The most noteworthy one will be the realtime.broadcast as it will be the one used to limit access to rooms by checking if an entry for a given room name and user id exists in the table public.rooms_users.
+
+> ⚠️ All the rules here are not the most secure, you should check your use case and refine them!
+#### public.profiles
```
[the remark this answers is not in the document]
w3b6x9: This is a good addition: https://github.com/supabase/supabase/blob/7617d45609dae7dabd270b19016c9cbc1f628271/examples/realtime/nextjs-authorization-demo/schema.png

but sql will help get devs up and running faster.

examples/realtime/nextjs-authorization-demo/README.md
filipecabaco: added a mention that we're overriding the supabase-js realtime-js dependency

examples/realtime/nextjs-authorization-demo/README.md
filipecabaco: this can actually impact postres_changes and in theory we could even improve a bit the performance of postgres_changes if we tell users they can run some of those queries on connect 👀 

Added a warning to mention it

w3b6x9 (approved):

PR #22513, opened 2024-04-05
example code fix inside redirect urls MDX

yashug:

YES

docs update


charislam:
Hmm I believe this would fail when using localhost, you would end up with `https://http://localhost:3000`.

To enforce secure HTTP, you would need to rule out localhost, and also check for the condition where the underlying URL already has 'http' (no 's') on it to overwrite that.

Do you want to look into this? Otherwise feel free to open an issue to that effect and we can get help from the rest of the community.

charislam (approved):

joshenlim:
Checked on latest master - this fix seems to already be present in a diff syntax [here](https://github.com/supabase/supabase/blob/master/apps/docs/content/guides/auth/redirect-urls.mdx?plain=1#L67)

will thus be closing this PR but please do feel free to reopen if i'm wrong here! 🙏

PR #22532, opened 2024-04-06
add env variables for Supabase AI in self-hosted version

ajayvignesh01:

YES

Adds Open AI API Key environment variable to .env and docker-compose.yml so that people know to add their keys in order to get Supabase AI working on the self-hosted version.

Trying to use Supabase AI in the SQL Editor for example, returns a long error that isn't too useful.

## What is the new behavior?

Supabase AI works across the self-hosted application

## Additional context

For some reason, the studio docker service requires both OPENAI_API_KEY & OPENAI_KEY to function. They can both point to the same value though.

Was also thinking of mentioning this in the docs, but wasn't sure about the best place to do it among the current sections: Before you begin, Running Supabase, Securing your services.


kiwicopple (approved):

ivasilov (approved):

charislam:
Thanks so much for this! If you want to make a docs update, I'd suggest under "Running Supabase", where the env file is copied

PR #22536, opened 2024-04-07
feat: add Nano instance to pricing page

w3b6x9:

YES

Pricing update

Pricing page doesn't include the new Nano compute instance that was added to docs: https://supabase.com/docs/guides/platform/compute-add-ons

## What is the new behavior?

- Pricing page includes mention and specs of Nano compute instance.
- Includes an update to the Compute pricing calculator where credits are not applied if all compute add ons are on the `Nano (Free)` but the moment this changes then credits kick in:

https://github.com/supabase/supabase/assets/5532241/f0c0638f-c89d-4344-8238-e66096656500

### DIFF
--- a/apps/www/components/Pricing/ComputePricingTable.tsx
+++ b/apps/www/components/Pricing/ComputePricingTable.tsx
@@ -28,20 +28,7 @@ const ComputePricingTable = () => {
           <tbody>
             {pricingAddOn.database.rows.map((row, i) => (
               <Fragment key={`row-${i}`}>
-                {i === 0 && (
-                  <tr>
-                    <td className="pb-1 bg-border-strong px-3 py-1 -mr-1 border-l-4 border-strong">
-                      <span>First instance is free on paid plans</span>
-                    </td>
-                  </tr>
-                )}
-                <tr
-                  key={i}
-                  className={cn(
-                    i % 2 === 0 ? 'bg-surface-100 rounded-lg' : '',
-                    i === 0 ? 'border-4 border-strong' : ''
-                  )}
-                >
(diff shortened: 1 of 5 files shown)


apps/www/components/Pricing/ComputePricingTable.tsx
```diff
@@ -28,20 +28,7 @@ const ComputePricingTable = () => {
...
-                    </td>
-                  </tr>
-                )}
-                <tr
-                  key={i}
-                  className={cn(
-                    i % 2 === 0 ? 'bg-surface-100 rounded-lg' : '',
-                    i === 0 ? 'border-4 border-strong' : ''
-                  )}
-                >
```
w3b6x9, line 44: Took this out b/c it looked weird with the introduce of the Nano instance in the row above. Also seemed redundant since there's mention of $10 of compute credits which is applied to make the first Micro instance free.

Here's what it would look like if it's kept in:

Screenshot 2024-04-06 at 17 54 45

apps/www/data/PricingAddOnTable.json
```diff
@@ -8,6 +8,25 @@
...
+            "title": "Connections: Direct",
+            "value": "60"
+          },
+          {
+            "key": "poolerConnections",
+            "title": "Connections: Pooler",
+            "value": "200"
+          }
+        ]
+      },
```
w3b6x9, line 29: This mirrors the Nano (Free) from the Compute Add-ons page: https://supabase.com/docs/guides/platform/compute-add-ons

kevcodez (changes requested):
Let's hold this back for now

PR #22567, opened 2024-04-08
fix: docs ssr build

alaister:

Bug fix

There are a couple of references to `window` in the docs, which is causing the build to fail.

## What is the new behaviour?

Guards those references so they work server side.


ivasilov (approved):

ivasilov:
There's already a fix in https://github.com/supabase/supabase/pull/22523

PR #22568, opened 2024-04-08
Query Performance Facelift

joshenlim:
Before:
image

After:
image


apps/studio/components/interfaces/QueryPerformanceV2/QueryPerformance.tsx
```diff
@@ -0,0 +1,201 @@
...
+    <>
+      <Tabs_Shadcn_
+        defaultValue={page}
+        onValueChange={(value) => {
+          setPage(value as QUERY_PERFORMANCE_REPORT_TYPES)
+          const { sort, search, ...rest } = router.query
+          router.push({ ...router, query: { ...rest, preset: value } })
+        }}
+      >
+        <TabsList_Shadcn_ className={cn('flex gap-0 border-0 items-end')}>
```
ivasilov, line 108: A better approach would be to use `NavMenu` and `NavMenuItem`. They're already used in https://supabase.com/dashboard/project/_/settings/vault/keys.

apps/studio/components/interfaces/QueryPerformanceV2/QueryPerformance.tsx
```diff
@@ -0,0 +1,201 @@
...
+}
+
+export const QueryPerformance = ({
+  queryHitRate,
+  queryPerformanceQuery,
+}: QueryPerformanceProps) => {
+  const router = useRouter()
+  const { preset } = useParams()
+  const { project } = useProjectContext()
+  const [page, setPage] = useState<
```
ivasilov, line 41: You can just use `useState<QUERY_PERFORMANCE_REPORT_TYPES>(` here, I think.

apps/studio/components/interfaces/QueryPerformanceV2/QueryPerformance.tsx
```diff
@@ -0,0 +1,201 @@
...
+    | QUERY_PERFORMANCE_REPORT_TYPES.MOST_TIME_CONSUMING
+    | QUERY_PERFORMANCE_REPORT_TYPES.MOST_FREQUENT
+    | QUERY_PERFORMANCE_REPORT_TYPES.SLOWEST_EXECUTION
+  >(
+    (preset as QUERY_PERFORMANCE_REPORT_TYPES) ?? QUERY_PERFORMANCE_REPORT_TYPES.MOST_TIME_CONSUMING
+  )
+  const [showResetgPgStatStatements, setShowResetgPgStatStatements] = useState(false)
+
+  const handleRefresh = async () => {
+    queryPerformanceQuery.runQuery()
```
ivasilov, line 51: Should these functions have `await`?

apps/studio/components/interfaces/QueryPerformanceV2/QueryPerformance.constants.ts
```diff
@@ -0,0 +1,37 @@
...
+    { id: 'query', name: 'Query', description: undefined, minWidth: 600 },
+    { id: 'rolname', name: 'Role', description: undefined, minWidth: undefined },
+    { id: 'avg_rows', name: 'Avg. Rows', description: undefined, minWidth: undefined },
+    { id: 'calls', name: 'Calls', description: undefined, minWidth: undefined },
+    { id: 'max_time', name: 'Max time', description: undefined, minWidth: undefined },
+    { id: 'mean_time', name: 'Mean time', description: undefined, minWidth: undefined },
+    { id: 'min_time', name: 'Min time', description: undefined, minWidth: undefined },
+    { id: 'total_time', name: 'Total time', description: 'latency', minWidth: 180 },
+  ],
+}
```
ivasilov, line 35: Add `as const` here for extra type safety and great VSCode suggestions.

apps/studio/components/interfaces/QueryPerformanceV2/QueryPerformance.constants.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export type QueryPerformancePreset = 'time' | 'frequent' | 'slowest'
```
ivasilov, line 37: This is unused?

apps/studio/data/database/index-advisor-enable-mutation.ts
```diff
@@ -0,0 +1,241 @@
...
+import type { ResponseError } from 'types'
+
+// [Joshen] This is just temporary while we wait for Index Advisor to be baked into the image itself
+
+export type IndexAdvisorEnableVariables = {
+  projectRef: string
+  connectionString?: string
+}
+
+export async function enableDatabaseWebhooks({
```
ivasilov, line 14: Copy-paste?

apps/studio/components/interfaces/QueryPerformanceV2/QueryPerformance.tsx
joshenlim: good point - shouldnt need to

apps/studio/data/database/index-advisor-enable-mutation.ts
joshenlim: hahahaha good catch 😛

apps/studio/components/interfaces/QueryPerformanceV2/QueryPerformance.tsx
MildTomato: I disagree, those components do not use tabs correctly

MildTomato: I agree with the concept of wrapping them though, but the NavMenu wrap is in this case the wrong one _I think_ as the radix tabs as primitives would mean you get keyboard "left" and "right" nav

MildTomato: use whatever you like for now @joshenlim @jordienr @ivasilov 
might update it later to different primitives, or offer a Tab a11y version

apps/studio/components/interfaces/QueryPerformance/QueryPerformance.tsx
```diff
@@ -34,14 +34,27 @@ interface QueryPerformanceProps {
   queryPerformanceQuery: DbQueryHook<any>
 }
 
+const QueryPerformanceLoadingRow = ({ colSpan }: { colSpan: number }) => {
+  return (
+    <>
+      {Array(4)
+        .fill('')
+        .map((_, i) => (
+          <tr key={'loading-' + { i }}>
```
alaister, line 43:
```suggestion
          <tr key={'loading-' + i}>
```

apps/studio/components/interfaces/QueryPerformanceV2/IndexSuggestion.tsx
```diff
@@ -0,0 +1,10 @@
+export const IndexSuggestion = () => {
+  return (
+    <div className="h-full overflow-auto flex flex-col gap-y-8 divide-y">
+      <div className="px-4 flex flex-col gap-y-2">
+        <p className="text-sm">Add new index</p>
+        <p>Hello</p>
```
alaister, line 6: Hello?

apps/studio/components/interfaces/QueryPerformanceV2/QueryPerformance.tsx
```diff
@@ -0,0 +1,210 @@
...
+                <div className="absolute top-0 left-0 w-full h-[1px] bg-foreground"></div>
+              )}
+
+              <div className="flex items-center gap-x-2">
+                <span className="">{tab.label}</span>
+                <Tooltip_Shadcn_>
+                  <TooltipTrigger_Shadcn_ asChild>
+                    <InformationCircleIcon className="transition text-foreground-muted w-3 h-3 data-[state=delayed-open]:text-foreground-light" />
+                  </TooltipTrigger_Shadcn_>
+                  <TooltipContent_Shadcn_ side="bottom">{tab.description}</TooltipContent_Shadcn_>
```
alaister, line 131: What do you think about making this `"top"`? That way it won't cover the info text underneath:

Screenshot 2024-04-11 at 11 25 30

```suggestion
                  <TooltipContent_Shadcn_ side="top">{tab.description}</TooltipContent_Shadcn_>
```

apps/studio/components/interfaces/QueryPerformanceV2/QueryPerformance.tsx
```diff
@@ -0,0 +1,210 @@
...
+      id: QUERY_PERFORMANCE_REPORT_TYPES.SLOWEST_EXECUTION,
+      label: 'Slowest execution',
+      description: 'Lists queries ordered by their maximum execution time',
+      isLoading: isLoadingMMF,
+      max:
+        (slowestExecutionTime ?? []).length > 0
+          ? Math.max(...(slowestExecutionTime ?? []).map((x: any) => x.max_time)).toFixed(2)
+          : undefined,
+    },
+  ]
```
alaister, line 92: Might be a pre-optimisation, but since we're using `Math.max()` in here it might be worth wrapping this in a `useMemo()`

apps/studio/data/database/index-advisor-enable-mutation.ts
```diff
@@ -0,0 +1,241 @@
...
+  projectRef: string
+  connectionString?: string
+}
+
+export async function enableIndexAdvisor({
+  projectRef,
+  connectionString,
+}: IndexAdvisorEnableVariables) {
+  const sql = `
+create or replace function index_advisor(
```
alaister, line 19: Is this function used in this PR? Also creating this function in the `public` schema is probably not the best idea since it can be called from postgrest

apps/studio/package.json
```diff
@@ -22,6 +22,7 @@
     "@graphiql/toolkit": "^0.9.1",
     "@hcaptcha/react-hcaptcha": "^1.8.1",
     "@headlessui/react": "^1.7.17",
+    "@heroicons/react": "^2.1.3",
```
alaister, line 25: Don't tell @kevcodez 👀

An icon wasn't available from lucide? If it's a one-off icon it might be better to just copy and paste the SVG rather than adding a whole package. wdyt?

alaister (commented):
Looking good! Just a few small things:

apps/studio/package.json
joshenlim: ^ this one's from jonny - i think the intention is we'll use icons from both lucide and hero perhaps cc @MildTomato

MildTomato: we use both sets

MildTomato: it's not a one off, we have a fair few in use that we should start using the lib

apps/studio/components/interfaces/QueryPerformanceV2/IndexSuggestion.tsx
joshenlim: ^ ahhh yup this will be updated in the next PR once this is merged - cause the UI is also feature flagged so not user-facing just yet

apps/studio/data/database/index-advisor-enable-mutation.ts
joshenlim: so this is just temporary until the index_advisor gets to staging/prod as a database extension - we'd remove this mutation at the end ideally

joshenlim: but yes will be used in the subsequent PR when we integrate the index advisor stuffs

alaister (approved):
LGTM!

PR #22471, opened 2024-04-04
feat: Templates for realtime policies

ivasilov:

PR #22573, opened 2024-04-08
feat: Realtime channel policies

ivasilov:


apps/studio/components/interfaces/Realtime/Policies/RealtimePoliciesChannelRow.tsx
```diff
@@ -0,0 +1,123 @@
...
+}: RealtimePoliciesChannelRowProps) => {
+  return (
+    <Panel
+      title={
+        <div className="flex w-full items-center justify-between">
+          <div className="flex items-center flex-row space-x-4">
+            <Antenna className="text-foreground-light w-4 h-4" />
+            <h4 className="text-lg">{channel.name}</h4>
+          </div>
+          <div className="flex space-x-1">
```
MildTomato, line 82: use `gap`

apps/studio/components/interfaces/Realtime/Policies/RealtimePoliciesChannelRow.tsx
```diff
@@ -0,0 +1,123 @@
...
+  onSelectPolicyEdit,
+  onSelectPolicyDelete,
+}: RealtimePoliciesChannelRowProps) => {
+  return (
+    <Panel
+      title={
+        <div className="flex w-full items-center justify-between">
+          <div className="flex items-center flex-row space-x-4">
+            <Antenna className="text-foreground-light w-4 h-4" />
+            <h4 className="text-lg">{channel.name}</h4>
```
MildTomato, line 80: more hardcoded tailwind going on

apps/studio/components/interfaces/Realtime/Policies/RealtimePoliciesChannelRow.tsx
```diff
@@ -0,0 +1,123 @@
...
+              New presence policy
+            </Button>
+          </div>
+        </div>
+      }
+    >
+      {policies.length === 0 ? (
+        <div className="p-4 px-6">
+          <p className="text-sm text-foreground-lighter">No policies created yet</p>
+        </div>
```
MildTomato, line 99: lots of opportunity to make components for empty states, headers, layouts and son on

apps/studio/components/layouts/RealtimeLayout/RealtimeMenu.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [showCreateChannelModal, setShowCreateChannelModal] = useState(false)
+  // service api key is required for creating a new channel
+  const canCreateChannels = useCheckPermissions(PermissionAction.READ, 'service_api_keys')
+
+  const page = router.pathname.split('/')[4] as undefined | 'policies' | 'inspector'
+
+  return (
+    <>
+      <Menu type="pills" className="my-6 flex flex-grow flex-col px-5">
+        <div className="mb-6 px-2">
```
MildTomato, line 25: as an example of why components matter to team:

this `Menu` code has been repeated so many times across the app. There is random tailwind in here,  `my-6 flex flex-grow flex-col px-5` and  `mb-6 px-2` which is now in every layout, of every section. 

It is now such a pain to update the side panel or update it.

apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
```diff
@@ -104,11 +110,10 @@ export const PolicyDetailsV2 = ({
...
+                <FormLabel_Shadcn_ className="flex items-center gap-x-1">
                   <p className="text-foreground-light text-sm">Table</p>
-                  <p className="text-foreground-light text-sm">
-                    <code className="text-xs">on</code> clause
-                  </p>
+                  <code className="text-xs">on</code>
+                  <p className="text-foreground-light text-sm">clause</p>
                 </FormLabel_Shadcn_>
                 <FormControl_Shadcn_>
                   <Select_Shadcn_
```
MildTomato, line 119: should be using `<FormItemLayout/>`

a lot of random tailwind in here.

joshenlim:
### UX thoughts:
1. It's odd that I don't see my new channel when I created it - I was wondering where my channels were until I checked the policies tab
    - Maybe we list them under the nav menu? clicking on the channel will start listening to the channel perhaps (or at least prefill "join a channel")
    - Then users can edit / delete channels via here too
    - image
2. Compile these under a single dropdown, otherwise there's too many CTAs at first glance
    - image
    - Maybe like - Add new policy: For channel, for broastcast, for presense
    - Probably will do better with some descriptions too, at first glance as a user I wouldn't know what i would create policies on them for (e.g What does it mean to create a policy for broadcast or presence)
4. Might be worth also directing users to the realtime RLS page from auth RLS if they're viewing the realtime schema there
    - image
5. A plus icon might be better here
    - Also as per jonny, may lets call this "persisted"
    - image
7. I'm guessing that users wouldn't have to create permanent channels to listen to them right? (like before) Maybe we should add a description here as to how 'permanent" or "persisted" channels differ from the normal channels (If there is any difference that might be important from the user POV)
    - e.g with permanent channels, you can use RLS to control access etc, but otherwise its anything goes for other channels 
    - image

### Bugs found:
1. Create a new permanent channel
    - There's no actual client side validation to enforce the rules in the input description (e.g I could create a channel named "Frontend Channel")
    - Toast message shows `[object Object]`
       image
2. Unrelated to this PR probably, but the height in the fields here are just slightlyyyyy off, maybe add some padding to the name field to match the height of the dropdown on the right
    - image
3. Create new channel button width seems smaller than the other pages
    - image
    - image

MildTomato:
> Unrelated to this PR probably, but the height in the fields here are just slightlyyyyy off, maybe add some padding to the name field to match the height of the dropdown on the right

let's not hack this. let's fix it in primitives

apps/studio/components/interfaces/Realtime/Policies/RealtimePoliciesChannelRow.tsx
ivasilov: Fixed.

ivasilov:
> ### UX thoughts:
> 
>     1. It's odd that I don't see my new channel when I created it - I was wondering where my channels were until I checked the policies tab
>        
>        * Maybe we list them under the nav menu? clicking on the channel will start listening to the channel perhaps (or at least prefill "join a channel")
>        * Then users can edit / delete channels via here too
>        * image
I don't want to build too much stuff since we're not sure how this UI is going to go, but I've made the `Policies` page first so when you click `add channel` you see in it policies.


>     2. Compile these under a single dropdown, otherwise there's too many CTAs at first glance
>        
>        * image
>        * Maybe like - Add new policy: For channel, for broastcast, for presense
>        * Probably will do better with some descriptions too, at first glance as a user I wouldn't know what i would create policies on them for (e.g What does it mean to create a policy for broadcast or presence)
Fixed with a dropdown button.

> 
>     3. Might be worth also directing users to the realtime RLS page from auth RLS if they're viewing the realtime schema there
>        
>        * image

Tried this approach in a separate PR https://github.com/supabase/supabase/pull/22471, didn't get much traction. 

About @MildTomato comments for removing duplicate classes, I'll leave that out for addressing in another effort. The layout duplication is way overdue for refactoring.

PR #22574, opened 2024-04-08
feat: add new server-side auth docs for sveltekit

charislam:

YES

Feature

No end-to-end guide for Server-Side Auth with SvelteKit

## What is the new behavior?

End-to-end guide for setting up Server-Side Auth with SvelteKit, covering:

- Protected routes
- Signin, signup, signout
- Protected data

## Additional context

https://docs-git-feat-new-svelte-ssr-docs-supabase.vercel.app/docs/guides/auth/server-side/sveltekit

### DIFF
--- a/apps/docs/content/guides/auth/server-side/sveltekit.mdx
+++ b/apps/docs/content/guides/auth/server-side/sveltekit.mdx
@@ -322,7 +322,7 @@ Create your first page. This example page calls Supabase from the server to get 
 
 This is an example of a public page that uses publicly readable data.
 
-To populate your database, run the [countries quickstart](/dashboard/project/_/sql/3ab7f123-675a-4342-a5d3-d342ed115a4b) from your dashboard.
+To populate your database, run the [countries quickstart](/dashboard/project/_/sql/quickstarts) from your dashboard.
 
 </StepHikeCompact.Details>
 
@@ -489,12 +489,12 @@ export const GET: RequestHandler = async ({ url, locals: { supabase } }) => {
     const { error } = await supabase.auth.verifyOtp({ type, token_hash })
     if (!error) {
       redirectTo.searchParams.delete('next')
-      return redirect(307, redirectTo)
+      return redirect(303, redirectTo)
     }
   }
 
   redirectTo.pathname = '/auth/error'
-  return redirect(307, redirectTo)
+  return redirect(303, redirectTo)
 }
 ```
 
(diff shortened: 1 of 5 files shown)


apps/docs/content/guides/auth/server-side/sveltekit.mdx
```diff
@@ -0,0 +1,636 @@
...
+
+<StepHikeCompact.Step step={7}>
+
+<StepHikeCompact.Details title="Create your first page">
+
+Create your first page. This example page calls Supabase from the server to get a list of countries from the database.
+
+This is an example of a public page that uses publicly readable data.
+
+To populate your database, run the [countries quickstart](/dashboard/project/_/sql/3ab7f123-675a-4342-a5d3-d342ed115a4b) from your dashboard.
```
J0, line 325: /nit might want to check if this path is fixed - id seems to be `13f0cb1b-1649-4897-aa28-eb1193e792c1` for me
CleanShot 2024-04-17 at 01 36 21@2x

apps/docs/content/guides/auth/server-side/sveltekit.mdx
```diff
@@ -0,0 +1,636 @@
...
+  if (token_hash && type) {
+    const { error } = await supabase.auth.verifyOtp({ type, token_hash })
+    if (!error) {
+      redirectTo.searchParams.delete('next')
+      return redirect(307, redirectTo)
+    }
+  }
+
+  redirectTo.pathname = '/auth/error'
+  return redirect(307, redirectTo)
```
J0, line 497: /tangential

Curious about what the thought process was like around 303 vs 307. Did read [this page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307)  but still not too clear about when to opt for 307.

If you could share about the process of deciding between the two would be much appreciated

J0 (approved):
Looks great, should hopefully resolve a sizeable chunk of the Svelte related issues.

apps/docs/content/guides/auth/server-side/sveltekit.mdx
charislam: damn I was hoping this is fixed, lemme see if I can figure out how to get a permanent link to it...

apps/docs/content/guides/auth/server-side/sveltekit.mdx
charislam: good question, I think I may have made it 307 because I was looking at the sveltekit redirect docs when I wrote this, and I think in this case it has the same behavior as a 303. Going to change to a 303 for consistency with other parts of this page that should be 303

PR #22606, opened 2024-04-09
Chore/add vite to connect UI

saltcod:
Split react into CRA and vite


alaister (approved):
LGTM!

One improvement might be to change the names of the env vars to something like:
- `VITE_SUPABASE_URL`
- `VITE_SUPABASE_ANON_KEY`

PR #22641, opened 2024-04-11
Feat/integrate index advisor

joshenlim:
Bringing in what we've got at [index_advisor](https://github.com/olirice/index_advisor/tree/main) into the dashboard 
Available in the query performance section when zooming into a specific query

Empty state:
image

After enabling the extension + an index can be suggested:
image

If no indexes are suggested (we also show what indexes are involved in the query too):
image


apps/studio/components/interfaces/QueryPerformanceV2/QueryPanel.tsx
```diff
@@ -0,0 +1,75 @@
...
+  after?: number
+}) => (
+  <div className={cn('flex flex-col gap-y-1', className)}>
+    <div className="flex items-center gap-x-2">
+      <span>{name}</span>
+      <Tooltip_Shadcn_>
+        <TooltipTrigger_Shadcn_ asChild>
+          <InformationCircleIcon className="transition text-foreground-muted w-3 h-3 data-[state=delayed-open]:text-foreground-light" />
+        </TooltipTrigger_Shadcn_>
+        <TooltipContent_Shadcn_ side="bottom" className="w-52 text-center">
```
alaister, line 40: What do you think of making this top as well? That way it doesn't cover the numbers

Screenshot 2024-04-11 at 15 31 52

```suggestion
        <TooltipContent_Shadcn_ side="top" className="w-52 text-center">
```

alaister (approved):
LGTM!

apps/studio/components/interfaces/QueryPerformanceV2/QueryIndexes.tsx
```diff
@@ -0,0 +1,326 @@
...
+                            You can improve this query's performance by{' '}
+                            <span className="text-brand">{totalImprovement.toFixed(2)}%</span> by
+                            adding the following suggested{' '}
+                            {index_statements.length > 1 ? 'indexes' : 'index'}
+                          </AlertDescription_Shadcn_>
+                        </Alert_Shadcn_>
+                      ) : (
+                        <p className="text-sm text-foreground-light">
+                          Creating the following {index_statements.length > 1 ? 'indexes' : 'index'}{' '}
+                          on <code className="text-xs">public.files</code> can improve this query's
```
bnjmnt4n, line 214: Spotted this hardcoded schema name which is inaccurate. Perhaps the line should be removed?

PR #22654, opened 2024-04-11
feat: add types for available regions in the docs

mansueli:

YES

Small improvement in the docs by documenting available types for invoking a function in a particular region.

## What is the new behavior?

## Additional context

### DIFF
--- a/apps/docs/spec/supabase_js_v2.yml
+++ b/apps/docs/spec/supabase_js_v2.yml
@@ -568,7 +568,7 @@ functions:
     notes: |
       - Requires either an email or phone number.
       - This method is used for passwordless sign-ins where a OTP is sent to the user's email or phone number.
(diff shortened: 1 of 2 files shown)


apps/docs/spec/supabase_js_v2.yml
```diff
@@ -4991,6 +4991,29 @@ functions:
             method: 'DELETE'
           })
           ```
+      - id: regional-invocation
+        name: Invoking a Function in the UsEast1 region
+        description: |
+          Here are the available regions:
+          - `FunctionRegion.Any`, `FunctionRegion.ApNortheast1`,
```
charislam, line 4998: I'm a bit confused by the pairing up of regions here, was expecting pairs to be somehow connected to each other but as far as I can tell, they're not. Could we just put every region as its own item instead?

charislam (approved):

PR #22655, opened 2024-04-11
fix: Escape new lines for DocsSearch

oakhtar147:

YES

Bug fix. In the docs search any `CommandItem` with a new line was breaking the search component.

Notice that `data-value` has a new line in the value, which was causing `querySelector` to break internally in `cmdk`.

https://github.com/supabase/supabase/assets/57761675/1026912c-c847-473e-b25f-5d80229d2eb7

Here's a [smaller repro](https://codesandbox.io/p/sandbox/stoic-framework-fhpdhm?file=%2Fsrc%2FApp.tsx%3A21%2C1) of the same issue. The version of `cmdk` is the same.

## What is the new behavior?

https://github.com/supabase/supabase/assets/57761675/a2e57ed0-70ce-4093-922b-3457df088655

## Additional context

I think the ideal solution would be to update `cmdk` to a later version. `cmdk@1.0.0` does not seem to have this issue.


charislam:
Thanks @oakhtar147! You're right about upgrading `cmdk`, but that requires more testing than we have time to tackle at present since it's a major version upgrade, so your PR is great.

It might be a bit more future-proof if we just do `encodeURIComponent`, actually, instead of continually discovering new symbols to escape. Do you mind switching out the escape logic to do that instead?

oakhtar147:
@charislam Ah yes, that's even better! I've removed `escapeDoubleQuotes` and replaced with `encodeURIComponent` (f3fcbc4f9).

charislam (approved):

oakhtar147:
@charislam, thanks for approving! There seems to be a Vercel deployment check failing that is preventing auto-merge.

PR #22722, opened 2024-04-15
chore: import edge runtime types.

thorwebdev:


examples/ai/edge-functions/supabase/functions/generate-embedding/index.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="https://esm.sh/v135/@supabase/functions-js@2.3.0/src/edge-runtime.d.ts" />
```
laktek, line 1: can we change the URL here to `https://esm.sh/@supabase/functions-js/src/edge-runtime.d.ts` it will be redirected to above. That way we don't need to update all references on a new release.

thorwebdev: I generally like to specify the exact versions as that means the code won't break with future changes. But I've updated it to remove.

soedirgo (approved):

PR #22725, opened 2024-04-15
fix: Fix the role impersonation in the realtime inspector

ivasilov:


apps/studio/components/interfaces/Realtime/Inspector/useRealtimeMessages.ts
```diff
@@ -170,6 +176,14 @@ export const useRealtimeMessages = ({
         }
       } else if (status === 'CLOSED') {
         // console.log(`Realtime Channel status: ${status}`)
+      } else if (status === 'CHANNEL_ERROR') {
+        toast.error(
+          'Failed to connect to the channel. This may be due to restrictive RLS policies. Check your role and try again.'
```
joshenlim, line 181: tinnyyyyyyy nit 

```suggestion
          'Failed to connect to the channel: This may be due to restrictive RLS policies. Check your role and try again.'
```

apps/studio/components/interfaces/Realtime/Inspector/RealtimeTokensPopover/index.tsx
```diff
@@ -59,27 +58,30 @@ export const RealtimeTokensPopover = ({ config, onChangeConfig }: RealtimeTokens
...
-    } else {
-      token = serviceRoleKey
+      if (
+        config.projectRef !== undefined &&
+        jwtSecret !== undefined &&
+        snap.role !== undefined &&
+        snap.role.type === 'postgrest'
+      ) {
+        token = anonRoleKey
+        await getRoleImpersonationJWT(config.projectRef, jwtSecret, snap.role)
```
alaister, line 72: Isn't `await` redundant here? Doesn't actually matter, just wondering 😅

alaister (approved):
LGTM!

apps/studio/components/interfaces/Realtime/Inspector/RealtimeTokensPopover/index.tsx
ivasilov: That's why the role impersonation didn't work, the bearer variable wasn't assigned when calling `onChangeConfig` so it was always null.

PR #22777, opened 2024-04-16
chore: only query repos with github auth

alaister:


apps/studio/components/layouts/AppLayout/EnableBranchingButton/EnableBranchingModal.tsx
```diff
@@ -123,9 +114,9 @@ const EnableBranchingModal = () => {
     defaultValues: { branchName: '' },
   })
 
-  const isLoading = isLoadingRepositories
-  const isError = isErrorRepositories || isErrorUpgradeEligibility
-  const isSuccess = isSuccessRepositories && isSuccessUpgradeEligibility
+  const isLoading = isLoadingConnections || isLoadingUpgradeEligibility
+  const isError = isErrorConnections || isErrorUpgradeEligibility
```
joshenlim, line 118: i just realised we're not using the error, isLoading, isError, isSuccess from `connections` - do we need to?

joshenlim (approved):

apps/studio/components/layouts/AppLayout/EnableBranchingButton/EnableBranchingModal.tsx
alaister: Yeah I think it was a mistake that they weren't included before. Most of the time the user doesn't see the loading state anyway since it's loaded in the background, but better safe than sorry if they manage to open the modal quick!

PR #22786, opened 2024-04-16
feat: Use URL query params in Auth Policies page

ivasilov:
This PR refactors the Auth Policies page to use URL query to keep state about selected schema and search string. It removes dependence on the Valtio state which is misused as a god object.


apps/studio/pages/project/[ref]/auth/policies.tsx
```diff
@@ -66,30 +63,24 @@ const onFilterTables = (
 }
 
 const AuthPoliciesPage: NextPageWithLayout = () => {
-  const { search, schema } = useParams()
+  const [params, setParams] = useUrlState<{
+    schema?: string
+    search?: string
+  }>()
+  const { schema = 'public', search: searchString = '' } = params
```
Comment (ivasilov) [inline apps/studio/pages/project/[ref]/auth/policies.tsx:70]: The idea I have is to refactor most of the UI to use this pattern (schema through URL query) and then instead of hardcoding `public` here, we put whatever is in local storage. That way, if a user comes to the page but doesn't have a schema in the URL, it will preselect the default one (instead of `public`).

alaister (approved):
LGTM!

PR #22796, opened 2024-04-16
fix: Make enums from other schemas work with public tables

ivasilov:


alaister (commented):
Looks good!
Would it be possible to prefix the enum name with its schema? If I have an enum called `wow`, have `public.wow` for example. At the moment if I have the same name enum in two schemas it's confusing:
Screenshot 2024-04-17 at 13 26 48

alaister (approved):
LGTM!

Hallidayo:

PR #22812, opened 2024-04-17
feat: new auth ia

charislam:


kangmingtay (approved):

apps/docs/components/MDX/oauth_pkce_flow.mdx
```diff
@@ -1,16 +1,49 @@
...
+<Tabs
+  scrollable
+  size="small"
+  type="underlined"
+  defaultActiveId="client"
+  queryGroup="environment"
+>
+<TabPanel id="client" label="Client">
+
+In the browser, `signInWithOAuth` automatically redirects to the SSO provider's authentication endpoint, which then redirects to your endpoint.
```
J0, line 12:
```suggestion
In the browser, `signInWithOAuth` automatically redirects to the OAuth provider's authentication endpoint, which then redirects to your endpoint.
```

/nit can we consider using OAuth? Aware that SSO is correct as well but had a ticket come in the past asking if they need to be on pro plan to use `signInWithOAuth`

J0 (approved):

J0 (approved):

packages/shared-data/config.ts
```diff
@@ -0,0 +1,104 @@
+const config = {
+  auth: {
```
saltcod, line 2: is this all auth config? should it be called that?

saltcod (approved):

packages/shared-data/config.ts
charislam: auth defaults is probably more accurate, lemme change 👍🏼

PR #22831, opened 2024-04-17
fix: typo in auth-sql-rules-examples.js

georgemoon:

YES

Bug fix

Typo in SQL code

## What is the new behavior?

Fixed typo

## Additional context

Current bug visible on: https://supabase.com/auth

### DIFF
--- a/apps/www/data/products/auth/auth-sql-rules-examples.js
+++ b/apps/www/data/products/auth/auth-sql-rules-examples.js
@@ -65,7 +65,7 @@ create table teams (
 
 create table members (
   team_id references team.id,
-  user_id referenced auth.users.id
+  user_id references auth.users.id
 );
 
 alter table teams enable row level security;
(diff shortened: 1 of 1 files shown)


georgemoon:
In fact this code block doesn't seem to execute in general, but "referenced" is not a Postgres command in any case :)

georgemoon:
Maybe it should be aligned to the code found here:
https://supabase.com/docs/guides/auth/row-level-security#using-joins

```
-- 1. Create a table of teams
create table teams (
  id serial primary key,
  name text
);

-- 2. Create many to many join
create table members (
  team_id bigint references teams,
  user_id uuid references auth.users
);

-- 3. Enable RLS
alter table teams enable row level security;

-- 4. Create Policy
create policy "Team members can update team details if they belong to the team"
  on teams
  for update using (
    (select auth.uid()) in (
      select user_id from members
      where team_id = id
    )
  );
```

charislam (approved):

charislam:
@georgemoon thanks! you're absolutely right, updated the code sample so it runs

Dustin4444 (commented):
#41695

PR #22865, opened 2024-04-18
feat(studio): db-pool postgrest config

soedirgo:
Allows self-serve configuration of Data API db pool size

Screenshot 2024-04-18 at 7 08 07 PM


alaister (approved):
LGTM!

PR #22888, opened 2024-04-19
fix: update magic link defaults

J0:

YES

We are reducing the default otp expriy duration which would affect magic links.


J0:
Closing since we're not updating the default

J0:
Looks like we did update the default

kangmingtay (approved):

PR #22891, opened 2024-04-19
feat: long otp expiry lint

alaister:

feature

## What is the new behaviour?

Screenshot 2024-04-19 at 13 08 18

### DIFF
--- a/apps/studio/components/interfaces/Linter/LinterDataGrid.tsx
+++ b/apps/studio/components/interfaces/Linter/LinterDataGrid.tsx
@@ -1,7 +1,13 @@
 import { ExternalLink, Eye, Table2, X } from 'lucide-react'
+import Link from 'next/link'
+import { useRef, useState } from 'react'
+import DataGrid, { Column, DataGridHandle, Row } from 'react-data-grid'
+import ReactMarkdown from 'react-markdown'
+
+import { useParams } from 'common'
+import { LINTER_LEVELS } from 'components/interfaces/Linter/Linter.constants'
 import { NoIssuesFound, lintInfoMap } from 'components/interfaces/Linter/Linter.utils'
-import DataGrid from 'react-data-grid'
-import { Row, DataGridHandle, Column } from 'react-data-grid'
+import { Lint } from 'data/lint/lint-query'
 import {
   Button,
   ResizableHandle,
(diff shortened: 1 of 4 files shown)


apps/studio/components/interfaces/Linter/LinterDataGrid.tsx
```diff
@@ -69,7 +69,11 @@ const LinterDataGrid = ({
       value: (row: any) => (
         <div className="flex items-center gap-1 text-xs">
           <span className="shrink-0">{entityTypeIcon(row.metadata?.type)}</span>
-          {`${row.metadata.schema}.${row.metadata.name}`}
+          {row.metadata &&
```
joshenlim, line 72: Do we need that many checks? What about just checking if schema and name are undefined?

Otherwise if possible lets shift this condition into a const so its a bit more readable

joshenlim: Same for below L223

joshenlim (approved):

apps/studio/pages/project/[ref]/database/security-advisor.tsx
```diff
@@ -31,12 +32,49 @@ const ProjectLints: NextPageWithLayout = () => {
...
+  // [Alaister]: checking this client side for speed, but should be moved into the query if possible
+  if (authConfig?.EXTERNAL_EMAIL_ENABLED && authConfig.MAILER_OTP_EXP > 3600) {
+    clientLints.push({
+      name: 'auth_otp_long_expiry',
+      level: 'WARN',
+      facing: 'EXTERNAL',
+      categories: ['SECURITY'],
+      description: 'OTP expiry exceeds recommended threshold',
+      detail:
+        'We have detected that you have enabled the email provider with the OTP expiry set to more than an hour. It is recommended to set this value to less than an hour.',
```
Comment (inian) [inline apps/studio/pages/project/[ref]/database/security-advisor.tsx:65]:
```suggestion
        'We have detected that you have enabled the email provider with the OTP expiry set to more than an hour. It is recommended to set this value to less than an hour to prevent brute force attacks.',
```

apps/studio/components/interfaces/Linter/LinterDataGrid.tsx
alaister: Cleaned up 👍

hf (approved):

PR #22894, opened 2024-04-19
Update docs to maybe optimize RBAC functions performance

felixgabler:

YES

Docs update

The RBAC docs do not use subqueries in the RLS policies and also do not mark the `authorize` function as `stable`.

## What is the new behavior?

I added `stable` to `authorize`, made it fetch the user role into a variable to save calls and also added subqueries to the RLS policies. I am not sure the subqueries really help in this case but generally it is advisable to get people to think about using subqueries for their auth functions in RLS so subquery caching can take effect. Curious for your thoughts!

## Additional context


charislam:
LGTM but I'm not super familiar with this flow, so I'll get @thorwebdev's second opinion as the author

felixgabler:
@thorwebdev what do you think about this PR?

pashkinelfe (approved):

thorwebdev:
@felixgabler very sorry, this slipped through the cracks. Would you mind resolving the conflicts and then we can get it merged in? Thanks for bearing with us here 💚

felixgabler:
@thorwebdev no worries, thanks for taking a look now! I have resolved the comments and it should be ready to merge now.

I hope it is okay if I use this opportunity to mention another issue that might have slipped through the cracks: https://github.com/supabase/auth-js/issues/912. This is blocking the Nuxt supabase integration from upgrading to `@supabase/ssr` through [this PR](https://github.com/nuxt-modules/supabase/pull/357). Maybe you could nudge someone for us? Thanks!

PR #23101, opened 2024-04-21
refactor: migrate guides to app router

charislam:
Migrate guides to App Router.

- [x] Checked I haven't broken metadata and Open Graph
- [x] Checked bundle size increase is reasonable
- [x] Checked mobile sizing and light/dark mode
- [x] Checked lazy loading of components still works
- [x] Checked command menu (search, docs AI)
- [x] Checked for no new console errors
- [x] Checked navigation between guides and reference pages 
- [x] Check that redirects still work
- [x] Check that 404s don't look too stupid

image


MildTomato:
Just had a brief look through changes.

What about using one Layout (which is a "use client") which uses next/navigation to determine which menu id to use.

Could also use one page file [using a catch all](https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes#catch-all-segments).

apps/docs/app/[...slug]/_page.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * looking for, which isn't possible with a regular 404 page (receives no
+ * params).
+ */
+
+import { redirect } from 'next/navigation'
+import { notFoundLink } from '~/features/recommendations/NotFound.utils'
+
+type Params = { slug: string[] }
+
+const CatchAllPage = async ({ params }: { params: Params }) => {
```
Comment (fsansalvadore) [inline apps/docs/app/[...slug]/_page.tsx:19]: I tried this on the preview but it responds with a 500 error page.

Screenshot 2024-07-01 at 11 07 03

Comment (fsansalvadore) [reply apps/docs/app/[...slug]/_page.tsx:19]: And maybe worth keeping this part as a separate PR to reduce scope? @charislam

fsansalvadore (changes requested):

fsansalvadore (approved):

PR #23148, opened 2024-04-22
chore: Update C# docs to reflect new Nuget Package names and Changelog

acupofjose:

YES

Updates docs


charislam (approved):

charislam:
Added some changes so we can maintain both v0 and v1 versions of the docs. Are there any changes to any of the library functions that we need to document as well? I scanned the release notes and it looks like there are some new properties and methods. If none of those changes are breaking, we can always document them in a follow-up PR.

acupofjose:
In terms of breaking functionality, no. Just the changes to the names of the libraries. There are some additional methods that need to be documented for parity with the JS clients, but that can be a different PR as you said! Thanks!

PR #23149, opened 2024-04-22
chore(docs): update swift docs with latest methods

grdsdev:

Update Swift docs with changes from upcoming release https://github.com/supabase/supabase-swift/pull/344


apps/docs/spec/supabase_swift_v2.yml
```diff
@@ -3040,6 +3141,37 @@ functions:
               )
             )
           ```
+      - id: regional-invocation
+        name: Invoking a Function in the UsEast1 region
```
charislam, line 3145:
```suggestion
        name: Invoking a Function in a specific region
```

charislam (approved):

PR #23154, opened 2024-04-22
fix: storage RLS example

bcihanc:
It's about the type fix.

ERROR: operator does not exist: text = uuid
Hint: No operator matches the given name and argument types. You might need to add explicit type casts.


charislam (approved):
Thanks @bcihanc!

PR #23159, opened 2024-04-22
enhance(educate): add section on how rls works with views

charislam:

YES

Docs update

Not clear how RLS works with views

## What is the new behavior?

Clarify how to protect views when using RLS

## Additional context

### DIFF
--- a/apps/docs/content/guides/auth/row-level-security.mdx
+++ b/apps/docs/content/guides/auth/row-level-security.mdx
@@ -125,6 +125,20 @@ on my_table for select
 using ( true );
 ```
 
+### Protect views using `security_invoker`
+
+Views bypass RLS by default because they are usually created with the `postgres` user. This is a feature of Postgres, which automatically creates views with `security definer`.
+
+In Postgres 15 and above, you can make a view obey the RLS policies of the underlying tables when invoked by `anon` and `authenticated` roles by setting `security_invoker = true`.
+
+```sql
+create view <VIEW_NAME>
+with(security_invoker = true)
+as select <QUERY>
+```
+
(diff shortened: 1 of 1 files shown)


apps/docs/content/guides/auth/row-level-security.mdx
```diff
@@ -125,6 +125,20 @@ on my_table for select
 using ( true );
 ```
 
+### Protect views using `security_invoker`
+
+Views do _not_ obey RLS policies by default. This is a feature of Postgres, which automatically creates views with `security definer`.
```
olirice, line 130: It is not 100% accurate to say that `security invoker` views don't follow RLS.

`security definer` means that the view follows the permissions of the role that created the view, rather than the role that is querying the view.

In almost all cases, the view is created by the `postgres` user, which bypasses RLS. For that reason, these views to not respect RLS rules, BUT if a role without RLS created the view, it would still follow those rules.

apps/docs/content/guides/auth/row-level-security.mdx
```diff
@@ -125,6 +125,20 @@ on my_table for select
...
+
+In Postgres 15 and above, you can make a view obey the RLS policies of the underlying tables by setting `security_invoker = true`.
+
+```sql
+create view
+security_invoker = true
```
olirice, line 136: this could read a bit nicer with placeholders for the other parts of the `create view` statements had placeholders. like `<view name>` etc

apps/docs/content/guides/auth/row-level-security.mdx
charislam: Very good point, thanks for the clarification!

mansueli (approved):

PR #23186, opened 2024-04-23
Reinstate link button for foreign keys in table editor side panel

joshenlim:
Original:
image

Empty state:
image

Relations in the midst of editing:
image

For relations that are already existing:
image

image

Composite foreign keys:
image

image


alaister (approved):
LGTM!

PR #23204, opened 2024-04-23
fix(ux): allow creating fk on new column after changing name

### LINKED ISSUE #22702: client doesnt update foregin key code when updating new column name
# Bug report

## Describe the bug

if you go to an existing table from the table editor, start creating a new column, define a foreign key relationship, confirm the foreign key relationship and then ** update the column name **, the foreign key relationship doesn't update the column name. all of this happens before confirming the creation of the table.

## To Reproduce

1. go to an existing project
2. go to table editor
3. create a new column
4. call the column "test"
5. make it be a foreign key to another column
6. click "save" (the foreign key, not the column!)
7. change the column name to "another_name"
8. click save (the column)
9. it will throw an error, because the column "test" doesn't exist.
10. this happens because you changed the name of the column, but the foreign key relationship didn't update to the new column name, so it will reference the old name you wrote for the column

## Expected behavior

(all this is before actually saving the new column) if you write a foreign key relationship and then change the name for the new column, the foreign key relationship query should change too, changing the column name referenced to the new one.  

## Screenshots

image
image
image

charislam:
Old behavior: If you create a new column, create an FK on that column, then change the name of the column before saving, FK creation will fail. This is because the name of the FK source column does not update to match the changed name.

New behavior: When you change the name of the new column, the FK source column changes to match, so the FK is created correctly.

Before:
image

After:
image

Resolves #22702


joshenlim (approved):

alaister (approved):
LGTM!

Huilensolis:
Thank god, it was so difficult to summarize the issue when i opened it,  but you made it very clear!

PR #23222, opened 2024-04-24
chore: Refactor the StorageExplorerStore to use API calls from the `data` folder

ivasilov:


apps/studio/data/storage/bucket-object-delete-mutation.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  if (error) handleError(error)
+  return data
+}
+
+export function useBucketObjectDeleteMutation() {
+  return useMutation({
+    mutationFn: deleteBucketObject,
+  })
+}
```
alaister, line 37: For all these new mutation functions is there a reason why we're not passing through the options like in the other mutations? Makes it easy to pass `onSuccess()` from outside for example

joshenlim: pushed up a change to make them consistent :) although they arent being used at the moment

alaister (approved):
LGTM!

PR #23235, opened 2024-04-24
enhance(sql editor): set result column width roughly based on contents

charislam:

YES

Enhancement

Column width of the SQL editor results always starts at 120 regardless of the content

## What is the new behavior?

Column width adjusts (very roughly) to content. Doesn't adjust for arrays and JSON (assuming those tend to be too long anyway), and for speed only adjusts based on the first returned row, not on the max length of all rows. Works quite well for things like UUIDs and timestamps though, where width is uniform across rows.

image
image

## Additional context

### DIFF
--- a/apps/studio/components/interfaces/SQLEditor/UtilityPanel/Results.tsx
+++ b/apps/studio/components/interfaces/SQLEditor/UtilityPanel/Results.tsx
@@ -56,26 +56,34 @@ const Results = ({ id, rows }: { id: string; rows: readonly any[] }) => {
   }
 
   const EST_CHAR_WIDTH = 8.25
-  const columns: CalculatedColumn<any>[] = Object.keys(rows?.[0] ?? []).map((key, idx) => ({
-    idx,
-    key,
-    name: key,
-    resizable: true,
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/SQLEditor/UtilityPanel/Results.tsx
```diff
@@ -55,14 +55,18 @@ const Results = ({ id, rows }: { id: string; rows: readonly any[] }) => {
...
     idx,
     key,
     name: key,
     resizable: true,
     parent: undefined,
     level: 0,
-    width: 120,
+    width:
+      typeof rows[0][key] === 'string' || typeof rows[0][key] === 'number'
+        ? Math.min(String(rows[0][key]).length * EST_CHAR_WIDTH, 500)
```
joshenlim, line 68: a better way is probably to consider _all_ rows - take the longest value from all the rows and use it as the multiplier with est_char_width

otherwise e.g if the first row for a column only has 1 char, that column will be extra tiny

joshenlim (approved):

PR #23255, opened 2024-04-25
Update CTA for manual disk size increase beyond limit

joshenlim:


apps/studio/components/interfaces/Settings/Database/DiskSizeConfiguration.tsx
```diff
@@ -129,51 +141,33 @@ const DiskSizeConfiguration = ({ disabled = false }: DiskSizeConfigurationProps)
...
-              >
-                {projectSubscriptionData?.plan?.id === 'free'
-                  ? 'Upgrade subscription'
-                  : 'Disable spend cap'}
-              </Link>
-            </Button>
-          }
-        >
-          <div>
+              : 'Disk size configuration is only available when disabling the spend cap.'}
```
alaister, line 168:
```suggestion
              : 'Disk size configuration is only available when the spend cap has been disabled.'}
```

alaister (approved):
LGTM! Left one small Terry-style English suggestion 😅

Also probably unrelated to this PR, but 200 is not more than 200:
Screenshot 2024-04-25 at 12 30 23

apps/studio/components/interfaces/Settings/Database/DiskSizeConfiguration.tsx
joshenlim: yeap agreed much better! active voice :)

PR #23275, opened 2024-04-25
Add ui-patterns alias

saltcod:
now components will autoimport in vscode from `ui-patterns` like they do from `ui`


charislam (approved):
should studio have this too?

ivasilov:
The fix doesn't change anything in my VSCode. Still can't get the autosuggest. Tried restarting the app and restarting the ts server.

PR #23280, opened 2024-04-25
refactor: clean up useClient

charislam:
Move `useClient`s to the actual file instead of index files, so that (a) we don't ship unnecessary client-side code and (b) we don't get the "can't dot into client component from server" error when trying to use MDX in RSCs.


packages/ui/src/components/Accordion/index.tsx
```diff
@@ -1 +1,3 @@
+'use client'
```
MildTomato, line 1: can we move these directly into the file?

MildTomato (dismissed):

MildTomato:
This doesn't seem right to me.

Screenshot 2024-04-26 at 12 48 50 PM

A lot of these components also cannot work server OR client, alot of them should be 'use client' as they have internal state - so this just confuses it further

charislam:
> This doesn't seem right to me.
> 
> Screenshot 2024-04-26 at 12 48 50 PM
> why do we need separate import files if we have already removed 'use client' from the index.tsx ?

just tested against the main migration pr and you're right, it works without, gonna change

packages/ui/src/lib/Markdown/index.tsx
```diff
@@ -1,4 +1,5 @@
-import * as React from 'react'
+'use client'
```
charislam, line 1: this one is on an index file because it exports an object and I think you will otherwise run into the "cannot dot into a client component" problem

saltcod (approved):

Issue #23225, opened 2024-04-24
CSS issue with side navigation panel of project dashboard : logo stacking beneath other items


JaykumarPatel1998:
# Bug report

- [✔ ] I confirm this is a bug with Supabase, not with my own application. 
- [ ✔] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Slight CSS issues in the side navigation panel of the project dashboard page: When the side panel is expanded and scrolled - supabase logo appears at the bottom layer and other items are layered above it.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to 'https://supabase.com/dashboard/projects'.
2. Open any project from your projects.
3. hover above the side navigation panel and scroll through the items.
4. Focus on the supabase logo at the top-left corner of the screen - it appears awkwardly positioned - IMO, it should appear at top and all items should scroll under it.

## Expected behavior

 Focus on the supabase logo at the top-left corner of the screen - it appears awkwardly positioned - IMO, it should appear at top and all items should scroll under it.

## Screenshots
before scroll:
![image](https://github.com/supabase/supabase/assets/85047980/208fb7ab-ead0-43cd-bafc-a99fc36c0cd1)

after scroll:
![image](https://github.com/supabase/supabase/assets/85047980/e14cc19d-aef4-4c43-8c81-d71b20645775)

after side panel is collapsed:
![image](https://github.com/supabase/supabase/assets/85047980/b8eaa87c-eb7a-4838-a521-1b7de0bdabf6)

smitgol:
I'll take a look at this one

Neeraj-gagat:
hi @JaykumarPatel1998  i am getting error while installing dependencies 
on npm install i come across this error

npm warn deprecated puppeteer@16.2.0: < 19.4.0 is no longer supported
npm error code 1
npm error path /home/neeraj/Desktop/repos-to-contribute/supabase/node_modules/libpg-query
npm error command failed
npm error command sh -c node-pre-gyp install --fallback-to-build
[...]

PR #23293, opened 2024-04-26
fix: supabase side menu logo was overlaying side menu items on small screens

liambirkin:

YES

Bug fix

Supabase side menu logo was overlaying side menu items on smalll screens.

https://github.com/supabase/supabase/issues/23225

## What is the new behavior?

Added background to nav items so that nav items cover the supabase logo on scroll (same behaviour/appearance as hover and active state)

## Additional context

Before:
CleanShot 2024-04-26 at 20 50 59@2x

After:
CleanShot 2024-04-26 at 20 52 39@2x
CleanShot 2024-04-26 at 20 52 55@2x


Julius-Java:
But I think the expected behavior is for the supabase logo to still be visible on scroll but have the other nav items scroll under the logo, that would be a better solution than covering the entire logo on scroll.

liambirkin:
Have added an alternative fix that behaves like that^

saltcod:
fixed in https://github.com/supabase/supabase/pull/25966

PR #23307, opened 2024-04-27
feat: llama3 in studio

gregnr:
Initial work to Bring-your-own-AI™ to Supabase studio. Requires an OpenAI-compatible API. Tested with Llama3 via Groq.

### Changes
In addition to the existing `OPENAI_API_KEY` env var, there are now 2 more:
- `OPENAI_BASE_URL`: point this to your openai-compatible API, eg https://api.groq.com/openai/v1
- `OPENAI_MODEL`: choose the language model, currently supports llama3, gpt-3.5, and gpt-4

In order for clippy to continue to work, we also added these env vars:
- `OPENAI_EMBEDDING_API_KEY`
- `OPENAI_EMBEDDING_BASE_URL`
- `OPENAI_EMBEDDING_MODEL`

These are necessary if we want to be able to eg. use groq + llama3 for the language model and openai + text-embedding-ada-002 for the embedding model.

### Other
Also did some work on the tokenizer to support open models like llama3 (we use the tokenizer to count tokens before sending them to the model).

### DIFF
--- a/apps/database-new/.env.local.example
+++ b/apps/database-new/.env.local.example
@@ -3,3 +3,5 @@
 NEXT_PUBLIC_SUPABASE_URL=your-project-url
 NEXT_PUBLIC_SUPABASE_ANON_KEY=your-anon-key
 OPENAI_API_KEY=your-openai-api-key
+OPENAI_BASE_URL=your-openai-compatible-base-url
+OPENAI_MODEL=your-language-model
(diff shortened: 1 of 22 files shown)


nfairbank:
Is there any appetite to push this through? Would be an amazing feature to allow this to plug into self-hosted models.

PR #23394, opened 2024-04-29
Add support for logs filtering under postgres logs

joshenlim:
API Edge Network and Supervisor logs already have RR filters, added for Postgres

However, still got some questions for infra, marking as draft meanwhile


alaister (approved):
LGTM!

PR #23409, opened 2024-04-29
Chore/add replica filter for query performance

joshenlim:
image


apps/studio/components/interfaces/QueryPerformance/QueryPerformance.tsx
```diff
@@ -108,6 +111,10 @@ export const QueryPerformance = ({
     slowestExecutionTime,
   ])
 
+  useEffect(() => {
+    state.setSelectedDatabaseId(ref)
+  }, [])
```
alaister, line 116: Are you intentionally omitting the `ref` from the deps? If so would you mind adding a comment on why 🙏🏻
```suggestion
  }, [ref])
```

apps/studio/hooks/analytics/useDbQuery.tsx
```diff
@@ -38,12 +53,12 @@ const useDbQuery = (
     isRefetching,
     refetch,
   } = useQuery(
-    ['projects', project?.ref, 'db', { ...params, sql: resolvedSql }, where, orderBy],
+    ['projects', project?.ref, 'db', { ...params, sql: resolvedSql, identifier }, where, orderBy],
     ({ signal }) => {
       return executeSql(
         {
           projectRef: project?.ref,
```
alaister, line 60: Just want to make sure this should still be `project?.ref` and not `identifier` still

apps/studio/components/interfaces/QueryPerformance/QueryPerformance.tsx
joshenlim: good point actually - yeah can add ref in the deps

apps/studio/hooks/analytics/useDbQuery.tsx
joshenlim: yup so on line 45 if read replicas is enabled, it'll use the selectedDatabaseId, other default back to project.ref

in this case, each replica will have its own query performance results, hence why added the identifier param in the query key

alaister (commented):
On a project with no read replicas I can still see the button and it seems to be constantly loading:
Screenshot 2024-05-02 at 15 15 22

joshenlim:
ahh good catch sorry 🙏 pushed a fix! it shouldnt show up for projects that dont have replicas enabled for

alaister (approved):
LGTM!

Issue #23454, opened 2024-05-01
docs: self-hosting - add alternative command to make checkout faster 


cyluss:
# Improve documentation

## Link 

https://supabase.com/docs/guides/self-hosting/docker

## Describe the problem

'Self-hosting with docker' document guides to checkout full project repository, which weighs around 766 MiB at 2024-05-01

## Describe the improvement

1. Two versions of self hosting commands are provided, 'General' and 'Advanced'.
2. Guide command with shallow clone is unmodified and presented as General for simplicity.
3. Guide command in git sparse checkout is added and presented as Advanced for faster and efficient checkout.
4. Newly added command would checkout only `docker` directory from project repository using git sparse checkout.

suggestion by @encima in #23455

General: receives 765.93 MiB of objects

```
# Get the code
git clone --depth 1 https://github.com/supabase/supabase

# Go to the docker folder
cd supabase/docker
```

Advanced: receives 24.98 MiB of objects, 3.2% of full checkout or 96.8% reduction
[git sparse checkout requires git 2.25 or later](https://github.com/git/git/blob/master/Documentation/RelNotes/2.25.0.txt#L67)
```
# Get the code using git sparse checkout
git clone --filter=blob:none --no-checkout https://github.com/supabase/supabase
cd supabase
git sparse-checkout set --cone docker && git checkout master

# Go to the docker folder
cd docker
```

## Additional context

General: 765.93 MiB
```
[root@localhost test]# time git clone --depth 1 https://github.com/supabase/supabase
Cloning into 'supabase'...
remote: Enumerating objects: 10213, done.
remote: Counting objects: 100% (10213/10213), done.
remote: Compressing objects: 100% (8907/8907), done.
remote: Total 10213 (delta 1203), reused 6876 (delta 794), pack-reused 0
Receiving objects: 100% (10213/10213), 765.93 MiB | 22.45 MiB/s, done.
Resolving deltas: 100% (1203/1203), done.
Updating files: 100% (9128/9128), done.

real    0m41.416s
user    0m13.667s
sys     0m4.124s
[root@localhost test]#
```

Advanced: 24.98 MiB (24.72 MiB + 261.82 KiB)

```
[root@localhost test2]# time git clone --filter=blob:none --no-checkout https://github.com/supabase/supabase
Cloning into 'supabase'...
remote: Enumerating objects: 199700, done.
remote: Counting objects: 100% (674/674), done.
remote: Compressing objects: 100% (341/341), done.
[...]

PR #23455, opened 2024-05-01
docs: self-hosting - checkout only docker directory from supabase project

cyluss:

YES

docs update

'Self-hosting with docker' document guides to checkout full project repository, which weighs around 766 MiB at 2024-05-01. Please refer supabase/supabase#23454.

## What is the new behavior?

'Self-hosting with docker' document guides to checkout only docker directory from project repository using git sparse checkout. Please refer supabase/supabase#23454.

## Additional context

supabase/supabase#23454

### DIFF
--- a/apps/docs/content/guides/self-hosting/docker.mdx
+++ b/apps/docs/content/guides/self-hosting/docker.mdx
@@ -20,12 +20,11 @@ Follow these steps to start Supabase locally:
 ```sh
 # Get the code
 git clone --filter=blob:none --no-checkout https://github.com/supabase/supabase
-pushd supabase && git sparse-checkout set --cone docker
-git checkout master
-popd
+cd supabase
+git sparse-checkout set --cone docker && git checkout master
 
 # Go to the docker folder
-cd supabase/docker
+cd docker
 
 # Copy the fake env vars
 cp .env.example .env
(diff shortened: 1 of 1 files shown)


encima (changes requested):
looks good and works fine. `pushd` is an optional extra but no preference of leaving it in or removing.

Imo, this is an optional addition to running it locally. Yes, the repo is large but the instructions are designed for all technical levels and the original instructions are more recognisable.

I would suggest having this as a tabbed option for: `General` and `Advanced` maybe.

encima:
@supabase/docs in case you have opinions

cyluss:
I tried to rebase my repository up to date but failed. Can we discuss on the issue #23454 or change issue into discussion?

cyluss:
Yes, this is optional and complicates the document. And it does not run in old versions of git which does not support clone filtering and sparse checkout.

However the repository will get bigger as time goes by, and some readers would copy and paste line-by-line what says on the document. IMO this is inherent complexity of monorepo and it is questionable to assume that all readers are on sufficiently fast broadband and cloning will end in short time period.

cyluss:
I agree with @encima's point and applied suggestion to the original issue. I will send another pull request in the weekend

PR #23648, opened 2024-05-02
Update supabase python reference docs with params

silentworks:

YES

Docs update

There are no parameters and their types in the docs for the functions

## What is the new behavior?

Parameters and their types are added to the docs for the functions

## Additional context


apps/docs/spec/supabase_py_v2.yml
```diff
@@ -781,26 +938,140 @@ functions:
...
           [JSON](/docs/guides/database/json) columns. Postgres offers some
           [operators](/docs/guides/database/json#query-the-jsonb-data) for
           querying JSON data.
         hideCodeBlock: true
+      - id: querying-referenced-table-with-inner-join
+        name: Querying referenced table with inner join
+        code: |
+          ```python
+          response = (
+              supabase.table("cities")
```
J0, line 964: New addition

apps/docs/spec/supabase_py_v2.yml
```diff
@@ -1456,155 +3072,313 @@ functions:
...
+              .text_search(
+                  "catchphrase",
+                  "'fat' & 'cat'",
+                  options={"type": "plain", "config": "english"},
+              )
+              .execute()
+          )
           ```
-          APIResponse(data=[{'id': 1, 'name': 'Michael', 'address': {'postcode': 90210}}], count=None)
+      - id: full-normalization
```
J0, line 3176: 🔥

J0 (commented):
Skimmed, looks great to me

apps/docs/spec/supabase_py_v2.yml
```diff
@@ -64,91 +155,267 @@ functions:
...
+                  {
+                    "id": "11111111-1111-1111-1111-111111111111",
+                    "user_id": "11111111-1111-1111-1111-111111111111",
+                    "identity_data": {
+                      "email": "hello@example.com",
+                      "sub": "11111111-1111-1111-1111-111111111111"
+                    },
+                    "provider": "email",
+                    "created_at": "2024-06-17T00:19:25.774522Z",
+                    "last_sign_in_at": "2024-06-17T00:19:25.774498Z",
```
J0, line 255: /nit Worried about keeping this in sync with schemachanges but not too big a concern.  I think there's an `is_anonymous` response now

J0 (approved):

apps/docs/spec/supabase_py_v2.yml
silentworks: Good catch. I will add it in a new PR with further updates. I think in order to keep schema changes in sync we will just have to make sure that the Pydantic model is kept up to date and it's response is always added to the docs.

PR #25475, opened 2024-05-06
chore: update logging docs

dragarcia:


darora (approved):
Approving - the underlying change was not implemented correctly in July 23, and this change updates the docs to reflect reality.

PR #25782, opened 2024-05-06
feat: Allow for configuring target envs for Vercel integration

kamilogorek:
This PR depends on https://github.com/supabase/infrastructure/pull/18002 changes, but can be already tested locally using code from the linked PR.

I want to update the Vercel integration UI to look like that (feel free to change it as much as you want; the top alert is gone in second commit).

image

From technical side, what should change is:
- `env_sync_targets` is now returned for connection endpoints and should be used to determine what toggles should be on/off (it returns an array of 3 possible string values)
- `{ env_sync_targets: string[] }` should be send to update Vercel connection endpoint

It "works" now, but I'm sure there is a better way to work with arrays in Shadcdn form components.

Also API types codegen is completely borked for me locally and it generates a lot of gibberish :( (I used `api-shared` fwiw).

### DIFF
--- a/apps/studio/components/interfaces/Reports/Reports.CreateReportModal.tsx
+++ b/apps/studio/components/interfaces/Reports/Reports.CreateReportModal.tsx
@@ -46,7 +46,6 @@ export const CreateReportModal = ({ visible, onCancel, afterSubmit }: CreateRepo
         },
       })
       toast.success('New report created')
-
       const newReportId = res[0].id
       router.push(`/project/${ref}/reports/${newReportId}`)
       afterSubmit()
--- a/apps/studio/components/interfaces/Settings/Integrations/VercelIntegration/VercelIntegrationConnectionForm.tsx
+++ b/apps/studio/components/interfaces/Settings/Integrations/VercelIntegration/VercelIntegrationConnectionForm.tsx
@@ -1,29 +1,23 @@
 import { zodResolver } from '@hookform/resolvers/zod'
 import { useForm } from 'react-hook-form'
 import toast from 'react-hot-toast'
+import * as z from 'zod'
+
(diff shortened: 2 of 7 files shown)


apps/studio/components/interfaces/Settings/Integrations/VercelIntegration/VercelSection.tsx
```diff
@@ -65,7 +66,39 @@ const VercelSection = ({ isProjectScoped }: { isProjectScoped: boolean }) => {
...
+        foreign_project_id: '123',
+        organization_integration_id: '123',
+        metadata: {
+          id: '123',
+          name: 'asd',
+          framework: 'nextjs' as VercelFramework,
+        },
+      },
+    ],
+  }
```
alaister, line 101: Is the dummy data still needed here? If yes, would you mind adding an explainer comment?

apps/studio/components/interfaces/Settings/Integrations/VercelIntegration/VercelSection.tsx
```diff
@@ -125,12 +158,14 @@ You can change the scope of the access for Supabase by configuring
...
-      ? vercelIntegration?.connections.filter(
-          (connection) => connection.supabase_project_ref === project?.ref
-        )
-      : vercelIntegration?.connections) || []
+  // let connections =
+  //   (isProjectScoped
+  //     ? vercelIntegration?.connections.filter(
+  //         (connection) => connection.supabase_project_ref === project?.ref
+  //       )
+  //     : vercelIntegration?.connections) || []
```
alaister, line 166: Still needed?
```suggestion
```

alaister (approved):

PR #25784, opened 2024-05-06
Pass exposed schemas to lint query

saltcod:
Pass a string of exposed schemas to the lint query

screenshot-2024-05-06-at-16 52 18

### DIFF
--- a/apps/studio/data/lint/lint-query.ts
+++ b/apps/studio/data/lint/lint-query.ts
@@ -2,9 +2,9 @@ import { UseQueryOptions, useQuery } from '@tanstack/react-query'
 
 import { executeSql } from '../sql/execute-sql-query'
 import { lintKeys } from './keys'
+import { useProjectPostgrestConfigQuery } from 'data/config/project-postgrest-config-query'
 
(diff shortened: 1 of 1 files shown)


apps/studio/data/lint/lint-query.ts
```diff
@@ -841,13 +844,19 @@ export type ProjectLintsError = unknown
 
 export const useProjectLintsQuery = <TData = ProjectLintsData>(
   { projectRef, connectionString }: ProjectLintsVariables,
```
charislam, line 846: very minor nit: `ProjectLintsVariables` doesn't omit `exposedSchemas` here, so it's possible to pass it in when calling the hook from elsewhere (gets ignored, but could be confusing DX)

charislam (approved):

apps/studio/data/lint/lint-query.ts
saltcod: aah great catch! I updated this — did I do it right @charislam ?

charislam: yep works now! 👍🏼

saltcod: nice, great catch!

PR #25872, opened 2024-05-08
chore/fix studio jest tests

joshenlim:
Attempt to reinstate our broken unit tests via jest - I've managed to get most of them working except:
- `logs-query.test.js`


apps/studio/jest.config.mjs
```diff
@@ -8,16 +8,19 @@ const createJestConfig = nextJest({
 const config = {
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
   moduleDirectories: ['<rootDir>', 'node_modules'],
-  setupFiles: ['jest-canvas-mock', './tests/setup/radix'],
-  testEnvironment: 'jsdom',
-  testTimeout: 10000,
-  testRegex: '(.*\\.test.(js|jsx|ts|tsx)$)',
   maxConcurrency: 3,
   maxWorkers: '50%',
   moduleNameMapper: {
```
Ziinc, line 13: Might need to do a module name mapper for the `get`to a mock file for the fetchers, don't think it is resolving correctly, resulting in it not getting mocked.
Possible that it was removed in the setup files, as the mock was done globally if my memory serves correctly.
You could also do a manual module mock in jest using `jest.mock("lib/...")`

jordienr (approved):

PR #25880, opened 2024-05-08
feat: add http auth hook docs

### LINKED ISSUE #27339: Missing documentation: Auth Hooks -> Send Email Webhbook
# Improve documentation

## Link 
https://supabase.com/docs/guides/database/webhooks

## Describe the problem
In the supabase admin (https://supabase.com/dashboard/project/[id]/auth/hooks) user can create new "Send Email Webhook". This webhook functionality is not documented anywhere.  

Screenshot 2024-06-18 at 05 32 18

## Describe the improvement
This webhook should be documented in 2 places:
* https://supabase.com/docs/guides/database/webhooks 
* have example in the Admin SQL Editor -> Templates

J0:

YES

We introduce hook docs for HTTP Hooks. This is done by introducing a `HTTP Examples` tab and renaming the existing examples to `SQL Examples`.


apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +170,177 @@ Here's an example:
...
+  { status: 429, headers: { 'Content-Type': 'application/json', 'retry-after': 'true' } }
+)
+```
+
+Note that all responses, including errors responses, need a `Content-Type` of `application/json` - not specifying the appropriate `Content-Type` will result in the function returning an error response. Supabase Auth will in turn return an Internal Server Error.
+
+</TabPanel>
+</Tabs>
+
+Outside of runtime errors, both HTTP Hooks and Postgres Hooks return timeout errors. Postgres Hooks have **two** seconds to complete processing while HTTP Hooks should complete in **five** seconds. Both HTTP Hooks and Postgres Hooks are run in a transaction do limit the duration of execution to avoid delays in authentication process.
```
charislam, line 215: Perhaps consider dropping these timeouts in the [shared config file](https://github.com/supabase/supabase/blob/master/packages/shared-data/config.ts) for easy updating? Then you can access it like so:

```
<SharedData data="config">auth.[WHATEVER THE REST OF YOUR PATH IS]</SharedData>
```

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +170,177 @@ Here's an example:
...
+  scrollable
+  size="small"
+  type="underlined"
+  queryGroup="language"
+>
+<TabPanel id="postgres-connect-hook-locally" label="SQL">
+Modify the `auth.hook.<hook_name>` field and set `uri` to a value of `pg-functions://postgres/<schema>/<function_name>`
+
+```
+[auth.hook.<hook_name>]
```
charislam, line 241: Do we need instructions for how to do this via dashboard as well?

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +170,177 @@ Here's an example:
...
-Ensure that your hooks complete within **2 seconds** to avoid any errors.
+```
 
-### Connect a hook
+For security, we recommend against the use the `security definer` tag. The `security definer` tag specifies that the function is to be executed with the privileges of the user that owns it. When a function is created via the Supabase dashboard with the tag, it will have the extensive permissions of the `postgres` role which make it easier for undesirable actions to occur.
 
-In the dashboard, navigate to [`Authentication > Hooks (Beta)`](/dashboard/project/_/auth/hooks) and select the appropriate PostgreSQL function from the dropdown menu.
+We recommend that you do not use any tag and explicitly grant permissions to `supabase_auth_admin` as described above.
+
+Read more about `security definer` tag [in our database guide](https://www.notion.so/docs/guides/database/functions#security-definer-vs-invoker).
```
charislam, line 278:
```suggestion
Read more about `security definer` tag [in our database guide](/docs/guides/database/functions#security-definer-vs-invoker).
```

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +170,177 @@ Here's an example:
...
+
+There are no restrictions as to what language can be used to write a Postgres Hook. If [PL/pgSQL](https://www.postgresql.org/docs/current/plpgsql.html) is too difficult consider using the [plv8](https://www.notion.so/docs/guides/database/extensions/plv8) extension which lets you use JavaScript to define functions.
```
charislam, line 280:
```suggestion
There are no restrictions as to what language can be used to write a Postgres Hook. If [PL/pgSQL](https://www.postgresql.org/docs/current/plpgsql.html) is too difficult consider using the [plv8](/docs/guides/database/extensions/plv8) extension which lets you use JavaScript to define functions.
```

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +170,177 @@ Here's an example:
...
+
+Once done, save your Auth Hook as a migration in order to version the Auth Hook and share it with other team members. Run [`supabase migration new`](https://supabase.com/docs/reference/cli/supabase-migration-new) to create a migration.
```
charislam, line 282:
```suggestion
Once done, save your Auth Hook as a migration in order to version the Auth Hook and share it with other team members. Run [`supabase migration new`](/docs/reference/cli/supabase-migration-new) to create a migration.
```

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +170,177 @@ Here's an example:
...
+
+<Admonition type="caution">
+  If you're using the Supabase SQL Editor, there's an issue when using the `?` (*Does the string
+  exist as a top-level key within the JSON value?*) operator. Use a direct connection to the
+  database if you need to use it when defining a function.
+</Admonition>
```
charislam, line 288:
```suggestion
<Admonition type="caution">

If you're using the Supabase SQL Editor, there's an issue when using the `?` (*Does the string exist as a top-level key within the JSON value?*) operator. Use a direct connection to the database if you need to use it when defining a function.
  
</Admonition>

```

Adding the spaces will prevent Prettier from inserting the weird line breaks

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -268,6 +514,70 @@ revoke all
...
+
+Your company requires that a user can make an incorrect MFA verification attempt no more than three times in an hour. Redis is chosen as a backing store out of convenience since it is already present in the company setup.
+
+Create a Redis database using the [Upstash Console](https://console.upstash.com/) or [Upstash CLI](https://github.com/upstash/cli) to track verification attempts.
+
+Select the Global type to minimize the latency from all edge locations. Copy the `UPSTASH_REDIS_REST_URL` and `UPSTASH_REDIS_REST_TOKEN` to your .env file. Also ensure that you have `AUTH_MFA_VERIFICATION_ATTEMPT_SECRET` set.
+
+Create a hook to read and write information from Redis, which will record the last time a user had an incorrect MFA verification attempt for a factor. For example:
+
+```sql
```
charislam, line 536:
```suggestion
```js
```

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -268,6 +514,70 @@ revoke all
...
+    if (!valid) {
+      const userKey = `${user_id}:mfa_fail`
+      await redis.hincrby('mfa-attempts', userKey, 1)
+
+      const failedAttemptsCount = await redis.hget('mfa-attempts', userKey);
+      await redis.expire(userKey, secondsBeforeNextRetry)
+
+
+      if (failedAttemptsCount >= 3) {
+        return new Response(JSON.stringify({ decision: 'reject' }), { status: 403, headers: { "Content-Type": "application/json" } });
```
charislam, line 567: I'm a bit confused here, and maybe that's because I don't understand the flow well enough.

Does `valid` denote that the MFA verification as checked by Supabase Auth failed? Then wouldn't it reject anyway on that basis alone? How does the user-supplied reject decision change the flow?

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -532,13 +964,189 @@ Optional Claims: `jti`, `iss`, `nbf`, `app_metadata`, `user_metadata`, `amr`
 - `authentication_method` the authentication method used to request the access token. Possible values include:
```
charislam, line 964: Missing `anonymous`?

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -712,24 +1320,1536 @@ revoke execute
...
+import {
+    Webhook
+} from "https://esm.sh/standardwebhooks@1.0.0"
+import {
+    readAll
+} from "https://deno.land/std/io/read_all.ts";
+import * as base64 from "https://denopkg.com/chiefbiiko/base64/mod.ts";
+
+Deno.serve(async (req) => {
+const payload = await req.text()
```
charislam, line 1391: Indentation here

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -712,24 +1320,1536 @@ revoke execute
...
+                "Content-Type": "application/json"
+            }
+        });
+    }
+});
+```
+</TabPanel>
+
+</Tabs>
+````
```
charislam, line 2855:
```suggestion
```

apps/docs/content/guides/auth/auth-hooks.mdx
J0: You're right, this should be based on the timestamp not the counter. Ditto for the password verification.

Thanks for flagging!

J0: I'm removing the examples for now as I can't find a use case that makes sense and doesn't require reading from the database.

I would like to avoid opening a connection to the DB where possible since the Password Verification Hook is unauthenticated and it might be risky to allow access via an edge function

apps/docs/content/guides/auth/auth-hooks.mdx
J0: Probably can expand on the section on [Deploying the Hook](content/guides/auth/auth-hooks.mdx) to establish a link between the two

J0:
I've left out HTTP examples for the Password Verification Hook and MFA Verification Hooks as I realized the previous examples I had in mind didn't quite make sense. 

Will reinstate after a general review of everything else by Auth team.

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -13,11 +13,13 @@ Hooks help you:
...
 - Add additional custom claims to your JWT
 
-You can use hooks at specific points along an Auth flow to perform custom behavior via Postgres Functions. The following hooks are available:
+You can use hooks at specific points along an Auth flow to perform custom behavior via a function. The following hooks are available:
 
 | Hook                          | Available on Plan    |
 | ----------------------------- | -------------------- |
 | Custom Access Token           | Free, Pro            |
+| Send SMS                      | Free, Pro            |
+| Send email                    | Free, Pro            |
```
J0, line 22:
```suggestion
| Send Email                    | Free, Pro            |
```

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +169,180 @@ Here's an example:
...
+
+Errors are responses which contain status codes 400 and above. On a retry-able error, such as an error with a `429` or `503` status code, HTTP Hooks will attempt up to three retries with a back-off of two seconds.
+
+Here’s a sample HTTP retry schedule:
+
+| Time Since Start (HH:MM:SS) | Event                   | Notes                                             |
+| --------------------------- | ----------------------- | ------------------------------------------------- |
+| 00:00:00                    | Initial Attempt         | Initial invocation begins.                        |
+| 00:00:05                    | Initial Attempt Timeout | Initial invocation must complete.                 |
+| 00:00:07                    | Retry Start #1          | After 2 sec delay, first retry begins.            |
```
J0, line 193: Replace 2 sec with the constant here and elsewhere in the table

```suggestion
| 00:00:07                    | Retry Start #1          | After 2 sec delay, first retry begins.            |
```

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +169,180 @@ Here's an example:
...
+```
+
+You also need to grant usage to `supabase_auth_admin`:
+
+```sql
+grant usage on schema public to supabase_auth_admin;
+
+```
+
+Also revoke permissions from the `authenticated` and `anon` roles to ensure the function is not accessible by Supabase Serverless APIs.
```
J0, line 264: Update this to revoke from public as well

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +169,180 @@ Here's an example:
...
+uri = "http://host.docker.internal:54321/functions/v1/sms_sender"
+secret = "env(SEND_SMS_HOOK_SECRET)"
+```
+
+`host.docker.internal` is a special DNS name used in Docker to allow a container to access the host machine's network. This allows the Auth container to reach the Functions container
+
+Fill in the Hook Secret in `supabase/functions/.env`
+
+```jsx
+SEND_SMS_HOOK_SECRET = '<base64_standard_secret>'
```
J0, line 323: Adjust this to use v1,whsec_... or adjust entry to drop v1,whsec_ prefix

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -118,15 +169,180 @@ Here's an example:
...
-Ensure that your hooks complete within **2 seconds** to avoid any errors.
+## Deploying a Hook
 
-### Connect a hook
+In the dashboard, navigate to [`Authentication > Hooks`](/dashboard/project/_/auth/hooks) and select the appropriate function type (SQL or HTTP) from the dropdown menu.
 
-In the dashboard, navigate to [`Authentication > Hooks (Beta)`](/dashboard/project/_/auth/hooks) and select the appropriate PostgreSQL function from the dropdown menu.
+## Available Hooks
+
+Each Hook description contains an example JSONSchema which you can use in conjunction with [JSONSchema Faker](https://json-schema-faker.js.org/) in order to generate a mock payload. For HTTP Hooks, you can also use [the Standard Webhooks Testing Tool](https://www.standardwebhooks.com/simulate) to simulate a request.
```
J0, line 345: Maybe expand more on how to simulate a request

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -712,24 +1168,1537 @@ revoke execute
...
+Your company wishes to add assign permissions via the role claim on the `app_metadata` field. Add the role claim to the token via a Hook.
 
+```javascript
+import { Webhook } from 'https://esm.sh/standardwebhooks@1.0.0'
+import { readAll } from 'https://deno.land/std/io/read_all.ts'
+import * as base64 from 'https://denopkg.com/chiefbiiko/base64/mod.ts'
+
+Deno.serve(async (req) => {
+  const payload = await req.text()
+  const base64_secret = Deno.env.get('AUTH_CUSTOM_ACCESS_TOKEN_SECRET')
```
J0, line 1191: Maybe drop the AUTH_ prefix since we don't specify it the example where we introduce the secret

J0 (commented):
Edit these in the morrow

flesicek:
Hey @charislam, I'm reviewing the documentation for the Email hook. Example is given for "signup" email_action_type. I'm missing field confirmation_url that should contain the full url, that user should go to.

There are fields for components of that url, but not the full url. Is this field just missing in the sample and in the production will be provided? Or there is a change in the approach and we need to start to assemble this url?

J0:
Hey @flesicek,

At the moment one will need to assemble the URL - I can check with the team if we can provide an assembled URL if there is demand though.

flesicek:
One more think @J0 > Does the email hook need to return any value? I'm assuming that when hook is enabled, it would stop existing email integration so the system will not send the email 2 times. Or is there any condition based on return value from the hook that could enable the existing email flow again?

flesicek:
> Yes this is correct. Enabling the Hook disables the existing email sending flow. Do you happen to have a use case where you'd need the existing flow?

There may be one. I'm going to use the hook to translate the emails. But majority of them will be still send in English. It will be easier if the hook determines that the language is english that it would return some value (if nothing is returned, supabase will not send any email). This would save then supabase edge function calls. But that feels like optimisation problem rather than hard requirement.

kangmingtay (approved):

florianjuengermann:
The example for limit failed password verification attempts has a problem. You can still try hundreds of wrong passwords. You will get 429s but as soon as you hit the correct password you are in. @kangmingtay

florianjuengermann:
No, the problem is not that you can try a lot over a long period. The problem is that you can try a lot over a very short period and still sign in if you hit the right password.

You need to first check for the rate limit and return 429 regardless of whether the password is correct or not. Otherwise, your account is not "locked" if you can still sign in with the correct password.

PR #25946, opened 2024-05-09
Custom error boundary

joshenlim:
Implement a custom error boundary that will notify Sentry if a full page crash occurs in an effort to ensure that our dashboard (ideally) never runs into that scenario, and that we take a pro-active stance towards such instances.

This also allows us to create a custom error page that's slightly more helpful than next.js's default where we add a button to let users report the issue via support, with the subject and message all filled out. We'll definitely relook at this one to make it prettier as well
image


apps/studio/components/ui/ErrorBoundaryState.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const ErrorBoundaryState = ({ error, resetErrorBoundary }: FallbackProps) => {
+  const router = useRouter()
+  const message = `Path name: ${router.pathname}\n\n${error.stack}`
+
+  return (
+    <div className="w-screen h-screen flex items-center justify-center flex-col gap-y-3">
+      <div className="flex items-center flex-col gap-y-1">
+        <p className="text-sm">
+          Application error: a client-side exception has occured (see browser console for more
```
alaister, line 16: lol okay misspell
```suggestion
          Application error: a client-side exception has occurred (see browser console for more
```

alaister (approved):
LGTM! 🔥

PR #25953, opened 2024-05-09
fix url query breaking active tab

jordienr:

YES

- fixes bug where items didnt show as active when we had query params in the url
- adds forwardRef to nav menu item component.


apps/studio/hooks/misc/useCurrentPath.tsx
```diff
@@ -0,0 +1,17 @@
+import { useRouter } from 'next/router'
+
+type UseCurrentPathOptions = {
+  withQuery: boolean
+}
+export const useCurrentPath = (options?: UseCurrentPathOptions) => {
```
alaister, line 6: is `options` needed here? doesn't seem to be used...

MildTomato (approved):

PR #25969, opened 2024-05-09
enhance(docs): reorg google auth guide (web)

charislam:
- rearrange the page
- revalidate and revise the web instructions


apps/docs/content/guides/auth/social-login/auth-google.mdx
```diff
@@ -516,66 +171,473 @@ Before you can use Sign in with Google, you need to obtain a [Google Cloud Platf
...
+   ```
+
+   Note that the nonce should be the same in both places, but because Supabase Auth expects the provider to hash it (SHA-256, hexadecimal representation), you need to provide a hashed version to Google and a non-hashed version to `signInWithIdToken`.
+
+   You can get both versions by using the in-built `crypto` library:
+
+   ```js
+   // Adapted from https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string
+
+   const nonce = btoa(
```
charislam, line 303: I tweaked this to work browser-side, would appreciate a second set of eyes

apps/docs/content/guides/auth/social-login/auth-google.mdx
```diff
@@ -516,66 +171,473 @@ Before you can use Sign in with Google, you need to obtain a [Google Cloud Platf
...
+       nonce: '<NONCE>',
+     })
+   }
+   ```
+
+   Note that the nonce should be the same in both places, but because Supabase Auth expects the provider to hash it (SHA-256, hexadecimal representation), you need to provide a hashed version to Google and a non-hashed version to `signInWithIdToken`.
+
+   You can get both versions by using the in-built `crypto` library:
+
+   ```js
```
charislam, line 300: Rewrote this to work browser-side since that seems to fit more with the example (script tags, functions in browser global scope). Would appreciate a second set of 👀 !

apps/docs/content/guides/auth/social-login/auth-google.mdx
```diff
@@ -1,26 +1,19 @@
 ---
 id: 'auth-google'
 title: 'Login with Google'
-description: 'Use Sign in with Google on the web, in native apps or with Chrome extensions'
+description: 'Use Sign In with Google on the web, in native apps or with Chrome extensions'
```
hf, line 4: Is this the new styling we apply? Maybe we should leave it for another PR where all `Sign in` is renamed to `Sign In`?

Your call!

apps/docs/content/guides/auth/social-login/auth-google.mdx
hf: I'm not sure I would class it as "browser side" since it's actually the olden days server side. But sure we can try it. I don't think I've seen anyone in Supabase use it.

hf (approved):
LGTM.

apps/docs/content/guides/auth/social-login/auth-google.mdx
charislam: Yeah I'll revert this 👍🏼

charislam, line 4:
```suggestion
description: 'Use Sign in with Google on the web, in native apps or with Chrome extensions'
```

PR #25974, opened 2024-05-10
Fix wrapper label not present

saltcod:
if user creates wrapper with sql and doesn't have a lable present, the page crashes

To reproduce: 
1. run this https://gist.github.com/saltcod/140a95d3161bf259c704efcd96dffecd
2. visit Database → wrappers → open Stripe wrapper
3. click Edit https://share.cleanshot.com/VMDcJcrG

Page crashes b/c label expected 

https://app.hubspot.com/contacts/19953346/record/0-5/2561901162

### DIFF
--- a/apps/studio/components/interfaces/Database/Wrappers/EditWrapper.tsx
+++ b/apps/studio/components/interfaces/Database/Wrappers/EditWrapper.tsx
@@ -338,7 +338,7 @@ const EditWrapper = () => {
                     ) : (
                       <div className="space-y-2">
                         {wrapperTables.map((table, i) => {
-                          const label = wrapperMeta.tables[table.index].label
+                          const label = wrapperMeta.tables[table.index]?.label
 
                           return (
                             <div
--- a/apps/studio/components/interfaces/Database/Wrappers/EditWrapper.tsx
+++ b/apps/studio/components/interfaces/Database/Wrappers/EditWrapper.tsx
@@ -41,6 +41,7 @@ import {
   makeValidateRequired,
 } from './Wrappers.utils'
 import WrapperTableEditor from './WrapperTableEditor'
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Database/Wrappers/EditWrapper.tsx
```diff
@@ -338,7 +338,7 @@ const EditWrapper = () => {
                     ) : (
                       <div className="space-y-2">
                         {wrapperTables.map((table, i) => {
-                          const label = wrapperMeta.tables[table.index].label
+                          const label = wrapperMeta.tables[table.index]?.label
```
charislam, line 341: If the table is not listed in wrapperMeta, you get a weird colon:
image

I guess a more general question, is there even a reason to go back and look up the constants to find the label? There are a lot of allowed tables in the Stripe FDW that don't seem to be included in the constants, so this will often be missing, and for the ones that are included, the label doesn't seem to add any information that's not immediately obvious from the table name (which is in the line immediately above). Would it be simpler to just drop it? Or rename it `Columns:`, which might be more descriptive?

joshenlim: ^ good point here, probably don't need to do the round trip for label - looking into it

joshenlim:
okay adding a few changes:

- Changed the label here to just "columns" as Charis suggested, agreed that its more descriptive
- Added a text field "Target" to visually show what the target of the wrapper is, I realised this is an important piece of info missing, in particular if the wrapper is clickhouse, mssql etc which targets arbitrary table names
image
- Added wrapper label name when editing 
image

alaister (approved):
LGTM!

saltcod:
Confirmed on prod.

PR #25981, opened 2024-05-10
feat: Add native sign in for Kakao with OIDC

nectarine:

YES

feature, docs update

https://github.com/orgs/supabase/discussions/23368
Kakao provides two different keys which are REST API key, and Native app key.
Supabase accept only one key(currently REST API key), so it should accept both keys to provide seamless sign in process in native apps. 

## What is the new behavior?

스크린샷 2024-05-10 오후 2 37 33

A local dev studio doesn't show me a `Provider` section in `Authentication` > `Configuration`, but I added a new field for the `Native app key`, similar to `Authorized Client IDs` of `Google`

Also, I add the docs with native app sign in with trying my best to preserve original author's tone.

## Additional context

supabase.auth also must be fixed to accept this new value `EXTERNAL_KAKAO_NATIVE_APP_KEY`.
(https://github.com/supabase/auth/blob/ed2b4907244281e4c54aaef74b1f4c8a8e3d97c9/internal/api/token_oidc.go#L77)
I'll make a new PR in that repository after this PR is accepted.

### DIFF
--- a/apps/studio/components/interfaces/Settings/General/CustomDomainConfig/CustomDomainVerify.tsx
+++ b/apps/studio/components/interfaces/Settings/General/CustomDomainConfig/CustomDomainVerify.tsx
@@ -61,7 +61,7 @@ const CustomDomainVerify = ({ projectRef, customDomain, settings }: CustomDomain
           </h4>
           <p className="text-sm text-foreground-light">
             Set the following TXT record(s) in your DNS provider, then click verify to confirm your
-            control over the domain
+            control over the domain.
           </p>
           <p className="text-sm text-foreground-light">
             Records which have been successfully verified will be removed from this list below.


kangmingtay:
hi @nectarine, thanks for contributing! can you please link the relevant docs from kakao that explains the differences between the "REST API Key" and "Native App Key"?

> I'll make a new PR in that repository after this PR is accepted.

Unfortunately, that isn't possible because adding a new field to support "Native App Key" without first making changes on the backend will cause alot of confusion for our users.

> supabase.auth also must be fixed to accept this new value `EXTERNAL_KAKAO_NATIVE_APP_KEY` 

The way this works with Google and Apple is that we accept a comma-separated string of client ids in the `EXTERNAL_XXX_CLIENT_ID` variable. The first `client_id` is assumed to be the client id used for Web OAuth and the others are accepted client ids for native / mobile use cases. I don't think we need to create a separate env var to store the `EXTERNAL_KAKAO_NATIVE_APP_KEY` since we can just reuse `EXTERNAL_KAKAO_CLIENT_ID` .

nectarine:
Absolutely.
https://developers.kakao.com/docs/latest/en/getting-started/app#app-key

Native app key: Used to [initialize](https://developers.kakao.com/docs/latest/en/android/getting-started#init) the Kakao SDK for Android or [initialize](https://developers.kakao.com/docs/latest/en/ios/getting-started#init) the Kakao SDK for iOS.

Just like Google, Kakao requires different keys with different platforms.

And of course, yes, it doesn't have to be a special field for Native App Key.
I just tried to reusing patterns in Google and Apple. They both have a single id field and have extra ids field or one more special field(bundle Id) ;) 

Modifying `EXTERNAL_KAKAO_CLIENT_ID` with a comma-separated string form would work.

I'll make a updated commit in a few days.

PR #26290, opened 2024-05-14
Chore/support checking replication lag in database reports

joshenlim:
image
image
image


alaister (approved):
LGTM!

Just want to check though, the unit is definitely seconds right...? ~53 hours seems very long for a replication lag...
Screenshot 2024-05-14 at 16 06 48

alaister (approved):
Double approve 🔥

PR #26304, opened 2024-05-14
Chore: block multiple account deletion requests

monicakh:


apps/studio/components/interfaces/Account/Preferences/DeleteAccountButton.tsx
```diff
@@ -26,6 +26,29 @@ import {
   Input_Shadcn_,
 } from 'ui'
 
+const DELETE_REQUEST_KEY = 'accountDeletionRequest'
```
ivasilov, line 29: Can we move this to `LOCAL_STORAGE_KEYS` in `supabase/apps/studio/lib/constants/index.ts`? Add a small comment to it and the string `accountDeletionRequest` should be prefixed with 
`supabase` (`supabase-account-deletion-request`).

It's easier to track which keys we're using for accessing local storage.

ivasilov (approved):

PR #26331, opened 2024-05-15
Chore/remove optimistic rendering for spinning up replicas

joshenlim:
- Remove optimistic rendering when creating replicas as API now immediately inserts a new row into the DB rather than through an async process with the worker
- Fix missing edge when load balancer initially renders


alaister (approved):
LGTM!

PR #26341, opened 2024-05-15
move linting to api

saltcod:
linting has been moved to the api 
this pr removes the original linting SQL and replaces with this api call


apps/studio/data/lint/lint-query.ts
```diff
@@ -1,923 +1,41 @@
...
-    },
-    signal
-  )
+  const { data, error } = await get(`/platform/projects/{ref}/run-lints`, {
+    params: { path: { ref: projectRef } },
+    signal,
+  })
 
-  return result
+  if (error) throw error
```
joshenlim, line 23: handleError(error)

joshenlim (approved):

PR #26378, opened 2024-05-16
Chore/support more granular updates via events for spinning up replica

joshenlim:
Spinning up replicas will show more granular status updates while the replica is initializing:
image

If estimates are available, hovering over the progress bar will show:
image

If an error happens during set up, error will show accordingly:
image


alaister (approved):
LGTM!

PR #26382, opened 2024-05-16
resolve find of undefined in api

joshenlim:


apps/studio/pages/project/[ref]/api/index.tsx
```diff
@@ -66,14 +66,9 @@ const DocView = observer(() => {
...
-  const { data, error } = useProjectApiQuery({
-    projectRef,
-  })
-
+  const { data, error } = useProjectApiQuery({ projectRef })
   const apiService = data?.autoApiService
-  const anonKey = apiService?.service_api_keys.find((x) => x.name === 'anon key')
-    ? apiService.defaultApiKey
-    : undefined
+  const anonKey = apiService?.defaultApiKey ?? undefined
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/api/index.tsx:71]: this is the main fix here, the rest are just rearranging stuff to make sure return statements are all the way at the bottom

Comment (joshenlim) [reply apps/studio/pages/project/[ref]/api/index.tsx:71]: was thinking here that the .find seems really unnecessary - but please lemme know if im overlooking something

alaister (commented):
Getting this on the API docs section:
Screenshot 2024-05-17 at 14 53 15

joshenlim:
pushed a fix up! thanks for the catch 🙏 🙏

alaister (approved):
LGTM!

PR #26384, opened 2024-05-16
First round of wrapping RQ errors with handleError

joshenlim:


apps/studio/data/analytics/org-daily-compute-stats-query.ts
```diff
@@ -34,7 +35,7 @@ export async function getOrgDailyComputeStats(
     params: { path: { slug: orgSlug }, query: { projectRef, startDate, endDate } },
     signal,
   })
-  if (error) throw error
+  if (error) throw handleError(error)
```
ivasilov, line 38: We shouldn't throw the result from the `handleError` function.

alaister (approved):
LGTM!

PR #26402, opened 2024-05-16
Chore/sentry fixes

saltcod:
— make the full page crash event show up as an error, instead of the default (Info)([screenshot](https://share.cleanshot.com/Gq6fRXfC))
— fix crash on settings/auth page when page is translated
— add key to fix console error


alaister (approved):
LGTM!

PR #26440, opened 2024-05-17
Chore: Add Validation to Account Name and Organization Name Text Fields to Prevent Malicious Input

monicakh:

Currently, the account name and organization name text fields lack validation, allowing any characters to be used. This vulnerability can be exploited for malicious purposes. For example, a malicious link can be saved in these text boxes. When users receive an invitation to join a New Relic account, these names render as valid links in email clients. Since the email is from a trusted domain like New Relic, users may click on these links, which could lead to harmful sites.

### Solution:

Added a regular expression validation to the relevant fields.

### Testing:

- Manually tested the account First name / Last name and the Organization name text fields to ensure that only valid characters are accepted.

image

image

### DIFF
--- a/apps/studio/components/interfaces/Account/Preferences/ProfileInformation.tsx
+++ b/apps/studio/components/interfaces/Account/Preferences/ProfileInformation.tsx
@@ -19,8 +19,18 @@ import type { Profile } from 'data/profile/types'
 import type { FormSchema } from 'types'
 
 const FormSchema = z.object({
-  first_name: z.string().optional(),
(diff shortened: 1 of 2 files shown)


MildTomato:
😢

charislam:
Hmm I would expect a URL to be a valid organization name, to be fair. I should be able to name my org `company.io` if I want to... would it be better to escape somehow in the HTML template instead to prevent email client auto-linking?

saltcod:
closing, can reopen if we still want to do this

PR #26474, opened 2024-05-17
fix: Add sourcemaps and error stacks in Sentry errors

ivasilov:


apps/studio/pages/_app.tsx
```diff
@@ -112,9 +112,7 @@ function CustomApp({ Component, pageProps }: AppPropsWithLayout) {
 
   const errorBoundaryHandler = (error: Error, info: ErrorInfo) => {
     console.error(error.stack)
-    Sentry.captureMessage(`Full page crash: ${error.message || 'Unknown error message'}`, {
-      level: 'error',
-    })
+    Sentry.captureException(error, { level: 'error' })
```
kamilogorek, line 115:
```suggestion
    Sentry.captureException(error)
```

This is a default.

apps/studio/pages/_app.tsx
```diff
@@ -111,9 +111,10 @@ function CustomApp({ Component, pageProps }: AppPropsWithLayout) {
   )
 
   const errorBoundaryHandler = (error: Error, info: ErrorInfo) => {
-    console.error(error.stack)
-    Sentry.captureMessage(`Full page crash: ${error.message || 'Unknown error message'}`, {
-      level: 'error',
+    Sentry.withScope(function (scope) {
+      console.error(error.stack)
+      scope.setTag('globalErrorBoundary', true)
+      Sentry.captureException(error, { level: 'error' })
```
kamilogorek, line 117: Change it to:

```js
    Sentry.withScope(function (scope) {
      scope.setTag('globalErrorBoundary', true)
      Sentry.captureException(error)
    })
    console.error(error.stack)
```

This way you won't have unnecessary console breadcrumb recorded which will make UI noisier.

saltcod (approved):

apps/studio/pages/_app.tsx
ivasilov: Fixed.

PR #26511, opened 2024-05-17
Update index_advisor.mdx

tylerfontaine:
Fix universal dashboard link to query-performance

YES

docs update, ...

The link to the dashboard has a * instead of a _ to allow uses to choose their project, which results in a 404.

## What is the new behavior?
It properly lets users select their own project.


alaister (approved):

tylerfontaine:
@alaister @charislam Not sure what else might be needed here? I see Alaister added the prettier ignore, but it didn't seem to work.

Also, had no idea underscores would need to be escaped. Thanks for the catch!

charislam:
Pushed a fix! Should work now 🤞🏼

PR #26532, opened 2024-05-18
bug: need to include "use client" in error.tsx file

dhairyathedev:

YES

Bug fix

In the latest Supabase documentation, there's a code snippet for error.tsx, which is currently set up as a server component. 
This causes an error in the Next.js application. To convert error.tsx into a client component, we need to add `use client` at the top of the file.

Docs Page: https://supabase.com/docs/guides/auth/server-side/nextjs

You can verify this at the official nextjs documentation - https://nextjs.org/docs/app/building-your-application/routing/error-handling

Screenshot 2024-05-18 at 4 26 29 PM

### DIFF
--- a/apps/docs/content/guides/auth/server-side/nextjs.mdx
+++ b/apps/docs/content/guides/auth/server-side/nextjs.mdx
@@ -382,6 +382,7 @@ export async function signup(formData: FormData) {
 ```
 
 ```ts app/error/page.tsx
+"use client";
 export default function ErrorPage() {
   return <p>Sorry, something went wrong</p>
 }


charislam:
Thanks @dhairyathedev! The difference between the tutorial and your case is that you're using an error-catching page (`error.tsx`) and the tutorial is using a normal page that just happens to be named `error` (`error/page.tsx`). Since it's a normal page, it's renderable server-side.

Throwing an error and catching it is certainly another possible pattern to use for Next.js, but since that's not what we do in the tutorial, I'm going to close this one for now.

Feel free to reopen if I've misunderstood the problem.

Issue #20370, opened 2024-01-12
`npm install` fails when installing repo dependencies for development


alt-glitch:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug
On running `npm install` after cloning the repository, it fails with the following log (some parts trimmed) :
```
glitch@instance-3:~/supabase$ npm i
npm ERR! code 1
npm ERR! path /home/glitch/supabase/node_modules/@gregnr/libpg-query
npm ERR! command failed
npm ERR! command sh -c node-pre-gyp install --fallback-to-build
npm ERR! make: Entering directory '/home/glitch/supabase/node_modules/@gregnr/libpg-query/build'
npm ERR!   CC(target) Release/obj.target/nothing/node_modules/node-addon-api/nothing.o
npm ERR! rm -f Release/obj.target/node_modules/node-addon-api/nothing.a Release/obj.target/node_modules/node-addon-api/nothing.a.ar-file-list; mkdir -p `dirname Release/obj.target/node_modules/node-addon-api/nothing.a`
npm ERR! ar crs Release/obj.target/node_modules/node-addon-api/nothing.a @Release/obj.target/node_modules/node-addon-api/nothing.a.ar-file-list
npm ERR!   COPY Release/nothing.a
npm ERR!   ACTION binding_gyp_queryparser_target_prebuild_dependencies libpg_query/include/pg_query.h
npm ERR! make[1]: Entering directory '/tmp/tmp.Ztle2IBnbL/libpg_query'
npm ERR! compiling src/pg_query.c
<--- trimmed --->
npm ERR! make[1]: Leaving directory '/tmp/tmp.Ztle2IBnbL/libpg_query'
npm ERR! ERROR: 'make' command failed
npm ERR! make: Leaving directory '/home/glitch/supabase/node_modules/@gregnr/libpg-query/build'
[...]

PR #26618, opened 2024-05-21
chore: upgrade libpg-query

gregnr:
[`libpg-query`](https://github.com/launchql/libpg-query-node/tree/15-latest) allows us to parse SQL using the real Postgres parser. We've been using a fork `@gregnr/libpg-query` that compiles the parser to WASM so that it can be used in the browser. The [WASM work](https://github.com/launchql/libpg-query-node/pull/34) has now been merged upstream, so this PR updates all `@gregnr/libpg-query` references to `libpg-query`.

### How it works under the hood
`libpg-query` used to only have Node NAPI bindings to the Postgres parsing logic (C code). Now that the WASM target is available, `libpg-query` will choose between the two: WASM when in a browser environment, NAPI when in a server environment. It uses `package.json` [exports](https://github.com/launchql/libpg-query-node/blob/f87e7294f999c5d469e0ec8ad0c088d73012b1e1/package.json#L20-L31) to do this.

Currently using the PG v15 version of the lib.

### DIFF
--- a/apps/database-new/lib/utils.ts
+++ b/apps/database-new/lib/utils.ts
@@ -1,5 +1,5 @@
-import { parseQuery } from '@gregnr/libpg-query'
 import dayjs from 'dayjs'
+import { parseQuery } from 'libpg-query'
 import { compact } from 'lodash'
 import { z } from 'zod'
 import { PostgresColumn, PostgresTable } from './types'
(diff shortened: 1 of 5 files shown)


saltcod (approved):

fsansalvadore:
Possibly fixes issue #20370

PR #26655, opened 2024-05-22
feat: automate mgmt api spec update

phamhieu:

YES

Automate mgmt api specs update


apps/docs/lib/refGenerator/helpers.ts
```diff
@@ -240,7 +240,9 @@ export function gen_v3(spec: OpenAPIV3.Document, dest: string, { apiUrl }: { api
         ...operation,
         path: key,
         fullPath,
-        operationId: slugify(operation.summary!),
+        operationId: isValidSlug(operation.operationId)
```
charislam, line 243: This change breaks the self-hosting references because they assume that the operation summary is used here, for example check /docs/reference/self-hosting-auth/get-a-user

charislam (changes requested):

apps/docs/lib/refGenerator/helpers.ts
phamhieu: It should be resolved now

charislam (approved):

PR #26670, opened 2024-05-22
fix/various improvements to Free plan downgrade [GEN-8955]

ecktoteckto:
The following was fixed/improved for the two-step downgrade to plan Free:

**Step 1**
- (fix) the addons that will be removed on downgrade were not shown due to missing parentheses
- (improvement) the text that the downgrade will take place at the end of the current billing cycle is now only shown for orgs billed via Fly

**Step 2 (Survey/final confirmation)**
- (fix) no hint regarding project restart was shown for projects that are created with Micro compute instead of being upgraded to Micro compute after project creation. Resulting in a mismatch between the projects listed in Step 1 and Step 2
- (improvement) the text that the downgrade will happen immediately after clicking confirm was changed for orgs billed via Fly
- (improvement) the success message after clicking confirm was adjusted regarding immediate/scheduled downgrade and projects being restarted
- (improvement) the text regarding credits that will be given is now only shown if the downgrade happens immediately

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/DowngradeModal.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/DowngradeModal.tsx
@@ -76,33 +76,33 @@ const DowngradeModal = ({
             <Alert
               withIcon
               variant="warning"
-              title="Downgrading to the free plan will lead to reductions in your organization's quota"
+              title="Downgrading to the Free plan will lead to reductions in your organization's quota"
             >
               <p>
-                If you're already past the limits of the free plan, your projects could become
+                If you're already past the limits of the Free plan, your projects could become
                 unresponsive or enter read only mode.
               </p>
             </Alert>
 
-            {(previousProjectAddons.length ?? 0) > 0 ||
-              (hasInstancesOnMicro && downgradingToNano && (
-                <Alert title={`Projects affected by the downgrade`} variant="warning" withIcon>
-                  <ul className="space-y-1 max-h-[100px] overflow-y-auto">
-                    {previousProjectAddons.map((project) => (
-                      <ProjectDowngradeListItem key={project.ref} projectAddon={project} />
-                    ))}
+            {((previousProjectAddons.length ?? 0) > 0 ||
+              (hasInstancesOnMicro && downgradingToNano)) && (
+              <Alert title={`Projects affected by the downgrade`} variant="warning" withIcon>
+                <ul className="space-y-1 max-h-[100px] overflow-y-auto">
+                  {previousProjectAddons.map((project) => (
+                    <ProjectDowngradeListItem key={project.ref} projectAddon={project} />
+                  ))}
 
-                    {projects
-                      .filter((it) => it.infra_compute_size === 'micro')
-                      .map((project) => (
-                        <li className="list-disc ml-6" key={project.id}>
-                          {project.name}: Compute will be downgraded. Project will also{' '}
-                          <span className="font-bold">need to be restarted</span>.
-                        </li>
-                      ))}
-                  </ul>
-                </Alert>
-              ))}
+                  {projects
+                    .filter((it) => it.infra_compute_size === 'micro')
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Support/Success.tsx
```diff
@@ -29,8 +38,15 @@ const Success = ({ sentCategory = '' }: SuccessProps) => {
       <div className="flex items-center flex-col space-y-2">
         <h3 className="text-xl">Support request successfully sent!</h3>
         <p className="text-sm text-foreground-light">
-          We will reach out to you at <span className="text-foreground">{respondToEmail}</span>
+          We will reach out to you at <span className="text-foreground">{respondToEmail}</span>.
         </p>
+        {selectedProject !== 'no-project' && (
```
kevcodez, line 43: Might be better to check for `project` being set here instead of selectedProject being some string

apps/studio/components/interfaces/Support/Success.tsx
```diff
@@ -29,8 +38,15 @@ const Success = ({ sentCategory = '' }: SuccessProps) => {
...
+          <p className="text-sm text-foreground-light">
+            Your ticket has been logged for the project{' '}
```
kevcodez, line 45: Given the customer just selected this, not sure if we're adding any value by repeating what they have selected?

If we add value, project name should be enough, no need to add a ref id or anything

kevcodez: Is this even something you changed? Seems kinda unrelated and might just wrongfully show up here?

apps/studio/components/interfaces/Support/SupportForm.tsx
```diff
@@ -40,9 +40,10 @@ const INCLUDE_DISCUSSIONS = ['Problem', 'Database_unresponsive']
 
 export interface SupportFormProps {
   setSentCategory: (value: string) => void
+  setSelectedProject: (value: string) => void
 }
 
-const SupportForm = ({ setSentCategory }: SupportFormProps) => {
+const SupportForm = ({ setSentCategory, setSelectedProject }: SupportFormProps) => {
```
kevcodez, line 46: Is this even something you changed? Seems kinda unrelated and might just wrongfully show up here?

apps/studio/components/ui/FilterPopover.tsx
```diff
@@ -27,6 +27,9 @@ export const FilterPopover = ({
 
   const formattedOptions = activeOptions.map((option) => {
     const base = options.find((x) => x[valueKey] === option)
+    if (!base || !base[labelKey]) {
```
kevcodez, line 30: Is this even something you changed? Seems kinda unrelated and might just wrongfully show up here?

apps/studio/pages/support/new.tsx
```diff
@@ -82,9 +86,16 @@ const SupportPage = () => {
             ].join(' ')}
           >
             {sentCategory !== undefined ? (
-              <Success sentCategory={sentCategory} />
```
kevcodez, line 85: Is this even something you changed? Seems kinda unrelated and might just wrongfully show up here?

apps/studio/components/interfaces/Support/Success.tsx
ecktoteckto: No, that's not my changes.

apps/studio/components/interfaces/Support/SupportForm.tsx
ecktoteckto: No, that's not my changes.

apps/studio/pages/support/new.tsx
ecktoteckto: No, that's not my changes.

apps/studio/components/ui/FilterPopover.tsx
ecktoteckto: No, that's not my changes.

kevcodez (approved):

PR #26676, opened 2024-05-22
enhance(refs): add params and return examples for database and auth

charislam:
- Make sure parameters and return examples are documented for all database and auth functions


mansueli:
LGTM

mansueli (approved):

apps/docs/spec/supabase_js_v2.yml
```diff
@@ -237,6 +237,45 @@ functions:
...
+                  A custom data object to store additional user metadata.
+              - name: captchaToken
+                isOptional: true
+                type: string
+              - name: channel
+                isOptional: true
+                type: sms | whatsapp
+                description: >
+                  The channel to use for sending messages.
+                  Only for phone signups.
```
gregnr, line 278: Safe to say this can't be generated upstream from `auth-js`?

gregnr (approved):

apps/docs/spec/supabase_js_v2.yml
charislam: They aren't currently generated (the TS parsing ignores union types). I have a PR to improve this but it needs a few more iterations to clean it up, and as a quick fix doing it manually was fastest

PR #26693, opened 2024-05-23
chore: switch to new slug based invoice endpoints

kevcodez:

MGMT API introduces new org-slug based endpoints for accessing invoices.

Additionally
- Ignore generated `api.d.ts` file when checking for spelling errors
- Remove misleading `codegen` command from studio, as it has moved to a shared package

## Additional context

Can only be merged after MGMT API prod deployment.


ecktoteckto:
**question:** Knowing that this endpoint is somehow different because of the authenticated user involved. But are we planning on moving the endpoint for overdue invoices from `/platform/stripe...` to `/platform/organizations...` as well?

ecktoteckto (approved):

PR #26696, opened 2024-05-23
Fix text confirmation modal not resetting input field

joshenlim:


packages/ui-patterns/Dialogs/TextConfirmModal.tsx
```diff
@@ -100,6 +100,10 @@ const TextConfirmModal = forwardRef<
       onConfirm()
     }
 
+    useEffect(() => {
```
MildTomato, line 103: this seems wrong.

react-hook-form probably has a sensible way to handle this

MildTomato:
approving, but we agreed this is a temporary solution

MildTomato (approved):

PR #26704, opened 2024-05-23
Chore/add backup warning

saltcod:
Adds alerts to backup ui that storage objects aren't part of backups

screenshot-2024-05-23-at-09 05 49
screenshot-2024-05-23-at-09 03 50


apps/studio/components/interfaces/Database/Backups/BackupsList.tsx
```diff
@@ -88,6 +89,14 @@ const BackupsList = () => {
                 title="You need additional permissions to trigger a scheduled backup"
               />
             )}
+            <Alert_Shadcn_ variant="default">
```
joshenlim, line 92: minor: maybe lets bring this into its own component, so scheduled and PITR can use the same component

joshenlim (approved):

PR #26707, opened 2024-05-23
Add note regarding branching being disabled for GitHub integration panel

kamilogorek:
Small QoL improvement, as it bit myself multiple times already.


alaister (approved):
LGTM!

PR #26709, opened 2024-05-23
chore: get lint results from api

soedirgo:
Use `/platform/projects/[ref]/run-lints` to get lint results.

Q: how does this affect self-hosted? Do we already expose advisors there? Do we need to keep the hardcoded SQL around for those?


soedirgo:
Closing in favor of https://github.com/supabase/supabase/pull/26341/files

PR #26792, opened 2024-05-27
Chore/gracefully handle large text or json in table editor

joshenlim:
Currently setting the limit for max characters we allow to render in the dashboard to be at 10Kb, although happy to debate what that number should be (I think it can be higher)

Within the grid, if the cell value is beyond the threshold:
image

Clicking load will then attempt to load the whole value:
image

Likewise within the side panel:
If the value is too large, the input field itself is disabled - CTA to edit via expanding the editor
image

Side note: updated the actions here to be behind a dropdown so it takes up less real estate
image

Same UX:
image


apps/studio/data/table-rows/table-rows-query.ts
```diff
@@ -94,22 +102,26 @@ export const getTableRowsSqlQuery = ({
...
 
-  let queryChains =
-    enumArrayColumns.length > 0
-      ? query
-          .from(table.name, table.schema ?? undefined)
-          .select(`*,${enumArrayColumns.map((x) => `"${x}"::text[]`).join(',')}`)
-      : query.from(table.name, table.schema ?? undefined).select()
+  let queryChains = query
+    .from(table.name, table.schema ?? undefined)
+    .select(`*,${truncatedColumns.join(',')}`)
```
alaister, line 124: Might need to add a check here for when truncatedColumns is empty. Getting this:
Screenshot 2024-05-27 at 20 54 04

joshenlim: pushed up a fix! 🙏

saltcod (approved):

alaister (approved):
LGTM!

PR #26793, opened 2024-05-27
[Dashboard] Postgrest API settings renamed to "Data API" settings. now using react-hook-form

MildTomato:

YES

- refactor API postgrest settings to react-hook-form
- schema selection stops submitting on change now 
- be aware, types have been changed

### DIFF
--- a/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
+++ b/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
@@ -197,21 +197,13 @@ const PostgrestConfig = () => {
                             <MultiSelectorInput placeholder="Select schemas for Data API..." />
                           </MultiSelectorTrigger>
                           <MultiSelectorContent>
-                            {schema.length <= 0 ? (
-                              <MultiSelectorList>
-                                <MultiSelectorItem key={'empty'} value={'no'}>
-                                  no
+                            <MultiSelectorList>
+                              {schema.map((x, i) => (
+                                <MultiSelectorItem key={x.id + '-' + i} value={x.name}>
+                                  {x.name}
                                 </MultiSelectorItem>
-                              </MultiSelectorList>
-                            ) : (
-                              <MultiSelectorList>
-                                {schema.map((x, i) => (
-                                  <MultiSelectorItem key={x.id + '-' + i} value={x.name}>
-                                    {x.name}
-                                  </MultiSelectorItem>
-                                ))}
-                              </MultiSelectorList>
-                            )}
+                              ))}
+                            </MultiSelectorList>
                           </MultiSelectorContent>
                         </MultiSelector>
                       )}
--- a/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
+++ b/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
@@ -140,7 +140,7 @@ const PostgrestConfig = () => {
       <form id={formId} onSubmit={form.handleSubmit(onSubmit)}>
         <FormPanel
           disabled={true}
-          header={<p>Data API Settings</p>}
+          header={<p>API Settings</p>}
           footer={
             <div className="flex px-8 py-4">
               <FormActions
--- a/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
+++ b/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
@@ -22,6 +22,7 @@ import {
   Form_Shadcn_,
   Input_Shadcn_,
   Skeleton,
+  Switch,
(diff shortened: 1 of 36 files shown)


apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -98,163 +129,213 @@ const PostgrestConfig = () => {
...
+                      label="Max rows"
+                      description="The maximum number of rows returned from a view, table, or stored procedure. Limits payload size for accidental or malicious requests."
+                    >
+                      <FormControl_Shadcn_>
+                        <Input_Shadcn_
+                          size="small"
+                          disabled={!canUpdatePostgrestConfig}
+                          {...field}
+                          type="number"
+                          {...form.register('maxRows', {
```
MildTomato, line 296: wondering why this one has valueAsNumber: true and the one below doesn't. 
just going to check

saltcod:
looks good & works!  love the new multiselect 

This does suffer from a similar problem to before — you make the change at the top of this panel in the Multiselect, and the save/cancel buttons are way down below.  It isn't super obvious that something has changed and needs saving. 


screenshot-2024-05-28-at-21 55 46

we could move the actions to the top — only showing them when something's changed 
screenshot-2024-05-28-at-21 55 26

could also maybe do something like this — a tooltip trigger that says "Unsaved changes" or something 
screenshot-2024-05-28-at-22 00 00
.
Don't think this is a huge problem — but I bet this is why it was autosave in the first place

apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -41,7 +55,13 @@ const PostgrestConfig = () => {
...
   })
 
-  const { data: config, isError } = useProjectPostgrestConfigQuery({ projectRef })
+  const {
+    data: config,
+    isLoading: isLoadingPostgrestConfigQuery,
+    isError,
+  } = useProjectPostgrestConfigQuery({ projectRef })
+
+  console.log('config', config)
```
saltcod, line 64: ^

apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -98,163 +129,213 @@ const PostgrestConfig = () => {
...
+        dbSchema: config.db_schema ? config.db_schema.replace(/ /g, '').split(',') : [],
+        maxRows: config.max_rows,
+        dbExtraSearchPath: config.db_extra_search_path,
+        dbPool: config.db_pool,
+      })
+    }
+  }
 
-            resetForm({ values, initialValues: values })
+  useEffect(() => {
```
saltcod, line 143: can you use the isSuccess callback here like Ivan suggested for that other useEffect issue?

saltcod (approved):

joshenlim (changes requested):
there seems to be some height style issue going on in the settings/api page here
Notice the empty space at the bottom

It seems like the MultiSelector component is the one causing it actually - i tried commenting out line 192 specifically and its okay

image

MildTomato:
> there seems to be some height style issue going on in the settings/api page here
Notice the empty space at the bottom

I can't repro this. anything in particular you did to repro ?

apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
MildTomato: thanks! fixed

apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
MildTomato: Can we change this in a follow up PR. Didn't want to get involved in changing how the form resets in this one.

apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -98,163 +127,213 @@ const PostgrestConfig = () => {
...
+                          <MultiSelectorTrigger>
+                            <MultiSelectorInput placeholder="Select schemas for Data API..." />
+                          </MultiSelectorTrigger>
+                          <MultiSelectorContent>
+                            {schema.length <= 0 ? (
+                              <MultiSelectorList>
+                                <MultiSelectorItem key={'empty'} value={'no'}>
+                                  no
+                                </MultiSelectorItem>
+                              </MultiSelectorList>
```
MildTomato, line 205: 😅 ah, that needs to go

joshenlim: i'll remove this

joshenlim (approved):

PR #26803, opened 2024-05-27
Fix value set for storage upload unit

saltcod:
storage upload limit should default to 50 _MB_
on new projects it sometimes gets set as 50 _bytes_

screenshot-2024-05-27-at-15 34 05

screenshot-2024-05-27-at-15 41 28


apps/studio/components/to-be-cleaned/Storage/StorageSettings/StorageSettings.tsx
```diff
@@ -31,201 +45,183 @@ const StorageSettings = () => {
...
-      onSuccess: (res) => {
-        const updatedValue = convertFromBytes(res.fileSizeLimit)
-        initialValues = {
-          fileSizeLimit: updatedValue.value,
-          unformattedFileSizeLimit: res.fileSizeLimit,
-        }
-        toast.success('Successfully updated storage settings')
-      },
-    })
+  const [selectedUnit, _] = useState(unit)
```
joshenlim, line 48: im guessing we no longer need this since this field is part of the form?

joshenlim:
Storage settings seems to be getting initialized to 0 MB if I navigate from another page (e.g settings/infrastructure) to settings/storage

Does not have if I refresh while on the settings/storage page itself

joshenlim:
Bug on preview: The "equivalent in bytes" value is not getting updated when I change the units (KB, MB, GB)
image

joshenlim (approved):

PR #26831, opened 2024-05-28
Fix #26802: Implement client-side only JWT secret generation

### LINKED ISSUE #26802: JWT secret should be generated client-side only
If I understand the HTTP network stream correctly, the JWT secret that is automatically generated here

https://github.com/supabase/supabase/blob/5e7817548b7972a72370dfa70550eda69a6664a1/apps/docs/components/JwtGenerator/JwtGenerator.tsx#L28

e.g. in context of https://supabase.com/docs/guides/self-hosting/docker#generate-api-keys, is actually generated on the server (SSR). I would prefer the JWT secret to be generated and visible to my local browser only. I am not a React dev (yet) but it seems using `dynamic()` for the `JwtGenerator` should do the trick.

O-Aditya:

YES

This PR introduces a bug fix.

Currently, the JWT secret is generated on both the client and server sides. This behavior is problematic because it can lead to security issues. The related issue is [#26802](link-to-the-issue).

## What is the new behavior?

With the changes in this PR, the JWT secret is now generated on the client side only. This is achieved by using Next.js's dynamic imports to ensure that the `JwtGenerator` component is only rendered on the client side.

## Additional context

This change improves the security of the application by ensuring that JWT secrets are only generated and visible on the client side.


charislam:
Thanks @O-Aditya! This does get replaced upon hydration on the client-side with a browser-generated string, but it's better to be extra safe. Pushed a commit to fix the build and also took the opportunity to improve the random string generation.

apps/docs/components/JwtGenerator/JwtGenerator.tsx
```diff
@@ -24,8 +24,41 @@ const serviceToken = `
...
+   * uniform distribution, we have to discard everything above the last full
+   * cycle, and pick again.
+   *
+   * The minus 1 is to account for 0-indexing.
+   */
+  const MAX = Math.floor(256 / CHARS.length) * CHARS.length - 1
+
+  const randomUInt8Array = new Uint8Array(1)
+
+  for (let i = 0; i < length; i++) {
```
kangmingtay, line 44: @charislam did you mean this instead?
```suggestion
  for (let i = 0; i < CHARS.length; i++) {
```

charislam: @kangmingtay, no I'm pretty sure this should be `length` from the arguments, which controls the length of the generated string. Each run through the loop generates one character for the random string

O-Aditya:
Hello @charislam ,

I've noticed that my pull request hasn't been approved yet. I'm eager to make any necessary changes to get it merged. Could you please provide some guidance on what improvements I need to make or any issues that need to be addressed?

Thank you for your time and consideration.

Best,
[Aditya]

charislam:
Hey @O-Aditya, thanks for following up! I made some changes to fix small things I noticed while reviewing your PR, so I've asked someone else at Supabase to do a review as well (so I'm not approving my own code). Your changes are great, and we'll take it from here! You can expect this to be merged in the next few business days 😄

kangmingtay (approved):

apps/docs/components/JwtGenerator/JwtGenerator.tsx
kangmingtay: oops missed that, ok lgtm then!

PR #26836, opened 2024-05-28
Chore/custom html elements cleanup

fsansalvadore:

YES

Keep CustomHTMLElements in ui and remove duplicate from docs.


charislam:
I have the slight suspicion that these files might have diverged a bit... let me double check

charislam (approved):
Actually looks like the `ui` version is the more updated one, so should be good!

PR #26814, opened 2024-05-28
add await to this return function filterSerializedResponseHeaders

kRaviteja7:
return await resolve(event, {
    filterSerializedResponseHeaders(name) {
      /**
       * Supabase libraries use the `content-range` and `x-supabase-api-version`
       * headers, so we need to tell SvelteKit to pass it through. */ return name === 'content-range' || name === 'x-supabase-api-version' }, })

## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES/NO

## What kind of change does this PR introduce?

Bug fix, feature, docs update, ...

## What is the current behavior?

Please link any relevant issues here.

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context

Add any other context or screenshots.

PR #26848, opened 2024-05-29
26733: Add await to resolve() in documentation for sveltekit.

jonlivingstone:
This prevents a race condition in which headers are modified (setting cookies) after being sent.

YES

Documentation update

Sometimes this code causes a race condition in which the code attempts to modify the cookies after the headers have been sent.
```
  return resolve(event, {
    filterSerializedResponseHeaders(name) {
      return name === 'content-range' || name === 'x-supabase-api-version'
    },
  })
 ```

## What is the new behavior?

The new behavior awaits on the `resolve()` so cookies are set before the return of the local function, and consequently before headers are sent.

```
  return await resolve(event, {
   filterSerializedResponseHeaders(name) {
     return name === 'content-range' || name === 'x-supabase-api-version'
   },
 })
```

This seems to correlate the [sveltekit samples](https://kit.svelte.dev/docs/hooks) in which `resolve()` is always awaited.

### DIFF
--- a/apps/docs/content/guides/auth/auth-helpers/sveltekit.mdx
+++ b/apps/docs/content/guides/auth/auth-helpers/sveltekit.mdx
@@ -94,7 +94,7 @@ export const handle = async ({ event, resolve }) => {
     return { session, user }
   }
 
-  return resolve(event, {
+  return await resolve(event, {
     filterSerializedResponseHeaders(name) {
       return name === 'content-range' || name === 'x-supabase-api-version'
     },
@@ -141,7 +141,7 @@ export const handle: Handle = async ({ event, resolve }) => {
     return { session, user }
   }
 
-  return resolve(event, {
+  return await resolve(event, {
     filterSerializedResponseHeaders(name) {
       return name === 'content-range' || name === 'x-supabase-api-version'
     },
@@ -677,7 +677,7 @@ async function supabase({ event, resolve }) {
     return { session, user }
   }
 
-  return resolve(event, {
+  return await resolve(event, {
     filterSerializedResponseHeaders(name) {
       return name === 'content-range' || name === 'x-supabase-api-version'
     },
@@ -703,7 +703,7 @@ async function authorization({ event, resolve }) {
     }
   }
 
-  return resolve(event)
+  return await resolve(event)
 }
 
 export const handle = sequence(supabase, authorization)
@@ -745,7 +745,7 @@ async function supabase({ event, resolve }) {
     return { session, user }
   }
 
-  return resolve(event, {
+  return await resolve(event, {
     filterSerializedResponseHeaders(name) {
       return name === 'content-range' || name === 'x-supabase-api-version'
     },
@@ -771,7 +771,7 @@ async function authorization({ event, resolve }) {
     }
   }
 
-  return resolve(event)
+  return await resolve(event)
 }
 
 export const handle: Handle = sequence(supabase, authorization)
@@ -970,7 +970,7 @@ export const handle: Handle = async ({ event, resolve }) => {
     return { session, user }
   }
 
-  return resolve(event, {
+  return await resolve(event, {
     filterSerializedResponseHeaders(name) {
       return name === 'content-range' || name === 'x-supabase-api-version'
     },
--- a/apps/docs/content/guides/auth/server-side/creating-a-client.mdx
+++ b/apps/docs/content/guides/auth/server-side/creating-a-client.mdx
@@ -493,7 +493,7 @@ export const handle: Handle = async ({ event, resolve }) => {
     return { session, user }
   }
 
-  return resolve(event, {
+  return await resolve(event, {
     filterSerializedResponseHeaders(name) {
       return name === 'content-range' || name === 'x-supabase-api-version'
     },
--- a/apps/docs/content/guides/auth/server-side/sveltekit.mdx
+++ b/apps/docs/content/guides/auth/server-side/sveltekit.mdx
@@ -118,7 +118,7 @@ const supabase: Handle = async ({ event, resolve }) => {
     return { session, user }
   }
 
-  return resolve(event, {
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by encima** [comment]
Closes #26733

**Comment by jonlivingstone** [comment]
> Hmm I'm a bit curious why this is working for you, `return await` should be redundant outside of a `try/catch`, and I don't see one in this file...

I don't understand the fix fully myself.

I don't understand this statement about async and try/catch, maybe because I am not knowledgeable in TypeScript. As far as I am aware, exception management and asynchronous processing are two very separate topics, we may need to synchronize execution independently of any error management.

In the present case, the issue is that the moment in the page lifecycle where cookies are set is not deterministic; there is a race condition if I use the example as specified on my website, node crashes on 20% of the requests because `cookies.set()` is called after the headers are sent. It doesn't do it anymore if I add `await` before `resolve`. Besides the [samples from sveltekit](https://kit.svelte.dev/docs/hooks) all await on `revolve`.

The specific instance that causes me problems is at:
https://github.com/supabase/supabase/blob/master/apps/docs/content/guides/auth/server-side/sveltekit.mdx?plain=1#L88

That can be the wrong (or insufficient) fix to a true problem. I wasn't able to follow the code up to when `cookies.set()` is called. Maybe we miss an `await` there and that's why the call to `set` is not guaranteed to return in time.

This is a documentation change so I am not blocked on this. :)

**Comment by charislam** [comment]
What I mean is that `return await` is equivalent to `return`, with the exception of within a `try/catch` block, where `return await` will trigger the `catch` block on error and `return` will not. I mentioned the exception just to double-check that it doesn't apply in this case, so I would expect the code before and after the proposed fix to be functionally equivalent.

I looked at the SvelteKit examples and I don't think they need the `await` either. My assumption is they put it in for consistency because they sometimes assign the returned promise value to a variable and alter the variable before returning it. But here you can see an example where they don't use it:

```
If unimplemented, defaults to ({ event, resolve }) => resolve(event).
```

> cookies.set() is called after the headers are sent

This is a good clue and I think gives me a hint what is happening. I'm going to tag this issue with "needs-analysis" because I think we need a deeper dig into the root cause. Super appreciate you bringing this up so we can investigate!

**Comment by jonlivingstone** [comment]
Thanks, no worries. If that helps I can try to get a minimal repro.
It happened as I started to put the website in prod (obviously...) on node 22 and it received a bunch of unpredictable non-user traffic not related to any route I had.
Maybe all what the await does is to switch threads and that's enough to mask the issue but not a guaranteed 'fix'.
Thanks for your time, I appreciate.

**Comment by charislam** [comment]
If you could get a minimal repro that would be lovely ❤️

**Comment by jonlivingstone** [comment]
I could not get to a minimum repro. I wonder if the issue is that the main app was doing something that would send the headers without waiting for supabase auth to complete.

Probably not worth it to continue any further then. Thanks for having answered.

FWIW here are some of stack trace from the server when running on node:
```
error: Cannot use `cookies.set(...)` after the response has been generated
at /home/ec2-user/bf-platform/rpx-frontend/build/server/index.js:4119:15
at set (/home/ec2-user/bf-platform/rpx-frontend/build/server/chunks/hooks.server-D5CYnYXm.js:15:9)
at /home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/ssr/dist/index.js:304:23
at /home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/ssr/dist/index.js:302:41
at map (:1:21)
at /home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/ssr/dist/index.js:302:13
at setItem (/home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/ssr/dist/index.js:299:30)
at /home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/auth-js/dist/main/lib/helpers.js:130:11
at setItemAsync (/home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/auth-js/dist/main/lib/helpers.js:129:29)
at /home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/auth-js/dist/main/GoTrueClient.js:1533:19
```

```
Error: Cannot use `cookies.set(...)` after the response has been generated
at event2.cookies.set (file:///home/ec2-user/bf-platform/rpx-frontend/build/server/index.js:4119:15)
at Object.set (file:///home/ec2-user/bf-platform/rpx-frontend/build/server/chunks/hooks.server-DRaHdJsB.js:15:23)
at /home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/ssr/dist/index.js:304:31
at Array.map (<anonymous>)
at Object.setItem (/home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/ssr/dist/index.js:302:20)
at setItemAsync (/home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/auth-js/dist/main/lib/helpers.js:130:19)
at SupabaseAuthClient._saveSession (/home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/auth-js/dist/main/GoTrueClient.js:1533:42)
at SupabaseAuthClient._callRefreshToken (/home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/auth-js/dist/main/GoTrueClient.js:147>
at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
at async SupabaseAuthClient.__loadSession (/home/ec2-user/bf-platform/rpx-frontend/node_modules/@supabase/auth-js/dist/main/GoTrueClient.js:8>
```

j4w8n:
@charislam, someone else submitted a lighter version of this, and it's been approved; if you wanna rein that one in before it gets merged. https://github.com/supabase/supabase/pull/26814

charislam:
Thanks @jonlivingstone! I'm going to close this then, but feel free to reopen if you figure out the problem!

And thanks for the heads up @j4w8n, I'll head over and comment.

PR #26580, opened 2024-05-20
Error boundary handle removeChild error

joshenlim:
image

PR #26857, opened 2024-05-29
Chore/sql editor pagination of results

joshenlim:
Add a limit field for the SQL editor that will automatically limit the results (unless the query itself has a limit)
Number of results have been shifted below and will have a suffix (auto limit n rows) if the auto limit is kicked in

image

image


apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityActions.tsx
```diff
@@ -46,48 +63,70 @@ const UtilityActions = ({
...
+            <DropdownMenuItem
+              className="justify-between"
+              onClick={() => {
+                setIntellisenseEnabled(!intellisenseEnabled)
+                toast.success(
+                  `Successfully ${intellisenseEnabled ? 'disabled' : 'enabled'} intellisense. ${intellisenseEnabled ? 'Please refresh your browser for changes to take place.' : ''}`
+                )
+              }}
+            >
+              <p>Intellisense enabled</p>
```
MildTomato, line 88: remove p ?

apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityActions.tsx
```diff
@@ -46,48 +63,70 @@ const UtilityActions = ({
...
 
       <DropdownMenu>
         <DropdownMenuTrigger asChild>
-          <Button
-            type="text"
-            className="px-1"
-            icon={<Keyboard size={14} className="text-foreground-light" />}
-          />
+          <Button asChild type="outline" iconRight={<ChevronDown size={14} />}>
+            <span>{ROWS_PER_PAGE_OPTIONS.find((opt) => opt.value === snap.limit)?.label}</span>
```
MildTomato, line 112: span ?

apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityActions.tsx
```diff
@@ -46,48 +63,70 @@ const UtilityActions = ({
...
+      </div>
 
       <DropdownMenu>
         <DropdownMenuTrigger asChild>
-          <Button
-            type="text"
-            className="px-1"
-            icon={<Keyboard size={14} className="text-foreground-light" />}
-          />
+          <Button asChild type="outline" iconRight={<ChevronDown size={14} />}>
```
MildTomato, line 111: Could be `default` ? That outline button looks a little nasty

MildTomato: which is my fault..

apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityPanel.tsx
```diff
@@ -111,19 +113,16 @@ const UtilityPanel = ({
...
-              `(${result.rows.length.toLocaleString()})`}
+            <span className="translate-y-[1px]">Results</span>
           </TabsTrigger_Shadcn_>
           <TabsTrigger_Shadcn_ className="py-3 text-xs" value="chart">
-            Chart
+            <span className="translate-y-[1px]">Chart</span>
           </TabsTrigger_Shadcn_>
+          {result?.rows && <ResultsDropdown id={id} />}
         </div>
         <div className="flex gap-1 h-full items-center">
```
MildTomato, line 125: these classes might not do anything now

MildTomato (commented):
All comments, non blocking :D

apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
```diff
@@ -311,13 +311,30 @@ const SQLEditor = () => {
...
 
+        // Remove lines and whitespaces
+        const cleanedSql = sql.trim().replaceAll('\n', ' ').replaceAll(/\s+/g, ' ')
+        // Check if need to auto limit rows
+        const appendAutoLimit =
+          snap.limit > 0 &&
+          cleanedSql.toLowerCase().startsWith('select') &&
+          !cleanedSql.endsWith('limit') &&
+          !cleanedSql.endsWith('limit;') &&
+          !!!cleanedSql.match('limit [0-9]*[;]?$')
```
alaister, line 322: triple bang?!?!

apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityActions.tsx
joshenlim: its a workaround for the removeChild / insertBefore error that we see when users use Google Translate (or some third party extension) - although i'm genuinely not sure whether this is worth pursuing to keep doing as a practice tbh ([ref](https://github.com/supabase/supabase/pull/26580))

atm inclined to remove the span tbh

apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
joshenlim: hahaha im wondering what would be a cleaner way to do this 😅 
wanted to check for falsey value of that 

maybe `(!!cleanedSql.match(...) === false)` might be better you reckon? for readability

saltcod:
should switching this dropdown rerun the query? 

screenshot-2024-05-29-at-12 58 44

joshenlim:
> should switching this dropdown rerun the query?
hmm I don't think so - would rather the user manually run the query again (at least thats how TablePlus does it, so am inclined to follow it for now). Its more explicit too which I prefer, cause if say the user ran `select * from table` that returns 200k rows, and changes the value to no limit, the UI performance will degrade then

apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
joshenlim: i'll change it to just 1 hahaha

alaister (commented):
To throw another spanner in the works, something like `select generate_series(1,1000);;;` is actually a valid SQL query since it ignores any number of semicolons

joshenlim:
I'm gonna opt to compress all those actions into a dropdown when the screen is small and the AI assistant is open like this - 
- I'd rather keep all "actions" in one place instead of several
- Those actions that i'm compressing into a dropdown are all "secondary" imo, they don't need a spot up front in the UI if the real estate doesnt support it
Screenshot 2024-05-30 at 14 43 36

joshenlim:
> To throw another spanner in the works, something like select generate_series(1,1000);;; is actually a valid SQL query since it ignores any number of semicolons

okay added a fix and also unit tests on this piece of the logic - but just FYI atm i'm leaning towards the check for when to append the limit be more restrictive (i.e want to avoid adding the limit param too easily and end up causing a syntax error for the user instead). otherwise i think we'd need to install a full pg parser in the dashboard to properly do this, but I'm thinking what we have now fits most of the use case? wdyt though @alaister

apps/studio/components/interfaces/SQLEditor/SQLEditor.utils.ts
```diff
@@ -109,3 +109,43 @@ export const compareAsNewSnippet = (sqlDiff: ContentDiff) => {
...
+    !hasComments &&
+    !hasMultipleQueries &&
+    cleanedSql.toLowerCase().startsWith('select') &&
+    !cleanedSql.endsWith('limit') &&
+    !cleanedSql.endsWith('limit;') &&
+    !cleanedSql.match('limit [0-9]*[;]?$')
+  return { cleanedSql, appendAutoLimit }
+}
+
+export const prefixWithLimit = (sql: string, limit: number = 0) => {
```
alaister, line 143: very much a nit but shouldn't this be called `suffixWithLimit`? 😅

alaister (approved):
LGTM!

apps/studio/components/interfaces/SQLEditor/SQLEditor.utils.ts
joshenlim: oh yes hahahaha that is correct

PR #26903, opened 2024-05-31
feat: add IPv4 add on direct conn notice

w3b6x9:

feature

## What is the new behavior?

- Users with IPv4 add on:
  - will no longer get IPv4 warning message `Direct database access is only available via IPv6-compatible networks or if your project has the IPv4 add on.` since they are able to connect to database directly via IPv4.
  - add message to strongly recommend direct database connection only if user is connecting to pooler using session mode.

https://github.com/supabase/supabase/assets/5532241/7e713270-b5fa-4f63-abb0-61b2687007f0
 
 - Users without IPv4 add on:
   - continue see the message  `Direct database access is only available via IPv6-compatible networks`.

https://github.com/supabase/supabase/assets/5532241/cab42d9d-5d3d-4792-bf05-392688f449e3

- Projects that still have their `Pool Mode` set to `Session` will have a warning prompting them to switch to `Transaction` when they toggle their connection pooler credentials in the mode dropdown.

https://github.com/supabase/supabase/assets/5532241/127fa362-d54a-44f9-b0fb-a99eaefd8e84

- Remove `'statement'` mode from Supavisor-related types b/c it doesn't have it.

### DIFF
--- a/apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
+++ b/apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
@@ -235,9 +235,7 @@ export const DatabaseConnectionString = ({ appearance }: DatabaseConnectionStrin
                 poolingMode === 'session' &&
                 ipv4Addon !== undefined &&
                 snap.usePoolerConnection && <IPv4AddonDirectConnectionNotice />}
-              {isSuccessAddons
-                ? ipv4Addon === undefined && !snap.usePoolerConnection && <IPv4DeprecationNotice />
-                : !snap.usePoolerConnection && <IPv4DeprecationNotice />}
+              {ipv4Addon === undefined && !snap.usePoolerConnection && <IPv4DeprecationNotice />}
               <Input
                 copy
                 readOnly
--- a/apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
+++ b/apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
@@ -67,7 +67,7 @@ export const DatabaseConnectionString = ({ appearance }: DatabaseConnectionStrin
   const { project: projectDetails, isLoading: isProjectLoading } = useProjectContext()
 
   const connectionStringsRef = useRef<HTMLDivElement>(null)
-  const [poolingMode, setPoolingMode] = useState<'transaction' | 'session' | 'statement'>('session')
+  const [poolingMode, setPoolingMode] = useState<'transaction' | 'session'>('transaction')
   const [selectedTab, setSelectedTab] = useState<
     'uri' | 'psql' | 'golang' | 'jdbc' | 'dotnet' | 'nodejs' | 'php' | 'python'
   >('uri')
--- a/apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
+++ b/apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
@@ -2,6 +2,7 @@ import { useParams, useTelemetryProps } from 'common'
 import { useRouter } from 'next/router'
 import { useEffect, useRef, useState } from 'react'
 
+import { getAddons } from 'components/interfaces/Billing/Subscription/Subscription.utils'
 import { useProjectContext } from 'components/layouts/ProjectLayout/ProjectContext'
 import AlertError from 'components/ui/AlertError'
 import DatabaseSelector from 'components/ui/DatabaseSelector'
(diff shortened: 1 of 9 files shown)


apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
```diff
@@ -218,7 +225,13 @@ export const DatabaseConnectionString = ({ appearance }: DatabaseConnectionStrin
...
                 onSelectPoolingMode={setPoolingMode}
               />
-              {!snap.usePoolerConnection && <IPv4DeprecationNotice />}
+              {isSuccessAddons &&
+                poolingMode === 'session' &&
+                ipv4Addon !== undefined &&
+                snap.usePoolerConnection && <IPv4AddonDirectConnectionNotice />}
+              {isSuccessAddons
+                ? ipv4Addon === undefined && !snap.usePoolerConnection && <IPv4DeprecationNotice />
+                : !snap.usePoolerConnection && <IPv4DeprecationNotice />}
```
joshenlim, line 234: this looks like it can be just like this? since in both ways we're checking `!snap.usePoolerConnection`?

```suggestion
              {isSuccessAddons && !snap.usePoolerConnection && <IPv4DeprecationNotice />}
```

w3b6x9: so here i was trying to make sure that the user would see some notice in case the call to add ons failed.

If call to add ons succeed then i want to check `ipv4Addon === undefined` as part of it.

can simplify it to `ipv4Addon === undefined && !snap.usePoolerConnection && <IPv4DeprecationNotice />`, similar to what you suggested, if we don't care about `isSuccessAddons` when it's `false`.

w3b6x9: that's how i had it here: https://github.com/supabase/supabase/pull/26903/files#diff-0affa75a26bfa91b825973f555005deb19360c715009b7647ffb9370ac70ff7cR144

apps/studio/components/interfaces/Settings/Database/IPv4ConnectionNotices.tsx
```diff
@@ -37,3 +36,32 @@ export const IPv4DeprecationNotice = () => {
...
+
+export const IPv4AddonDirectConnectionNotice = () => {
+  return (
+    <Alert_Shadcn_ variant="default">
+      <AlertTitle_Shadcn_>
+        Direct database connections is recommended if you're connecting with session mode.
+      </AlertTitle_Shadcn_>
+      <AlertDescription_Shadcn_ className="space-y-3">
+        <p>
+          We strongly recommend you connect directly to your database instead of connection pooling
```
inian, line 48: If you are not connecting to the database from serverless environments, we strongly recommend you connect to your database directly instead of using connection pooling for lower latency.

Since your project has the IPv4 add on, the connection pooler isn't required to connect to the database from IPv4 only environments. 

(wanted to add a note about serverless environments)

packages/api-types/types/api.d.ts
```diff
@@ -4449,17 +4449,17 @@ export interface components {
       default_pool_size: number | null
       max_client_conn: number | null
       /** @enum {string|null} */
-      pool_mode: 'transaction' | 'session' | 'statement' | null
```
w3b6x9, line 4452: I don't think `null` is possible for `pool_mode`.

joshenlim: btw we shouldn't update this file - this is auto generated from the API, which means this needs to be fixed on the API instead

w3b6x9: Gotcha, thanks

joshenlim: ahh sorry - i assumed you did this manually 🤦 just to check if you ran npm run codegen in packages/api-types?

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
joshenlim: ahh okay i see what you mean here

packages/api-types/types/api.d.ts
w3b6x9: i didn't, wasn't sure how it worked.

w3b6x9: thanks for doing it! now i know

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
joshenlim: ^ yeah i think we don't need to check for isSuccessAddons here, just `ipv4Addon === undefined && !snap.usePoolerConnection && <IPv4DeprecationNotice />` as you mentioned will work - i'll make this change

joshenlim (approved):

apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseConnectionString.tsx
w3b6x9: Sounds good thanks

PR #26912, opened 2024-05-31
Fix icon size in storage settings

joshenlim:


alaister (approved):

MildTomato:
ha i saw this before and was wondering why it was only happening in my branch

PR #26992, opened 2024-06-04
Update supabase_js_v2.yml

f00f:

YES

docs update

Incorrect response.
The row with the id 1 has (name == "null"). It has not (name is null)

## What is the new behavior?

Correct response.
.is('column', null) tests for NULL values, not for strings that are == "null".

## Additional context

Just tested it in my supabase instance.


charislam (approved):
Thanks @f00f!

PR #27016, opened 2024-06-04
feat(docs): track last-changed date

charislam:
Add a table to track the last-changed date for docs content, and a script to seed the database with last-changed dates based on Git commit dates.

After the schema changes are in prod, I'll run the script to seed the table, and make a followup PR to update the timestamps when PRs are merged into master.

The point of doing this rather than just relying on Git commits is that the timestamp updates on merge will check section checksums, to avoid prematurely updating timestamps for _sections_ that were not edited even though the _parent file_ was updated. (In other words, we might have a very old section in a file where some other section was very recently edited, and we want to be able to detect the very old section.)

### DIFF
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -97,6 +97,7 @@
     "jest-environment-jsdom": "^29.7.0",
     "npm-run-all": "^4.1.5",
     "openapi-types": "^12.1.3",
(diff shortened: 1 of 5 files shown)


fsansalvadore (approved):
Magician ✨

PR #27019, opened 2024-06-05
feat(command menu v2): complete state management

charislam:
Add all the outstanding pieces for state management of new command menu:

- Pages state (handles routing between command menu views, e.g., docs search, AI, etc.)
- View state (handles opening and closing and initiation (needed for lazy loading))
- Query state (handles active search term)


MildTomato:
Should we get an example / demo working in design system ?
Not really sure how this works and the DX for us

MildTomato (approved):

PR #27038, opened 2024-06-05
feat: update last_changed table daily

charislam:
Keeps track of fine-grained (per section) edit times for docs content.

Once daily, a GitHub Action runs that:

- Checks whether content hashes have changed for each section
- Updates the table that tracks content edit times if the hashes have changed

**Note**: The cron job isn't scheduled yet. I'll run the Action manually a few times to validate it, then turn it on in another PR.

## Testing

You'll have to test locally because that's the only way to have the right env variables:

1. Reset your local database: `npx supabase db reset --local`. Make sure your `apps/docs/.env` file has the right values for the Supabase URL and service role key.
2. Seed the `last_changed` table by running the script: `npm --prefix=apps/docs run last-changed`
3. Check localhost:54321 to make sure the table is populated.
4. Open a random content MDX file and change a random heading.
5. Run the script again: `npm --prefix=apps/docs run last-changed`. The number of updated sections should be the same as for the last run, but the number of removed sections should be 1 (the row with the old heading should be removed).
6. Check localhost:54321 to see that the row with the old heading is gone and the row with the new heading is added.
7. Check to see that timestamps are updated: `select distinct last_checked from last_changed`

### DIFF
--- a/.github/workflows/docs-last-changed.yml
+++ b/.github/workflows/docs-last-changed.yml
@@ -3,11 +3,8 @@ name: [Docs] Update last-changed dates
 on:
   # Temporarily turned off until the action is validated manually
   #
-  # push:
-  #   branches:
-  #     - master
-  #   paths:
-  #     - 'apps/docs/content/**'
+  # schedule:
+  #   - cron: '0 0 * * *'
   workflow_dispatch:
     inputs:
       reset:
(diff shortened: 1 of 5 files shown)


gregnr (approved):
Nice one!

PR #27044, opened 2024-06-05
feat(cmd menu): add hooks

charislam:
Hook the new Command Menu state up to React.  Add tests for hooks.


apps/studio/components/interfaces/Support/SupportForm.tsx
```diff
@@ -70,7 +71,12 @@ export interface SupportFormProps {
 }
 
 const SupportForm = ({ setSentCategory, setSelectedProject }: SupportFormProps) => {
-  const { handleDocsSearchDebounced, searchState, searchState: state } = useDocsSearch()
+  const supabaseClient = useSupabaseClient()
```
charislam, line 74: Injecting the client as a dependency instead of baking it into the hook because I want to minimize imports in `common` (where the hook is).

We didn't have the right package declarations for this before (but everything still worked fine because all the apps installed `@supabase/auth-helpers-react` for other reasons anyway). But now with the new tests it breaks testing if we don't fix it

packages/common/hooks/useDocsSearch.ts
```diff
@@ -199,24 +198,26 @@ const useDocsSearch = () => {
...
-            type: 'resultsReturned',
-            key: localKey,
-            sourcesLoaded,
-            results: data,
-          })
-        }
-      })
+      supabaseClient
+        .rpc('docs_search_fts', { query: query.trim() })
+        .then(({ data, error }: { data: any; error: any }) => {
```
charislam, line 203: Only change is the types here. These used to be implicit `any` anyway, but tests will fail with TS complaint if we don't define it explicitly now.

apps/studio/components/interfaces/Support/SupportForm.tsx
MildTomato: this is worrying me a bit as `@supabase/auth-helpers-react` is deprecated

charislam: yeah all the apps are behind on this one still... 😢 we should probably have a plan to move off this (but not in this PR)

packages/ui-patterns/CommandMenu/api/hooks/commandsHooks.test.tsx
```diff
@@ -0,0 +1,254 @@
+/**
```
saltcod, line 1: look at these fancy tests!!

saltcod (approved):

PR #27048, opened 2024-06-06
chore: adjust project transfer texts

kevcodez:

With recent changes to architecture, the previous texts are no longer accurate for all cases. Adjusted the texts and cleaned up the modal to concentrate on the relevant bits.

This is prep work for some API changes that will allow project transfers between paid and free plans with addons (will be auto-removed)


alaister (approved):
LGTM!

PR #27073, opened 2024-06-06
fix(docs): fix broken link

charislam:


carusocv (approved):
LGTM 

(But also I'm not an eng here so probably get another 🤷🏼 )

PR #27077, opened 2024-06-06
fix: revoke role on public for hooks

J0:

YES

Currently , we ask for users to create a function in the schema and then revoke access from `anon` and `authenticated` so that only `supabase_auth_admin` can access the function. However, since anon and authenticated inherit public's execute permission we need to revoke access from `public` too to prevent `anon` and `authenticated` from inheriting access.

We revoke `public` from using  the function as it is not sufficient to revoke `anon` and `authenticated` to prevent `public` from running the function.

Search was done via: `rg 'revoke' -B 4 -A 4`


apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -77,12 +77,12 @@ You also need to grant usage to `supabase_auth_admin`:
 grant usage on schema public to supabase_auth_admin;
 ```
 
-Also revoke permissions from the `authenticated` and `anon` roles to ensure the function is not accessible by Supabase Serverless APIs.
+Also revoke permissions from the `authenticated`, `public`, and `anon` roles to ensure the function is not accessible by Supabase Serverless APIs. We revoke access on the `public` role as it has access to functions created on `public.*` by default and `anon` and `authenticated` inherit permissions from the `public` role. Consequently, we need to revoke permissions from `public` in order to prevent the `anon` and `authenticated` roles from inheriting permissions to invoke the Postgres Hook.
```
J0, line 80:
```suggestion
Also revoke permissions from the `authenticated`, `public`, and `anon` roles to ensure the function is not accessible by Supabase Serverless APIs. The `public` role has  access to functions created on `public.*` by default and `anon` and `authenticated` inherit permissions from the `public` role.  Permission is revoked from the`public` role In order to prevent the `anon` and `authenticated` roles from inheriting permissions to invoke the Postgres Hook.
```

apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -602,15 +602,15 @@ grant execute
...
+  from authenticated, anon, public;
 
 grant all
   on table public.profiles
   to supabase_auth_admin;
 
 revoke all
   on table public.profiles
-  from authenticated, anon;
+  from authenticated, anon, public;
```
J0, line 613: There is no need to revoke from the tables on `public`. However we do this for consistency with function revocation.

charislam (approved):

PR #27097, opened 2024-06-07
Set the primary key as the default sort if no other sort exists

saltcod:
Prevent table editor jumping around when editing by setting the primary key to be the default sort

https://www.notion.so/supabase/Table-Editor-jumpy-editing-experience-insanity-45bea4c2823948659e08ceee678659e5?pvs=4


apps/studio/data/table-rows/table-rows-query.ts
```diff
@@ -138,9 +154,19 @@ export const getTableRowsSqlQuery = ({
...
       queryChains = queryChains.filter(x.column, x.operator, value)
     })
-  sorts.forEach((x) => {
-    queryChains = queryChains.order(x.table, x.column, x.ascending, x.nullsFirst)
-  })
+
+  // If sorts is empty, use the primary key as the default sort
+  if (sorts.length === 0) {
+    const primaryKey = getPrimaryKeyColumn(table)
+    console.log({ primaryKey })
```
MildTomato, line 161: random console log

apps/studio/data/table-rows/table-rows-query.ts
```diff
@@ -34,6 +34,13 @@ type GetTableRowsArgs = {
 // [Joshen] We can probably make this reasonably high, but for now max aim to load 10kb
 export const MAX_CHARACTERS = 10 * KB
 
+// return the primary key column if exists, otherwise return the first column
+// to use as a default sort
+const getPrimaryKeyColumn = (table: SupaTable) => {
```
MildTomato, line 39: maybe better named `defaultSortingColumn` since it might not be a PK

MildTomato: `getDefaultOrderByColumn` or maybe better to use pg terms

joshenlim (approved):
LGTM - tested
- Default sort works when no sort applied
- Updating a row when no sort applied doesnt cause data to shift
- Applying a sort will override default sort

PR #27172, opened 2024-06-11
fix: remove warning on branching that services are not available

J0:

YES

Enforces the distinction between the capabilities of Fly Postgres and the capabilities of branching. We also convert the inactivity duration to a constant in config so it is easier to keep in sync.

Removes the callout mentioning that branches do not have all features due to use of Fly Postgres. This cuts down on confusion and allows us to roll out support for different features for both of those at different points in time, rather than having them unified

What it looks like post change:

CleanShot 2024-06-11 at 11 57 51@2x

Relevant Internal Discussion: https://supabase.slack.com/archives/C01D6TWFFFW/p1718096770383279

### DIFF
--- a/apps/docs/content/guides/platform/branching.mdx
+++ b/apps/docs/content/guides/platform/branching.mdx
@@ -22,7 +22,7 @@ If you understand Git, you already understand Supabase Branching.
 
 Supabase Branching works with Git. You can test changes in a separate, temporary environment without affecting your production setup. When you're ready to ship your changes, merge your branch to update your production instance with the new changes.
 
-You can run multiple Preview Branches for every Supabase project. Branches contain all the Supabase features with their own API credentials. Preview Environments pause automatically after <SharedData data="config">branching.inactivity_period_in_minutes</SharedData> minutes of inactivity.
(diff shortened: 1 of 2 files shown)


apps/docs/content/guides/platform/branching.mdx
```diff
@@ -22,15 +22,7 @@ If you understand Git, you already understand Supabase Branching.
...
-You can run multiple Preview Branches for every Supabase project. Branches contain all the Supabase features with their own API credentials. Preview Environments pause automatically after 5 minutes of inactivity.
-
-<Admonition type="caution" label="Some features may not work as expected on Preview Branches">
-
-Due to Preview Branches running on Fly Postgres instances, some features like `pg-cron` are not fully supported.
-
-For a full list of limitations, see [Fly Postgres limitations](/docs/guides/platform/fly-postgres#limitations).
-
-</Admonition>
+You can run multiple Preview Branches for every Supabase project. Branches contain all the Supabase features with their own API credentials. Preview Environments pause automatically after <SharedData data="config">branching.inactivity_period_in_minutes</SharedData> minutes of inactivity.
```
pcnc, line 25: Might be worth mentioning that `pg_cron` executions will also be impacted by inactivity-related pausing

alaister (approved):

PR #27176, opened 2024-06-11
fix: Make `studio` work without importing the `storybook` dependency

ivasilov:
This PR reverts the https://github.com/supabase/supabase/pull/27174 and fixes the bug properly. 

The bug (`process.env` was undefined so all environment variables couldn't load in `studio`) was caused by the flag  `fallbackNodePolyfills`. The flag was added so that the frontend bundle would be smaller because it disables all Node libraries (including `process.env`). 

The functionality worked before today because we had the `node-process` dependency included by accident. Once we removed `storybook` (and its transitive dependency `process`) it stopped working.


alaister (approved):
LGTM

PR #27179, opened 2024-06-11
Chore/fix limit in sql query

saltcod:
fetch first was conflicting with our auto-appended limit

**To test:** 
this should fail before and pass after `select * from countries fetch first 5 rows only` 

 screenshot-2024-06-11-at-10 19 20


joshenlim:
lgtm! tested SQL editor with various syntax for the "first first" command - limit is not appended correctly

joshenlim (approved):

PR #27195, opened 2024-06-12
Update humans.txt

saltcod:
add shadcn to humans.txt


charislam (approved):
I wasn't aware we had this section and at first I was like "you're adding WHO"? 😺

saltcod:
Checked docs and www before merging.

PR #27215, opened 2024-06-12
fix: Make the onCopyUrl work on Safari

ivasilov:


saltcod (approved):

joshenlim (approved):
LGTM! Tested on preview
- copy URL with expiry on storage explorer works in safari
- copy URL for other parts of the dashboard using copyToClipboard still works (they actually work on my safari without this change for some reason, but they still do with this change)
  - e.g Table editor copy cell content

PR #27221, opened 2024-06-12
refactor(docs): migrate home page to app router

charislam:
Migrate docs home page to App Router.

Biggest changes touching shared packages are:
- Need to put `use client` on a bunch of stuff
- Need to switch out for `next/compat/router` (this prevents straight-out erroring if Pages Router isn't mounted, but doesn't provide a replacement for the actual functionality)

- [x] Checked to make sure I haven't broken meta/OpenGraph
- [x] Checked responsive styles
- [x] Checked build size -- minor increase expected, nothing awful
- [x] Checked load time navigating between routers -- again minor increase and slight flicker expected, nothing awful
- [x] Checked for console errors -- only ones are 404s on the RSC prefetches, which are expected


apps/docs/app/layout.tsx
```diff
@@ -0,0 +1,81 @@
...
+  },
+}
+
+const viewport: Viewport = {
+  themeColor: '#1E1E1E',
+}
+
+const RootLayout = ({ children }: { children: React.ReactNode }) => {
+  return (
+    <ShortcutPreviewBuild>
```
fsansalvadore, line 54: @charislam what do you say if we abstract some or all of the providers into one component so we don't need to duplicate lots of code between `pages/_app.tsx` and `app/layout.tsx`? Could be in another PR but I think it's worth doing this upfront, before risking of building on top of the duplications.

fsansalvadore: Something like 

```tsx
 <Providers>
   <body>{children}</body>
 </Providers>
```

apps/docs/app/page.tsx
```diff
@@ -0,0 +1,342 @@
...
+const HomePage = () => (
+  <HomeLayout>
+    <div className="flex flex-col">
+      <h2 id="products">Products</h2>
+      <ul className="grid grid-cols-12 gap-6 not-prose [&_svg]:text-brand-600">
+        {products.map((product) => {
+          return (
+            <li key={product.title} className="col-span-12 md:col-span-4">
+              <Link href={product.href} passHref>
+                <GlassPanelWithIconPicker {...product}>
```
fsansalvadore, line 213: This results in a slightly different design.
But I think it looks good 😎

apps/docs/app/layout.tsx
fsansalvadore: This would also help to make the entry files more readable and reduce their scope.

charislam: Done!

fsansalvadore (approved):
Nice. Checked all previews and everything looks good 👍

charislam:
Checked all previews. Also tested log explorer on studio to make sure URL search params still work properly there

PR #27225, opened 2024-06-12
fix: quick fix for changelog

charislam:
quick and dirty (non-ideal) by just ignoring all errors when parsing markdown from github discussions

### DIFF
--- a/apps/www/pages/changelog.tsx
+++ b/apps/www/pages/changelog.tsx
@@ -88,9 +88,9 @@ export const getServerSideProps: GetServerSideProps = async ({ res, query }) => 
     28108378,
   ]
 
-  const releases = await (
-    await fetchGitHubReleases()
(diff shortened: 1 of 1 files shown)


apps/www/pages/changelog.tsx
```diff
@@ -165,43 +165,51 @@ export const getServerSideProps: GetServerSideProps = async ({ res, query }) =>
...
+
+        return {
+          ...item,
+          source: discussionsMdxSource,
+          type: 'discussion',
+          created_at,
+          url: item.url,
+        }
+      } catch (err) {
+        console.error(`Problem processing discussion MDX: ${err}`)
```
kevcodez, line 186: We should probably get this into Sentry and catching+console logging would make this a bit invisible

You could explicitly do `Sentry.captureException` - not sure about best practice here though

kevcodez (approved):
Approved to unblock and ensure the page can be opened again

PR #27242, opened 2024-06-13
docs: rewrite ssr examples to use `getAll`, `setAll`

hf:
Rewrites the SSR examples to use the [new `getAll`, `setAll`](https://github.com/orgs/supabase/discussions/27037) functions from the upcoming `@supabase/ssr@^0.4.0` release.

### DIFF
--- a/apps/docs/content/guides/auth/server-side/creating-a-client.mdx
+++ b/apps/docs/content/guides/auth/server-side/creating-a-client.mdx
@@ -137,7 +137,7 @@ Server Components only have read access to cookies. Check the `Middleware` tab f
 </Admonition>
 
 ```ts page.tsx
-import { createServerClient, type CookieOptions } from '@supabase/ssr'
+import { createServerClient } from '@supabase/ssr'
 import { cookies } from 'next/headers'
 
 export default async function Page () {
@@ -148,8 +148,8 @@ export default async function Page () {
     process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY!,
     {
       cookies: {
-        get(name: string) {
-          return cookieStore.get(name)?.value
+        getAll() {
+          return cookieStore.getAll()
         },
       },
     }
@@ -174,10 +174,8 @@ import { createServerClient, type CookieOptions } from '@supabase/ssr'
 import { NextResponse, type NextRequest } from 'next/server'
 
 export async function middleware(request: NextRequest) {
-  let response = NextResponse.next({
-    request: {
-      headers: request.headers,
-    },
+  let supabaseResponse = NextResponse.next({
+    request,
   })
 
   const supabase = createServerClient(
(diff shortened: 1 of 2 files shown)


apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -732,19 +719,19 @@ import { type LoaderFunctionArgs } from '@remix-run/node'
...
 
   const supabase = createServerClient(process.env.SUPABASE_URL!, process.env.SUPABASE_ANON_KEY!, {
     cookies: {
-      get(key) {
-        return cookies[key]
-      },
-      set(key, value, options) {
-        headers.append('Set-Cookie', serialize(key, value, options))
+      getAll() {
+        return request.headers.get('Cookie')?
```
hf, line 727: Possibly use `parse` here.

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -764,19 +751,19 @@ import { type ActionFunctionArgs } from '@remix-run/node'
...
```
hf, line 759: Possibly use `parse` here.

apps/docs/content/guides/auth/server-side/nextjs.mdx
```diff
@@ -626,25 +608,15 @@ Create a `utils/supabase` folder with a file for each type of client. Then copy
...
-import { createServerClient, type CookieOptions, serialize } from '@supabase/ssr'
+import { createServerClient, serialize } from '@supabase/ssr'
 import { type GetServerSidePropsContext } from 'next'
 
 export function createClient(context: GetServerSidePropsContext) {
   const supabase = createServerClient(
     process.env.NEXT_PUBLIC_SUPABASE_URL!,
     process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY!,
     {
-      cookies: {
```
hf, line 637: ?

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -174,61 +174,58 @@ import { createServerClient, type CookieOptions } from '@supabase/ssr'
...
-            },
-          })
-          response.cookies.set({
-            name,
-            value: '',
-            ...options,
+        setAll(cookiesToSet) {
+          cookiesToSet.forEach(({ name, value, options }) =>
+            request.cookies.set(name, value, options)
+          )
```
vhpx, line 192: doesn't seem like Next.js middleware's NextRequest allow you to set the options along with each cookie. My workaround was to temporarily remove the `options` argument:
```tsx
cookiesToSet.forEach(({ name, value }) =>
    request.cookies.set(name, value)
);
```

hf: Ah, you're right yes. Will fix.

J0 (approved):

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -174,61 +174,56 @@ import { createServerClient, type CookieOptions } from '@supabase/ssr'
...
+  // 1. Pass the request in it, like so:
+  //    const myNewResponse = NextResponse.next({ request })
+  // 2. Copy over the cookies, like so:
+  //    myNewResponse.cookies.setAll(supabaseResponse.cookies.getAll())
+  // 3. Change the myNewResponse object to fit your needs, but avoid changing
+  //    the cookies!
+  // 4. Finally:
+  //    return myNewResponse
+  // If this is not done, you may be causing the browser and server to go out
+  // of sync and terminate the user's session prematurely!
```
J0, line 224: Unrelated to the PR but to confirm my understanding - do we include the config below in our example so that folks don't forget to do so when they copy?

apps/docs/content/guides/auth/server-side/nextjs.mdx
```diff
@@ -626,25 +608,15 @@ Create a `utils/supabase` folder with a file for each type of client. Then copy
...
-        get(name: string) {
```
charislam, line 638: @hf, why were these cookie methods removed? getServerSideProps can access cookies so shouldn't cookie methods be provided here in case the dev wants to check auth, get user info, etc.?

https://nextjs.org/docs/pages/api-reference/functions/get-server-side-props#context-parameter

charislam (approved):

apps/docs/content/guides/auth/server-side/nextjs.mdx
hf: They're not removed, they're replaced with the `getAll` and `setAll` equivalents, because those are the new API.

charislam: But it looks like they're just replaced with an empty object? See line 617:

```
cookies: {},
```

Apologies if I'm missing something here

PR #27245, opened 2024-06-13
chore: Address the `Failed to construct 'URL': Invalid URL` error report in Sentry

ivasilov:
If the error is `Failed to construct 'URL': Invalid URL`, send it only 1% of the cases it happened.


saltcod (approved):

saltcod:
checked staging and www

PR #27247, opened 2024-06-13
Account for br and hr tags

saltcod:
Tags need to be self-closing. We're handling  already, but not <br> and <hr>. There's others as well, but these are the most likely to be included in Changelog posts.


ivasilov (approved):

saltcod:
checked www

MildTomato (approved):

MildTomato:
Approved with no preview check

PR #27256, opened 2024-06-13
fix(docs): clarify function revokes

charislam:


mansueli:
I think that we should also document how to revoke from a single specific function as that might also be what the user is looking for. Sometimes they have just one or two functions that need extra protection.

mansueli (approved):

charislam:
Checked www and docs

PR #27275, opened 2024-06-14
feat: Unify ESLint configs for Next.js apps

ivasilov:
This PR adds a new package with ESLint configuration for the Nextjs apps. 

There were several instances of the `eslint/next` package (which contains lint rules), they're now unified.

It also adds eslint to the `www` app.


alaister (approved):
LGTM!

- tested all available previews

Issue #27312, opened 2024-06-17
Self-host: Storage API is not working (Docker container unhealthy)


RoadToDev101:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I self-host Supabase with Docker on my machine as per the instructions on [Supabase Docs](https://supabase.com/docs/guides/self-hosting/docker). After successfully composing, the containers `supabase/storage-api:v1.0.6` and `supabase/studio` are in an unhealthy status. I can create buckets in the dashboard but fail to create folders in the bucket or upload files.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Follow the instructions on the Supabase self-hosting Docker guide.
2. Run docker-compose up.
3. Check the status of the containers supabase/storage-api and supabase/studio.
4. Try creating a bucket, then attempt to create a folder in the bucket or upload a file.

## Expected behavior

The `supabase/storage-api` and `supabase/studio` containers should be in a healthy status, allowing the creation of folders and uploading of files within buckets.

## Screenshots

![image](https://github.com/supabase/supabase/assets/115443243/2860a31f-7fd2-482e-9c6b-9f7e1378c947)

## System information

- Operating System: Ubuntu 22.04.4 LTS              
- Kernel: Linux 5.15.0-107-generic
- Browser (if applies) Chrome
- Version of Node.js: v20.11.1

![image](https://github.com/supabase/supabase/assets/115443243/63b1a66b-391f-4c4c-8825-7fd2e931571c)

encima:
Thanks for opening @RoadToDev101 
This is a bug in the self-hosted docker-compose as the storage api seems not to listen on `localhost`, so it reports as unhealthy when it is fine.
This can be confirmed with `docker logs $STORAGE_CONTAINER_ID`

We will make a PR to address this. That said, the issue of not being able to create folders cannot be reproduced, can you provide logs from the storage container, please?

RoadToDev101:
Hi @encima,

This is the log I get from the storage container when render the storage page on dashboard and create a new bucket name 'design_views'

`{"level":30,"time":"2024-06-18T01:26:18.129Z","pid":1,"hostname":"a122cace4088","msg":"Server listening at http://0.0.0.0:5000"}
[...]

PR #27316, opened 2024-06-17
Update healthchecks for all containers

encima:

YES

Fix/Feature

Current healthchecks are run through localhost or not using docker's networking which can result in failing checks

Studio, in particular, fails healthchecks by default as it does not expose the running app over localhost. 

A few issues exist regarding this, one recent example is https://github.com/supabase/supabase/issues/27312

## What is the new behavior?

This PR adds a healthcheck that retrieves the IP of the container (as `curl` and `wget` are not available so the container name cannot be used).

For all other containers, their internal name is used for Docker to resolve their respective addresses.

## Additional context

The current image for `studio` would need to be updated as this introduces a `docker-healthcheck.js` file that will only be included in subsequent builds.


cmantsch:
Was about to also create a PR, I guess you were faster 😄 

Main issue seemed to be the localhost, the service does not bind to 127.0.0.1 (however for storage it fails on linux amd64, but works fine on apple arm64).

The healthcheck for studio can also be fixed without the `healthcheck.js` file by adding quotes around the JS command (needs to be escaped):

```yaml
healthcheck:
      test:
        [
          "CMD",
          "node",
          "-e",
          "\"require('http').get('http://studio:3000/api/profile', (r) => {if (r.statusCode !== 200) throw new Error(r.statusCode)})\""
        ]
      timeout: 5s
      interval: 5s
      retries: 3
```

encima:
@cmantsch nice! that seems better to incorporate

Hallidayo:
Hi @encima - We’re doing a bit of housekeeping and plan to close out PRs that have been inactive for more than 3 months.

PR #27317, opened 2024-06-17
feat: Add eslint/turbo rule

ivasilov:
This PR adds an `eslint/turbo` rule which will report if some of the env vars in the apps are missing from the `turbo.json` config.


alaister (approved):
LGTM!

- Checked all available previews

PR #27288, opened 2024-06-15
Feat: add critical warning modal for removing members

monicakh:
## What kind of change does this PR introduce?

- Added a modal that displays a critical warning when removing a member from the organization Team page. 

image

PR #27336, opened 2024-06-18
Warn that deleting a user will also delete their snippets

saltcod:
Deleting a user will also delete any SQL snippets they have. These could be important. We should warn admins before taking this action. 

screenshot-2024-06-17-at-22 13 41


MildTomato:
maybe duplicate of https://github.com/supabase/supabase/pull/27288

PR #27364, opened 2024-06-18
Chore/allow index advisor for postgrest queries

saltcod:
To check, create a Query Performance entry:

1. Create the countries table in your test project using the SQL Editor Quickstart
2. go to the table's API side panel and grab code to do a curl https://share.cleanshot.com/NYsmXBbB
3. swap your actual anon key for the `SUPABASE_CLIENT_ANON_KEY`
4. run this curl in your terminal
5. this will give you an entry in the [Query Performance log](https://supabase.com/dashboard/project/_/database/query-performance) that you can search for
6. you should see a `with pgrst_source as` query  https://share.cleanshot.com/mfbg2gKM

this query should now have an index suggestions tab


monicakh (approved):
LGTM! Tested it on Studio.

PR #27376, opened 2024-06-19
fix: Minor fixes for turborepo in Vercel and Docker

ivasilov:


alaister (approved):
LGTM!

- still checked studio preview

PR #27400, opened 2024-06-20
Update connecting-to-postgres.mdx to mention clients

rogersm-uwosh:
It seems (I am not an expert by any means) that beyond the 3 strategies described here, you can also connect to Supabase using the clients -- but that doesn't get mentioned here. For those just learning Supabase, like me, I feel like this would make the documentation clearer.

I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

Update to the docs, to indicate that a 4th way of connecting to Supabase databases is via client software

What is the current behavior?

The connection strategies don't mention clients

What is the new behavior?

It mentions that you can connect to Supabase using clients, which I think is hugely important.

Additional context


charislam (approved):
Thanks @rogersm-uwosh!

PR #27402, opened 2024-06-20
feat: add rate limit otp

J0:
Exposes the rate limit var for GOTRUE_RATE_LIMIT_OTP which allows developers to impose per IP Rate limits on OTP sending.

- Depends on: https://github.com/supabase/infrastructure/pull/18616 so that needs to go in

- Also need to sanity check the copy after infra is merged


apps/studio/components/interfaces/Auth/RateLimits/RateLimits.tsx
```diff
@@ -394,6 +401,35 @@ const RateLimits = () => {
                   />
                 </FormSectionContent>
               </FormSection>
+              <FormSection
+                id="otp"
+                header={
+                  <FormSectionLabel
+                    description={
+                      <p className="text-foreground-light text-sm">
+                        Number of OTP sign-ins and sign up request that can be made per hour per IP
```
J0, line 410:
```suggestion
                        Number of OTP sign-ins and sign up requests that can be made per hour for a given IP
```

kangmingtay:
Can we name it rate limit for sign ups and sign-ins so it's clearer since it applies to both email links and OTPs

apps/studio/components/interfaces/Auth/RateLimits/RateLimits.tsx
```diff
@@ -394,6 +401,35 @@ const RateLimits = () => {
...
+                        address.
+                      </p>
+                    }
+                  >
+                    Rate limit for OTP sign ins
```
J0, line 415:
```suggestion
                    Rate limit for OTP sign up and sign ins
```

kangmingtay:
@J0 Removed the OTP wording since it introduces some confusion when thinking about how it works with auto-confirmed emails / phone numbers - this rate limit essentially caps the number of signups and sign-ins so i've changed the wording to capture that

kangmingtay:
I've tested with a local project and can confirm that the rate limit changes are applied successfully

J0:
Thanks!

ivasilov (approved):

PR #27412, opened 2024-06-20
ref: Do not disable branching when removing GitHub connection

kamilogorek:
We don't want to do this automatically for people, as we might accidentally remove branches that they wanted to keep, and removing branches is not reverseable.


alaister (approved):
LGTM!

- Tested on preview:
Screenshot 2024-06-20 at 20 43 29

PR #27440, opened 2024-06-21
fix: Updating EU tax ids

kevcodez:
Previously, some EU tax ids couldn't be set. To reproduce, go to prod and try adding a greek tax id.

Changes
- Fix for setting EU tax ids
- Remove sanitization, people should just enter their full tax id, sanitization rather leads to more problems


ecktoteckto:
There's the following issue now when adding Tax-ID for Austria:
1. Select "Austria AT VAT" from dropdown
2. Type in "12345678" (i.e. without the prefix)
3. Save => Success message is shown
4. Reload page => Hungary is selected as country and set in Stripe

saltcod:
trust y'all to know better than us what the implications are for these, but this file has had a bunch of fixes over the years.

All the EU coutnries were handled separately & I recall an issue as well with Cypress 

https://github.com/supabase/supabase/commits/436bdb10aeae3d1c10a145f859fe151b45474b51/studio/components/interfaces/Organization/BillingSettings/TaxID/TaxID.constants.ts?browsing_rename_history=true&new_path=apps/studio/components/interfaces/Organization/BillingSettings/TaxID/TaxID.constants.ts&original_branch=dd26541beb1b912477059d9a1b57d72076963ee1

ecktoteckto:
Found another issue (which is actually not related to this change though).

Select "United Kingdom Northern Ireland - XI VAT" from dropdown
Type in "XI123456789"
Save => Success message is shown and info is saved in Stripe correctly
Reload page => "United Kingdom - GB VAT" is selected

The reason is that attribute "type" is set to "eu_vat" for "United Kingdom" (without Northern Ireland) in `TaxID.constants.ts`. Is probably supposed to be "gb_vat"?

ecktoteckto (approved):

kevcodez:
> Select "United Kingdom Northern Ireland - XI VAT" from dropdown
> Type in "XI123456789"
> Save => Success message is shown and info is saved in Stripe correctly
> Reload page => "United Kingdom - GB VAT" is selected
> 
> The reason is that attribute "type" is set to "eu_vat" for "United Kingdom" (without Northern Ireland) in TaxID.constants.ts. Is probably supposed to be "gb_vat"?

Am aware, unfortunately both map to the same tax code (eu_vat) with the same country code (GB). So the only thing to properly map it would be to consider the tax prefix (XI) as well. I think it's fine for now, the display there doesn't matter much and Great Britain is not really wrong either.

The country dropdown is more of a help for the placeholder, we could even think about removing the dropdown and just showing a country flag + value in the future once it is set up. Can revisit in the future.

PR #27483, opened 2024-06-24
Handle if flag is false for restore disable check

joshenlim:


alaister (commented):
LGTM!

- Tested with a live and paused project on staging

alaister (approved):

PR #27494, opened 2024-06-24
fix minio healthcheck

terion-labs:
Minio has no curl and this is not correct healthcheck (and it fails)

See https://github.com/minio/minio/issues/18389

Also questionable to have in this compose file `storage` and `imgproxy` services — because if to start both, basic compose and compose.s3 — this will produce conflict in services


Hallidayo:
Hi @terion-name - We’re doing a bit of housekeeping and plan to close out PRs that have been inactive for more than 3 months.

PR #27495, opened 2024-06-24
chore: Regenerate `api-types` file with sorted option

ivasilov:
This PR adds the `alphabetize` option to the `codegen` command in `packages/api-types`. This will sort all types alphabetically which should improve the git diffs in the future.


joshenlim (approved):
No visual / functionality changes, minimally tested that the staging preview works on studio, docs and www

PR #27499, opened 2024-06-24
:bug: fix: wrap column around quotes for creating index sql command

### LINKED ISSUE #27447: Creating indexes on Supabase dashboard does not handle uppercase columns effectively
# Bug report

## Describe the bug

If a column name in a Postgres table has any uppercase characters and the user tries to create an index on that column, Supabase will not put quotes around the column and will keep it lowercase, throwing an error as the column does not exist

## To Reproduce

1. Create Postgres table with a column that has an uppercase name (such as "Visible")
2. Create an index on that column using the indexes tab in the Supabase dashboard
3. Error is thrown

## Expected behavior

The column should be wrapped with quotes so it enforces capitals to be recognized instead of assuming the column to be all lowercase. This makes it so the index is created without any issue for the user.

## Video

An example of this behavior is shown below. Instead of the index being created, an error is thrown, and I need to manually go to the SQL tab to wrap quotes around the column. It would be super helpful to the user if these quotes were wrapped already.

https://github.com/supabase/supabase/assets/59634395/70335af9-5369-4cb6-bdd9-9cc73522dbdf

## System information

- OS: macOS
- Browser (if applies): Chrome

girishsavlesha:

YES

#27447 

Bug fix

## What is the new behavior?

https://github.com/supabase/supabase/assets/29174602/eec53e86-c5d3-44f1-90f4-868cb88d49cc

## Additional context


ivasilov (approved):

ivasilov (approved):
Tested on studio staging.

PR #27506, opened 2024-06-24
Chore/longer toast time

saltcod:
Feedback submit success toast disappears before it can be read


charislam (approved):
Checked: opened preview

PR #27511, opened 2024-06-25
docs: Remove extra `await` from Swift sample

roz0n:
Removed extra await in `AppView` example

Yes

Docs update

There is an extra `await` in a Swift code example which prevents the project from building:

```
Group {
  if isAuthenticated {
    ProfileView()
  } else {
    AuthView()
  }
}
.task {
  for await state in await supabase.auth.authStateChanges {     // this is invalid syntax
    if [.initialSession, .signedIn, .signedOut].contains(state.event) {
      isAuthenticated = state.session != nil
    }
  }
}
```

## What is the new behavior?

The extra `await` is removed and the application builds successfully as a result.

## Additional context

### DIFF
--- a/apps/docs/content/guides/getting-started/tutorials/with-swift.mdx
+++ b/apps/docs/content/guides/getting-started/tutorials/with-swift.mdx
@@ -272,7 +272,7 @@ struct AppView: View {
       }
     }
     .task {
-      for await state in await supabase.auth.authStateChanges {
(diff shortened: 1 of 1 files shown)


grdsdev (approved):
Thanks for the fix.


Compiler issues just a warning for that, it doesn't prevent from building in my case.

But thanks for that.

PR #27522, opened 2024-06-25
Trim column name in table editor when updating column

joshenlim:


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
```diff
@@ -123,18 +123,20 @@ export const generateUpdateColumnPayload = (
...
+  const name = field.name.trim()
   const type = field.isArray ? `${field.format}[]` : field.format
-  const comment = (field.comment?.length ?? '') === 0 ? null : field.comment
+  const comment = ((field.comment?.length ?? '') === 0 ? null : field.comment)?.trim()
+  const check = field.check?.trim()
 
   const payload: Partial<UpdateColumnPayload> = {}
-  if (!isEqual(originalColumn.name, field.name)) {
-    payload.name = field.name
+  if (!isEqual(originalColumn.name, name)) {
```
alaister, line 132: So we don't accidentally rename columns that already contain a space is it worth adding a `trim()` here?
```suggestion
  if (!isEqual(originalColumn.name.trim(), name)) {
```

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
```diff
@@ -123,18 +123,20 @@ export const generateUpdateColumnPayload = (
...
+    payload.name = name
   }
   if (!isEqual(originalColumn.comment, comment)) {
```
alaister, line 135: Renaming the comment should be fine unless they're using the comments for something else like postgraphile etc and they need some crazy format or something... so maybe worth adding a `.trim()` here as well?

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
joshenlim: hmm very fair point - otherwise might bring down their app
i'll add a comment as well

alaister (approved):
LGTM!

- Tested adding and renaming column names, descriptions, and checks

PR #27533, opened 2024-06-25
Connection strings to be prefixed with postgresql:// instead of postgres://

joshenlim:


alaister (approved):
LGTM!

- Tested on preview

PR #27553, opened 2024-06-26
Chore/optimizations for table editor query

joshenlim:


alaister (approved):
LGTM!

- Tested using a table with ~2 million rows, saw roughly a 2x speed-up in query time
- Tested with sorts and filters
- Tested using a table with no text columns
- Tested SQL editor with a limit/offset query

PR #27559, opened 2024-06-26
Add bullet point for custom otp emails using admin api or hooks

encima:

YES

Docs

Feedback has noted that the current docs for https://supabase.com/docs/reference/javascript/auth-admin-generatelink is hard to find, so linking to it from https://supabase.com/docs/reference/javascript/auth-signinwithotp

## What is the new behavior?

Now we gots ourselves some links. Also added a shoutout to Auth Hooks as they are a newer feature


charislam (approved):
Checks: opened preview

PR #27572, opened 2024-06-26
Chore/add lint

saltcod:
We're not handling lint 017, breaking the ui

https://supabase.com/docs/guides/database/database-linter?lint=0017_foreign_table_in_api

screenshot-2024-06-26-at-16 54 24


apps/studio/components/interfaces/Linter/Linter.utils.tsx
```diff
@@ -161,13 +161,23 @@ export const lintInfoMap: LintInfo[] = [
   {
     name: 'materialized_view_in_api',
     title: 'Materialized View in API',
-    icon: <Eye className="text-foreground-muted" size={15} strokeWidth={1.5} />,
+    icon: <Table2 className="text-foreground-muted" size={15} strokeWidth={1.5} />,
```
alaister, line 164: Are these icons the wrong way around now? Feel like the view should be the eye and the table should be the table? 😅

joshenlim: ^ yeah agree, should be consistent with the table editor?

joshenlim: i'll make the change in this PR so that we can get this through, can update if we got this one wrong thereafter

joshenlim (approved):
LGTM - tested on staging preview by creating a FDW table, then checking security advisor UI

PR #27577, opened 2024-06-27
fix: linting issues

charislam:


supabase/migrations/20240626184716_misc_database_fixes.sql
```diff
@@ -0,0 +1,151 @@
...
+
+-- Return a setof page_section so that we can use PostgREST resource embeddings (joins with other tables)
+create or replace function match_page_sections_v2(
+  embedding vector(1536),
+  match_threshold float,
+  min_content_length int
+)
+returns setof page_section
+language plpgsql
+set search_path = ''
```
gregnr, line 15: Any reason we don't use the `alter function` syntax for these like for `update_last_changed_checksum` above (anything else we're changing)?

charislam: I had to add in the explicit schemas (otherwise the function will fail to run because it doesn't know where to search for the relations). The functions above that I could just alter already had explicit schemas so a a simple alter was fine

gregnr (approved):

PR #27581, opened 2024-06-27
move code snippet above tip that references it

Trevato:

YES

docs update

A tip says to "use the script above" but the script is below lol.

## What is the new behavior?

I just moved the script above the text so it makes sense.

## Additional context

I'm trying to get more involved in open source so this was mostly an exercise for myself.


Hallidayo (changes requested):
Hi @Trevato - thank you the PR. I think instead of moving the block could you change the text to:

`Add the following script to your`

Trevato:
> Hi @Trevato - thank you the PR. I think instead of moving the block could you change the text to:
> 
> 
> 
> `Add the following script to your`

Updated.

Hallidayo (approved):
Brill thank you @Trevato 👍

PR #27592, opened 2024-06-27
Fix: pooler connection limit warning message

monicakh:

Bug fix

The warning message indicates the maximum connection limit for direct connections, not for the connection pooler.

## What is the new behavior?

Screenshot 2024-06-27 at 2 02 05 PM


w3b6x9 (changes requested):
The original is correct; we want it to be 80% of DB's max conn.

PR #27602, opened 2024-06-27
format dates in default renderer

jordienr:

YES

Default renderer for rows in logs explorer doesnt format the dates

Now, if the column contains `timestamp` it will format the value


apps/studio/components/interfaces/Settings/Logs/LogTable.tsx
```diff
@@ -121,6 +122,8 @@ const LogTable = ({
           return JSON.stringify(value)
         } else if (value === null) {
           return 'NULL'
+        } else if (String(v).includes('timestamp')) {
```
Ziinc, line 125: should not render to local tz by default for logs explorer because a query might explicitly convert to a different timezone but still have the same column name
```
select datetime(current_timestamp(), 'Asia/Tokyo') as timestamp
``` 
This will then get shown as Italy time if user was in italy.

Would be more explicit if the default query was changed to `datetime(timestamp, 'UTC')` or something similar.

Ziinc (changes requested):

PR #27612, opened 2024-06-28
Optimize table editor count

joshenlim:
`select count(*) from table` can be really expensive in particular if your table is really large, we've seen this being the cause of some problems where the query was excessively consuming resources on the database. 

As such we're going to retrieve an _estimate_ count of the table if the table has more than 50,000 rows, otherwise we will return the exact count. This applies to tables, whereas for other entities like views, materialized views, etc we will always fetch the estimate count. The exact count can then be fetched on demand.

image

A warning will also show up when trying to retrieve exact count on demand for tables with row count greater than 50,000
image

However, in the case for views etc (not tables), if the estimated row count is below 50,000 we'll skip the warning


apps/studio/data/table-rows/table-rows-count-query.ts
```diff
@@ -9,31 +9,87 @@ import { formatFilterValue } from './utils'
...
-  return sql
+with approximation as (
+    select reltuples as estimate
+    from pg_class
+    where oid = ${table.id}
+)
+select 
+  case 
+    when estimate = -1 then (select pg_temp.count_estimate('${selectBaseSql}'))
+    when estimate > ${THRESHOLD_COUNT} then ${filters.length > 0 ? `pg_temp.count_estimate('${selectBaseSql}')` : 'estimate'}
```
alaister, line 79: Single quotes in these two need to be escaped (for example `select * from posts where title = 'My Post'` should become `select * from posts where title = ''My Post''`

alaister (changes requested):
Looking good! Running into an error due to some missing quotes though:

apps/studio/data/table-rows/table-rows-count-query.ts
joshenlim: ahh good catch! will fix these 🙏

apps/studio/components/grid/components/footer/pagination/Pagination.tsx
```diff
@@ -122,96 +111,107 @@ const Pagination = () => {
...
   }
 
+  useEffect(() => {
+    if (page && page > totalPages) {
+      snap.setPage(totalPages)
+    }
+  }, [page, totalPages])
+
+  // [Joshen] Oddly without this, state.selectedRows will be stale
+  useEffect(() => {}, [state.selectedRows])
```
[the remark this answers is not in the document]
joshenlim: i've tried to break this when L132 is removed - and it seems alright actually
Tested:
- checking rows,
- unchecking rows,
- checking all rows
- unchecking all rows
- doing things when rows are checked (e.g deleting rows)

i'll remove this line

apps/studio/data/table-rows/table-rows-count-query.ts
```diff
@@ -9,31 +9,85 @@ import { formatFilterValue } from './utils'
...
-  return sql
+with approximation as (
+    select reltuples as estimate
+    from pg_class
+    where oid = ${table.id}
+)
+select 
+  case 
+    when estimate = -1 then (select pg_temp.count_estimate('${selectBaseSql.replace("'", "''")}'))
+    when estimate > ${THRESHOLD_COUNT} then ${filters.length > 0 ? `pg_temp.count_estimate('${selectBaseSql.replace("'", "''")}')` : 'estimate'}
```
alaister, line 79: sneaky little one
```suggestion
    when estimate = -1 then (select pg_temp.count_estimate('${selectBaseSql.replaceAll("'", "''")}'))
    when estimate > ${THRESHOLD_COUNT} then ${filters.length > 0 ? `pg_temp.count_estimate('${selectBaseSql.replaceAll("'", "''")}')` : 'estimate'}
```

alaister (changes requested):

apps/studio/data/table-rows/table-rows-count-query.ts
joshenlim: ahh thanks for the catch!! 🙏

alaister (approved):
LGTM!

- Tested on preview with various filters on tables and views

PR #27615, opened 2024-06-28
Fix changing provider doesnt change default region

joshenlim:


alaister (approved):
LGTM!

- Tested on preview
- Changed provider and region on new project form

Issue #27505, opened 2024-06-24
NextJS Middleware Auth Examples recommends using function that doesn't exist 


ethanniser:
# Improve documentation

## Link 

https://supabase.com/docs/guides/auth/server-side/nextjs?queryGroups=router&router=app

## Describe the problem

The main issue is that the current example is structured in a way were it assumes the only thing you will ever do in your middleware is authenticate with supabase, but often I would think that is not the case.

My current project has some complex redirection logic, so I can't just simply return the `supabaseResponse`. The docs suggest making a new `NextResponse` and manually setting the cookies
```ts
const myNewResponse = NextResponse.next({ request })
myNewResponse.cookies.setAll(supabaseResponse.cookies.getAll())
```

The only problem is the `cookies` property on `NextResponse` is of type `ResponseCookies` which does not actually have a `setAll` method.

## Describe the improvement

It would be fantastic if there was:

 - A recommended way to pass on the Supabase cookies to new responses
 - An example where the middleware does more than simply `return await updateSession` 
 - Maybe an example repo to go off of and could be linted as to avoid recommending non existent or depreciated apis 

---

I'm upgrading my auth from 0.6 `@supabase/auth-helpers-nextjs` and I was actually in the process of writing a separate issue earlier asking for the docs to no longer recommend `get`, `set` and `delete`, but #27242 literally merged as I was writing it (thanks @hf), but I've run into this separate issue now. If any other additional information is needed just let me know- happy to help.

Thanks so much

ErikPetersenDev:
Just to link things together I thought I'd share that there's also a Discussion I just came across that covers a couple other documentation issues on the same page (https://supabase.com/docs/guides/auth/server-side/nextjs?queryGroups=router&router=app).

Discussion: https://github.com/orgs/supabase/discussions/27619

To summarize quickly:

- one instance of `data: user` should be `data: { user}` (otherwise `if` block is never executed)
- one instance of `NextResponse.redirect('/login')` doesn't work (referance: https://nextjs.org/docs/messages/middleware-relative-urls)... I believe it should be:
```
const url = request.nextUrl.clone();
url.pathname = "/login";
return NextResponse.redirect(url);
```

wadjeroudi:
@ErikPetersenDev running in this issue, could you propose a solution for the setAll issue ?
In the last updated doc, it still has the setAll call on the new response but it doesn't exist, like @ethanniser said

My previous middleware but using the new documentation, I loose the cookies

```js
[...]

PR #27622, opened 2024-06-28
Fix nextjs example in ssr docs

ErikPetersenDev:

YES

This PR makes some fixes to code examples in Next.js SSR docs for auth.   The modifications affect documentation that was updated within the past week via PR: https://github.com/supabase/supabase/pull/27242.

The fixes are to this page: https://supabase.com/docs/guides/auth/server-side/nextjs?queryGroups=router&router=app

There are a few small issues in the code example for `utils/supabase/middleware.ts`

The first problem is that `user` is not destructured from the getUser response.  Due to this, the "user" is always truthy and the `if` block that follows will never execute.  Changing `data: user` to `data: { user }` resolves this issue.

Once that's fixed an error is thrown from the `if` block of code because Next.js middleware does not allow relative urls.  The error points to these Next.js docs: https://nextjs.org/docs/messages/middleware-relative-urls.

The suggested/recommended fix is to replace `NextResponse.redirect('/login')` with:
```
const url = request.nextUrl.clone();
url.pathname = "/login";
return NextResponse.redirect(url);
```

Once that's addressed one more fix is required:  the middleware executes on every route, so we need to skip the login route to avoid infinite redirects.  Therefore, the `if` statement needs to change from `if (!user)` to `if (!user && !request.nextUrl.pathname.startsWith("/login"))`.

## What is the new behavior?

The revised code destructures the user property, updates the redirect to the code recommended by the Next.js docs, and skips the login route when redirecting.

## Additional context

These issues were brought up in a Discussion here: https://github.com/orgs/supabase/discussions/27619

## Not addressed

Once the example is fixed, the code will redirect to the login page for _all_ routes, include the root (`/`).  If this is not the intended behavior then further changes to the example will be required.  However, this PR at the moment only addresses the bugs and not the intent.   An alternate would be to comment out the fixed version of the `if` block with redirect and leave it there as an example.  Otherwise, someone testing this out will be redirected to login for every route instead of only protected routes (which I think is what the examples are trying to demonstrate).

There's a separate open Issue about the same documentation page.  Specifically related to some code located within a comment.  That issue is not addressed by this change.  Issue:  https://github.com/supabase/supabase/issues/27505


charislam (approved):
Thanks @ErikPetersenDev!

ErikPetersenDev:
@charislam Looks like all the checks are passing.  I had to push one change after your approval to fix the formatting with Prettier.  Please let me know if there's anything else you need from me!

charislam:
Thanks! Everything looks great, we really appreciate the help!

PR #27642, opened 2024-06-29
fix docs creating-a-client.mdx

MiryangJung:

YES

fix docs code example

## Additional context

The data returned by the `getUser` cannot be verified with `!data` because it is of the following type.

```
data: {user: User} | {user: null}
```


ErikPetersenDev:
I submitted a PR yesterday (approved/merged) that fixed this same error in a different location in the docs.  There are more bugs though once you fix the destructuring issue.  One is related to the use of a relative URL in middleware and the other related to a redirect loop.

Take a look at that PR linked below.  We should probably make the exact same changes in both places to keep the guidance consistent.  I wasn't aware of this other instance at the time (should've searched!).

Here's the PR that was approved/merged, which includes an explanation of the other changes that were made:
https://github.com/supabase/supabase/pull/27622

PR #27655, opened 2024-06-30
docs: Kotlin docs misc updates

dshukertjr:

YES

Kotlin docs update.

- Updating the "installing" section to place the code blocks on the right by using `RefSubLayout.Details` and `RefSubLayout.Examples`
- Adding decoding example on `invoke()` method
- Updating realtime code samples to launch the flows in a new coroutine.

### DIFF
--- a/apps/docs/docs/ref/kotlin/installing.mdx
+++ b/apps/docs/docs/ref/kotlin/installing.mdx
@@ -8,7 +8,7 @@ custom_edit_url: https://github.com/supabase/supabase/edit/master/web/spec/supab
 ### Add one or more modules to your project
 
 <RefSubLayout.EducationRow>
-    <RefSubLayout.Examples>
+    <RefSubLayout.Details>
         <div style={{ display: 'flex', flexDirection: 'row', alignItems: 'center' }}>
             <a href="https://github.com/supabase-community/supabase-kt/releases" style={{ marginRight: '8px' }}>
                 <img src="https://img.shields.io/github/release/supabase-community/supabase-kt?label=stable" alt="latest stable supabase-kt version"/>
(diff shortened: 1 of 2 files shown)


dshukertjr:
@jan-tennert If you could also take a look that would be awesome!

apps/docs/docs/ref/kotlin/installing.mdx
```diff
@@ -17,54 +17,19 @@ custom_edit_url: https://github.com/supabase/supabase/edit/master/web/spec/supab
...
         - [**storage-kt**](https://github.com/supabase-community/supabase-kt/tree/master/Storage)
         - [**functions-kt**](https://github.com/supabase-community/supabase-kt/tree/master/Functions)
         - [**postgrest-kt**](https://github.com/supabase-community/supabase-kt/tree/master/Postgrest)
-        - [apollo-graphql](https://github.com/supabase-community/supabase-kt/tree/master/plugins/ApolloGraphQL) - Creates an [Apollo GraphQL Client](https://github.com/apollographql/apollo-kotlin) for interacting with the Supabase API.
-        - [compose-auth](https://github.com/supabase-community/supabase-kt/tree/master/plugins/ComposeAuth) - Provides easy Native Google & Apple Auth for Compose Multiplatform targets.
-        - [compose-auth-ui](https://github.com/supabase-community/supabase-kt/tree/master/plugins/ComposeAuthUI) - Provides UI Components for Compose Multiplatform.
-        - [coil-integration](https://github.com/supabase-community/supabase-kt/tree/master/plugins/CoilIntegration) - Provides a [Coil](https://github.com/coil-kt/coil) Integration for displaying images stored in Supabase Storage.
```
jan-tennert, line 63: If we remove GraphQL here, we should probably also remove the GraphQL snippet in the "Initializing" section

apps/docs/spec/supabase_kt_v2.yml
```diff
@@ -4393,7 +4393,10 @@ functions:
         isSpotlight: true
         code: |
           ```kotlin
-          supabase.functions.invoke("function_name")
+          val response = supabase.functions.invoke("function_name")
+
+          // Decode the response body to a serializable class
+          val data = res.body<FunctionResponse>()
```
jan-tennert, line 4399:
```suggestion
          val response = supabase.functions.invoke("function_name")

          // Decode the response body to a serializable class
          val data = response.body<FunctionResponse>()
```
If we expand the snippet here, should we also do this with the other function snippets? Also, if we introduce the `body` function, we should probably add a comment in the description. Maybe something like "The return type of the invoking methods is an `HttpResponse`. You can access the response body via `response.body<T>()`, where T can be any serializable type."

jan-tennert (commented):
Two small suggestions, otherwise looks good!

apps/docs/docs/ref/kotlin/installing.mdx
dshukertjr: It's really up to you, but I personally think it's alright to leave the code snippet in the `initializing` section as it is still a valid example, and we still have the mentions of those additional plugins in the docs just all bundled together.

apps/docs/spec/supabase_kt_v2.yml
dshukertjr: >If we expand the snippet here, should we also do this with the other function snippets?

When we do add more functions examples, we probably should add these decoding examples wherever appropriate. For now, I think we are good.

jan-tennert (approved):
LGTM!

charislam (approved):

PR #27659, opened 2024-06-30
docs: fix Next.js create client docs to match fix in ssr auth docs

ErikPetersenDev:

Yes

This is a documentation update to ensure fixes made to a documentation code example Friday are applied to another location in the docs where the identical code exists.

There are a couple issues in the current docs for creating a client here: https://supabase.com/docs/guides/auth/server-side/creating-a-client?queryGroups=environment&environment=middleware

The exact same code issues were fixed two days ago in another location in the Next.js docs, but I missed that the same example code was located in two places.  That PR is here, which includes a detailed description of the issues and the changes:  https://github.com/supabase/supabase/pull/27622.   The live version of that page, including the changes made Friday, is here: https://supabase.com/docs/guides/auth/server-side/nextjs?queryGroups=router&router=app

## What is the new behavior?

This PR updates the Next.js "creating a client" docs to fix the issues and match exactly what was merged to the other Next.js page.

Summary of three changes made:

- Destructure response from `getUser` to get the `user` object.  The existing code results in a "user" that is always truthy, which causes the subsequent `if` block to never run.
- The existing redirect (which doesn't execute at the moment) doesn't work in middleware.  The code is updated to match the recommended solution noted in the Next.js docs here: https://nextjs.org/docs/messages/middleware-relative-urls
- Finally, the `/login` path needs to be exempted from the `if` statement to prevent a redirect loop.
- EDIT (see comment below):  The `/auth` path also needs to be exempted.

## Additional context

As noted in the previous PR, this does not address whether the intent is to always redirect unauthorized users to login, even for visits to the root (`/`).


ErikPetersenDev:
@charislam FYI.  I found out the documentation changes I made Friday need to be applied to a different page in the docs that includes the same exact code example.  Only pinging you here since you were involved in that one, but someone else can review/approve of course!

Merged PR from Friday:
https://github.com/supabase/supabase/pull/27622

ErikPetersenDev:
A comment was made on the previous PR indicating that it's still not fully working.  Specifically, routes starting with `auth` need to be skipped as well otherwise they'll get redirected to `/login` as well (any "public" route needs to be explicitly listed in the current code example).

I've pushed a new commit to this PR to add the `auth` route to both locations in the docs.  Open to other, more extensible solutions this as well, but it would require larger changes (e.g., a separate place in the code to centralize "public" routes and then match them)... for now it's just handled in the `if` statement but that wouldn't scale as additional public routes are added.

Here's the comment:  https://github.com/supabase/supabase/commit/9a18d013c29b11851e922b5b1d19780ca94184cb#commitcomment-143698171

charislam (approved):
Thanks @ErikPetersenDev! Agreed that we need to document a more extensible pattern for determining when to redirect, but happy to have a quick fix for now! ❤️

PR #27738, opened 2024-07-01
Invalid SQL syntax in rls bypass section

GaryAustin1:
Unless there is some new SQL I'm not familiar the code shown: `grant bypassrls on "table_name" to "role_name";`
Will not work and instead should be:
`alter role "role_name" with bypassrls;`

YES

 docs update, ...

image


charislam (approved):
Thanks @GaryAustin1!

PR #27751, opened 2024-07-02
Realtime Inspector adjustments

saltcod:
misc fixes for the realtime inspector policies

**addresses these:** 

https://www.notion.so/Feedback-Request-for-Task-and-Bug-Fix-f9809d743bda479099463b528e1df9d9

https://www.notion.so/supabase/Remove-realtime-schema_migrations-and-realtime-subscriptions-from-Realtime-Policy-Creation-Form-53c86773713f42dda48585e93f80f8ef


apps/studio/components/interfaces/Auth/Policies/PolicyEditorModal/PolicyEditorModal.constants.ts
```diff
@@ -192,23 +192,7 @@ using ( realtime.messages.extension = 'broadcast' );`.trim(),
...
-ON realtime.messages for update
-TO authenticated
-using ( realtime.messages.extension = 'broadcast' )
-with check ( realtime.messages.extension = 'broadcast' );`.trim(),
-      name: 'Allow pushing broadcasts for authenticated users only',
-      definition: "realtime.messages.extension = 'broadcast'",
-      check: "realtime.messages.extension = 'broadcast'",
-      command: 'UPDATE',
-      roles: ['authenticated'],
-    },
```
w3b6x9, line 211: @saltcod we still want the policies, just not the `UPDATE` ones. Can you just replace these `UPDATE` policies with `INSERT` ones?

w3b6x9 (commented):
@saltcod left you a comment.

Can you also take out the UPDATE, DELETE, and ALL radio selections from here?
Screenshot 2024-07-02 at 14 54 06

apps/studio/components/interfaces/Auth/Policies/PolicyEditorModal/PolicyEditorModal.constants.ts
saltcod: gotcha gotcha

w3b6x9 (approved):
@saltcod lgtm, thanks!

apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
```diff
@@ -131,63 +131,11 @@ export const PolicyDetailsV2 = ({
                   </p>
                 </FormLabel_Shadcn_>
                 <FormControl_Shadcn_>
-                  <Popover_Shadcn_ open={open} onOpenChange={setOpen} modal={false}>
```
joshenlim, line 134: hold on this component is used in multiple places (primarily in the auth/policies) page - the changes here will affect the UI wherever this component is used and not just for the realtime inspector

joshenlim (changes requested):
Reference to https://github.com/supabase/supabase/pull/27751/files#r1663519641

The changes here affects more than just the realtime inspector which shouldn't be the case. Probably needs some prop + conditional handling instead

apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
saltcod: oh god, thanks Joshen! That was a giant misread on my part!

apps/studio/components/interfaces/Realtime/Inspector/useRealtimeMessages.ts
```diff
@@ -190,7 +203,7 @@ export const useRealtimeMessages = (
 
         newChannel.unsubscribe()
         setChannel(undefined)
-        setRealtimeConfig({ ...config, enabled: false })
+        setRealtimeConfig({ ...config, channelName: '', enabled: false })
```
joshenlim, line 206: just a nit here - any reason why we need to reset the channel name input field if there's an error?
just thinking from the user's POV - if say this turned out to be a mistake and i wanna test, i'd need to key in the channel name again then

joshenlim, line 206: nit feedback (no need to action in this PR)
am thinking a better UX when running into this bug here is that
- The Channel popover shows up
- The input field of the channel should retain as the channel name that failed to connect
- input field perhaps should auto focus

But shall we update the toast.error to also include the channel name that error-ed out too? am thinking otherwise theres no other indication of what channel error-ed out other than remembering what you typed into the input field prior, something like "Failed to connect to channel 'channel_name': ..."

joshenlim (approved):
Nit comment but otherwise LGTM!

Tested on staging preview
- Auth/policies, UI is unaffected
- Realtime/policies, No longer have update delete and all commands for policy creator

apps/studio/components/interfaces/Realtime/Inspector/useRealtimeMessages.ts
saltcod: Yeah good shout. Added the channel name to the toast. 
Needs a little re-plumbing to do the rest, let's leave for the moment.

saltcod:
checked preview

PR #27756, opened 2024-07-02
docs: document more ways to secure the data api

charislam:
Document ways to increase the security of Data API, including:

- Disabling it entirely
- Exposing a custom schema instead of the `public` schema

Also found a duplicated link in the nav for CLS, so removed it.

### DIFF
--- a/apps/docs/content/guides/database/hardening-data-api.mdx
+++ b/apps/docs/content/guides/database/hardening-data-api.mdx
@@ -2,7 +2,7 @@
 title: 'Hardening the Data API'
 ---
 
-Your database's automatically generated Data API exposes the `public` schema by default. Since the `public` schema is often used by other tools as a default space, you might want to lock down this schema. This helps prevent accidental exposure of data that's automatically added to `public`.
+Your database's automatically generated Data API exposes the `public` schema by default. If your `public` schema is used by other tools as a default space, you might want to lock down this schema. This helps prevent accidental exposure of data that's automatically added to `public`.
 
 There are two levels of security hardening for the Data API:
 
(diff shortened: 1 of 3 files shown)


apps/docs/content/guides/database/hardening-data-api.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+    ```sql
+    create schema if not exists api;
+    ```
+
+1.  Grant the `anon` and `authenticated` roles usage on this schema. Also grant any other permissions they will need, such as ability to read and edit tables:
+
+    ```sql
+    grant usage on schema api to anon, authenticated;
+    grant select, insert, update, delete on all tables in schema api to anon, authenticated;
```
steve-chavez, line 46: This would be an insecure setup. `anon` should not be granted any privilege by default. `authenticated` should also be granted privileges explicitly for each table and operation.

As it is users will just copy paste this and end up with a similar setup as Supabase defaults.

For now I suggest:

```suggestion
  grant select on table <your_table> to anon; -- only read access for anon users  
  grant select, insert, update, delete on table <your_table> to authenticated; -- read and write access for authenticated users  
```

steve-chavez (approved):
LGTM for starters :+1:

apps/docs/content/guides/database/hardening-data-api.mdx
```diff
@@ -0,0 +1,57 @@
+---
+title: 'Hardening the Data API'
+---
+
+Your database's automatically generated Data API exposes the `public` schema by default. Since the `public` schema is often used by other tools as a default space, you might want to lock down this schema. This helps prevent accidental exposure of data that's automatically added to `public`.
```
kiwicopple, line 5:
```suggestion
Your database's automatically generated Data API exposes the `public` schema by default. If your `public` schema is used by other tools as a default space, you might want to lock down this schema. This helps prevent accidental exposure of data that's automatically added to `public`.
```

kiwicopple (approved):

PR #27781, opened 2024-07-03
enhance: move not-found data fetch to client side

charislam:
Fetching recommendations on the server takes too long before the first load, so it's faster to load the basic not-found page and generate the recommendations on the client side.

### DIFF
--- a/apps/docs/app/not-found/page.tsx
+++ b/apps/docs/app/not-found/page.tsx
@@ -17,7 +17,7 @@ const metadata: Metadata = {
 const NotFoundPage = async ({ searchParams: { page } }: { searchParams: { page?: string } }) => {
   return (
     <article className="prose max-w-[80ch]">
-      <h1>We couldn&apos;t find that page</h1>
+      <h1>404: We couldn&apos;t find that page</h1>
       <p>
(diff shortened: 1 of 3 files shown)


apps/docs/app/not-found/page.tsx
```diff
@@ -15,25 +15,34 @@ const metadata: Metadata = {
...
-    ' '
-  )
-  if (!query) return null
-
-  const supabase = createClient<Database>(
-    process.env.NEXT_PUBLIC_SUPABASE_URL!,
-    process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY!
+  return (
+    <article className="prose max-w-[80ch]">
+      <h1>We couldn&apos;t find that page</h1>
```
fsansalvadore, line 20: Should we keep a mention to `404` in the title?
Like `404: page not found`, to maintain familiarity with the 404 pages pattern @charislam

fsansalvadore (approved):

apps/docs/app/not-found/page.tsx
charislam, line 20:
```suggestion
      <h1>404: We couldn&apos;t find that page</h1>
```

PR #27787, opened 2024-07-03
Add labels and grid to sql editor charts

jordienr:

YES

--

this makes it way better for sharing screenshots of the charts.

CleanShot 2024-07-04 at 00 41 29@2x

### DIFF
--- a/apps/studio/components/interfaces/SQLEditor/UtilityPanel/ChartConfig.tsx
+++ b/apps/studio/components/interfaces/SQLEditor/UtilityPanel/ChartConfig.tsx
@@ -1,3 +1,4 @@
+import { Checkbox } from '@ui/components/shadcn/ui/checkbox'
 import BarChart from 'components/ui/Charts/BarChart'
 import NoDataPlaceholder from 'components/ui/Charts/NoDataPlaceholder'
 import dayjs from 'dayjs'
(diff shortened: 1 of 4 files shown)


alaister (approved):
LGTM! Although there seems to be a TS error that needs fixing

- Tested on staging with some dummy data

apps/studio/components/interfaces/SQLEditor/UtilityPanel/ChartConfig.tsx
```diff
@@ -168,6 +188,28 @@ export function ChartConfig({ results = { rows: [] }, config, onConfigChange }:
...
               </Label_Shadcn_>
             </div>
+            <div className="flex items-center">
+              <Checkbox_Shadcn_
+                id="showLabels"
+                name="showLabels"
+                checked={config.showLabels}
+                onClick={() => onConfigChange({ ...config, showLabels: !config.showLabels })}
+              />
+              <Label_Shadcn_ className="text-foreground-light p-1.5" htmlFor="showLabels">
```
MildTomato, line 198: It's possible these labels won't be interactive unless you wrap a <form> around the whole thing. a weird Radix thing, it only inserts a hidden `<input />` when inside a `<form />`

MildTomato (dismissed):

apps/studio/components/interfaces/SQLEditor/UtilityPanel/ChartConfig.tsx
jordienr: seems to work fine (clicking either the label or the checkbox) but added the <form> justincase something changes in the future 🤔

PR #27818, opened 2024-07-05
Support filtering projects by status, only either paused or active

joshenlim:
image

### DIFF
--- a/apps/studio/components/interfaces/HomePageActions.tsx
+++ b/apps/studio/components/interfaces/HomePageActions.tsx
@@ -5,27 +5,36 @@ import { useRouter } from 'next/router'
 import { useOrganizationsQuery } from 'data/organizations/organizations-query'
 import { useIsFeatureEnabled } from 'hooks/misc/useIsFeatureEnabled'
 import { EMPTY_ARR } from 'lib/void'
+import { Filter, Search } from 'lucide-react'
 import {
   Button,
+  Checkbox_Shadcn_,
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
   DropdownMenuLabel,
   DropdownMenuTrigger,
-  IconSearch,
   Input,
+  Label_Shadcn_,
+  PopoverContent_Shadcn_,
+  PopoverTrigger_Shadcn_,
+  Popover_Shadcn_,
 } from 'ui'
 
 interface HomePageActionsProps {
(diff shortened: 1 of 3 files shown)


alaister (approved):
LGTM!

- Tested on staging

Small nit: probably don't need to show the "select only" option when there's only two options:

Screenshot 2024-07-05 at 18 57 09

saltcod (approved):

apps/studio/components/interfaces/HomePageActions.tsx
```diff
@@ -62,14 +71,65 @@ const HomePageActions = ({
...
+            <div className="px-3 pt-3 pb-2 flex flex-col gap-y-2">
+              <p className="text-xs">Filter projects by status</p>
+              <div className="flex flex-col">
+                {[
+                  { key: 'ACTIVE_HEALTHY', label: 'Active' },
+                  { key: 'INACTIVE', label: 'Paused' },
+                ].map(({ key, label }) => (
+                  <div key={key} className="group flex items-center justify-between py-0.5">
+                    <div className="flex items-center gap-x-2">
+                      <Checkbox_Shadcn_
```
MildTomato, line 102: Might need to wrap all this in a html form so that the checkbox labels are clickable

MildTomato: @joshenlim

apps/studio/components/interfaces/HomePageActions.tsx
```diff
@@ -62,14 +71,65 @@ const HomePageActions = ({
...
+                      />
+                      <Label_Shadcn_ htmlFor={key} className="capitalize text-xs">
+                        {label}
+                      </Label_Shadcn_>
+                    </div>
+                    <Button
+                      size="tiny"
+                      type="default"
+                      onClick={() => setFilterStatus([key])}
+                      className="transition opacity-0 group-hover:opacity-100 h-auto px-1 py-0.5"
```
MildTomato, line 122: This will possibly break with keyboard only usage, since the buttons will focus and not be visible

apps/studio/components/interfaces/HomePageActions.tsx
joshenlim: oh are they not? i just checked on prod - the checkbox labels are clickable (which will toggle the checked state of the checkboxes)

PR #27828, opened 2024-07-05
fix: revalidate fetches every day

charislam:
The Vercel Data Cache caches data across deployments, and we want to check for updates in federated content, so let's revalidate the data once a day.

It may be more efficient to revalidate on-demand, since we don't change most of these sources once a day, but that requires setting up an API endpoint (and figuring out how to authenticate it), so this is quick-and-easy for now.


fsansalvadore (approved):
Looks good to me 👍

PR #27857, opened 2024-07-08
ref: Cleanup vercel_configuration_id metadata

kamilogorek:
This is a stale data that we do not need to keep around anymore.


joshenlim (approved):
Tested on staging preview minimally with creating a new project via the dashboard (not vercel integration)

PR #27858, opened 2024-07-08
feat: Reorder new project panels

ivasilov:
This PR reorders the sections in the new project dialog so that the `compute billing` is near the `compute size` section (I've also removed the separator line between them). Previously, it was after the `Advanced Options` (renamed to `Security Options`).

Previously:
Screenshot 2024-07-08 at 15 36 10

After:
Screenshot 2024-07-08 at 15 35 59


alaister (approved):
LGTM!

- Tested changing compute size
- Created a new project on staging

PR #27894, opened 2024-07-10
Support hardening api via custom schema in api settings

joshenlim:
image


apps/studio/components/interfaces/Settings/API/HardenAPIModal.tsx
```diff
@@ -0,0 +1,287 @@
...
+        db_extra_search_path: config?.db_extra_search_path,
+      },
+    })
+  }
+
+  return (
+    <Dialog open={visible} onOpenChange={onClose}>
+      <DialogContent size="xlarge">
+        <DialogHeader>
+          <DialogTitle>Prevent accidental exposure of data via the API</DialogTitle>
```
kiwicopple, line 112:
```suggestion
          <DialogTitle>Switch the default API schema</DialogTitle>
```

apps/studio/components/interfaces/Settings/API/HardenAPIModal.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+        <DialogSectionSeparator />
+
+        <DialogSection className="text-sm text-foreground-light">
+          <p>
+            If you want to use the Data API with increased security, we recommend exposing a custom
+            schema instead of the <code className="text-xs text-foreground">public</code> schema to
+            get more <span className="text-brand">conscious control</span> over your exposed data.
+            Any data, views, or functions that should be exposed need to be deliberately put within
+            your custom schema instead.
```
kiwicopple, line 127:
```suggestion
            By default, the <code className="text-xs text-foreground">public</code> schema is used to
            generate API routes. In some cases it's better to use a custom schema.
```

kiwicopple (commented):
some suggestions

apps/studio/components/interfaces/Settings/API/HardenAPIModal.tsx
joshenlim: agreed this is much better

alaister (commented):
Looking good! Small one, but I can really make sure the `public` schema is removed from exposed schemas:

https://github.com/supabase/supabase/assets/10985857/88f910c9-989c-4c07-b81d-8c5496316597

joshenlim:
oh whoops! sorry - just pushed to remove an unnecessary check

alaister (commented):
uhh it seems the check might have been somewhat necessary, now the second option is always checked for me (even on brand new projects):
Screenshot 2024-07-11 at 17 14 22
Screenshot 2024-07-11 at 17 14 33

joshenlim:
ahh sorry! managed to repro on a new project - its cause `.split(', ')` in the check wasn't working cause `db_schema` would start off as `public,pg_graphql` instead of `public, pg_graphql`

alaister (approved):
LGTM!

- Tested on preview
- Tried re-enabling and then removing public again

PR #27918, opened 2024-07-11
Support using API schema in create project

joshenlim:
image


alaister (changes requested):
Found a small bug: new project -> use dedicated schema for data api -> only connection string
Screenshot 2024-07-11 at 17 07 12

joshenlim:
thanks so much for the catch! 🙏 🙏  pushed a fix as well with some small UI updates

alaister (approved):
LGTM!

- Tested creating a project with default settings
- Tested creating a project with hardened settings
- Tested creating a project with connection only settings

PR #27922, opened 2024-07-11
Add placeholder in database functions editor

joshenlim:


apps/studio/components/ui/CodeEditor/CodeEditor.tsx
```diff
@@ -61,6 +95,21 @@ const CodeEditor = ({
     if (autofocus) editor?.focus()
   }
 
+  const onChangeContent: OnChange = (value) => {
+    hasValue.current.set((value ?? '').length > 0)
+
+    const placeholderEl = document.getElementById(placeholderId) as HTMLElement | null
```
ivasilov, line 101: Can we use a state variable here instead of the imperative `getElementById`? If Monaco is causing issues with rerenders, you can always use a `useRef` to the div element.

joshenlim: oh yup good point gotcha on this

alaister (commented):
Small bug spotted:
Screenshot 2024-07-12 at 11 31 42

Additionally I think we should only show this placeholder for plpgsql functions (sql functions don't use it)

joshenlim:
> Additionally I think we should only show this placeholder for plpgsql functions (sql functions don't use it)

Good point! pushed a fix for this

> Small bug spotted:

Can I check how you ran into this? I can't seem to reproduce this

ivasilov:
The placeholder only shows up for me on second opening of the sheet.
Video

The placeholder is also shown when editing an existing function
Screenshot 2024-07-12 at 13 46 06

alaister (approved):
LGTM!

- Tested creating and saving a new function
- Tested editing existing functions (sql, plpgsql)

PR #27923, opened 2024-07-11
Fix project status filter

joshenlim:


alaister (approved):
LGTM!

- tested on preview and could see all projects

PR #27930, opened 2024-07-11
feat: Use expanding text area in the Ai Assistant (SQL Editor)

ivasilov:


alaister (approved):
LGTM!

- Tested on preview with multi-line strings

One suggestion might be to add a limit to the max height of the textarea, as currently it's possible to expend it past the height of the screen with enough lines of text

PR #27931, opened 2024-07-11
refactor(docs): type spec parsing + testing setup

charislam:
This produces more complete type information (unions, following type alias refererences that aren't successfully dereferenced at an earlier stage, etc.), in a format that is much easier to query from UI components. Also adds a snapshot test for easier future iteration.

I had a lot of trouble configuring Jest for the snapshot tests (something to do with imports and compile targets), and there are zero other tests in this directory right now, so I just tore out the test setup and replaced it with Vitest. Much faster!


apps/docs/lib/constants.ts
```diff
@@ -1,8 +1,8 @@
 export const API_URL = (
   process.env.NODE_ENV === 'development'
     ? process.env.NEXT_PUBLIC_API_URL || 'http://localhost:8080'
-    : process.env.NEXT_PUBLIC_API_URL!
-).replace(/\/platform$/, '')
+    : process.env.NEXT_PUBLIC_API_URL
+)?.replace(/\/platform$/, '')
```
charislam, line 5: vitest gets stuck on this otherwise

fsansalvadore (approved):
PR seems good and docs preview also looks all good.
Approving to unblock

PR #27935, opened 2024-07-11
Chore: Removed link to Maergo website

mikeridley:
business.

In order to do that, I had to update the slug template to add some conditional logic that only displays the link and icon in the case where the customer_url property is not null.

YES

Small update to Maergo customer story to remove the link to their website. In order to support that, had to add conditional logic to the slug template.

Page links to a non-functional website for a customer that went out of business.

Current template still prints the link icon even in cases where there is no customer_url specified.

## What is the new behavior?

Maergo page no longer has a link to the website.

Template no longer renders the link when customer_url is not set.

## Additional context

Maergo went out of business but it's still a great case study, so we want to keep it.

### DIFF
--- a/apps/www/_customers/maergo.mdx
+++ b/apps/www/_customers/maergo.mdx
@@ -14,7 +14,6 @@ og_image: /images/customers/og/maergo.jpg
 tags:
   - supabase
 date: '2024-06-12'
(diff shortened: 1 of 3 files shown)


fsansalvadore (approved):
Looks good 👍

PR #27937, opened 2024-07-11
fix(docs): footenote links from toc

charislam:
The footnote link/anchor was incorrect because the getAnchor logic doesn't check if the component comes with its own id (which overrides the generated id), and the anchor must match the id. This is the case for footnote headings, which are auto-generated by remarkGfm.


fsansalvadore (approved):
I checked the preview and [this link](https://docs-git-fix-footntes-supabase.vercel.app/docs/guides/platform/access-control#footnote-label) now works properly 👍

PR #27944, opened 2024-07-12
chore: downgrade for usage in sagemaker

thorwebdev:


loong (approved):
for aws supabase hackathon

PR #27985, opened 2024-07-15
chore: update URL for codegen

thebengeu:
Verified that there's no change in the output of `npm run codegen`


alaister (approved):
LGTM!

- Opened all available previews

PR #27999, opened 2024-07-15
Display up-to-date customer balance [GEN-10135]

ecktoteckto:
The goal is to show the latest/up-to-date customer balance in the Dashboard instead of stale data.

Before this PR we fetched a customer's balance via Mgmt-API endpoint `/platform/organizations/{slug}/customer`. This endpoint uses database view `billing_customers` which can contain stale customer data since there's no Orb webhook notifying us about a change in a customer's balance.

The customer's balance is also returned by another endpoint `/platform/organizations/{slug}/billing/subscription`. This endpoint already makes a request to the Orb-API, i.e. we get the latest data.

In the long term we want Orb to provide a respective webhook. For now we go with that quick solution.

There'll be an extra PR for fixing customer balance regarding negative/positive value and cents/dollars.

**Implementation details:**
- my first thought was to hand in the subscription to component `CreditBalance` instead of fetching the subscription from inside the component. But it seemed to be break too much with the approach in many other components (at least in that context). I'm not 100% sure about the caching behavior, but in my local tests there was still only one request towards the Mgmt-API and hence towards the Orb API.

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/CreditBalance.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/CreditBalance.tsx
@@ -8,20 +8,27 @@ import {
 import AlertError from 'components/ui/AlertError'
 import NoPermission from 'components/ui/NoPermission'
 import ShimmeringLoader from 'components/ui/ShimmeringLoader'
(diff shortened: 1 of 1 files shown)


kevcodez (approved):
Tested on staging

PR #28020, opened 2024-07-16
fix: Fix the Sheet closing transition

ivasilov:


alaister (approved):
LGTM!

- Tested by opening a bunch of sheets in the dashboard
- Also opened www + docs as a sanity check

PR #28030, opened 2024-07-17
feat: use TUS for file uploads in dashboard

alaister:

feature/chore

File uploads in the dashboard use the regular supabase-js library, which is unstable for large files or unstable networks.

## What is the new behavior?

- TUS
- Also adds a fallback mime lookup when the browser doesn't know which type to use


joshenlim (approved):
- Uploading a folder via drag and drop
- Uploading several files via drag and drop
- Uploading several files via Upload button
- Verified that new checks work (0 file size + replace logic)

PR #28034, opened 2024-07-17
chore: tus optimisations

alaister:

- Adds `x-source` header for dashboard uploads
- `chunkSize` is based on file size
- Adds `onShouldRetry` to ignore 400 status codes


apps/studio/localStores/storageExplorer/StorageExplorerStore.tsx
```diff
@@ -664,7 +680,13 @@ class StorageExplorerStore {
               objectName: formattedPathToFile,
               ...fileOptions,
             },
-            chunkSize: 6 * 1024 * 1024, // NOTE: it must be set to 6MB
+            chunkSize,
+            onShouldRetry(error) {
+              const status = error.originalResponse ? error.originalResponse.getStatus() : 0
+              const doNotRetryStatuses = [400, 403, 404, 429, 409, 404]
+
+              return doNotRetryStatuses.includes(status)
```
alaister, line 688:
```suggestion
              const doNotRetryStatuses = [400, 403, 404, 409, 429]

              return !(doNotRetryStatuses.includes(status))
```

joshenlim (approved):
LGTM! Although some comments for improvements. Tested on staging preview
- Uploading a folder with nested contents
- Uploading a 1.5 GB file to observe the upload process
- Cancelled uploading when uploading the 1.5 GB file

Some feedback:
-  when uploading multiple small files (e.g a folder) the time estimation seems to be flashing (appears and disappears)
  - i think i understand now, the time estimation is on a per file basis, but in the case of uploading a folder, i’d expect the time estimation to be for the whole upload process
  - also i now realise why i might have ran into that client side error when i uploaded a folder, its probably cause i was uploading multiple files - and the files are so small perhaps when i hit cancel, that file was already uploaded?
  - am thinking what should the “cancel” button behaviour be in the case of uploading a folder, im guessing hitting cancel will just cancel the current file, but continue to the next one
- let’s replace the loading toast with a success when the user hits cancel? “Successfully cancelled file upload”
  - and also a loading indicator when the cancel button was clicked?

PR #28046, opened 2024-07-17
fix(storage): change replaced character to U+202F

charislam:
The character that was breaking screenshots was the Narrow No-Break Space (U+202F), which is apparently not the same thing as the No-Break Space (U+00A0). (TIL there are 17 different Unicode space separators.)

Tested on local by uploading a screenshot taken with default 12-hour timestamp (which was the broken case before).


joshenlim (approved):
LGTM! tested on the preview by uploading a file that Charis sent over which name includes the narrow nbsp

PR #28047, opened 2024-07-17
Fix getAllItemsAlongFolder break

joshenlim:


alaister (approved):
LGTM!

- Tested downloading a folder with >100 nested files
- Tested deleting a folder with >100 nested files

PR #28048, opened 2024-07-17
feat: replace docs command menu with v2

charislam:
Replace docs command menu with Command Menu V2 from `ui-patterns/CommandMenu`.

This introduces the `prepackaged` folder for the shared Command Menu, which contains commands that are shared across the sites (for example, docs search, theme switcher).

Commands that are only applicable to a single site are now defined within that site's own app folder (put them wherever makes sense for your app folder organization, as long as you use the hooks within a global `CommandProvider` it will work...)

Tested navigating around, using search and AI on laptop and mobile.


packages/ui-patterns/CommandMenu/prepackaged/ai/queryAi.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        if (content) {
+          answer += content
+        }
+      } catch (err) {
+        handleError(err)
+      }
+    })
+
+    eventSource.stream()
```
fsansalvadore, line 67: Not sure this is the right place to leave this comment on, but it seems like the ai answers don't come in in a stream fashion in the docs cmdk menu. But rather show up only when the answer is complete. See screen recording.

https://github.com/user-attachments/assets/cb34a45e-76cb-4bc8-9760-e682a8b0734e

packages/ui-patterns/CommandMenu/api/CommandList.tsx
```diff
@@ -29,7 +29,7 @@ const CommandList = forwardRef<
   return (
     <CommandList_Shadcn_
       ref={setRef}
-      className={cn('overflow-y-auto overflow-x-hidden bg-transparent', className)}
+      className={cn('max-h-[initial] overflow-y-auto overflow-x-hidden bg-transparent', className)}
```
fsansalvadore, line 32: Not sure if this is intentional and not sure if this is where it happens, but the width of the modal changes between sections.

https://github.com/user-attachments/assets/b4a3a63b-db34-4494-9ae2-30d8834b85dd

fsansalvadore: And it looks like some additional padding snuck in the ai chat

Screenshot 2024-07-18 at 10 29 53

charislam: The resizing one is intentional (people wanted more room to read the AI responses), but let me take a look at the padding...

jordienr (approved):
Tested in chrome (mac) and safari/chrome(ios) and navigation, scrolling, clicking, and AI chat worked fine!

PR #28064, opened 2024-07-18
fix: new id in content query

alaister:

Bug fix

`enableFolders && id !== 'new' && typeof snapV2.addSnippet === 'function'` `enableFolders` is `undefined` so that expression returns `undefined` instead of `false`, which then passes through to `enabled && typeof projectRef !== 'undefined' && typeof id !== 'undefined'` and now since enabled is `undefined` that whole expression returns `undefined` and when you don’t pass enabled to react-query (also known as `undefined`), it defaults to `true`


joshenlim (approved):

joshenlim (changes requested):
wait

joshenlim (approved):

PR #28071, opened 2024-07-18
fix: use gotrue_id instead of email

kangmingtay:

YES

* Fixes an issue where it is not possible to manage access for an sso user if the sso user has the same email

image

## What is the new behavior?
image

## Additional context
We should also show if a member is an SSO user - but that will be done in a FLUP PR


joshenlim (approved):
- "You" label still shows up correctly on my user in the organization settings team tab

PR #28072, opened 2024-07-18
fix: temporary remove default sorting

MildTomato:

YES

reverts table editor default sorting behaviour 

## Additional context

### DIFF
--- a/apps/studio/data/table-rows/table-rows-query.ts
+++ b/apps/studio/data/table-rows/table-rows-query.ts
@@ -74,16 +74,16 @@ export const fetchAllTableRows = async ({
     })
 
   // If sorts is empty, use the primary key as the default sort
-  if (sorts.length === 0) {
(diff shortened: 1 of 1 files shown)


MildTomato:
[old version for reference](https://github.com/supabase/supabase/commit/d6f9cb53f90d2a9e68f05ce0b2e901715ca7954d#diff-8f7c022d69ec3b549b88c06a3b9738ba29630e2883a1bdaf8e5519a1ebe47579)

joshenlim (approved):
- Created the countries table
- Updated one of the cell values in the table editor
- Updated row gets shifted to the bottom, no longer on same page (verifies that there no longer is a default sort)

ivasilov (approved):
Tested on SQL editor (with a small table). The generated query in the API call is correct. When adding (single and multiple) columns for sorting, the query is correct.

PR #28080, opened 2024-07-19
Set up opt in by slug for PLP

joshenlim:


alaister (approved):
LGTM!

Tested as instructed

> For orgs with no access to PLP (update role, invite, accept invite)
For orgs with access to PLP (update role, invite, accept invite)

PR #28083, opened 2024-07-19
chore: remove mermaid, use static SVG

kevcodez:
Mermaid is a very heavy dependency with lots of transitive dependencies and it not very actively maintained. We're only using it in a single place.

Rather than integrating the entire Mermaid toolchain, we can simply use a static image (exported from the Mermaid code)


ivasilov (approved):
Tested the https://docs-git-nuke-mermaid-supabase.vercel.app/docs/guides/auth/auth-mfa page, shown properly. 

You missed the best reason why we should remove mermaid, there's an active vulnerability that we can't upgrade (latest `mermaid` depends on old vulnerable `puppeteer`).

PR #28084, opened 2024-07-19
Add set search path for query performance query

joshenlim:


ivasilov (approved):
Tested on Query Performance page via an owner account and read-only account. The read only account saw mostly `insuficient privilege` lines. Works as intended.

PR #28095, opened 2024-07-19
Refactor/app router refs

charislam:
Migrates client SDK References to App Router. (Management and CLI API references aren't migrated yet, nor are self-hosting config references.)

Some notes:

- Big changes to the way crawler pages are built and individual section URLs (e.g., `javascript/select`) are served. All of these used to be SSG-generated pages, but the number of heavy pages was just too much to handle -- slow as molasses and my laptop sounded like it was taking off, and CI sometimes refuses to build it all at all.

  Tried various tricks with caching and pre-generating data but no dice.

  So I changed to only building _one copy_ of each SDK+version page, then serving the sub-URLs through a response rewrite. That's for the actual user-visible pages.
  
  For the bot pages, each sub-URL needs to be its own page, but prebuilding it doesn't work, and rendering on demand from React components is too slow (looking for super-fast response here for SEO). Instead I changed to using an API route that serves very minimal, hand-crafted HTML. It looks ugly, but it's purely for the search bots.

  You can test what bots see by running `curl --user-agent "Mozilla/5.0 (iPhone; CPU iPhone OS 6_0 like Mac OS X) AppleWebKit/536.26 (KHTML, like Gecko) Version/6.0 Mobile/10A5376e Safari/8536.25 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)" <URL_OF_PAGE>`
  
  Also added some smoke tests to run against prod for the crawler routes, since we don't keep an eye on those regularly, and Vercel config changes could surprise-break them. Tested the meta images on Open Graph and all seems to work fine.
  
  With this approach, full production builds are _really fast_: ~5 minutes

- Starts using the new type spec handling, which is better at finding params automatically, so I could remove some of the manually written ones from the spec files.


apps/docs/docs/ref/kotlin/v1/introduction.mdx
```diff
@@ -0,0 +1,13 @@
+This reference documents every object and method available in Supabase's Kotlin Multiplatform library, [supabase-kt](https://github.com/supabase-community/supabase-kt). You can use supabase-kt to interact with your Postgres database, listen to database changes, invoke Deno Edge Functions, build login and user management functionality, and manage large files.
```
charislam, line 1: Also copied this from Kotlin V2

apps/docs/docs/ref/swift/v1/installing.mdx
```diff
@@ -0,0 +1,63 @@
+---
```
charislam, line 1: Copied from Swift v2

apps/docs/docs/ref/swift/v1/introduction.mdx
```diff
@@ -0,0 +1 @@
+This reference documents every object and method available in Supabase's Swift library, [supabase-swift](https://github.com/supabase/supabase-swift). You can use supabase-swift to interact with your Postgres database, listen to database changes, invoke Deno Edge Functions, build login and user management functionality, and manage large files.
```
charislam, line 1: Copied from Swift V2

apps/docs/features/docs/Reference.generated.script.ts
```diff
@@ -0,0 +1,136 @@
+import { keyBy, isPlainObject } from 'lodash'
```
charislam, line 1: This script now runs before dev and build, to pre-write the modified type spec and function details to generated files. Was trying to get build time down by not having to repeat this work over and over for each page.

apps/docs/app/api/crawlers/route.smoke.test.ts
```diff
@@ -0,0 +1,72 @@
+import { load } from 'cheerio'
```
saltcod, line 1: great!

apps/docs/features/docs/Reference.generated.script.ts
saltcod: cool, nice!

saltcod:
a lot happening here but seems to do as described!   :^)

saltcod (approved):

PR #28106, opened 2024-07-20
fix:Email validation now checks for period in email format

PLyn:

YES

Update email validation on create new user modal to check for "." in the regex after the @ character.

On this page: https://supabase.com/dashboard/project/_/auth/users

Any email that has the @ is valid. Eg. test@testcom

## What is the new behavior?

Validate for the "." to ensure valid email. Eg. test@test.com for it to be valid


PLyn:
Going to use zod's email validation instead. Will update this soon

alaister (approved):
LGTM!

- Tested locally creating a user with valid and invalid emails
- Also tested with auto confirm on and off

Issue #27921, opened 2024-07-11
Select rows is incorrect when switch pages in Table editor


jingsam:
# Bug report

- [ ] I confirm this is a bug with Supabase, not with my own application. 
- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Select some rows in Table editor, then go to next page, the selected rows is incorrect.

For example, I select one row with id = 1

<img width="548" alt="image" src="https://github.com/supabase/supabase/assets/1522494/819aa068-916d-44ba-ae52-186fd9b5b641">

Then I go to next page, as you see, the id=101 is selected.

<img width="365" alt="image" src="https://github.com/supabase/supabase/assets/1522494/25c2e473-e259-48d9-a795-553621ca9e68">

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

saltcod:
Ideally, we'd deselect any selected rows on page change.

harshsbhat:
This issue is already fixed. Can someone close this

PR #28128, opened 2024-07-22
fix(pagination): Calculate any selectedRows in useMemo and Ensure cancel is called after confirm

sinantalhakosar:

YES

Bug fix

Select some rows in Table editor, then go to next page, since confirmation modal is not being open, the selected rows are incorrect. But since the problem is that state changes are not being reflected in pagination component, it only works problematic if select only 1 row.

https://github.com/supabase/supabase/issues/27921

Also, I noticed that we should call `onCancel` after clicking confirm in the opening confirmation modal

## What is the new behavior?

If there is any selected rows before changing the page, we are showing a confirmation modal now.

## Additional context

- Calculated anySelectedRow exist information from state in useMemo to ensure state changes are reflected when selectedRows.size >= 1.
- Added a call to cancel after confirm in the confirmation modal.


alaister (approved):
LGTM! Thanks again @sinantalhakosar!

PR #28136, opened 2024-07-23
chore: only lint in CI, not on Vercel

kevcodez:
Same as typechecking, we will lint through GH actions, but no longer on Vercel to speed up builds.

Also added the typecheck skips for www


alaister (approved):
LGTM!

- Opened studio, www, docs

PR #28148, opened 2024-07-23
fix: handle inherited tables definitions

leomos:

YES

Bug fix

After creating two simple tables like such:

```sql
create table public.parent (foo text null);
create table public.child (bar text null) inherits (parent);
```

Checking the table definition for `child` errors with a `400` and displays a blank page:

image

## What is the new behavior?

After creating the same tables as above, the definition of the table is displayed correctly:

image

## Additional context

Currently, when displaying the definition of a table that was created as inherited from another table (via the `inherit` keyword), the server errors.

This is due to the fact that in the `CREATE_PG_GET_TABLEDEF_SQL` query the variable `v_colrec` is used when still not initialized.

This commit fixes this by using the correct variable: `in_table`.

Reference: https://github.com/MichaelDBA/pg_get_tabledef/commit/c330a897d6afc710dabd21f1632ad767b3fafaa1#diff-455c3b153a7da564b1a3c4660a53bad68ac681f08c37d2cbd6da76375a36f76dR277

### DIFF
--- a/apps/studio/data/database/database-table-definition.ts
+++ b/apps/studio/data/database/database-table-definition.ts
@@ -206,7 +206,7 @@ export const CREATE_PG_GET_TABLEDEF_SQL = minify(
             WHERE t.table_schema=s.table_schema AND t.table_name=s.table_name AND t.table_schema = in_schema AND t.table_name = in_table AND t.table_type = 'BASE TABLE');      		  
             
             --Issue#19 put double-quotes around SQL keyword column names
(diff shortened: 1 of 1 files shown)


leomos:
I noticed that the resulting query after the fix does not show the added (non-inherited) columns.

This might be solved by updating the whole query to the latest version (from https://github.com/MichaelDBA/pg_get_tabledef/).

Unfortunately when trying to do so, I was facing some other syntax problems that I couldn't fix.

leomos:
I managed to get the latest code from `pg_get_tabledef` working, but it is still not showing the additional columns on the `child` table.

saltcod:
fixed in https://github.com/supabase/supabase/pull/28278

PR #28158, opened 2024-07-24
fix: minor invite flow fixes

alaister:

- redirecting to login if the user isn't logged in yet
- redirecting to projects without a full refresh after accepting an invitation
- removing `OrganizationInviteOld.tsx` – it is no longer needed as the new endpoints support both org/project permissions
- fixing decline button to simply redirect to `/projects` as the user does not have permission to delete invites


phamhieu (dismissed):

jordienr (approved):
✅ 
tested the invitation flow and everything worked

PR #28181, opened 2024-07-25
fix chart widths

jordienr:

YES

- fixes left margin in some charts (it was due to the Y axis being min 60px in width even when its empty)
- fixes the chart in sql editor not fitting the viewport


alaister (approved):
LGTM!

- Tested on preview by viewing some charts in sql editor

I'm guessing "some charts" doesn't include these ones?
Screenshot 2024-07-26 at 11 20 21

jordienr:
> I'm guessing "some charts" doesn't include these ones?
oops yeah, fixing that

Issue #27614, opened 2024-06-28
OAuth breaks when using provided auth/callback example in docs when using custom domain


saendu:
# Improve documentation
Use X-Forwarded-Host header to not get redirected to ``http://localhost:8080``
add something like tis code to the docs: 
```javascript
...
const { error } = await supabase.auth.exchangeCodeForSession(code)
    if (!error) {
      const forwardedHost = request.headers.get("x-forwarded-host") // original origin before load balancer
      const isLocalEnv = process.env.NODE_ENV === "development"
      if (isLocalEnv) {
        // we can be sure that there is no load balancer in between, so no need to watch for X-Forwarded-Host
        return NextResponse.redirect(`${origin}${next}`)
      } else if (forwardedHost) {
        return NextResponse.redirect(`https://${forwardedHost}${next}`)
      } else {
        return NextResponse.redirect(`${origin}${next}`)
      }
    }
...
```

## Link 
https://supabase.com/docs/guides/auth/social-login/auth-azure#add-login-code-to-your-client-app

## Describe the problem

When using a custom domain and Azure Oauth provider, the provided /auth/callback example in the docs is not working, since the {origin} will be http://localhost:8080. 
Use instead x-forwarded-host header. See code above. 

## Describe the improvement

A clear and concise description of the improvement.

## Additional context
Probabbly only a problem when using a custom domain where you get a dedicated instance.

encima:
Thanks for opening this! Feel free to open a PR to add this info :)

PRIZZA21:
Hello, I am interested in Supabase project and would like to make contributions. I think I can start with this. .Can I work on it ? Thanks.

a4vg:
Hi @PRIZZA21, have you started with it? I was planning on creating a PR as well.

PR #28183, opened 2024-07-25
fix: update next.js oauth pkce flow docs

aryan-mehrabi:

YES

fix for issue #27614 

#27614 

As described in the issue: 

> When using a custom domain and Azure Oauth provider, the provided /auth/callback example in the docs is not working, since the {origin} will be http://localhost:8080/.

## What is the new behavior?

Oauth section in the docs got updated as the following:

```js
const { error } = await supabase.auth.exchangeCodeForSession(code)
    if (!error) {
      const forwardedHost = request.headers.get("x-forwarded-host") // original origin before load balancer
      const isLocalEnv = process.env.NODE_ENV === "development"
      if (isLocalEnv) {
        // we can be sure that there is no load balancer in between, so no need to watch for X-Forwarded-Host
        return NextResponse.redirect(`${origin}${next}`)
      } else if (forwardedHost) {
        return NextResponse.redirect(`https://${forwardedHost}${next}`)
      } else {
        return NextResponse.redirect(`${origin}${next}`)
      }
    }
```

## Additional context

[Doc Link](https://supabase.com/docs/guides/auth/social-login/auth-azure#add-login-code-to-your-client-app)

Note: as the oauth pkce flow file get updated, it is expected that other pages that uses this file gets affected.

### DIFF
--- a/apps/docs/components/MDX/oauth_pkce_flow.mdx
+++ b/apps/docs/components/MDX/oauth_pkce_flow.mdx
@@ -71,7 +71,16 @@ export async function GET(request: Request) {
     const supabase = createClient()
     const { error } = await supabase.auth.exchangeCodeForSession(code)
     if (!error) {
-      return NextResponse.redirect(`${origin}${next}`)
+      const forwardedHost = request.headers.get('x-forwarded-host') // original origin before load balancer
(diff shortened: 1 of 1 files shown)


charislam (approved):
Thanks @aryan-mehrabi!

PR #28208, opened 2024-07-26
feat(orb): support new usage stats

kevcodez:
- Add new metric Disk Size GB-Hrs (only exposed to paid plans on Orb)
- Adjusted some texts as preparation of new billing plans
- Fixed some visual bugs with the usage summary grid
- Project allocations in tooltips are now sorted by usage (descending)
- Project allocations with too little usage (<0.01) now display "<0.01" instead of just "0"
- Usage formatter supports usage that is broken down to hours
- Stop showing percentages when the usage is 0 (previously showed "<1%"

This change is backwards compatible and can be merged. Note that there will be full overhaul in the feature of the entire usage page, this is just the least invasive change to make the transition work.


alaister (approved):
LGTM!

- Checked billing and usage pages on preview build

PR #28209, opened 2024-07-26
chore: official supabase-py

grdsdev:

Remove community tag from supabase-py, and mark it as official.


charislam (approved):
Check: opened all previews

PR #28241, opened 2024-07-28
Update Supabase Bootstrap NextJS example to use new SSR methods

PLyn:

YES

Update NextJS cookie-based Auth example that Supabase BootStrap uses. Now uses the new SSR methods and server actions.

Uses new SSR package but still using Deprecated methods which still work for now.

## What is the new behavior?

Uses the new methods (GetAll, SetAll etc) and use server actions for login and signup.

## Additional context

Also updated package.json to use the new 0.4.1 version

### DIFF
--- a/examples/auth/nextjs/app/auth/callback/route.ts
+++ b/examples/auth/nextjs/app/auth/callback/route.ts
@@ -1,6 +1,5 @@
 import { type EmailOtpType } from '@supabase/supabase-js'
 import { type NextRequest, NextResponse } from 'next/server'
-
 import { createClient } from '@/utils/supabase/server'
 import { redirect } from 'next/navigation'
 
--- a/examples/auth/nextjs/app/auth/callback/route.ts
+++ b/examples/auth/nextjs/app/auth/callback/route.ts
@@ -1,19 +1,28 @@
-import { createClient } from "@/utils/supabase/server";
-import { NextResponse } from "next/server";
+import { type EmailOtpType } from '@supabase/supabase-js'
+import { type NextRequest, NextResponse } from 'next/server'
 
-export async function GET(request: Request) {
-  // The `/auth/callback` route is required for the server-side auth flow implemented
-  // by the SSR package. It exchanges an auth code for the user's session.
-  // https://supabase.com/docs/guides/auth/server-side/nextjs
-  const requestUrl = new URL(request.url);
-  const code = requestUrl.searchParams.get("code");
-  const origin = requestUrl.origin;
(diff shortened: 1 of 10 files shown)


charislam (approved):

ssudaraka:
Hi @PLyn,

1. In the callback route (`/app/auth/callback/route.ts`), the current implementation is correct. It should look for the `code` in the query parameter and exchange it for a session to support the PKCE flow. It looks like this part is missing in our [Next.js guide](https://supabase.com/docs/guides/auth/server-side/nextjs?queryGroups=router&router=app).

    What you have added to the callback route should go to the email confirmation route (`/app/auth/confirm/route.ts`). It's for redirecting the user when they click the link on their confirmation email.

2. We should keep the versions for `@supabase/ssr` and `@supabase/supabase-js` packages to the `latest` in `package.json`.

3. Currently, the example code uses double quotes in import statements and strings. Let's keep that code style intact.

Other changes look good to me. Thanks for moving the actions to their own page.

examples/auth/nextjs/app/auth/callback/route.ts
ssudaraka, line 1: This path is used to exchange the code in the PKCE flow. The current implementation is correct.

charislam: IIRC both are correct, depending on how you set up your confirmation emails (whether you send the code or the token hash). We lean towards recommending the token hash these days because it removes the same-browser limitation. We should probably document that somewhere though (step 6 here: https://supabase.com/docs/guides/auth/server-side/nextjs?queryGroups=router&router=app), maybe in the readme?

ssudaraka: Ah, you're correct. Since this example shows the email auth flow, it only needs to have the email confirmation route. I confused it with the PKCE flow we had earlier that we use for Social auth and SAML auth.

It's already documented with the relevant social auth flows ([Google auth](https://supabase.com/docs/guides/auth/social-login/auth-google?queryGroups=platform&platform=web&queryGroups=environment&environment=client&queryGroups=framework&framework=nextjs#signing-users-in), [Facebook auth](https://supabase.com/docs/guides/auth/social-login/auth-facebook?queryGroups=language&language=js&queryGroups=environment&environment=client&queryGroups=framework&framework=nextjs)) and we probably need to mention it somewhere in SAML docs too.

@PLyn, sorry about the confusion. Not having the callback route in this example should be fine.

PLyn: No problem. Thanks both of you!

PR #28248, opened 2024-07-29
Add WAL disk usage to Disk Usage docs

encima:

YES

docs update

Queries and descriptions for database size

## What is the new behavior?

Additional info to see how much disk space is consumed by WAL and what it does.

## Additional context

Could potentially add default values here and/or information on when WAL is cleared (clarify that this cannot be done manually)


apps/docs/content/guides/platform/database-size.mdx
```diff
@@ -31,6 +31,17 @@ Depending on your billing plan, your database can go into read-only mode which c
 
 </Admonition>
 
+### Disk space usage
+
+Your database size is part of the disk usage for your Supabase project, there are many components to Postgres that consume additional disk space. One of the primary components, is the [Write Ahead Log (WAL)](https://www.postgresql.org/docs/current/wal-intro.html).
+Postgres will store database changes in log files that are cleared away after they are applied to the database. These same files are also used by [Read Replicas](/docs/guides/platform/read-replicas) or other replication methods.
```
charislam, line 37:
```suggestion
Your database size is part of the disk usage for your Supabase project, there are many components to Postgres that consume additional disk space. One of the primary components, is the [Write Ahead Log (WAL)](https://www.postgresql.org/docs/current/wal-intro.html). Postgres will store database changes in log files that are cleared away after they are applied to the database. These same files are also used by [Read Replicas](/docs/guides/platform/read-replicas) or other replication methods.
```

If these are meant to be separate paragraphs, add a blank line, otherwise don't put a line break. (I know it's standard-ish Markdown that line breaks disappear but we use MDX so they don't and it displays weirdly.)

charislam (approved):

PR #28279, opened 2024-07-30
fix: sign up redirect

alaister:

Bug fix

`/sign-up` and `/sign-in-sso` are inaccessible via direct link


jordienr (approved):

ivasilov (approved):
Tested with Github login (with and without MFA), SSO login. All work ok.

PR #28282, opened 2024-07-30
fix/ deep-dark theme name issue 

MildTomato:

YES/NO

## What is the new behavior?

## Additional context

### DIFF
--- a/packages/common/Providers.tsx
+++ b/packages/common/Providers.tsx
@@ -7,7 +7,12 @@ export function ThemeProvider({ children, ...props }: ThemeProviderProps) {
   // @ts-ignore next-themes is old :/
 
   return (
-    <NextThemesProvider {...props} themes={['dark', 'light']}>
+    <NextThemesProvider
+      {...props}
+      themes={['dark', 'light']}
+      defaultTheme="dark"
+      storageKey={'supabase-theme'}
+    >
       {children}
     </NextThemesProvider>
   )
--- a/packages/common/Providers.tsx
+++ b/packages/common/Providers.tsx
@@ -7,7 +7,7 @@ export function ThemeProvider({ children, ...props }: ThemeProviderProps) {
   // @ts-ignore next-themes is old :/
 
   return (
-    <NextThemesProvider {...props} themes={['dark', 'light', 'deep-dark']}>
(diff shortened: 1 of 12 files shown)


packages/common/Providers.tsx
```diff
@@ -7,7 +7,7 @@ export function ThemeProvider({ children, ...props }: ThemeProviderProps) {
   // @ts-ignore next-themes is old :/
 
   return (
-    <NextThemesProvider {...props} themes={['dark', 'light', 'deep-dark']}>
+    <NextThemesProvider {...props} themes={['dark', 'light']}>
```
ivasilov, line 10: the `themes` prop overrides any themes passed in via props. Should the unstructured props be at the end?

ivasilov (changes requested):
This PR doesn't solve the issue if the user has `deep-dark` in the localstorage value.

packages/common/Providers.tsx
MildTomato: we don't use 'classic-dark' on anything but Studio, so we override it in studio

packages/common/Providers.tsx
```diff
@@ -7,7 +7,12 @@ export function ThemeProvider({ children, ...props }: ThemeProviderProps) {
   // @ts-ignore next-themes is old :/
 
   return (
-    <NextThemesProvider {...props} themes={['dark', 'light', 'deep-dark']}>
+    <NextThemesProvider
+      {...props}
+      themes={['dark', 'light']}
+      defaultTheme="dark"
+      storageKey={'supabase-theme'}
```
MildTomato, line 14: @ivasilov this should fix it

MildTomato: > This PR doesn't solve the issue if the user has deep-dark in the localstorage value.

what about now ?

ivasilov (approved):

packages/ui/build/css/themes/dark-combined.css
```diff
@@ -1,3 +1,5 @@
+[data-theme='deep-dark'],
+.deep-dark,
```
MildTomato, line 2: @ivasilov workaround without losing the saved themes for those unaffected

ivasilov (approved):

PR #28287, opened 2024-07-31
fix: remove unnecessary condition in MemberRow

alaister:

Bug fix

If a user has 0 projects their role will not show

## What is the new behavior?

It always shows


joshenlim (approved):
LGTM! Tested locally
- For an org that has access to project level permissions, and has 0 projects

PR #28304, opened 2024-07-31
feat: local docs search

charislam:
Improve the speed of search by running it locally for users with fast connections. Search indexes are copied from the remote database into a PGLite instance running in a Shared Worker.

Requires using Transformers.js embeddings to also cut out the OpenAI call.

To avoid large data transfers for users on slow or data-limited connections, the conditions for using local search are:
- Effective data connection speed of 4g
- User has not requested data saving

Because the browser APIs required to check these conditions are only exposed in Chromium, local search is not available in Firefox and Safari. In these browsers (and before the worker is ready in Chromium), the existing remote search is used as a fallback.

One side-effect: `/api/ai/docs` routes had to be moved to App Router + Node Runtime. Vercel always includes the > 2 MB PGLite module in the server bundle, even though the edge routes don't need it, which causes it to exceed the Edge Runtime size limit. I tried the experimental `outputFileTracingExcludes` but couldn't get it to work.

## Testing

Test on Chrome and one other browser (which will cover local search and fallback remote search). Note that on Chrome remote search is still used as a fallback until the worker is ready. To make sure you're actually testing local search, open the console and wait until you see the worker `READY` event logged. You can also run a search with the Network tab open and make sure no network requests are made.

(Keep in mind that first startup even for remote search may be a little slow on preview due to it running on a branch DB. I'd prewarm things by running a first search and letting the loading spinner spin, then hard refreshing and starting over.)


packages/ui-patterns/CommandMenu/prepackaged/DocsSearchLocal/DocsSearchLocal.worker.ts
```diff
@@ -0,0 +1,127 @@
...
+        import.meta.url
+      ),
+    },
+  })
+  await db.exec(CREATE_VECTOR_EXTENSION)
+  await db.exec(CREATE_PAGE_TABLE)
+  await db.exec(CREATE_PAGE_SECTION_TABLE)
+
+  const supabase = createClient(supabaseUrl, supabaseAnonKey)
+  // Preload the extraction pipeline before signalling that the worker is
```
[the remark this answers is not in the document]
charislam: 🇨🇦 (jk)

apps/docs/scripts/search/generate-embeddings.ts
```diff
@@ -3,6 +3,9 @@ import dotenv from 'dotenv'
 import { parseArgs } from 'node:util'
 import { OpenAI } from 'openai'
 import { v4 as uuidv4 } from 'uuid'
+
+import { setupSingletonExtractor } from 'ui-patterns/CommandMenu/prepackaged/DocsSearchLocal/DocsSearchLocal.shared.llm'
```
saltcod, line 7: 1) what

charislam: that is ... a ridiculous place to put a shared util, is it not? 😂 moving...

apps/docs/scripts/search/generate-embeddings.ts
```diff
@@ -185,6 +197,7 @@ async function generateEmbeddings() {
               content,
               token_count: embeddingResponse.usage.total_tokens,
               embedding: responseData.embedding,
+              hf_embedding: featureArray,
```
gregnr, line 200: I wonder if it would be better to name embedding columns by their model vs platform? Could give us more flexibility in the future. What do you think?

packages/ui-patterns/CommandMenu/prepackaged/DocsSearchLocal/DocsSearchLocal.shared.llm.ts
```diff
@@ -0,0 +1,27 @@
+import type { Tensor } from '@xenova/transformers'
+
+const MODEL = 'Supabase/gte-small'
+
+export interface Extractor {
+  extract: (input: string) => Promise<Tensor>
+}
+
+export function setupSingletonExtractor() {
```
gregnr, line 9: Our docs and most places will call this `embed()` vs `extract()` - maybe we should use similar terminology.

With respect to the singleton pattern - it might be clearer to just rename `setupSingletonExtractor()` itself to just `embed()` which still uses a singleton pattern but hides the details ([example](https://github.com/supabase-community/postgres-new/blob/358ad8b66db774b85fd8c1ee9a688cd623b58118/apps/postgres-new/lib/embed/index.ts#L6-L31)). Might be easier for new devs to understand when they see it for the first time.

gregnr: Also worth considering - inference is somewhat CPU intensive so it's usually better to offload to Web Worker so that the main thread/UI isn't blocked. You can use [`comlink`](https://github.com/GoogleChromeLabs/comlink) to simplify communication with the web worker.

Example of how we do this in `postgres-new`:
https://github.com/supabase-community/postgres-new/blob/main/apps/postgres-new/lib/embed/index.ts

---

**Edit:** I see now that you might only be calling this from a Web Worker already. If so this might not be relevant.

packages/ui-patterns/CommandMenu/prepackaged/DocsSearchLocal/DocsSearchLocal.worker.sql.ts
```diff
@@ -0,0 +1,124 @@
...
+			group by page.id
+			order by match_score desc
+			limit 20
+		)
+	)
+)
+where disambiguator = 1
+order by match_score desc
+limit 10
+;
```
gregnr, line 123: Worth sticking with 2 space indents vs. tabs?

packages/ui-patterns/CommandMenu/prepackaged/DocsSearchLocal/DocsSearchLocal.worker.sql.ts
```diff
@@ -0,0 +1,124 @@
...
+				'{}' as headings,
+				'{}' as slugs,
+				'fts' as match_type,
+				least(1,
+					-- Weighting factor determined by trial and error
+					ts_rank('{0.01, 0.1, 0.2, 1.0}', page.tsv, websearch_to_tsquery('simple', $3), 1) * 100
+				) as match_score
+			from page
+			join page_section on page_section.page_id = page.id
+			where page.tsv @@ websearch_to_tsquery('simple', $3)
```
gregnr, line 113: PGlite fixed FTS issues as of v0.2.2 so no more need for `simple`

gregnr (commented):
Nice @charislam 🎉 this is great work and will be really cool to have in docs. A few thoughts/questions:
- [`comlink`](https://github.com/GoogleChromeLabs/comlink) simplifies Web Worker communication quite a bit. I see you already have a bunch of Web Worker communication infra in place, but the benefits of simplicity might be worth the switch. Some examples of how we've done that [here](https://github.com/supabase-community/postgres-new/blob/main/apps/postgres-new/lib/embed/index.ts) and [here](https://github.com/supabase-community/postgres-new/blob/main/apps/postgres-new/lib/db/index.ts).
- Is `DocsSearchLocal` its own component/UI? I may have missed something but have we considered reusing the same component for both remote/local search and just swap out the search mechanism?
- Big one: ElectricSQL's [sync engine](https://next.electric-sql.com/product/electric) is actually designed for this exact use case (pulling in remote data to local DB). When we add new docs in the future, this can transparently pull just the changes from the remote DB. What do you think about using something like this?

PR #28318, opened 2024-08-01
fix(self-hosted): postgrest config route location

alaister:

Bug fix

Currently the `/config/postgrest` route is in the wrong location.

## What is the new behavior?

Moves it under the `/projects/[ref]` folder where it's expected to be.


saltcod (approved):

saltcod:
checked preview

PR #28326, opened 2024-08-01
fix(docs): search button on not-found page

charislam:
I missed a place when swapping the old command menu out for the new one.


saltcod (approved):

saltcod:
checked preview

PR #28327, opened 2024-08-01
ci: remove migration push from search action

charislam:
We don't need this for migration pushes since the Supabase integration will do it instead. Removing it makes this action less brittle since embedding updates get blocked on CLI problems.

Test: successful run here https://github.com/supabase/supabase/actions/runs/10203661699/job/28230464361


alaister (approved):
LGTM!

PR #28328, opened 2024-08-01
fix(docs): duplicate heading ids

charislam:
Duplicate heading IDs interferes with the Table of Contents highlighting, so add custom unique IDs for those sections.


loong (approved):
👍

PR #28332, opened 2024-08-02
Add an Authenticator Assurance Level toggle for impersonating users in Studio

sleepdotexe:

YES

New feature

Currently in Studio, when using the RoleImpersonationPopover to impersonate a specific user, the JWT that is created is locked to AAL1.

Given [supabase recommends adding RLS policies to ensure tables are protected by MFA](https://supabase.com/docs/guides/auth/auth-mfa#enforce-rules-for-mfa-logins) where the expected JWT AAL is AAL2, this makes it effectively impossible to inspect or debug tables and policies using the user impersonator since no rows will ever be returned.

There doesn't appear to be an easy way within the studio to "upgrade" an impersonated user to AAL2.

See [this discussion](https://github.com/orgs/supabase/discussions/27986).

## What is the new behavior?

With this change, a toggle has been added into the user selector within the ImpersonationPopover to allow the user to specify whether they want the impersonated user to be at AAL1 or AAL2.

**Impersonate a user at AAL1**:
image

**Impersonate a user at AAL2**:
image

The AAL level can only be set before the JWT is created, so the toggle is placed inside the conditionally-rendered user selector that only displays if no currently-impersonated user exists.

**Currently impersonating a user at AAL2**:
image

**Note** that the AAL level shows in the impersonated user row as well as the RoleImpersonationPopover trigger.

The AAL option for impersonated roles is defined in the `type PostgrestImpersonationRole` definition from `lib/role-impersonation.ts`, where `role` is `authenticated`. This makes it automatically accessible wherever an impersonated role with role type of `authenticated` is used.

Since it's technically a boolean state between AAL1 and AAL2, I have used a switch in the UI but put labels on each side of the switch to make it very clear which AAL is being selected (instead of only having 1 label which could provide a confusing "on/off" state; ie. having a switch between AAL1 <-> AAL2 is more clear than just having AAL2 be on or off).

I have hard-coded the display names as "AAL1" and "AAL2" instead of doing something like `aal.toUpperCase()` in case they change in future, or we want to override the display names with some custom label. I also added a `type AuthenticatorAssuranceLevels` definition as I couldn't find anywhere else where this was already specified, but I'm assuming there should be a source of truth somewhere for what the values for `aal` within the JWT can be?

## Additional context

- [x] Ran `npm run build:studio` successfully!
- [x] Ran `test:prettier` successfully!

I tested that this new feature worked locally using the following steps:

1. Run a local copy of supabase Studio using `npm run dev:studio` and `docker compose up`.
2. Create a new user on the Authentication page.
3. Add a new table and enable RLS:
```sql
create table "public"."profiles" (
    "id" bigint generated by default as identity not null,
    "user_id" uuid not null
);

alter table "public"."profiles" enable row level security;

CREATE UNIQUE INDEX profiles_pkey ON public.profiles USING btree (id);
alter table "public"."profiles" add constraint "profiles_pkey" PRIMARY KEY using index "profiles_pkey";
alter table "public"."profiles" add constraint "profiles_user_id_fkey" FOREIGN KEY (user_id) REFERENCES auth.users(id) ON UPDATE CASCADE ON DELETE CASCADE not valid;
alter table "public"."profiles" validate constraint "profiles_user_id_fkey";
```
4. Add a row to the table for the newly created user.
```sql
insert into public.profiles (user_id) VALUES ('<your user id here>');
```
5. Inspect the table impersonating the created user, with the toggle at either AAL. At this point we expect to see no rows since our table has enabled RLS with no policies.

image

6. Add an RLS policy to the table that allows that user to view their own rows:
```sql
create policy "Users can access their own rows"
on "public"."profiles"
as permissive
to public
using (
    (SELECT auth.uid()) = user_id
);
```
7. Inspect the table impersonating the created user, with the toggle at AAL1. At this point we expect to see 1 row since our table has an RLS policy allowing us to view rows.

image

8. Add an RLS policy to the table that forces users to be AAL2:
```sql
create policy "Enforce MFA for all end users"
on "public"."profiles"
as restrictive
to authenticated
using (
    ((( SELECT auth.jwt() AS jwt) ->> 'aal'::text) = 'aal2'::text)
);
```
9. Inspect the table impersonating the created user, with the toggle at AAL1. At this point we expect to see no rows since our restrictive policy will deny AAL1 users.

image

10. Inspect the table impersonating the created user, with the toggle at AAL2. At this point we expect to see our row again, since the created JWT should be at AAL2 which is permitted by our RLS policy.

image

---
Personally, since I work on applications with MFA enabled, this would seriously improve my DX and productivity within the supabase studio if this was merged. First time contributing, so feedback is welcome.

Thank you!


alaister (approved):
LGTM! Thanks for the super detailed description @sleepdotexe – made testing a delight!

JFYI, I made a couple of small additions:
- fixed a small light mode text colour issue
- adds "Advanced options" collapsible
Screenshot 2024-08-02 at 15 37 57

- adds info tooltip linking to MFA docs
Screenshot 2024-08-02 at 15 37 40

sleepdotexe:
Hi @alaister – sorry to bring up an old thread, but I was just wondering if these changes would be coming to Studio soon for local dev environments (ie. Docker)? I'm on the latest CLI version 1.200.3 but even in newly initialized projects I can't seem to see this option as being available. Or am I missing a way to update my local instances?

Given most of my work and testing is done on local, it'd be great to have access to this feature so I don't have to keep removing my AAL RLS policies locally.

PR #22793, opened 2024-04-16
docs: add error codes auth

hf:
Adds docs about Auth's error codes.

PR #28334, opened 2024-08-02
docs: Remove error codes section from Flutter, Swift, and Python reference docs.

### LINKED ISSUE #28320: Copy/Paste problem in the auth error codes documentation
# Improve documentation

## Link 

https://supabase.com/docs/reference/dart/auth-error-codes

## Describe the problem

The documentation describes the way AuthError works for the javascript client, and not for other clients. It's just a copy/paste of the javascript description. This is confusing because the names aren't the same, and the way it works seems to be different too.
For example, it is said that the error class "AuthError" would wrap all different errors, but in dart the class is not "AuthError" but "AuthException".
Also, i can't find the error code in the exception class, or the "isAuthApiError" function, for example.

## Describe the improvement

An improvement would be to document the way error management works for each type of client.

dshukertjr:

YES

- Updates the reference docs to remove the error codes section on certain libraries until they are implemented.
- Removes the `meta.shared` property on the `common-client-libs-sections.json` file introduced in https://github.com/supabase/supabase/pull/22793, because 1. it was only used for the error codes, 2. it was tricky to only display them on certain client libraries. 
- Created a shared auth-error-codes-table component to display the list of error codes, while allowing the content out side of the table to be editable per client, so that the descriptions could be adjusted to each client libraries' lingo

Closes https://github.com/supabase/supabase/issues/28320#issuecomment-2263612334


dshukertjr:
@guiguicdd @juancarlospaco I looked at the code of our Swift and Python libraries and didn't find anything that seemed like the auth error codes, so I'm hiding the auth error code section for those two as well, but let me know if I'm wrong!

apps/docs/components/Navigation/NavigationMenu/NavigationMenuRefList.tsx
```diff
@@ -1,5 +1,6 @@
 import { useCommonSections, useSpec } from './NavigationMenu.utils'
 import NavigationMenuRefListItems from './NavigationMenuRefListItems'
+import { ICommonCategory } from '~/components/reference/Reference.types'
```
dshukertjr, line 3:
```suggestion
```

apps/docs/components/Navigation/NavigationMenu/NavigationMenuRefList.tsx
```diff
@@ -28,6 +28,12 @@ const NavigationMenuRefList = ({
   }
 
   const filteredSections = commonSections.filter((section) => {
+    if (section.type === 'category') {
+      section.items = section.items.filter((item) => {
+        return !item.excludes?.includes(id)
+      })
+    }
+
```
dshukertjr, line 36: This allows items inside a section to be removed from the side nav when `exclude` is used.

charislam (approved):
Thank you! ❤️

PR #28336, opened 2024-08-02
fix: LW ticket site auth state conflict with studio

alaister:

Bug fix

When a logged in dashboard user gets a LW ticket, they get logged out of the dashboard

## What is the new behavior?

This is somewhat guessing at the root error but I think what's happening is:

1. User gets redirected back to www site after clicking claim ticket
2. The two Supabase clients both try and grab the tokens from the URL (since `detectSessionInUrl` is default `true`)
3. The new tokens are invalid for use with the dashboard app, so auth-js removes them (even if there were existing ones)

This may not be the root issue, but I can see no reason why www would need to log a user in for the dashboard (only needs to check whether they are already logged in), so I've added `NEXT_PUBLIC_AUTH_DETECT_SESSION_IN_URL = false` in Vercel for www.


hf:
Sounds sane.

ivasilov (approved):

PR #28345, opened 2024-08-02
fix(docs): hydration error on js references

charislam:
Our syntax highlighter doesn't recognize HTML (tried to install it but it's not on the list of supported languages) and this somehow causes the server and client to make different highlighting decisions, with the server attempting to highlight the code block and the client not.

Removing the html language specifier removes the hydration error (and since the client never properly highlighted it anyway, we don't lose much in terms of syntax highlighting.)


fsansalvadore (approved):
That seems to fix it 👍

PR #28347, opened 2024-08-02
Chore/rls nudge

saltcod:
https://supabase.slack.com/archives/C0161K73J1J/p1722538753077259

change the default for this checkbox from false to true
screenshot-2024-08-02-at-16 28 31


joshenlim:
Found a bug - check expression should not show up for SELECT (which is the default command when creating a new policy)
image

there's a couple of unrelated UI bugs noticed but we can fix them separately
  - When changing the command back to SELECT, the placeholder no longer shows up cause there's some white spaces
  - There's a bit of a shift in UI when the CHECK expression shows up

https://github.com/user-attachments/assets/ea16293e-d050-43fb-800a-13c7654a4546

joshenlim:
ahh for point 2, here's what i meant:
https://github.com/user-attachments/assets/f802823f-1bda-4072-9f75-d1b8a8a874e9

but i think no worries about this for now - its also an existing problem on prod i realised so we can tackle this separately in another PR

joshenlim:
so sorry one last bug found
- Select "UPDATE" command doesnt check the "Use check expression" block by default (like "ALL" command)

joshenlim:
LGTM! Tested on the staging preview
- Creating a policy -> select "ALL" / "UPDATE" command from "SELECT", with check block shows up by default
  - If I uncheck "use check expression", then change the command to "SELECT" then to "ALL" again, with check block will show up as expected
- Editing a policy with the ALL command, with check block shows up
- Editing a policy with SELECT command, with check block doesnt show up

joshenlim (approved):

PR #28383, opened 2024-08-05
Open folder of snippet when opening a snippet on fresh browser session

joshenlim:
Related to SQL folders: 
- Open the folder which the snippet belongs to when opening a snippet on a fresh browser session
  - Only if the snippet belongs to a folder
  - And also fetch the contents of the folder


alaister (approved):
LGTM!

- Refreshed while having a snippet in a folder selected
- Also clicked around the UI

PR #28385, opened 2024-08-05
Chore/fix autodocs stored procedure menu

joshenlim:
- Fix Autodocs V2 stored procedure client crash when clicking on a store procedure that returns a TABLE type
  - Fix was just referencing how we obtained the parameters from Autodocs V1 which was through the post property rather than get
- Fix Autodocs V1 stored procedures menu not highlighting selected stored procedure
- Replace Database functions CTA buttons with ButtonTooltip component
  - Tooltip wasn’t showing up due to legacy components


alaister (approved):
LGTM!

- Clicked around auto docs V2

PR #28388, opened 2024-08-05
fix: upload data during creation only for files above the chunk size

fenos:

YES

Improvement

## What is the new behaviour?

With this PR we only upload data during the Upload URL creation for files equal to or lower than the defined chunk size.
This will improve DX as well as the calculation of the elapsed time.


alaister (approved):
LGTM!

- tested by uploading files above and below 6mb

PR #28395, opened 2024-08-05
fix: Disable sorting by json/jsonb columns

ivasilov:
This PR removes the json/jsonb columns from the sort popover. Json/jsonb columns can only be used in a order by clause if you also specify a key in the JSON object.


saltcod (approved):

saltcod:
checked preview

PR #28397, opened 2024-08-05
fix: fix the invoice status label tooltip for invoices in status 'issued' [GEN-10357]

ecktoteckto:

The status label tooltip for an invoice in status "Issued" does not contain any text

gen-10357-before

## What is the new behavior?

The status label tooltip for an invoice in status "Issued" does contain text depending on whether there was a payment attempt or not.

**Payment Attempt**

gen-10357-after-payment-attempt

**No Payment Attempt**

gen-10357-after-no-payment-attempt

## Additional context
- we decided on not introducing a new label but instead go with the already existing label "Outstanding" in case of an invoice in status "Issued" with no payment attempts. We don't want to overcomplicate things. "Outstanding" is not "Overdue".
- theoretically we could also take into consideration whether "auto-collection" is activated for an invoice or not. But again, let's not overcomplicate things and only a very small set of users don't have "auto-collection" enabled.

### DIFF
--- a/apps/studio/components/interfaces/Billing/InvoiceStatusBadge.tsx
+++ b/apps/studio/components/interfaces/Billing/InvoiceStatusBadge.tsx
@@ -5,6 +5,7 @@ import { InvoiceStatus } from './Invoices.types'
 
 interface InvoiceStatusBadgeProps {
   status: InvoiceStatus
+  paymentAttempted: boolean
 }
 
 const invoiceStatusMapping: Record<
@@ -40,7 +41,7 @@ const invoiceStatusMapping: Record<
   },
 }
 
-const InvoiceStatusBadge = ({ status }: InvoiceStatusBadgeProps) => {
+const InvoiceStatusBadge = ({ status, paymentAttempted }: InvoiceStatusBadgeProps) => {
   const statusMapping = invoiceStatusMapping[status]
 
   return (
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Billing/InvoiceStatusBadge.tsx
```diff
@@ -63,13 +64,21 @@ const InvoiceStatusBadge = ({ status }: InvoiceStatusBadgeProps) => {
...
+            ) &&
+              (paymentAttempted ? (
+                <p className="text-xs text-foreground">
+                  We were not able to collect the money. Make sure you have a valid payment method
+                  and enough funds. Outstanding invoices may cause restrictions. You can manually
+                  pay the invoice using the "Pay Now" button.
+                </p>
+              ) : (
+                <p className="text-xs text-foreground">
+                  The invoice will soon be charged for. You can manually pay the invoice using the
```
kevcodez, line 78: For a subscription that uses payment via invoice (no auto collection) but hasn't done the payment yet, but is beyond the due date, this wouldn't be true as we'd never charge for that invoice

ecktoteckto: The text is a bit off even if the payment via invoice is not overdue since we don't actively charge but rely on the customer's action. But back to your actual point. So this kind of invoice (non auto-collection, overdue) is not picked up by our dunning process at some point?

Maybe we should expose the info "auto_collection true/false" via the Mgmt-API rather sooner than later so that we are able to provide spot on texts here. Also to avoid potential support cases.

Maybe for now we can go with something along the lines of

> "The invoice will soon be charged for. In case invoice auto collection is not enabled for your organization, please make sure to make the payment in a timely manner via the link in our invoice emails. You can also pay the invoice now using the "Pay Now" button." ?

kevcodez: Subscription endpoint already exposes payment via invoice or card

'card' | 'invoice'

kevcodez: But let's not overcomplicate it and just keep the test generic enough to match all cases

kevcodez: > The invoice will soon be charged for. In case you pay via invoice instead of card, please make sure to make the payment in a timely manner. You can also pay the invoice using your card now using the "Pay Now" button.

apps/studio/components/interfaces/Billing/InvoiceStatusBadge.tsx
```diff
@@ -63,13 +64,22 @@ const InvoiceStatusBadge = ({ status }: InvoiceStatusBadgeProps) => {
...
-                enough funds. Outstanding invoices may cause restrictions. You can manually pay the
-                using the "Pay Now" button.
-              </p>
-            )}
+            {[InvoiceStatus.OPEN, InvoiceStatus.ISSUED, InvoiceStatus.UNCOLLECTIBLE].includes(
+              status
+            ) &&
+              (paymentAttempted ? (
+                <p className="text-xs text-foreground">
+                  We were not able to collect the money. Make sure you have a valid payment method
```
kevcodez, line 72:
```suggestion
                  We were not able to collect the payment. Make sure you have a valid payment method
```

money sounds a bit off here, let's align it with your other text, payment also sounds more professional

kevcodez (approved):
Tested on staging

PR #28411, opened 2024-08-06
Show correct number of private snippets for SQL folders

joshenlim:
Dependent on API changes here:
https://github.com/supabase/infrastructure/pull/19245

Count for private snippets under the new SQL Editor nav UI is not accurate as queries within folders are loaded on demand
This should address it


apps/studio/data/content/content-id-query.ts
```diff
@@ -3,7 +3,6 @@ import { useQuery, UseQueryOptions } from '@tanstack/react-query'
 import { get, handleError } from 'data/fetchers'
 import { ResponseError } from 'types'
 import { contentKeys } from './keys'
-import { components } from 'api-types'
```
joshenlim, line 6: unused import

alaister (approved):
LGTM!

- Tested by moving queries around in private / shared / favourites / folders

PR #28416, opened 2024-08-06
Support sorting buckets in storage explorer

joshenlim:
Support searching + sorting buckets in storage explorer, similar UX to table editor

image

### DIFF
--- a/apps/studio/components/layouts/StorageLayout/StorageMenu.tsx
+++ b/apps/studio/components/layouts/StorageLayout/StorageMenu.tsx
@@ -17,6 +17,13 @@ import { useCheckPermissions } from 'hooks/misc/useCheckPermissions'
 import { useSelectedProject } from 'hooks/misc/useSelectedProject'
 import { AlertDescription_Shadcn_, AlertTitle_Shadcn_, Alert_Shadcn_, Menu } from 'ui'
 import BucketRow from './BucketRow'
+import {
+  InnerSideBarFilters,
+  InnerSideBarFilterSearchInput,
+  InnerSideBarFilterSortDropdown,
+  InnerSideBarFilterSortDropdownItem,
+} from 'ui-patterns'
+import { useLocalStorage } from '@uidotdev/usehooks'
 
 const StorageMenu = () => {
   const router = useRouter()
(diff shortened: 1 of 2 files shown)


apps/studio/components/layouts/StorageLayout/StorageMenu.tsx
```diff
@@ -61,6 +78,34 @@ const StorageMenu = () => {
...
           </ButtonTooltip>
+
+          <InnerSideBarFilters className="px-0">
+            <InnerSideBarFilterSearchInput
+              name="search-buckets"
+              aria-labelledby="Search buckets"
+              placeholder="Search buckets..."
+              value={searchText}
+              onChange={(e) => {
+                setSearchText(e.target.value.trim())
```
alaister, line 89: Maybe move the `.trim()` inside the `.sort()` call? Doing it here breaks the spacebar when typing into the input lol

apps/studio/components/layouts/StorageLayout/StorageMenu.tsx
```diff
@@ -39,12 +52,16 @@ const StorageMenu = () => {
 
   const { data, error, isLoading, isError, isSuccess } = useBucketsQuery({ projectRef: ref })
   const buckets = data ?? []
+  const sortedBuckets =
+    sort === 'alphabetical'
+      ? buckets.sort((a, b) => a.name.localeCompare(b.name))
```
alaister, line 57: related to previous comment, something like:
```suggestion
      ? buckets.sort((a, b) => a.name.toLowerCase().trim().localeCompare(b.name.toLowerCase().trim()))
```
wdyt?

apps/studio/components/layouts/StorageLayout/StorageMenu.tsx
joshenlim: oh yes! good catch

apps/studio/components/layouts/StorageLayout/StorageMenu.tsx
joshenlim: okay to do this - but does doing this cover any edge case that the original implementation wouldn't?

alaister: Should be pretty similar

joshenlim: okay gotcha! will just update it :)

alaister (approved):
LGTM!

- Tested on preview

PR #28419, opened 2024-08-06
fix: Fix a bug when updating a row which contains truncated values

ivasilov:
This PR fixes an issue where the user tried to update a value in a row which contains truncated JSON value. The update statement tries to update the whole row but we can be specific about which columns to update and skip the truncated columns. If the user loaded the full values, they will be included in the update if they're changed.

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.utils.ts
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.utils.ts
@@ -1,8 +1,9 @@
 import type { PostgresTable } from '@supabase/postgres-meta'
-import type { Dictionary } from 'types'
 import dayjs from 'dayjs'
 import { compact, find, isEqual, isNull, isString, isUndefined, omitBy } from 'lodash'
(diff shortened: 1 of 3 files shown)


saltcod (approved):

saltcod:
Checked several tables, looks good

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.utils.ts
```diff
@@ -190,25 +196,41 @@ export const generateRowObjectFromFields = (
...
       const originalFormattedOut = convertInputDatetimeToPostgresDatetime(type, originalFormatted)
       if (originalFormattedOut !== rowObject[property]) {
         payload[property] = rowObject[property]
       }
+    } else if (type !== undefined && JSON_TYPES.includes(type)) {
+      // don't update if the value is truncated. This is to enable the user to change cell values on rows which have
+      // truncated JSON values. If the user
+      const isTruncated = isValueTruncated(field?.value)
+      if (!isTruncated) {
+        console.log('passThrough')
```
joshenlim, line 220: sneaky console log

PR #28424, opened 2024-08-06
Refine lint warning on homepage

saltcod:
Refine the homepage Security Issues badge: 

separate warn and error lints — link to the warn tab if only warn lints exist 

To confirm: 

1. create a foreign data wrapper using Stripe with defaults. This will generate a warn lint.
2. this will show an amber dot and link to the warn tab in Security Advisor
3. next, run the countries quickstart in the SQL Editor. This will generate an error lint. 
4. this will show a red dot and link to the error tab in the Security Advisor 

screenshot-2024-08-06-at-15 14 25

### DIFF
--- a/apps/studio/components/interfaces/Home/SecurityStatus.tsx
+++ b/apps/studio/components/interfaces/Home/SecurityStatus.tsx
@@ -1,26 +1,27 @@
 import { CheckCircle2, Loader2 } from 'lucide-react'
 import { useState } from 'react'
 import { Button, PopoverContent_Shadcn_, PopoverTrigger_Shadcn_, Popover_Shadcn_, cn } from 'ui'
-
 import { useProjectLintsQuery } from 'data/lint/lint-query'
 import { useSelectedProject } from 'hooks/misc/useSelectedProject'
-
 import Link from 'next/link'
 import { WarningIcon } from 'ui/src/components/StatusIcon'
 
 export const SecurityStatus = () => {
   const project = useSelectedProject()
   const [open, setOpen] = useState(false)
-  const { data, isLoading } = useProjectLintsQuery({
-    projectRef: project?.ref,
-  })
+  const { data, isLoading } = useProjectLintsQuery({ projectRef: project?.ref })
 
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Home/SecurityStatus.tsx
```diff
@@ -1,26 +1,27 @@
...
+  const { data, isLoading } = useProjectLintsQuery({ projectRef: project?.ref })
 
-  const securityLints = (data ?? []).filter(
-    (lint) =>
-      lint.categories.includes('SECURITY') && (lint.level === 'ERROR' || lint.level === 'WARN')
-  )
+  const securityLints = (data ?? []).filter((lint) => lint.categories.includes('SECURITY'))
+  const errorLints = securityLints.filter((lint) => lint.level === 'ERROR')
+  const warnLints = securityLints.filter((lint) => lint.level === 'WARN')
+  const noIssuesFound = errorLints.length === 0 && warnLints.length === 0
```
charislam, line 17: question: security lints will never be `INFO` level?

charislam (approved):

charislam:
Tested as documented, instructions are great 😄

apps/studio/components/interfaces/Home/SecurityStatus.tsx
saltcod: there are info level lints, but I figure they're not high prio enough to trigger these notifications

apps/studio/components/interfaces/Home/SecurityStatus.tsx
```diff
@@ -33,8 +34,12 @@ export const SecurityStatus = () => {
             ) : (
               <div
                 className={cn(
-                  `w-2 h-2 rounded-full`,
-                  noIssuesFound ? 'bg-brand' : 'bg-destructive-600'
+                  'w-2 h-2 rounded-full',
+                  noIssuesFound
```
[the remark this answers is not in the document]
saltcod: aah good catch thanks!

PR #28426, opened 2024-08-06
Validate pg version in new project setup form

saltcod:
We can currently create a project with an invalid pg version. We should validate the input rather than just showing a failed toast.

context: https://supabase.slack.com/archives/C0161K73J1J/p1722937534699409

screenshot-2024-08-06-at-17 04 08

### DIFF
--- a/apps/studio/pages/new/[slug].tsx
+++ b/apps/studio/pages/new/[slug].tsx
@@ -271,11 +271,10 @@ const Wizard: NextPageWithLayout = () => {
         .min(3, 'Project name must be at least 3 characters long.') // Minimum length check
         .max(64, 'Project name must be no longer than 64 characters.'), // Maximum length check
       postgresVersion: z
-        .string({
-          required_error: 'Please enter a Postgres version.',
-        })
-        .regex(
-          /^1[2-9]\.\d+(\.\d+)?(-\d+)?$/,
+        .string()
+        .optional()
+        .refine(
+          (val) => !val || /^1[2-9]\.\d+(\.\d+)?(-\d+)?$/.test(val),
           'Invalid Postgres version: should start with a number between 12-19, a dot and additional characters, i.e. 15.2 or 15.2.0-3'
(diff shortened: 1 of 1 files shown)


apps/studio/pages/new/[slug].tsx
```diff
@@ -270,9 +270,14 @@ const Wizard: NextPageWithLayout = () => {
...
         .max(64, 'Project name must be no longer than 64 characters.'), // Maximum length check
-      postgresVersion: z.string({
-        required_error: 'Please enter a Postgres version.',
-      }),
+      postgresVersion: z
+        .string({
+          required_error: 'Please enter a Postgres version.',
+        })
+        .regex(
+          /^1[2-9]\.\d+(\.\d+)?(-\d+)?$/,
```
Comment (charislam) [inline apps/studio/pages/new/[slug].tsx:278]: Should this be kept in sync with the check on line 375? That regex is `/1[2-9]\..*/`

apps/studio/pages/new/[slug].tsx
```diff
@@ -270,9 +270,14 @@ const Wizard: NextPageWithLayout = () => {
         .min(1, 'Please enter a project name.') // Required field check
         .min(3, 'Project name must be at least 3 characters long.') // Minimum length check
         .max(64, 'Project name must be no longer than 64 characters.'), // Maximum length check
-      postgresVersion: z.string({
-        required_error: 'Please enter a Postgres version.',
-      }),
+      postgresVersion: z
+        .string({
```
Comment (charislam) [inline apps/studio/pages/new/[slug].tsx:274]: Should be optional, right? Current behavior is it will default to latest if left blank

apps/studio/pages/new/[slug].tsx
Comment (saltcod) [reply apps/studio/pages/new/[slug].tsx]: aah shoot missed this, updated!

apps/studio/pages/new/[slug].tsx
Comment (saltcod) [reply apps/studio/pages/new/[slug].tsx]: yes! good catch, updated

PR #28432, opened 2024-08-06
feat: add new command menu to studio

charislam:
Add new command menu to studio. Ports over commands from old menu (with some minor reorganization), then adds:

- Branch switcher
- Project switcher


apps/studio/components/interfaces/App/CommandMenu/Support.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { COMMAND_MENU_SECTIONS } from './CommandMenu.utils'
+
+const useSupportCommands = () => {
+  const commands = useMemo(
+    () =>
+      [
+        {
+          id: 'support',
+          name: 'Support',
+          route: '/support',
```
saltcod, line 16: this gives me a 404 — prob just on staging? 
https://studio-staging-git-feat-new-command-menu-studio-supabase.vercel.app/dashboard/support

apps/studio/components/layouts/AdvisorsLayout/Advisors.Commands.tsx
```diff
@@ -0,0 +1,28 @@
+import { useParams } from 'common'
```
saltcod, line 1: can remove this guy now?

saltcod (approved):

apps/studio/components/layouts/AdvisorsLayout/Advisors.Commands.tsx
charislam: I think we should keep right? This is just the navigation item (I removed the one that shows the lints)

apps/studio/components/interfaces/App/CommandMenu/Support.tsx
charislam: Ahhh, no, good catch, this is under www not studio so needs to be a full route

PR #28439, opened 2024-08-07
Support copy cell content in logs explorer

joshenlim:
image


alaister (approved):
LGTM!

- Tested by copying various cells from rows in logs explorer

PR #28442, opened 2024-08-07
Auth policies dont show warning if schema is protected

joshenlim:
Comparison: schema that's not protected vs schema thats protected

public schema:
image

realtime schema:
image


alaister (approved):
LGTM!

- Tested on preview with various schemas

PR #28443, opened 2024-08-07
Trim empty spaces in table editor filter popover inputs

joshenlim:
Mainly to prevent confusion when there's accidentally a whitespace at the front or back, and hence returning 0 results


alaister (approved):
LGTM!

- Tested with uuids and regular texts

PR #28444, opened 2024-08-07
Use expanding text area fort RLS assistant

joshenlim:
To support multi line prompts - SQL editor assistant already has this, so just bringing it in here

[One sec] Need to update both places to use AssistantChat component ([ref](https://supabase-design-system.vercel.app/design-system/docs/fragments/assistant-chat))

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/AIPolicyChat.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/AIPolicyChat.tsx
@@ -15,11 +15,11 @@ import { useProfile } from 'lib/profile'
 import { useAppStateSnapshot } from 'state/app-state'
 import {
   Button,
+  ExpandingTextArea,
   FormControl_Shadcn_,
   FormField_Shadcn_,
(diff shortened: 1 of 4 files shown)


apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/AIPolicyChat.tsx
```diff
@@ -175,7 +184,7 @@ export const AIPolicyChat = ({
             )}
           />
         </form>
-      </Form_Shadcn_>
+      </Form_Shadcn_> */}
```
alaister, line 187: This is meant to be commented out right?

alaister (approved):
LGTM!

- Added some multi line text into the sql editor ai assistant

apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/AIPolicyChat.tsx
joshenlim: oh yes! meant to remove it

PR #28458, opened 2024-08-07
Chore/update without where warning

saltcod:
Adds a new warning when a user tries to run an update query without a where clause 
 
Shows one or both warnings, depending on the query.

screenshot-2024-09-17-at-17 51 22

To test:

SQL Editor → Quickstarts → countries query 

--should show modal
`drop table countries;`  

--should show modal
`UPDATE public.countries SET name = 'New Name';` 

--should not show modal
`UPDATE public.countries SET name = 'New Name' WHERE id = 1;`  

--should show modal
`delete from countries; update countries set name = 'hello';`

--should show modal
`update countries set name = 'hello'; delete from countries;`


apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
```diff
@@ -296,9 +300,13 @@ const SQLEditor = () => {
...
 
-        const containsDestructiveOperations = checkDestructiveQuery(sql)
+        const destructiveOperations = checkDestructiveQuery(sql)
+        const updateWithoutWhereClause = isUpdateWithoutWhere(sql)
+
+        setContainsDestructiveOperations(destructiveOperations)
+        setContainsUpdateWithoutWhereClause(updateWithoutWhereClause)
 
-        if (!force && containsDestructiveOperations) {
+        if (!force && (containsDestructiveOperations || containsUpdateWithoutWhereClause)) {
```
charislam, line 309:
```suggestion
        if (!force && (destructiveOperations || updateWithoutWhereClause)) {
```

otherwise these will contain the stale values (set is asynchronous) and the modal will be bypassed

apps/studio/components/interfaces/SQLEditor/SQLEditor.utils.ts
```diff
@@ -95,7 +95,14 @@ export function getDiffTypeDropdownLabel(diffType: DiffType) {
...
 
 export function checkDestructiveQuery(sql: string) {
-  return destructiveSqlRegex.some((regex) => regex.test(removeCommentsFromSql(sql)))
+  const cleanedSql = removeCommentsFromSql(sql)
+  return destructiveSqlRegex.some((regex) => regex.test(cleanedSql))
+}
+
+// Function to check for UPDATE queries without WHERE clause
+export function isUpdateWithoutWhere(sql: string): boolean {
+  const updateWithoutWhereRegex = /^update\s+[\w.]+\s+set\s+[\w\W]+(?!where\s)/is
```
charislam, line 104: just tested this on quoted identifiers and I think the regex needs to be updated to account for those: `"public"."countries"`

apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
saltcod: yes good catch!

apps/studio/components/interfaces/SQLEditor/SQLEditor.utils.ts
saltcod: great thanks!   Updated this — catches all of them now for me 

UPDATE countries SET name = 'New Name';
UPDATE public.countries SET name = 'New Name';
UPDATE "public"."countries" SET name = 'New Name';

charislam (approved):

joshenlim (changes requested):
Found an issue:
- If you try to run both a delete _and_ update statement, e.g:
`delete from countries;
update countries set name = 'hello';`
- It'll show the delete warning modal correctly, and if i click 'Run', the query is then run without showing a warning for the update statement (and vice versa)

i think it should show both warnings?

saltcod:
Thanks @joshenlim — updated this with that in mind.  The PR description has the new test cases.

joshenlim:
hmm we shouldn't be showing 2 separate modals at once here, it's an odd UX imo

perhaps instead of 2 separate modal warnings, we should combine them into 1 single warning modal? Something like this
image

joshenlim (approved):
Okay LGTM! Tested on the preview that the modal prevents accidental running of queries. I adjusted the styling a bit to use the alert property of the ConfirmationModal + shifted the tests into SQLEditor.utils.test.ts

PR #28471, opened 2024-08-08
Chore/fix realtime inspector not being able to select latest message if previous latest message was selected

joshenlim:
Fixes a bug in the realtime inspector whereby if you have the latest message currently selected (top most), and a new message comes in, clicking on the latest message will not select the message - due to an oddity with RDG's onSelectedCellChange not triggering

I'm guessing that internally its using the row index to track which cell is selected, and because originally the selected row idx is 0, the new message that comes in is also row idx 0, so it doesnt get triggered


alaister (approved):
LGTM!

- Tested in preview realtime inspector

PR #28477, opened 2024-08-08
Only query eligibility endpoint when project is active healthy

joshenlim:
As per title, there has been feedback that the eligibility endpoint is getting queried while the project is being upgraded.  Change here just prevents the RQ from triggering unless project status is active healthy (I can't imagine a scenario where we need to query the upgrade eligibility when the project is not active)


alaister (approved):
LGTM!

- Opened a project on preview

PR #28479, opened 2024-08-08
Only render RestrictionBanner if user is logged in

joshenlim:
RestrictionBanner uses the useOrganizationQuery, which will be throwing a 401 if the user is not logged in as this component lives at the app level (e.g on the sign in screen). The change in this PR will help 
- prevent unnecessary calls to the GET organization endpoint when user is not logged in
- reduce confusion in unrelated console errors while on the sign in page


alaister (approved):
LGTM!

- Opened preview

PR #28504, opened 2024-08-09
Support deleting project in restore failed state

joshenlim:
As per title:
image

Opting to put it behind a dropdown as I'd avoid raising the delete project CTA in this context - its not always the case that a user would want to delete the project for a failed restore state (and may unnecessarily raise anxiety seeing that CTA if the failed restore was indeed something went wrong)


alaister (approved):
LGTM!

- Tested on local by deleting an restore failed project

PR #28547, opened 2024-08-12
SQL Editor support opening cell into side panel for read only detailed view

joshenlim:
Accessible via right click context menu on the cell
image

For textual values, you have the option to either view in raw or markdown
image
image

For JSON values, it'll automatically be prettified
image

### DIFF
--- a/apps/studio/components/interfaces/SQLEditor/UtilityPanel/CellDetailPanel.tsx
+++ b/apps/studio/components/interfaces/SQLEditor/UtilityPanel/CellDetailPanel.tsx
@@ -1,0 +1,67 @@
+import { useEffect, useState } from 'react'
+import remarkGfm from 'remark-gfm'
+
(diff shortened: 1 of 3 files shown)


alaister (commented):
Working well! One small issue is you have to right click on the cells text content to get the custom menu. Clicking slightly below the text doesn't work

joshenlim:
Gotcha on this - just pushed up to refactor the context menu in SQL editor results to use the UI lib's context menu component which should help to mitigate it

alaister (commented):
Something else I've run into – viewing the value of a NULL column just shows as empty. Is this okay? or should we be showing NULL?

joshenlim:
Great point! i'll make it such that it'll show NULL placeholder in the code editor
image

also pushed up a small style fix in the results table to capitalize null and make it darker
image

alaister (approved):
LGTM!

- Tested on preview with text, jsonb, and markdown

PR #28548, opened 2024-08-12
Style fixes for invite member button modal

joshenlim:
- Fix searching of projects not searching for project name correctly
- Fix width issue for long project names when project is selected

### DIFF
--- a/apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
+++ b/apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
@@ -75,9 +75,9 @@ export const InviteMemberButton = () => {
   const orgScopedRoles = (allRoles?.org_scoped_roles ?? []).sort(
     (a, b) => b.base_role_id - a.base_role_id
   )
-  const orgProjects = (projects ?? []).filter(
-    (project) => project.organization_id === organization?.id
-  )
+  const orgProjects = (projects ?? [])
+    .filter((project) => project.organization_id === organization?.id)
+    .sort((a, b) => a.name.localeCompare(b.name))
   const canReadSubscriptions = useCheckPermissions(
     PermissionAction.BILLING_READ,
     'stripe.subscriptions'
@@ -87,8 +87,8 @@ export const InviteMemberButton = () => {
     { enabled: canReadSubscriptions }
   )
   const currentPlan = subscription?.plan
-  const isOptedIntoProjectLevelPermissions =
-    useIsOptedIntoProjectLevelPermissions(slug as string) && currentPlan?.id === 'enterprise'
+  const isOptedIntoProjectLevelPermissions = true
+  // useIsOptedIntoProjectLevelPermissions(slug as string) && currentPlan?.id === 'enterprise'
 
   const userMemberData = members?.find((m) => m.gotrue_id === profile?.gotrue_id)
   const hasOrgRole =
@@ -326,7 +326,7 @@ export const InviteMemberButton = () => {
                               type="default"
                               role="combobox"
                               size="small"
-                              className="justify-between"
+                              className="justify-between max-w-[470px]"
                               iconRight={
                                 <ChevronsUpDown className="ml-2 h-4 w-4 shrink-0 opacity-50" />
                               }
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
```diff
@@ -87,8 +87,8 @@ export const InviteMemberButton = () => {
     { enabled: canReadSubscriptions }
   )
   const currentPlan = subscription?.plan
-  const isOptedIntoProjectLevelPermissions =
-    useIsOptedIntoProjectLevelPermissions(slug as string) && currentPlan?.id === 'enterprise'
+  const isOptedIntoProjectLevelPermissions = true
+  // useIsOptedIntoProjectLevelPermissions(slug as string) && currentPlan?.id === 'enterprise'
```
alaister, line 91: Just for testing?

apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
```diff
@@ -341,13 +341,16 @@ export const InviteMemberButton = () => {
...
+                                  className={cn(
+                                    (orgProjects || []).length > 7 &&
+                                      'max-h-[210px] overflow-y-auto'
+                                  )}
                                 >
                                   {orgProjects.map((project) => {
                                     return (
                                       <CommandItem_Shadcn_
                                         key={project.ref}
-                                        value={project.ref}
```
MildTomato, line 350: won't this break if 2 projects have same name ?

MildTomato (changes requested):

apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
```diff
@@ -341,13 +341,16 @@ export const InviteMemberButton = () => {
...
+                                        value={project.name}
                                         onSelect={() => {
```
MildTomato, line 354: why not onSelect={(e) => ...}

e = value of the item

MildTomato (changes requested):

apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
MildTomato: also this will break any form validation I guess

joshenlim: > won't this break if 2 projects have same name ?

ahh good point - i could add the ref as part of the value in order to ensure each item's uniqueness, but then we won't be able to do your suggestion [as below](https://github.com/supabase/supabase/pull/28548#discussion_r1713343166)

> this will break any form validation

this one's okay - the only validation here is just to not be empty, so none broken

joshenlim: pushed up a fix for the first comment - to add ref and name to the value
`onSelect` in this case will be status quo then, won't change that

alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
```diff
@@ -336,20 +336,39 @@ export const InviteMemberButton = () => {
                             </Button>
                           </PopoverTrigger_Shadcn_>
                           <PopoverContent_Shadcn_ sameWidthAsTrigger className="p-0">
-                            <Command_Shadcn_>
+                            <Command_Shadcn_
+                              // [Joshen] Let's update this to use keywords in CommandItem once cmdk is updated
+                              filter={(value, search) => {
```
MildTomato, line 341: maybe simpler 

```
filter = (value, search) => {
  const project = orgProjects.find(project => project.ref === value);
  const projectName = project?.name?.toLowerCase();
  return (projectName?.includes(search.toLowerCase()) || value.includes(search)) ? 1 : 0;
}
```

lowercase issue can be removed when updated to cmdk 1.0

MildTomato, line 341: isn't it just this ?

```
 filter={(value, search) => {
   if (value.includes(search)) return 1
   return 0
 }}
```

apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
```diff
@@ -336,20 +336,39 @@ export const InviteMemberButton = () => {
...
+                                  )}
                                 >
                                   {orgProjects.map((project) => {
                                     return (
                                       <CommandItem_Shadcn_
                                         key={project.ref}
                                         value={project.ref}
-                                        onSelect={() => {
-                                          form.setValue('projectRef', project.ref)
+                                        onSelect={(ref) => {
```
MildTomato, line 370: suggest using "value" as var name to avoid confusion for next person

MildTomato (approved):

PR #28549, opened 2024-08-12
Estimate the estimated count in table editor

joshenlim:
Use rounded numbers instead of the exact "estimated" value (the exactness of the estimate can easily throw users off).

Before: 
image

After:
image


alaister (approved):
LGTM!

- Tested on preview

PR #28560, opened 2024-08-12
Set schema dropdown if a table is selected

saltcod:
Sets the right selectedSchema in the dropdown if we go directly to a table

**To reproduce:** 
- in the Table Editor, select Realtime from the schema selector dropdown, then select a table https://share.cleanshot.com/FZ3KfvfD 
- refresh the page (or grab the url, open a new tab and open that table in a new tab)
- see that `public` is set as the selected schema again

### DIFF
--- a/apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
+++ b/apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
@@ -60,7 +60,6 @@ const TableEditorMenu = () => {
     isSuccess,
     isError,
     error,
-    refetch,
     hasNextPage,
     isFetchingNextPage,
     fetchNextPage,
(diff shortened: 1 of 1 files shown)


apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
```diff
@@ -90,17 +91,28 @@ const TableEditorMenu = () => {
   const schema = schemas?.find((schema) => schema.name === snap.selectedSchemaName)
   const canCreateTables = useCheckPermissions(PermissionAction.TENANT_SQL_ADMIN_WRITE, 'tables')
 
-  const refreshTables = async () => {
```
saltcod, line 93: this looks like it's been unused for a while

charislam (approved):
tested as described

PR #28573, opened 2024-08-12
feat: snippet commands

charislam:
Add commands to studio cmd k for searching and running snippets (doesn't actually run them directly cause that would take a lot of duplicated logic, just links to sql editor with them loaded)

Not visible if you haven't signed in or haven't selected a project:
image

If you are in a project, snippet command shows:
image

Empty state:
image

Snippet selector view:
image


saltcod (approved):

saltcod:
looks good!

charislam:
> small nit: the names of the snippets could be better — ex: there's countries and countries table (one is a view and one is the countries table quickstart)

I think that's the AI generation at work, I just used the existing snippet names. But I just added the beginning of the snippet SQL to the search value, so hopefully that'll make it easier to disambiguate when searching.

joshenlim:
i'm not too sure what might be happening but i seem to running into this error when hovering some of my queries? It's happening for e.g when i hover my snippet called "Slack clone" which i got it from the quickstarts
image

joshenlim:
this seems to be the error for the above comment
image

charislam:
@joshenlim my bad, happened when I added the beginnings of the SQL content because the value is used as an attribute and needs to be escaped, I pushed a commit that fixes it just now

joshenlim (approved):
okay LGTM! Tested on the preview:
- Cmd K -> Snippets -> Select "Run snippets"
- Able to see my list of snippets, even those in folders
- Selecting snippets opens them up in the SQL editor
- Selecting "Create new snippet" opens up /sql/new

PR #28579, opened 2024-08-13
Chore/auth policies page fixes

joshenlim:
- Replace manual tooltip buttons with the new ButtonTooltip component (the tooltips weren't showing so the refactor is necessary)
- Created new DropdownMenuItemTooltip component and use it in PolicyRow
- Update UI to allow access to the Edit Policy panel for a user who doesn't have update access BUT prevent saving from there
  - This is to allow such users to still be able to read the policy definitions despite not being able to save (without having to do any update to the Auth Policies UI atm)
- Minor refactors around to use lucide react icons

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
@@ -32,6 +32,8 @@ import {
   Select_Shadcn_,
 } from 'ui'
 import { MultiSelectV2 } from 'ui-patterns/MultiSelectDeprecated/MultiSelectV2'
+import { PermissionAction } from '@supabase/shared-types/out/constants'
+import { useCheckPermissions } from 'hooks/misc/useCheckPermissions'
 
 interface PolicyDetailsV2Props {
   schema: string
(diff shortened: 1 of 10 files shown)


apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
```diff
@@ -60,6 +62,7 @@ export const PolicyDetailsV2 = ({
 }: PolicyDetailsV2Props) => {
   const { project } = useProjectContext()
   const [open, setOpen] = useState(false)
+  const canUpdatePolicies = false // useCheckPermissions(PermissionAction.TENANT_SQL_ADMIN_WRITE, 'tables')
```
alaister, line 65: Testing? 👀

apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/PolicyRow.tsx
```diff
@@ -54,89 +58,68 @@ const PolicyRow = ({
...
+              <TooltipContent_Shadcn_ side="bottom" align="center">
+                {policy.roles.slice(3).join(', ')}
+              </TooltipContent_Shadcn_>
+            </Tooltip_Shadcn_>
+          )}
         </div>
       </div>
       <div>
-        {canUpdatePolicies ? (
+        {true ? (
```
alaister, line 83: Also testing? 👀

apps/studio/components/interfaces/Auth/Policies/AIPolicyEditorPanel/PolicyDetailsV2.tsx
joshenlim: very much yes! 🙏

apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/PolicyRow.tsx
joshenlim: ahh yes this too!

alaister (approved):
LGTM!

- Tested as read-only user on preview

PR #28581, opened 2024-08-13
Refactor icons from ui to lucide part 1

joshenlim:


apps/studio/components/grid/components/formatter/ForeignKeyFormatter.tsx
```diff
@@ -56,18 +57,17 @@ export const ForeignKeyFormatter = (props: Props) => {
...
               className="translate-y-[2px]"
-              onClick={() => {}}
               style={{ padding: '3px' }}
-              asChild
             >
               <Link
                 href={`/project/${projectRef}/editor/${targetTable?.id}?filter=${relationship?.target_column_name}%3Aeq%3A${value}`}
                 onClick={() => snap.setSelectedSchemaName(relationship.target_table_schema)}
               >
-                <IconArrowRight size="tiny" />
```
MildTomato, line 70: don't need size

apps/studio/components/grid/components/grid/SelectColumn.tsx
```diff
@@ -144,7 +145,7 @@ function SelectCellFormatter({
               type="text"
               size="tiny"
               className="rdg-row__select-column__edit-action"
-              icon={<IconMaximize2 size="tiny" strokeWidth={1.5} className="text-foreground" />}
+              icon={<Maximize2 strokeWidth={1.5} className="text-foreground" />}
```
MildTomato, line 148: don't need text-foreground anymore

apps/studio/components/grid/components/formatter/ForeignKeyFormatter.tsx
MildTomato: when an icon is in a `<Button />` that is

apps/studio/components/grid/components/grid/SelectColumn.tsx
MildTomato: when an icon is in a `<Button />` that is

apps/studio/components/grid/components/formatter/ForeignKeyFormatter.tsx
joshenlim: i think in this case we need the size? since the icon is used standalone

with size:
image

without size:
image

alaister (approved):
LGTM!

- Tested by clicking through a bunch of pages on preview and couldn't see any wonky icons

PR #28582, opened 2024-08-13
Fix limit in sql editor

joshenlim:
Fix auto limit logic in SQL editor not updating


alaister (approved):
LGTM!

- Tested with a large table on preview

PR #28585, opened 2024-08-13
Chore/composite foreign key fixes

joshenlim:
- Fix inability to select button to view referencing row from Table Editor grid if the column has both FK and is also a PK, should now see the button
image
- Fix inaccuracy of column mapping for composite foreign key columns
  - Before: Notice both are `public.properties.property_id`
image
  - After:
image
- Fix inability to update composite foreign key columns via RowEditor
  - It's related to the above problem, and also a UX problem where users are expected to select values separately for the columns under the composite foreign key, when just one row is needed
  - Selecting the target row will now update the values for all input fields of columns that are under the same composite foreign key

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.utils.ts
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.utils.ts
@@ -1,4 +1,4 @@
-import type { PostgresTable } from '@supabase/postgres-meta'
+import type { PostgresRelationship, PostgresTable } from '@supabase/postgres-meta'
 import dayjs from 'dayjs'
 import { compact, find, isEqual, isNull, isString, isUndefined, omitBy } from 'lodash'
 import type { Dictionary } from 'types'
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.utils.ts
```diff
@@ -36,17 +37,26 @@ export const generateRowFields = (
...
 
-    const foreignKey = find(relationships, (relationship) => {
-      return (
-        relationship.source_schema === column.schema &&
-        relationship.source_table_name === column.table &&
-        relationship.source_column_name === column.name
-      )
+    const foreignKey = foreignKeys.find((fk) => {
+      return fk.columns.map((x) => x.source).includes(column.name)
     })
```
alaister, line 42: Just to check, checking the `schema` is not important here anymore?

joshenlim: actually yeah good point - should add that check just to be sure as there could be same column names across schemas

joshenlim: oh yeap this is okay - cause we already filter the foreign keys one level above in RowEditor
https://github.com/supabase/supabase/blob/e2f09af1009da3b355054da915355f2ae273b614/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.tsx#L67

apps/studio/state/table-editor.tsx
```diff
@@ -182,7 +183,8 @@ export const createTableEditorState = () => {
         sidePanel: { type: 'cell', value: { column, row } },
       }
     },
-    onEditForeignKeyColumnValue: (foreignKey: ForeignKey) => {
+    onEditForeignKeyColumnValue: (foreignKey: ForeignKeyState) => {
+      console.log('onEditFK', { foreignKey })
```
alaister, line 187: Cheeky one!
```suggestion
```

alaister (approved):
LGTM!

- Tested on preview with tables with composite primary keys and foreign keys

Issue #28590, opened 2024-08-13
The placeholder in "CHECK Constraint" input is not suggesting the correct example


sokhuong-uon:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I created a table with the `duration_min` field as `int4`.
This is what I see in the check constraint
![image](https://github.com/user-attachments/assets/05deb4ae-807e-4b2b-8cfe-b63baac7624b)

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Create a table which include a `duration_min` field. Set `duration_min` to `int4`.
2. Click save. So the table is created now.
3. Click to view the newly created table.
4. Click on the dropdown menu of the `duration_min` column
5. Click "Edit Column"
6. Scroll down and see
![image](https://github.com/user-attachments/assets/05deb4ae-807e-4b2b-8cfe-b63baac7624b)
7. Follow the suggestion will lead to error.

## Expected behavior

Show `e.g duration_min > 0` or something correct.

## System information

- OS: Ubuntu
- Browser Chrome

## Additional context

I use supabase.com in the browser.

sokhuong-uon:
Clarification:

The type of the field is `int4`.
I want the placeholder to be something like:
- `duration_min > 0`
- `duration_min < 0`
- `duration_min < -100` (doesn't make sense but it's ok because we don't understand the semantic of the field name)

I think that checking the length of an integer doesn't seems right.
In fact, we will get a postgresql error if we follow the suggestion in placeholder.

major-winter:
Hi @sokhuong-uon , this issue occurs with other types like `int2, int8,...` as well. Do I understand correctly that you want a dynamic placeholder that can adjust to match the selected data type?

sokhuong-uon:
yes, you're right.

major-winter:
Great, I'm working on it and will be creating a PR soon.

praniket01:
Is anyone working on this currently ?

major-winter:
> Is anyone working on this currently ?
> 
> 

There's an open PR for this issue. It's awaiting approval.

Issue #28536, opened 2024-08-11
Rename Supabase in Hindi Translation of README


iAbhinav:
# Improve documentation

## Link 

[Hindi](https://github.com/supabase/supabase/blob/master/i18n/README.hi.md) translation of the Readme.md.

## Describe the problem

Supabase is mis-spelled in multiple places.

## Describe the improvement

It should be `सुपाबेस` and not `सुपाबसे`

✋🏻 Assign me the issue, i am a native hindi speaker.

charislam:
Thanks for the report @iAbhinav! We don't assign issues here (just take the first working PR that comes in), so feel free to work on it!

PR #28606, opened 2024-08-13
fix: The placeholder in "CHECK Constraint" suggestion wrong

zerone0x:

YES

Bug fix, Doc

### The placeholder didn't have correct suggesstion
e.g length(duration_min) ‹ 500
[ The placeholder in "CHECK Constraint" input is not suggesting the correct example #28590 ](https://github.com/supabase/supabase/issues/28590)

### Wrong name: सुपाबसे
[Rename Supabase in Hindi Translation of README #28536](https://github.com/supabase/supabase/issues/28536)

## What is the new behavior?

- e.g duration_min > 0 
- Fix new name to सुपाबेस

## Additional context

I updated placeholder to show correct suggestions. And some doc of hindi. Pls help review it, thanks!


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.tsx
```diff
@@ -357,7 +357,7 @@ const ColumnEditor = ({
           <Input
             label="CHECK Constraint"
             labelOptional="Optional"
-            placeholder={`e.g length(${columnFields?.name || 'column_name'}) < 500`}
+            placeholder={`e.g length(${columnFields?.name || 'column_name'}) > 0`}
```
sokhuong-uon, line 360: This is not what I meant.

The type of the field is `int4`.
I want the placeholder to be something like:
- `duration_min > 0`
- `duration_min < 0`
- `duration_min < -100` (doesn't make sense but it's ok because we don't understand the semantic of the field name)

I don't meant the placeholder should be changed to `> 0` to fit the meaning of my table's field's name.

I just think that checking the length of an integer doesn't make sense.

PR #28607, opened 2024-08-13
fix: repeated responses in docs search

charislam:
Fixes a bug in docs search from duplicated code (I moved the useEffect into a new component and forgot to delete the old one, so it runs the on-entry query submit twice, resulting in a double conversation in production.)

In dev, there is actually a _quadruple_ -_-" conversation, due to useEffect running twice as designed, with no cleanup. Added a cleanup function to reset after the first useEffect, so it only runs once on dev as well.

This only happens when there is an existing query to send when you hit the AI page, so to reproduce the current bug, type in the command input _before_ selecting "Supabase AI". If you select "Supabase AI" before entering your prompt, it won't occur, which is how I guess this bug slipped past manual testing.

https://supabase.slack.com/archives/C023E4L60R3/p1723551853199999

### DIFF
--- a/packages/ui-patterns/CommandMenu/prepackaged/DocsAi/DocsAiPage.tsx
+++ b/packages/ui-patterns/CommandMenu/prepackaged/DocsAi/DocsAiPage.tsx
@@ -55,11 +55,6 @@ const DocsAiPage = () => {
     setIsLoading,
   })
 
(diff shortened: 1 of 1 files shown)


fsansalvadore (approved):
Can confirm this solved the bug in the preview ✅👍

PR #28609, opened 2024-08-13
fix: apply grants on create / update / delete auth postgres hook

kangmingtay:

YES

* Currently, users need to manually run these SQL statements if they create a postgres hook (https://supabase.com/docs/guides/auth/auth-hooks#security-model). If they don't remember to run them, the postgres function acting as the webhook will be insecure as it is accessible from the data API. This PR aims to improve the DX by running these statements on the postgres function selected as the webhook.

* Users have to remember to manually run these statements to secure the postgres function configured as the hook

## What is the new behavior?
* Run these statements when the postgres function is configured as the hook - this significantly improves the DX and makes it secure-by-default since the user doesn't have to remember to run them.

* When a hook is created, we display the statements that will be run on the postgres function:
image

* When a hook is updated, we restore the original permissions to the old function and apply the new permissions to the new function (note the revoke statements applied to `hook_foo`):
image

* When a hook is deleted, we restore the original permissions to the function (edit: made modal bigger):
image

## Additional context
* All of these statements are executed through `pg_meta` 

### DIFF
--- a/apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
+++ b/apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
@@ -412,14 +412,17 @@ export const CreateHookSheet = ({
                       </FormItemLayout>
                     )}
                   />
-                </div>
-                <div className={cn('h-64 w-full')}>
-                  <CodeEditor
-                    id="postgres-hook-editor"
-                    isReadOnly={true}
-                    language="pgsql"
-                    value={statements.join('\n\n')}
-                  />
+                  <div className={cn('h-72 w-full col-span-2')}>
+                    <p className="py-4 text-sm text-foreground-light">
+                      {`The following statements will be executed on the function:`}
+                    </p>
+                    <CodeEditor
+                      id="postgres-hook-editor"
+                      isReadOnly={true}
+                      language="pgsql"
+                      value={statements.join('\n\n')}
+                    />
+                  </div>
                 </div>
               </>
             ) : (
--- a/apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
+++ b/apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
@@ -116,7 +116,6 @@ export const CreateHookSheet = ({
   const { ref: projectRef } = useParams()
   const { project } = useProjectContext()
   const { mutate: updateAuthConfig, isLoading: isUpdatingConfig } = useAuthConfigUpdateMutation()
-  const [statements, setStatements] = useState<string[]>([])
   const httpsAuthHooksEnabled = useFlag('httpsAuthHooksEnabled')
 
   const definition = useMemo(
@@ -124,14 +123,16 @@ export const CreateHookSheet = ({
     [title]
   )
 
-  const hook: Hook = {
-    ...definition,
-    enabled: authConfig?.[definition.enabledKey] || false,
-    method: extractMethod(
-      authConfig?.[definition.uriKey] || '',
-      authConfig?.[definition.secretsKey] || ''
-    ),
-  }
+  const hook: Hook = useMemo(() => {
+    return {
+      ...definition,
+      enabled: authConfig?.[definition.enabledKey] || false,
+      method: extractMethod(
+        authConfig?.[definition.uriKey] || '',
+        authConfig?.[definition.secretsKey] || ''
+      ),
+    }
+  }, [definition, authConfig])
 
   // if the hook has all parameters, then it is not being created.
   const isCreating = !isValidHook(hook)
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
```diff
@@ -219,6 +238,15 @@ export const CreateHookSheet = ({
         },
       }
     )
+
+    // grant permissions to new function
+    if (values.postgresValues.statements.length > 0) {
+      await executeSql({
```
ivasilov, line 244: This doesn't run after `updateAuthConfig`, runs in parallel because there's no await on it. You can move this code into `onSuccess` and add something in `onError` which reverses the revokation.

apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
```diff
@@ -300,52 +328,107 @@ export const CreateHookSheet = ({
...
+                        description="Postgres function which will be called by Supabase Auth each time the hook is triggered."
+                      >
+                        <FormControl_Shadcn_>
+                          <FunctionSelector
+                            size="small"
+                            schema={values.postgresValues.schema}
+                            value={field.value}
+                            onChange={(value) => {
+                              setStatements([])
+                              if (hook.method.type === 'postgres') {
```
ivasilov, line 370: No need for this magic. Move all of this into a `useMemo` and use the `values` from line 252 to construct the statements.

apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
```diff
@@ -300,52 +328,107 @@ export const CreateHookSheet = ({
...
+                          description="These permission changes will be applied on the postgres function configured as the hook."
+                        >
+                          <div className={cn('h-64')}>
+                            <FormControl_Shadcn_>
+                              <CodeEditor
+                                id="postgres-hook-editor"
+                                isReadOnly={true}
+                                language="pgsql"
+                                value={statements.concat(grantStatements).join('\n\n')}
+                                onInputChange={field.onChange}
```
ivasilov, line 422: Can the user change the statements? No need for all this logic, just a simple `div` will be ok.

kangmingtay: nope it should be read-only, the user shouldn't be able to modify the statements

ivasilov: also, make the div have the same padding as the rest of the form.

apps/studio/components/interfaces/Auth/Hooks/HooksListing.tsx
```diff
@@ -112,6 +117,7 @@ export const HooksListing = () => {
           if (!selectedHookForDeletion) {
```
ivasilov, line 117: Set to size of the modal to be bigger for easier reading of the SQL code.

apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
kangmingtay: @ivasilov might need some help with this one - i can't seem to figure out why the `useMemo` doesn't recalculate when the functions selector change multiple times

apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
```diff
@@ -215,14 +226,71 @@ export const CreateHookSheet = ({
       {
         onSuccess: () => {
           toast.success(`Successfully created ${values.hookType}.`)
+          if (statements.length > 0) {
```
ivasilov, line 229: I think this logic is wrong or duplicated with the one at line 258. What is the idea, revoke permissions before changing, change then grant the permissions again, right?

apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
ivasilov: Fixed.

apps/studio/components/interfaces/Auth/Hooks/CreateHookSheet.tsx
kangmingtay: we just want to apply the permission changes to the postgres functions (old and new) when the config is updated

the old function will be restored to it's original state and the new function will have the new permissions applied

kangmingtay: the order in which it's being executed here doesn't really matter, as long as the permission changes are applied when the auth config update endpoint is called it's fine

ivasilov (approved):

PR #28629, opened 2024-08-14
fix sort bug in fk select

jordienr:

YES


joshenlim (approved):
LGTM, tested on the staging preview
- Table editor for a table which has a column that's a foreign key to another, insert row via RowEditor -> "Select record" -> sort in the table there works

PR #28641, opened 2024-08-14
chore(docs): remove preview build shortcuts

charislam:
Now that production builds are fast, let's remove the preview build shortcuts. (One of them didn't work anyway.) Having different preview and production builds can lead to edge case bugs making it into master, and we won't see much further build time improvement that's worth the risk.


ivasilov (approved):
Tested few pages on the docs app, works ok.

PR #28647, opened 2024-08-14
feat: create new table command

charislam:
image


joshenlim (approved):
LGTM - tested on the preview
- Cmd + K -> "Create" -> Select "Create table"
- Opens up table editor with the table editor side panel open

Some nits though
- This one's probably not related but i tried to type "Create" while on the home page and saw "Create RLS policy" -> wondering if we should hide that unless the user is in a project? Like create table
image

- Wondering if we could use the table icon instead of plus? am presuming we'll eventually add more "Create" commands, so am thinking having contextual icons will make it easier to find what the user needs
image
image

PR #28654, opened 2024-08-15
Fix ColumnEditor removing default value when updating foreign key

joshenlim:
In the table editor when editing a column, if you have a column that has a default value and you try to set a foreign key for it, the default value input field will be cleared. This PR fixes that


alaister (approved):
LGTM!

- Tested on preview by updating a foreign key with a default value

PR #28250, opened 2024-07-29
feat: Replace all toasts with sonner

ivasilov:

PR #28670, opened 2024-08-15
feat/export-as-sql

saltcod:
Add the option to export tables as SQL 

**To test:**
- here's a kitchen sink of data types https://gist.github.com/saltcod/c9046c97ba72d803c77bf86c82ff5ee3
- create this table, import the data
- export the data to sql 
- remove the id column
- import the sql 
 

**header:**
screenshot-2024-08-15-at-12 53 54

**table sidebar menu:** 
screenshot-2024-08-16-at-09 57 39

https://www.notion.so/supabase/Export-table-editor-content-as-SQL-396b9276783f495d99ab65e55cafef6f


apps/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
```diff
@@ -333,7 +394,18 @@ const EntityListItem: ItemRenderer<Entity, EntityListItemProps> = ({
               }}
             >
               <Download size={12} />
-              <span>Export as CSV</span>
+              <span>Export table as CSV</span>
+            </DropdownMenuItem>
+            <DropdownMenuItem
```
joshenlim, line 399: i'd opt to do a secondary dropdown menu for this - otherwise the dropdown menu here seems to be getting larger and clunkier

apps/studio/components/grid/components/header/Header.tsx
```diff
@@ -327,6 +327,52 @@ const RowHeader = ({ table, sorts, filters }: RowHeaderProps) => {
...
+      .map((row) => {
+        const filteredRow = { ...row }
+        delete filteredRow.idx
+
+        const columns = Object.keys(filteredRow)
+          .map((col) => `"${col}"`)
+          .join(', ')
+        const values = Object.values(filteredRow)
+          .map((val) => (typeof val === 'object' && val !== null ? JSON.stringify(val) : val))
+          .map((val) => `'${val}'`)
```
joshenlim, line 366: hmm i'd be careful here RE correctness in representation of data
we'd technically stringify numerical values here too which, depending on the column's data type, might not be want the user wants

ideally we'd also have some unit tests around this to ensure its correctness since we're writing the exporting logic ourselves

apps/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
```diff
@@ -140,6 +140,67 @@ const EntityListItem: ItemRenderer<Entity, EntityListItemProps> = ({
     }
   }
 
+  const exportTableAsSQL = async () => {
```
joshenlim, line 143: is this to export the table or the table rows? might be more accurate to label the dropdown menu as export table data perhaps? i initially thought it'll include the "CREATE table" syntax

joshenlim: also can we try to reuse the function that you wrote under "Header.tsx"? at first glance they look like they can share the same function (and we can write one test to cover these two cases)

apps/studio/components/grid/components/header/Header.tsx
joshenlim: also this looks like `n` number of INSERT statements which i'm guessing might be rough on the DB esp if theres gonna be a lot of rows - we should export this into a single INSERT statement with multiple values imo

joshenlim: i'm currently taking a look at this - will push up some changes

apps/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
saltcod: hah yes of course, thanks!

apps/studio/components/interfaces/TableGridEditor/TableEntity.utils.ts
```diff
@@ -24,19 +24,30 @@ export const getEntityLintDetails = (
...
+            typeof val === 'object' && val !== null ? JSON.stringify(val) : `'${val}'`
+          )
+          .join(', ') +
+        ')'
+      )
     })
-    .join('\n')
+    .join(', ')
+
+  return `INSERT INTO ${table} (${columns}) VALUES ${valuesSets};`
```
saltcod, line 52: changed this as the original version was doing an insert inside the .map, so would return an insert on every iteration

joshenlim:
pushed up some changes that will enable formatTableRowsToSQL to:
- consider the schema in the insert statement
- consider array + json column types and format them correctly

i think we should be good to go now!

apps/studio/components/grid/components/header/Header.tsx
```diff
@@ -327,6 +328,37 @@ const RowHeader = ({ table, sorts, filters }: RowHeaderProps) => {
     setIsExporting(false)
   }
 
+  async function onRowsExportSQL() {
+    setIsExporting(true)
+
+    if (allRowsSelected && totalRows > MAX_EXPORT_ROW_COUNT) {
+      toast.error(
+        `Sorry! We're unable to support exporting of SQL for row counts larger than ${MAX_EXPORT_ROW_COUNT.toLocaleString()} at the moment.`
```
alaister, line 336: Worth giving the user a workaround here? Linking to a guide on how to use pg_dump or something... https://supabase.com/docs/reference/cli/supabase-db-dump

alaister (approved):
LGTM!

- Tested as per instructions on preview

apps/studio/components/grid/components/header/Header.tsx
joshenlim: good point!

joshenlim: and also i realised this validation is not in EntityListItem either - will fix that

joshenlim: okay pushed up the fixes ^ will look like this (ignore the 5, i made the limit smaller to trigger the toast)
image

just FYI though i don't think the toast here communicates clearly RE the docs, but we'll update it once this PR https://github.com/supabase/supabase/pull/28250 goes through, cause I think using sonner will be better, but I don't wanna introduce a 3rd toast in this PR just yet

PR #28678, opened 2024-08-15
Add caution on Update Policy section for need to create Select Policy

### LINKED ISSUE #28559: UPDATE RLS policy requires SELECT RLS policy too
# Improve documentation

## Link 

https://github.com/supabase/supabase/blob/master/apps/docs/content/guides/database/postgres/row-level-security.mdx#update-policies

## Describe the problem

UPDATE RLS policy doesn't mention that  it also needs a SELECT POLICY even for update action from client without chaining select .

e.g. Following code from the client will work on the RLS enabled `profiles` table only when there are both UPDATE and SELECT POLICIES.

```
    await supabase
        .from('profiles')
        .update({'name': "test"}).eq('id', userId);
```

```
--RLS POLICY TO ALLOW THE USERS TO UPDATE THEIR ROWS
create policy "Users can update their own profile."
on profiles for update
using ( (select auth.uid()) = id )       -- checks if the existing row complies with the policy expression
with check ( (select auth.uid()) = id ); -- checks if the new row complies with the policy expression
```

```
--RLS POLICY TO ALLOW THE USERS TO SELECT THEIR ROWS
create policy "Users can select their own profile."
on profiles for select
using ( (select auth.uid()) = id )       -- checks if the existing row complies with the policy expression
```

Without the aforementioned SELECT policy, Update policy wouldn't work but the documentation seems to suggest that just the UPDATE policy is needed. Many supabase users have faced the [issue](https://community.weweb.io/t/supabase-does-not-update-despite-success/6286/18) of UPDATE not working after setting requisite RLS policy.

## Describe the improvement

Add a note and example  to the UPDATE policy for the relevant SELECT policy to go along with it.

## Additional context

Right now there is a separate SELECT policy example in the documentation but its not relevant to the UPDATE policy.

Add any other context or screenshots that help clarify your question.

bulutyerli:

YES

Docs update, ...

The Update Policy section of the documentation did not mention the need to create a Select Policy for Row-Level Security.

## What is the new behavior?

Added a caution note in the Update Policy section to inform users that they must also create a Select Policy to ensure proper Row-Level Security implementation.

## Additional context

This change closes [#28559](https://github.com/supabase/supabase/issues/28559).

Supabase SS


apps/docs/content/guides/database/postgres/row-level-security.mdx
```diff
@@ -172,6 +172,12 @@ with check ( (select auth.uid()) = user_id ); -- checks if the new row complies
 
 If no `with check` expression is defined, then the `using` expression will be used both to determine which rows are visible (normal USING case) and which new rows will be allowed to be added (WITH CHECK case).
 
+<Admonition type="caution">
+
+To perform an `UPDATE` operation, a corresponding <a href="#select-policies">`SELECT` policy</a> is required. Without a `SELECT` policy, the `UPDATE` operation will not work as expected.
```
charislam, line 177:
```suggestion
To perform an `UPDATE` operation, a corresponding [`SELECT` policy](#select-policies) is required. Without a `SELECT` policy, the `UPDATE` operation will not work as expected.
```

charislam (approved):
Thanks @bulutyerli!

PR #28679, opened 2024-08-15
Prevent number inputs from changing via scroll

saltcod:
Prevent this insanity
https://x.com/cdt5058/status/1822986663850627169

### DIFF
--- a/packages/ui/src/components/shadcn/ui/input.tsx
+++ b/packages/ui/src/components/shadcn/ui/input.tsx
@@ -1,5 +1,5 @@
 import { VariantProps, cva } from 'class-variance-authority'
-import * as React from 'react'
+import { forwardRef, useEffect, useImperativeHandle, useRef } from 'react'
 import { SIZE_VARIANTS, SIZE_VARIANTS_DEFAULT } from '../../../lib/constants'
 import { cn } from '../../../lib/utils/cn'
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/shadcn/ui/input.tsx
```diff
@@ -1,5 +1,5 @@
 import { VariantProps, cva } from 'class-variance-authority'
```
charislam, line 1:
```suggestion
'use client'

import { VariantProps, cva } from 'class-variance-authority'
```

needed to not fail the design-system and docs builds because app router

packages/ui/src/components/shadcn/ui/input.tsx
```diff
@@ -18,12 +18,32 @@ const InputVariants = cva('aria-[]', {
   },
 })
 
-const Input = React.forwardRef<HTMLInputElement, InputProps>(
+const Input = forwardRef<HTMLInputElement, InputProps>(
   ({ className, type, size = 'small', ...props }, ref) => {
+    const inputRef = useRef<HTMLInputElement>(null)
+
+    useImperativeHandle(ref, () => inputRef.current!)
```
charislam, line 25: 🤯 so much better than the shenanigans I do to merge refs, will be copying this trick 🧠

charislam (commented):
Hmm still seems to be happening for me? are these ones using a different input?


https://github.com/user-attachments/assets/4e8aba5e-0295-4a48-b38b-08a54a354267

joshenlim:
> Hmm still seems to be happening for me? are these ones using a different input?

Hmm you're right - there's different input components being used here `Input` and `InputNumber`, i'll look into this

joshenlim:
I think i found the fix but I have no idea why it works - but just to make sure we're on the same page on the bug that we're trying to fix first
- For an input field of type number, while the mouse is over the input field, scrolling _should_ update the value of the input field but not scroll the page. If the mouse is not over the input field, the page should then scroll while the value of the input field should _not_ update. Correct me if I'm wrong here though but this is the normal native browser input html behaviour? and we shouldn't be preventing that
- The bug here however is that for `InputNumber` in the SMTP Provider settings, while the mouse is over the input field and we scroll, both the value updates a little and the page scrolls too

Was trying to find what's the difference in usage between the `InputNumber` components from `BasicAuthSettings` and `SmtpForm` since the bug was only happening in the latter, and realised that the `size` prop was missing? Adding a `size="small"` to the `InputNumber`s under SMTP Provider settings _somehow_ seems to have fixed the issue.

I'd opt to just ^ do that for now since end goal is that we move away from the old `Input` and `InputNumber` components anyways to use the new Shadcn ones

joshenlim:
this problem does also exist in auth/rate-limits fields where we're using the new Shadcn Input components with type set to number - but I don't think the fix here is to remove updating the input field value via scroll since thats a native browser behaviour

will probably need @MildTomato to chime in on this since it involves changes to packages/ui

joshenlim:
@MildTomato RE the bug that we're trying to solve here
- Notice that for an Input field of type number, when my mouse is over the input field and i scroll, both the page scrolls + the input value updates

https://github.com/user-attachments/assets/e8748da7-5c51-41cc-8a7a-0fa512868b18

PR #28683, opened 2024-08-15
feat: command to query a table

charislam:
Creates this subpage:

image

Selecting a table brings you to the SQL editor with this prefilled:

image


joshenlim:
hahaha caught this in my previous stress testing scenarios 😂 
im guessing the highlighting logic is ignoring cases on the table name?
image

charislam:
@joshenlim 

> my table names give this vibe 😂

I love it 😂 

good catch with the highlighting, nice stress testing there

I quoted all identifiers that need to be quoted in SQL, in the interest of "educating users about SQL" (didn't feel right to write it as `schema.tableName` when it's not valid), but I guess we don't do that anywhere else in the dashboard, so I'll take it out

charislam:
@joshenlim should be fixed now!

jordienr:
lgtm! tested in preview environment

jordienr (approved):

PR #28687, opened 2024-08-16
Fix sorting in roles

joshenlim:
Fix the sorting of roles as highest to lowest privileges:
image


alaister (approved):
LGTM!

- Tested on preview

PR #28706, opened 2024-08-16
fix: disable sms provider validation when sms hook is enabled

kangmingtay:

YES

* When the SMS hook is enabled, we should disable the validation for the SMS provider selected. The SMS hook will be used in place of the SMS provider configured. 
* Only the configuration related to the SMS provider (e.g. twilio, messagebird, vonage, textlocal) should be disabled - other configs like `SMS_AUTOCONFIRM`, `SMS_OTP_LENGTH`, etc should remain unaffected since they are still applicable when the hook is configured

* A user will not be able to enable the phone provider without added the configuration for the SMS provider (see [comment](https://github.com/supabase/auth/issues/1582#issuecomment-2292810035)

## What is the new behavior?
* We show an alert in the "Phone" provider form section if the SMS hook is enabled and disable the sms provider fields.
* If the hook is disabled, we should apply the validation for the SMS provider fields.

image

## Additional context

### DIFF
--- a/apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersForm.tsx
+++ b/apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersForm.tsx
@@ -12,7 +12,7 @@ import {
   Button,
   WarningIcon,
 } from 'ui'
-import { PROVIDERS_SCHEMAS } from '../AuthProvidersFormValidation'
+import { getPhoneProviderValidationSchema, PROVIDERS_SCHEMAS } from '../AuthProvidersFormValidation'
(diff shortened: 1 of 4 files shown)


joshenlim:
Generally LGTM - but i'll push up some small changes to the copy and styling
- I don't think we need a warning Alert for this one specifically
- Copy can be a bit more informative (e.g explaining why is it disabled)

joshenlim:
Pushed up the following changes:

- Update Alert in Phone provider settings
image


- Added description in SMS hook side panel
image

joshenlim (approved):
LGTM otherwise! Tested locally
- SMS Provider specific settings are disabled while SMS hook is enabled

PR #28716, opened 2024-08-17
doc: Fix reference to non-existant schema in SQL snippet of "Custom Claims & RBAC"

tvollstaedt:

YES

Docs update

The original example snipped references a schema `public.users` that does not exist by default. While there is a reference to the full schema file, we want to provide simple c&p templates to get the user up and running.

## What is the new behavior?

We could either add a `public.users` schema, but since the referenced one just links to the auth db and does not fulfill any relevant job in the example, let's simply reference the `auth` schema directly.


charislam (approved):
Thanks @tvollstaedt!

PR #28733, opened 2024-08-19
feat: control increase disk modal via url state

alaister:

feat

The increase disk is unable to be controlled via the URL, so we can't link directly to it.

## What is the new behavior?

Add `?show_increase_disk_size_modal=true` on the `/database` settings page to open the modal.


pamelachia (approved):

Pyr33x (approved):

Pyr33x (approved):

joshenlim (approved):
LGTM - tested on the preview
- Going to settings/database with the URL param, modal opens
- Modal has the appropriate validations depending on the plan of the org
- Modal has the loading states required

PR #28766, opened 2024-08-20
fix: Fix various bugs in Text Editor and JSON editor

ivasilov:
This PR fixes 3 separate bugs when editing a specific row in table editor:
- The `expand editor` option in a dropdown on a text field now opens the Text editor.
- When you open a `JSON` field with a big JSON payload, the JSON editor offers to load the full JSON. If you load it, press cancel, and go back again to the JSON editor, the value will be truncated (with no option to load the full payload).
- When you open a `text` field with a big payload, the Text editor offers to load the full text. If you load it, press cancel, and go back again to the JSON editor, the value will be truncated (with no option to load the full payload).


saltcod:
screenshot-2024-08-20-at-16 58 01
getting an error when I expand a json object

saltcod:
hit return too fast sorry ^^

get that error when I load full json → then click expand 
screenshot-2024-08-20-at-16 59 56

joshenlim (approved):
- RowEditor: Expand editor -> Loading large JSON on demand, then going back, then going back to expand editor, can confirm prompt to load full data is present (same for text)
- Can confirm expand editor CTA for text is working
  - On prod it wasn't working for a column that's unique somehow

PR #28781, opened 2024-08-21
Clarify language in `range` documentation

nanewalt:

YES

docs update

The python [range documentation](https://supabase.com/docs/reference/python/range) has the following issues:
1. Inconsistent naming - The parameter list uses `start` and `end` while the description uses `from` and `to`
2. Incorrect offset end description - The description states that the result is limited by "starting at an offset (`from`) and ending at the offset (`from + to`)". This is incorrect as the end index is just `to`. E.g. `range(20, 30)` returns 11 items (20-30), not 51 items (20-20+30)

## What is the new behavior?

1. Range indices are references as `from` and `to` in parameter list. This is aligned with the documentation for other languages.
2. Description states that range offset ends at just `to`

## Additional context

[relevant implementation](https://github.com/supabase/postgrest-py/blob/main/postgrest/base_request_builder.py#L598-L608)

### DIFF
--- a/apps/docs/spec/supabase_py_v2.yml
+++ b/apps/docs/spec/supabase_py_v2.yml
@@ -4762,9 +4762,9 @@ functions:
         type: string
         description: Set this to limit rows of foreign tables instead of the parent table.
     notes: |
-      Limit the query result by starting at an offset (`from`) and ending at the offset (`from + to`). Only records within this range are returned. This respects the query order and if there is no order clause the range could behave unexpectedly.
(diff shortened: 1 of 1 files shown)


silentworks:
This should remain with `start` and `end` as that is what the parameters are called in the Python postgrest library. If this documentation change should be made then there needs to be a change made to the `postgrest-py` library first.

ChrisChinchilla:
I think this was duplicated in https://github.com/supabase/supabase/pull/30989 which has been merged, let me know if that wasn't correct

PR #28797, opened 2024-08-21
Fix ai debug

saltcod:
To reproduce: 
- go to the SQL Editor and create a new snippet by adding `select * from no_exist;`
- refresh the page
- run the `select * from no_exist;` query
- when that fails, click Debug with AI 
- this should fail to even run and give you an error toast

### DIFF
--- a/apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
+++ b/apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
@@ -414,7 +414,15 @@ const SQLEditor = () => {
         )
       }
     }
-  }, [debugSql, entityDefinitions, id, snap.results, snap.snippets])
+  }, [
+    debugSql,
+    entityDefinitions,
+    id,
+    snap.results,
+    snap.snippets,
+    snapV2.results,
+    snapV2.snippets,
+  ])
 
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
```diff
@@ -414,7 +414,15 @@ const SQLEditor = () => {
...
     }
-  }, [debugSql, entityDefinitions, id, snap.results, snap.snippets])
+  }, [
+    debugSql,
+    entityDefinitions,
+    id,
+    snap.results,
+    snap.snippets,
+    snapV2.results,
+    snapV2.snippets,
```
joshenlim, line 424: Hmm maybe results and snippets could be one variable here up top like

const results = enableFolders ? snapV2.results : snap.results

Likewise for snippets - so we reduce the amount of vars in the dependency array

saltcod: hummm I think we need the snapv2.results in the array?  if we cache them at the top of this function, they won't trigger a change right? Or I'm misunderstanding

Flowsmooth (approved):

PR #28817, opened 2024-08-22
guard events

fsansalvadore:


apps/www/pages/_app.tsx
```diff
@@ -2,53 +2,47 @@ import '@code-hike/mdx/styles'
...
+      {
+        category: 'www',
+        action: 'page_views',
+        label: url,
       },
-    })
+      telemetryProps,
+      router
+    )
   }
```
phamhieu, line 43: why converting from `telemetry/page` to `/telemetry/event`?

phamhieu: on studio, we still use `/telemetry/page` https://github.com/supabase/supabase/blob/cd55df330df8d9415c7ff183aa92538c14c32e06/apps/studio/components/ui/PageTelemetry.tsx#L80-L90

phamhieu: `/telemetry/page` and `/telemetry/event` are different. I don't think we can create a page_view event to replace the call to `/telemetry/page`."

fsansalvadore: Ah you're right. I'm gonna revert this "refactoring" on the `www/_app` then.

pamelachia (approved):

4L3k51 (approved):

PR #28818, opened 2024-08-22
Update to 4o-mini

saltcod:
Update SQL Editor chat model

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4292,9 +4292,9 @@
       }
     },
     "node_modules/@babel/helper-plugin-utils": {
-      "version": "7.24.7",
-      "resolved": "https://registry.npmjs.org/@babel/helper-plugin-utils/-/helper-plugin-utils-7.24.7.tgz",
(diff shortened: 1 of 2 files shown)


kiwicopple (approved):

packages/ai-commands/src/sql/chat.ts
```diff
@@ -58,7 +58,7 @@ export async function chatSql(
 
   try {
     const response = await openai.chat.completions.create({
-      model: 'gpt-3.5-turbo-0125',
+      model: 'gpt-4o-mini',
```
ivasilov, line 61: We should use a specific version of `gpt-4o-mini-2024-07-18` from https://platform.openai.com/docs/models/gpt-4o-mini. This is to help with reproducibility.

package-lock.json
```diff
@@ -4292,9 +4292,9 @@
       }
```
ivasilov, line 4292: The package-lock.json doesn't need to be updated, copy it over from latest master.

ivasilov (approved):

Issue #22242, opened 2024-03-26
emailRedirectTo is sending the config URL rather than the custom URL


thereis:
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When using the JS SDK with the admin privileges, in a backend application, if you signup a new user and declare the `emailRedirectTo` to `localhost:3000`, the value that the backend will execute is the `Site URL` from the `URL Configuration`.

`localhost:3000` is already added to the allowed `Redirect URLs`.

`{{ .ConfirmationURL }}` is the variable declared in the `Email Template` in the `Confirm Signup` tab.

My current code:

```ts
const { data: supabaseUser, error: supabaseError } =
        await supabaseAdminClient.auth.signUp({
          email: input.email,
          password: input.confirmPassword,
          options: {
            emailRedirectTo: 'http://localhost:3000',
            data: {
              first_name: input.firstName,
              last_name: input.lastName,
            },
          },
        });
```

## Expected behavior

To send the e-mail with the custom URL.

## Screenshots

```
https://[REDACTED].supabase.co/auth/v1/verify?token=[REDACTED]&type=signup&redirect_to=https://[MY_PRODUCTION_URL_INSTEAD_OF_LOCALHOST]/auth/callback?token_hash=[REDACTED]&type=email
```

## System information

- OS: macOS
- Chrome
- Version of supabase-js: `^2.39.7`
- Version of Node.js: `v20.4.0`

UpDriveMD:
I have the same issue

bluedaniel:
This `emailRedirectTo` appears to do nothing, bit of an urgent bug isn't it?

Not sure if there's any workarounds for this currently

bluedaniel:
After lots of trial and error, the only way I got this working was to ensure there's a trailing slash in the `emailRedirectTo`

Not sure if the docs mention this (or if that solves other people's issues)

thereis:
Hello @bluedaniel, I will give it a try in a few days and I will let you know.

iantanwx:
@bluedaniel I had this issue developing locally; what fixed it was adding the callback url to `supabase/config.toml`. 


```
[auth]
enabled = true
# The base URL of your website. Used as an allow-list for redirects and for constructing URLs used
# in emails.
[...]

PR #28821, opened 2024-08-22
fix: add confirmation url for Send Email Auth Hook

J0:

YES

Update Auth Hook Send Email to include confirmation URL generation

### DIFF
--- a/apps/docs/content/guides/auth/auth-hooks/send-email-hook.mdx
+++ b/apps/docs/content/guides/auth/auth-hooks/send-email-hook.mdx
@@ -630,6 +630,11 @@ const templates = {
   },
 }
 
+function generateConfirmationURL(email_data) {
+  // TODO: replace the ref with your project ref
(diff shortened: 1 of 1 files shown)


LinusHojaGER:
@J0 At the moment i am having issues on my self hosted supabase, where i try to get deeplinks working for my app. On the cloud hosted supabase it works, but on my self hosted instance it is not injecting the right redirect url, it uses the default. Is your change fixing this behavior? If so, we should really try to get this merge thru. There is a discussion and an open issue, that would be fixed. 

Discussion: https://github.com/orgs/supabase/discussions/27250
Open issue: https://github.com/supabase/supabase/issues/22242

J0:
Hey @LinusHojaGER,

Thanks for raising this. I skimmed the thread and I think the issue is likely unrelated - it doesn't look like folks are using a Hook in the linked thread. It could be related to a misconfigured redirect, will look further and reply there

charislam (approved):

PR #28832, opened 2024-08-23
Fix missing region in replica primary node

joshenlim:
Fixes missing region in settings infra primary node for the new regions


alaister (approved):
Code LGTM!

- Opened preview for good measure 😅

PR #28835, opened 2024-08-23
Chore/table editor fixes

joshenlim:
Fixes a couple of issues with the table editor around updating cell values of columns that have a foreign key
- Fixes default sort behaviour on tables with composite primary keys, currently it only picks one of the columns from the composite PK and applies the sort on it, which means that the rows will still move around after updating a cell
  - Threshold of default sort to be applied only for tables below 50k rows still apply
- For a table that has 2 columns with foreign keys targeting the same table, fix to update the correct column when the foreign row is selected
- For a table that has a composite foreign key, support updating the foreign key columns from the grid editor itself (it currently works only via the RowEditor side panel)

### DIFF
--- a/apps/studio/components/grid/components/grid/Grid.tsx
+++ b/apps/studio/components/grid/components/grid/Grid.tsx
@@ -146,6 +146,7 @@ export const Grid = memo(
 
         const fk = data?.find(
           (key: any) =>
+            key.source_columns == columnName &&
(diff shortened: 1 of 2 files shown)


alaister (approved):
LGTM!

- Tested sorting on a table with a composite primary key
- Tested updating a foreign value on a table with multiple foreign keys
- Tested updating a composite foreign key from the table editor

PR #28838, opened 2024-08-23
Fix active state in withSidebar

joshenlim:


ivasilov (approved):
Tested on studio, works ok.

MildTomato (approved):

PR #28846, opened 2024-08-23
Chore/update opt in message

saltcod:
Add a warning if users aren't opted-in to sending anon data with their queries 

screenshot-2024-08-23-at-13 13 26

Allow them to opt-in right from the SQL Editor
screenshot-2024-08-23-at-13 14 17

Also adds a message inline in the chat — right after the first user message, and only shows one time per chat
screenshot-2024-08-23-at-13 59 21

Also adds context to the schema selector: 
screenshot-2024-08-23-at-18 54 30


apps/studio/components/interfaces/Organization/GeneralSettings/OptInToOpenAIToggle.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react'
```
saltcod, line 1: moved this so I could use it on the new modal and on the org page

alaister (approved):
LGTM!

- Tested on preview

PR #28877, opened 2024-08-26
Revert "docs: Update s3-uploads to include File URL Structure"

fsansalvadore:
Reverts supabase/supabase#28827


inian (approved):
how to access objects is documented here - https://supabase.com/docs/guides/storage/serving/downloads

PR #28897, opened 2024-08-26
fix: spend cap permissions tooltip

alaister:

Bug fix

Currently the tooltip won't show because the button is disabled. This PR introduces a wrapping span which separates the tooltips disabled state from the buttons disabled state.

Screenshot 2024-08-26 at 21 17 09


kevcodez (approved):
Tested locally

ivasilov (approved):

PR #28913, opened 2024-08-27
fix: sql editor folder name editing

alaister:

Bug fix

https://github.com/user-attachments/assets/441d37ee-b657-4e5d-a20d-8cca9e30544b

## What is the new behavior?

https://github.com/user-attachments/assets/7f9b215a-b172-4ea7-a9b1-b1517be79025

## Additional context

This also fixes another bug where a user would double click a on a new folder that hadn't been created yet. This would cause 'new-folder' to be sent in place of an actual uuid


apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SQLEditorTreeViewItem.tsx
```diff
@@ -83,6 +83,7 @@ export const SQLEditorTreeViewItem = ({
 
   // [Joshen] Folder contents are loaded on demand too
   const onOpenFolder = async (id: string) => {
+    console.log('2 id:', id)
```
alaister, line 86:
```suggestion
```

ivasilov (approved):
Tested on Studio, SQL editor.

PR #28935, opened 2024-08-28
fix: do not show compute badge on paused projects

kevcodez:


alaister (approved):
LGTM!

- Checked on preview with a paused project

PR #28939, opened 2024-08-28
fix layouts and make sure buttons are visible in smol screens

jordienr:
tip to test this:
- put browser in 150% zoom
- navigate screens
- make sure buttons are visible or can be scrolled to

the goal here is to make everything usable in smaller screens or for people using zoom, not necessarily pretty. 

screens updated:
- table editor (headers)
- sql editor (headers)
- table list (headers)
- db roles (headers)
- database functions (table)
- database triggers
- database types
- database indexes
- database publications
- sql editor results: moved max results to footer.


saltcod:
not sure what's different with these?

screenshot-2024-08-28-at-16 59 49

jordienr:
> not sure what's different with these?
> 
> screenshot-2024-08-28-at-16 59 49

Fixed spacing between insert/rls and made it scrollable
CleanShot 2024-08-29 at 11 01 04

joshenlim (commented):
Generally LGTM - tested on staging preview under 150% zoom
- Table editor header
- SQL editor with assistant panel open
- Database pages

Other issues that I think we can address separately
- API Docs side panel from the table editor covers the whole screen, cannot be closed on mobile (on desktop can be closed via Esc key)
- Database/types -> Docs button is out of alignment
- Database/indexes -> Docs button is out of alignment

Just need to double check on SQL Editor RE limit tooling before I approve as I don't see that UI anymore (was it removed or shifted?)

joshenlim (approved):
LGTM! SQL editor all okay now :)

PR #28942, opened 2024-08-28
feat: upgrade via pricing page

alaister:

feature

Users will often accidentally create new paid orgs when they meant to upgrade an exisiting org.

## What is the new behavior?

If a user is logged in, the buttons on the pricing page will show an "Upgrade now" button instead of "Get started". They will then be able to select an existing org they wish to upgrade and be redirected to the relevant page in the dashboard.


apps/www/lib/fetchWrapper.ts
```diff
@@ -1,18 +1,46 @@
...
+    headers.set('Authorization', `Bearer ${accessToken}`)
+  }
+
   return fetch(url, {
     method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
-      Accept: 'application/json',
-    },
+    headers: headers,
```
kevcodez, line 41:
```suggestion
    headers,
```

apps/www/lib/fetchWrapper.ts
```diff
@@ -1,18 +1,46 @@
...
+    Accept: 'application/json',
+  })
+
+  if (accessToken) {
+    headers.set('Authorization', `Bearer ${accessToken}`)
+  }
+
+  return fetch(url, {
+    method: 'GET',
+    headers: headers,
```
kevcodez, line 21:
```suggestion
    headers,
```

kevcodez (approved):
Tested on staging

apps/www/components/Pricing/PricingComparisonTable.tsx
```diff
@@ -327,54 +362,70 @@ const PricingComparisonTable = () => {
...
+                          )}
+                        </p>
+                      </span>
+                      <span className="flex flex-col justify-between h-full pb-2">
+                        {isUpgradablePlan && hasExistingOrganizations ? (
+                          <UpgradePlan
+                            organizations={organizations}
+                            onClick={() =>
+                              sendTelemetryEvent(
+                                gaEvents[`www_pricing_comparison_${plan.name.toLowerCase()}_mobile`]
```
fsansalvadore, line 400: This should be "not" `_mobile`, like line 416.
The "Upgrade plan"s could have their own separate event btw.

apps/www/components/Pricing/UpgradePlan.tsx
```diff
@@ -0,0 +1,158 @@
...
+                  </CommandGroup_Shadcn_>
+                </CommandList_Shadcn_>
+              </Command_Shadcn_>
+            </PopoverContent_Shadcn_>
+          </Popover_Shadcn_>
+        </DialogSection>
+
+        <DialogSection>
+          <DialogDescription className="text-xs">
+            Upon continuing, you will be redirected to the organization&apos;s billing page where
```
fsansalvadore, line 132: Maybe `By continuing you will...`?

fsansalvadore (approved):
Left two comments but approving to unblock.

apps/www/components/Pricing/PricingComparisonTable.tsx
alaister: Nice catch! Removed `_mobile` there and added a new event for the upgrade buttons :)

PR #28951, opened 2024-08-28
chore: link to upgrade existing org

kevcodez:
Billing papercut: Customers regularly create a new organization on a paid plan instead of upgrading their existing one. To mitigate that we will now show a small text allowing customers to jump into the subscription upgrade side panel of an existing organization.

Screenshot 2024-08-28 at 17 57 59

Screenshot 2024-08-28 at 17 58 04


alaister (approved):
LGTM!

- Tested on preview

PR #28967, opened 2024-08-29
Show log errors as different colors in chart

jordienr:
EDIT 02/09
Moving this back to draft, needs further testing and improvements in chart.

- if there are errors show them as different colors in the bar chart
- update the filter when you click the bar chart so that it shows the correct range

### DIFF
--- a/apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
+++ b/apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
@@ -296,6 +296,24 @@ const calcChartStart = (params: Partial<LogsEndpointParams>): [Dayjs, string] =>
   return [its.add(-extendValue, trunc), trunc]
 }
 
+function getErrorCondition(table: LogsTableName): string {
+  switch (table) {
+    case 'edge_logs':
+      return 'response.status_code >= 400'
+    case 'postgres_logs':
+      return "parsed.error_severity IN ('ERROR', 'FATAL', 'PANIC')"
+    case 'function_logs':
+      return "metadata.level IN ('error', 'fatal')"
+    case 'auth_logs':
+      return "metadata.level = 'error' OR metadata.status >= 400"
+    case 'function_edge_logs':
+      return 'response.status_code >= 400'
+    // Add conditions for other log types as needed
+    default:
+      return 'false' // Default to no errors if table type is unknown
+  }
+}
+
 /**
  *
  * generates log event chart query
@@ -308,13 +326,16 @@ export const genChartQuery = (
   const [startOffset, trunc] = calcChartStart(params)
   const where = genWhereStatement(table, filters)
 
+  const errorCondition = getErrorCondition(table)
+
   let joins = genCrossJoinUnnests(table)
 
   return `
 SELECT
 -- log-event-chart
   timestamp_trunc(t.timestamp, ${trunc}) as timestamp,
-  count(t.timestamp) as count
+  count(t.timestamp) as count,
+  LOGICAL_OR(${errorCondition}) as has_error
 FROM
   ${table} t
   ${joins}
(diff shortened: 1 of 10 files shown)


apps/studio/components/interfaces/Settings/Logs/LogsPreviewer.tsx
```diff
@@ -229,10 +230,18 @@ export const LogsPreviewer = ({
             <LogEventChart
               data={eventChartData}
               onBarClick={(isoTimestamp) => {
+                // from should be 1 hour before the bar
+                // to should be 1 hour after the bar
```
jordienr, line 234: not sure about this approach tbh

problem is, with current approach sometimes you click a bar and dont see the logs, because in the chart the error is at 11:12 and the timestamp you limited to was 11:00.

apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
```diff
@@ -296,6 +296,24 @@ const calcChartStart = (params: Partial<LogsEndpointParams>): [Dayjs, string] =>
...
+    case 'edge_logs':
+      return 'response.status_code >= 400'
+    case 'postgres_logs':
+      return "parsed.error_severity IN ('ERROR', 'FATAL', 'PANIC')"
+    case 'function_logs':
+      return "metadata.level IN ('error', 'fatal')"
+    case 'auth_logs':
+      return "metadata.level = 'error' OR metadata.status >= 400"
+    case 'function_edge_logs':
+      return 'response.status_code >= 400'
```
jordienr, line 310: gotta review this conditions, can be ignored for now, only tested in edge_logs

apps/studio/pages/project/[ref]/logs/edge-logs.tsx
```diff
@@ -19,6 +19,6 @@ export const LogPage: NextPageWithLayout = () => {
   )
 }
 
-LogPage.getLayout = (page) => <LogsLayout title="Edge Logs">{page}</LogsLayout>
+LogPage.getLayout = (page) => <LogsLayout title="API Gateway">{page}</LogsLayout>
```
Comment (jordienr) [inline apps/studio/pages/project/[ref]/logs/edge-logs.tsx:22]: we renamed it to API Gateway in the sidebar a while back but the title was still showing the old name

apps/studio/components/interfaces/Settings/Logs/LogsPreviewer.tsx
Ziinc: 1 hour diff might be a bit big if we're mitigating the second step differential, maybe just step back by 1 sec is enough

Ziinc: Would require knowing what interval type it is at though, if it's minutes then would need to step back by a min and so on...

apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
```diff
@@ -308,13 +326,16 @@ export const genChartQuery = (
...
+
   let joins = genCrossJoinUnnests(table)
 
   return `
 SELECT
 -- log-event-chart
   timestamp_trunc(t.timestamp, ${trunc}) as timestamp,
-  count(t.timestamp) as count
+  count(t.timestamp) as count,
+  LOGICAL_OR(${errorCondition}) as has_error
```
Ziinc, line 338: Not really necessary for logical_or if you aren't working with arrays, fine to just use bare condition I think wrapped in a parentheses

apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
Ziinc: Looks fine, best to verify on a staging project

Ziinc (approved):

alaister (approved):
LGTM!

- Tested on preview

PR #28971, opened 2024-08-29
chore: credit card enhancements

alaister:

chore/feat

## What is the new behavior?

If a credit card is expired or expiring, we now show a badge on that card:
Screenshot 2024-08-29 at 11 45 59

We also now show the default payment method on the invoices screen:
Screenshot 2024-08-29 at 12 45 11


kevcodez (approved):
Tested on staging

ivasilov (approved):

PR #28976, opened 2024-08-29
Chore/move sort out by itself

saltcod:
Moves the sorting hat (button) out of the "Search queries..." input for better discoverability 

screenshot-2024-08-29-at-14 05 05


joshenlim:
taking a look at this but i think we’ll likely need @MildTomato to chime in and this is sort of a design pattern decision here

the table editor + storage explorer also have this “sort” UX and them, so if we’re gonna change it here, we should also change it for the others too

saltcod:
Obsidian also has this on top of their file folders

screenshot-2024-09-04-at-23 19 21

PR #28978, opened 2024-08-29
chore: Update RateLimits.tsx

mansueli:
Make it clearer

YES

image

There's already hour mentioned in the other side.

## What is the new behavior?

## Additional context

### DIFF
--- a/apps/studio/components/interfaces/Auth/RateLimits/RateLimits.tsx
+++ b/apps/studio/components/interfaces/Auth/RateLimits/RateLimits.tsx
@@ -408,8 +408,8 @@ const RateLimits = () => {
                   <FormSectionLabel
                     description={
                       <p className="text-foreground-light text-sm">
-                        Number of sign up and sign-in requests that can be made per IP
-                        address (excludes anonymous users).
+                        Number of sign up and sign-in requests that can be made per IP address
+                        (excludes anonymous users).
                       </p>
                     }
                   >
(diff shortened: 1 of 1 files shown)


J0 (approved):
Thanks!

apps/studio/components/interfaces/Auth/RateLimits/RateLimits.tsx
```diff
@@ -408,8 +408,8 @@ const RateLimits = () => {
                   <FormSectionLabel
                     description={
                       <p className="text-foreground-light text-sm">
-                        Number of sign up and sign-in requests that can be made per hour per IP
-                        address (excludes anonymous users).
+                        Number of sign up and sign-in requests that can be made per IP address
+                        (excludes anonymous users).
```
alaister, line 412: We don't say anywhere that this is in a 5 minute interval though right?
```suggestion
                        Number of sign up and sign-in requests that can be made in a 5 minute interval per IP address
                        (excludes anonymous users).
```

mansueli: But the hint below the editable field shows the per hour rate. I am not sure where you are getting this 5 minutes from. 🤔

joshenlim: just chatted with Joel - our docs need some updating, alaister is right here RE the 5 minute interval although we currently don't mention about it in our docs

J0: Updated here: https://github.com/supabase/supabase/pull/29176/

ivasilov (approved):

PR #28996, opened 2024-08-30
feat: add saml encrypted assertion support

hf:
See: https://github.com/supabase/auth/pull/1752

Merge after infra and Auth version is deployed.


alaister (approved):
LGTM!

- Tested on preview with a dummy URL

PR #28999, opened 2024-08-30
fix: Fix the caching of the selected schema

ivasilov:
This PR fixes a bug where the user can add a table in wrong schema by changing the schema in another tab. To replicate the bug:

1. Open both tabs on `[supabase.com/project/_/database/tables](http://supabase.com/project/_/database/tables). They should be on the same schema. It's important that you have no ?schema=something at the end
2. Click new Table in tab 1. Take note what the side panel header says (Create a new table under public, for example)
3. Go to tab 2 and switch the schema to something else
4. When you go to tab 1 the header will change to the new schema


joshenlim (approved):
LGTM! Tested on staging preview as per reproduction steps
Can verify that table is created in the proper schema for tab 1

PR #29000, opened 2024-08-30
fix: Change all apps to use the same version of lucide-react package

ivasilov:


alaister (approved):
LGTM!

- Opened all previews

fsansalvadore (approved):
Checked docs and www previews, looks all good 👍

PR #29002, opened 2024-08-30
fix: Add CommandList wrapper to fix a crash in the cmdk feature

ivasilov:
This PR fixes a crash which happened on docs and studio when choosing the "Ask Supabase AI" option in the cmdk feature.


saltcod (approved):

saltcod:
checked all apps, looks good

PR #29034, opened 2024-09-02
Silence config cat warnings

joshenlim:
Done by basically just ensuring that we pass _some_ param into `getAllValuesAsync()` to silence the warnings
The reason why we're doing so as this has been known to cause a lot of red herrings whenever users are debugging UI issues (despite the config cat warnings being completely unrelated all the time)

### DIFF
--- a/apps/studio/lib/configcat.ts
+++ b/apps/studio/lib/configcat.ts
@@ -18,7 +18,5 @@ function getClient() {
 }
 
 export async function getFlags(user?: User) {
-  return user?.email !== undefined
-    ? await getClient().getAllValuesAsync(new configcat.User(user.email))
-    : await getClient().getAllValuesAsync(new configcat.User(''))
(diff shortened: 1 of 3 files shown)


apps/studio/lib/configcat.ts
```diff
@@ -20,5 +20,5 @@ function getClient() {
 export async function getFlags(user?: User) {
   return user?.email !== undefined
     ? await getClient().getAllValuesAsync(new configcat.User(user.email))
-    : await getClient().getAllValuesAsync()
+    : await getClient().getAllValuesAsync(new configcat.User(''))
```
kamilogorek, line 23:
```js
return await getClient().getAllValuesAsync(new configcat.User(user?.email ?? ''))
```

joshenlim: oh right that's way more succinct hahaha

ivasilov (approved):
Tested on studio, works ok.

PR #29047, opened 2024-09-03
fix: update rate limits in docs to 2

J0:

YES

We drop rate limits back to 2: https://github.com/supabase/infrastructure/pull/19614/files#diff-54be3ab8fa566471e8818985465fec1272f1b05a92fb752bfc55a83d539d2291


charislam (commented):
I think this should be updated via https://github.com/supabase/supabase/blob/878b8f44936312e8030fa34dc8096707d3736b86/packages/shared-data/config.ts#L9?

Unless it's talking about a different limit altogether

J0:
You're right - we should use the shared config. Thanks for catching

charislam (approved):

PR #29060, opened 2024-09-03
fix: cmdk error in AiMessages component

charislam:
- Add a dummy CommandList descendant
- Needed for cmdk to not throw an error
- No easy way to remove the Command Menu because it's needed in the empty state
- Somewhat hacktastic but solves the error

### DIFF
--- a/packages/ui-patterns/CommandMenu/prepackaged/DocsAi/DocsAiPage.tsx
+++ b/packages/ui-patterns/CommandMenu/prepackaged/DocsAi/DocsAiPage.tsx
@@ -177,71 +177,78 @@ function PromptInput({
 
 function AiMessages({ messages }: { messages: Array<Message> }) {
   return (
-    <div
-      className={cn(
-        'flex-grow min-h-0 overflow-auto p-4',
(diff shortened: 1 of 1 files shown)


packages/ui-patterns/CommandMenu/prepackaged/DocsAi/DocsAiPage.tsx
```diff
@@ -177,71 +177,78 @@ function PromptInput({
...
-                      'w-7 h-7',
-                      'border border-muted bg-background shadow-sm rounded-full',
-                      'flex items-center justify-center',
-                      'text-foreground-lighter'
+    <>
+      {/* cmdk menus need a CommandList descendent in order to not throw an
+      error. This display doesn't actually need the Command Menu, but it's
+      needed for the empty state to work with the input, so this is a somewhat
+      hacktastic way of making it not error. */}
+      <CommandList_Shadcn_ />
```
charislam, line 185: This is the only change, everything else is just a whitespace shift.

ivasilov (approved):

PR #29062, opened 2024-09-03
Chore/empty table errors

saltcod:
Empty tables currently cause the Table Editor to crash.

to reproduce: 
- create table example() in the sql editor
- goto the table editor, click this table and see a full page crash


apps/studio/data/table-rows/table-rows-query.ts
```diff
@@ -133,11 +133,14 @@ export const getTableRowsSqlQuery = ({
...
-    )
-    .map((column) => `"${column.name}"::text[]`)
+  const arrayBasedColumns =
+    table.columns.length > 0
+      ? table.columns
+          .filter(
+            (column) => (column?.enum ?? []).length > 0 && column.dataType.toLowerCase() === 'array'
+          )
+          .map((column) => `"${column.name}"::text[]`)
+      : []
```
joshenlim, line 143: nit: this one's an unnecessary check cause the filter will either way return an empty array (and the subsequent map will also return an empty array)

joshenlim (approved):
LGTM - just a tiny nit comment. Checked on staging preview
- Create a table via `create table example()`
- Open table in table editor - no crash

PR #29066, opened 2024-09-03
feat: improve timeouts for webhooks

mansueli:

YES

We are getting quite a few timeouts in the edge runtime and this will ensure a better default. 

The default timeout is 1s, which results in a high number of 499 requests that aren't logged in the edge runtime. 

## What is the new behavior?

This increases the default to 5s and the max to 10s. 

## Additional context


alaister (approved):
LGTM!

- Tested on preview

joewinke:
@mansueli thank you for pushing this - I was able to update the timeout to 10000ms. However cosmetically, the user ux shows a max of 5000 I think it should be updated to 10000.
Screenshot 2024-10-11 at 1 40 03 PM

PR #29092, opened 2024-09-04
feat: add configure org command

charislam:
Add a cmdk option to switch organizations.

Also fixes a bug found while testing where CJK IME input does not work with the Command Input.


apps/studio/components/interfaces/App/CommandMenu/OrgProjectSwitcher.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { useProjectsQuery } from 'data/projects/projects-query'
+import { useOrganizationsQuery } from 'data/organizations/organizations-query'
+import { PageType, useRegisterCommands, useRegisterPage, useSetPage } from 'ui-patterns/CommandMenu'
+import { IS_PLATFORM } from 'common'
+import { COMMAND_MENU_SECTIONS } from './CommandMenu.utils'
+
+const PROJECT_SWITCHER_PAGE_NAME = 'Switch project'
+const ORGANIZATION_SWITCHER_PAGE_NAME = 'Switch organization'
+
+export function useProjectSwitchCommand() {
```
charislam, line 13: This hook is just copied from the original ProjectSwitcher.tsx before the rename

alaister (approved):
LGTM!

- Tested on studio preview
- Also tested on www/docs to make sure it doesn't show up

PR #29120, opened 2024-09-05
feat(docs): main troubleshooting page

charislam:
feat: add troubleshooting pages to docs

preview: https://docs-git-feat-docs-troubleshooting-2-supabase.vercel.app/docs/guides/troubleshooting

not in nav yet until we port a significant amount of these over


apps/docs/app/guides/(with-sidebar)/[...slug]/page.tsx
Comment (charislam) [inline apps/docs/app/guides/(with-sidebar)/[...slug]/page.tsx:1]: Moved all of these under a nested directory so we could make the troubleshooting pages have a different layout

apps/docs/components/Navigation/NavigationMenu/NavigationMenu.constants.ts
```diff
@@ -172,6 +172,11 @@ export const GLOBAL_MENU_ITEMS: GlobalMenuItems = [
       label: 'Resources',
       menuItems: [
         [
+          /*           {
```
charislam, line 175: will eventually go here but hidden for now because we don't have many migrated yet

apps/docs/components/Navigation/NavigationMenu/NavigationMenu.constants.ts
```diff
@@ -172,6 +172,11 @@ export const GLOBAL_MENU_ITEMS: GlobalMenuItems = [
...
+            label: 'Troubleshooting',
+            icon: 'contributing',
```
charislam, line 177: @fsansalvadore any suggestions for icons? currently contributing uses the wrench icon, which probably should be transferred to troubleshooting, but then I'm not sure what to use for contributing

apps/docs/content/troubleshooting/monitor-supavisor-postgres-connections.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+[[errors]]
+http_status_code = 403
+code = "forbidden"
+message = "Not allowed to access this resource"
+
+[[errors]]
+http_status_code = 500
+code = "internal_server_error"
+message = "Internal server error"
```
charislam, line 27: Gonna remove these later as none of them are actually relevant to the issue, just needed some dummy data to see how the display looks

apps/docs/app/guides/(with-sidebar)/[...slug]/page.tsx
Comment (charislam) [inline apps/docs/app/guides/(with-sidebar)/[...slug]/page.tsx:1]: moved all of these under a nested directory so we could have a sidebarless layout for troubleshooting

apps/docs/components/Navigation/NavigationMenu/NavigationMenu.constants.ts
fsansalvadore: `Contributing` could either be a simple _heart_ or a _"branching"_ icon, maybe?

Screenshot 2024-09-11 at 14 24 08

And I agree troubleshooting would be the wrench.

fsansalvadore:
@charislam what if the filtering form and topics were in the sidebar? 
It would keep the layout more consistent.

Plus, can we also reuse the breadcrumbs component?

charislam:
@fsansalvadore what do you think with the sidebar?

image

image

fsansalvadore (approved):
The mobile version needs a bit of design love, but since this isn't user facing yet we can keep it moving 👍 

we need Dashboard team to approve also, not sure why that `apps/studio/next-env.d.ts` keeps changing...

saltcod (approved):

apps/docs/content/troubleshooting/monitor-supavisor-postgres-connections.mdx
charislam, line 27:
```suggestion
```

PR #29135, opened 2024-09-06
fix: Clarify that you can no longer use auth.uid() in RLS policies when using Auth0 Third Party Auth

dagingaa:

YES

This change modifies the limitations for Auth0 Third Party Auth to include that you can no longer use the `auth.uid` function in RLS policies, and need to update these to instead use `auth.jwt() -> 'sub'`

`auth.uid()` requires the return type of the jwt sub claim to be a uuid, and will throw an `invalid input syntax for type uuid` error when you attempt to use it for RLS policies.

In addition, according to https://github.com/supabase/auth/pull/484/files we should be using auth.jwt() always anyway, but that message hasn't made it to the docs.

Currently this is not informed in the docs which is confusing.

## What is the new behavior?

Users understand the limitations


charislam:
@J0 a quick one just for you to double-check 🙏🏼

PR #29136, opened 2024-09-06
Chore/persist schema visualizer node positions

joshenlim:
- Persist positions of nodes in the schema visualizer by storing in local storage
- Add a button to automatically arrange the layout of the nodes in the schema visualizer
image
- Note: new tables added into the schema will have their positions default to `n * 10, n * 10` where n is the idx for each new table,  if there's an existing set of stored data of positions for the schema in local storage
  - Users can either manually drag the nodes to the desired location OR
  - Hit auto layout to let the dagre library handle the layout (which is the original default behaviour everytime a user opens the schema visualizer)

### DIFF
--- a/apps/studio/components/interfaces/Database/Schemas/Schemas.utils.ts
+++ b/apps/studio/components/interfaces/Database/Schemas/Schemas.utils.ts
@@ -1,0 +1,211 @@
+import dagre from '@dagrejs/dagre'
+import type { PostgresSchema, PostgresTable } from '@supabase/postgres-meta'
+import { LOCAL_STORAGE_KEYS } from 'lib/constants'
+import { uniqBy } from 'lodash'
+import { Edge, Node, Position } from 'reactflow'
+import 'reactflow/dist/style.css'
+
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Database/Schemas/Schemas.utils.ts
```diff
@@ -0,0 +1,223 @@
...
+
+export async function getGraphDataFromTables(
+  ref: string,
+  schema: PostgresSchema,
+  tables: PostgresTable[]
+): Promise<{
+  nodes: Node<TableNodeData>[]
+  edges: Edge[]
+}> {
+  if (!tables.length) {
```
saltcod, line 34: is this right?

saltcod (approved):

apps/studio/components/interfaces/Database/Schemas/Schemas.utils.ts
joshenlim: yup! this is just to handle empty state

PR #29145, opened 2024-09-06
Replace 3.5 with 4o

saltcod:


charislam (approved):
Ran AI responses on all site previews, tried RLS AI assistant in studio

PR #29169, opened 2024-09-09
Minor improvement to schema visualizer layout of nodes without saved positions

joshenlim:
Instead of default the starting position of nodes without saved positions to 0, 0
We default it to be in the negative y axis - that way for layouts that were originally auto layout-ed via dagre, it'll not overlap with the first node at 0, 0

This is still in no way optimal understandably, will see how we can iterate over this to make it better


alaister (approved):
LGTM!

- Opened schema visualizer on preview

saltcod (approved):

PR #29170, opened 2024-09-09
Allow access to the database backups page irregardless of project status

joshenlim:
As per PR title, but except for paused, pause failed and pausing states


alaister (approved):
LGTM!

- Tested on preview with projects in various states

PR #29173, opened 2024-09-09
Add client side validation for payload size in sql editor

joshenlim:
Prevent users from running into "request body is too large" errors (and also save an unnecessary hammer on the API)
image


alaister (approved):
LGTM!

- Tested on preview with a giant query

PR #29181, opened 2024-09-09
chore: exclude Log Drains and MFA Phone from list of removed addons in downgrade modal

ecktoteckto:

Adjusts copy on "Downgrade to Free" modal regarding addons that will be removed.

Log Drains and MFA Phone are listed as addons that will be removed from the project once downgrading.

before

## What is the new behavior?
Log Drains and MFA Phone are no longer listed.

after

**Additional**
- fixed typo in Adanved MFA Phone section
- fixed spacing on Project Transfer modal


kevcodez (approved):
Tested on staging that modal still works fine

PR #29188, opened 2024-09-10
feat: show new org page on first login for subset of users

loong:

YES

More user guidance, showing new org creation page right after first time login for a subset of users


apps/studio/pages/projects.tsx
```diff
@@ -39,6 +40,15 @@ const ProjectsPage: NextPageWithLayout = () => {
       else if (localStorageSlug && verifiedSlug) router.push(`/org/${localStorageSlug}`)
       else router.push(`/org/${organizations[0].slug}`)
     }
+
+    if (orgCreationExperimentGroup === 'group-b' && !navLayoutV2 && isSuccess && hasWindowLoaded) {
+      // navigate to new page exactly once
+      const hasShownNewPage = localStorage.getItem('newPageShown')
+      if (!hasShownNewPage && organizations.length === 0) {
+        localStorage.setItem('newPageShown', 'true')
```
loong, line 48: not sure if this is a good idea

joshenlim: hmm just to double check:
- this intends to only target a subset of _new_ users who just signed up?
- note that a default org is already created for new users (just that it happens async), so once the new user creates a new org with this flow, it means that the user will end up with 2 orgs yeah?

loong: whoops didn't reply here

I made another commit to keep the local storage key where you suggest.

>this intends to only target a subset of new users who just signed up?
Yes only new users and only a subset. We use configcat to split the groups and run an a/a/b test.

>note that a default org is already created for new users (just that it happens async), so once the new user creates a new org with this flow, it means that the user will end up with 2 orgs yeah?
There's another infra PR (already approved) that disables the auto-creation

joshenlim (approved):

PR #29195, opened 2024-09-10
feat: Add a commit sha to the headers in studio app

ivasilov:
This PR adds the commit SHA from which the Studio app was built to the headers.


alaister (approved):
LGTM!

- Verified that this header is sent on preview

kevcodez:
This likely won't land in api due to api gateway header allow list, see slack thread

PR #29204, opened 2024-09-10
Chore/sql editor fixes

joshenlim:
- Add new folder: Add validation for creating a new folder of the same name as one of the existing folders
- Move query: Add validation for creating a new folder of the same name as one of the existing folders
- Move query: Add validation for new folder name to be minimum length of 1
- Rename folder: Handle state update back to idle if update failed (due to duplicate name)
- Rename folder: Add validation for renaming a folder to a same name as one of the existing folders


alaister (approved):
LGTM!

- Tested as per PR description on preview

PR #29166, opened 2024-09-09
feat: New connection management docs

TheOtherBrian1:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

A complete rewrite of our connecting to database docs

## What is the current behavior?

Everything related to connections is shoe horned into the connecting to database docs and it has become unruly.

## What is the new behavior?

Rewrites the connecting to database docs to be easier to read:
- Moves sections to connect to ORMs into their own guides (new docs)
  -  Adds one for Prisma 
  -  Adds a Prisma troubleshooting guide (a bit dense - could be streamlined)
- Moves Supavisor debugging section into the troubleshooting guides (new docs)
- Creates a connection management guide (new docs)
- Updates the IPv4 Add-On page to address everything Postgres/IPv4 related

PR #29210, opened 2024-09-10
fix: clarify that ipv4 now billed hourly

charislam:
ipv4 is now billed hourly at the prorated former monthly rate (see [changelog](https://github.com/orgs/supabase/discussions/28438)), but we don't mention this in the dashboard, which discourages quick experimentation

context here: https://github.com/supabase/supabase/pull/29166#discussion_r1750596095

image

### DIFF
--- a/apps/studio/components/interfaces/Settings/Addons/IPv4SidePanel.tsx
+++ b/apps/studio/components/interfaces/Settings/Addons/IPv4SidePanel.tsx
@@ -172,11 +172,12 @@ const IPv4SidePanel = () => {
                       <p className="text-foreground-light">
                         Allow direct database connections via IPv4 address
                       </p>
-                      <div className="flex items-center space-x-1 mt-2">
+                      <div className="grid grid-cols-[max-content,1fr] space-x-1 mt-2">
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Settings/Addons/IPv4SidePanel.tsx
```diff
@@ -172,11 +172,12 @@ const IPv4SidePanel = () => {
...
                         Allow direct database connections via IPv4 address
                       </p>
-                      <div className="flex items-center space-x-1 mt-2">
+                      <div className="grid grid-cols-[max-content,1fr] space-x-1 mt-2">
                         <p className="text-foreground text-sm">{formatCurrency(option.price)}</p>
-                        <p className="text-foreground-light translate-y-[1px]">
+                        <p className="text-foreground-light translate-y-[2px]">
                           / month / database
                         </p>
+                        <span className="col-start-2 text-foreground-light">(billed hourly)</span>
```
kevcodez, line 180: This is only true for Orb-based customers

Orb-based customers also have a note at the bottom of the modal

> There are no immediate charges. The addon is billed at the end of your billing cycle based on your usage and prorated to the hour.

PR #29219, opened 2024-09-11
Chore/disk attributes mangement

joshenlim:
Updates new UI for disk management under settings/database
image


apps/studio/components/interfaces/DataWarehouse/FormFooterChangeBadge.tsx
```diff
@@ -0,0 +1,43 @@
+import { AnimatePresence, motion } from 'framer-motion'
+import { Badge } from 'ui'
+
+interface FormFooterChangeBadgeProps {
+  formState: {
+    dirtyFields: Record<string, boolean>
+  }
+}
+export const FormFooterChangeBadge = ({ formState }: FormFooterChangeBadgeProps) => {
```
alaister, line 9: Is this actually used anywhere in this PR?

alaister (commented):
NaN while input is empty:
Screenshot 2024-09-11 at 11 48 53

alaister (commented):
spacing between the text and the warning is toight
Screenshot 2024-09-11 at 11 52 53

apps/studio/components/interfaces/DataWarehouse/FormFooterChangeBadge.tsx
joshenlim: yup! DiskManagementPanelForm L566

joshenlim:
> NaN while input is empty:
I'm actually not sure how to customize the error message here, so maybe we leave it as it is for now (altho not optimal understandably). PostgrestConfig under settings/api also has this same issue too

joshenlim:
@alaister pushed up the fix for the other comment :)

packages/ui/src/components/radio-group-card.tsx
```diff
@@ -31,17 +31,17 @@ const RadioGroupCardItem = React.forwardRef<
...
         'flex flex-col gap-2',
         'w-48',
-        'bg-overlay/50',
+        'bg-overlay',
         'rounded-md',
         'border',
         'p-2',
-        'shadow-sm',
-        'hover:bg-surface-300',
+        // 'shadow-sm dark:shadow-none',
```
alaister, line 38:
```suggestion
```

alaister (approved):
LGTM!

A few non-blocking nits:

apps/studio/components/interfaces/DiskManagement/DiskManagementPanelForm.tsx
```diff
@@ -0,0 +1,600 @@
...
+  const throughputPrice = calculateThroughputPrice({
+    storageType: form.getValues('storageType') as DiskType,
+    newThroughput: form.getValues('throughput') || 0,
+    oldThroughput: form.formState.defaultValues?.throughput || 0,
+  })
+
+  useEffect(() => {
+    // Initialize field values properly when data has been loaded
+    if (isSuccess) form.reset(defaultValues)
+  }, [isSuccess])
```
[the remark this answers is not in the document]
joshenlim: oh yeah you're right, this useEffect is unnecessary - will remove

apps/studio/components/interfaces/DiskManagement/DiskManagement.constants.tsx
```diff
@@ -0,0 +1,53 @@
...
+    maxIops: 256000,
+  },
+}
+
+export const DISK_TYPE_LABELS = {
+  [DiskType.GP3]: 'General Purpose SSD (gp3)',
+  [DiskType.IO2]: 'Provisioned IOPS SSD (io2)',
+}
+
+type PlanType = 'free' | 'pro' | 'team' | 'enterprise'
```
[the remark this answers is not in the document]
joshenlim: yup! you're right - can get it from API types

apps/studio/components/interfaces/DiskManagement/DiskManagementPanelForm.tsx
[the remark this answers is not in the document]
joshenlim: hmm... actually thinking might need this. if i put this in onSuccess its causing the following issue:

Divit ran into an issue that i couldn't reproduce such that the form value resets when he switches tabs?
i'm thinking its because RQ is just revalidating and triggers onSuccess, whereas I'd only want to set the form once (which can be achieved with the useEffect since the value of isSuccess doesnt change)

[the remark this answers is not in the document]
joshenlim: actually we might need this useEffect, was previously thinking the onSuccess was sufficient

Divit is running into an issue that i could repro, such that if we edit the form, go to another tab and wait a bit then come back to the same tab, I think RQ revalidates and triggers the onSuccess, hence resetting the form (we only want the form to reset once at page load, and if we were waiting for the requested changes to be enacted)

if we use the useEffect then we can guarantee it runs only once, cause the value of isSuccess will not change

apps/studio/components/interfaces/DiskManagement/DiskManagementPanelForm.tsx
```diff
@@ -0,0 +1,602 @@
...
+                                  shouldValidate: true,
+                                })
+                                trigger('provisionedIOPS')
+                                trigger('throughput')
+                              }}
+                              min={includedDiskGB}
+                            />
+                          </FormControl_Shadcn_>
+                          <div className="border border-strong bg-surface-300 rounded-r-md px-3 flex items-center justify-center">
+                            <span className="text-foreground-lighter text-xs font-mono">GiBs</span>
```
delgado3d, line 480:
```suggestion
                            <span className="text-foreground-lighter text-xs font-mono">GiB</span>
```

apps/studio/components/interfaces/DiskManagement/DiskManagementPanelForm.tsx
```diff
@@ -0,0 +1,602 @@
...
+                          afterPrice={Number(diskSizePrice.newPrice)}
+                          show={
+                            formState.isDirty &&
+                            formState.dirtyFields.totalSize &&
+                            !formState.errors.totalSize &&
+                            diskSizePrice.oldPrice !== diskSizePrice.newPrice
+                          }
+                        />
+                        <div className="text-xs text-foreground-light">
+                          Your plan includes {includedDiskGB}GB of disk size.
```
delgado3d, line 516:
```suggestion
                          Your plan includes {includedDiskGB}
                          GiB of disk size.
```

delgado3d, line 516:
```suggestion
                          Your plan includes {includedDiskGB}GiB of disk size.
```

apps/studio/components/interfaces/DiskManagement/DiskSpaceBar.tsx
```diff
@@ -0,0 +1,200 @@
...
+                <Tooltip_Shadcn_>
+                  <TooltipTrigger_Shadcn_ asChild>
+                    <div className="absolute right-full bottom-0 border mr-2 px-2 py-1 bg-surface-400 rounded text-xs text-foreground-light whitespace-nowrap">
+                      Autoscaling
+                    </div>
+                  </TooltipTrigger_Shadcn_>
+                  <TooltipContent_Shadcn_ side="bottom" className="w-[310px] flex flex-col gap-y-1">
+                    <p>
+                      Supabase expands your disk storage automatically when the database reached 90%
+                      of the disk size. However, auto-scaling can only take place once every 6
```
delgado3d, line 173:
```suggestion
                      of the disk size. However, any disk modifications, including auto-scaling can only take place once every 6
```

PR #29223, opened 2024-09-11
Project filter status, active will only show active projects, not all

joshenlim:
Currently having the filter with only "Active" still shows _all_ projects, this fixes it to only show active projects
image


alaister (approved):
LGTM!

- Tested on preview

PR #29224, opened 2024-09-11
fix: usage summary for stripe org [GEN-10963]

ecktoteckto:

Fixes the usage summary for Stripe orgs

Cells for metrics that are not returned by the Mgmt-API are not displayed correctly. For the metrics "Monthly Active Third Party Users" and "Log Drain Events" no usage data is returned from Mgmt-API because these metrics do not exist for Stripe orgs.

10963-before

### Free Plan
10963-after-free

### Pro Plan
10963-after-pro


kevcodez (approved):
Looks good on staging

PR #29232, opened 2024-09-11
Add docs for pgmq

olirice:
We're adding the pgmq extension https://github.com/supabase/postgres/pull/1120

This PR adds docs for the extension at `Database -> Extensions -> pgmq: Queues`

These just the extension docs. For any "Supabase Queue" product style docs we'll add a separate "Product" with dedicated docs/examples

### DIFF
--- a/apps/docs/content/guides/database/extensions/pgmq.mdx
+++ b/apps/docs/content/guides/database/extensions/pgmq.mdx
@@ -6,7 +6,6 @@ description: 'pgmq: Managed queues in Postgres'
 
 pgmq is a lightweight message queue built on Postgres.
 
-
 ## Features
 
 - Lightweight - No background worker or external dependencies, just Postgres functions packaged in an extension
@@ -14,7 +13,6 @@ pgmq is a lightweight message queue built on Postgres.
 - API parity with AWS SQS and RSMQ
 - Messages stay in the queue until explicitly removed
 - Messages can be archived, instead of deleted, for long-term retention and replayability
-
 
 ## Enable the extension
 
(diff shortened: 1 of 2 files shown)


apps/docs/content/guides/database/extensions/pgmq.mdx
```diff
@@ -0,0 +1,708 @@
...
+- Lightweight - No background worker or external dependencies, just Postgres functions packaged in an extension
+- "exactly once" delivery of messages to a consumer within a visibility timeout
+- API parity with AWS SQS and RSMQ
+- Messages stay in the queue until explicitly removed
+- Messages can be archived, instead of deleted, for long-term retention and replayability
+
+## Enable the extension
+
+```sql
+create extension queue;
```
charislam, line 20: is this a separate extension that's a dependency or did you mean `create schema`?

apps/docs/content/guides/database/extensions/pgmq.mdx
```diff
@@ -0,0 +1,708 @@
...
+          2
+```
+
+---
+
+### Reading Messages
+
+#### read
+
+Read 1 or more messages from a queue. The VT specifies the amount of time in seconds that the message will be invisible to other consumers after reading.
```
charislam, line 105:
```suggestion
Read 1 or more messages from a queue. The VT specifies the delay in seconds between reading and the message becoming invisible to other consumers.
```

charislam (approved):

apps/docs/content/guides/database/extensions/pgmq.mdx
olirice: good catch, that was meant to be `create schema queue`;
ty

PR #29245, opened 2024-09-12
Improve table editor pagination input field

joshenlim:
UX improvement - allows proper backspacing in the input field, and only go to page upon enter


alaister (changes requested):
It seems value is not updated when using the buttons instead of typing:

Screenshot 2024-09-12 at 16 44 13

alaister (approved):
LGTM!

- Tested locally

PR #29254, opened 2024-09-12
fix: move Auth error codes to top level

J0:

YES

Moves the Auth error codes

internal context: https://supabase.slack.com/archives/C022071RB2L/p1726139029552159


apps/docs/spec/common-client-libs-sections.json
```diff
@@ -402,6 +402,7 @@
         "slug": "auth-error-codes",
         "type": "markdown",
         "excludes": [
+          "reference_javascript_v2",
```
J0, line 405: Opted not to exclude entirely just yet as [Python uses error codes](https://supabase.com/docs/reference/python/auth-error-codes) and the implementation might differ. Can remove after checking with library owner

apps/docs/content/guides/auth/debugging/error-codes.mdx
```diff
@@ -11,7 +20,10 @@ Errors originating from the server API classed as `AuthApiError` always have a `
 
 <AuthErrorCodesTable />
```
J0, line 21: We can probably move the status codes out similar to what storage does. Are we okay with classifying them as "Legacy Error Codes"?

charislam: Are they actually legacy? Or are they always going to convey the same information (just that this information is too coarse-grained to be useful)? We could call them "HTTP status codes" and put them beneath the other error codes

apps/docs/content/guides/auth/debugging/error-codes.mdx
```diff
@@ -1,3 +1,12 @@
+---
+id: 'auth-error-codes'
+title: 'Error Codes'
+description: 'Supabase Auth Error Codes'
+subtitle: 'Learn about the Auth error codes and how to resolve them'
+---
+
+## Auth Error Codes
+
 Supabase Auth can throw or return various errors when using the API. All errors originating from the `supabase.auth` namespace of the client library will be wrapped by the `AuthError` class.
```
charislam, line 10: One problem with this brought up by Tyler (and why he abstracted the AuthErrorCodes table in the first place) is that the intro is too JavaScript-centric (the listed classes are all JavaScript-based, for example). We might need to have tabs for the different language details

e.g., this is what Tyler wrote for [Dart](https://docs-git-j0-moveerrorcodes-supabase.vercel.app/docs/reference/dart/auth-error-codes)

apps/docs/spec/common-client-libs-sections.json
charislam: See comment above, the error codes table should be shared, we just need to move the intro text into tabs I think

apps/docs/content/guides/auth/debugging/error-codes.mdx
J0: I think they convey the same information but the information is too coarse-grained. HTTP status codes sounds fine yup

charislam (approved):

PR #29259, opened 2024-09-12
Chore: Add links to documentation to the  Auth Providers page

monicakh:

- Added a "Learn more" button with links to setup guides for each auth provider.

### DIFF
--- a/apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersForm.types.ts
+++ b/apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersForm.types.ts
@@ -8,6 +8,7 @@ export interface Provider {
   $schema: string
   type: 'object'
   title: string
+  link: string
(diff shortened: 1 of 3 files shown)


joshenlim:
hmm this button is oddly placed - i can take a look at this tomorrow to see where might be a better position
image

charislam (approved):
Approving to unblock, with one minor nit about the layout

Would it look less obtrusive if we put the button in the header like so?
Screenshot 2024-09-12 at 11 33 51

(just a suggestion)

joshenlim:
i think lets do it like this:
image

am thinking you'd typically want docs for a provider only when you're planning to edit it, so putting it within the Collapsible will be better. That and it doesn't distract the "Enabled and "Disabled" status badge too - it looks a bit too cluttered otherwise
image

joshenlim (approved):
LGTM - tested on staging preview that each documentation button works

PR #29264, opened 2024-09-12
fix(redirects): remove unnecessary redirect

charislam:
We actually have a page for this now on the docs site, so we can remove it.


alaister (approved):
LGTM!

- checked the `/docs/reference` route to ensure the redirect no longer occurs

PR #29268, opened 2024-09-12
refactor(docs): convert self-hosting to app router

charislam:


fsansalvadore (approved):
All self-hosting pages seem to work fine looking at the preview 👍

PR #29272, opened 2024-09-12
fix(cmdk): empty state shows up when it shouldn't

charislam:
Can't remember why on earth I put this in a useMemo instead of a useEffect...

To reproduce:

There's a race condition somewhere, so it can be hard to reproduce the bug. One method that works ~90% of the time for me:

1. Open command menu in a project with branching enabled.
2. Use the command menu to switch branches.
3. Open command menu and search "branch" again. You might get the bug at this stage, or you might get the empty state (unrelated bug, this should show other available branches). If you don't get the bug yet, continue:
4. Press "Escape" to clear the search (only works with Escape, not Backspace!)
5. See the empty state showing up on top of a clearly non-empty list:
image

Fixed:
image

### DIFF
--- a/apps/studio/components/interfaces/BranchManagement/Branch.Commands.tsx
+++ b/apps/studio/components/interfaces/BranchManagement/Branch.Commands.tsx
@@ -17,7 +17,7 @@ export function useBranchCommands() {
 
   let { data: branches } = useBranchesQuery(
     {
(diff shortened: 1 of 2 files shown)


joshenlim (approved):
LGTM - tested on the staging preview to switch branches via cmdk

PR #29282, opened 2024-09-13
prevent use of WITH, ILIKE or wildcards (*) in logs explorer

jordienr:
to test
go to logs explorer
try a query with a WITH/ILIKE/wildcard statement
should show new error and send you to docs.

```
WITH test AS
 select
 cast (timestamp as datetime) as timestamp,
  event_message,
  metadata
from postgres_logs
where
  NOT REGEXP_CONTAINS (event_message, 'n')
SELECT event_message FROM test;
```

### DIFF
--- a/packages/ui/src/components/CodeBlock/CodeBlock.tsx
+++ b/packages/ui/src/components/CodeBlock/CodeBlock.tsx
@@ -3,7 +3,7 @@
 import { Check, Copy } from 'lucide-react'
 import { useTheme } from 'next-themes'
 import { Children, ReactNode, useState } from 'react'
-import * as CopyToClipboard from 'react-copy-to-clipboard'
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/CodeBlock/CodeBlock.tsx
```diff
@@ -3,7 +3,7 @@
 import { Check, Copy } from 'lucide-react'
 import { useTheme } from 'next-themes'
 import { Children, ReactNode, useState } from 'react'
-import * as CopyToClipboard from 'react-copy-to-clipboard'
+import { CopyToClipboard } from 'react-copy-to-clipboard'
```
jordienr, line 6: this import breaks vitest for some reason

saltcod (approved):

PR #29287, opened 2024-09-13
chore: pnpm

alaister:


.github/workflows/docs-mgmt-api-update.yml
```diff
@@ -11,31 +11,37 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          ref: master
+          sparse-checkout: |
+            apps/docs
+
+      - uses: pnpm/action-setup@v4
+        name: Install pnpm
+        with:
+          version: 9.10.0
```
ivasilov, line 23: We shouldn't have the version here. According to the [docs](https://github.com/pnpm/action-setup?tab=readme-ov-file#version), it'll grab it from the package.json.

alaister:
Closing for now as this will be easier now that https://github.com/supabase/supabase/pull/29610 has been merged

PR #29293, opened 2024-09-13
refactor(docs): move last pages over to app router

charislam:
The end of the Great Migration is here!!!

Moves the last pages over, deleting pages like the FAQ that we don't use and that contain duplicated information anyway.

Dev secret auth page URL had to change as App Router doesn't like the leading underscores in the path.

Also fixes the not-found recommendations to use the proper Next.js not-found page so it will return a 404 as it should. (I was under the erroneous impression that I couldn't get the pathname in not-found.tsx, that is not true, so this works better.)


fsansalvadore (approved):
🔥

PR #29313, opened 2024-09-16
Chore/disk mgt UI updates

joshenlim:
- Refrain from using "GB", all to use "GiB"
- Refrain from using "TiB", all to use "GiB"
- Add note to IOPS input field RE final usable IOPS, if value exceeds the max IOPS supported by the current compute size
- Adjust language for IOPS and Throughput fields to be more accurate regarding their max values, and added tooltips for more details and clarity


alaister (approved):
LGTM!

- Clicked through on staging

Small visual bug I did notice:
Screenshot 2024-09-16 at 13 45 19
The reset button is being slightly cut off

joshenlim:
great catch - i'll shift the description down, so that it supports most viewport widths
image

PR #29321, opened 2024-09-16
ref: Use local state to improve GitHub integration security

kamilogorek:
Use locally-stored random value to prevent CSRF issues.

docs: https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps#web-application-flow
ref: https://www.notion.so/supabase/GitHub-integration-should-use-state-parameter-for-authorization-eefcb2a9680c44fc951c2b1baecebb96?pvs=4


alaister (approved):
LGTM!

- Tested locally

PR #29326, opened 2024-09-16
Chore/fix move folder 2

saltcod:
Fixes the broken move SQL snippets

To reproduce: 
- create a folder using the top, create folder button
- try moving a query into that folder

screenshot-2024-09-16-at-12 52 25


joshenlim (approved):
LGTM! Tested on staging preview moving both single and multiple queries in and out of folders

PR #29333, opened 2024-09-17
Chore/more disk mgt UI updates

joshenlim:
- Only show new Disk Mgt UI for AWS projects
- Fix autoscaling marker to 90% of disk size
- Change GiB to GB, MiB to MB and MiBps to MB/s


alaister (approved):
LGTM!

- Tested on preview

PR #29337, opened 2024-09-17
fix: error on invalid mime types

alaister:

Bug fix

If an invalid mime type is uploaded, the storage explorer will continually retry.

## What is the new behavior?

Screenshot 2024-09-17 at 15 07 03


joshenlim (approved):
LGTM - tested on staging preview that the error surfaces and upload no longer retries
JFYI as well the multiple toasts isn't ideal but we'll look into that when we refactor the storage store

PR #29339, opened 2024-09-17
Support downloading storage objects in project paused state

joshenlim:
As per PR title, corresponding BE PR: https://github.com/supabase/infrastructure/pull/19728
- Am putting downloading backup + storage objects within a dropdown instead
- Screenshot below shows the state if there's no storage archive available for download, with the corresponding tooltip to communicate that
- Otherwise the dropdown item will not be disabled and can be downloaded by clicking (tested locally by hardcoding a ZIP file URL)
image


alaister (approved):
LGTM!

- Tested locally as instructed

w3b6x9:
@joshenlim what happens to the button and messaging when storage is >= 2gb?

joshenlim:
@w3b6x9 so when the user hits "Download storage objects", it first triggers a POST to /archive, which will validate the 2GB on the API end. If the storage has more than 2GB, FE will show a toast to reflect that message from API `The project storage size of ${storageSizeGb} GB exceeds the limit for the storage archive."

PR #29342, opened 2024-09-17
fix: specify which auth0 package to import for tpa

J0:

YES

Specifies package to import for Auth0 tpa

## Additional context

Initialization options taken [from this repo](https://github.com/auth0/auth0-spa-js)

### DIFF
--- a/apps/docs/content/guides/auth/third-party/auth0.mdx
+++ b/apps/docs/content/guides/auth/third-party/auth0.mdx
@@ -27,8 +27,8 @@ const auth0 = new Auth0Client({
   domain: '<AUTH0_DOMAIN>',
   clientId: '<AUTH0_CLIENT_ID>',
   authorizationParams: {
-    redirect_uri: '<MY_CALLBACK_URL>'
-  }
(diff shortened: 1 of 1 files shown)


kangmingtay (approved):

apps/docs/content/guides/auth/third-party/auth0.mdx
```diff
@@ -21,6 +21,15 @@ Auth0 can be used as a third-party authentication provider alongside Supabase Au
 
 ```typescript
 import { createClient } from '@supabase/supabase-js'
+import createAuth0Client from '@auth0/auth0-spa-js'
```
kangmingtay, line 24: i think the import is incorrect - shouldn't it be `Auth0Client`? (disclaimer: didn't really check the package if that method exists tho)

J0: ah good catch, from the repo description both methods exist but `createAuth0Client` is used as an async method - must have forgotten to change when I swapped out.

Thanks!

PR #29360, opened 2024-09-18
fix: update http to https and pathname to include **

J0:

YES

Receive the following error when loading the Auth guides locally.  The linked error seems to suggest a pattern mismatch and modifying the pattern seems to patch the error
image

- Also change two URLs from `http` to `https` for avoidance of doubt on pattern matching

Can't load Auth guides

## What is the new behavior?

Auth guides load


charislam (approved):
Thanks! 🙏🏼

fsansalvadore (approved):

PR #29385, opened 2024-09-19
chore: hide compute usage section for Free Plan

kevcodez:
There is no need to show the compute section for Free Plan customers, it likely only leads to confusion. The old endpoint is also pretty expensive to call.


alaister (approved):
LGTM!

- Tested on preview with both free and paid orgs

PR #29393, opened 2024-09-19
feat: add email restriction banner

hf:
Add email restriction banner based on resource warnings about Auth email restrictions. See: https://github.com/supabase/infrastructure/pull/19866


joshenlim (approved):
LGTM, tested locally by hardcoding resource-warnings-query to return `auth_restricted_email_sending` - just made some small tweaks

PR #29395, opened 2024-09-19
fix(seo): docs robots tag in previews

charislam:
Docs preview sites are getting indexed in Google. Last time we "fixed" this problem, we corrected the "X-Robots" header, but we forgot to change the robots meta tag in the HTML, so we didn't actually solve the problem. -_-


apps/docs/app/layout.tsx
```diff
@@ -19,8 +19,8 @@ const metadata: Metadata = {
   metadataBase: new URL('https://supabase.com'),
   icons: genFaviconData(BASE_PATH),
   robots: {
-    index: true,
-    follow: true,
+    index: IS_PRODUCTION ? true : false,
```
fsansalvadore, line 22:
```suggestion
    index: !!IS_PRODUCTION,
```

fsansalvadore (approved):
Approving with a suggestion 👍

apps/docs/app/layout.tsx
charislam: Ahh good one

PR #29423, opened 2024-09-20
chore: avoid querying usage endpoint unnecessarily

kevcodez:
For plans with usage-based billing enabled we don't need to check whether they exceeded their quotas. We can save us that network call.


alaister (approved):
LGTM!

- Opened on staging

Note: I don't have any projects that are exceeding usage currently so the testing here is limited

PR #29448, opened 2024-09-23
chore: max io2 disk size to 60 TB

delgado3d:

Bug fix

io2 max disk size limited to 16 TB

## What is the new behavior?

Increased to 60 TB


darora (approved):

joshenlim (approved):
LGTM RE dashboard changes - can verify on staging preview that the disk size limits are different for GP3 and IO2, and that form validation kicks in when changing disk type

PR #29462, opened 2024-09-24
Fix project roles not showing which project

joshenlim:
As per PR title
Was checking to show that info only if the org had access to project level roles - which checks for the org subscription, but if you’re a developer, you dont have access to the subscription info

Updated to just check if org already has _some_ project level roles instead

image


alaister (approved):
LGTM!

- Tested on preview

PR #29467, opened 2024-09-24
chore: explicit docs about reducing disk size

kevcodez:
It was a bit hard to spot for customers, so now making it more explicit on how to reduce disk size.

### DIFF
--- a/apps/docs/content/guides/platform/database-size.mdx
+++ b/apps/docs/content/guides/platform/database-size.mdx
@@ -95,12 +95,6 @@ Due to provider restrictions, disk expansions can occur only once every six hour
 
 </Admonition>
 
-<Admonition type="note">
-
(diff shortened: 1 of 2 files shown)


apps/docs/content/guides/platform/database-size.mdx
```diff
@@ -137,3 +131,9 @@ Once you have reclaimed space, you can run the following to disable [read-only](
 ```sql
 set default_transaction_read_only = 'off';
 ```
+
+### Reducing disk size
+
+Disks don't automatically downsize during normal operation. Once you have (reduced your database space usage)[/docs/guides/platform/database-size#database-space-management], they _will_ automatically "right-size" when a Supabase (or Postgres) update is applied. Their final size after the update is 1.2x the size of the database (min 8 GB). For example, if your database is 100GB in size, and you have a 200GB disk, the size after a Supabase update will be 120GB.
```
darora, line 137: Ideally this should link to the upgrades doc - which lists all the caveats involved, the downtime, etc

darora (approved):

PR #29468, opened 2024-09-24
Chore/user management v2 panel

joshenlim:
Clicking on a user will open the user's information up in a panel
- All actions previously behind a context menu is now available here (and context menu is removed as well)

image


apps/studio/components/interfaces/Auth/Users/UserOverview.tsx
```diff
@@ -0,0 +1,500 @@
...
+    'send_magic_link' | 'send_recovery' | 'send_otp'
+  >()
+  const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false)
+  const [isDeleteFactorsModalOpen, setIsDeleteFactorsModalOpen] = useState(false)
+
+  const { data } = useAuthConfigQuery({ projectRef })
+  const mailerOtpExpiry = data?.MAILER_OTP_EXP ?? 0
+  const minutes = Math.floor(mailerOtpExpiry / 60)
+  const seconds = Math.floor(mailerOtpExpiry % 60)
+  const formattedExpiry = `${mailerOtpExpiry > 60 ? `${minutes} minutes ${seconds > 0 ? 'and' : ''} ` : ''}${seconds > 0 ? `${seconds} seconds` : ''}`
```
alaister, line 77: Will minutes / seconds ever be `1`? If so we should handle the singular language here

apps/studio/components/interfaces/Auth/Users/Users.utils.ts
```diff
@@ -10,28 +10,35 @@ export const isAtBottom = ({ currentTarget }: UIEvent<HTMLDivElement>): boolean
...
+    const providers: string[] = user.raw_app_meta_data?.providers ?? []
 
     return {
       id: user.id,
       email: user.email,
       phone: user.phone,
       created_at: user.created_at,
       last_sign_in_at: user.last_sign_in_at,
-      provider: user.is_anonymous ? '-' : provider,
+
```
alaister, line 21: for some reason this space is upsetting me 😅
```suggestion
```

alaister (approved):
LGTM!

- Tested on preview with email and github based users

couple of tiny nits:

saltcod:
Not blocking & this affects a bunch of newer data grids, but up/down keyboard nav would be great through the rows, and hitting Esc should close the side panel 

screenshot-2024-09-24-at-09 32 03

saltcod:
this "waiting for notification" message 
used to be in further to the left — its now only visible if you scroll. 
Seems like an important piece of info to be always visible — wonder can we move that column over? (just imo, feel free to disagree)
screenshot-2024-09-24-at-09 34 30

saltcod:
can also tackle this separately, but curious where the pending state is for search 

it's currently down at the bottom, out of view on long tables
screenshot-2024-09-24-at-09 41 38

saltcod:
I feel like these two sidebar areas would benefit from a heading to make consistent with the others. Not sure the word "Actions" is right though. "Resend / reset" maybe?


screenshot-2024-09-24-at-09 52 26

joshenlim:
>this "waiting for notification" message
used to be in further to the left — its now only visible if you scroll.

hmm thinking about this - i reckon the best solution here is eventually to make the table columns sortable (and persisted via local storage), every team has their own preferred order of the columns i think, will look into this subsequently

> pending state is for search

Added the loading line! 🙏 

>I feel like these two sidebar areas would benefit from a heading to make consistent with the others. Not sure the word "Actions" is right though. "Resend / reset" maybe?

Tbh i was thinking of just shifting the data under "Invite information" to the first section - not sure if there's added benefit keeping it separated. Also thinking of adding separators - so we've got 3 sections (1) User info + provider (2) actions (3) danger, cause i can't think of a good title/description for (2) either (the UI itself feels quite clear what they are)

PR #29469, opened 2024-09-24
feat: add collection retention days

jordienr:
allows users to set retention days in collections.

CleanShot 2024-09-24 at 11 19 16@2x

### DIFF
--- a/apps/studio/components/interfaces/DataWarehouse/CreateWarehouseCollection.tsx
+++ b/apps/studio/components/interfaces/DataWarehouse/CreateWarehouseCollection.tsx
@@ -1,6 +1,6 @@
 import { zodResolver } from '@hookform/resolvers/zod'
 import { PermissionAction } from '@supabase/shared-types/out/constants'
-import { PlusIcon } from 'lucide-react'
+import { Info, PlusIcon } from 'lucide-react'
 import { useRouter } from 'next/router'
 import { useEffect, useState } from 'react'
 import { useForm } from 'react-hook-form'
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/DataWarehouse/CreateWarehouseCollection.tsx
```diff
@@ -103,6 +127,21 @@ export const CreateWarehouseCollectionModal = () => {
...
+                control={form.control}
+                name="retention_days"
+                render={({ field }) => (
+                  <FormItemLayout
+                    label="Retention days"
+                    layout="horizontal"
+                    description={retentionDescription()}
+                  >
+                    <FormControl_Shadcn_>
+                      <Input type="number" {...field} />
```
joshenlim, line 140: let's follow how we put units in our input fields? ref settings/auth
and we can also remove "days" from the label too - so just "Retention"

image

apps/studio/components/interfaces/DataWarehouse/CreateWarehouseCollection.tsx
```diff
@@ -103,6 +127,21 @@ export const CreateWarehouseCollectionModal = () => {
                   </FormItemLayout>
                 )}
               />
+              <FormField_Shadcn_
```
joshenlim, line 130: might need some gap-y-1 maybe between the 2 fields

image

apps/studio/components/interfaces/DataWarehouse/CreateWarehouseCollection.tsx
```diff
@@ -103,6 +127,21 @@ export const CreateWarehouseCollectionModal = () => {
...
+                control={form.control}
+                name="retention_days"
+                render={({ field }) => (
+                  <FormItemLayout
+                    label="Retention days"
+                    layout="horizontal"
```
joshenlim, line 136: thinking of using labelOptional for the upsell - something like this, but wondering wdyt as well though @saltcod 

image

joshenlim: hmm actually i realised - we don't allow reconfiguring the retention period after the collection is created right?

joshenlim: what about this though - wdyt?
- Changing the layout to vertical
- labelOptional to highlight that you can't change retention
- Upsell in an Admonition instead

image

joshenlim:
this is separate though but am thinking given that the retention can't be updated after the collection is created, should we add in some info about the retention here then? so users know what's the retention they set (and wont be confused if they start losing logs)
image

apps/studio/components/interfaces/DataWarehouse/CreateWarehouseCollection.tsx
jordienr: 🫡
CleanShot 2024-09-25 at 18 50 07@2x

apps/studio/components/interfaces/DataWarehouse/CreateWarehouseCollection.tsx
jordienr: @joshenlim checked with chase and we plan to allow updating retention for a collection. I think it's a good idea to upsell anyway if they need more than the free tier allows.

apps/studio/components/interfaces/DataWarehouse/CreateWarehouseCollection.tsx
```diff
@@ -56,9 +73,16 @@ export const CreateWarehouseCollectionModal = () => {
     createCollection({
       projectRef: ref,
       name: vals.name,
+      retention_days: vals.retention_days,
     })
   })
 
+  const retentionDescription = () => {
+    if (planLoading) return 'Max 90 days'
```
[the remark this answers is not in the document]
jordienr: https://github.com/user-attachments/assets/5b222be7-2d29-4ede-9679-4a5e64fb53c6

much better, thanks!

apps/studio/components/interfaces/DataWarehouse/CreateWarehouseCollection.tsx
saltcod: +++ to the upgrade modal there. I think customers will know what they want right there and then, let them upgrade in place if they want to 👍

saltcod:
I find something a bit off about having the two of these right together. Is the upgrade alert going at the bottom per discussion above? Or is it going out like this? If like this, let's think about how we can display these two pieces of info better

 
screenshot-2024-09-25-at-16 07 44

saltcod (approved):

PR #29471, opened 2024-09-24
Add check for saveNodePositions

joshenlim:
Basically don't save if there's no nodes present (which i think makes sense)


jordienr (approved):

ivasilov (approved):
The code looks ok (fixes all Sentry issues).

PR #29480, opened 2024-09-24
Chore/download types schemas

saltcod:
Download typescript types buttons do not include user-defined custom schemas

To reproduce: 

- On production, 
- create a custom schema and a new table: 

``` 
create SCHEMA pets;
 
CREATE TABLE
  pets.dogs (
    id BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
    NAME TEXT NOT NULL,
    breed TEXT NOT NULL,
    age INTEGER,
    created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
  );
  ```

- Download typescript types for your project https://share.cleanshot.com/qndRStRp
- see that `pets` schema is not included


charislam (approved):
Tested:

- Generated types include exposed custom schemas
-  Generated types do not include unexposed custom schemas

PR #29486, opened 2024-09-24
Chore/fix duplicate table

saltcod:
Fix database/tables duplicate table operation

Issue: 
- the side panel wasn't getting the current table and was going through the flow creating a new table instead of duplicating the current one 

screenshot-2024-09-24-at-15 37 21


joshenlim (approved):
LGTM! Verified on staging preview that clicking "Duplicate table" in the database/tables page opens the duplicate table panel. Also tested actually duplicating, and with RLS / realtime toggled on/off

PR #29502, opened 2024-09-25
Show all projects which role applies to, and link project from hover card

joshenlim:
Only applies to project level permissions which is available only on the Enterprise plan

image


apps/studio/components/interfaces/Organization/TeamSettings/MemberRow.tsx
```diff
@@ -61,6 +68,8 @@ export const MemberRow = ({ member }: MemberRowProps) => {
       return appliedProjects.length === 0
     }).length > 0
 
+  console.log({ orgProjects })
+
```
alaister, line 72: A cheeky one
```suggestion
```

alaister (approved):
LGTM!

- Tested on preview

One small one:

PR #29516, opened 2024-09-25
fix docs alert z index

fsansalvadore:

Fix alert visibility.

Screenshot 2024-09-25 at 17 04 20

## What is the new behavior?

Screenshot 2024-09-25 at 17 04 59


jgoux (approved):
When in doubt, level-up z-index! 😄

PR #29529, opened 2024-09-26
ref: Unify /v1 api requests, remove API_ADMIN_URL constant

kamilogorek:
- Replace deprecated `put` with `data/fetchers`
- Deprecated `put` from `lib/common/fetch/put.ts` is not used anymore and can be safely removed
- Remove all `API_ADMIN_URL` usage and use standardized `data/fetchers` urls to query `/v1` directly
- `API_ADMIN_URL` can be removed from Vercel env variables

All changed behavior verified locally.


alaister (approved):
LGTM!

- Tested on preview
  - Viewed general settings page for upgrade status query
  - Created an oauth app
  - Updated SSL enforcement settings

PR #29534, opened 2024-09-26
Redo column type dropdown

saltcod:
Replaces the side panel column type picker with new components

Maybe contentious change: moved the enums to the bottom instead of the top. imo, this input should always be consistent, with customization at the bottom.

 
screenshot-2024-09-26-at-10 25 12


ivasilov (approved):
The gap between the trigger and popover is too big, but otherwise looks ok. 
Screenshot 2024-09-26 at 16 27 00
Screenshot 2024-09-26 at 16 27 10

PR #29536, opened 2024-09-26
fix: vercel integration config form submit bug

alaister:

Bug fix

Two forms with the same id, submitting one submits both

## What is the new behavior?

Unique ids per form


kamilogorek (approved):

ivasilov (approved):
Tested on staging on a project with 2 Vercel connections.

PR #29547, opened 2024-09-26
Chore/move simple codeblock

saltcod:
Move SimpleCodeBlock to shared packages


joshenlim (approved):
LGTM - just tested briefly on the preview on the auth policy templates UI that things look okay, but presuming everything else is okay since the build went through properly.

Maybe just remove the sitemap changes?

PR #29553, opened 2024-09-27
fix: update to make explicit that there can be multiple secrets

J0:

YES

## What is the change

Better express that an Auth Hook can make use of multiple secrets

Will set to ready after catching up on the discussion on the [standard-webhooks repo](https://github.com/standard-webhooks/standard-webhooks) (the library we use) on the plans to handle multiple secrets on the client (if any)

On Auth side we have an existing [convention](https://github.com/supabase/auth/blob/master/internal/conf/configuration.go#L564)

### DIFF
--- a/apps/docs/content/guides/auth/auth-hooks.mdx
+++ b/apps/docs/content/guides/auth/auth-hooks.mdx
@@ -102,11 +102,15 @@ When a HTTP hook is created, the secret generated should be of the following for
 - `whsec_` signifies that the secret is symmetric
 - `<standard-base64-secret>` implies a Standard Base64 encoded secret which can contain the characters `+`, `/` and `=`
 
+
 The secret is used to verify the payload received in your hook. Create an entry in your `.env.local` file to store the secret for each hook that you have. For example:
 
 ```bash
-HOOK_SECRET=<standard-base64-secret>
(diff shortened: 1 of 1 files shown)


apps/docs/content/guides/auth/auth-hooks.mdx
```diff
@@ -102,20 +102,24 @@ When a HTTP hook is created, the secret generated should be of the following for
...
 
+
 The secret is used to verify the payload received in your hook. Create an entry in your `.env.local` file to store the secret for each hook that you have. For example:
 
 ```bash
-HOOK_SECRET=<standard-base64-secret>
+HOOK_SECRETS=<standard-base64-secret>
 ```
 
+There field is expressed in plural rather than singular as there are plans to allow for asymmetric signing and multiple hook secrets for ease of secret rotation. For instance: `<standard-base-64-secret>|<another-standard-base-64-secret>`.
```
J0, line 112:
```suggestion
There field is expressed in plural rather than singular as there are plans to allow for asymmetric signing and multiple hook secrets for ease of secret rotation. For instance, a possible convention could be: `<standard-base-64-secret>|<another-standard-base-64-secret>`. 
```

jgoux (approved):
LGTM!

PR #29586, opened 2024-09-30
Chore/hide disk io chart for 4xl and above

joshenlim:
Replaces Disk IO chart with this admonition for projects on 4XL and above

image


alaister (approved):
LGTM!

- Tested on preview with small and >= 4xl instances

darora (approved):

PR #29605, opened 2024-10-01
Small UI tweak to security issues dropdown

joshenlim:
### Before:
image

### After:
- Updated descriptions to be more concise
- Reduced width of dropdown (felt a bit odd that a dropdown was that large)
- Added header + aligned bottom button to the right for consistency
- Clicking on error, warn or info will direct to the security with the right tab
- Security advisor button just directs to the security advisor page without param
image


alaister (approved):
LGTM!

- Tested on preview

saltcod (approved):

PR #29607, opened 2024-10-01
chore: remove branch management flag

alaister:

chore - remove the `branchManagement` flag


joshenlim (approved):
LGTM code wise - tested authorizing GH locally but ran into issues adding/removing GH integrations to be project due to staging/local set up issues, but alaister helped to verify those are working 🙏

PR #29610, opened 2024-10-01
chore: Prep work for `pnpm`

ivasilov:
This PR fixes some issues in the monorepo to let us use pnpm:
- each package (or app) needs to be have all referenced-in-code dependencies specified in its `package.json`. No new dependencies have been introduced.
- remove circular dependencies (`ui-patterns` depends on `ui` which reexports `admonition` from `ui-patterns`)
- add some reexports to ui components to be used in the apps instead of the apps adding more dependencies just for util functions and types.
- for packages and apps which use vitest, add imports for all vitest functions. This used to be work because `jest` is providing ambient types for the functions, but it's better if they're declared. 
- migrated the `ui` package to use `vitest` instead of `jest`.


alaister (approved):
LGTM!

- Opened all previews and clicked around

fsansalvadore (approved):

PR #29614, opened 2024-10-01
docs(python): async realtime

grdsdev:

Update all python ref examples to use the async variant, as that is the only supported way.

### DIFF
--- a/apps/docs/spec/supabase_py_v2.yml
+++ b/apps/docs/spec/supabase_py_v2.yml
@@ -6425,72 +6425,81 @@ functions:
         isSpotlight: true
         code: |
           ```python
-          channel = supabase.channel("room1")
+          channel = await supabase.channel("room1")
 
           def on_subscribe(status, err):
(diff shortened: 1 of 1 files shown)


thearperson:
Thanks for adding this, super helpful. Curious: is there any standalone Python sample code for this?

apps/docs/spec/supabase_py_v2.yml
```diff
@@ -6425,72 +6425,81 @@ functions:
         isSpotlight: true
         code: |
           ```python
-          channel = supabase.channel("room1")
+          channel = await supabase.channel("room1")
```
thearperson, line 6428: From this snippet alone, it's not very clear where the `supabase` object comes from.

After a bit digging, it seems to be an `AClient` object?

Hallidayo:
Hi  @grdsdev - We’re doing a bit of housekeeping and plan to close out PRs that have been inactive for more than 3 months.

PR #29629, opened 2024-10-01
Clarify how to use the service role key with the rest API

mmkal:
I struggled with this for a while. v0 was able to figure this out from the docs, but I wasn't, so I thought an explicit example which makes it easy to go from values you see in the dashboard to what-do-i-need-to-do would help.

YES

Docs update

People like me getting confused as to how to use the service role key when working with the rest API directly (not using a client library)

## What is the new behavior?

A `curl` example showing how to use it

## Additional context

There is an example saying that you need to pass both the `apikey` and `authorization` header when auth'ing a *user*, but it doesn't for a service role.

I reiterated that people should only use the service role key from a secure environment like a server because I don't want to be responsible for people not doing that. But I can remove if it feels like too much.

### DIFF
--- a/apps/docs/content/guides/auth/jwts.mdx
+++ b/apps/docs/content/guides/auth/jwts.mdx
@@ -162,6 +162,14 @@ You'll notice that this token is quite a bit longer, since it contains informati
 }
 ```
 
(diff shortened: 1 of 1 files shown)


charislam (approved):
Thanks @mmkal! Absolutely agree about reiterating the warning to do this in a secure environment

PR #29636, opened 2024-10-02
Fix filtering on SAML provider

joshenlim:
As per PR title - this is for the user management UI


alaister (approved):
LGTM!

- Tested on preview

PR #29637, opened 2024-10-02
Add number of database connections charts in database reports

joshenlim:
Dependent on BE API PR here:
https://github.com/supabase/infrastructure/pull/20071

And their deployment here:
https://github.com/supabase/infrastructure/pull/20090

image


apps/studio/data/projects/project-upgrade-mutation.ts
```diff
@@ -12,7 +12,8 @@ export type ProjectUpgradeVariables = {
 export async function upgradeProject({ ref, target_version }: ProjectUpgradeVariables) {
   const { data, error } = await post('/v1/projects/{ref}/upgrade', {
     params: { path: { ref } },
-    body: { target_version },
+    // @ts-ignore release_channel param is optional on API end, but not typed properly from generated types
```
alaister, line 15:
```suggestion
    // @ts-expect-error release_channel param is optional on API end, but not typed properly from generated types
```

alaister (approved):
LGTM!

- Tested on preview

PR #29642, opened 2024-10-02
Chore/user management v2 part 3

joshenlim:
Introduces the logs panel in the user management UI, starting with relevant auth logs
image


apps/studio/components/interfaces/Auth/Users/Users.utils.tsx
```diff
@@ -132,8 +133,15 @@ export function getDisplayName(user: User, fallback = '-'): string {
     first_name,
   } = user.raw_user_meta_data ?? {}
 
-  const last = familyName || family_name || surname || lastName || last_name
-  const first = givenName || given_name || firstName || first_name
+  const last =
+    (custom_claims as any)?.last_name ||
```
joshenlim, line 137: i realised this custom_claims property exists for SAML users

apps/studio/hooks/analytics/useLogsPreview.tsx
```diff
@@ -42,11 +43,17 @@ interface LogsPreviewHook {
   setFilters: (filters: Filters | ((previous: Filters) => Filters)) => void
   setParams: Dispatch<SetStateAction<LogsEndpointParams>>
 }
-function useLogsPreview(
-  projectRef: string,
-  table: LogsTableName,
+function useLogsPreview({
+  projectRef,
```
joshenlim, line 47: just making the arguments here into an object so its easier / more flexible to call this hook

apps/studio/components/interfaces/Auth/Users/Users.utils.tsx
alaister: Good catch. It may not be called `last_name`/`first_name` every time though, so maybe if we can go through all the options like we normally do on `custom_claims` first, then callback to the normal variables?

apps/studio/components/interfaces/Auth/Users/Users.utils.tsx
```diff
@@ -172,7 +180,7 @@ export function getAvatarUrl(user: User): string | undefined {
     profileImageURL ||
     profile_image_url) as string | undefined
 
-  return SUPPORTED_CSP_AVATAR_URLS.some((x) => url?.startsWith(x)) ? url : undefined
+  return SUPPORTED_CSP_AVATAR_URLS.some((x) => (url ?? '')?.startsWith(x)) ? url : undefined
```
alaister, line 183: Am I going crazy?
```suggestion
  return SUPPORTED_CSP_AVATAR_URLS.some((x) => (url ?? '').startsWith(x)) ? url : undefined
```

joshenlim: oh yeap! wouldnt need the ternary then

joshenlim:
> Good catch. It may not be called last_name/first_name every time though, so maybe if we can go through all the options like we normally do on custom_claims first, then callback to the normal variables?

you mean like here?
https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Auth/Users/Users.utils.tsx#L133

separately, im also wondering how sustainable this one is - esp if we do end up allowing users to update the raw_user_meta_data
this would be a free form input and we wouldn't be able to reliably get the display name anymore

alaister:
> you mean like here? https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Auth/Users/Users.utils.tsx#L133

Yep, exactly like that!

> separately, im also wondering how sustainable this one is - esp if we do end up allowing users to update the raw_user_meta_data this would be a free form input and we wouldn't be able to reliably get the display name anymore

This is more of a nice to have rather than something that should always work... so we can do our best here but if people are going to have crazy solutions for naming their users I don't think we'll ever be able to account for everything!

joshenlim:
Closing in favor of https://github.com/supabase/supabase/pull/29656

PR #29644, opened 2024-10-02
Fix scroll issue in settings infra reactflow

joshenlim:
Currently if you try to mouse wheel scroll in settings/infrastructure while your cursor is in the react flow canvas, nothing happens. This fixes that by ensuring the page scroll works, since we've disabled zoom on scroll for our react flow anyways


alaister (approved):
LGTM!

- Tested on preview

PR #29650, opened 2024-10-02
Chore: Replaced any with interface CSSProperties 

### LINKED ISSUE #29576: replace type any with interface CSSProperties
# Chore
I've noticed that in some places where css is used as props, there is a usage of type "any". We can clearly replace this with usage of interface "CSSProperties" from "react"

## Describe the chore

Replace all instances of type "any" with interface "CSSProperties" where css is used as props.

## Additional context

Take a look at this before, after code example:
image

0xkar4n:
This PR is to update the any with interface CSSProperties where CSS is accepted as a prop which is mentioned in issue #29576
Fixes #29576 

YES

Chore

image

## What is the new behavior?
image


JoyShaheb:
@0xkar4n The purpose is to search, find and replace all instances of the same issue. Not just the file I pointed out, there's more.

fsansalvadore:
@JoyShaheb actually as far as the specific issue #29576, this is actually the correct fix.
Only the `style` type needs to be of CSSProperties type, not all "any".

Thanks anyway for the joining the discussion!

fsansalvadore (approved):
Thanks @0xkar4n 👍

PR #29653, opened 2024-10-02
search and replaced instances of props being type any, to type CSSPropterties

alinagarib:

YES/NO

YES

replaces every instance where type "any" is used as a prop with CSSProperties

repeated usage of props of type any

https://github.com/supabase/supabase/issues/29576

## What is the new behavior?

same, with instances of props: any replaced with props: CSSProperties

## Additional context


JoyShaheb:
@alinagarib, seems like you misunderstood the usage of `CSSPropterties`. The original purpose was to replace the `any` which were used to insert CSS styles as props. Do you need more explanations or some examples?

PR #29660, opened 2024-10-03
Chore/improvements to datetime editing in table editor

joshenlim:
### Small improvements to datetime editing
- In the grid: Date time picker automatically opens when you click to edit a cell
- In the side panel: Added an action for time input types, to either set to NULL or set to now
  - This should alleviate a problem on Safari as the default browser date picker UI does not support setting the value to NULL (See bottom for browser date picker references)
  - On Safari you still can't set a time type field to NULL within the grid editor still though, I haven't been able to figure out a good solution for this just yet as its a little more complicated with the grid cell's focusing state and all
image

Reference for default browser pickers:
- Chrome: 
image
- Safari: 
image
- Firefox:
image

### DIFF
--- a/apps/studio/components/grid/components/editor/DateEditor.tsx
+++ b/apps/studio/components/grid/components/editor/DateEditor.tsx
@@ -1,10 +1,5 @@
-import * as React from 'react'
+import { ChangeEvent, useEffect, useRef } from 'react'
 import type { RenderEditCellProps } from 'react-data-grid'
-
-function autoFocusAndSelect(input: HTMLInputElement | null) {
-  input?.focus()
-  input?.select()
-}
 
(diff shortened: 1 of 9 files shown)


alaister (approved):
For columns with a not-null constraint, we don't need to show the set null option right?
Screenshot 2024-10-03 at 16 20 47

joshenlim:
> For columns with a not-null constraint, we don't need to show the set null option right?
Oh right yeap good point - also found a fix with the set to now for date type (not timestamptz), will fix!

joshenlim:
Okay just pushed up a fix to verify that "Set to now" works on date, time, timetz, timestamp, and timestamptz

alaister (approved):
LGTM!

- Tested on preview

PR #29661, opened 2024-10-03
Chore/add validation for supabase prefix in edge functions secrets management

joshenlim:
- Add validation for SUPABASE_ prefix when creating new secret
image

- Add validation for SUPABASE_ prefix when deleting secrets
image

### DIFF
--- a/apps/studio/components/interfaces/Functions/EdgeFunctionSecrets/AddNewSecretModal.tsx
+++ b/apps/studio/components/interfaces/Functions/EdgeFunctionSecrets/AddNewSecretModal.tsx
@@ -1,10 +1,15 @@
-import { useParams } from 'common'
-import { useRef, useState } from 'react'
+import { zodResolver } from '@hookform/resolvers/zod'
+import { useEffect, useRef, useState } from 'react'
+import { SubmitHandler, useForm } from 'react-hook-form'
 import { toast } from 'sonner'
-import { Button, Form, Input, Modal } from 'ui'
+import z from 'zod'
 
+import { useParams } from 'common'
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Functions/EdgeFunctionSecrets/AddNewSecretModal.tsx
```diff
@@ -16,27 +21,44 @@ const AddNewSecretModal = ({ visible, onClose }: AddNewSecretModalProps) => {
   const submitRef = useRef<HTMLButtonElement>(null)
   const [showSecretValue, setShowSecretValue] = useState(false)
 
+  const FormSchema = z.object({
+    name: z
+      .string()
+      .min(1, 'Please provide a name for your secret')
+      .refine((value) => !value.match(/^(SUPABASE_).*/), {
```
alaister, line 28: Are we already validating that these secrets are uppercase? Am I allowed to create secrets with `supabase_` (lowercase)?

joshenlim: these are the checks that API does on their end, so gonna follow them here (so yup lowercase should be fine)
https://github.com/supabase/infrastructure/blob/develop/api/src/routes/v1/projects/ref/secrets/secrets.controller.ts#L106
https://github.com/supabase/infrastructure/blob/develop/api/src/routes/v1/projects/ref/secrets/secrets.dto.ts#L10

alaister (approved):
LGTM!

- Tested on preview

PR #29665, opened 2024-10-03
add edge logs page

jordienr:


joshenlim (approved):
able to verify on staging preview that the page renders with no errors, endpoint is not throwing errors, but none of my staging projects have logs for edge functions though

PR #29669, opened 2024-10-03
Project Status improvements

jordienr:
- add logs links
- set status as coming up when project is less than 10 min old

https://github.com/user-attachments/assets/d0deae35-505b-4409-9059-69035b2d923f


MildTomato (approved):

awalias (changes requested):
can we solve for restoring as well

awalias (approved):
approving just so I'm not blocking but let's def solve for restoring as well

apps/studio/components/interfaces/Home/ServiceStatus.tsx
```diff
@@ -113,13 +132,52 @@ const ServiceStatus = () => {
...
+    if (isProjectNew) return 'Coming up...'
+    if (isSuccess) return 'No issues'
+    return 'Unable to connect'
+  }
+
+  const StatusIcon = ({ isLoading, isSuccess }: { isLoading: boolean; isSuccess: boolean }) => {
+    if (isLoading || isProjectNew) return <Loader2 size={14} className="animate-spin" />
+    if (isSuccess) return <CheckCircle2 className="text-brand" size={18} strokeWidth={1.5} />
+    return <AlertTriangle className="text-warning" size={18} strokeWidth={1.5} />
+  }
```
alaister, line 150: Can these be defined outside of the render method? I find there are often random bugs when defining other components inside components!

apps/studio/components/interfaces/Home/ServiceStatus.tsx
```diff
@@ -105,6 +123,7 @@ const ServiceStatus = () => {
             docsUrl: 'https://supabase.com/docs/guides/functions/troubleshooting',
             isLoading,
             isSuccess: edgeFunctionsStatus?.healthy,
+            logsUrl: '/logs/edge-functions-logs',
```
alaister, line 126: Does this route actually exist?

alaister (approved):
LGTM!

- Tested on preview

A couple of small things:

apps/studio/components/interfaces/Home/ServiceStatus.tsx
joshenlim: ahh yup! it's in here - just merged it
https://github.com/supabase/supabase/pull/29665

apps/studio/components/interfaces/Home/ServiceStatus.tsx
joshenlim: oh yes gotcha! overlooked this one 🙏

PR #29671, opened 2024-10-03
new Integrations page

saltcod:
wip

screenshot-2024-10-09-at-09 17 20


ivasilov (commented):
We should probably move integrations above API docs (maybe even in the second group along DB, Auth because the third group is a group with read-only features).

The webhooks should be the first choice in the sidebar because they're dynamic (may not appear for older projects).

ivasilov (approved):

PR #29678, opened 2024-10-04
feat: workflow run logs

alaister:

feature

https://github.com/user-attachments/assets/305337f1-0591-4820-a0c4-c4f1724c0a79


joshenlim:
Some issues found:
- Probably needs to handle "View logs" for old branches which don't have logs? (Update: might be another issue that's not about old branches)
image

- Nit: small inconsistent padding top bottom here
image

Thinking instead of opening the button, we make the button disabled (using ButtonTooltip) with the text "No logs available for this branch". Cause the modal here feels like a bit of an overkill
image

There doesn't seem to be a way to select another log after choosing one
image

- Unrelated: This can probably be an admonition now, like how it works in the text confirm modal (when you delete a branch)
image

joshenlim (approved):
I can't really test the correctness of the UI as I'm running into workflow logs issue on my staging project but since this is feature flagged - okay to let this go through 👍

PR #29682, opened 2024-10-04
Prevent deploying new replicas if org has outstanding invoices

joshenlim:
Creating new project:
image

Deploying new RR:
image

### DIFF
--- a/apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureConfiguration/DeployNewReplicaPanel.tsx
+++ b/apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureConfiguration/DeployNewReplicaPanel.tsx
@@ -5,12 +5,16 @@ import { toast } from 'sonner'
 
 import { useParams } from 'common'
 import {
-  DISK_LIMITS,
   DISK_PRICING,
   DiskType,
 } from 'components/interfaces/DiskManagement/DiskManagement.constants'
+import {
+  calculateIOPSPrice,
+  calculateThroughputPrice,
+} from 'components/interfaces/DiskManagement/DiskManagement.utils'
 import { useDiskAttributesQuery } from 'data/config/disk-attributes-query'
 import { useEnablePhysicalBackupsMutation } from 'data/database/enable-physical-backups-mutation'
+import { useOverdueInvoicesQuery } from 'data/invoices/invoices-overdue-query'
 import { useProjectDetailQuery } from 'data/projects/project-detail-query'
 import { Region, useReadReplicaSetUpMutation } from 'data/read-replicas/replica-setup-mutation'
 import {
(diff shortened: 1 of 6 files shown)


alaister (approved):
LGTM!

- Tested locally

kevcodez (changes requested):
The backend only restricts this for orgs not on team and enterprise, front end should match that accordingly

Ideally front end should stop implementing all this billing business logic and we should rather have a validation/prereq endpoint that outputs whether the action is possible

joshenlim:
> Ideally front end should stop implementing all this billing business logic

Yeap defo agreed on this - there's multiple places where these sort of validations are required

joshenlim:
> The backend only restricts this for orgs not on team and enterprise, front end should match that accordingly

Just pushed a fix for this one! 🙏 @kevcodez

apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureConfiguration/DeployNewReplicaPanel.tsx
```diff
@@ -78,9 +78,11 @@ const DeployNewReplicaPanel = ({
   const { data: subscription } = useOrgSubscriptionQuery({ orgSlug: org?.slug })
   const { data: diskConfiguration } = useDiskAttributesQuery({ projectRef })
 
-  const hasOverdueInvoices =
-    (allOverdueInvoices ?? []).filter((x) => x.organization_id === project?.organization_id)
-      .length > 0
+  const overdueInvoices = (allOverdueInvoices ?? []).filter(
```
kevcodez, line 81: I think it would be neater if we simply disabled the entire overdueInvoices call for these customers (enabled flag on the query), so there is no unnecessary network call

apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureConfiguration/DeployNewReplicaPanel.tsx
```diff
@@ -78,9 +78,11 @@ const DeployNewReplicaPanel = ({
...
+    (x) => x.organization_id === project?.organization_id
+  )
+  const ignoreOverdueInvoices = ['team', 'enterprise'].includes(subscription?.plan.id ?? '')
```
kevcodez, line 84: I think it would be neater if we simply disabled the entire overdueInvoices call for these customers (enabled flag on the query), so there is no unnecessary network call

joshenlim: gotcha just pushed a change! sorry only just saw this comment 🙏

apps/studio/pages/new/[slug].tsx
```diff
@@ -203,16 +204,22 @@ const Wizard: NextPageWithLayout = () => {
   const { data: organizations, isSuccess: isOrganizationsSuccess } = useOrganizationsQuery()
   const currentOrg = organizations?.find((o: any) => o.slug === slug)
 
+  const { data: orgSubscription } = useOrgSubscriptionQuery({ orgSlug: slug })
+
+  const { data: allOverdueInvoices } = useOverdueInvoicesQuery({
+    enabled: !['team', 'enterprise'].includes(orgSubscription?.plan.id ?? ''),
```
Comment (kevcodez) [inline apps/studio/pages/new/[slug].tsx:210]: This is going to fire immediately as orgSubscription will be undefined while it's still loading

```suggestion
    enabled: orgSubscription && !['team', 'enterprise'].includes(orgSubscription.plan.id),
```

Comment (joshenlim) [reply apps/studio/pages/new/[slug].tsx:210]: oh right gotcha on this! will update for DeployNewReplicaPanel too

kevcodez (approved):
Tested on staging

PR #29685, opened 2024-10-04
Fix auth users count when searching

joshenlim:


alaister (approved):
LGTM!

- Tested on preview

PR #29688, opened 2024-10-04
Fix wrong style width value in solid tutorial

vincentfretin:

YES

Fix issue in code in the solid tutorial.

The previous code gave a ts error:
`Type 'number' is not assignable to type 'Width<0 | (string & {})> | undefined'.ts(2322)`

## What is the new behavior?

The error goes away.

## Additional context

This is using the same logic than the div below. An alternative is just to remove the style on this div, all children have the width specified.


charislam (approved):
Thanks @vincentfretin!

PR #29692, opened 2024-10-05
chore: use amount_due for invoice total

kevcodez:
To not confuse customers thinking they paid double - we will display the amount_due (amount after credits) for now.

We'll do a more sophisticated iteration later, this is mainly to avoid customers from thinking they paid multiple times for the time of the Orb migration.


alaister (approved):
LGTM!

- Tested on preview

PR #29720, opened 2024-10-07
chore: kaizen changes for disk size billing

kevcodez:
- Improved disk size docs on "How billing works" docs
- Fixed a link on the disk management docs
- Moved from tooltip on hover over the pie chart to using a hover card that is on the whole element, making it easier to spot
- Removed the static PieChart icon for compute hours or when usage is filtered by project
- Improved tooltip for disk size metric

Screenshot 2024-10-07 at 20 21 33


alaister (approved):
LGTM!

- Tested on preview

note I couldn't fully test the hovercards, as none of my staging orgs seem to have any usage at all...

PR #29723, opened 2024-10-07
fix: Add special handling for array of enums when fetching table rows

ivasilov:
This PR fixes an issue which occurred when exporting a table to CSV and the table has an enum array field.


joshenlim (approved):
LGTM! Verified on staging preview - hitting "Select all rows" then "Export to CSV" for a table with a enum array column that the output is ["enum1", "enum2"], instead of {enum1, enum2}

separate note though - we probably need to think about how to represent arrays here (exporting data, side panel editor etc). We currently treat arrays in the dashboard as JS arrays, but postgres arrays are different with `{}` instead of `[]`, just wondering if this might be tripping some users up

PR #29724, opened 2024-10-07
fix: allow editing of rate limit when SendEmailHook is enabled

J0:

YES

For developers who only use the Send Email Hook and not the default or custom smtp


J0:
The handling on infra side needs to change as well

PR #29734, opened 2024-10-08
chore: improve table grid loading skeleton

alaister:

chore

Currently we have a spinner, followed by a skeleton.

## What is the new behavior?

Skeleton only.

## Additional context

https://github.com/user-attachments/assets/aef7b3e2-ba26-4b2b-ba26-43d719e8c44e


joshenlim (approved):
nice one! LGTM - tested on staging preview, loading states are much more fluid, less shifting of layouts

PR #29745, opened 2024-10-08
Refactor telemetry event into RQ hook, and update to support PH

joshenlim:
No visual change for this one

Dashboard: (Verified locally)
- Refactored calling sendEvent and sendIdentify from Telemetry to use an RQ mutation hook instead
- Refactored posting to /telemetry/page to use an RQ mutation hook instead
- Conditionally format the request body and version header based on a feature flag
- Deprecate lib/telemetry


joshenlim:
closing in favor of https://github.com/supabase/supabase/compare/chore/refactor-telemetry-calls-for-dashboard-to-support-PH?expand=1

PR #29747, opened 2024-10-08
Chore/refactor telemetry calls for dashboard to support PH

joshenlim:
No visual change for this one - update Telemetry calls to go to PH instead of GA depending on feature flag

Dashboard: (Verified locally)
- Refactored calling sendEvent and sendIdentify from Telemetry to use an RQ mutation hook instead
- Refactored posting to /telemetry/page to use an RQ mutation hook instead
- Conditionally format the request body and version header based on a feature flag
- Deprecate lib/telemetry

Docs
- Just adding ts-ignore here to prevent the typescript check from failing for now 

NOTE: Docs and WWW do not have feature flags set up, and I'm thinking it doesn't make sense for them to either since feature flags require users to be signed in to work, and these 2 apps do not enforce users to be signed in. We'll still need to make the necessary changes for them, but instead of feature flagging, we probably need to keep them in a PR until they're ready to go.


apps/studio/data/telemetry/send-event-mutation.ts
```diff
@@ -0,0 +1,124 @@
...
+        : ({ ...vars, ...payload } as SendEventGA)
+      return sendEvent(type, body)
+    },
+    {
+      async onSuccess(data, variables, context) {
+        await onSuccess?.(data, variables, context)
+      },
+      async onError(data, variables, context) {
+        if (onError === undefined) {
+          toast.error(`Failed to send Telemetry event: ${data.message}`)
```
saltcod, line 116: is this user facing?

apps/studio/data/telemetry/send-identify-mutation.ts
```diff
@@ -0,0 +1,80 @@
...
+        : ({ user: userInfo, ...payload } as SendIdentifyGA)
+      return sendIdentify(type, body)
+    },
+    {
+      async onSuccess(data, variables, context) {
+        await onSuccess?.(data, variables, context)
+      },
+      async onError(data, variables, context) {
+        if (onError === undefined) {
+          toast.error(`Failed to send Telemetry identify: ${data.message}`)
```
saltcod, line 72: same

saltcod (approved):

alaister (approved):
LGTM!

- Tested on preview

PR #29749, opened 2024-10-08
Update telemetry calls for docs to support PH

joshenlim:
***Note: This should only be merged after @pamelachia gives us the cue***

Ensure that https://github.com/supabase/supabase/pull/29747 is merged, and then we merge master into this merge so that this branch gets the latest API types changes (if we do update the API types now, then dashboard typescript checks are gonna fail)

- Updates calls to telemetry (/page and /event) to support the new version 2 endpoint for both docs and www
- Update existing TELEMETRY_CONSENT property in LOCAL_STORAGE_KEYS from common to reset user’s preferences for opting in
- Update PrivacySettings under `ui-patterns` to trigger a call to telemetry reset when opting out
- Update telemetry stuffs from `common` to use new endpoint and payload for telemetry page


charislam (approved):

pamelachia:
Not sure what other PRs are involved but posthog is a go and we can push out all the necessary PRs and changes live.

4L3k51 (approved):

PR #29756, opened 2024-10-08
update logs empty state

jordienr:
- fix height issue with empty state
- update ui copy in PostgREST logs
- componentize empty state to make it easy to update the rest

CleanShot 2024-10-08 at 21 11 05@2x

### DIFF
--- a/apps/studio/pages/project/[ref]/logs/postgrest-logs.tsx
+++ b/apps/studio/pages/project/[ref]/logs/postgrest-logs.tsx
@@ -19,7 +19,7 @@ export const LogPage: NextPageWithLayout = () => {
       EmptyState={
         <LogsTableEmptyState
           title="No results found"
-          description="By default only errors are logged into PostgREST logs. If you want to see HTTP requests, check the API Gateway logs."
+          description="Only errors are captured into PostgREST logs by default. Check the API Gateway logs for HTTP requests."
         />
       }
     />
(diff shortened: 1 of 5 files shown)


apps/studio/pages/project/[ref]/logs/postgrest-logs.tsx
```diff
@@ -15,6 +16,12 @@ export const LogPage: NextPageWithLayout = () => {
       queryType="postgrest"
       projectRef={ref as string}
       tableName={LogsTableName.POSTGREST}
+      EmptyState={
+        <LogsTableEmptyState
+          title="No results found"
+          description="By default only errors are logged into PostgREST logs. If you want to see HTTP requests, check the API Gateway logs."
```
Comment (saltcod) [inline apps/studio/pages/project/[ref]/logs/postgrest-logs.tsx:22]: "...are captured in PostgREST logs..." ?

saltcod (approved):

apps/studio/pages/project/[ref]/logs/postgrest-logs.tsx
Comment (joshenlim) [reply apps/studio/pages/project/[ref]/logs/postgrest-logs.tsx:22]: ^ i think thats better yup!

joshenlim (approved):
Verified on staging preview that the empty state height takes up the full height now instead of not taking up the whole height

One really minor nit though sorry 😂 it's now slightlyyyyy scrollable hahahaha
Edit: Ahh okay found the issue with the max height provided to LogsTable, if its possible we should avoid these arbitrary numbers as much as possible but we can look into that next time 🙏 

https://github.com/user-attachments/assets/22563783-db51-406f-a5a3-81a821df6c3e

PR #29761, opened 2024-10-08
Include "to authenticated" in example rls policies

enzzzooo:
Update custom-claims-and-role-based-access-control-rbac.mdx and added "to authenticated" in the rls policy which only allows logged in users to query instead of non logged in users as per "Specify roles in your policies" found here https://supabase.com/docs/guides/database/postgres/row-level-security#specify-roles-in-your-policies

YES

Include "to authenticated" in example rls policies

No roles and uses extra resources for no reason


charislam (approved):
Thanks @enzzzooo!

PR #29762, opened 2024-10-09
fix: new common hook: useIsomorphicLayoutEffect

alaister:

Bug fix

image (2)

## What is the new behavior?

A peaceful console


kevcodez (changes requested):
We shouldn't introduce dependencies for saving a single line of code

kevcodez (approved):

PR #29773, opened 2024-10-09
fix: Fix Studio crashes when using Google Translate

ivasilov:
Wrap all selects which use SelectValue with spans to avoid crashes when using Google Translate on its pages.


alaister (approved):
LGTM!

- Tested on preview by checking a bunch of the changed dropdowns

PR #29784, opened 2024-10-09
Prevent deploying replicas if on pro plan with spend cap enabled

joshenlim:
image


alaister (approved):
LGTM!

- Tested on preview with a spend capped project

PR #29785, opened 2024-10-09
Add sendReset call when user opts out of telemetry

joshenlim:


alaister (approved):
LGTM!

- Tested on preview

PR #29791, opened 2024-10-09
update collection form

jordienr:
- create component for form
- use new component in create and update collection dialogs
- allows users to update retention days


alaister (approved):
LGTM!

- Tested on preview

PR #29793, opened 2024-10-09
Update send identify RQ to send user gotrue id instead of user id

joshenlim:


saltcod:
Confirmed working on preview & that IdentifyUserBody has gotrue_id property

saltcod (approved):

PR #29794, opened 2024-10-09
fix: allow submitting support form without orgs

kevcodez:
Currently customers are forced to select an organization to submit the form. However, when the customer has no orgs, they are in a deadlock.

This PR allows customer to submit the form without an org selected, BUT only if they do not have any orgs. That is to prevent people from forgetting to select their org while creating a ticket related to an org. This is the least invasive change while still allowing customers to create a ticket.

Shadcn SelectItem doesn't allow null/undefined/empty string so we gotta do some nonsense to represent the empty value state here.


apps/studio/components/interfaces/Support/SupportFormV2.tsx
```diff
@@ -324,12 +337,15 @@ export const SupportFormV2 = ({ setSentCategory, setSelectedProject }: SupportFo
...
-                        {isLoadingSubscription && <Loader2 size={14} className="animate-spin" />}
+                        {isLoadingSubscription && organizationSlug !== 'no-org' && (
+                          <Loader2 size={14} className="animate-spin" />
+                        )}
                         {isSuccessSubscription && (
                           <Badge variant="outline" className="capitalize">
                             {subscriptionPlanId}
                           </Badge>
                         )}
+                        {organizationSlug === 'no-org' && <span>No specific org</span>}
```
joshenlim, line 348: let's spell this out fully, same for below
```suggestion
                        {organizationSlug === 'no-org' && <span>No specific organization</span>}
```

joshenlim:
seems like the loader stays there
image

apps/studio/components/interfaces/Support/SupportFormV2.tsx
```diff
@@ -338,6 +354,9 @@ export const SupportFormV2 = ({ setSentCategory, setSelectedProject }: SupportFo
                       {organizations?.map((org) => (
                         <SelectItem_Shadcn_ value={org.slug}>{org.name}</SelectItem_Shadcn_>
                       ))}
+                      {organizations?.length === 0 && (
```
joshenlim, line 357: just to check - do we want this option only available if the user has 0 organizations? or make it available for everyone?

i'm okay with either tbh - the current implementation will not have any change for most users i guess which is also good

kevcodez: > do we want this option only available if the user has 0 organizations? 

this, see PR description for reasining

joshenlim: ahhh gotcha! sorry skipped over that 🙏  gotcha on this yup thats all fine

joshenlim:
Closing in favor of https://github.com/supabase/supabase/pull/29809

PR #29802, opened 2024-10-09
store selected logs to url

jordienr:
stores selected log to url
fixes issues with nuqs in tests (thx ivan)
fixes layout scroll issues without those ugly magic numbers i had (this was the hardest part)

move state up to reduce debt in LogTable and make it easier to reuse between warehouse and logs
improve empty state in sidebar


apps/studio/components/layouts/LogsLayout/LogsSidebarMenuV2.tsx
```diff
@@ -276,14 +277,16 @@ export function LogsSidebarMenuV2() {
...
-            </Alert_Shadcn_>
-          </div>
+          <InnerSideBarEmptyPanel
+            className="mx-4"
+            title="No queries created yet"
+            description="Create and save your queries to use them in the explorer"
+            actions={
+              <Link href={`/project/${ref}/logs/explorer`}>
+                <Button type="default">Create query</Button>
+              </Link>
```
joshenlim, line 287: usually we write it this way for consistency:

```
<Button asChild>
  <Link ...>
    Create query
  </Link>
</Button>
```

joshenlim (commented):
LGTM and tested on staging preview that URL state for logs works
- went into Postgres logs
- clicked on a link
- copied and opened the URL in another tab
- selected log shows up

Possible improvement:
- selected log should be scrolled to and highlighted in the data grid when opening the URL in another tab, currently does not
image
image

Just double checking if the panel being always open now is intended (on prod currently it only shows up when a log is selected)

joshenlim (approved):
side panel looks good now!

PR #29806, opened 2024-10-09
Chore/link to storage settings

saltcod:
PR Does a few things: 

- Add a link to storage settings if the file is larger than the global upload max size

screenshot-2024-10-09-at-21 22 10

- Update the 413 (too large) error to be more reader friendly (before/after)
 
screenshot-2024-10-10-at-10 49 44

- add 413 to doNotRetryStatuses. Previously, files that were too  large would retry 7 times before finally failing, when they could just fail immediately


ivasilov (changes requested):
Screenshot 2024-10-10 at 11 33 31
There's also a bucket limit which is not clear how it interacts with the global limit. What does "takes precedence" mean? The highest or the lowest limit is applied? The link you added to the limit may be on the bucket.

ivasilov (approved):

PR #29809, opened 2024-10-10
Fix support form again

joshenlim:
Related to https://github.com/supabase/supabase/pull/29794, just starting from scratch

From @kevcodez in the attached PR:
Currently customers are forced to select an organization to submit the form. However, when the customer has no orgs, they are in a deadlock.

This PR allows customer to submit the form without an org selected, BUT only if they do not have any orgs. That is to prevent people from forgetting to select their org while creating a ticket related to an org. This is the least invasive change while still allowing customers to create a ticket.

Shadcn SelectItem doesn't allow null/undefined/empty string so we gotta do some nonsense to represent the empty value state here.


alaister (approved):
LGTM!

- Tested on preview by submitting a ticket from an org with and without orgs

PR #29810, opened 2024-10-10
Chore/minor fix in events telemetry

joshenlim:
- Use API types for SendEventGA and SendEventPH
- Update userAgent to user_agent
- Update API types from api


alaister (approved):
LGTM!

- Tested on preview

PR #29819, opened 2024-10-10
fix(docs): ref scroll

charislam:
Fix various scrolling bugs with reference pages:

- Take out smooth scrolling as it tends to lead to buggy scrolling position
- Add overscroll-behavior: contain to the nav, otherwise overscrolling on the nav scrolls the main page, which causes the nav to jump around unexpectedly
- Put aria-current calculation in useEffect as otherwise it doesn't trigger properly on first load

See https://supabase.slack.com/archives/C023E4L60R3/p1728116359237709

### DIFF
--- a/apps/docs/features/docs/Reference.navigation.client.tsx
+++ b/apps/docs/features/docs/Reference.navigation.client.tsx
@@ -11,9 +11,9 @@ import { createContext, useCallback, useContext, useEffect, useMemo, useRef, use
 
 import { cn } from 'ui'
 
-import { BASE_PATH } from '~/lib/constants'
 import type { AbbrevApiReferenceSection } from '~/features/docs/Reference.utils'
 import { isElementInViewport } from '~/features/ui/helpers.dom'
+import { BASE_PATH } from '~/lib/constants'
 
 export const ReferenceContentInitiallyScrolledContext = createContext<boolean>(false)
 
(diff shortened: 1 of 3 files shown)


apps/docs/features/docs/Reference.navigation.client.tsx
```diff
@@ -60,14 +58,33 @@ export function ReferenceNavigationScrollHandler({
...
-      currentLink.scrollIntoView({
-        block: 'center',
+      // Calculate the offset of the current link relative to scrollingParent
+      // and scroll the parent to the top of the link.
+      const offsetTop = currentLink.offsetTop
+      const parentOffsetTop = parentRef.current?.offsetTop ?? 0
+      const scrollPosition = offsetTop - parentOffsetTop
+
+      parentRef.current?.scrollTo({
+        top: scrollPosition - 60 /* arbitrary padding */,
```
avallete, line 87: **question**

If that's meant to offset the scrolling, doesn't that match the "top navbar" height (50px) + a bit of padding ?

avallete (approved):
LGTM

fsansalvadore (approved):

apps/docs/features/docs/Reference.navigation.client.tsx
charislam: yep that's true, will update comment 👍🏼

PR #29825, opened 2024-10-10
Chore/search user json

saltcod:
Add search to json panel in new User Management  table

screenshot-2024-10-10-at-17 20 41

### DIFF
--- a/apps/studio/components/interfaces/Auth/Users/UserPanel.tsx
+++ b/apps/studio/components/interfaces/Auth/Users/UserPanel.tsx
@@ -5,6 +5,7 @@ import { useState } from 'react'
 import {
   Button,
   cn,
+  Input_Shadcn_,
   ResizableHandle,
   ResizablePanel,
   Tabs_Shadcn_,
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Auth/Users/UserPanel.tsx
```diff
@@ -76,8 +93,26 @@ export const UserPanel = ({ selectedUser, onClose }: UserPanelProps) => {
             value="raw"
             className={cn('mt-0 flex-grow min-h-0 overflow-y-auto', PANEL_PADDING)}
           >
+            <div className="flex items-center mb-2">
+              <Input_Shadcn_
+                autoFocus
+                type="text"
+                placeholder="Search..."
```
alaister, line 100: This is more of a filter don't you think?
```suggestion
                placeholder="Filter..."
```

alaister (approved):
LGTM!

- Tested on preview

One future improvement: flatten out the nested objects like `raw_app_meta_data` so I can filter by provider for example

PR #29848, opened 2024-10-11
add loki log drain form

jordienr:
you can use something like [webhook.cool](https://webhook.cool/) to test the log drains


apps/studio/components/interfaces/LogDrains/LogDrainDestinationSheetForm.tsx
```diff
@@ -155,6 +161,8 @@ export function LogDrainDestinationSheetForm({
       url: defaultValues?.config?.url || '',
       api_key: defaultValues?.config?.api_key || '',
       region: defaultValues?.config?.region || '',
+      header1_key: defaultValues?.config?.header1_key || '',
```
Ziinc, line 164: Should use the same header logic as http destination, can be one or more headers with the application/json header set by default

Ziinc: Should be the `headers` key

Ziinc (changes requested):

Ziinc (approved):

saltcod:
screenshot-2024-10-21-at-12 06 52

saltcod (approved):

PR #29851, opened 2024-10-11
fix: update curl snippet for invite user

J0:

YES

Invite is an admin endpoint and should use a service role key. 

Updates a snippet, using old terminology for now (service_role_key) we will need to update this when we overhaul the naming for keys.

image


saltcod (approved):

saltcod:
checked preview

PR #29866, opened 2024-10-13
Update docs with "make" requirement

kjrocker:

YES

Docs update

Doc recommends git, npm, node, and docker

## What is the new behavior?

Adding requirement for `make` or some equivalent to the `build-essentials` package, based on my experience with the installation on a totally fresh system.

## Additional context

Fairly self explanatory. I ran the setup on a very fresh system, and ran into problems running `npm install` because some packages had build steps that required `make`.


charislam (approved):
Thanks @kjrocker!

PR #29881, opened 2024-10-14
feat: support for restarting and resizing statuses

delgado3d:
1. Adds support for restarting and resizing statuses.
2. Adds a new Restarting (resizing) page to be used while compute is being resized
3. Have tested these changes work as expected:
    - compute size changes
    - Project restart & database restart
    - Read replica viewer & restarts (Infrastructure page)

Have used 'restarting' language in few places while project is being resized to indicate that a full project restart is involved.
^^ lmk if you think it is better to explicitly say 'resizing' instead.


joshenlim (approved):
LGTM! Tested on preview
- Updating compute instance
- Restarting a replica
- Restarting a project

PR #29883, opened 2024-10-14
fix: remove incorrect usage of restoring state for restarts

delgado3d:
- Database status is not expected to be RESTORING anymore for restarts
- RESTORING is also not expected to be a valid / healthy status for database anymore
- Goes on top off https://github.com/supabase/supabase/pull/29881


joshenlim (approved):
Code wise LGTM - tested on staging preview by restarting a replica - correct badge is shown
image

PR #29891, opened 2024-10-15
Update PH consent banner

joshenlim:
- Add telemetry consent check within all telemetry RQ mutations (except send-reset since that shouldn't need a check)
  - Ensures that cookies are only placed in users' browsers after they've opted in
- Update consent banner copy
- Add new local storage key for telemetry PH based on feature flag to reset user's preferences for opting in


alaister (approved):
LGTM!

-Tested on preview

note:

PR #29896, opened 2024-10-15
chore: Bump sentry to v8+

ivasilov:


saltcod (approved):
checked preview, sentry & code looks good

PR #29899, opened 2024-10-15
fix: Don't send the request twice when saving a policy from the Policy editor

ivasilov:


saltcod (approved):

saltcod:
Tested creating and editing policies.

PR #29914, opened 2024-10-16
Add telemetry page leave for www and docs

joshenlim:


4L3k51 (approved):

apps/www/pages/_app.tsx
```diff
@@ -64,6 +64,14 @@ export default function App({ Component, pageProps }: AppProps) {
...
   }
 
+  function handlePageLeaveTelemetry(url: string) {
+    return post(`${API_URL}/telemetry/page-leave`, {
+      page_url: url,
+      page_title: title,
+      pathname: router.pathname,
+    })
+  }
+
```
alaister, line 74: Any reason this needs to be in a separate function? Could it be moved into `handleBeforeUnload`?

joshenlim: not really tbh - was following the same way handlePageTelemetry was written in case there's some dev pattern going on - but good point i'll just shift it in 🙏

alaister (approved):
LGTM!

- Approving on behalf of dashboard

PR #29924, opened 2024-10-16
fix(docs): integrations sidebar linking to missing partners

charislam:
Two related bugs:
- Not filtering partners on approved: true
- Not revalidating occasionally to pick up DB changes


apps/docs/app/guides/(with-sidebar)/layout.tsx
```diff
@@ -1,8 +1,12 @@
-import { type PropsWithChildren } from 'react'
+import { cache, type PropsWithChildren } from 'react'
 
 import { supabaseMisc } from '~/lib/supabaseMisc'
 import Layout from '~/layouts/guides'
 
+// Revalidate occasionally to pick up changes to partners
+// 60 seconds/minute * 60 minutes/hour * 24 hours/day
+export const revalidate = 86_400
```
ivasilov, line 8: We used to have this set to 5 hours and it still showed outdated info. The most common issue was when the request to Supabase would fail and the page would show 404 for the next 5 hours. Can we handle that error case with app layouts?

charislam: With App Router, it shouldn't 404 on error, it will just continue showing stale data and retrying on each request until it fetches successfully.

https://nextjs.org/docs/canary/app/building-your-application/data-fetching/incremental-static-regeneration#handling-uncaught-exceptions

fsansalvadore (approved):

PR #29936, opened 2024-10-17
Add client side perms checking for disk config

joshenlim:
As per PR title - Developer roles should not be able to update disk config


alaister (approved):
LGTM!

- Tested on preview

PR #29937, opened 2024-10-17
fix: storage permissions

alaister:

fix


phamhieu (approved):

joshenlim (approved):
LGTM! Test on preview

- As a developer role i can upload files, rename files, and delete files
- As a read only role i _cannot_ upload files, rename files, and delete files

lemme know if i might have missed something here!

PR #29950, opened 2024-10-17
Fix: unify Client ID handling for Google provider

monicakh:

CLIENT_ID fields should be allowed  to accept multiple client IDs for both Apple and Google authentication. Now, users can specify more than one client ID by providing them as a comma-separated string. 

## Additional context

image

image


apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
```diff
@@ -833,8 +833,9 @@ const EXTERNAL_PROVIDER_GOOGLE = {
       type: 'boolean',
     },
     EXTERNAL_GOOGLE_CLIENT_ID: {
-      title: 'Client ID (for OAuth)',
-      description: 'Client ID to use with the OAuth flow on the web.',
+      title: 'Client IDs',
+      description:
+        'Comma-separated list of client IDs for OAuth, Android apps, One Tap, and Chrome extensions.',
```
kangmingtay, line 838:
```suggestion
        'Comma-separated list of client IDs for Web, OAuth, Android apps, One Tap, and Chrome extensions.',
```

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
```diff
@@ -843,78 +844,32 @@ const EXTERNAL_PROVIDER_GOOGLE = {
...
-            'At least one Authorized Client ID is required when not using the OAuth flow.'
+      .required('At least one Client ID is required when Google sign-in is enabled.'),
+    EXTERNAL_GOOGLE_SECRET: string().when('EXTERNAL_GOOGLE_ENABLED', {
+      is: true,
+      then: (schema) =>
+        schema
+          .required('Client Secret is required when using the OAuth flow.')
+          .matches(
+            /^[a-z0-9.\/_-]*$/i,
+            'Invalid characters. Google OAuth Client Secrets usually contain letters, numbers, dots, dashes, and underscores.'
```
kangmingtay, line 870: does this handle the case where one should be allowed to add a client id without a client secret?

kangmingtay (commented):
@monicakh we also need to update the validation rules for all other providers that use the "Additional Client IDs" field (e.g. apple)

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
monicakh: Just tested it, and it does. It treats the client secret as an optional field. 

image

joshenlim:
@monicakh just wanted to check in - is this PR ready for us to take a look?

monicakh:
> @monicakh just wanted to check in - is this PR ready for us to take a look?

monicakh:
@joshenlim Should be ready when @kangmingtay accept the new changes for apple provider.

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
monicakh: The Secret Key is only required if Google or Apple provider is enabled.

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
kangmingtay: hmm this should also mention `Web` as per the suggestion right?

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
```diff
@@ -473,47 +467,27 @@ const EXTERNAL_PROVIDER_APPLE = {
...
+      .matches(
+        /^([a-z0-9-]+\.[a-z0-9-]+(\.[a-z0-9-]+)*(,[a-z0-9-]+\.[a-z0-9-]+(\.[a-z0-9-]+)*)*)$/i,
+        'Invalid characters. Each ID should follow a reverse-domain style string (e.g. com.example.app). Use commas to separate multiple IDs.'
+      )
+      .when('EXTERNAL_APPLE_ENABLED', {
+        is: true,
         then: (schema) =>
           schema.required(
-            'At least one Authorized Client ID is required when not using the OAuth flow.'
+            'At least one Service ID or bundle ID is required when Apple sign-in is enabled.'
```
kangmingtay, line 490: maybe this is more inline with the description of the field?
```suggestion
            'At least one Client ID is required when Apple sign-in is enabled.'
```

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
kangmingtay: hmm but if the provider is not enabled, how do they edit those fields? I was under the impression that you can't edit the client ID and secret fields without first enabling the provider? 

also, what happens if the provider is enabled? will it allow the user to add only the client id field?

monicakh: @kangmingtay I didn’t make changes to this functionality. If you check in production, users can currently edit the Client ID and Secret fields even when the provider isn’t enabled. 

But if the provider is enabled, users will be required to add at least one Client ID and a Secret now. Should I set it up so that only the Client ID is required?

kangmingtay: > Should I set it up so that only the Client ID is required?
yup this should be allowed - can you also check if it allows saving only the client ID when the provider is enabled? if it's doesn't allow that, i can help to make the corresponding changes in the backend too

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
```diff
@@ -399,9 +399,8 @@ const EXTERNAL_PROVIDER_APPLE = {
       type: 'boolean',
     },
     EXTERNAL_APPLE_CLIENT_ID: {
-      title: 'Service ID (for OAuth)',
-      description: `Client identifier used in the OAuth flow on the web.
-[Learn more](https://developer.apple.com/documentation/sign_in_with_apple/configuring_your_environment_for_sign_in_with_apple)`,
+      title: 'Client IDs (OAuth, iOS, macOS, watchOS, tvOS bundle IDs or service IDs)',
```
kangmingtay, line 402:
```suggestion
      title: 'Client IDs (Web, OAuth, iOS, macOS, watchOS, tvOS bundle IDs or service IDs)',
```

kangmingtay (approved):
changes seem good - one thing to note is that we have some docs that still refer to the "Authorized Client IDs" field like [this one](https://supabase.com/docs/guides/auth/social-login/auth-google) which might be confusing if it's not updated

also, users who are building mobile apps may find this change confusing as the "Authorized Client IDs" field is now removed and combined with the "Client IDs" field - i'm thinking if we should add a callout for this in the form section?

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
```diff
@@ -399,9 +399,8 @@ const EXTERNAL_PROVIDER_APPLE = {
       type: 'boolean',
     },
     EXTERNAL_APPLE_CLIENT_ID: {
-      title: 'Service ID (for OAuth)',
-      description: `Client identifier used in the OAuth flow on the web.
-[Learn more](https://developer.apple.com/documentation/sign_in_with_apple/configuring_your_environment_for_sign_in_with_apple)`,
+      title: 'Client IDs (Web, OAuth, iOS, macOS, watchOS, tvOS bundle IDs or service IDs)',
```
MildTomato, line 402: Screenshot 2024-11-04 at 17 53 24

this is an enormous label

MildTomato: the content in parenthesis is probably a description, or optional label

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
```diff
@@ -843,78 +815,34 @@ const EXTERNAL_PROVIDER_GOOGLE = {
...
-      /^([a-z0-9-]+([.][a-z0-9-]+)+)?$/i,
-      'Invalid characters. Google OAuth Client IDs are usually a domain-name (e.g. 01234567890-abcdefghijklmnopqrstuvwxyz012345.apps.googleusercontent.com).'
-    ),
-    EXTERNAL_GOOGLE_ADDITIONAL_CLIENT_IDS: string()
+    EXTERNAL_GOOGLE_CLIENT_ID: string()
+      .matches(/^\S+$/, 'Client IDs should not contain spaces.')
       .matches(
-        /^([a-z0-9-]+([.][a-z0-9-]+)*(,\s*[a-z0-9-]+([.][a-z0-9-]+)*)*,*\s*)?$/i,
-        'Invalid characters. Google Client IDs are usually a domain-name style string (e.g. com.example.com.app or *.apps.googleusercontent.com).'
+        /^([a-z0-9-]+\.[a-z0-9-]+(\.[a-z0-9-]+)*(,[a-z0-9-]+\.[a-z0-9-]+(\.[a-z0-9-]+)*)*)$/i,
```
MildTomato, line 830: could we not do a OR rule here.

if there's no comma we shouldn't be returning a validation error about being comma separated.

MildTomato (changes requested):

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
MildTomato: might be a mix up here, as the whole form with all values get submitted on each submit.
So 'just changing the ID' will mean all values are submitted.

MildTomato: in other words, all required fields need to have a value if 'enabled' is true.
you can however just change 1 input and fire the form, but technically your submitting all values.

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
monicakh: @MildTomato We can change it to: `Client IDs (Web, OAuth, Apple Platforms, or Service IDs)`

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
monicakh: @MildTomato The validation error is needed here because users can now submit multiple IDs in that field. Also, if a user inputs only one client ID, it won’t raise any validation errors, as the check only applies when multiple IDs are expected in a comma-separated format so we can parse them later.

monicakh: I've tested it again for both Apple and Google and it doesn't return any error when one Client ID is submitted:

image

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
joshenlim: agree with Jonny here - this label is really long and looks odd
+1 on moving it to description instead

Something like this feels better:
image

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
```diff
@@ -399,9 +399,8 @@ const EXTERNAL_PROVIDER_APPLE = {
       type: 'boolean',
     },
     EXTERNAL_APPLE_CLIENT_ID: {
-      title: 'Service ID (for OAuth)',
-      description: `Client identifier used in the OAuth flow on the web.
-[Learn more](https://developer.apple.com/documentation/sign_in_with_apple/configuring_your_environment_for_sign_in_with_apple)`,
+      title: 'Client IDs (Web, OAuth, iOS, macOS, watchOS, tvOS bundle IDs or service IDs)',
+      description: `Comma separated list of allowed Apple app bundle IDs for native sign in, or service IDs for Sign in with Apple JS. [Learn more](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js)`,
```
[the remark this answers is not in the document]
joshenlim: +1 on this - the markdown component should already handle this actually

joshenlim:
Found a bug

If i save one Apple Client ID, it works okay, `EXTERNAL_APPLE_CLIENT_ID` is saved correctly
POST request payload:
image
Response:
image

But if i save multiple, `EXTERNAL_APPLE_ADDITIONAL_CLIENT_IDS` gets updated instead and `EXTERNAL_APPLE_CLIENT_ID` is now null - this causes the input field to be empty when i refresh the page
POST request payload:
image
Response:
image
After refreshing the page:
image

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
MildTomato: that's better

apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
joshenlim: ^ okay can verify those cases pass!

MildTomato: ah it looks ok, just checked it myself

joshenlim (approved):
LGTM! Tested on preview that saving client IDs for apple and google works. Also validated the regex for checking the client Ids

MildTomato:
this feels like we've made the API worse. 

why do we have a EXTERNAL_ID and then a EXTERNAL_ADDITONAL_IDS ?

joshenlim:
we're removing `EXTERNAL_ADDITONAL_IDS` - thats the change here i think

MildTomato (approved):

PR #29963, opened 2024-10-17
Chore/table header simplification

saltcod:
- tighten up header area
- implement intermediate state in select all button
- hide grid header actions when selections in place
 
screenshot-2024-11-01-at-16 53 32


alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/grid/components/footer/Footer.tsx
```diff
@@ -20,6 +21,9 @@ const Footer = ({ isRefetching }: FooterProps) => {
   const { id: _id } = useParams()
   const id = _id ? Number(_id) : undefined
 
+  const state = useTrackedState()
+  const { selectedRows, allRowsSelected } = state
```
joshenlim, line 25: this seems unused - i'll remove it

apps/studio/components/grid/components/grid/SelectColumn.tsx
```diff
@@ -183,13 +183,27 @@ function SelectCellHeader({
...
+  const state = useTrackedState()
+  const { selectedRows, allRowsSelected } = state
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  // indeterminate state === some rows are selected but not all
+  const isIndeterminate = selectedRows.size > 0 && !allRowsSelected
+
+  useEffect(() => {
+    if (inputRef.current) {
+      inputRef.current.indeterminate = isIndeterminate
```
joshenlim, line 195: hmm what does this do? are we using inputRef.indeterminate?

joshenlim: will remove - seems like its unused

apps/studio/components/grid/components/header/Header.tsx
```diff
@@ -377,23 +380,24 @@ const RowHeader = ({ table, sorts, filters }: RowHeaderProps) => {
...
-        </span>
-        {!allRowsSelected && totalRows > allRows.length && (
-          <Button type="link" onClick={() => onSelectAllRows()}>
-            Select all rows in table
-          </Button>
-        )}
-      </div>
-      <div className="h-[20px] border-r border-strong" />
       <div className="flex items-center gap-2">
+        <Tooltip_Shadcn_>
```
joshenlim, line 384: hmm the order of the buttons feel weird it terms of context
- we've got clear selection (which is to do with row selection)
- followed by export and delete
- then select all and the number of rows (which is to do with row selection)

image

joshenlim: any reason why we're changing from the current order of buttons?
image

apps/studio/components/grid/components/header/Header.tsx
```diff
@@ -420,7 +424,7 @@ const RowHeader = ({ table, sorts, filters }: RowHeaderProps) => {
               <Button
                 type="default"
                 size="tiny"
-                icon={<Trash />}
+                icon={<Trash className="text-destructive-600" />}
```
joshenlim, line 427: hmm from a consistency POV: is this something we want to keep doing as a UI pattern?

joshenlim (approved):
LGTM! Tested header changes on the preview
- with at least one row selected but not all, header checkbox has a "-" icon
- Export to CSV + SQL works for selected rows (basic testing, didn't stress test this)

PR #29980, opened 2024-10-18
fix copy btn alignment, fix titles

jordienr:
before
CleanShot 2024-10-18 at 12 09 38@2x

after
CleanShot 2024-10-18 at 12 10 02@2x


alaister (approved):
LGTM!

- Tested on preview

One tiny thing I noticed that I'm not sure if it was intentional is the copy button here is quite far:

Screenshot 2024-10-21 at 18 14 38

PR #29983, opened 2024-10-18
added vercel-postgres migration guide 'Migrate from Vercel Postgres t…

burggraf:
This adds a new migration guide:
"Migrate from Vercel Postgres to Supabase"

### DIFF
--- a/apps/docs/content/guides/platform/migrating-to-supabase/vercel-postgres.mdx
+++ b/apps/docs/content/guides/platform/migrating-to-supabase/vercel-postgres.mdx
@@ -4,9 +4,6 @@ description: 'Migrate your existing Vercel Postgres database to Supabase.'
 subtitle: 'Migrate your existing Vercel Postgres database to Supabase.'
 ---
 
-This is a guide for migrating your Vercel Postgres database to [Supabase](https://supabase.com).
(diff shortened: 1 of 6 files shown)


apps/docs/content/guides/platform/migrating-to-supabase/vercel-postgres.mdx
```diff
@@ -0,0 +1,101 @@
+---
+title: 'Migrate from Vercel Postgres to Supabase'
+description: 'Migrate your existing Vercel Postgres database to Supabase.'
+subtitle: 'Migrate your existing Vercel Postgres database to Supabase.'
+---
+
+This is a guide for migrating your Vercel Postgres database to [Supabase](https://supabase.com).
+Supabase is a robust and open-source platform. Supabase provide all the backend features developers need to build a product: a Postgres database, authentication, instant APIs, edge functions, realtime subscriptions, and storage. Postgres is the core of Supabase—for example, you can use row-level security and there are more than 40 Postgres extensions available.
```
charislam, line 8:
```suggestion
```

Let's take this out, this is more of a how-to than a marketing pitch for Supabase, I don't think anyone really gets around to reading a migration guide without already being somewhat familiar with the product they're migrating to

charislam (approved):

apps/docs/content/guides/platform/migrating-to-supabase/vercel-postgres.mdx
burggraf: Sounds good.  I just lifted this from the postgres migration guide, so we might want to delete it on that one, too.

burggraf (commented):
Thanks Charis

PR #29986, opened 2024-10-18
Pass search term to entity query

saltcod:
The entities query powering the table dropdown only shows 100 tables by default.  There's a search input field here, but the search term isn't getting passed to the query. So if you're looking to index your 101st table, you won't find it in this list. 

https://supabase.com/dashboard/project/_/database/indexes

The search term will filter this local list, but it won't cause the query to refetch results

screenshot-2024-10-18-at-16 17 41


jordienr (approved):

jordienr:
tested by creating 100+ tables and searching, all good

PR #29987, opened 2024-10-18
chore: aggressive prefetching

alaister:


apps/studio/components/interfaces/TableGridEditor/EmptyState.tsx
```diff
@@ -29,6 +30,7 @@ const EmptyState = ({}: EmptyStateProps) => {
     connectionString: project?.connectionString,
     schemas: [selectedSchema],
     sort,
+    filterTypes: Object.values(ENTITY_TYPE),
```
joshenlim, line 33: maybe this can be a default value on useEntityTypesQuery - i realised we sort of do this here
but filterTypes should have been defaulted in the prop declaration instead
image

apps/studio/components/layouts/ProjectLayout/NavigationBar/NavigationBar.utils.tsx
```diff
@@ -11,8 +13,9 @@ import {
   Storage,
   TableEditor,
 } from 'icons'
+import { IS_PLATFORM, PROJECT_STATUS } from 'lib/constants'
+import { Blocks } from 'lucide-react'
```
joshenlim, line 17: tinyyy tinyy nit - can merge with the lucide react import above

apps/studio/data/database/foreign-key-constraints-query.ts
```diff
@@ -107,19 +129,32 @@ export const useForeignKeyConstraintsQuery = <
       projectRef,
       connectionString,
       sql: getForeignKeyConstraintsQuery({ schema }),
-      queryKey: ['foreign-key-constraints'],
+      queryKey: ['foreign-key-constraints', schema],
```
joshenlim, line 132: nit: `[schema, 'foreign-key-constraints']`? if we wanna invalidate foreign key constraints (assumingly at a schema level) it'll be easier

apps/studio/data/database/schemas-query.ts
```diff
@@ -39,3 +44,20 @@ export const useSchemasQuery = <TData = SchemasData>(
 export function invalidateSchemasQuery(client: QueryClient, projectRef: string | undefined) {
   return client.invalidateQueries(sqlKeys.query(projectRef, ['schemas', 'list']))
 }
+
+export function prefetchSchemas(
```
joshenlim, line 48: this is just a thought - should we follow the same api as what you wrote for `usePrefetchEditorTablePage` ? 

so this becomes a hook that returns a callback, then we don't need to separately declare `const queryClient = useQueryClient()` and pass it in - queryClient can be declared in usePrefetchSchemas instead, one prop less

joshenlim: ^ will apply for `prefetchEntityType` too

apps/studio/data/prefetchers/project.$ref.editor.$id.tsx
```diff
@@ -0,0 +1,182 @@
...
+              schema: entity?.schema,
+              tableName: entity?.name,
+            }),
+            undefined,
+          ]
+
+          switch (entity?.type) {
+            case ENTITY_TYPE.TABLE:
+            case ENTITY_TYPE.PARTITIONED_TABLE:
+              promises[3] = prefetchTable(queryClient, {
```
joshenlim, line 84: just a nit - i'd prefer immutability usually - should we just do promises.push() instead of reassigning index 3 instead? then we don't need the undefined as the last item in promises

alaister: Yeah I'd rather use `.push()` too, the reason it's this way is to make typescript happy since I'm using a tuple instead of a regular array. So it knows `[3]` should be of type whatever `prefetchTable` returns

joshenlim (approved):
LGTM - tested on preview that perceived loading speed around the table editor is improved, although we need to check on one thing though that the call to `/views` seems to be triggered each time i hover over a table link in the table editor

joshenlim (approved):
LGTM once again! Tested the prefetching logic in table editor and also checking that the table def query doesnt error out with a foreign table!

PR #29996, opened 2024-10-19
Fixes item description and unit price of Billing Breakdown

ecktoteckto:
Fixes the Billing Breakdown regarding item description and unit price.

**Item description:** for "Log Drain", "Auth MFA Phone" and "Auth MFA WebAuthn" the "(Hours)" was missing
**Unit price:** was not shown for items without project breakdown

**Side note regarding formatting:** Now that the unit price description is also shown for items without a project breakdown we are running out of space for prices with Tiered Pricing (e.g. Auth MFA Phone) with ugly word wrapping in column "Item". I played around with abbreviations (e.g. "hr" instead of "hour", "$0.1027/hr" instead of "$0.1027 per hour) but that didn'f feel great. So I limited the max width of the column "Unit price". It's a bit dense but at least complete and understandable.

Screenshot 2024-10-19 at 19 17 43


kevcodez (approved):
Checked staging, nothing broke, but given usage pipeline on staging is not working, will merge and double-check on prod

PR #30022, opened 2024-10-22
Update auth hooks to allow selection of fns that return void

joshenlim:
As per PR title, cc @J0 - we should be able to use a DB function that returns void specifically for the send email hook only


J0 (approved):

alaister (approved):
LGTM!

- Tested on preview

PR #30024, opened 2024-10-22
fix: allow developer role to update buckets

alaister:

Bug fix

`PermissionAction.STORAGE_ADMIN_WRITE` is used.

## What is the new behavior?

Use `PermissionAction.STORAGE_WRITE` instead.


joshenlim (approved):
LGTM! Tested the changes locally earlier can verify that developer role can update and create buckets now

PR #30041, opened 2024-10-22
fix: Optimize json data when including it in the www nav

ivasilov:
This PR adds a new script to the `www` app which reads the blog posts data and picks fields from it which are to be used in the top nav. This avoids bundling all blog posts (including content) into the nav on every single page.


fsansalvadore (approved):
Nice, the preview looks all good 👍

PR #30053, opened 2024-10-23
fix: avatar url startsWith error

alaister:

Bug fix

Sometimes `x.startsWith is not a function` errors due to users being able to put anything in these fields

## What is the new behavior?

Only check startsWith on strings


joshenlim (approved):
LGTM - i'm not sure how to repro this issue on the preview, but i forced a repro locally, seem okay!

PR #30055, opened 2024-10-23
Chore/minor fix scratch pad

joshenlim:
- Tiny padding fix to ensure header for scratch pad matches height of assistant header
- Only allow opening scratch pad if feature preview is enabled


alaister (approved):
LGTM!

- Tested on preview

PR #30071, opened 2024-10-24
fix log drain permissions

jordienr:


joshenlim (approved):
LGTM - this looks right, settings -> should use ADMIN_WRITE (similar to storage)

PR #30079, opened 2024-10-24
chore: increase react-query staleTime on some queries

alaister:

A few queries are always stale causing lots of excess fetching for little benefit since we manually invalidate when things change. We also offer refresh buttons on most data views in the dashboard.

## What is the new behavior?

Most queries now follow the same rule:
`react-queries are revalidated at most once every 5 minutes or when manually invalidated`


alaister:
Superseded by #30437

PR #30080, opened 2024-10-24
Loki log drain update & ui fix

jordienr:
- truncate log drains table text
- add username and password inputs for loki log drains


joshenlim (approved):
LGTM! Tested on preview creating a log drain with username and password + stress test truncation. Just added a title prop so users can hover over and read the whole text if truncated

PR #30088, opened 2024-10-25
Ensure commands only show up if assistant is enabled

joshenlim:


alaister (approved):
LGTM!

- Tested on preview

PR #30095, opened 2024-10-25
Fix foreign key mgt table editor open in new tab

joshenlim:
For example: clicking on `auth.users` here will open the `auth.users` table in the table editor,  but if you're midway editing the table (e.g adding a new foreign key) the panel closes and you lose your progress
image

Change here ensures that the button opens the referenced table in a new tab (understandably you just want to see the table, you don't necessarily want to switch context entirely)

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/ForeignKeysManagement/ForeignKeyRow.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/ForeignKeysManagement/ForeignKeyRow.tsx
@@ -78,9 +78,10 @@ export const ForeignKeyRow = ({
               }
             >
               <EditorTablePageLink
+                target="_blank"
+                rel="norefererer"
                 projectRef={ref}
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/ForeignKeysManagement/ForeignKeyRow.tsx
```diff
@@ -78,9 +78,10 @@ export const ForeignKeyRow = ({
               }
             >
               <EditorTablePageLink
```
alaister, line 80: If we're opening in a new tab there's no reason to spend network resources on prefetching table data... can you please swap this for a regular next/link? 🙏

joshenlim: oh yes! that makes sense haha

alaister (approved):
LGTM!

- Tested on preview

PR #30096, opened 2024-10-25
fix: api types + remove old telemetry

alaister:

Bug fix

API types were out of sync due to https://github.com/supabase/infrastructure/pull/20333

## What is the new behavior?

- Fixes API types
- Removes codepaths for old telemetry endpoints + `enablePosthogChanges` feature flag


apps/studio/components/ui/PageTelemetry.tsx
```diff
@@ -34,41 +33,34 @@ const PageTelemetry = ({ children }: PropsWithChildren<{}>) => {
 
   const consentToastId = useRef<string | number>()
   const previousPathname = usePrevious(router.pathname)
-  const enablePostHogTelemetry = useFlag('enablePosthogChanges')
 
-  const consentFlag = enablePostHogTelemetry
-    ? LOCAL_STORAGE_KEYS.TELEMETRY_CONSENT_PH
-    : LOCAL_STORAGE_KEYS.TELEMETRY_CONSENT
+  const consentFlag = LOCAL_STORAGE_KEYS.TELEMETRY_CONSENT
```
joshenlim, line 37: this should be TELEMETRY_CONSENT_PH right?

apps/studio/data/telemetry/send-event-mutation.ts
```diff
@@ -48,22 +38,21 @@ export const useSendEventMutation = ({
   'mutationFn'
 > = {}) => {
   const router = useRouter()
-  const usePostHogParameters = useFlag('enablePosthogChanges')
 
   const consent =
```
joshenlim, line 42: nit - just wondering whether this can be shifted to sendEvent now, then we don't need to pass consent as a prop

previously i did this just because of the feature flag check

joshenlim: but would apply for all the other telemetry mutations

alaister: Moved!

joshenlim (approved):
LGTM! Tested on staging preview - telemetry events are getting sent successfully when opted in, otherwise if not opted in, no telemetry events are sent

PR #30097, opened 2024-10-25
Fix missing enforceExactcount param

joshenlim:


apps/studio/components/grid/components/header/Header.tsx
```diff
@@ -270,6 +271,7 @@ const RowHeader = ({ table, sorts, filters }: RowHeaderProps) => {
   }
 
   const onRowsDelete = () => {
+    console.log({ totalRows, countData })
```
saltcod, line 274:
```suggestion

```

saltcod:
created a table with 500k rows, confirmed expected behaviour

saltcod (approved):

PR #30098, opened 2024-10-25
feat: use full cloudflare hostname for DNS lookup

delgado3d:
Some networks can block direct IP address access (`https://1.1.1.1/`) which can error out the custom domain setup process. Updated it to use `https://one.one.one.one/` instead which also additionally supports IPv6-only networks.

Have successfully tested setting up custom domain on the staging preview.


pcnc (approved):

joshenlim (approved):
LGTM! Testing on preview that the new URL works, also pushed a smol fix on the "cancel" loading state, as well as an unrelated icon fix in oauth apps

PR #30099, opened 2024-10-25
Use POSTGRES_DB variable everywhere in docker-compose.yml

TBG-FR:

YES ✅ 

Replacing remaining `_supabase` references with the appropriate .env variable `POSTGRES_DB` inside `docker-compose.yml`.

`POSTGRES_DB` variable is used to configure `db` service (and its pg database), but some services keep using a `_supabase` string instead of this variable (or a mix of both 🤯). An user that would be configuring `POSTGRES_DB` to something else could break his setup, if he doesn't see and change `_supabase` strings.

## What is the new behavior?

`POSTGRES_DB` variable is used everywhere, hence an user that would be configuring `POSTGRES_DB` to something else would have nothing to do, `analytics` and `supavisor` would have the correct database name right away.

## Additional context

N/A

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -343,7 +343,7 @@ services:
     environment:
       LOGFLARE_NODE_HOST: 127.0.0.1
       DB_USERNAME: supabase_admin
-      DB_DATABASE: _supabase
+      DB_DATABASE: ${POSTGRES_DB}
       DB_HOSTNAME: ${POSTGRES_HOST}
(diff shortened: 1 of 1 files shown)


djsisson:
@TBG-FR  The _supabase db, was to keep logs and tenants out of the main postgres database

Hallidayo:
Hi @TBG-FR  - We’re doing a bit of housekeeping and plan to close out PRs that have been inactive for more than 3 months.

PR #30102, opened 2024-10-25
fix log filters

jordienr:


saltcod (approved):

saltcod:
checked preview, logs filters now work

PR #30106, opened 2024-10-25
Chore/fix billing metric hovercard

saltcod:
fix link inside link which is messing up hover state here
https://supabase.com/dashboard/org/_/usage

screenshot-2024-10-25-at-17 08 04


kevcodez (approved):
Tested on staging, both the hovercard and link work

PR #30133, opened 2024-10-28
fix: users displayName error when not strings

alaister:

Bug fix

If an object is in `raw_user_meta_data` for any of the name fields, React crashes with an error.

## What is the new behavior?

Always cast the name to a string. Most people shouldn't have objects for their name fields, so this is more just a fallback to prevent crashing


joshenlim (approved):
LGTM! Verified locally by manually updating a user's raw_user_meta_data object

PR #30163, opened 2024-10-29
Update all documentation buttons

joshenlim:
- Change icon from ExternalLink to OpenBook
- Shorten text from "Documentation" to "Docs" -> in efforts to take less space in areas that are a little more cramped
  - Am opting to not abbreviate the text if used in an admonition as we don't need to worry about the space then (Example below)

### Before:
image

### After:
image

Screenshot 2024-10-29 at 16 49 13


apps/studio/components/interfaces/Organization/TeamSettings/UpdateRolesPanel/UpdateRolesPanelOld.tsx
```diff
@@ -1,232 +0,0 @@
-import { ExternalLink } from 'lucide-react'
```
alaister, line 1: Can this file be deleted?

alaister (approved):
LGTM!

- Clicked around the preview dashboard and clicked various docs buttons

saltcod (approved):

apps/studio/components/interfaces/Organization/TeamSettings/UpdateRolesPanel/UpdateRolesPanelOld.tsx
joshenlim: yup! this one's no longer being used as it's the old org team role management panel before project level permissions

PR #30177, opened 2024-10-29
fix: fixes org usage summary

ecktoteckto:
Org usage summary broke, now that Mgmt-API returns "hours" for all metrics (introduced [here](https://github.com/supabase/infrastructure/pull/20401)). It's not the most elegant fix but should do for now. I'm happy to do find a more elegant solution tomorrow morning.

**Beore GB vs Byte:**
before-1

**After GB vs Byte:**
after-1

**Before Metric not billed by hour:**
before-2

**After Metric not billed by hour:**
after-2


saltcod (approved):

saltcod:
tested & looks good

PR #30184, opened 2024-10-30
chore(perf): table editor query optimisation

alaister:

chore/performance

The table editor sends 5 queries to load a table/view. The previous query flow:
- Fetch entity type (table/view/materialized view etc)
  - Fetch foreign key constraints
  - Fetch encrypted columns
  - Fetch table info based on entity type
    - Fetch table data based on table info

(indentation represents the waterfall)

## What is the new behavior?

It now only sends 2.

- Fetch table info
  - Fetch table data based on table info

^ the way it should be

## Additional changes

- Fixed some missing default fallback booleans on react-query `enabled` flag, causing queries to be sent when they shouldn't.
- Added `deletion_action` and `update_action` to the tables relationships, meaning we don't have to fetch all the column key constraints to check these.


apps/studio/components/grid/components/formatter/ForeignKeyFormatter.tsx
```diff
@@ -23,11 +23,13 @@ export const ForeignKeyFormatter = (props: Props) => {
   const { projectRef, tableId, row, column } = props
   const id = tableId ? Number(tableId) : undefined
 
-  const { data: selectedTable } = useTableQuery({
+  const { data } = useTableEditorQuery({
     projectRef: project?.ref,
     connectionString: project?.connectionString,
     id,
   })
+  const selectedTable = data?.table_info
```
joshenlim, line 31: I'm just leaving annotations for myself to double check when I test your PR later so feel free to ignore (unless you have the answer at hand hahah)

do we not need `getTableLikeFromTableEditor` like here? need to check what's the diff between `getTableLikeFromTableEditor(data)` and `data.table_info`

apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
```diff
@@ -96,13 +100,18 @@ const TableEditorMenu = () => {
...
+    id,
+  })
 
   useEffect(() => {
+    const selectedTable = getTableLikeFromTableEditor(tableData)
     if (selectedTable) {
       setSelectedSchema(selectedTable.schema)
     }
-  }, [selectedTable])
+  }, [tableData])
```
joshenlim, line 114: Minor nit: am thinking the dependency array here should be`tableData.id`?

Given that this useEffect is just to set the schema, `tableData` being JSON - this will always trigger this useEffect even if the schema didn't change (e.g if i add a new column - verified locally) Although the consequence isn't anything big

apps/studio/data/tables/get-table.ts
joshenlim, line 1: Another one that i'll check later - but can we actually deprecate `getTable` then? I see that it's being used in EditHookPanel, SidePanelEditor utils, EntityListItem -> but reckon those could use the new table-editor-query? (i might be wrong / overlooking something at this time)

apps/studio/components/grid/components/formatter/ForeignKeyFormatter.tsx
alaister: Yeah I'm actually not sure of the answer here either... I did it this way because before it was only fetching the table and not handling all the other types...

joshenlim:
doing some testing locally and i noticed this: my schema has 1 foreign table and 2 normal tables
saw that `foreign_table_info` returns a different table - it should be null right?

image

alaister:
> doing some testing locally and i noticed this: my schema has 1 foreign table and 2 normal tables am console.log-ing out `tableData` from `TableEditorMenu` saw that `foreign_table_info` returns a different table - it should be null right? although its a nit i suppose since `getTableLikeFromTableEditor` would retrieve the correct information

wow that is very odd considering I'm filtering by id... it definitely shouldn't be returning both there

joshenlim:
ahh found it! we're missing the ID check
image

alaister:
hahaha I spotted also – pushed up a fix 😅

joshenlim:
just to double check as well - are partitioned tables covered by this?

alaister:
partitioned tables should be covered by the regular tables query 👍

joshenlim:
side note - i think we can remove encrypted columns now
cause it seems like we're deprecating pg sodium and TCE altogether
https://supabase.com/docs/guides/database/extensions/pgsodium#encryption-key-location

joshenlim (approved):
LGTM! Tested the following:
- Tables, Partitioned Tables, Foreign tables, Views, Materialized Views render correctly in the Table Editor
  - Can be filtered / sorted
- Column foreign key deletion and update action render correctly

There shouldn't be any visual changes
- Added a small fix whereby update action for a column with foreign key wasn't showing in the tooltip
- Screenshot 2024-10-31 at 12 16 46
  - Separately, reckon this should be a hover card instead, but can look into it separately
- Also can deprecate encrypted_columns separately too

PR #30188, opened 2024-10-30
Add spam validation check in auth email templates

joshenlim:
Related API PR: https://github.com/supabase/infrastructure/pull/20234

Adds a spam validation check in the auth email templates page. We're using SpamAssassin for the check
- Show risk of spam if > 1 spam rule with score > 0 is returned
- Prevent saving if > 1 spam rule with score > 0 is returned while using built in email service

Validation check happens on 2 events
- Debounce after 1s when editing content
- When going between templates

Example spam detected:
image

Example spam not detected:
image


alaister (approved):
LGTM!

- Tested on preview

PR #30189, opened 2024-10-30
Chore/replace boolean input with select

joshenlim:
As per PR title - shouldn't have any visual change

To test - create a boolean column that's not nullable and no default value. Then head to add a row

Before:
image

After:
image


ivasilov:
The select doesn't have a default value selected on columns with default values.
Screenshot 2024-10-30 at 11 51 57
 Here's the table definition:
 ```
 create table
  public."testTable" (
    id bigint generated by default as identity not null,
    created_at timestamp with time zone not null default now(),
    default_true boolean not null default true,
    default_false boolean not null default false,
    default_true_nullable boolean null default true,
    default_false_nullable boolean null default false,
    default_null_nullable boolean null,
    constraint test - boolean - table_pkey primary key (id)
  ) tablespace pg_default;
  ```

ivasilov:
There's still the issue with "select a value" when editing a row with null values, but that's not a big deal.
Screenshot 2024-10-31 at 10 35 30

ivasilov (approved):

PR #30198, opened 2024-10-30
Chore/self hosted assistant disabled

saltcod:
in self-hosted, if you don't have a OPENAI_API_KEY set, the assistant will look like it's working, but will just silently fail.

added an api route to check for the existence of the key, and an alert to notify if not present

screenshot-2024-10-30-at-16 40 00

### DIFF
--- a/apps/studio/components/interfaces/SQLEditor/AiAssistantPanel/index.tsx
+++ b/apps/studio/components/interfaces/SQLEditor/AiAssistantPanel/index.tsx
@@ -325,7 +325,7 @@ export const AiAssistantPanel = ({
         <AssistantChatForm
           textAreaRef={inputRef}
           loading={isLoading}
-          disabled={isLoading || isApiKeySet}
+          disabled={isLoading || !isApiKeySet}
           icon={
             <AiIconAnimation
               allowHoverEffect
--- a/apps/studio/components/interfaces/SQLEditor/AiAssistantPanel/index.tsx
+++ b/apps/studio/components/interfaces/SQLEditor/AiAssistantPanel/index.tsx
@@ -4,11 +4,13 @@ import { Info } from 'lucide-react'
 import { useEffect, useMemo, useRef, useState } from 'react'
 import { toast } from 'sonner'
 
+import { Alert, AlertDescription, AlertTitle } from '@ui/components/shadcn/ui/alert'
 import { Message as MessageType } from 'ai'
 import { useChat } from 'ai/react'
 import { useParams } from 'common'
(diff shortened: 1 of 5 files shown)


apps/studio/components/interfaces/SQLEditor/AiAssistantPanel/index.tsx
```diff
@@ -112,6 +115,25 @@ export const AiAssistantPanel = ({
...
+        const response = await fetch(`${BASE_PATH}/api/ai/sql/check-api-key`)
+        setIsApiKeySet(response.status === 200)
+      } catch (error) {
+        console.error('Error checking API key:', error)
+        setIsApiKeySet(false)
+      }
+    }
+
+    checkApiKey()
+  }, [])
```
joshenlim, line 137: this could probably be a react query - will be cleaner i think

apps/studio/pages/api/ai/sql/check-api-key.ts
```diff
@@ -0,0 +1,9 @@
+import { NextApiResponse } from 'next'
+
+export default function handler(res: NextApiResponse) {
+  if (process.env.OPENAI_API_KEY) {
+    res.status(200).send('OK')
+  } else {
+    res.status(401).send('API key not found')
+  }
+}
```
joshenlim, line 9: this doesn't seem to work (running into 500) and is also not consistent with the way we write our next js endpoints

apps/studio/components/interfaces/SQLEditor/AiAssistantPanel/index.tsx
joshenlim: also thinking - probably doesnt need to be self hosted specific? this check should also work for hosted too, probably a good safety net for us actually haha

joshenlim:
Pushed up the fixes for the issues identified
Also added the same check to the global assistant UI

apps/studio/components/interfaces/SQLEditor/AiAssistantPanel/index.tsx
joshenlim: ^ actually sorry think if we do this API call on the hosted dash it'll probably incur unnecessary functions cost

joshenlim (approved):
LGTM! Tested locally for if API key is not set, and on staging preview for hosted
@saltcod will let you merge whenever ready - probably good to double check the changes i made here before merging too 🙏

saltcod:
amazing, thanks @joshenlim! Checked staging and local, merging.

PR #30206, opened 2024-10-31
Update securing-your-api.mdx to fix  broken example

GaryAustin1:

YES

Doc Fix.

The example for checking additional API keys has a security flaw and bad code.

It checked current_role for anon to do security, but because it is a security definer function the role will never be anon.

Added to check for the role claim in the jwt.

Also the table used for keys is UUID and the type from the header is text for the key.  Cast it to UUID.

[Please link any relevant issues here.](https://supabase.com/docs/guides/api/securing-your-api?queryGroups=pre-request&pre-request=use-additional-api-key#examples)

## What is the new behavior?

See Above comments

## Additional context

image


charislam (approved):
Thanks @GaryAustin1!

PR #30211, opened 2024-10-31
feat: Redirect to create org when the user has no orgs to authorize

ivasilov:
This PR adds changes the oAuth authorization to send the user to create an org if the user doesn't belong to an organization.


saltcod (approved):

saltcod:
Tested flow on local with new account, works as expected

PR #30215, opened 2024-10-31
docs(auth): update password reset function name in PKCE flow

LukasKri:

YES

docs update

I was setting up password rest functionality and the wrong name of this function caught me off guard. Spent 5-10 minutes googling until I understand that I don't have to update `signUpNewUser` function, which is mentioned at the top of the page, by including auth `auth.resetPasswordForEmail()` functionality, but rather create a new separate function for this purpose.

## What is the new behavior?

Less confusion for the reader.


charislam (approved):
Thanks @LukasKri!

PR #30258, opened 2024-11-04
Fix https://supabase.sentry.io/issues/6037628167/

joshenlim:
- Reproduced by clicking "Add new row" on a table that has 0 columns
- If a table has 0 columns, postgrest returns the `columns` property as null instead of an empty array
- This caused generateRowFields to error out, also due to the unnecessary `!` parameter
- Hide Add new row action if table has 0 columns

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnForeignKey.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnForeignKey.tsx
@@ -116,13 +116,20 @@ const ColumnForeignKey = ({
           table={{
             id: table.id,
             name: table.name,
-            columns:
-              column.isNewColumn && column.name
-                ? (table.columns as any[]).concat(column)
-                : (table.columns as any[]).map((c) => {
-                    if (c.id === column.id) return { ...c, name: column.name }
-                    else return c
-                  }),
+            columns: (column.isNewColumn && column.name
+              ? (table.columns ?? []).concat(column as any)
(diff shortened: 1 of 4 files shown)


MildTomato:
confirmed fixed on preview

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnForeignKey.tsx
```diff
@@ -116,13 +116,20 @@ const ColumnForeignKey = ({
...
             name: table.name,
-            columns:
-              column.isNewColumn && column.name
-                ? (table.columns as any[]).concat(column)
-                : (table.columns as any[]).map((c) => {
-                    if (c.id === column.id) return { ...c, name: column.name }
-                    else return c
-                  }),
+            columns: (column.isNewColumn && column.name
+              ? (table.columns ?? []).concat(column as any)
```
MildTomato, line 120: `any` ?

MildTomato: it was `any` before so not a blocker.

joshenlim: hmm actually lemme just clean this up too

joshenlim: okay pushed a refactor up!

alaister (approved):
LGTM!

- Tested on preview

PR #30283, opened 2024-11-04
Chore/move advisors

saltcod:
- moves Query Performance to Advisors
- renames Security → Security Advisor, Performance → Performance Advisor to be less werid with "Query Performance"
- makes the schema visualizer the top level item under /database/


joshenlim (approved):
LGTM! Verified changes on the preview 🙂

PR #30294, opened 2024-11-05
Add resize info in postgres upgrade modal

joshenlim:
To inform users that their disk will be resized back to the original value when they upgrade their postgres version

image


alaister (approved):
LGTM!

- Tested on preview by upgrading a project with an enlarged disk

PR #30310, opened 2024-11-06
Chore/allow export from non editable tables

saltcod:
Should still be allowed to export from tables in protected schemas

screenshot-2024-11-05-at-20 56 59


joshenlim (approved):
LGTM! Tested on preview with the auth schema

PR #30315, opened 2024-11-06
chore: remove tableEditorPrefetching feature flag

alaister:


joshenlim (approved):
LGTM - checked through logs that there doesn't seem to be any significant trends in error rates caused by this change (which is now toggled for all users), should be safe to remove but we'll keep an eye

PR #30323, opened 2024-11-06
fix: table editor fallback to undefined

alaister:

Bug fix

If you try and query something on the table editor that isn't a table like object (for example an index), the query returns `null` and errors out the dashboard.

## What is the new behavior?

Fallback to `undefined` in this case so the 404 screen is correctly shown


joshenlim (approved):
LGTM! Verified on staging preview entering the table editor with an index's ID

Separately though (not urgent), just wondering how the user ran into this. Hypothesis is a bad URL somewhere in the dashboard but let's observe first

PR #30345, opened 2024-11-07
Chore/cron nl

saltcod:
wip


packages/ai-commands/src/sql/cron.ts
```diff
@@ -0,0 +1,97 @@
...
+    name: 'generateCronSyntax',
+    description: stripIndent`
+      Generates a cron expression for a given natural language description.
+    `,
+    parameters: generateCronSyntaxSchema.schema as Record<string, unknown>,
+  }
+
+  try {
+    const response = await openai.chat.completions.create({
+      model: 'gpt-4o-2024-05-13',
```
gregnr, line 61: This is probably something that gpt-4o-mini could be good at. If no real quality difference, this will save money

joshenlim:
this is totally separate from what this PR is, but wondering if its small enough to piggy back haha
reckon our loading state should show the header and description, like webhooks here for e.g
image

right now the whole section is a skeleton loader

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
```diff
@@ -1,107 +1,221 @@
...
+                      {isGeneratingCron ? <CronSyntaxLoader /> : presetValue || '* * * * * *'}
+                    </span>
+                  ) : (
+                    <span className="text-xl font-mono text-foreground-lighter">
+                      {isGeneratingCron ? <CronSyntaxLoader /> : presetValue || '* * * * * *'}
+                    </span>
+                  )}
+                  {scheduleString ? (
+                    <span className="text-foreground-light flex items-center gap-2">
+                      The cron will be run {/* lowercase the first letter */}
```
joshenlim, line 188: this is an extension - but im thinking it might also be helpful to show when the job will run next - wdyt?

e.g if we selected every 5 minutes, and its currently 11.30pm, we can show "Job will run next at 11.35pm"

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
```diff
@@ -1,107 +1,221 @@
...
+                        setInputValue('')
+                        setPresetValue('')
+                        setScheduleString('')
+                      }}
+                    />
+                    <p className="text-sm text-foreground-light">Use natural language</p>
+                  </div>
+
+                  <div className="mt-4">
+                    <p className="text-sm text-foreground-light">Use a preset</p>
```
joshenlim, line 136: probably don't need this label - i think the buttons communicate clear enough that they're presets

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
joshenlim: e.g for the preset "Every hour, at 30 minutes" not sure if its just me but im still not sure how the job will run
clicking it shows the description "The cron will be run at 30 minutes past the hour." under the schedule which makes me wonder - "it's now 11.43pm, hmm does that mean it's gonna run at 12.30am?"

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
```diff
@@ -1,107 +1,221 @@
...
-                    .
-                  </span>
-                ) : (
-                  !fieldState.error && <span className="h-5" />
-                )}
+
+              <FormMessage_Shadcn_ className="mt-4">
+                <div className="mt-2 bg-surface-100 p-4 rounded grid gap-4 border">
+                  <h4 className="text-sm text-foreground">Schedule</h4>
+                  {scheduleString ? (
```
joshenlim, line 170: this feels like repeated information with the input field imo
image

joshenlim: im wondering if maybe the description here is all we need "the cron will be run at ___"

joshenlim: hmm a bit on the fence for either way actually - will relook at this again

but i reckon RE the syntax chart, we could potentially line it up with the cron syntax in the schedule box - will be even clearly perhaps (and also one less repetition of * * * * * *
image

apps/studio/data/ai/cron-syntax-mutation.ts
```diff
@@ -0,0 +1,52 @@
...
+import { isResponseOk, post } from 'lib/common/fetch'
+import { BASE_PATH } from 'lib/constants'
+import type { ResponseError } from 'types'
+
+export type CronSyntaxGenerateVariables = {
+  prompt: string
+}
+
+export async function generateCronSyntax({ prompt }: CronSyntaxGenerateVariables) {
+  const response = await post<string>(BASE_PATH + '/api/ai/sql/cron', {
```
joshenlim, line 13: just remember we need to add this to `middleware.ts`

and i reckon we should also try to figure out how to just cover all API routes under /api/ai/sql in there too - guessing its likely all endpoints under there will be used by hosted too

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
saltcod: Yeah it's meant to be a kind of "here's the summary of what's happening" 
it's redundant here in this view, but not when in natural language view. 

https://share.cleanshot.com/hJlwj6CX

apps/studio/data/ai/cron-syntax-mutation.ts
saltcod: sorry, i've got this in here — is there something missing?

https://github.com/supabase/supabase/pull/30345/files#diff-3aeded5f052bc3f0f64fa6ba9db7afacecc1b07ad9df802c82798a31679baede

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
saltcod: (that's the current reasoning, always open to suggestion tho!)

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
saltcod: Yes I think that's when it's set to run. Agreed on this one. I'll replace with an example that's less ambiguous, since docs / ai will prob do the heaviest lifting for people here.

packages/ai-commands/src/sql/cron.ts
```diff
@@ -0,0 +1,96 @@
+import type OpenAI from 'openai'
+import { ContextLengthError } from '../errors'
+import { jsonrepair } from 'jsonrepair'
+
+// Responds to a natural language request for cron syntax.
+export async function chatCron(openai: OpenAI, prompt: string) {
```
gregnr, line 6: Might be worth renaming this to something like `generateCron()`. Currently we use the `chat___` convention for conversation-style AI that are multi-turn.

packages/ai-commands/src/sql/cron.ts
```diff
@@ -0,0 +1,96 @@
...
+  const initMessages: OpenAI.Chat.Completions.ChatCompletionMessageParam[] = [
+    {
+      role: 'system',
+      content: `
```
gregnr, line 10: Maybe we prefix with `codeBlock` template tag from `common-tags` which will strip indents (so that we're not passing indented prompts to OpenAI).

apps/studio/data/ai/cron-syntax-mutation.ts
joshenlim: ahh yes that will do!

joshenlim (approved):
LGTM! tested on preview and locally the new natural language prompt for cron :)

PR #30350, opened 2024-11-07
chore: remove kpsVersion property

kevcodez:
KPS has long been removed and the property is never set for projects

Angelico also gave the go to clean it up.


alaister (approved):
LGTM!

- Tested on preview

PR #30352, opened 2024-11-07
chore: use service versions endpoint

alaister:

chore

Service versions are shown from the project detail endpoint.

## What is the new behavior?

These fields are now moved to a dedicated endpoint.

## Additional context

We're planning to remove the service versions from the project detail endpoint as it slows it down a lot.

### DIFF
--- a/apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureInfo.tsx
+++ b/apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureInfo.tsx
@@ -127,7 +127,7 @@ const InfrastructureInfo = () => {
                         <Input
                           readOnly
                           disabled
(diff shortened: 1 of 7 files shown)


joshenlim (approved):
LGTM! Tested on preview on different types of projects
- Project that can be upgraded but has read replicas
- Project that can be upgraded
- Project on Fly
- Project that's already on latest version

Minor issue found that's probably only affecting internal though - if we use a custom postgres version, we don't seem to render the version properly. Probably can just render the whole version string instead of slicing if the version string doesn't have `supabase-postgres` ([ref](https://github.com/supabase/supabase/blob/chore/use-service-versions-endpoint/apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureInfo.tsx#L58))
image
image

PR #30357, opened 2024-11-07
Hook up RLS to new assistant if feature flag is enabled

joshenlim:
As per PR title


packages/ai-commands/src/sql/chat.ts
```diff
@@ -87,6 +89,10 @@ export async function chatSql(
     initMessages.push(...messages)
   }
 
+  console.log('\n----- START: NEW MESSAGE -----')
```
jordienr, line 92: 👮‍♂️🚨🚔 get him

jordienr (approved):
Code wise LGTM
left some UX feedback in slack but can be done in a different PR if you prefer

packages/ai-commands/src/sql/chat.ts
joshenlim: 😂  caught red handed in 4K

PR #30358, opened 2024-11-07
Chore/disk mgt UI fixes

joshenlim:
Add warning if at 90% + within cooldown period
Screenshot 2024-11-07 at 18 11 13

Add warning if in read only mode
Screenshot 2024-11-07 at 18 35 12

Also updated bar to be red if the utilization is >= 90% (since DB will enter read only mode at 95%)

### DIFF
--- a/apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
+++ b/apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
@@ -313,7 +313,7 @@ export function DiskManagementForm() {
             <>
               <div className="flex flex-col gap-y-3">
                 <DiskCountdownRadial />
-                {usedPercentage >= 90 && isWithinCooldownWindow && (
+                {!isReadOnlyMode && usedPercentage >= 90 && isWithinCooldownWindow && (
                   <Admonition
                     type="destructive"
                     title="Database size is currently over 90% of disk size"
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
```diff
@@ -310,6 +311,35 @@ export function DiskManagementForm() {
           />
           {!isFlyArchitecture && (
             <>
+              <div className="flex flex-col gap-y-3">
+                <DiskCountdownRadial />
+                {usedPercentage >= 90 && isWithinCooldownWindow && (
+                  <Admonition
+                    type="destructive"
+                    title="Database is currently at 90% of disk size"
```
jordienr, line 319: should be `is currently over 90%` right?

apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
```diff
@@ -310,6 +311,35 @@ export function DiskManagementForm() {
           />
           {!isFlyArchitecture && (
             <>
+              <div className="flex flex-col gap-y-3">
+                <DiskCountdownRadial />
+                {usedPercentage >= 90 && isWithinCooldownWindow && (
```
jordienr, line 316: should it have `&& !isReadOnlyMode` too? Or is that checked in `isWithinCooldownWindow`

jordienr (approved):

apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
joshenlim: oh yeap! good point ahha

apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
joshenlim: ahhh yes great catch thank you so much! 🙏 will need to add this check or 2 warnings will show i think

PR #30372, opened 2024-11-08
Deprecate old org roles query

joshenlim:


alaister (approved):
LGTM!

- Tested on preview

PR #30377, opened 2024-11-08
Add check for project active healthy status in some RQs that are unnecessary when project is paused

joshenlim:
Mainly the ones which are dependent on the database being online
- /run-lints
- any pg-meta ones, i've only added in for extensions since thats the one thats getting triggered while on a paused state, but technically any pg-meta RQs should have that check
- any that are executing SQL on the DB


alaister (approved):
LGTM!

- Tested all queries on preview

PR #30379, opened 2024-11-08
Chore/update create and update user to use platform endpoints

joshenlim:
As per PR title, tested locally creating user, banning / unbanning user


alaister (approved):
LGTM!

- Tested on preview

PR #30398, opened 2024-11-10
chore: use in-browser sha256

kevcodez:
Browsers should be able to do sha256 just fine with the crypto.subtle.digest API - no need for a dependency.

Screenshot 2024-11-10 at 16 40 36

Verified locally the the base64 encoded user id is the same as before.


alaister (approved):
LGTM!

- tested on preview by verifying `supabase-sentry-user-id` was set in local storage

PR #30430, opened 2024-11-12
fix: use nulls last when sorting users

alaister:

Bug fix

When sorting on last sign in at descending, you have to scroll past a bunch of users that haven't ever signed in before you can see your most recent sign in

## What is the new behavior?

Moves these users to the end

## Additional context
Before:
Screenshot 2024-11-12 at 16 13 01


jordienr (approved):

joshenlim (approved):
LGTM! Verified that null values are shifted to the bottom when sorted

PR #30433, opened 2024-11-12
fix: allow downgrade to Free Plan with no active projects

kevcodez:
The member exceeding Free Plan limits modal should not be displayed if the org to be downgraded has no active projects.

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
@@ -102,7 +102,9 @@ const PlanUpdateSidePanel = () => {
   } = useOrganizationBillingSubscriptionPreview({ tier: selectedTier, organizationSlug: slug })
 
   const availablePlans: OrgPlan[] = plans?.plans ?? []
-  const hasMembersExceedingFreeTierLimit = (membersExceededLimit || []).length > 0
+  const hasMembersExceedingFreeTierLimit =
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
```diff
@@ -102,7 +102,9 @@ const PlanUpdateSidePanel = () => {
   } = useOrganizationBillingSubscriptionPreview({ tier: selectedTier, organizationSlug: slug })
 
   const availablePlans: OrgPlan[] = plans?.plans ?? []
-  const hasMembersExceedingFreeTierLimit = (membersExceededLimit || []).length > 0
+  const hasMembersExceedingFreeTierLimit =
+    (membersExceededLimit || []).length > 0 &&
+    orgProjects.filter((it) => it.status !== 'INACTIVE' && it.status !== 'GOING_DOWN').length > 0
```
ecktoteckto, line 107: Theoretically we don't have to check for `GOING_DOWN` since the Mgmt-API endpoint we use to retrieve projects doesn't include projects in status `GOING_DOWN`. If we don't want to rely on the endpoint's filtering, we would also have to include `REMOVED` here.

ecktoteckto (approved):

ecktoteckto:
Tested on staging

joshenlim (approved):
LGTM! Tested on staging with a pro plan that has 1 paused project, and a member who has reached the free project limit

PR #30437, opened 2024-11-12
chore: remove useExecuteSqlQuery (part 1)

alaister:

chore

## What is the new behavior?

- `useExecuteSqlQuery()` -> to regular `useQuery()`.
  - Real keys instead of piggybacking off sql keys. This will help with proper invalidation.
  - Better types as we no longer need to use `select()`.
  - Increased `staleTime` on these queries as `useExecuteSqlQuery()` defaults to `0`. This should help reduce some load on the API.
- Swapped all queries requiring `schema` and `table` to simply require `id`.
  - Mainly for consistency and simplicity and also easy of invalidation.


apps/studio/data/database/entity-definition-query.ts
```diff
@@ -79,32 +84,55 @@ export type EntityDefinitionVariables = {
...
-  return useExecuteSqlQuery(
+  const sql = getEntityDefinitionSql({ id, type })
+  const { result } = await executeSql(
     {
       projectRef,
       connectionString,
-      sql: getEntityDefinitionQuery({ id, type }),
-      queryKey: databaseKeys.entityDefinition(projectRef, id),
+      sql,
+      queryKey: ['entity-definition', id],
```
ivasilov, line 101: Should this be `databaseKeys.entityDefinition(projectRef, id)`?

alaister: In this case it's just a key for the URL query string – do you think that would benefit from the whole query key and not just this part?

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
```diff
@@ -740,12 +742,8 @@ export const updateTable = async ({
 
   await Promise.all([
     queryClient.invalidateQueries(sqlKeys.query(projectRef, ['foreign-key-constraints'])),
-    // invalidate list of tables, as well as visible individual tables
-    queryClient.invalidateQueries(['projects', projectRef, 'tables']),
```
joshenlim, line 744: just to double check - we're removing invalidating projects here?

alaister: I couldn't see this query key in use anymore

apps/studio/data/database/entity-definition-query.ts
joshenlim: yup this should be fine - executeSql doesnt cache the results in the RQ state (its not a getter since it just fires a POST request)

joshenlim:
Found an unrelated issue - we probably should remove the data/definition toggle in the table editor if the table is a foreign table (im guessing might apply for partition tables too?)

saltcod:
Tested: 
- table crud
- row crud (inline & expanded side panel)
- column crud 
- database/tables crud for both columns and tables

joshenlim (approved):
LGTM! Tested on the preview
- Create/update/delete table
- Create/update/delete column
- Create/update/delete row
- Add/Remove foreign key constraint

PR #30451, opened 2024-11-13
Update Login with WorkOS docs to include latest UI and best practices

zackproser-workos:

YES

Updates the docs and screenshots for Login with WorkOS to the latest best practices, including using [AuthKit](https://authkit.com).

The current docs reference out-of-date:
-  WorkOS dashboard screenshots 
- navigation instructions 
-  older versions of the WorkOS login experience

## What is the new behavior?

These changes: 
1. update all WorkOS dashboard screenshots to reference the latest UI
1. update instructions with two tested paths: A. Use Supabase Auth with WorkOS as a provider and B. use WorkOS AuthKit 

## How I verified these changes

I started with [an official Supbase example application](https://github.com/supabase/supabase/tree/master/examples/auth/nextjs#clone-and-run-locally) and then I ran the supabase docs project locally from my branch containing these changes. 

I sequentially applied the modifications specified by the instructions updated in this pull request.

I verified that the following workflows succeeded end-to-end: 
- login
- user session persistence
- logout


charislam:
Thanks @zackproser-workos! Really appreciate the PR and how detailed it is, especially for the OAuth section.

Just a quick question: I skimmed through the AuthKit docs for WorkOS, and I'm wondering if this connects to Supabase Auth in any way, or if it's just purely WorkOS Auth? If it's the latter, it might be a bit confusing to include it, since all our documented Auth methods assume that you will be authenticated to Supabase, and all the corresponding JWT authorization/database access stuff will work with that Auth method.

charislam:
Ahh thanks for pointing the way to the repo @zackproser-workos! After looking at the Supabase integration from the example app, I'm leaning towards leaving out the AuthKit bit because it will probably cause confusion with other sections of the docs (that assume you have the right cookies to call RLS-protected routes on Supabase, whereas doing it the AuthKit way will only give you either the anon-scoped data or, if you use the service role key on the server side, you'll have to handle the filtering yourself).

Happy to re-review and accept this PR with the OAuth section! Thanks again 🙏🏼

zackproser-workos:
> Ahh thanks for pointing the way to the repo @zackproser-workos! After looking at the Supabase integration from the example app, I'm leaning towards leaving out the AuthKit bit because it will probably cause confusion with other sections of the docs (that assume you have the right cookies to call RLS-protected routes on Supabase, whereas doing it the AuthKit way will only give you either the anon-scoped data or, if you use the service role key on the server side, you'll have to handle the filtering yourself).
> 
> Happy to re-review and accept this PR with the OAuth section! Thanks again 🙏🏼

@charislam - Understood - thanks for the feedback. 

I've removed the AuthKit information from the PR - would appreciate another look when you have the cycles.

Thanks so much.

charislam (approved):

charislam:
Thanks @zackproser-workos! Looks great, made a few minor edits and they should be good to publish!

zackproser-workos:
@charislam - thank you kindly for your swift assistance! 🙏

PR #30480, opened 2024-11-14
ci: linter v2

charislam:
PR to start introducing Linter V2!

Primary improvements from this linter:
- Better parsing of MDX (fewer false positives)
- More flexibility
- And suggested autofixes! (see auto-generated PR comment below for how that works)

Turning this on for only a small portion of the repo at first to avoid an avalanche of alerts, will slowly expand from there.

image


fsansalvadore (approved):
Nice, looks good!

PR #30494, opened 2024-11-15
feat: add project settings option to mark compliance projects

staaldraad:

YES

feature

## Additional context

Projects within organizations with the HIPAA add-on can be mark as "sensitive". These projects will have additional security advisor checks applied and other compliance related guardrails.

API enabled in related PR

Screenshot 2024-11-15 at 16 25 47


doublethink (approved):

apps/studio/components/interfaces/Settings/General/ComplianceConfig/ProjectComplianceMode.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+import { Switch, TooltipContent_Shadcn_, TooltipTrigger_Shadcn_, Tooltip_Shadcn_ } from 'ui'
+
+const ComplianceConfig = () => {
+  const { ref } = useParams()
+  const [isSensitive, setIsSensitive] = useState(false)
+  const selectedOrganization = useSelectedOrganization()
+  const { data: subscription } = useOrgSubscriptionQuery({ orgSlug: selectedOrganization?.slug })
+  const hasHipaaAddon = subscriptionHasHipaaAddon(subscription)
+  const { data: settings, isLoading, isSuccess } = useProjectSettingsV2Query({ projectRef: ref })
```
kevcodez, line 27:
```suggestion
  const { data: settings, isLoading, isSuccess } = useProjectSettingsV2Query({ projectRef: ref }, { enabled: hasHipaaAddon })
```

Would prevent the request from being triggered unless customer actually has the HIPAA add-on

kevcodez, line 27: just double checking - Is is correct that the compliance settings come out of the regular settings endpoint (GET), but the update goes against a separate sensitivity endpoint?

staaldraad: yes, the settings endpoint already returns information such as `ssl_enforced` so we added this field to it. Idiomatically it is probably more correct to have a GET sensitivity end-point, however saving on making another api call felt like a small wiin

apps/studio/components/interfaces/Settings/General/ComplianceConfig/ProjectComplianceMode.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const toggleIsSensitive = async () => {
+    if (!ref) return console.error('Project ref is required')
+    setIsSensitive(!isSensitive)
+    updateComplianceConfig({ projectRef: ref, isSensitive: !isSensitive })
+  }
+
+  // this is only setable on compliance orgs, currently that means HIPAA orgs
+  if (!hasHipaaAddon) {
+    return
```
joshenlim, line 66:
```suggestion
    return null
```

apps/studio/components/interfaces/Settings/General/ComplianceConfig/ProjectComplianceMode.tsx
```diff
@@ -0,0 +1,126 @@
...
+              className="lg:col-span-7"
+              description={
+                <div className="space-y-4">
+                  <p className="text-sm text-foreground-light">
+                    Apply additional compliance controls to this project
+                  </p>
+                </div>
+              }
+            >
+              Project processes sensitive data (HIPAA)
```
joshenlim, line 88: I don't quite understand what this label means - ideally the language should be actionable since its a switch

perhaps "Apply additional compliance controls to this project" should be the label, and the description could expand on what compliance controls are involved

apps/studio/components/interfaces/Settings/General/ComplianceConfig/ProjectComplianceMode.tsx
```diff
@@ -0,0 +1,126 @@
...
+  // this is only setable on compliance orgs, currently that means HIPAA orgs
+  if (!hasHipaaAddon) {
+    return
+  }
+
+  return (
+    <div id="compliance-configuration">
+      <div className="flex items-center justify-between mb-6">
+        <FormHeader className="mb-0" title="High Compliance Configuration" description="" />
+        <DocsButton href="https://supabase.com/docs/guides/deployment/shared-responsibility-model#managing-healthcare-data" />
```
joshenlim, line 73: does the docs need updating? i don't see any related content from this URL about this switch in the dashboard

apps/studio/components/interfaces/Settings/General/ComplianceConfig/ProjectComplianceMode.tsx
```diff
@@ -0,0 +1,126 @@
...
+        <FormSection
+          header={
+            <FormSectionLabel
+              className="lg:col-span-7"
+              description={
+                <div className="space-y-4">
+                  <p className="text-sm text-foreground-light">
+                    Apply additional compliance controls to this project
+                  </p>
+                </div>
```
joshenlim, line 85: this seems unnecessary - can just be a string

apps/studio/data/config/keys.ts
```diff
@@ -24,4 +24,6 @@ export const configKeys = {
     ['projects', organizationSlug, cloudProvider, dbRegion, 'available-creation-versions'] as const,
   projectUnpausePostgresVersions: (projectRef: string | undefined) =>
     ['projects', projectRef, 'available-unpause-versions'] as const,
+  projectComplianceConfig: (projectRef: string | undefined) =>
```
joshenlim, line 27: there's no GET query for this config right? so this key is unnecessary

apps/studio/data/config/project-compliance-config-mutation.ts
```diff
@@ -0,0 +1,65 @@
...
+  'mutationFn'
+> = {}) => {
+  const queryClient = useQueryClient()
+
+  return useMutation<ComplianceConfigUpdateData, ResponseError, ComplianceConfigUpdateVariables>(
+    (vars) => updateComplianceConfig(vars),
+    {
+      async onSuccess(data, variables, context) {
+        const { projectRef } = variables
+        await queryClient.invalidateQueries(configKeys.projectComplianceConfig(projectRef))
```
joshenlim, line 52: noticed that the is_sensitive data is from the project settings V2 query, the invalidation should invalidate that query instead then so that the UI is updated

apps/studio/components/interfaces/Settings/General/ComplianceConfig/ProjectComplianceMode.tsx
```diff
@@ -0,0 +1,126 @@
...
+            >
+              Project processes sensitive data (HIPAA)
+            </FormSectionLabel>
+          }
+        >
+          <FormSectionContent loading={false} className="lg:!col-span-5">
+            <div className="flex items-center justify-end mt-2.5 space-x-2">
+              {(isLoading || isSubmitting) && (
+                <Loader2 className="animate-spin" strokeWidth={1.5} size={16} />
+              )}
```
joshenlim, line 96: FE standardization - we need an error handler here to handle the error state from ProjectSettingsV2 query
we typically just render the AlertError component if isError

apps/studio/data/config/project-compliance-config-mutation.ts
```diff
@@ -0,0 +1,65 @@
...
+import type { ResponseError } from 'types'
+import { configKeys } from './keys'
+import { handleError, patch } from 'data/fetchers'
+
+export type ComplianceConfigUpdateVariables = {
+  projectRef: string
+  isSensitive: boolean
+}
+
+export type ComplianceConfigUpdateResponse = {
```
joshenlim, line 13: is this being used? it should be coming from the API type generation either way

joshenlim (approved):
Changes LGTM! Checked on staging preview on an org with the hipaa addon

PR #30521, opened 2024-11-18
feat: display trigger condition

alaister:

feature

Trigger condition is not displayed.

## What is the new behavior?

Screenshot 2024-11-18 at 18 03 22

### DIFF
--- a/apps/studio/components/interfaces/Database/Triggers/CreateTrigger.tsx
+++ b/apps/studio/components/interfaces/Database/Triggers/CreateTrigger.tsx
@@ -8,6 +8,7 @@ import { Badge, Button, Checkbox, Input, Listbox, SidePanel } from 'ui'
 import ConfirmationModal from 'ui-patterns/Dialogs/ConfirmationModal'
 
 import { useProjectContext } from 'components/layouts/ProjectLayout/ProjectContext'
+import CodeEditor from 'components/ui/CodeEditor/CodeEditor'
 import FormEmptyBox from 'components/ui/FormBoxEmpty'
 import NoTableState from 'components/ui/States/NoTableState'
 import {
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Database/Triggers/CreateTrigger.tsx
```diff
@@ -492,6 +497,33 @@ const SelectEnabledMode = observer(({}) => {
...
 })
 
+const DisplayCondition = observer(() => {
+  const _localState = useContext(CreateTriggerContext)
+
+  return (
+    <div className="text-sm leading-4 grid gap-2 md:grid md:grid-cols-12">
+      <div className="flex flex-col space-y-2 col-span-4">
+        <label className="block text-foreground-light text-sm leading-4 break-all">
+          Condition (read-only)
```
joshenlim, line 507: probably can remove the read only label here since we already mention it in the description on how to update the condition

alaister: yeah I'm torn here – I actually had it without this label initially but then I can just see a user getting frustrated while trying to edit the field because they didn't read the description – it is easy to miss I think. What do you think?

joshenlim: hmm im thinking theres 3 ways
- Add a separator between condition and the other fields, make it contextual clear that the top section is editable, bottom is not
- New line on the "To update..." part and make the text "text-foreground" to stand out
- Instead of the text (read-only), we use the Badge component

joshenlim: maybe #2 might be best - and then we see if we get any feedback

alaister: 2 done! 👍

joshenlim (approved):

PR #30524, opened 2024-11-18
Add self-hosting password change guide and update docker image versions

### LINKED ISSUE #22605: Also Add Password Changing Guidelines
# Improve documentation

## Link 

https://supabase.com/docs/guides/self-hosting/docker

## Describe the problem

When password is changes like for postgress in .env , It causes an error when started again (`docker compose up -d`) , there is no mention of fixing that or aany passwd changing guidelines, After many hours i found out  you need to `docker compose down -v`

# Remove Postgres data:
`rm -rf volumes/db/data/`

## Describe the improvement

After many hours i found out  you need to `docker compose down -v`

# Remove Postgres data:
`rm -rf volumes/db/data/`

encima:

YES

Docs update and docker image versions update

Info for changing postgres password is incomplete and changing it in the env file does not update it

## What is the new behavior?

- Docker image versions have been updated to be in line with the CLI
- Add a section for password updating to the self hosting docs

## Additional context

Closes #22605

### DIFF
--- a/apps/docs/content/guides/self-hosting/docker.mdx
+++ b/apps/docs/content/guides/self-hosting/docker.mdx
@@ -195,7 +195,7 @@ If you have already started the deployment, you can change the password by alter
 
 For example:
 ```bash
-$ docker exec -it supabase-db psql -U supabase_admin -W
+docker exec -it supabase-db psql -U supabase_admin -W
 # Enter the original password from the .env file
 ALTER USER supabase_admin WITH PASSWORD 'new-password';
 ALTER USER postgres WITH PASSWORD 'new-password';
--- a/apps/docs/content/guides/self-hosting/docker.mdx
+++ b/apps/docs/content/guides/self-hosting/docker.mdx
@@ -187,11 +187,37 @@ Run this form twice to generate new `anon` and `service` API keys. Replace the v
 
 You will need to [restart](#restarting-all-services) the services for the changes to take effect.
 
+### Change database password
+
+If you have not yet started the deployment, you can change the password in the `./docker/.env` file before running `docker compose up -d`.
+
+If you have already started the deployment, you can change the password by altering the `postgres` user in the database with the new password and then updating the `POSTGRES_PASSWORD` in `./docker/.env`
+
+For example:
+```bash
+$ docker exec -it supabase-db psql -U supabase_admin -W
+# Enter the original password from the .env file
+ALTER USER supabase_admin WITH PASSWORD 'new-password';
+ALTER USER postgres WITH PASSWORD 'new-password';
+ALTER USER authenticator WITH PASSWORD 'new-password';
+ALTER USER pgbouncer WITH PASSWORD 'new-password';
+ALTER USER supabase_auth_admin WITH PASSWORD 'new-password';
+ALTER USER supabase_functions_admin WITH PASSWORD 'new-password';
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yml
```diff
@@ -411,21 +391,14 @@ services:
       - ./volumes/db/pooler.sql:/docker-entrypoint-initdb.d/migrations/99-pooler.sql:Z
       # Use named volume to persist pgsodium decryption key between restarts
       - db-config:/etc/postgresql-custom
+    ports:
+      - '${POSTGRES_PORT}:${POSTGRES_PORT}'
```
avallete, line 395: **question**

Since we have the supavisor container handling the pooling bellow, is it necessary to expose the postgres container port directly ?

encima: ah true, probably not. I would prefer to expose both with PG in the default port but I don't have a strong opinion. 
Happy to remove if we only want one?

avallete (approved):

12138zhen:
It just so happens that I'm confused about this

apps/docs/content/guides/self-hosting/docker.mdx
```diff
@@ -187,11 +187,37 @@ Run this form twice to generate new `anon` and `service` API keys. Replace the v
...
 
+### Change database password
+
+If you have not yet started the deployment, you can change the password in the `./docker/.env` file before running `docker compose up -d`.
+
+If you have already started the deployment, you can change the password by altering the `postgres` user in the database with the new password and then updating the `POSTGRES_PASSWORD` in `./docker/.env`
+
+For example:
+```bash
+$ docker exec -it supabase-db psql -U supabase_admin -W
```
charislam, line 198:
```suggestion
docker exec -it supabase-db psql -U supabase_admin -W
```

apps/docs/content/guides/self-hosting/docker.mdx
```diff
@@ -187,11 +187,37 @@ Run this form twice to generate new `anon` and `service` API keys. Replace the v
...
+ALTER USER supabase_functions_admin WITH PASSWORD 'new-password';
+ALTER USER supabase_storage_admin WITH PASSWORD 'new-password';
+exit
+```
+
+Then update the `POSTGRES_PASSWORD` in the `./docker/.env` file with the new password and run
+
+```bash
+$ docker compose down
+$ docker compose up -d
```
charislam, line 214:
```suggestion
docker compose down
docker compose up -d
```

charislam (approved):
minor nit: leaving out the prompt when writing terminal commands makes them easier to copy

encima:
> minor nit: leaving out the prompt when writing terminal commands makes them easier to copy

that makes sense! how do you usually distinguish between a terminal session and a psql one? or should they be separate blocks then?

charislam:
> > minor nit: leaving out the prompt when writing terminal commands makes them easier to copy
> 
> that makes sense! how do you usually distinguish between a terminal session and a psql one? or should they be separate blocks then?

Separate blocks will work well I think

tvogel:
I was looking for this information today. Any reasons not to go forward with this PR?

tvogel:
The spurious docker-compose.yml changes probably would need to be removed.

encima:
We can dismiss these changes and aim to only include the work in the docs [here](https://github.com/supabase/supabase/pull/38941)

PR #30527, opened 2024-11-18
Chore/cron ai nudges

saltcod:
- fix syntax issues accounting for seconds
- input now allows for normal cron syntax (* * * * *) and seconds in this format:  "10 seconds" 
- prevent submit on natural language keydown 
- set every 5 minutes as the initial default for readability

### DIFF
--- a/apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
+++ b/apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
@@ -32,11 +32,11 @@ interface CronJobScheduleSectionProps {
 }
 
 const presets = [
-  { name: 'Every minute', expression: '*/1 * * * *' },
+  { name: 'Every minute', expression: '* * * * *' },
   { name: 'Every 5 minutes', expression: '*/5 * * * *' },
   { name: 'Every first of the month, at 00:00', expression: '0 0 1 * *' },
-  { name: 'Every Monday at midnight', expression: '0 0 * * 1' },
-  { name: 'Every night at midnight', expression: '0 0 0 * * *' },
+  { name: 'Every night at midnight', expression: '0 0 * * *' },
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
```diff
@@ -113,13 +113,18 @@ export const CronJobScheduleSection = ({ form }: CronJobScheduleSectionProps) =>
...
+                        }
+                      }}
                       onChange={(e) => setInputValue(e.target.value)}
                     />
                   ) : (
                     <Input_Shadcn_
                       {...field}
                       autoComplete="off"
-                      placeholder="* * * * *"
+                      placeholder="* * * * * *"
```
joshenlim, line 127: slightly confused here - when should it have 5 and 6 placeholders? from the syntax here it looks like just 5
image

joshenlim:
Managed to repro a client crash by changing the last asterisk to 1
image
image

joshenlim:
unrelated feedback - thinking that adding (GMT) here isn't that useful as timezones are only applicable when the cron is set to run at a specific time. should we add it to the description instead whenever there's a specific time involved? (and also guessing this is +0000?)
image

saltcod:
re: the timezone — this is just exposing a system setting 
https://github.com/citusdata/pg_cron?tab=readme-ov-file#setting-up-pg_cron

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
saltcod: resolving this comment — this is basically the heart of the PR

joshenlim (approved):
LGTM! Tested on staging - can verify reported issues no longer happen, and am able to create a job with the new 30 seconds template

PR #30546, opened 2024-11-19
fix: Cron schedule

ivasilov:
Small fixes:
- Simplify the implementation in ScheduleSection.
- Fix the schedule input to turn red when its value has error.
- Fix the schedule input to clear the errors when set via presets or AI.
- Fix the ErrorBoundary to handle non-Error errors.
- Minor wording fixes.


apps/studio/components/ui/ErrorBoundaryState.tsx
```diff
@@ -1,15 +1,33 @@
...
 export const ErrorBoundaryState = ({ error, resetErrorBoundary }: FallbackProps) => {
   const router = useRouter()
-  const message = `Path name: ${router.pathname}\n\n${error.stack}`
-  const isRemoveChildError = error.message.includes("Failed to execute 'removeChild' on 'Node'")
+
+  let errorMessage = ''
+  let urlMessage = ''
+  let isRemoveChildError = false
+  if (isError(error)) {
+    errorMessage = error.message
```
joshenlim, line 27: could we use ternary for this? in favor of immutability

`const errorMessage = isError(error) ? error.message ?? ''`

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
```diff
@@ -76,47 +78,22 @@ export const CronJobScheduleSection = ({ form, supportsSeconds }: CronJobSchedul
...
-        setScheduleString('Invalid cron expression')
-        return
-      }
-
-      setScheduleString(CronToString(schedule))
-    } catch (error) {
-      setScheduleString('Invalid cron expression')
-      console.error('Error converting cron expression to string:', error)
+  const schedule = form.watch('schedule')
+  let scheduleString = ''
```
joshenlim, line 89: hmm can we use ternary for this? we can wrap the `.test` in a separate function that has a try catch

in favor of immutability + consistency

joshenlim:
unrelated - i'm seeing this error in the cron overview page
image

joshenlim:
also seeing this error when i try to generate cron with AI (not sure if its an issue though)
image

apps/studio/components/interfaces/Integrations/CronJobs/CronJobScheduleSection.tsx
joshenlim: pushed a fix for this!

apps/studio/components/interfaces/Integrations/CronJobs/CronJobs.utils.ts
```diff
@@ -171,5 +174,13 @@ export function getScheduleMessage(scheduleString: string, schedule: string) {
     .map((s, i) => (i === 0 ? s.toLowerCase() : s))
     .join(' ')
 
-  return `The cron will be run ${readableSchedule}.`
+  return `The cron will run ${readableSchedule}.`
+}
+
+export const checkIsSeconds = (value: string) => {
+  try {
+    return secondsPattern.test(value)
```
ivasilov, line 182: This function doesn't throw, `ToCronString` throws.

joshenlim (approved):
Tested once again on staging - lgtm other than the last console error [here](https://github.com/supabase/supabase/pull/30546#issuecomment-2516921412)! :)

PR #30555, opened 2024-11-19
Update ColumnType.tsx

hexronuspi:
When creating a column in a table, attempting to save without selecting a data type currently fails silently—it does not trigger an error, nor does it proceed with the save. This update introduces a clear alert to notify users if the data type is not selected, ensuring the user is guided to provide the necessary input before saving.

YES

Bug fix, feature

Saving a column without a data type fails without any information of, "what went wrong?".

## What is the new behavior?

A clear alert to notify users if the data type is not selected

The UI is not in accordance of Supabase, I just wanted to point the issue, I can further work on it.


Hallidayo:
Hi @hexronuspi - got a conflict to resolve but should be all good once resolved.

hexronuspi:
Hi @Hallidayo I left working on it due to less knowledge about JS during that time, will work on it.

PR #30565, opened 2024-11-20
chore: update invoice pay now redirect

alaister:

chore

Redirect URL is static

## What is the new behavior?

Uses the new endpoint from https://github.com/supabase/infrastructure/pull/20666


kevcodez (approved):
Tested on staging

PR #30581, opened 2024-11-21
fix: response error codes

alaister:

Bug fix

- Upgrades `openapi-fetch` & `openapi-typescript`
- Swaps generic error type for a `ResponseError`, with the `code` property
- Improves type safety

We're checking for error code in a few places but it doesn't actually exist.

## What is the new behavior?

We're now using `ResponseError` as a real type, so we can use it more safely.


apps/studio/data/ai/sql-debug-mutation.ts
joshenlim, line 1: I actually think this file isn't being used anymore with the new AI assistant

Under [SQLEditor.tsx line 156](https://github.com/supabase/supabase/blob/feat/new-project-creation-experiment/apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx#L156) debugSql is not being called, but only being used in the dependency array - i think we can remove this now

apps/studio/data/integrations/vercel-connection-update-mutate.ts
```diff
@@ -3,7 +3,7 @@ import { toast } from 'sonner'
 
 import { handleError, patch } from 'data/fetchers'
 import type { ResponseError } from 'types'
-import type { UpdateConnectionPayload } from './integrations.types'
+import type { EnvironmentTargets, UpdateConnectionPayload } from './integrations.types'
```
joshenlim, line 6: This import being used?

joshenlim (commented):
Tested on docs:
- POST via telemetry opt in

Tested on studio:
- Reading / creating / deleting access tokens
- AI renaming of untitled queries in SQL editor
- Edge functions overview page (Request 200 for functions.req-stats, functions.resource-usage call)
- Org settings -> Usage (Request 200 for org-daily-stats call)
- Project home page (Request 200 for usage.api-counts call)
- Deleting MFA factors in auth/users
- Reset password in auth/users
- Send magic link in auth/users
- Enabling database extension (to check execute-sql-query)
Local set up general smoke testing

Did not test:
- Send OTP in auth/users
- Updating JWT secret in settings/api

Bugs found on studio: Works on staging, but not on staging preview
- GET notifications endpoint is currently 400
- AI renaming of SQL snippet isn't saving the snippet name (Request is 200 nonetheless)
- Settings/API -> Project API Keys just shows "JWT secret is being updated" and then "Failed to update JWT secret"

alaister:
@joshenlim Thanks for the review! I've addressed most of what you found.

Could you please clarify what: "AI renaming of SQL snippet isn't saving the snippet name (Request is 200 nonetheless)", means? Is this when the user clicks rename manually or the auto renaming? (manually seems to work for me).

joshenlim:
Yup! Sorry should've added a screen grab - when you're on a new snippet then run it, we'd rename the snippet using our /title endpoint. The endpoint is triggering correctly and returns 200, but we're not actually setting the name of the snippet thereafter from the response of the /title call

https://github.com/user-attachments/assets/de38e37a-be0d-4fa7-b2db-3a32bd06ce61

alaister:
@joshenlim I'm not sure how it was working beforehand to be honest... anyway updated now!

joshenlim (approved):
Okay i think all should be good now - did a quick smoke test on staging preview but checked more thoroughly in the SQL editor since that was where it was last bugging out

saltcod (approved):

PR #30584, opened 2024-11-21
chore: declare fonts bucket in config

sweatybridge:

chore

## What is the new behavior?

Checks fonts file in code so they can be uploaded to storage from config.

## Additional context


alaister (approved):
LGTM!

PR #30591, opened 2024-11-21
Chore/trigger name length

saltcod:
prevent long trigger names from overflowing the triggers table

**test data**:  https://gist.github.com/saltcod/c2f0ec9294563e23f37231f647b7bcec

**Note**: 
trigger names are limited to 63 chars and are automatically truncated
https://share.cleanshot.com/fldH3dY6

screenshot-2024-11-21-at-13 25 13


TheOtherBrian1:
I'm not sure if this should be done or not, but it may be worthwhile to place a title attribute on the trigger name:
```
<span title={x.name} className="truncate max-w-48 inline-block">{x.name}</span>
```

joshenlim (approved):
LGTM! Tested on preview RE tooltip
JFYI if this is the pattern to follow we'll probably need to do it for the other tables on the various database pages (unless we choose to go down the data grid path for all of them)

PR #30592, opened 2024-11-21
fix: Handle Trigger name length in Triggers page

monicakh:

Bug fix

**Problem:** The trigger name is up to 63 characters; when having such a long name, the triggers table is cut off and the only way to see it is to zoom out.

**Steps to reproduce:**

create a trigger with the 63-character name, for example, "The_most_valuable_experiences_come_from_stepping_outside_your_c"

refresh the page, the table should be cut off.

Tested in Chrome, Firefox and Brave - everywhere looks the same and the table is indeed cut off, however, it is visible when zooming out on 80%. 

100% Chrome

image

80%

image

## What is the new behavior?

image


monicakh:
Dupe: https://github.com/supabase/supabase/pull/30591

PR #30618, opened 2024-11-22
Chore/spam scores

saltcod:
We currently show spam rules for anything greater than 0. 
Adjust to show just 1.5+

Test:
screenshot-2024-11-22-at-14 18 54


phamhieu:
The score doesn’t hold much significance in our use case. If the validation check returns any matching rules, it indicates that the email may be considered spam.
Afaik, the email header rules typically has a very high score, while the content rules score is lower. SpamAssassin heavily depends on the email header.

Moreover, the spamAssassin report appears to round the score values. When it returns 0, it should actually be 0.xxxx, as rules with a score of 0 are ignored during spam checks. We should change the logic to >= 0, as many content rules appear with a score of 0 in the report.

Screenshot 2024-11-05 at 8 38 12 AM

joshenlim:
Closing PR as per Hieu's comment above 🙂🙏 will be addressing this on the API end:
https://github.com/supabase/infrastructure/pull/20712

PR #30629, opened 2024-11-23
fix - logout in nextJS bootstrap auth example

PLyn:

YES

bug fix

log out button errors in bootstrap NextJS example errors for Supabase.Auth cannot be found

## What is the new behavior?

Log out now works without errors.

## Additional context

The createClient call in AuthButton.tsx for the logout logic was missing the await to prevent errors when logging out. This is specifically for the NextJS Auth cookie-based template for Bootstrap in the examples folder.


charislam (approved):
Thanks

PR #30630, opened 2024-11-23
Fix command in documentation for typescript 

quick007:

YES

Docs update

Command doesn't work in console
image

## What is the new behavior?

The command has been corrected and works without error

## Additional context

I am using the supabase CLI tool downloaded directly but I assume the behavior is the same


charislam (approved):
Thanks @quick007!

PR #30641, opened 2024-11-25
Chore/misc fixes 251124

joshenlim:
- Fix key prop in CreateIndexSidePanel
- Fix creating FK in ColumnEditor not validating types properly for array and non array types
- Fix creating FK in ColumnEditor on new column not updating types properly when target column is array


alaister (approved):
LGTM!

- Tested on preview

PR #30647, opened 2024-11-25
Feat/assistant additions

SaxonF:
I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

Makes some improvements to the assistant

there’s a bit here where the “Thinking…” loading state goes away too early and I think the request has failed
Show Query” when hidden, “Hide Query” when shown
border accidentally too short
Passes the current table and schema as context if we have it
Added "suggestions" when editing a policy


joshenlim (approved):
LGTM - tested on preview

- Suggestions when editing policy with assistant
- schema and table gets passed to generate-v3 when querying on table editor
- table is not passed if on a page that also has ID param (e.g sql editor)
- Current SQL snippet gets passed as context when opening assistant in SQL editor

PR #30648, opened 2024-11-25
Feat/assistant convert supabasejs

SaxonF:

YES

This adds the ability for the AI assistant to convert sql to supabase-js client code via the [sql-to-rest](https://github.com/supabase-community/sql-to-rest?tab=readme-ov-file#sql-is-a-very-open-language---how-can-it-all-translate-to-rest) package. If a query is not supported, it will suggest creating a function and give the supabase-js code to call the function.

It attempts to create supabase-js code but often gets it wrong.

## What is the new behavior?

It first calls the new tool, then if a query is not supported, it will create a function along with the supabase-js code to call that function.

## Additional context

image

image


apps/studio/pages/api/ai/sql/generate-v3.ts
```diff
@@ -105,6 +105,9 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse) {
       - Default to create or replace whenever possible for updating an existing function, otherwise use the alter function statement
       Please make sure that all queries are valid Postgres SQL queries
 
+      # You convert sql to supabase-js client code
```
joshenlim, line 108: minor nit - does the `#` need to be a `-` for consistency? not sure if that really matters

joshenlim (approved):
LGTM very cool! Tested asking Assistant to convert SQL to JS, and also asking it to convert it to RPC then using it in JS as well

Left a non-blocking comment :)

PR #30662, opened 2024-11-25
docs: improve connection docs

charislam:


TheOtherBrian1 (approved):
Much more succinct. There's nothing in particular I would change, but as we build out the quickstart guides for more ORMs, it'll be best to separate GUIs from them

PR #30666, opened 2024-11-26
chore: replace Markdown component in some generic studio components

kevcodez:
We no longer want to use Markdown when not necessary. I checked the props for these components and we were only passing in a regular text without any actual markdown

### DIFF
--- a/apps/studio/components/ui/Forms/FormHeader.tsx
+++ b/apps/studio/components/ui/Forms/FormHeader.tsx
@@ -1,6 +1,3 @@
-import ReactMarkdown from 'react-markdown'
-
-import { Markdown } from 'components/interfaces/Markdown'
 import { ReactNode } from 'react'
 import { cn } from 'ui'
 import { DocsButton } from '../DocsButton'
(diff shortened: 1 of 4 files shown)


apps/studio/components/ui/Forms/FormHeader.tsx
```diff
@@ -21,12 +18,8 @@ const FormHeader = ({
...
     <div className={cn(`mb-6 flex items-center justify-between gap-x-4 ${className}`)}>
       <div className="space-y-1">
-        <h3 className="text-foreground text-xl">
-          <ReactMarkdown unwrapDisallowed disallowedElements={['p']}>
-            {title}
-          </ReactMarkdown>
-        </h3>
-        {description && <Markdown content={description} className="max-w-full" />}
+        <h3 className="text-foreground text-xl prose">{title}</h3>
+        {description && <div className="prose text-sm max-w-full">{description}</div>}
```
alaister, line 22: For this one have you checked that we're not passing markdown to this component?

kevcodez: Yes - just plain text for both title and description

alaister (approved):
LGTM!

- Tested on preview

PR #30594, opened 2024-11-21
fix: supavisor migration issue

robsoncombr:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

Fix: problem during the first supabase startup in docker environment where supavisor is not able to apply migrations successfully due to lack of permissions to _supabase

## What is the current behavior?

When starting supabase for first time on docker env supavisor is not able to run migration due to lack of permissions to _supabase.

## What is the new behavior?

After change the user used in connection URL from postgres to supabase_admin following pattern from analytics and introduction to supavisor service in [PR 29596](https://github.com/supabase/supabase/pull/29596)

## Additional context

None

PR #30590, opened 2024-11-21
Supabase vector fix in docker-compose.yml

shawal-mbalire:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES/NO

## What kind of change does this PR introduce?

Bug fix

## What is the current behavior?

When you try docker compose vector doesn't run due to referencing /etc as etc

Please link any relevant issues here.

## What is the new behavior?

Now this error doesn't come file correctly referenced

Feel free to include screenshots if it includes visual changes.

PR #27487, opened 2024-06-24
Solution to fix #27312

seanpham99:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

Bug fix

## What is the current behavior?

[Please link any relevant issues here.](https://github.com/supabase/supabase/issues/27312)

## What is the new behavior?

The containers are healthy
image

PR #30674, opened 2024-11-26
fix(self-hosted): self hosted docker compose

avallete:

YES

- fix services health-checks
- fix supavisor user
- fix migrations to connect back to postgres database to match cli behavior
- add a smoke test ensuring changes made to the stack don't break the `docker-compose up` 

## Integrate changes from:
- https://github.com/supabase/supabase/pull/30594
- https://github.com/supabase/supabase/pull/30590

## What is the new behavior?

## Additional context


docker/docker-compose.yml
```diff
@@ -17,9 +17,9 @@ services:
           "CMD",
           "node",
           "-e",
-          "require('http').get('http://localhost:3000/api/profile', (r) => {if (r.statusCode !== 200) throw new Error(r.statusCode)})"
+          "require('http').get('http://studio:3000/api/profile', (r) => {if (r.statusCode !== 200) throw new Error(r.statusCode)})"
         ]
-      timeout: 5s
+      timeout: 10s
```
avallete, line 22: **remark**

Noticed that this heahtcheck for some reason is very slow (maybe node starting time) on my tests it was taking ~6-7s to return OK.

docker/docker-compose.yml
```diff
@@ -330,7 +330,7 @@ services:
     container_name: supabase-analytics
     image: supabase/logflare:1.4.0
     healthcheck:
-      test: [ "CMD", "curl", "http://localhost:4000/health" ]
+      test: [ "CMD", "curl", "http://analytics:4000/health" ]
```
sweatybridge, line 333: This command is run from within container. So localhost should be correct. Same goes for other health checks.

docker/docker-compose.yml
```diff
@@ -92,7 +92,7 @@ services:
           "--no-verbose",
           "--tries=1",
           "--spider",
-          "http://localhost:9999/health"
+          "http://auth:9999/health"
```
avallete, line 95: **remark**

`wget` doesn't necessarily love the "localhost", while `curl` will translate it to `127.0.0.1` and pass, `wget` depending on versions will error out. So I replaced all the mentions of localhost by the container network name so the behavior remain consistent.

docker/docker-compose.yml
vincentfretin: Yes when I tested the healthcheck, it took a few seconds to spawn the nodejs process, really slow.
I also had to change the host for the healthcheck to work.
Currently I'm running with the changes from https://github.com/supabase/supabase/pull/27487

avallete: Yes, I finally understood where the issue was with those hosts.

On the `studio` the issue lies into the built `server.js` file that contain this:

```
root@be346e9b78f9:/app# cat apps/studio/server.js 
const path = require('path')

const dir = path.join(__dirname)

process.env.NODE_ENV = 'production'
process.chdir(__dirname)

const currentPort = parseInt(process.env.PORT, 10) || 3000
const hostname = process.env.HOSTNAME || '0.0.0.0'

let keepAliveTimeout = parseInt(process.env.KEEP_ALIVE_TIMEOUT, 10)
```

Here we can see that the server is actually ran under `HOSTNAME` if defined in the env, which is the case within the `docker-compose` network container. 

Running `netstat` confirm that the server binding is over the docker alias and not `0.0.0.0` which explain why, within the container, trying to `curl http://studio:3000/api/profile` will work while `curl http://127.0.0.1:3000` or `curl http://localhost:3000` will fail.

And I think the issue is similar for the `storage` container even tough the "server" start file is much less readable, we can see that the actual host that we bind onto `0.0.0.0` and while this is similar for what we have for the `analytics` container, we can see in this one that the issue is that while `analytics` use `curl` which automatically alias `localhost` to `127.0.0.1` which match `0.0.0.0`, `wget` doesn't do the same. And will fail if you don't provide `0.0.0.0` or `127.0.0.1`

```
/opt/app/rel/logflare/bin # netstat -tuln | grep 4000
tcp        0      0 0.0.0.0:4000            0.0.0.0:*               LISTEN      
/opt/app/rel/logflare/bin #  wget --no-verbose --tries=1 --spider "http://localhost:4000/health"
Connecting to localhost:4000 ([::1]:4000)
wget: can't connect to remote host: Connection refused
/opt/app/rel/logflare/bin #  wget --no-verbose --tries=1 --spider "http://0.0.0.0:4000/health"
Connecting to 0.0.0.0:4000 (0.0.0.0:4000)
remote file exists
/opt/app/rel/logflare/bin #  curl -v "http://localhost:4000/health"
* processing: http://localhost:4000/health
*   Trying [::1]:4000...
* connect to ::1 port 4000 failed: Connection refused
*   Trying 127.0.0.1:4000...
* Connected to localhost (127.0.0.1) port 4000
> GET /health HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/8.2.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< cache-control: max-age=0, private, must-revalidate
< content-length: 112
< content-type: application/json; charset=utf-8
< date: Tue, 26 Nov 2024 15:25:38 GMT
< logflare-node: logflare@127.0.0.1
< server: Cowboy
< x-request-id: GAuOHZA13KjXRaIAACui
< 
* Connection #0 to host localhost left intact
{"nodes": ["logflare@127.0.0.1"],"nodes_count":1,"proc_count":804,"status":"ok","this_node":"logflare@127.0.0.1"}
```

We can see a similar setup with the `vector` container, where we can see that the `wget` health-check require to use the container name alias rather than the `localhost`.

@sweatybridge I see two ways to go with this, either we fix the underlying server builds to always bind to `0.0.0.0` but we will still need to change the `wget` healthchecks to be performed on either `127.0.0.1` or `0.0.0.0` rather than localhost. Or we can do like we currently do for the `vector` container, and use the container network alias as the static value that we can be sure should work the same across commands.

docker/docker-compose.yml
```diff
@@ -17,14 +17,16 @@ services:
...
         ]
-      timeout: 5s
+      timeout: 10s
       interval: 5s
       retries: 3
     depends_on:
       analytics:
         condition: service_healthy
+    ports:
+      - "3000:3000"
```
vincentfretin, line 29: About the way we expose the ports in this file, on ubuntu, docker adds an iptables rule automatically that will expose 3000 on the network. Same for the other ports in this docker-compose.yml file. That's something to be aware of if you don't have an additional security group in front of your server.
On my own deployment I rewrite all the ports section to add "127.0.0.1:" in front of each port, for example "127.0.0.1:4000:4000" to not expose it on the network interface.
There is probably the same issue on the cli, we may just expose the different ports to any public Wi-Fi.

avallete: Good catch this new export wasn't intended.

We might want to review the rest of the port expose in a dedicated PR, feel free to open an issue to discuss about this.

kiwicopple (approved):

docker/docker-compose.yml
sweatybridge: Using `auth` hostname sounds good to me. Thanks for digging into the root cause for this.

sweatybridge (approved):

PR #30675, opened 2024-11-26
chore: paginate user content

alaister:

chore

Requires: https://github.com/supabase/infrastructure/pull/20663


joshenlim:
## Just a reminder to flip the page limit to 50 (currently at 3)

Just need to do this before we merge in the PR, but the current limit at 3 is useful for testing. Also just wondering if we might be able to raise the page limit to 100? Ideally I'd like this change to not be too visible for most users

## If i've got a mixture of shared and private snippets, the first page of private snippets doesn't seem to load the full page of 3 results on initial render



Clicking load more the first time doesn't give more results in "Private" too - i have to click several times. This is my full snippet list for reference



## We should show the number of shared and favourites on first render



## Searching will need to be done on the API end

Snippet searching is currently done client side, but will need to be shifted to the API end, otherwise the pagination will limit the search results (e.g If sorted alphabetically and I'm looking for a snippet which name starts with "Z", UI will show 0 results which is not accurate)

## Empty state required for no search results



Related to the above RE searching to be done on API end

## "Load more" button seems to show up everytime i open the folder without refreshing the page (Non-blocking)



Not a biggie, but it's not consistent with the root  - but i reckon we don't need it, cause private snippets within folders won't be dynamic (new snippets can only be added by the user, not other users, so don't need load more to check if there's more snippets)

## If there's only 2 snippets here (given page limit 3), shouldn't need to show "Load more" here



## Sorting will need to be done on the API end (Non-blocking)

Snippet sorting is currently done on the client side, but will need to be shifted to the API end, otherwise we'll have weird behaviour like clicking "Load more" loads snippets at the top of the list

## Snippet preview not showing up in Cmd K (Non-blocking)

How it looks like on PR:


Reference how it looks like on staging:

saltcod:
it's showing load more here, but I'm pretty sure I have just 3 snippets — when should load more show?
screenshot-2024-12-16-at-16 20 01

saltcod:
moving from a folder back to the root, I got this error 

screenshot-2024-12-16-at-16 21 08

saltcod:
sharing a snippet didn't invalidate the list

to reproduce: 
1. create a snippet
2. add it to a private folder
3. move from private folder to shared 
4. see that shared count has been updated, but the snippet doesn't show up there even after you refresh

screenshot-2024-12-16-at-16 22 10

saltcod:
what's the new paginated amount per page? Is it actually 3? 
not sure what the rationale is there, but def feels punitively low. Can we not do a proper number there? 20-30 or so on first load at least?

joshenlim:
JFYI the page number will be increased to 100 before we merge this PR, it's currently set to 3 temporarily so that we can test it on staging without having to create too many snippets. The pagination change shouldn't visually affect most users as most users have less than 100 snippets

> it's showing load more here, but I'm pretty sure I have just 3 snippets — when should load more show?

The load more button shows up unless the current page that we're fetching has less results than the page size. This includes the slight edge case where you have the same number of snippets as the page size, in which case hitting load more then validates that we've reached the last page

> moving from a folder back to the root, I got this error

gotcha great catch! will get that fixed 🙏 

> sharing a snippet didn't invalidate the list

gotcha - there's an API issue that I've found RE sharing a query from a private folder, will get that fixed 🙏

joshenlim (commented):
Tried my best to test as many scenarios as possible locally, LGTM so far:

Creating snippets/folders
- [x] Creating a new snippet
- [x] Creating a new snippet from quick start
- [x] Creating a new snippet from template
- [x] Create folder
- [x] Rename folder
- [x] Rename folder with snippets
- [x] Delete folder
- [x] Delete folder with snippets

When a snippet has yet to be opened (clicked on) - tested all these while on /sql/new
- [x] In root -> share query
- [x] In root -> rename query
- [x] In root -> move to folder
- [x] In root -> Delete query
- [x] In a folder -> share query
- [x] In a folder -> rename query
- [x] In a folder -> move to another folder
- [x] In a folder -> move to root
- [x] In a folder -> delete query
- [x] Unshare query

When a snippet has been opened (clicked on) - tested all these while having the query open
- [x] In root -> share query
- [x] In root -> rename query
- [x] In root -> move to folder
- [x]  In root -> Delete query
- [x] In a folder -> share query
- [x] In a folder -> rename query
- [x] In a folder -> move to another folder
- [x] In a folder -> move to root
- [x] In a folder -> delete query
- [x] Unshare query

Will hold back the ✅  until we bump the page limit to 100

joshenlim (approved):
LGTM - whole team has tested this properly
there's a couple of small UX kinks but we can address them separately as they are non blocking (to prevent this PR from getting bigger than it already is)

PR #30682, opened 2024-11-26
Fix/ef selection

saltcod:
In the cron ui, the EF you choose saves fine, 
but whenever you re-open the Cron sheet, it chooses 
the first edge function it finds — edgeFunctions[0]

screenshot-2024-11-26-at-15 27 54

### DIFF
--- a/apps/studio/components/interfaces/Integrations/CronJobs/CronJobsTab.tsx
+++ b/apps/studio/components/interfaces/Integrations/CronJobs/CronJobsTab.tsx
@@ -81,7 +81,7 @@ export const CronjobsTab = () => {
                   })
                 }
               >
-                Create a queue
+                Create a cron job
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Integrations/CronJobs/CronJobsTab.tsx
```diff
@@ -81,7 +81,7 @@ export const CronjobsTab = () => {
                   })
                 }
               >
-                Create a queue
+                Create a cron job
```
saltcod, line 84: unrelated, but tucking this tiny change in here too

joshenlim (approved):
LGTM! Verified on preview that editing a cron job which uses an edge function will select the correct edge function, instead of the first available edge function

PR #30685, opened 2024-11-27
Assistant refinements

SaxonF:

YES/NO

- UI refinements to blank state including DotGrid instead of gradients, collapsables for button groups, ensure fade div hides property on load
- Refine system prompt as it was being a bit too pushy with supabase-js conversions

image

### DIFF
--- a/apps/studio/components/ui/DotGrid.tsx
+++ b/apps/studio/components/ui/DotGrid.tsx
@@ -1,4 +1,3 @@
-import React from 'react'
 import { motion } from 'framer-motion'
 
 interface DotGridProps {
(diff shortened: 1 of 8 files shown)


apps/studio/components/ui/DotGrid.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react'
+import { motion } from 'framer-motion'
+
+interface DotGridProps {
+  rows: number
+  columns: number
+  count: number
+}
+
+const DotGrid: React.FC<DotGridProps> = ({ rows, columns, count }) => {
```
joshenlim, line 10: in general let's avoid using FC, but use PropsWithChildren instead ([ref](https://medium.com/raccoons-group/why-you-probably-shouldnt-use-react-fc-to-type-your-react-components-37ca1243dd13))

also in this case since we're not expecting a children, just DotGridProps will do

joshenlim (approved):
LGTM! Changes look good - although just need to double check on the output where the assistant renders a chart when im asking for a list of data

e.g "for the countries table, can I check which are in asia and europe?" -> expecting a list of results, but assistant renders an empty chart instead

able to verify that it happens on this branch, staging renders a list as expected

PR #30692, opened 2024-11-27
Prevent non owners from editing shared snippets

joshenlim:
As per PR title


saltcod (approved):

saltcod:
Confirmed in preview: 
- I can share a snippet and still edit it
- team cannot edit it
- team cannot delete it
- team can duplicate it

PR #30693, opened 2024-11-27
ref: Remove experts page and setup permanent redirects

kamilogorek:
Probably needs some adjustment for `apps/www/pages/partners/index.tsx`, as there's some empty spacing now:

image

Note: it doesn't redirect `/partners/integrations/<name>` when given partner is `expert` type, as we prebuild them and redirects cannot be used when prerendering. It's fine though as we never used those in the first place and all entries were marked as `unapproved` in the database so they are not accessible anyway.


fsansalvadore (approved):
looks all good 👍

PR #30695, opened 2024-11-27
Remove the `queue_` prefix from queues API endpoints

olirice:
Removes the `queue_` prefix from the Data API endpoints that are created when the API is toggled on for the Supabase Queues integration

For example
`pgmq_public.queue_send(...)` becomes `pgmq_public.send(...)`

and from client libraries the call would look like 
```python
supabase: Client = create_client(url, key, options = SyncClientOptions(schema='pgmq_public']))

response = supabase.rpc(
    "send", # updated from queue_send
    {
        "queue_name": "qux",
        "message": {"foo": "bar"},
         "sleep_seconds": 0 
    }
).execute()
```

This will make the endpoints slightly less discoverable in the OpenAPI spec but we have created an internal task to improve OpenAPI specs to be grouped by schema, which will gets us the best of both worlds.


joshenlim (approved):
LGTM! Tested on staging preview - enabling / disable expose postgrest settings in queues

PR #30698, opened 2024-11-27
fix: invalid migration

charislam:
Fixes an invalid migration that was preventing other migration files from being pushed on top. The two `alter table` changes are already in prod migration history under this timestamp, but the `comment` change is not, so we should be safe to just delete it and add it in a separate migration.

### DIFF
--- a/supabase/migrations/20241127183137_comment_meetup_timezone.sql
+++ b/supabase/migrations/20241127183137_comment_meetup_timezone.sql
@@ -1,0 +1,1 @@
+comment on column meetups.timezone is 'Needs to be in America/Los_Angeles format.';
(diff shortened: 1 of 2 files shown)


supabase/migrations/20241127183137_comment_meetup_timezone.sql
```diff
@@ -0,0 +1 @@
+comment on column meetups.timezone is 'Needs to be in America/Los_Angeles format.';
```
fsansalvadore, line 1: This wasn't a very vital comment, but fair enough 😁

fsansalvadore (approved):

PR #30705, opened 2024-11-27
Update access-control.mdx (type cast owner_id to uuid)

sojs-coder:

YES

Docs update.

Page for explaining how to enable RLS on objects has the following code snippet:

```sql
create policy "Individual user Access"
on storage.objects for select
to authenticated
using ( (select auth.uid()) = owner_id );
```
In the schema, owner_id is type `text`:

image

This snippet compares two different types, and throws an error.
In order to fix this, we need to type case owner_id to uuid so the proper comparison can be made.

```sql
create policy "Individual user Access"
on storage.objects for select
to authenticated
using ( (select auth.uid()) = owner_id::uuid );
```

## What is the new behavior?

Added `::uuid` type cast.


charislam (approved):
Thanks @sojs-coder!

PR #30706, opened 2024-11-28
Add inline completions to SQL Editor

SaxonF:

YES

Adds the ability to make edits to selected SQL directly within the editor.

You have to open up the assistant, chat, and then re-save it back in.

## What is the new behavior?

Generate and compare diff directly within editor.

image


apps/studio/pages/api/ai/monaco/complete.ts
```diff
@@ -0,0 +1,101 @@
...
+    const result = await streamText({
+      model: openai('gpt-4o-mini'),
+      maxSteps: 5,
+      tools: getTools({ projectRef, connectionString, authorization, includeSchemaMetadata }),
+      system: `${SQL_SYSTEM_PROMPT}
+
+      Follow these instructions:
+      - First look at the list of provided schemas and if needed, get more information about a schema. You will almost always need to retrieve information about the public schema before answering a question. If the question is about users, also retrieve the auth schema.
+
+      Here are the existing database schema names you can retrieve: ${schemas}.`,
```
gregnr, line 72: Recommend using [`common-tags`](https://github.com/zspecza/common-tags) here to strip indents so that indentation spaces aren't sent downstream:

```ts
{
  ...
  system: codeBlock`
    ${SQL_SYSTEM_PROMPT}
    Follow these instructions:
    - First look at the list of provided schemas and if needed, get more information about a schema. You will almost always need to retrieve information about the public schema before answering a question. If the question is about users, also retrieve the auth schema.
    Here are the existing database schema names you can retrieve: ${schemas}.
  `,
  ...
}
```

They have a bunch of tags available like `stripIndent`, `oneLine`, etc. `codeBlock` is like `stripIndent` except it also strips intents for placeholder values like you have here.

gregnr (approved):

saltcod (approved):

PR #30707, opened 2024-11-28
chore: automatically mark default payment method [GEN-12003]

kevcodez:
When customers go through the upgrade flow via the plan upgrade modal, we want to automatically confirm any newly added payment method as default payment method. We've seen some issues on Orb that customers would upgrade without a default payment method. This change should help mitigate it further.


alaister (approved):
LGTM!

staaldraad (approved):

PR #30708, opened 2024-11-28
fix: add new auth error codes

J0:

YES

As per title


apps/docs/next-env.d.ts
charislam, line 1: Was this auto-deleted? Never had that happen before...

charislam (approved):
Approving to unblock but would prefer it if you alphabetized the list 🙏🏼

apps/docs/next-env.d.ts
J0: I accidentally added it and then accidentally deleted it. Please ignore, will reset in a bit

PR #30714, opened 2024-11-28
Validate spam on save

joshenlim:
Alter spam validation behaviour to be on save instead of debounce


kangmingtay (approved):

alaister (approved):
LGTM!

- Tested on preview with a valid and spammy templates

Issue #30528, opened 2024-11-18
Cron job creation tool generates wrong command


L3g7:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When creating a cron job using the [Create a new cron job]-Button in Dashboard -> Integrations -> Cron-jobs that invokes an edge function via HTTP GET, the generated net.http_get contains a `body` argument which it should not have, resulting in the following error:
```
ERROR:  function net.http_get(url => unknown, headers => jsonb, body => jsonb, timeout_milliseconds => integer) does not exist
LINE 3:       net.http_get(
              ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

## To Reproduce

1. Go to a project dashboard -> Integrations -> Cron Jobs
4. Click on Create a new cron job
5. Select Type = Supabase Edge Function and Method = Get
6. Create cron job
7. See wrong command

## Expected behavior

I expected the `body` argument to be omitted.

GaryAustin1:
I worked with this user and confirm this is a bug.

GaryAustin1:
@saltcod This should be fixed before launch...

saltcod:
Yup thanks @GaryAustin1 on the radar.

saltcod:
Marking this one as closed per https://github.com/supabase/supabase/pull/30721

thanks for the report!

PR #30721, opened 2024-11-28
fix: Polish stuff for Cron

ivasilov:
Few fixes:
- Change the naming of the Cron extension.
- Previous runs now has its own button.
- Fix a bug where body would be sent on GET requests. [Issue](https://github.com/supabase/supabase/issues/30528)
- Replace the http parameters with body.
- Use query param to show/hide the create cron job sheet.


saltcod (approved):

saltcod (changes requested):
issue saving fields

saltcod (approved):

saltcod:
Tested creating and saving a cron job with a webhook.

PR #30730, opened 2024-11-29
More assistant refinements

SaxonF:

YES

- Refines styles like alignment, spacing, typography
- Moves info text to header
- Updates AI Icon

https://github.com/user-attachments/assets/50cbd57d-ade9-460d-8767-0958c9bf3693


joshenlim:
do we want the hover effect always on though? there's a type error because of that
i'm assuming no though - e.g on like "edit with assistant" buttons we wouldn't want that

i'll fix it to default to false for now (allowHoverAnimation property)

joshenlim (approved):
LGTM generally! Tested on staging preview - pushed up some minor refactors and fixes

PR #30779, opened 2024-12-02
Remove assistant V2 from feature preview

joshenlim:


apps/studio/components/interfaces/SQLEditor/AiAssistantPanel/AiMessagePre.tsx
```diff
@@ -20,6 +20,7 @@ interface AiMessagePreProps {
   className?: string
 }
 
+// [DEPRECATED] To delete
```
alaister, line 23: for all of these if you use this syntax vs code will understand and make it clearer not to use them. Not needed if these are gonna be immediately deleted
```suggestion
/**
 * @deprecated to delete
 */
```

alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/interfaces/SQLEditor/AiAssistantPanel/AiMessagePre.tsx
joshenlim: yup! will look into removing these guys next week after LW :)

PR #30782, opened 2024-12-02
fix: Add missing dependency to fix the docker build

ivasilov:


alaister (approved):
LGTM!

- Ensured preview still opens

PR #30784, opened 2024-12-02
fix: oid types in text editor

alaister:

Bug fix

Double clicking on an oid column value explodes

## What is the new behavior?

Value is cast to a string so it can be rendered correctly by Monaco


joshenlim (approved):
LGTM - verified on an oriole project with oid columns

PR #30811, opened 2024-12-02
Chore/cron footer

saltcod:
fix the cron footer so it works with infinite lists

screenshot-2024-12-02-at-13 55 53


ivasilov (approved):

saltcod:
checked preview, footer works on infinite list pages

PR #30823, opened 2024-12-03
Feat/save assistant messages

SaxonF:

YES

This PR saves aiAssistantState to local storage so that chat history is saved. It adds user message to state and then once useChat finishes, it adds assistant messages to state. valtio [subscribe](https://valtio.dev/docs/how-tos/how-to-persist-states) is used to watch for state changes and save to storage. 

The added is isClient state added to [fix hydration error](https://nextjs.org/docs/messages/react-hydration-error#solution-1-using-useeffect-to-run-on-the-client-only) however I'm not sure if this is the best approach. Would it be better to remove the open check on project layout, and render the component dynamically instead with a conditional inside the component to return null if open is false?

### DIFF
--- a/apps/studio/state/app-state.ts
+++ b/apps/studio/state/app-state.ts
@@ -76,14 +76,20 @@ const getInitialState = () => {
   const urlParams = new URLSearchParams(window.location.search)
   const aiAssistantPanelOpenParam = urlParams.get('aiAssistantPanelOpen')
 
-  const parsedAiAssistant = stored
-    ? JSON.parse(stored, (key, value) => {
+  let parsedAiAssistant = INITIAL_AI_ASSISTANT
+
+  try {
+    if (stored) {
+      parsedAiAssistant = JSON.parse(stored, (key, value) => {
         if (key === 'createdAt' && value) {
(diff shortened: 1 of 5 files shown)


apps/studio/state/app-state.ts
```diff
@@ -51,9 +47,70 @@ const INITIAL_AI_ASSISTANT: AiAssistantPanelType = {
...
+  const aiAssistantPanelOpenParam = urlParams.get('aiAssistantPanelOpen')
+
+  const parsedAiAssistant = stored
+    ? JSON.parse(stored, (key, value) => {
+        if (key === 'createdAt' && value) {
+          return new Date(value)
+        }
+        return value
+      })
+    : INITIAL_AI_ASSISTANT
```
alaister, line 86: There's no reason that this value should be improperly formatted, but since it's user controlled we should properly err on the side of caution and wrap it with a try/catch?

alaister (approved):
LGTM!

- Tested on preview
  - confirmed `aiAssistantPanelOpen` param works
  - refresh to see value is persisted correctly
  
 One small suggestion:

PR #30827, opened 2024-12-03
feat: improved disk size insights

kevcodez:
Numerous kaizen improvements to disk size insights

- Disk Size usage is broken down to Database/WAL/System (instead of just showing "used space")
- Fixed a few docs links
- Added docs link to "Compute and disk" sections
- Clarified disk pricing in our docs
- Slight copy improvements to the database report
- Added docs for disk size distribution and some info around WAL size

disk-size-distribution

There are still a lot of things we can do in terms of WAL insights and easier ways to dig into database space (database reports page), but that's not in the scope of this PR.


alaister (approved):
LGTM!

- Tested on preview

PR #30608, opened 2024-11-22
add cron jobs events

jordienr:

PR #30874, opened 2024-12-04
Add telemetry for cron

joshenlim:
As per https://github.com/supabase/supabase/pull/30608

### DIFF
--- a/apps/studio/components/interfaces/Integrations/CronJobs/CronJobCard.tsx
+++ b/apps/studio/components/interfaces/Integrations/CronJobs/CronJobCard.tsx
@@ -8,6 +8,8 @@ import { SQLCodeBlock } from 'components/interfaces/Auth/ThirdPartyAuthForm/SqlC
 import { useProjectContext } from 'components/layouts/ProjectLayout/ProjectContext'
 import { CronJob } from 'data/database-cron-jobs/database-cron-jobs-query'
 import { useDatabaseCronJobToggleMutation } from 'data/database-cron-jobs/database-cron-jobs-toggle-mutation'
+import { useSendEventMutation } from 'data/telemetry/send-event-mutation'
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Integrations/CronJobs/CronJobCard.tsx
```diff
@@ -68,19 +73,47 @@ export const CronJobCard = ({ job, onEditCronJob, onDeleteCronJob }: CronJobCard
...
                   </DropdownMenuItem>
                   <DropdownMenuSeparator />
-                  <DropdownMenuItem onClick={() => onDeleteCronJob(job)}>
+                  <DropdownMenuItem
+                    onClick={() => {
+                      sendEvent({
+                        action: TELEMETRY_EVENTS.CRON_JOBS,
+                        value: TELEMETRY_VALUES.CRON_JOB_DELETE_CLICKED,
+                        label: 'User clicked delete cron job button',
+                      })
```
alaister, line 113: This event triggers only when the modal opens and not when they actually delete a cron job – is that okay?
(same with the edit one above)

alaister (approved):
LGTM!

- Tested on preview and verified network console to see event requests

apps/studio/components/interfaces/Integrations/CronJobs/CronJobCard.tsx
joshenlim: yeah - i think this is okay for now, we'll track both "clicked" and "actually deleted" - which reminds me i need to add the latter haha

joshenlim: we need to think through properly though what we want to track and what for - but since we're launching today, we'll just add this in to get _some_ telemetry

PR #30875, opened 2024-12-04
Polish scrolling on cron job history

joshenlim:
Previously:
- Small gap at the top
- Page can be scrolled
image

After:
- No more gap
- Only the data grid is scrollable
image


alaister (approved):
LGTM!

- Tested on preview

PR #30913, opened 2024-12-05
Chore/fix error handling in generate v3

joshenlim:
generate-v3 is not an edge API route, so needs a res.send for error handling


alaister (approved):
LGTM

PR #30929, opened 2024-12-05
Update postgis.mdx

Jordaneisenburger:

image

Docs update

The 4th option in the list isn't available in the Supabase dashboard hence I remove it from the docs, since we don't name it `gis.geography(POINT) not null` is causing an error as it doesn't recognize `gis`.

## What is the new behavior?
The documentation is now correct and runs without errors


charislam:
Thanks for opening the issue @Jordaneisenburger!

Just double-checking, you don't see this modal when you enable the gis extension? (just double-checked on my account and it showed for me)

image

Jordaneisenburger:
@charislam Ahh I see, my bad I just added it to the extensions schema which was the default and made sense to me so didn't even check other options. Not sure if adding it to the extension schema is actually a good thing todo but good to know it is indeed working as documented.

PR #30955, opened 2024-12-05
refactor(docs): guides toc headers

charislam:
Remove the dangerouslySetInnerHTML which is prone to breakage


joshenlim (approved):
LGTM! Verified that the TOC in `/docs/guides/local-development/cli/config#auth.external.provider.enabled` still rendered correctly + skimmed other pages

PR #30966, opened 2024-12-06
Adjust behaviour of alert error support

joshenlim:


alaister (approved):
LGTM!

- Tested on preview and verified that the error was attached in network request

PR #30978, opened 2024-12-06
feat: update emails to use a valid email address

cstockton:
The purpose of this change is to reduce the bounce back rate of the default email provider at Supabase. This change will allow users to continue copying examples from the docs. Without it, an upcoming change in supabase/auth[1] will begin rejecting these invalid emails.

[1] https://github.com/supabase/auth/pull/1845


cstockton:
I would be happy to make additional pull requests for any other places the docs team can point out that I may have missed.

kiwicopple (approved):

kangmingtay (approved):

PR #31016, opened 2024-12-09
fix: add frequently asked questions to auth identity linking docs

J0:

Adds a frequently asked questions section in response to a support ticket with the intention of making scenarios clearer for users coming in fresh to Supabase Auth. 

This is not exactly linking related but guessing the questions if any would typically come when the developer interacts with this page. Open to exposing this information on another separate page if folks have one in mind

### DIFF
--- a/apps/docs/content/guides/auth/auth-identity-linking.mdx
+++ b/apps/docs/content/guides/auth/auth-identity-linking.mdx
@@ -173,7 +173,7 @@ if google_identity:
 </TabPanel>
 </Tabs>
 
-## Frequently Asked Questions
+## Frequently asked questions
 
 ### How do I add email with password login to my OAuth account?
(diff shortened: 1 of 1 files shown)


apps/docs/content/guides/auth/auth-identity-linking.mdx
```diff
@@ -172,3 +172,11 @@ if google_identity:
 
 </TabPanel>
 </Tabs>
+
+## Frequently asked questions
+
+### How to add email/password login to an OAuth account?
+Call the `updateUser({ password: 'validpassword'})` to add email with password authentication to an account created with an OAuth provider (Google, GitHub, etc.).
+
+### What happens if I sign up with email after OAuth?
```
J0, line 181:
```suggestion
### Can you sign up with email if already using OAuth?
```

charislam (approved):

PR #31032, opened 2024-12-10
docs: add realtime benchmarks docs

egor-romanov:

- docs: add realtime benchmarks docs


apps/docs/content/guides/realtime/benchmarks.mdx
```diff
@@ -0,0 +1,153 @@
...
+
+### Realtime's Postgres Changes
+
+Realtime systems usually require forethought because of their scaling dynamics. For the `Postgres Changes` feature, every change event must be checked to see if the subscribed user has access. For instance, if you have 100 users subscribed to a table where you make a single insert, it will then trigger 100 "reads": one for each user.
+
+There can be a database bottleneck which limits message throughput. If your database cannot authorize the changes rapidly enough, the changes will be delayed until you receive a timeout.
+
+Database changes are processed on a single thread to maintain the change order. That means compute upgrades don't have a large effect on the performance of Postgres change subscriptions. You can estimate the expected maximum throughput for your database below.
+
+If you are using Postgres Changes at scale, you should consider using separate "public" table without RLS and filters. Alternatively, you can use Realtime server-side only and then re-stream the changes to your clients using a Realtime Broadcast.
```
charislam, line 145:
```suggestion
If you are using Postgres Changes at scale, you should consider using a separate "public" table without RLS and filters. Alternatively, you can use Realtime server-side only and then re-stream the changes to your clients using a Realtime Broadcast.
```

apps/docs/public/img/guides/realtime/payload-size-50kb.png
charislam, line 1: Is this image used? Didn't notice it when I was reading

charislam (approved):

apps/docs/public/img/guides/realtime/payload-size-50kb.png
egor-romanov: not yet, i wanted Filipe to take a look at the changes too 

and if Filipe thinks we should add it too, then we can do it easily, or otherwise ill just remove the file

filipecabaco (approved):

PR #31033, opened 2024-12-10
Reinstate reset db password

joshenlim:
Was accidentally hidden with the new connect UI feature flag, placed it back in the same position as before

### DIFF
--- a/apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseSettings.tsx
+++ b/apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseSettings.tsx
@@ -102,135 +102,129 @@ const DatabaseSettings = () => {
   }, [primaryConfig?.pool_mode])
 
   return (
-    <>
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Settings/Database/DatabaseSettings/DatabaseSettings.tsx
```diff
@@ -102,135 +102,129 @@ const DatabaseSettings = () => {
...
+                  value={`reference=${projectRef}`}
+                  label="Options"
                 />
-              </>
-            )}
-          </Panel.Content>
-        </Panel>
-      </section>
-
-      <ResetDbPassword disabled={isLoading || isError} />
```
joshenlim, line 232: i actually dont see why there was a need to disable this based on the statuses of read replicas, addons, and pooler config - if anyone might remember

alaister (approved):
LGTM!

- Tested on preview

MildTomato (approved):

PR #31048, opened 2024-12-10
fix: Change the users infinite query to use like instead of ilike

ivasilov:
This PR changes the Users dashboard to use `LIKE` instead of `ILIKE` when searching for a user by id, email or phone. This is technically ok because all ids, emails or phones are already lowercased on insertion.


saltcod (approved):

joshenlim (approved):
LGTM! Tested searching for users on staging preview

PR #31063, opened 2024-12-11
Allow multiple r2np

jordienr:
- Allows users to restore to new projects more than once

The R2NP tab will show a list with previous restores (with links to the new projects) and the available backups to restore from.

CleanShot 2024-12-13 at 09 34 38@2x

## When you visit a project that was restored 
CleanShot 2024-12-13 at 12 38 31@2x

### DIFF
--- a/apps/studio/pages/project/[ref]/database/backups/restore-to-new-project.tsx
+++ b/apps/studio/pages/project/[ref]/database/backups/restore-to-new-project.tsx
@@ -201,9 +201,7 @@ const RestoreToNewProject = () => {
           If you need to restore a project to multiple other projects, please reach out via [support](/support/new?ref=${project?.ref}).`}
         />
         <Button asChild type="default">
-          <Link
-            href={`/dashboard/project/${(cloneStatus?.cloned_from?.source_project as any)?.ref || ''}`}
-          >
+          <Link href={`/project/${(cloneStatus?.cloned_from?.source_project as any)?.ref || ''}`}>
             Go to original project
           </Link>
         </Button>
--- a/apps/studio/pages/project/[ref]/database/backups/restore-to-new-project.tsx
+++ b/apps/studio/pages/project/[ref]/database/backups/restore-to-new-project.tsx
@@ -225,37 +225,6 @@ const RestoreToNewProject = () => {
     )
   }
 
-  // if (lastClone?.status === 'FAILED') {
-  //   return (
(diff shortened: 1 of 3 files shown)


joshenlim (commented):
Some UX feedback based on observations:
- It’s odd that the whole UI switches when we’re restoring to a project
  - Is it possible to show the “Previous Restorations” table still while we’re currently restoring to a project?
  - The project that we’re currently restoring could show up there too
  - If we’re preventing multiple restores to new projects concurrently, we can disable the PITR picker
- The UI doesn’t automatically update once the restore has completed
  - We used to have some long polling logic iirc, but I can verify that there's no network requests being made while the restoration is happening
- What if the project that we restored to is now deleted?
  - Should it still be under the “Previous restorations” list?
  - Minimally i think we should mark it as deleted if we wanna keep it there
  
  
 LGTM otherwise - tested on the preview that restoring to multiple projects work 🙂

joshenlim (approved):
LGTM - Tested on staging preview :)

apps/studio/pages/project/[ref]/database/backups/restore-to-new-project.tsx
```diff
@@ -108,13 +105,81 @@ const RestoreToNewProject = () => {
...
+        <div className="grid grid-cols-4 gap-2 text-sm p-4 group">
+          <div className="min-w-24 truncate">{(clone.target_project as any).name}</div>
+          <div>
+            <StatusBadge status={clone.status} />
+          </div>
+          <div>
+            <TimestampInfo
+              className="font-mono text-xs text-foreground-lighter"
+              value={clone.inserted_at ?? ''}
+              // isUTC={true}
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/database/backups/restore-to-new-project.tsx:154]: remove if not needed?

apps/studio/pages/project/[ref]/database/backups/restore-to-new-project.tsx
```diff
@@ -108,13 +105,81 @@ const RestoreToNewProject = () => {
...
+        >
+          <div className="min-w-24 truncate">{(clone.target_project as any).name}</div>
+          <div>
+            <StatusBadge status={clone.status} />
+          </div>
+          <div>
+            <TimestampInfo
+              className="font-mono text-xs text-foreground-lighter"
+              value={clone.inserted_at ?? ''}
+              // isUTC={true}
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/database/backups/restore-to-new-project.tsx:173]: remove if not needed?

PR #31067, opened 2024-12-11
Standardize PH events

joshenlim:
As per PR title
- Add TS enum for `TelemetryActions`, avoid using arbitrary strings so that we can easily reference what events we're tracking through the dashboard
- Simplified `sendEvent` mutation to just accept `action` and `properties`
- Removed unnecessary event sending when signing up, or signing in via MFA


alaister (approved):
LGTM!

- Tested on preview by verifying various events are being sent correctly

joshenlim:
i'll wait for @4L3k51 to give the green light before merging this 🙂 🙏

PR #31070, opened 2024-12-11
fix: Lowercase the search term before searching users table

ivasilov:


apps/studio/components/interfaces/RoleImpersonationSelector/UserImpersonationSelector.tsx
```diff
@@ -95,7 +95,7 @@ const UserImpersonationSelector = () => {
               )
             }
             placeholder="Search for a user.."
-            onChange={(e) => setSearchText(e.target.value.trim())}
+            onChange={(e) => setSearchText(e.target.value.trim().toLocaleLowerCase())}
```
joshenlim, line 98: for this one actually we shouldn't be doing trim and lowerCase in onChange setSearchText, as that means that the user can enter "space" in the input, nor enter uppercase letters

for this one specifically, we should shift the trim and toLowerCase to L35 here probably
https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/RoleImpersonationSelector/UserImpersonationSelector.tsx#L35
image

joshenlim (approved):
LGTM! Tested on staging searching on the role impersonation UI + users UI

PR #31072, opened 2024-12-11
postgres version: refetch on window focus false

jordienr:


apps/studio/components/interfaces/ProjectCreation/PostgresVersionSelector.tsx
```diff
@@ -58,11 +58,16 @@ export const PostgresVersionSelector = ({
...
-  })
+  } = useProjectCreationPostgresVersionsQuery(
+    {
+      cloudProvider,
+      dbRegion,
+      organizationSlug,
+    },
+    {
+      refetchOnWindowFocus: false,
+    }
```
joshenlim, line 69: this one seems unnecessary? the ticket that we're addressing is only about the region selector

postgres version selector is currently not available to users too, only internal

apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
```diff
@@ -35,9 +35,14 @@ export const RegionSelector = ({ cloudProvider, field }: RegionSelectorProps) =>
...
-  const { isLoading: isLoadingDefaultRegion } = useDefaultRegionQuery({
-    cloudProvider,
-  })
+  const { isLoading: isLoadingDefaultRegion } = useDefaultRegionQuery(
+    {
+      cloudProvider,
+    },
+    {
+      refetchOnWindowFocus: false,
+    }
```
joshenlim, line 44: i don't think doing this here will have any difference though, as this isn't mutating the form values

joshenlim: this is probably where it might make more sense, but the refetchOnWindowFocus is already false
https://github.com/supabase/supabase/blob/d4c928ca72dcae65cfd4b072b3c8a8aea06cd8dd/apps/studio/pages/new/%5Bslug%5D.tsx#L168-L177

PR #31082, opened 2024-12-11
Chore/www session consent

saltcod:


packages/common/fetchWrappers.tsx
```diff
@@ -3,14 +3,17 @@ interface DataProps {
 }
 
 export const post = (url: string, data: DataProps, options = {} as { [key: string]: any }) => {
-  const { optionHeaders, ...otherOptions } = options
+  const { headers: optionHeaders, ...otherOptions } = options
```
saltcod, line 6: this wasn't working — always returning a 400 error. 
so I borrowed the implementation from www/lib/fetchWrapper.ts

this function is only used twice, here and in feedback submission in docs. Checked and that still works.

packages/common/telemetry.ts
```diff
@@ -1,7 +1,12 @@
 import { post } from './fetchWrappers'
 
 export function handlePageTelemetry(API_URL: string, route: string, telemetryProps: any) {
-  const { page_url, search, language, viewport_height, viewport_width } = telemetryProps
+  const { page_url } = telemetryProps
+
+  // check if ph exists, if not use an empty object
```
saltcod, line 6: the original destructuring wasn't working correctly

packages/common/telemetry.ts
```diff
@@ -1,7 +1,12 @@
...
+  const ph = telemetryProps?.ph || {}
```
joshenlim, line 7: this looks wrong though - from ConstentToast we pass in `telemetryProps` which is derived from `useTelemetryProps()`, which returns this
https://github.com/supabase/supabase/blob/master/packages/common/hooks/useTelemetryProps.ts#L19

`ph` is not part of the return object

packages/common/telemetry.ts
```diff
@@ -1,7 +1,12 @@
 import { post } from './fetchWrappers'
 
 export function handlePageTelemetry(API_URL: string, route: string, telemetryProps: any) {
```
joshenlim, line 3: I reckon it's best we define the shape of telemetryProps here - rather than leaving it as `any`

at least TS can help us check for issues

packages/common/telemetry.ts
ivasilov: The API endpoint is defined in `api.d.ts`. We should just use it here.

ivasilov (approved):

4L3k51 (approved):

PR #31095, opened 2024-12-12
ref: Use non-deprecated delete user endpoint

kamilogorek:
Fixes an issue when someone modifies `auth.users` schema and tries to delete a user by passing whole user object to the endpoint.


kangmingtay (approved):

kiwicopple (approved):

alaister (approved):
LGTM!

- Tested on preview

PR #31099, opened 2024-12-12
chore: Downgrade prettier to v3

ivasilov:
This PR downgrades prettier version to v3. The v4 prettier had a bug where it imports a package which imports a v3.2 prettier which then results in usage of the v3.2 as the main one.


joshenlim (approved):
LGTM! Minimally tested logging into the staging preview

PR #31100, opened 2024-12-12
chore: Bump nanoid to non-vuln version

ivasilov:


joshenlim (approved):
LGTM - should be okay we don't use nanoid in studio
did some basic smoke test just to be sure on preview

PR #31104, opened 2024-12-12
chore: Fix the env vars for turbo

ivasilov:
This PR fixes all apps and puts their env vars in the proper category (normal or pass-through). The difference is whether they're used as cache keys in turbo (pass-through env vars are not used as keys).


alaister (approved):
LGTM!

- Tested all available previews

staaldraad (approved):

fsansalvadore (approved):
lgtm!

PR #31113, opened 2024-12-13
Chore/update rules in generate v3

joshenlim:
- Always have a primary key
- Automatically create an index for foreign keys
- RLS should be enabled by default
- Example prompt: give me tables for countries and cities with foreign keys between them
Screenshot 2024-12-13 at 11 30 48

- Do not expose auth.users in a public.<name> view
Screenshot 2024-12-13 at 10 59 17

- When writing RLS policies using auth.<function>(), make sure the calls are wrapped with `()`
Screenshot 2024-12-13 at 10 47 56

- Use security invoker for views
image

- Do not create materialized views in public schema
image

- Do not create foreign tables in public schema 
image

- Extensions should never be installed in the public schema
image


SaxonF (approved):
Looks good :shipit:

PR #31150, opened 2024-12-16
fix(studio): custom schema enum column creation

avallete:

YES

- Bug fix an issue that cause the creation of new columns to fail if the column is an enum type not declared on the public schema by using the `format` property that prefix the enum name with the schema if required instead of `name` on the frontend.

https://github.com/user-attachments/assets/c57b47be-ed8e-478b-bddf-8246fa45d6d5

## What is the new behavior?

https://github.com/user-attachments/assets/ce55ef57-e122-4956-a76e-68e92ca9be06

## Additional context

Should fix: https://app.hubspot.com/live-messages/19953346/inbox/6612091758#comment


avallete:
Just noticed now that a PR was created 8month ago for similar issue. Looping up people involved at the time for review:

https://github.com/supabase/supabase/pull/22796

alaister (approved):
LGTM!

- Tested on preview

PR #31167, opened 2024-12-16
Update TimestampInfo to accept a UTC timestamp by default

jordienr:
Issue: TimestampInfo expects a local timestamp, most of the time (optimally all of the time) we are working with UTC timestamp, this updates the component to always expect a UTC timestamp and renames the value prop to be explicit about it to prevent errors.

Fixes the issue with timestamps in Backups → Restore to new project


joshenlim (approved):
looks great! checked restore to new project + postgres logs, timestamps look good 🙂

PR #31169, opened 2024-12-16
fix: Update env variables to include !

tariknh:

YES

This removes the red squiggly lines you otherwise would get by using the example client createClient() example

Red squiggly lines under the env variables

## What is the new behavior?

No error lines

## Additional context


charislam:
Thanks for opening the issue @tariknh! I'm going to close this because the change is being made to the JavaScript example, and that will annoy users of JavaScript who will see invalid syntax.

If you click through to the TypeScript example, you can copy-paste the TypeScript version, which already has the non-null specifiers!

image

PR #31189, opened 2024-12-17
feat: capture more info from connection string copy

loong:

YES/NO

Captures the language used (e.g. python) and the connection method (e.g. direct, transaction_pooler etc)


joshenlim (approved):
LGTM - Verified on staging preview that new properties lang and connectionMethod are getting sent correctly, existing property connectionType is also getting sent properly

PR #31209, opened 2024-12-18
Chore/toggle for storage client connection s3

joshenlim:
Requires https://github.com/supabase/infrastructure/pull/21003 to be deployed first

Added a toggle under settings/storage > S3 Connection
image

Creation of access keys will be disabled while s3 connection is disabled
image


alaister (approved):
LGTM!

- Tested on preview

PR #31229, opened 2024-12-18
fix(docs): update SvelteKit tutorial to use Svelte 5 syntax

### LINKED ISSUE #31114: Supabase Auth SvelteKit tutorial uses old Svelte syntax
# Improve documentation

## Link 

https://supabase.com/docs/guides/getting-started/tutorials/with-sveltekit

## Describe the problem

It is using old Svelte 4 syntax

## Describe the improvement

It needs to be updated to use Svelte 5 syntax

olegpolin:

YES

Docs update

Tutorial uses Svelte 4 syntax

## What is the new behavior?

Tutorial uses Svelte 5 syntax

## Additional context

Fixes #31114


CoolOppo:
Thank you for taking the time to do this, I'll be using this as my reference. Hope it is reviewed and merged soon.

charislam (approved):
Thanks @olegpolin!

CoolOppo:
I have a question. In `src/routes/layout.ts`, it says, "It's fine to use `getSession` here, because on the client, `getSession` is safe, and on the server, it reads `session` from the `LayoutData`, which safely checked the session using `safeGetSession`."

I'm not sure about how the data loading actually works and couldn't easily find it in the docs how it would work. We return the "unsafe" `session` in the universal load function in `src/routes/layout.ts`, but also return the "safe" `session` in `src/routes/+layout.server.ts`.

Basically my question is: is it guaranteed the returned `session` from the Server Load function in the `+layout.server.ts` always will override the one in the `+layout.ts` and populate the `data` object on the server side properly? Is there anywhere in the docs that specifically says that when you return the same parameter name from both a Universal AND Server load, that the `data` object will always be populated by the one from the one that returned it in the Server load, or is this a bug?

Thanks...could be a simple thing here that's actually correct, but I wanted to make sure because it wasn't easy to find a definitive answer to that question

ChrisChinchilla:
I have updated things to use Svelte 5 in https://github.com/supabase/supabase/pull/37646 so I may need to do some manual changes to this PR, that said I am planning a general overhaul of the auth section, but in the meantime this is some good work to get live.

olegpolin:
Thanks, hopefully this gets merged soon

ChrisChinchilla:
@olegpolin I am still in progress… Early next week it should be done.

ChrisChinchilla:
@olegpolin Done, I will also need team reviews, but please add any feedback you have :)

olegpolin:
Looks good! Just not sure why Supabase docs don't use semicolons in any of the js/ts files

ChrisChinchilla:
@olegpolin I am not sure… But I will add that to the list of things to look into :)

ChrisChinchilla (approved):

charislam (approved):

Issue #15316, opened 2023-06-23
Return a promise for channel subscriptions?


MrAlexLau:
Per [the documentation](https://supabase.com/docs/guides/realtime/broadcast#send-messages) for broadcasting messages, a channel must first be subscribed to before it can call the `send` method. 

However, `channel.subscribe` does not return a promise and only executes its callback. This means that if you're calling this example from the documentation...
```javascript
const channel = supabase.channel('test')

channel.subscribe((status) => {
  if (status === 'SUBSCRIBED') {
    channel.send({
      type: 'broadcast',
      event: 'supa',
      payload: { org: 'supabase' },
    })
  }
})
```

...in a short-lived context like an AWS Lambda function, the process may die before the callback is executed. In this particular case, it would be more natural to say something like:

```javascript
const channel = supabase.channel('test')

await channel.subscribed();

channel.send({
  type: 'broadcast',
  event: 'supa',
  payload: { org: 'supabase' },
})
```

Does that make sense? Am I overlooking existing functionality?

Of course, one option developers have is to write their own promise. I was able to write the following code to accomplish the same task:

```javascript
const channel = supabase.channel("test");

const channelSubscribed = new Promise((resolve, reject) => {
  channel.subscribe((status) => {
    if (status === 'SUBSCRIBED') {
      resolve(channel);
    } else {
       reject("Could not subscribe to channel");
    }
  })
});

await channelSubscribed.then((channel) => {
  channel.send({
    type: 'broadcast',
    event: 'supa',
    payload: { org: 'supabase' },
  })
})
```

_Originally posted by @MrAlexLau in https://github.com/supabase/supabase/discussions/15236_

anudeep652:
Thanks,  @MrAlexLau. I am working on a React Native project, and I wasn't able to broadcast messages. After wrapping `channel.send()` inside a promise, I was able to broadcast it.

encima:
I am not sure it is needed to implement this in the library itself but adding this to the docs might be very useful, please feel free to open a PR!

bcw117:
Does this still need to be added to docs?

Hallidayo:
Hi @bcw117 - Please feel free to open a PR 👍

Hallidayo:
Hi everyone, A a pull request has been merged that fixes the issue I'm going to close this one now.

PR #31232, opened 2024-12-19
docs: add async/await JS example to message broadcasting

kjrocker:

YES

Documentation.

Current Javascript examples use the callback syntax. 

## What is the new behavior?

This introduces at least one example early in the page of using promises and Async/Await.

## Additional context

This is based on https://github.com/supabase/supabase/issues/15316

I'm thinking about creating `subscribeAsync` that would expose a Promise, that might be the better solution long-term. Open to feedback on that.

Credit for the specific example goes to https://github.com/MrAlexLau


charislam (commented):
Thanks @kjrocker!

Definitely agree that it's good to have an async/await example. But it seems weird to create an awaitable async function, just to immediately chain it with `then`. What do you think about altering the example to use async/await throughout?

kjrocker:
Yeah, I could have done that better. Let me know what you think of the new version.

I could also make `channel.subscribeAsync` if we wanted to do this at the library level :smile:

charislam:
Yeah, this seems to make a bit more sense. We should check what the possible statuses are, right now this will throw if you get any status other than `SUBSCRIBED`, but I'm wondering if there's an initializing or pending state that we need to ignore

kjrocker:
Nope, doesn't look like it.

https://github.com/supabase/realtime-js/blob/666ebb796cf97e096118a69d22e79931cc474c1f/src/RealtimeChannel.ts#L104

charislam (approved):
Thanks for checking @kjrocker! Makes sense then, approving ❤️

PR #31237, opened 2024-12-19
Feat/new project creation experiment

SaxonF:
This introduces an experimental project creation flow which allows a user to generate a schema as part of the project creation flow. It also introduces a new design giving greater prominence to the physicality of a database.

We'll be running this as an experiment just to see whether it positively helps people get to the point of connecting to their database.

https://github.com/user-attachments/assets/d0524024-a626-4faa-8978-4372d7d9b360


apps/studio/pages/new/[slug].tsx
```diff
@@ -147,6 +147,12 @@ const Wizard: NextPageWithLayout = () => {
   const { data: organizations, isSuccess: isOrganizationsSuccess } = useOrganizationsQuery()
   const currentOrg = organizations?.find((o: any) => o.slug === slug)
 
+  // TODO: Remove this after project creation experiment
+  const projectCreationExperimentGroup = useFlag<string>('projectCreationExperimentGroup')
+  if (currentOrg && projectCreationExperimentGroup === 'group-b') {
+    router.push(`/new/v2/${currentOrg.slug}`)
```
Comment (alaister) [inline apps/studio/pages/new/[slug].tsx:153]:
```suggestion
    router.replace(`/new/v2/${currentOrg.slug}`)
```
Using `.push()` here breaks the browsers back button. Would also mind popping this inside a `useEffect`? Half nit but more than anything makes me feel weird having right in render 😅

alaister (commented):
I reckon most people will arrive on the `/new` screen via the new project button, so could you also add the flag checking logic there so there's no need to go via the old `/new` page?

One other related small fix:

apps/studio/components/interfaces/ProjectCreation/SchemaGenerator.tsx
```diff
@@ -0,0 +1,207 @@
...
+                } else {
+                  sendEvent({
+                    action: TelemetryActions.PROJECT_CREATION_SECOND_STEP_PROMPT_INTENDED,
+                    properties: {
+                      isNewPrompt,
+                    },
+                  })
+                }
+                setPromptIntendSent(true)
+              }
```
loong, line 184: This is a bit whack way of doing it, so anyone feel free to change it if you have a better idea.

What this basically does that it checks each time a key is pressed in the form if the prompt length has at least 5 characters. If yes, it will send the event

loong:
@pamelachia Added a bunch of events, can you check the [telemetry.ts](https://github.com/supabase/supabase/pull/31237/files#diff-192e9c02bc700d99aa1f1a5ec3e761ae3235b23dc760207e6d55e97c40be9322) and see if the documentation is clear to you

pamelachia (approved):

apps/studio/components/interfaces/Database/Schemas/Schemas.utils.ts
```diff
@@ -12,14 +12,14 @@ const NODE_SEP = 25
...
-  tables: PostgresTable[]
+  ref?: string,
+  schema?: PostgresSchema,
+  tables?: PostgresTable[]
 ): Promise<{
   nodes: Node<TableNodeData>[]
   edges: Edge[]
 }> {
-  if (!tables.length) {
+  if (!tables?.length) {
```
joshenlim, line 22: this might cause a client error if tables is undefined (`undefined.length`)
initialize table as an empty array?

apps/studio/components/interfaces/Database/Schemas/Schemas.utils.ts
```diff
@@ -12,14 +12,14 @@ const NODE_SEP = 25
 const RANK_SEP = 50
 
 export async function getGraphDataFromTables(
-  ref: string,
-  schema: PostgresSchema,
-  tables: PostgresTable[]
+  ref?: string,
```
joshenlim, line 15: i might find the answer as i check through down, but any reason why these need to be optional?

apps/studio/components/interfaces/ProjectCreation/InitialStep.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onStartBlank: () => void
+  onMigrate: () => void
+  onSqlGenerated: (sql: string) => void
+  onServicesUpdated: (services: SupabaseService[]) => void
+  onTitleUpdated: (title: string) => void
+  isLoading: boolean
+}) => {
+  const [value, setValue] = useState('')
+
+  const handleSubmit = (e: React.FormEvent) => {
```
joshenlim, line 30: there's some unused variables here - will help clean them up along with unused imports as i go through

apps/studio/components/interfaces/ProjectCreation/InitialStep.tsx
joshenlim, line 1: is this file being used? i don't see it being imported anywhere

apps/studio/components/interfaces/ProjectCreation/ProjectVisual.tsx
```diff
@@ -0,0 +1,208 @@
...
+  instanceLabel: (size: string) => string
+}
+
+export const ProjectVisual = ({
+  sqlStatements,
+  showInfo = true,
+  services,
+  selectedRegion,
+  projectDetails,
+  instanceLabel,
```
joshenlim, line 43: unused param - will clean up

apps/studio/components/interfaces/ProjectCreation/SchemaFlow.tsx
```diff
@@ -0,0 +1,139 @@
...
+ * Component to detect React Flow container resizes.
+ * Calls `onResize` when `width` or `height` changes.
+ *
+ * Debounces at 200ms by default.
+ *
+ * It's better to use this child component instead of the
+ * `useOnResize` hook directly in order to prevent a large
+ * amount of re-renders on the main component.
+ */
+function ResizeHandler({ onResize, debounce = 200 }: ResizeHandlerProps) {
```
joshenlim, line 119: unused function, will clean up

apps/studio/components/interfaces/ProjectCreation/SchemaFlow.tsx
```diff
@@ -0,0 +1,139 @@
...
+  useOnResize(onResize, debounce)
+
+  return null
+}
+
+export function useDebounce<T>(value: T, delay: number) {
```
joshenlim, line 125: any reason why we can't use lodash's debounce? 
or if this is better then we should shift this to lib/helper and update everywhere else accordingly

apps/studio/components/interfaces/ProjectCreation/SchemaFlow.tsx
```diff
@@ -0,0 +1,139 @@
...
+          className="[&>*]:stroke-foreground-muted opacity-[25%]"
+          variant={BackgroundVariant.Dots}
+          color={'inherit'}
+        />
+      </ReactFlow>
+    </div>
+  )
+}
+
+function useOnResize(fn: () => void, debounce = 200) {
```
joshenlim, line 91: unused function, will clean up

SaxonF: yes sorry, we can remove all that including useDebounce which is unused

apps/studio/components/interfaces/SchemaVisualizer/index.tsx
joshenlim, line 1: in general for this file - for where we're console.erroring, should we show a toast instead so that the errors don't fail silently?

apps/studio/pages/new/v2/[slug].tsx
```diff
@@ -0,0 +1,1211 @@
...
+  const hasMembersExceedingFreeTierLimit = (membersExceededLimit || []).length > 0
+
+  const showNonProdFields = process.env.NEXT_PUBLIC_ENVIRONMENT !== 'prod'
+
+  const freePlanWithExceedingLimits =
+    orgSubscription?.plan?.id === 'free' && hasMembersExceedingFreeTierLimit
+
+  const isManagedByVercel = currentOrg?.managed_by === 'vercel-marketplace'
+
+  const canCreateProject =
```
Comment (joshenlim) [inline apps/studio/pages/new/v2/[slug].tsx:233]: unused variable - although i think this should be checked?

apps/studio/pages/new/v2/[slug].tsx
```diff
@@ -0,0 +1,1211 @@
...
+      refetchOnMount: false,
+      refetchOnWindowFocus: false,
+      refetchInterval: false,
+    }
+  )
+
+  const {
+    mutate: createProject,
+    isLoading: isCreatingNewProject,
+    isSuccess: isSuccessNewProject,
```
Comment (joshenlim) [inline apps/studio/pages/new/v2/[slug].tsx:212]: unused variables, although to double check if we're supposed to be using these for any UI responses/transitions?

apps/studio/components/interfaces/ProjectCreation/InitialStep.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onTitleUpdated,
+  isLoading,
+}: {
+  onSubmit: (value: string) => void
+  onStartBlank: () => void
+  onMigrate: () => void
+  onSqlGenerated: (sql: string) => void
+  onServicesUpdated: (services: SupabaseService[]) => void
+  onTitleUpdated: (title: string) => void
+  isLoading: boolean
```
joshenlim, line 26: is this variable meant to be used?

apps/studio/pages/new/v2/[slug].tsx
Comment (joshenlim) [inline apps/studio/pages/new/v2/[slug].tsx:1]: this is a really huge file - can we break this into smaller components to improve readability and maintainability?

packages/shared-data/regions.ts
joshenlim, line 1: we also have a copy of lon lats [here](https://github.com/supabase/supabase/blob/feat/new-project-creation-experiment/apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureConfiguration/InstanceConfiguration.constants.ts#L30)
- can use to double check correctness (if its the same units)
- can we de-dupe accordingly? happy to use this and remove the AWS_REGIONS_COORDINATES variable

SaxonF: I wish I knew that before looking up long lat for each haha. yep lets use existing

joshenlim:
## Tested:
- Create from scratch flow
- Migrate from existing database flow

## Bugs founds:
- No proper loading state when "Create project" was clicked, i ended up clicking multiple times which created multiple projects

- Create from scratch flow and Migrate existing database flow feels exactly the same - the latter seems to be missing any migration action?

- Minor: References to external tables should be de-duped here. Also, for someone completely new to Supabase, i think i'd be wondering what "auth.user.id" is (i wouldn't have context what are the schemas created by Supabase at this point)
image

- Globe is re-rendering each time an input field is updated

https://github.com/user-attachments/assets/7edb66ce-56f4-4156-986a-a1b7a332d1a6

- Selecting a different organization here, redirects me back to "Start blank" / "Migrate existing" which i found the UX confusing 
image

- If i submit a prompt, i get redirected to the "create project" state. If i then were to go back to the previous state (e.g from the current bug where if i select a new organization i get redirected back, OR if i created the project, then go to create a new project again), the assistant response doesn't clear when i go back to the main page. If i submit a new prompt from here, i get redirected to the "create project" state, but theres no generated schema then.
image


## Some feedback from testing on preview:

- Should we have some UI to navigate back to the initial view where i can choose "Start blank" or "Migrate existing"? Otherwise if i hit the Supabase logo, i'm redirected back to the projects page which the UX felt confusing
image

- If im on an organization that I don't have permissions to create projects for, I'm unable to change my organization while on this page 
image

- Unless maybe i can select my organization here?
image

- The globe's green dot for the selected region (which im presuming is the bigger green dot here) doesn't stand out as much as the other dots on the globe, so I didn't really know what I was looking at
image

- Is there no way to minimize the advanced settings after I've clicked on it? 
image

- It wasn't clear that i needed to hit enter to generate the schemas here or see any output - I was wondering if the tables will be created after the project was created or something which didn't give me the confidence of a proper starting point. Maybe the Up arrow icon should be an enter icon?
image

- I feel like including "Cron" and "queues" as our product offering feels weird - our main products are DB, auth, storage, edge functions and realtime, cron and queues are part of the database, not their own products?
image

SaxonF:
@joshenlim I've resolved most of these (I hope). Some of the small style ones could be follow up's. 

> Create from scratch flow and Migrate existing database flow feels exactly the same - the latter seems to be missing any migration action?

Right now there is nothing in onboarding that changes. This leads into homepage redesign where we can start to guide users based on what they chose during project creation.

> Selecting a different organization here, redirects me back to "Start blank" / "Migrate existing" which i found the UX confusing

I agree this is annoying. Could be a follow up PR? The fix is not so simple.

joshenlim:
> Could be a follow up PR? The fix is not so simple.

Yeap that's all good no worries since its not user facing atm. I did notice that this behaviour only happens when selecting an organization that's not been selected before if it might help to zoom into the problem. (e.g if i initially am creating a project in Org A, and then i select Org B, i get thrown back to the first step. But if i then go back into the create project page and select Org A, that's fine then i don't get thrown back)

joshenlim:
Minimally verified that all the bugs identified have now been resolved (less the selecting org getting redirected one)

Will do one last code sweep tomorrow to double check and we should be good to go 🙂

apps/studio/components/interfaces/ProjectCreation/InitialStep.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+export const InitialStep = ({
+  onSubmit,
+  onStartBlank,
+  onMigrate,
+  onSqlGenerated,
+  onServicesUpdated,
+  onTitleUpdated,
+  isLoading,
```
joshenlim, line 18: double check on this one as this param is not being used atm

apps/studio/components/interfaces/ProjectCreation/ProjectVisual.tsx
```diff
@@ -0,0 +1,233 @@
...
+}
+
+export const ProjectVisual = memo(
+  ({
+    sqlStatements,
+    showInfo = true,
+    services,
+    selectedRegion,
+    projectDetails,
+    instanceLabel,
```
joshenlim, line 58: double check on this one as this param is not being used

apps/studio/components/interfaces/ProjectCreation/SchemaFlow.tsx
```diff
@@ -0,0 +1,102 @@
...
+          className="[&>*]:stroke-foreground-muted opacity-[25%]"
+          variant={BackgroundVariant.Dots}
+          color={'inherit'}
+        />
+      </ReactFlow>
+    </div>
+  )
+}
+
+export function useDebounce<T>(value: T, delay: number) {
```
joshenlim, line 88: double check on this one - any reason why we can't use lodash's debounce?
or if this is better then we should shift this to lib/helper and update everywhere else accordingly

apps/studio/components/interfaces/ProjectCreation/SchemaGenerator.tsx
```diff
@@ -0,0 +1,215 @@
+import { useChat } from 'ai/react'
+import { ArrowUp } from 'lucide-react'
```
joshenlim, line 2: unused import here

apps/studio/components/interfaces/SchemaVisualizer/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+          raw_app_meta_data jsonb,
+          raw_user_meta_data jsonb,
+          is_super_admin boolean,
+          created_at timestamp with time zone,
+          updated_at timestamp with time zone
+      );
+      ALTER TABLE ONLY auth.users
+      ADD CONSTRAINT users_email_key UNIQUE (email);
+      ALTER TABLE ONLY auth.users
+      ADD CONSTRAINT users_pkey PRIMARY KEY (id);
```
joshenlim, line 57: this seems duplicate with below at L75 - could we pull this out into a const?
okay for this to be done in another PR

joshenlim (approved):
All good now! 😄

PR #31259, opened 2024-12-19
fix: increase snippet limit within folders

alaister:

Bug fix

Loads 3 snippets within a folder at a time (leftover from testing).

## What is the new behavior?

Loads 100


joshenlim (approved):
LGTM - tested on preview

pushed an unrelated fix where the move query modal is missing a word "move n _snippets_"
image

PR #31271, opened 2024-12-20
Support bulk deletes in auth users

joshenlim:
image


apps/studio/components/interfaces/Auth/Users/InviteUserModal.tsx
```diff
@@ -44,52 +44,49 @@ const InviteUserModal = ({ visible, setVisible }: InviteUserModalProps) => {
   }
 
   return (
-    <div>
```
joshenlim, line 47: just removed an unnecessary div here

alaister (commented):
If you delete all your users at once, the side panel seems to stay open

Screenshot 2024-12-20 at 21 18 08

joshenlim:
Oh good catch! Should close the panel if the user thats in view is deleted 🙏

joshenlim:
@alaister just pushed a fix up! 🙂 🙏

alaister (approved):
LGTM!

- Tested on preview

alaister (approved):

PR #31301, opened 2024-12-23
Fix camelcase enum types in column editor

### LINKED ISSUE #31260: Uppercase enums do not work
Enums with an uppercased name do not work in the dashboard, but lowercase ones do.

joshenlim:
Resolves https://github.com/supabase/supabase/issues/31260

ColumnEditor has been using the `format` property from `enumTypes` which adds escaped double quotes in the enum name, if the enum is camelcased. Fix is to use the `name` property instead (i actually can't recall why we were using format tbh, but name seems more logical and correct)

Before:
image

After: 
image


joshenlim:
Closing in favor of https://github.com/supabase/supabase/pull/31303

PR #31303, opened 2024-12-23
Fix using camel case enum from public schema erroring out in ColumnEditor

### LINKED ISSUE #31260: Uppercase enums do not work

joshenlim:
Resolves https://github.com/supabase/supabase/issues/31260

- Removed escaped double quotes for camel cased enums when displaying them in the UI
- Stripe double quotes for enum types in the public schema only

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnType.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnType.tsx
@@ -1,7 +1,19 @@
 import * as Tooltip from '@radix-ui/react-tooltip'
 import { noop } from 'lodash'
+import {
+  Calendar,
+  Check,
+  ChevronsUpDown,
+  ExternalLink,
+  Hash,
+  ListPlus,
+  ToggleRight,
+  Type,
+} from 'lucide-react'
 import Link from 'next/link'
 import { ReactNode, useState } from 'react'
+
+import type { EnumeratedType } from 'data/enumerated-types/enumerated-types-query'
 import {
   AlertDescription_Shadcn_,
   AlertTitle_Shadcn_,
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnType.tsx
```diff
@@ -67,8 +67,11 @@ const ColumnType = ({
...
 }: ColumnTypeProps) => {
   const [open, setOpen] = useState(false)
-  // @ts-ignore
-  const availableTypes = POSTGRES_DATA_TYPES.concat(enumTypes.map((type) => type.format))
+  const availableTypes = POSTGRES_DATA_TYPES.concat(
+    enumTypes.map((type) =>
+      type.schema === 'public' ? type.format.replaceAll('"', '') : type.format
+    )
+  )
   const isAvailableType = value ? availableTypes.includes(value) : true
```
alaister, line 75: Just to confirm, this `value` doesn't need a `.replaceAll('"', '')` as well?

joshenlim: value here should be okay since its a passed prop
- When selecting a new column type, we'd "sanitize" it when we selected the option, so value here will have the double quotes stripped already if needed
- When viewing a column's type though, i noticed that if the enum is not from the public schema, the user will not be able to edit the column, but i feel like this needs to be fixed on pg-meta's end? it seems like the type of the column does not contain the type's schema
image

joshenlim: ^ this bug is an existing issue though, so probably can address separately

joshenlim: hmm one sec - i noticed this can be fixed from dashboard, not pgmeta

joshenlim: okay updated this, should be fixed now 😄

alaister (approved):
LGTM!

- Tested on preview

PR #31305, opened 2024-12-23
Add last and next run in CronJobCard

joshenlim:
image
image


apps/studio/components/interfaces/Integrations/CronJobs/CronJobRow.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const { mutate: sendEvent } = useSendEventMutation()
+  const { mutate: toggleDatabaseCronJob, isLoading } = useDatabaseCronJobToggleMutation()
+
+  // pg_cron can also use "30 seconds" format for schedule. Cronstrue doesn't understand that format so just use the
+  // original schedule when cronstrue throws
+  let schedule = job.schedule
+  try {
+    const scheduleString = CronToString(job.schedule)
+    schedule = scheduleString
+  } catch {}
```
jordienr, line 46: I think it'd be cleaner to move this to its own function outside the component but it's clear what it does so no blocking

function safeCronToString(schedule: string) → trycatch → returns string, wdyt?

jordienr (approved):
Tested in preview, no errors ✅

apps/studio/components/interfaces/Integrations/CronJobs/CronJobRow.tsx
joshenlim: oh yes! good point 🙏

PR #31306, opened 2024-12-23
Fix API docs introction

joshenlim:
Adds some content for the API introduction page, referenced from our docs site to fix the alignment

Before:
image

After:
image


jordienr (approved):
tested in preview lgtm

PR #31311, opened 2024-12-23
fillTimeseries: Skip filling if theres a minimum of datapoints

jordienr:

YES

fillTimeseries is a function that adds datapoints with some default values, if we have enough data, we dont need to use it, this change allows to set the minimum of data points to check before filling.

https://github.com/user-attachments/assets/5dab9bb7-4656-49c7-8705-f54de4dd498a


alaister (approved):
LGTM!

- Tested on preview

PR #31390, opened 2024-12-26
fix: nodejs and golang direct connection strings

alaister:

Bug fix

Node.js and Go showing the pooler connection string instead of the direct connection string.

## What is the new behavior?

Properly showing the direct connection string.


joshenlim (approved):
LGTM! Verified on preview for Node and Go Connect UI

PR #31615, opened 2024-12-26
Update react user management tutorial docs

matthewliuswims:

YES

Docs change

People get 2 errors that they can google and follow the solution here: https://stackoverflow.com/a/74468038/11303031. But ideally, [the guide](https://supabase.com/docs/guides/getting-started/tutorials/with-react) should "just work" when someone follows it. Example of an error is below.

Screenshot 2024-12-26 at 3 03 40 PM

## What is the new behavior?

Tell the user they need to actually 1) create the bucket 2) create the policies before it'll work

### DIFF
--- a/apps/docs/content/guides/getting-started/tutorials/with-react.mdx
+++ b/apps/docs/content/guides/getting-started/tutorials/with-react.mdx
@@ -396,4 +396,4 @@ return (
 )
 ```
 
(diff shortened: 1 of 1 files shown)


matthewliuswims:
closed in favor of the above https://github.com/supabase/supabase/pull/31616

PR #31616, opened 2024-12-26
update docs for user management tutorial

matthewliuswims:

YES

Docs change

When following the tutorial word-by-word, people will get 2 errors (one that the bucket doesn't exist and 2 that there is a row level security issue). People can google and follow the solution here: https://stackoverflow.com/a/74468038/11303031. But ideally, [the guide](https://supabase.com/docs/guides/getting-started/tutorials/with-react) should "just work" when someone follows it. Example of an error is below.

Screenshot 2024-12-26 at 3 03 40 PM

## What is the new behavior?

Tell the user they need to actually 1) create the bucket 2) create the policies before it'll work

### DIFF
--- a/apps/docs/content/guides/getting-started/tutorials/with-angular.mdx
+++ b/apps/docs/content/guides/getting-started/tutorials/with-angular.mdx
@@ -525,4 +525,4 @@ export class AccountComponent implements OnInit {
 }
 ```
 
-At this stage you have a fully functional application!
+Now follow the [docs](https://supabase.com/docs/guides/storage/quickstart#add-security-rules) to create a new bucket called `avatars` and add policies to allow authenticated users to write and read to the bucket. With that, you should have a fully functional application!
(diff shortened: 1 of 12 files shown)


matthewliuswims:
nvm -- this should be taken care of in the database schema part of the tutorial. 
```
-- Set up Storage!
insert into storage.buckets (id, name)
  values ('avatars', 'avatars');

-- Set up access controls for storage.
-- See https://supabase.com/docs/guides/storage/security/access-control#policy-examples for more details.
create policy "Avatar images are publicly accessible." on storage.objects
  for select using (bucket_id = 'avatars');

create policy "Anyone can upload an avatar." on storage.objects
  for insert with check (bucket_id = 'avatars');

create policy "Anyone can update their own avatar." on storage.objects
  for update using ((select auth.uid()) = owner) with check (bucket_id = 'ava
  ```
Not sure why it didn't work for me though...

PR #31699, opened 2024-12-27
docs(pgmq): clarify wording around reading messages

bstncartwright:

YES

The original documentation for the `read` operation in PGMQ made it sound like the `vt` parameter specified _how long until the message became_ invisible, when in reality it is how long the message _is_ invisible. See: https://github.com/tembo-io/pgmq/tree/main?tab=readme-ov-file#read-messages
> Read 2 message from the queue. Make them invisible for 30 seconds. If the messages are not deleted or archived within 30 seconds, they will become visible again and can be read by another consumer.

## Additional context


charislam (approved):
Thanks @bstncartwright!

PR #32505, opened 2025-01-01
fix(dashboard): Improve scrollbar position on Table Editor list

kylerummens:

YES

Bug fix

The existing padding of the list of tables/views makes it so that the scrollbar is inside the padding, making the scrollbar cover the content of the list items.

## What is the new behavior?

The scrollbar exists outside of the padding and not cover the list item content. The overall spacing and alignment of the list items is preserved. Also, the `InfiniteList` component was not altered, so no other places in the dashboard are negatively affected by this change.

## Additional context

The diff for the `EntityListItem.tsx` file looks large, but all I did was wrap the `<EditorTablePageLink>` component in a `<div className="px-2">`

Before:
Screenshot 2024-12-31 at 7 34 33 PM

After:
Screenshot 2024-12-31 at 7 18 06 PM

### DIFF
--- a/apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
+++ b/apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
@@ -168,7 +168,7 @@ const TableEditorMenu = () => {
             )}
           </div>
         </div>
-        <div className="flex flex-auto flex-col gap-2 pb-4 px-2">
+        <div className="flex flex-auto flex-col gap-2 px-2">
           <InnerSideBarFilters>
             <InnerSideBarFilterSearchInput
               autoFocus={!isMobile}
@@ -269,7 +269,7 @@ const TableEditorMenu = () => {
                 />
               )}
               {(entityTypes?.length ?? 0) > 0 && (
-                <div className="flex flex-1" data-testid="tables-list">
+                <div className="flex flex-1 -mx-2" data-testid="tables-list">
                   <InfiniteList
                     items={entityTypes}
                     ItemComponent={EntityListItem}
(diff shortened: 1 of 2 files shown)


apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
```diff
@@ -168,7 +168,7 @@ const TableEditorMenu = () => {
             )}
           </div>
         </div>
-        <div className="flex flex-auto flex-col gap-2 pb-4 px-2">
+        <div className="flex flex-auto flex-col gap-2 px-2">
```
alaister, line 171: Any reason you've removed `pb-4` here? Now the last table is very close to the bottom of the browser.

alaister (commented):
Thanks @kylerummens!

This looks good, just one quick question:

apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
kylerummens: The padding on the bottom seemed off to me with the scroll as well. If you like, I can add it back.

alaister: What do you mean by "seemed off"? Adding it back would be great or if you can think of a better way that won't seem off that would be great 🙏

kylerummens: Just the fact that the scrollbar didn't extend to the bottom of the container. It appeared that the only other way to fix that would have been to add that padding in the InfiniteList component, which would have affected more places in the app.

kylerummens: I just added that padding back per your request. Thanks

alaister (approved):
LGTM! Thanks, @kylerummens!

PR #32524, opened 2025-01-02
fix: Add year to the footer

ivasilov:
Found a sane way to add the current year to the footer instead of manual updates each year.

Screenshot 2025-01-02 at 15 25 21


ivasilov:
This has been marked as "not a priority" for this year. Maybe next year.

PR #32542, opened 2025-01-03
chore(www): reduce dashboard button layout shift

alaister:

chore

No buttons are displayed until the user state has finished loading. This causes quite a big shift once the buttons load in.

## What is the new behavior?

The regular sign in / start your project buttons are shown while the user state is loading. There will still be some layout shift as the sizes are different, but I think it's better than the layout shift caused by starting from nothing.


saltcod (approved):

fsansalvadore (approved):
Nice!

PR #32549, opened 2025-01-03
chore: remove LW8 animation

kevcodez:
To avoid continuous maintenance burden for upgrades and dependencies and also slower build times / installs, this PR removes the animated "8" and gets rid of a couple of dependencies.


fsansalvadore (approved):
Good enough!

PR #32550, opened 2025-01-03
chore: Expand prettier commands to also run on packages

ivasilov:


saltcod (approved):
Checked all previews & looks good!

Issue #23618, opened 2024-05-02
Add an OAuth example to the new SvelteKit SSR documentation


coppinger:
# Improve documentation

https://supabase.com/docs/guides/auth/server-side/sveltekit

https://github.com/supabase/supabase/blob/master/apps/docs/content/guides/auth/server-side/sveltekit.mdx

## Describe the problem

I found it slightly difficult to figure out how to use OAuth with the new SvelteKit SSR guide as there isn't an example of how to handle it—seemingly by redirecting to data.url

ex:
```
		if (error) {
			console.error(error);
			return redirect(303, '/auth/error');
		} else {
			return redirect(303, data.url);
		}
```

## Describe the improvement

It'd be great to address this in the guide. I tried to make a PR, but couldn't quite figure out how best to approach adding this info.

Aishwary2004Gupta:
Hey, can I work on this issue?

charislam:
@Aishwary2004Gupta, of course! We don't assign issues, so feel free to open a PR if no one else has gotten there first! 😄

ShubhSingh07:
Hi, I'd like to work on this issue. Could you please assign it to me?

melaksenay:
> Hi, I'd like to work on this issue. Could you please assign it to me?

For future reference, there is no assigning people to issues! You can work on them as you please and just make sure to let others know you are working on a given issue by commenting!

sarthaksharma27:
Chat! I am working on this issue #23618

Khan4218:
Hi! I’d like to work on this issue. Can I get assigned?

nikhil008-git:
Heyy, can i work on this issue?

dhunganapramod9:
Hey! i am working on this

aparajitaks:
Hi! I’d like to work on this issue.

I can add an OAuth example for the new SvelteKit SSR documentation. I’ll prepare a clear, minimal example and open a PR once it’s ready.

Please let me know if there are any specific requirements or preferred formats for documentation examples.

RudraKo:
Hello! I’d like to take this issue.

I can implement the OAuth example for the updated SvelteKit SSR docs. I’ll follow the existing documentation structure and open a PR soon.

If there are any guidelines I should follow for doc contributions, I’d be happy to review them.

harikapadia999:
Hi! I'd like to work on this documentation issue.

**Plan:**
I'll add a comprehensive OAuth example to the SvelteKit SSR documentation covering:
[...]

PR #32575, opened 2025-01-05
code now uses signInWithOAuth as per the SDK recommendations

Aishwary2004Gupta:

YES

docs update

The SvelteKit SSR guide does not include an example of how to handle OAuth redirection for error handling and successful login scenarios. This has been noted as an area of improvement in [issue #23618](https://github.com/supabase/supabase/issues/23618).

## What is the new behavior?

The SvelteKit SSR guide now includes an example demonstrating how to handle OAuth errors and successful logins by using redirect(303, data.url) for successful authentication and redirect(303, '/auth/error') for error handling. This addition improves clarity for developers integrating OAuth with Supabase in SvelteKit projects.

docs improvement

## Additional context

This addition aims to simplify the developer experience for handling OAuth in Supabase when using SvelteKit. This change addresses the concern raised in issue #23618. No breaking changes have been introduced.


apps/docs/content/guides/auth/server-side/sveltekit.mdx
```diff
@@ -619,8 +619,30 @@ export const load: PageServerLoad = async ({ depends, locals: { supabase } }) =>
...
+### Handling OAuth in SSR
+
+When using OAuth with SvelteKit SSR, you need to handle redirects for successful or failed authentication. 
+Here’s an example:
+
+```javascript
+import { redirect } from '@sveltejs/kit';
+
+export const load = async ({ fetch }) => {
+    const res = await fetch('/auth/oauth');
```
charislam, line 631: Curious about this line: in general our recommendation is to use the `signInWithOAuth` function in the [SDK](https://supabase.com/docs/reference/javascript/auth-signinwithoauth) to implement OAuth. What's the use case for your example?

charislam (commented):
Thanks @Aishwary2004Gupta! Just a quick clarification question below!

apps/docs/content/guides/auth/server-side/sveltekit.mdx
Aishwary2004Gupta: The example provided here is meant to show how to handle OAuth flows in scenarios where you may want to delegate the OAuth redirection and callback handling to a server-side endpoint (/auth/oauth).

Aishwary2004Gupta: I could update the example to align with the SDK’s best practices if you'd like!

charislam: Ahhh, I see. In that case, the API would be up to the dev creating the custom endpoint, though, so let's just use the SDK example, and anyone creating a custom server-side endpoint should be able to build on top of that

Aishwary2004Gupta: Thank you for the feedback! I've updated the example to align with the Supabase SDK's recommendation to use signInWithOAuth. 

updated code

j4w8n:
@Aishwary2004Gupta have you created your own demo app with SvelteKit, OAuth, and Supabase's ssr library?

Aishwary2004Gupta:
@j4w8n yes I have

j4w8n:
> @j4w8n yes I have

Ok. I wanted to check because the code is importing a function from @supabase/supabase-js that doesn't exist. Plus you'd wanna import from the ssr library, to create a client. Or at least that's been my experience.

Aishwary2004Gupta:
Could you review my changes. I've updated the example to align with the Supabase SDK's recommendation to use signInWithOAuth.

j4w8n:
> Could you review my changes. I've updated the example to align with the Supabase SDK's recommendation to use signInWithOAuth.

The changes from several hours ago are the latest I see. I'll make some comments within the files.

j4w8n:
On a sidenote, I'm not much of a git expert, but does anyone know why this PR is now showing over 74k diffs?

apps/docs/content/guides/auth/server-side/sveltekit.mdx
```diff
@@ -625,18 +625,32 @@ When using OAuth with SvelteKit SSR, you need to handle redirects for successful
 Here’s an example:
 
 ```javascript
+import { createSupabaseClient } from '@supabase/supabase-js';
```
j4w8n, line 628: createSupabaseClient should be createClient. But, like I said, you should be using auth-related functions from @supabase/ssr for server side auth, not @supabase/supabase-js

apps/docs/content/guides/auth/server-side/sveltekit.mdx
```diff
@@ -625,18 +625,32 @@ When using OAuth with SvelteKit SSR, you need to handle redirects for successful
...
-export const load = async ({ fetch }) => {
-    const res = await fetch('/auth/oauth');
-    const data = await res.json();
+// Initialize the Supabase client
+const supabase = createSupabaseClient(
+    'https://your-supabase-project-url.supabase.co',
+    'your-anon-key'
+);
+
+export const load = async ({ url }) => {
```
j4w8n, line 637: I'm not sure how everyone else does it, but I thought the "SvelteKit way" would be to have a page form that calls an action; therefore the `signInWithOAuth` code would be in an action not a load function?

j4w8n: Then, if that's the case, you wouldn't need to create a supabase client in this file either, but rather destructure it from the action's `event`.

Aishwary2004Gupta (commented):
I've updated the code to use createClient and ensured that the @supabase/ssr package is used to properly align with server-side rendering best practices, also I've moved the OAuth logic to the action handler. This keeps the load function separate and ensures the SvelteKit convention of handling side effects via actions.

apps/docs/content/guides/auth/server-side/sveltekit.mdx
Aishwary2004Gupta: updated the code to use createClient and ensured that the @supabase/ssr package is used to properly align with server-side rendering best practices.

apps/docs/content/guides/auth/server-side/sveltekit.mdx
Aishwary2004Gupta: So I've moved the OAuth logic to the action handler. This keeps the load function separate and ensures the SvelteKit convention of handling side effects via actions. It should now be cleaner and more aligned with best practices.

j4w8n:
I'm still seeing some issues with this implementation.

And again, why are there so many non-related diff changes?

PR #32592, opened 2025-01-06
Fix next run data for cron job not using UTC

joshenlim:
"Next run" currently generates a timestamp (using cron-parser) that's in the local timezone already, whereas it should be returned as UTC since cron jobs are all in UTC TZ (Refer to screenshot under "Before")

Before:
image

After:
image


apps/studio/components/interfaces/Integrations/CronJobs/CronJobCard.tsx
```diff
@@ -47,6 +47,7 @@ export const CronJobCard = ({ job, onEditCronJob, onDeleteCronJob }: CronJobCard
   })
   const lastRun = data?.start_time ? dayjs(data.start_time).valueOf() : undefined
   const nextRun = getNextRun(job.schedule, data?.start_time)
+  console.log(job.jobname, { nextRun: dayjs(nextRun).format('DD MMM YYYY HH:mm:ss (ZZ)') })
```
alaister, line 50: cheeky
```suggestion
```

alaister (approved):
LGTM!

- Tested on preview

PR #32594, opened 2025-01-06
fix(docs): accepting consent

alaister:


joshenlim (approved):
Tested across www studio and docs previews that accepting telemetry works

although there seems to be 2 POST requests to /page each time we accept

PR #32595, opened 2025-01-06
feat: enforce event type defs

loong:

YES

From now on, if any of the events do not comply to the event definitions in telemetry.ts, it will cause a type error during building time.


joshenlim (approved):
Changes LGTM! Minimally verified logging in + checking POST event when submitting a prompt via the assistant

pamelachia (approved):
tested preview sends 5 different events successfully. No sendEvent action that uses string.

PR #32598, opened 2025-01-06
Persist last selected database only for sql editor via local storage

joshenlim:


joshenlim:
ahh sorry i was missing a useEffect hence why - have pushed a fix up! 🙏

alaister (approved):
LGTM!

- Tested on preview

PR #32603, opened 2025-01-06
Fix impersonation and searching users escape single quote

joshenlim:
Escape single quotes in SQL for impersonation and searching users


alaister (approved):
LGTM!

- Tested on preview

PR #32605, opened 2025-01-06
docs: update reset script copy

Ziinc:
Updates the reset.sh copy and adds in more details about the .env deletion

### DIFF
--- a/docker/reset.sh
+++ b/docker/reset.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-echo "WARNING: This will remove all containers and their data. This action cannot be undone!"
+echo "WARNING: This will remove all containers and container data, and will reset the .env file. This action cannot be undone!"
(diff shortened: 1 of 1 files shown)


kiwicopple (approved):

docker/reset.sh
```diff
@@ -22,21 +22,23 @@ for DIR in "${BIND_MOUNTS[@]}"; do
...
+    echo "Directory $DIR does not exist. Skipping bind mount deletion step..."
   fi
 done
 
 echo "Resetting .env file..."
 if [ -f ".env" ]; then
   echo "Removing existing .env file..."
   rm -f .env
+else
+  echo "No .env file found. Skipping .env removal step..."
```
avallete, line 34: **suggestion**

Maybe instead of removing it, we could `mv .env .env.old` so it isn't destructive 🤔 

I can see an issue where this wouldn't be properly gitignored and accidentally comitted tough.

avallete (approved):

PR #32612, opened 2025-01-07
fix: do not block project/replica creation with due invoices on team/enterprise

kevcodez:
Just setting `enabled` flag on react-query is not sufficient, as the data may be preloaded already. To ensure we're not blocking team/enterprise customers with outstanding invoices from launching new projects or replicas, we have to add an additional check.

Backend is already validating this properly, only a frontend issue.


alaister (approved):
LGTM!

- Tested on preview

PR #32614, opened 2025-01-07
feat: add Hono example and guide.

thorwebdev:


apps/docs/content/guides/getting-started/quickstarts/hono.mdx
```diff
@@ -0,0 +1,95 @@
+---
```
dshukertjr, line 1: I'm wondering if we should have something more than just calling the bootstrap command. The bootstrap command is awesome, but following this guide doesn't really teach anything about how to use Supabase with Hono.

dshukertjr (approved):
Nice one!

I did leave one commend about the getting start guide, but I think it's okay to merge for now and think about improving it later on!

PR #32617, opened 2025-01-07
fix: deprecation comments

kevcodez:
Using `//` deprecations are not properly recognized as JSDoc doesn't work like that.


joshenlim (approved):
code wise LGTM! only comment changes so should be okay

PR #32620, opened 2025-01-07
fix: mfa phone alert fixed to show up only if phone is not configured

hf:
MFA with Phone now supports all phone providers.


joshenlim (approved):
LGTM, verified changes on staging - although the callout should show up when the option is selected, not after save, will look into this separately

PR #32622, opened 2025-01-07
TimestampInfo: Prevent onMouseDown and onPointerDown

jordienr:
- Fixes issue where the onMouseDown or onPointerDown events will trigger parent element events when used inside Dropdowns.

## Before
CleanShot 2025-01-07 at 13 15 07

## After
CleanShot 2025-01-07 at 13 14 19


joshenlim (approved):

PR #32632, opened 2025-01-07
r2np: new cost

jordienr:

YES

- Updates the cost estimate for the R2NP projects to use the target project specs, since the specs can be slightly different than the current project

- Shows cost of compute based on the current project instead of the new project that will be created.

## What is the new behavior?

- Shows cost of compute (same as current project)
- Shows cost of new disk size (its a bit higher than current project)
- Shows total

## Additional context

CleanShot 2025-01-08 at 13 41 06@2x


alaister (approved):
LGTM!

- Tested on preview

PR #32633, opened 2025-01-07
fix: swift ref docs filter params

### LINKED ISSUE #31354: Incorrect Type Signature in Swift Documentation for Filter Method
The documentation for the `filter()` method in the Supabase Swift client has an incorrect type signature. The documentation shows the `operator` parameter as an enum type (`.in`, `.cs`), while the actual implementation accepts a `String`.

## Current Behavior
Documentation shows usage examples with enum-style operators:
```swift
.filter("id", operator: .in, value: "(5,6,7)")
.filter("arraycol", operator: .cs, value: #"{"a","b"}"#)
```

## Actual Implementation
The actual method signature in the codebase is:
```swift
public func filter(
    _ column: String,
    operator: String,
    value: String
) -> PostgrestFilterBuilder
```

## Impact
This discrepancy can lead to compilation errors when developers copy code directly from the documentation, as the enum syntax shown won't match the expected String parameter.

## Proposed Solution
Either:
1. Update the documentation to show string literals instead of enum values:
```swift
.filter("id", operator: "in", value: "(5,6,7)")
.filter("arraycol", operator: "cs", value: #"{"a","b"}"#)
```

OR

2. Modify the implementation to use an enum type for the operator parameter to match the current documentation.

## Links
- Documentation: [supabase_swift_v2.yml#L2822-L2830](https://github.com/supabase/supabase/blob/6f1ccc5d214550c525ffd45b574d62b3fdbbf8ac/apps/docs/spec/supabase_swift_v2.yml#L2822C1-L2830C10)
- Implementation: [PostgrestFilterBuilder.swift#L510](https://github.com/supabase/supabase-swift/blob/4d33c9c3964399cce8a102d32c26b273a0789235/Sources/PostgREST/PostgrestFilterBuilder.swift#L510)

charislam:
Parameters given for `filter` function of Swift SDK don't match actual implementation. Documented as enum, implementation uses string.

Resolves #31354

### DIFF
--- a/apps/docs/spec/supabase_swift_v2.yml
+++ b/apps/docs/spec/supabase_swift_v2.yml
@@ -32,7 +32,7 @@ functions:
         code: |
           ```swift
           let client = SupabaseClient(
-            supabaseURL: URL(string: "https://xyzcompany.supabase.co")!, 
+            supabaseURL: URL(string: "https://xyzcompany.supabase.co")!,
             supabaseKey: "public-anon-key",
             options: SupabaseClientOptions(
               db: .init(
@@ -62,7 +62,7 @@ functions:
                 schema: "other_schema"
               )
             )
-          ) 
+          )
           ```
         description: |
           By default the API server points to the `public` schema. You can enable other database schemas within the Dashboard.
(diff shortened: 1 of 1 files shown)


charislam:
@grdsdev a sanity check please, as I don't actually know Swift, but the original issue and linked sources made sense to me so made a PR 😁

grdsdev (approved):
Thanks @charislam

PR #32680, opened 2025-01-09
feat: credit top up

kevcodez:
Adds the credit top up functionality to the org billing page, changes include:

- New credit top up modal
- Refactored section to use FormPanel to align styling
- Changed copy of credits section
- Functionality is feature flagged and only enabled on staging/local right now

Screenshot 2025-01-09 at 20 37 24
Screenshot 2025-01-09 at 20 37 29

### DIFF
--- a/apps/studio/components/interfaces/Billing/Payment/AddPaymentMethodForm.tsx
+++ b/apps/studio/components/interfaces/Billing/Payment/AddPaymentMethodForm.tsx
@@ -111,7 +111,7 @@ const AddPaymentMethodForm = ({
   }
 
   return (
-    <form onSubmit={handleSubmit}>
+    <div>
       <Modal.Content
         className={`transition ${isSaving ? 'pointer-events-none opacity-75' : 'opacity-100'}`}
       >
@@ -147,16 +147,17 @@ const AddPaymentMethodForm = ({
         </Button>
         <Button
           block
-          htmlType="submit"
+          htmlType="button"
           size="small"
           type="primary"
           loading={isSaving}
           disabled={isSaving}
+          onClick={handleSubmit}
         >
           Add payment method
         </Button>
       </Modal.Content>
-    </form>
+    </div>
   )
 }
 
--- a/apps/studio/components/interfaces/Organization/BillingSettings/CreditBalance.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/CreditBalance.tsx
@@ -1,4 +1,4 @@
-import { PermissionAction, SupportCategories } from '@supabase/shared-types/out/constants'
+import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { useParams } from 'common'
 import {
   ScaffoldSection,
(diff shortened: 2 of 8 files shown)


apps/studio/components/interfaces/Billing/Payment/AddPaymentMethodForm.tsx
```diff
@@ -111,7 +111,7 @@ const AddPaymentMethodForm = ({
   }
 
   return (
-    <form onSubmit={handleSubmit}>
```
kevcodez, line 114: Removed the form here, as otherwise form on a higher level component are broken and submitted as well causing issues, we dont wanna have forms inside these inlined components

apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+    if (paymentIntentConfirmation.paymentIntent?.status === 'succeeded') {
+      onSuccessfulPayment()
+    }
+  }
+
+  const onSuccessfulPayment = async () => {
+    onTopUpDialogVisibilityChange(false)
+    await queryClient.invalidateQueries(subscriptionKeys.orgSubscription(slug))
+    toast.success(
```
joshenlim, line 179: nit: Input could be reset to 100 after a successful top up? Else if i open the modal again, my previous value is persisted

apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
```diff
@@ -0,0 +1,369 @@
+import { PermissionAction, SupportCategories } from '@supabase/shared-types/out/constants'
```
joshenlim, line 1: nit: we sort our imports as such:

```
// external imports
import { useState } from 'react'
// line break here
//internal imports
import { Component } from './Component'
```

example [here](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureConfiguration/InstanceNode.tsx#L5)

apps/studio/components/interfaces/Organization/BillingSettings/CreditBalance.tsx
```diff
@@ -55,29 +54,30 @@ const CreditBalance = () => {
...
         ) : (
           <>
-            {isLoading && (
-              <div className="space-y-2">
-                <ShimmeringLoader />
-                <ShimmeringLoader className="w-3/4" />
-                <ShimmeringLoader className="w-1/2" />
-              </div>
-            )}
+            <FormPanel footer={<CreditTopUp slug={slug} />}>
```
joshenlim, line 57: tiny nit: extra bottom border when feature flag is off
image

apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+  const [paymentIntentSecret, setPaymentIntentSecret] = useState('')
+  const [paymentIntentConfirmation, setPaymentIntentConfirmation] = useState<PaymentIntentResult>()
+
+  const onSubmit: SubmitHandler<CreditTopUpForm> = async ({ amount, paymentMethod }) => {
+    setPaymentIntentConfirmation(undefined)
+
+    const token = await initHcaptcha()
+
+    await topUpCredits(
```
joshenlim, line 129: no need for await here since this is `mutate` not `mutateAsync`

apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+          <DialogSectionSeparator />
+
+          <Form_Shadcn_ {...form}>
+            <form id={FORM_ID} onSubmit={form.handleSubmit(onSubmit)}>
+              <DialogSection className="flex flex-col gap-2">
+                <FormField_Shadcn_
+                  control={form.control}
+                  name="amount"
+                  render={({ field: { ref, ...rest } }) => (
```
joshenlim, line 247: is the destructured `ref` being used?

joshenlim:
probably need to pull master in? seeing some unrelated TS errors

apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
kevcodez: Think it's fine, pretty unlikely to do two top ups in a row and if so the amount can stay the same

apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
kevcodez: I think it's still cleaner to use await given this is an async function that can fail and it's clearer this way, especially if someone adds a line after

joshenlim: hmm if thats the case then swap the above to use `mutateAsync`? otherwise the await isn't waiting here

joshenlim (approved):

staaldraad (approved):

ecktoteckto (approved):
Tested on staging by doing a few tens of top ups with different cards (With/Without 3DS, insufficient funds etc.)

PR #32683, opened 2025-01-09
Fix view connection string

joshenlim:
Update "view connection string" CTA from read replicas to open the new Connect UI instead of redirecting to settings/database


alaister (approved):
LGTM!

- Tested on preview

PR #32698, opened 2025-01-10
Only show Connect UI in layout header if project is active healthy

joshenlim:
Connect UI wouldn't be relevant unless the project is active healthy


alaister (approved):
LGTM!

- Tested on preview

PR #32699, opened 2025-01-10
Temporarily disable enabling of queues exposure via postgrest

joshenlim:
Disable enabling of queues exposure, but enable toggling it off

When currently disabled:
image

When already enabled:
image


alaister (approved):
LGTM!

- Tested on preview

PR #32736, opened 2025-01-13
Thor/fix hono example

thorwebdev:
* Hono example app was missing `supabase init` to work correctly with `bootstrap`.


apps/docs/content/guides/getting-started/quickstarts/hono.mdx
```diff
@@ -8,12 +8,6 @@ hideToc: true
 <StepHikeCompact>
 
   <StepHikeCompact.Step step={1}>
-    
-    <QuickstartDbSetup />
-
-  </StepHikeCompact.Step>
-
-  <StepHikeCompact.Step step={2}>
```
dshukertjr, line 16: Might need to update the step number on 3, 4, and 5 to 2, 3, and 4.

thorwebdev: Ahh, good catch, thank you!

dshukertjr (approved):

PR #32742, opened 2025-01-13
feature: Adds logs chart with stacked charts with total errors

jordienr:

YES

- Adds simpler component for the logs bar chart shows on top of logs result
- Adds stacked bars to show errors and successes in logs
- Fixes infinite render loop (removes a useeffect)
- Adds tooltip to logs chart component
- lil cleanup 
- uses new ui pkg in chart
- adds LogsBarChart to ui-patterns

- Currently it uses the barchart component that doesnt allow stacked bars 

## What is the new behavior?

CleanShot 2025-01-13 at 20 52 15@2x

## Additional context

Decided to create a new simpler component for this use case. Reasons:
- Current Bar Chart component already has a ton of props and conditionals inside it. It's very easy to udpate this component and break it in some use case. The new component is just for log results meaning if we update and change it it won't break anything anywhere else without us realising.


apps/studio/components/ui/Charts/LogsBarChart.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return (
+    <div className={cn('flex flex-col gap-y-3')}>
+      <ChartContainer
+        config={
+          {
+            error_count: {
+              label: 'Errors',
+            },
+            ok_count: {
+              label: 'Success',
```
Ziinc, line 42: Not such a good idea to label it success, as this would include warning logs.
`Other` would work

apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
```diff
@@ -321,14 +321,16 @@ export const genChartQuery = (
...
   let joins = genCrossJoinUnnests(table)
 
-  return `
+  const q = `
 SELECT
 -- log-event-chart
   timestamp_trunc(t.timestamp, ${trunc}) as timestamp,
-  count(t.timestamp) as count
+  count(CASE WHEN NOT (${errorCondition}) THEN 1 END) as ok_count,
+  count(CASE WHEN ${errorCondition} THEN 1 END) as error_count,
```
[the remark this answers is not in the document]
Ziinc: might want to consider adding a warning condition as well...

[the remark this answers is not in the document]
Ziinc: @jordienr do test the change, not sure if the `not` will work 😅

apps/studio/components/ui/Charts/LogsBarChart.tsx
```diff
@@ -0,0 +1,122 @@
+import dayjs from 'dayjs'
```
MildTomato, line 1: Can you move this to patterns ?

MildTomato: Seems more like a pattern rather than an atom component

MildTomato: Also able to write some docs on it ?

apps/studio/components/interfaces/Settings/Logs/LogsPreviewer.tsx
```diff
@@ -239,23 +225,28 @@ export const LogsPreviewer = ({
...
+              onBarClick={(datum) => {
+                if (!datum?.timestamp) return
                 handleSearch('event-chart-bar-click', {
                   query: filters.search_query as string,
-                  to: isoTimestamp as string,
+                  to: datum.timestamp.toString(),
                   from: null,
                 })
               }}
+              EmptyState={
```
joshenlim, line 243: just to check - when should this show up? I tried on my staging project for storage logs, there's no data but it shows this UI instead

joshenlim (approved):
Generally LGTM, just a small question above ^

Tested on staging preview Postgres logs, verified the charts and tooltip show up nicely (although mine doesnt have any error showing up cause the staging project isn't used haha

PR #32748, opened 2025-01-13
Use whitespace-pre in TextConfirmModal

joshenlim:
Address an edge case whereby the text to confirm has multiple spaces in between words - e.g `Test   Project`

Before:
image

After:
image


saltcod (approved):

saltcod:
Checked and looks good

PR #32749, opened 2025-01-13
Chore/fix set to now

saltcod:
There's a bug with the inline implementation of the new date editor

Set to Now causes an error .

screenshot-2025-01-13-at-12 56 10

This PR reuses the logic from the Sheet panel editor.
screenshot-2025-01-13-at-12 55 49


joshenlim (approved):
Changes LGTM code wise, can see why the bug occured
Unable to verify on staging preview though as the bug doesnt show up on my end (minimally verified that its still saving correctly)

PR #32762, opened 2025-01-14
Add set null cta in ForeignRowSelector

joshenlim:
Only if the source column is nullable
image

Also update InputField for foreign keys in RowEditor to have a consistent UI
image


alaister (approved):
LGTM!

- Tested on preview

PR #32765, opened 2025-01-14
Show all cron jobs even those with no names

joshenlim:
image


saltcod (approved):
confirmed with a no-name job, looks good!

PR #32767, opened 2025-01-14
Chore/fix cron job card

joshenlim:
For CronJobCard
- Opt to use CodeBlock component instead of SqlCodeBlock
- Show copy button to allow copying easily the cron job command (and copying accurately)
  - User was running into an issue whereby he tried to copy the command from another cron job that was using an edge function, to create a new cron job using the SQL snippet from the edge function cron job. But when editing the new cron job after it was created, the UI thinks that the type of the cron job is HTTP request (which leads to a bit of a confusing UX)

For CreateCronJobSheet
- The Snippet option is now a escape hatch for the other options. For example, whatever you change in HTTP request form, it will be reflected in the SQL Snippet textbox. This will enable the users to build the command but further customize if they want in the SQL Snippet form.
- Refactor the `parseCronJobCommand` function to handle additional cases like:
  - plain json in headers (not using the `jsonb_build_object`
  - edge functions to other supabase projects are parsed as HTTP requests now.
- Handle an error in RQ when the first run returns undefined. RQ cache doesn't accept undefined as a value so the value is forced to null.


apps/studio/components/interfaces/Auth/ThirdPartyAuthForm/AddRLSPolicyForFirebaseDialog.tsx
joshenlim, line 1: unused file

alaister (approved):
LGTM!

- Tested on preview

PR #32768, opened 2025-01-14
Fix vault missing exts alert

joshenlim:
The upgrade database alert was missing for vault integrations - which led to a confusion for why we couldn't find the keys and secrets tabs on one of our projects. (Screenshot below is just for reference as I don't have a project which is on PG14)

image

### DIFF
--- a/apps/studio/components/interfaces/Integrations/Landing/Integrations.constants.tsx
+++ b/apps/studio/components/interfaces/Integrations/Landing/Integrations.constants.tsx
@@ -56,7 +56,7 @@ const supabaseIntegrations: IntegrationDefinition[] = [
     id: 'queues',
     type: 'postgres_extension' as const,
     requiredExtensions: ['pgmq'],
-    missingExtensionsAlert: <UpgradeDatabaseAlert />,
+    missingExtensionsAlert: <UpgradeDatabaseAlert minimumVersion="15.6.1.143" />,
     name: `Queues`,
     icon: ({ className, ...props } = {}) => (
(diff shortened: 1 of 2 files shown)


alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/interfaces/Integrations/Landing/Integrations.constants.tsx
```diff
@@ -173,6 +173,7 @@ const supabaseIntegrations: IntegrationDefinition[] = [
     id: 'vault',
     type: 'postgres_extension' as const,
     requiredExtensions: ['supabase_vault'],
+    missingExtensionsAlert: <UpgradeDatabaseAlert />,
```
ivasilov, line 176: This component is specific to Queues. It says `The extension is available on version 15.6.1.143 and higher.`. Technically, it's correct (you can upgrade to a lower version than that) but the extension is available since PG v15 (I think).

ivasilov: Should we add another field for minimum version and use that in the alert?

joshenlim: > should we add another field for minimum version

that makes sense! although it wouldn't matter too much atm since we don't support selecting a specific version to upgrade to (upgrades are always just the latest version iinw)

i'll still add a minimum version prop nonetheless, and default it to 15.6 for now

joshenlim: okay pushed :)

PR #32769, opened 2025-01-14
SQL Chart fixes

jordienr:

YES

- fix labels for X axis when there are too many results
- add labels to select inputs in chart configuration form
- add domain hopefully to fix issue with Y axis cutting out (cant repro)

CleanShot 2025-01-14 at 16 39 36@2x

## What is the new behavior?

CleanShot 2025-01-14 at 16 40 01@2x


joshenlim (approved):
Found some bugs which I've pushed up fixes for 

- Chart config is not y scrollable, couldn't fully access the bottom checkbox
- For a chart that has x axis as dates, and y axis as numbers, hitting Flip causes a client crash which cannot be reverted as the config is stored in local storage (Bar is unable to render string values on the y axis) -> I reckon we only support flipping if both axes are fully numerical values only perhaps
  - [Edit] Hmm, actually i just tried to chart number of countries per continent and flipping the axes look fine actually... We can look into that separately as its not related to this PR

PR #32770, opened 2025-01-14
docs: add minimal jwt example with custom access token hook

hf:
Adds an example on how to use the custom access token hook to cut down on JWT size.


hf:
https://github.com/supabase/auth/pull/1913

charislam (approved):

PR #32778, opened 2025-01-14
fix: troubleshooting sync workflow

charislam:
Workflow was missed when updating to pnpm

### DIFF
--- a/.github/workflows/docs-sync-troubleshooting.yml
+++ b/.github/workflows/docs-sync-troubleshooting.yml
@@ -29,17 +29,22 @@ jobs:
           sparse-checkout: |
             apps/docs
 
-      - name: Set up Node.js
+      - uses: pnpm/action-setup@v4
+        name: Install pnpm
+        with:
+          run_install: false
+
+      - name: Use Node.js
         uses: actions/setup-node@v4
         with:
           node-version-file: '.nvmrc'
-          cache: 'npm'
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-sync-troubleshooting.yml
```diff
@@ -29,11 +29,17 @@ jobs:
           sparse-checkout: |
             apps/docs
 
-      - name: Set up Node.js
+      - uses: pnpm/action-setup@v4
+        name: Install pnpm
+        with:
+          version: 9.10.0
```
ivasilov, line 35: You can leave out the version. According to the [docs](https://github.com/pnpm/action-setup?tab=readme-ov-file#version), it will use the `packageManager` key in `package.json`.

.github/workflows/docs-sync-troubleshooting.yml
```diff
@@ -29,11 +29,17 @@ jobs:
...
+
+      - name: Use Node.js
         uses: actions/setup-node@v4
         with:
           node-version-file: '.nvmrc'
-          cache: 'npm'
+          cache: 'pnpm'
 
       - name: Install dependencies
         run: npm ci
```
ivasilov, line 45: Should all mentions of npm be refactored to `pnpm`?

ivasilov (approved):

PR #32788, opened 2025-01-15
Revert "Temp disable count query"

alaister:
- Reverts supabase/supabase#31256
- Also updates the code to use the new count API


joshenlim (approved):
LGTM! Verified locally that the count values are working and accurate

PR #32792, opened 2025-01-15
AI Assistant Performance

SaxonF:

This PR makes several changes in an effort to improve performance when using the Assistant. There were a few issues with the current implementation, but the major culprit that caused fps drops was the re-rendering of the QueryBlock component while streaming content. 

The fix was to delay the rendering until the sql snippet was complete. There is no reliable way to do this so the solution on this branch introduces a debounce component. 

The branch also reduces re-rendering across other components.

The result is a more stable FPS while streaming content.


joshenlim (approved):
Changes LGTM! Tested on staging preview and all seems okay too 🙂

PR #32796, opened 2025-01-15
fix: Various fixes

ivasilov:
Small fixes:
- Fix a folder path in local-studio-tests package;
- Don't show the database selector in self-hosted studio.
- Instantiate openAi inside the code instead of in global to avoid causing an HTML error if the API key is not present.
- Move aiTools file out of pages/api because it was accessible via API and may end up causing a vulnerability.


Hallidayo:
Hi @ivasilov - We’re doing a bit of housekeeping and plan to close out PRs that have been inactive for more than 3 months.

PR #32804, opened 2025-01-15
Billing docs revamp - cost control [GEN-12430]

ecktoteckto:


charislam (approved):

apps/docs/components/Navigation/NavigationMenu/NavigationMenu.constants.ts
```diff
@@ -2090,6 +2090,10 @@ export const platform: NavMenuConstant = {
           name: 'Your monthly invoice',
           url: '/guides/platform/your-monthly-invoice',
         },
+        {
+          name: 'How to control your costs',
```
kevcodez, line 2094: Maybe shorter without losing any context: Control your costs

apps/docs/content/guides/platform/cost-control.mdx
```diff
@@ -0,0 +1,56 @@
+---
+id: 'cost-control'
+title: 'How to control your costs'
+---
+
+## Spend Cap
+The Spend Cap determines whether your organization can exceed the subscription plan's free quota allowance for any metric. Scenarios that could lead to high usage—and thus high costs—include system attacks or bugs in your software. The Spend Cap can protect you from these unexpected costs for certain metrics.
```
kevcodez, line 7: We might want to clarify two things early on:

- If you are on the free plan, you will never be charged at all
- Spend Cap is only possible on Pro Plan

apps/docs/content/guides/platform/cost-control.mdx
```diff
@@ -0,0 +1,56 @@
...
+- Function Invocations
+- Monthly Active Users
+- Monthly Active SSO Users
+- Monthly Active Third Party Users
+- Realtime Messages
+- Realtime Peak Connections
+- Storage Image Transformations
+- Storage Size
+
+### Metrics not covered by the Spend Cap
```
kevcodez, line 35: We might wanna add a sentence as to *why* these are not covered: They have predictable costs and are user choices

kevcodez (approved):

apps/docs/components/Navigation/NavigationMenu/NavigationMenu.constants.ts
ecktoteckto: Yeah, was thinking about that. Was afraid that it might be understood as an "imperative". But you're probably right. Gonna change it.

apps/docs/content/guides/platform/cost-control.mdx
ecktoteckto: Done

PR #32821, opened 2025-01-16
chore: remove react fiber

kevcodez:
Unused, Alaister just added it after I removed it because his subconscious likes to see me suffer


alaister (approved):
💚

MildTomato (approved):
@alaister confirms it can be removed.

PR #32831, opened 2025-01-16
docs(billing): Revamp - Subscription [GEN-12451]

ecktoteckto:
This page is in kind of an in-between state. But good enough for feedback I'd say.

Things to consider:
- might make sense to refactor the parts "About billing on Supabase" and "Change your subscription" into separate pages
- I wanted to describe the general concept of credits. Not sure though how detailed it should be in that place. Whether the top up feature should be explained here as well. Leaving it for now and will come back to it later once we have a more complete picture
- we might change the diagram that describes the org / billing account structure. But the one we have should do for now


apps/docs/content/guides/platform/manage-your-subscription.mdx
```diff
@@ -0,0 +1,60 @@
+---
+id: 'manage-your-subscription'
+title: 'Manage your subscription'
+---
+
+## About billing on Supabase
+Supabase bills separately for each organization. That means that each organization has its own subscription, including a unique plan (Free, Pro, Team, or Enterprise), billing cycle, and invoices.
+
+Different plans cannot be mixed within a single organization. For example, you cannot have both a Pro Plan project and a Free Plan project in the same organization. To have projects on different plans, you must create separate organizations. See [Project Transfers](docs/guides/platform/project-transfer) if you need to move a project to a different organization.
```
charislam, line 9:
```suggestion
Different plans cannot be mixed within a single organization. For example, you cannot have both a Pro Plan project and a Free Plan project in the same organization. To have projects on different plans, you must create separate organizations. See [Project Transfers](/docs/guides/platform/project-transfer) if you need to move a project to a different organization.
```

charislam (approved):

apps/docs/content/guides/platform/manage-your-subscription.mdx
```diff
@@ -0,0 +1,60 @@
+---
+id: 'manage-your-subscription'
+title: 'Manage your subscription'
+---
+
+## About billing on Supabase
+Supabase bills separately for each organization. That means that each organization has its own subscription, including a unique plan (Free, Pro, Team, or Enterprise), billing cycle, and invoices.
```
kevcodez, line 7:
```suggestion
Supabase bills separately for each organization. That means that each organization has its own subscription, including a unique plan (Free, Pro, Team, or Enterprise), payment method, billing cycle, and invoices.
```

kevcodez (approved):

PR #32834, opened 2025-01-16
Fix chart click

jordienr:

YES

Fixes the way dates are calculated when a user clicks the chart in logs.

It will sometimes calculate dates in a way that it shows no results.
CleanShot 2025-01-16 at 15 55 38

## What is the new behavior?

CleanShot 2025-01-16 at 15 54 24


apps/studio/components/interfaces/Settings/Logs/LogsPreviewer.tsx
```diff
@@ -154,14 +155,6 @@ export const LogsPreviewer = ({
         iso_timestamp_start: nextStart,
         iso_timestamp_end: nextEnd,
       }))
-      router.push({
```
jordienr, line 157: not necessary since I moved this state to nuqs a while ago

saltcod (approved):
looks good on preview!

PR #32849, opened 2025-01-17
chore: add tests pg meta package

avallete:
In the effort to help migrate the current `pgMeta` server more and more toward a library I've setup the structure to migrate tests from https://github.com/supabase/postgres-meta/tree/master/test to this new lib with minimal changes/refactor.

So we can keep coverage, and expand bit by bit. Started by porting the `schema` tests as a first example.

Also added some utils to reduce side effects of tests between each others by scoping each of them into their own freshly seeded database before dropping it. It'll allow for some nice parallel testing at some point.


packages/pg-meta/src/pg-meta-schemas.ts
```diff
@@ -127,7 +127,7 @@ function remove(
     id?: number
     name?: string
   },
-  { cascade = false }: SchemaRemoveParams = {}
+  { cascade = false }: SchemaRemoveParams = { cascade: false }
```
soedirgo, line 130: We already do `{ cascade = false }` so this is redundant

packages/pg-meta/src/pg-meta-schemas.ts
```diff
@@ -4,9 +4,9 @@ import { SCHEMAS_SQL } from './sql/schemas'
 import { z } from 'zod'
 
 const pgSchemaZod = z.object({
-  id: z.number(),
-  name: z.string(),
-  owner: z.string(),
+  id: z.coerce.number(),
+  name: z.coerce.string(),
+  owner: z.coerce.string(),
```
soedirgo, line 9: Why do we need `coerce`?

packages/pg-meta/test/schemas.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    try {
+      await fn(db)
+    } finally {
+      await db.cleanup()
+    }
+  })
+}
+
+// Example of refactored tests using the wrapper
+withTestDatabase('list with system schemas', async ({ executeQuery }) => {
```
soedirgo, line 28: 🔥🔥🔥

packages/pg-meta/src/pg-meta-schemas.ts
avallete: I'm not sure how we plan to use this `api` wise. But from my understanding, it's supposed to be used to retrieve the expected result of the `sql` query. However, using `executeRaw` with something like `postgrest-js` adapter for instance (like I do in the tests) will result in `id` being a `string`. Containing a valid number.

So then, using a `zod.parse(result)` here would fail because zod would say that we got a string instead of a number.

I think in most case tough, we want to use zod to coerce the type if it can, and fail if it can't.

avallete: cc @soedirgo little ping on this one

soedirgo: I see what you mean - which type ends up being returned depends on the Postgres adapter used to make the query. `postgres` uses string for `oid`s but another adapter may use something else (iirc `pg` uses number by default).

For `id` specifically, ~all of them are of type `oid` whose max value is well under JS's `Number.MAX_SAFE_INTEGER`, so there's no issue with using numbers to represent them. I'll update the SQL to not convert it into `int8` (I don't remember why I did that in the first place, probably some quirks with `pg`).

I think it's better to be strict on which type is used if possible, because we can't control for adapters doing something weird (e.g. converting timestamp strings to `Date`, which is lossy). Maybe for now we can just follow whatever `postgres` returns by default, as it tends to do the sane thing afaict.

soedirgo (approved):

PR #32738, opened 2025-01-13
ReportsV2 P1: Add Query block component

joshenlim:
## Add QueryBlock component, replaces the SqlSnippet component in Assistant
image
Part 1 for ReportsV2 (More context in the section below), heavily referenced from Saxon's work on `feat/reports-v2`.

- Visually identical to existing component, but added an additional action here "View data" 
  - Swap between viewing as chart (if the assistant output is a chart data) and table which shows the raw data from the query
image
  - Support dynamically viewing data as cumulative
image
- Add loading state when query result is not loaded yet (Primarily when opening the dashboard from a fresh session and assistant is open)
image
- Fixes user messages not getting persisted across browser sessions (currently only assistant messages are, some bug with stale state when saving to local storage)

## Context on what ReportsV2 is
We've generally decided on the direction for ReportsV2 and will be doing the changes in incremental small(ish) PRs, briefly:

- [x] Create a `QueryBlock` component that will a shared component between the Assistant (replaces `SqlSnippet`) and ReportsV2
- [ ] Implement SQL blocks in Reports section that will use this new `QueryBlocks` component, and support saving custom report with SQL blocks (SQL blocks will be rendered from SQL snippets from the SQL Editor)
- [ ] Support concept of parameters in Reports section

^ Once the above is done, we can also look into additional features (not in any particular order):
  - [ ] Add some UI indication to link SQL editor to Reports (to educate users about this)
  - [ ] Support dragging a `QueryBlock` from Assistant to Reports

PR #32851, opened 2025-01-17
ReportsV2 P2: Support adding SQL snippets to Reports

joshenlim:
## Support adding SQL snippets to Reports

_This is currently behind a feature flag so no user-facing changes_

Again heavily referenced from Saxon's work on `feat/reports-v2`, You can now add a SQL snippet created from the SQL editor as a block in custom reports as such:
image
image
image
image

You may change the view between data table and chart (depending on how you want to visualize the data), update the axes, toggle cumulative - all these will get saved and persisted into the custom report. There's an edit button as well which will allow you to edit the SQL in the SQL Editor, so the reports page is purely just for viewing data.

SQL blocks will only be rendered if the query is entirely read only (no mutations). For e.g if i try to render the Countries quick start, we'll render this:
image

Also unrelated some fixes and improvements (these will be user-facing once merged):
- Fix re-ordering blocks not showing a save / cancel option
- Show placeholder behind blocks when dragging blocks for clearer indication of its position

## Context on what ReportsV2 is
We've generally decided on the direction for ReportsV2 and will be doing the changes in incremental small(ish) PRs, briefly:

- [x] Create a `QueryBlock` component that will a shared component between the Assistant (replaces `SqlSnippet`) and ReportsV2 ([PR](https://github.com/supabase/supabase/pull/32738))
- [x] Implement SQL blocks in Reports section that will use this new `QueryBlocks` component, and support saving custom report with SQL blocks (SQL blocks will be rendered from SQL snippets from the SQL Editor)
- [ ] Support concept of parameters in Reports section

^ Once the above is done, we can also look into additional features (not in any particular order):
  - [ ] Add some UI indication to link SQL editor to Reports (to educate users about this)
  - [ ] Support dragging a `QueryBlock` from Assistant to Reports
  - [ ] UI update to make charts and SQL blocks consistent


SaxonF:
Before I begin the review I did run into a `Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.` error. 

I got this when prompting in assistant for a count of tasks by week. I can confirm it works fine in master but not in this branch. I would assume some changes to QueryBlock would be causing it.

SaxonF:
One more bug, switching reports doesn't reset the state, so old report remains.

SaxonF:
I agree the pushing down of grid items is annoying. I think we should resolve that before pushing.

When loading the report , it shows the not readonly message even if they are.

image

A little quality of life improvement, even if we add as a follow up along with the other tasks, is that we should hide the date picker if no default charts are added as it has no effect on custom snippets.

Other than the issues noted above, its working well in my testing. I'll do a full code review on Monday.

joshenlim:
Hmm it actually seems to be an open issue with react-grid-layout RE the pushing items down problem
https://github.com/react-grid-layout/react-grid-layout/issues/1982
https://github.com/react-grid-layout/react-grid-layout/issues/720

joshenlim:
Okay pushed up fixes for the following:
- Fix client crash when querying assistant panel
- Fix dragging charts downwards pushes the item below it down
- Fix switching reports doesn't update the report
- Hide date picker unless a chart of provider daily-stats or infra-monitoring is in the report

SaxonF (approved):
Tested well and code looks good! LGTM.

PR #32860, opened 2025-01-17
chore: Rename all uses of `Tooltip_Shadcn_` to just `Tooltip`

ivasilov:
This PR is last PR for migrating to use shadcn Tooltips. It's find-and-replace of all uses of `Tooltip_Shadcn_` to just `Tooltip`.


staaldraad (approved):

joshenlim (approved):
Minimally tested logging in and basic smoke test on table / sql editor
If the build passes then it should be okay haha

PR #32866, opened 2025-01-17
feat: Add a banner for clock skew in studio

ivasilov:
This PR adds a banner which will appear if the user has clock skew on their computer. Clock skew is important because it's causing issues with JWT verification and can make the whole studio unresponsive.

Screenshot 2025-01-17 at 17 42 09


staaldraad (approved):
CSP change approved.

saltcod (approved):

saltcod:
Checked against preview and looks good!

PR #32898, opened 2025-01-18
Assistant chat check keycode enter

joshenlim:
## Enforce `keyCode` 13 when hitting enter in the AssistantChatForm before submitting prompt

Context:
For certain languages, like Chinese or Japanese for example, typing in characters allow choosing other characters when you hit the down arrow key as such:

image

The typically UX will be to find the character that the user wants, then hit enter to select. However, because the textarea in AssistantChatForm listens to the onKeyDown "Enter" event, this will prematurely submit the prompt. Text areas (afaik) do not really support triggering the "onSubmit" event on forms as the default behaviour for hitting "Enter" in textareas is to create a new line instead.

 I did notice that hitting "Enter" when selecting the japanese character fires a different keyCode of 229 ([more information here](https://github.com/facebook/react/issues/14512)), and hitting Enter when there's no selection fires a keyCode of 13 (despite both "key" properties returning "Enter"). So this should do the trick for now


dshukertjr (approved):
I tested it in the preview, and it works like a charm! Thanks for the quick fix!

Screenshot 2025-01-18 at 20 10 43

alaister (approved):
LGTM!

- Tested on preview

PR #32904, opened 2025-01-18
fix: search index upload

charislam:
Search index pipeline shouldn't be trying to read troubleshooting section yet (those are still searched via GH discussions). Trying to read it errors as they don't have the same frontmatter shape as other guides content.


kevcodez (approved):
Rubber stamping

PR #32920, opened 2025-01-20
docs: add example to order parent table by referenced table

soedirgo:
[Context](https://www.notion.so/supabase/Add-example-of-ordering-parent-table-by-foreign-table-with-inner-bb2751e43dc04b71bdec507360877555?pvs=4)

### DIFF
--- a/apps/docs/spec/supabase_dart_v2.yml
+++ b/apps/docs/spec/supabase_dart_v2.yml
@@ -4683,6 +4683,56 @@ functions:
             }
           ]
           ```
+      - id: order-parent-table-by-a-referenced-table
+        name: Order parent table by a referenced table
+        code: |
+          ```dart
+          final data = await supabase
+              .from('cities')
+              .select('''
+                name,
+                country:countries (
+                  name
+                )
+              ''')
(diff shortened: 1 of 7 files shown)


grdsdev (approved):

apps/docs/spec/supabase_dart_v2.yml
```diff
@@ -4683,6 +4683,56 @@ functions:
...
+          ```dart
+          final data = await supabase
+              .from('cities')
+              .select('''
+                name,
+                country:countries (
+                  name
+                )
+              ''')
+              .order('country(name)', ascending: true)
```
dshukertjr, line 4698: Dart requires the comma at the end.
```suggestion
              .order('country(name)', ascending: true);
```

dshukertjr (approved):
Just a small change on Dart code, but otherwise looks good! Thanks for this!

apps/docs/spec/supabase_dart_v2.yml
soedirgo: 🙏 thanks!

charislam (approved):

PR #32923, opened 2025-01-20
fix: don't retry queries on 4xx errors

alaister:

Bug fix/chore

All errors other than 404 are retried up to 3 times.

## What is the new behavior?

All 400 class errors are not retried. It's unlikely that retrying these type of errors will result in a different response, and the cost of retrying them is the user having to wait a long time before they see the error message – which is usually a result of a bad input on the users part.

## Testing

An easy way to test this is input the wrong type in the filter input in the table editor and you'll immediately see an error now:
Screenshot 2025-01-20 at 13 34 32


joshenlim (approved):
LGTM! Tested on preview in table editor

PR #32934, opened 2025-01-20
chore: skip invalidation on sql only changes

alaister:

chore

On text changes in the sql editor, we invalidate all the queries and counts in the sidebar.

## What is the new behavior?

There's no need to invalidate anything, so we don't.


joshenlim (approved):
LGTM! Tested on preview that count doesn't trigger when updating snippet

PR #32942, opened 2025-01-20
fix: remove duplicate archive function from pgmq_public

alaister:

Bug fix

Two identical archive functions.

## What is the new behavior?

Only one.


ivasilov (approved):

olirice:
The duplicate archive function is an error but we're not exposing batch functionality on purpose

If the user falls submits more than a few ids it will fall off index and result in timeouts when queues are large.

We can expose the functionality but should restrict it to 5-6 ids at a time with an error message. Not great, but a better UX than timeouts.

For now I'd suggest removing the redundant code instead

PR #32944, opened 2025-01-20
Fix: Update PostGIS tutorial documentation

### LINKED ISSUE #28310: Unclear examples in PostGIS extension documentation
# Improve documentation

## Link 

https://supabase.com/docs/guides/database/extensions/postgis

## Describe the problem

I followed the example under "Enable the extension" exactly, first through the "Dashboard" tab. In the fourth step, I used "Create a new schema" and picked the name `gis`, just as in the example.

I then went on to try and create a table with a column using a new PostGIS field. I ran the exact query mentioned in the example on the page (except renaming the table). Running the query gives me an error in the Result tab:

```
ERROR:  42704: type "geography" does not exist
```

I eventually got it to work by prefixing the `geography` field with the schema name I used in the previous section, `gis`. So instead of

```
location geography(POINT) not null
```

I used

```
location gis.geography(POINT) not null
```

which made it work.

## Describe the improvement

I'm fairly new to both Supabase and PostgreSQL, but it seem to me that something is missing in the documentation here? Is it expected that the fields need to be prefixed with a schema name if it's not in the `public` schema? I was thinking about suggesting an exact improvement, but hesitating as I don't feel I understand the whole problem. Hopefully this could be made clearer by someone who knows the details around this.

## Additional context

Also note that when trying to create a "geography" field thought the Table Editor, I get the same issue. And in here, I can't tell if I'm even able to connect the field to its schema (`gis`) in the same way I did in the SQL Editor.

Screenshot 2024-08-01 at 8 49 25 AM

coreAux:

YES

Updating docs for PostGIS.

Resolves #28310


GaryAustin1:
I confirmed the tutorial was bugged with the user here.  And the fix of allowing usage of the gis schema is needed.
I'm not positive the search path with just gis is sufficient as shown, but does work for the example.

charislam (commented):
Thanks @coreAux!

Instead of setting the search path to gis, what do you think of setting search_path to '' and qualifying the necessary tables/function names? Bit more of a generally safe solution

tejasnasre:
your fixes are correct i was getting type geometry not found mine issue solved thanks for the solution. happy coding

coreAux:
> Thanks @coreAux!
> 
> Instead of setting the search path to gis, what do you think of setting search_path to '' and qualifying the necessary tables/function names? Bit more of a generally safe solution

Hey @charislam!

If I was a bit more experienced with SQL I could probably adjust that easily. What does setting `search_path` to `"` do? When I try it in the SQL editor it throws an error.

Even if I set the `search_path` to `gis`, without granting access to the `gis` for `anon` I get the following error when trying to call it from my client:

```bash
'{"error":{"code":"42704","details":null,"hint":null,"message":"type \\"geometry\\" does not exist"}}' 
```

charislam:
@coreAux this is what I mean:

```
create or replace function nearby_restaurants(lat float, long float)
returns table (id public.restaurants.id%TYPE, name public.restaurants.name%TYPE, lat float, long float, dist_meters float)
set search_path = ''
language sql
as $$
  select id, name, gis.st_y(location::gis.geometry) as lat, gis.st_x(location::gis.geometry) as long, gis.st_distance(location, gis.st_point(long, lat)::gis.geography) as dist_meters
  from public.restaurants
  order by location operator(gis.<->) gis.st_point(long, lat)::gis.geography;
$$;
```

Setting search_path to an empty string now means you have to schema-qualify every reference in your function definition, and protects against some varieties of Postgres vulnerabilities that depend the search path order. You can read more about it in our [docs](https://supabase.com/docs/guides/database/database-advisors?queryGroups=lint&lint=0011_function_search_path_mutable) and the [Postgres docs](https://www.postgresql.org/docs/current/runtime-config-client.html).

And I think you're right about granting usage, you'll need to do that as well

coreAux:
> @coreAux this is what I mean:
> 
> ```
> create or replace function nearby_restaurants(lat float, long float)
> returns table (id public.restaurants.id%TYPE, name public.restaurants.name%TYPE, lat float, long float, dist_meters float)
> set search_path = ''
> language sql
> as $$
>   select id, name, gis.st_y(location::gis.geometry) as lat, gis.st_x(location::gis.geometry) as long, gis.st_distance(location, gis.st_point(long, lat)::gis.geography) as dist_meters
>   from public.restaurants
>   order by location operator(gis.<->) gis.st_point(long, lat)::gis.geography;
> $$;
> ```
> 
> Setting search_path to an empty string now means you have to schema-qualify every reference in your function definition, and protects against some varieties of Postgres vulnerabilities that depend the search path order. You can read more about it in our [docs](https://supabase.com/docs/guides/database/database-advisors?queryGroups=lint&lint=0011_function_search_path_mutable) and the [Postgres docs](https://www.postgresql.org/docs/current/runtime-config-client.html).
> 
> And I think you're right about granting usage, you'll need to do that as well

Awesome! Thanks for the explanation and for throwing in the example as well! I would've never been able to guess that syntax, especially the operator one, coming from JS. 

I made a new commit updating both functions in the tutorial.

apps/docs/content/guides/database/extensions/postgis.mdx
```diff
@@ -203,15 +203,22 @@ Sorting datasets from closest to farthest, sometimes called nearest-neighbor sor
...
-  order by location <-> st_point(long, lat)::geography;
+  order by location operator(gis.<->) gis.st_point(long, lat)::gis.geography;
 $$;
 ```
 
-You can call this function from your client using `rpc()` like this:
+Before being able to call this function from our client we need to grant access to our `gis` schema:
+
+```sql
+grant usage on schema gis to anon, authenticated, service_role;
```
charislam, line 218:
```suggestion
grant usage on schema gis to anon, authenticated;
```

charislam (approved):
Awesome thanks @coreAux!

PR #32945, opened 2025-01-21
Add missing users delete endpoint for local/self-host

joshenlim:
As per PR title - although I'm not sure how did I miss this 🙈


saltcod:
Checked & looks good!

saltcod (approved):

tecoad:
@joshenlim When will this be shipped to production????
Can understand how such things are placed in production...

silentworks:
@tecoad this is not an issue on hosted Supabase, so when you say production this isn't an issue on production. If you are asking when will it be shipped to the CLI then that will depend on when the next version of the CLI gets released. You can also use an older version of the CLI where this feature works until this is released. Also bugs happen in every code base, so a bug getting released isn't something new to development, what matters is that it gets fixed.

waymond91:
I am having this exact issue on a self hosted supabase deployment we did yesterday with docker.
It seems like this PR is related to my issue, or am I mistaken?
Do I need to do anything special to get this change into my docker deployment?
image

silentworks:
@waymond91 I don't know how you configure your docker deploys but if you are using the `docker-compose.yml` provided in this repo then just update this line https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml#L14 to the latest tag of studio which is currently `20250127-7ad4ad6`.

PR #32955, opened 2025-01-21
fix: sign users out if profile request 401s

alaister:

Bug fix

If the user has a bad auth token, auth-js won't know about it and will think the user is authenticated, leading to endless loading states.

## What is the new behavior?

If the profile request returns 401, signs the user out.


jordienr (approved):

jordienr:
tested by breaking the token manually in the preview, got redirected to sign in ✅

PR #32960, opened 2025-01-21
Fix SQL editor menu wrongly hiding results

joshenlim:
For favourites and shared, we were depending on whether their corresponding `count` values (from `/content/count`) to decide what to render.

Because the result from `/content/count` didn't tally with what's coming from `/content`:
- e.g `/content/count` returns that shared snippets count is 0, but `/content` returns 2 results for shared snippets
- Hence why FE didn't render the shared snippets results due to the incorrect `/content/count` value

However, looking twice at this, I reckon we don't need to rely on the `/content/count` value to determine the UI rendering, can solely depend on the results from `/content/count`


alaister (approved):
LGTM!

- Tested on preview

PR #32962, opened 2025-01-21
fix: show database charts at 1h resolution

Ziinc:
Fixes resolution issue where `min` aggregation is not displaying correct values. Also improves multi-day UX so that users can narrow in to the specific hours where issues arise.


apps/studio/pages/project/[ref]/reports/database.tsx
```diff
@@ -100,7 +100,15 @@ const DatabaseUsage = () => {
                 value={'7d'}
                 options={TIME_PERIODS_INFRA}
                 currentBillingPeriodStart={undefined}
-                onChange={setDateRange}
+                onChange={(values) => {
+                  // override datepicker interval, show DB charts at hourly intervals.
+                  // https://linear.app/supabase/issue/ANL-640/disk-io-consumption-not-displaying-correctly-in-db-report
```
Comment (jordienr) [inline apps/studio/pages/project/[ref]/reports/database.tsx:105]: no need to add this in a comment IMO we can always find this PR through the commit

jordienr (approved):

PR #32963, opened 2025-01-21
Temp fix to read meta property from jwt update status query

joshenlim:


apps/studio/data/config/jwt-secret-updating-status-query.ts
```diff
@@ -29,7 +29,8 @@ export async function getJwtSecretUpdatingStatus(
 
   if (error) handleError(error)
 
-  const meta = data.jwtSecretUpdateStatus
+  // @ts-expect-error API typing issue here
+  const { meta } = data.jwtSecretUpdateStatus
```
kamilogorek, line 33:
```suggestion
  const meta = data.jwtSecretUpdateStatus?.meta ?? data.jwtSecretUpdateStatus
```

This way it'll work without any updates when API is fixed.

PR #32989, opened 2025-01-22
Feat/reports v2/misc wrap up

joshenlim:
- Add Telemetry events for:
  - Adding SQL block to custom report
  - Dragging SQL block from Assistant to custom report
- Add UI indication in SQL Editor Chart Options for new functionality
image
- Add UI indication in SQL block within Assistant Panel for new functionality (currently only shows up while in a custom report)
image


apps/studio/lib/constants/telemetry.ts
```diff
@@ -512,3 +514,11 @@ export interface AssistantEditInSqlEditorClickedEvent {
     isInNewSnippet: boolean
   }
 }
+
```
pamelachia, line 517: For every event you should add comments above them describing the precise action taken and where it would happen. Both events sound the same to me. If you're not sure how to name the event, describe the precise action and I can suggest. The event should always end with a verb and the verb attached should also be something [we commonly use](https://www.notion.so/supabase/Event-tracking-standardization-1195004b775f80f98ee3fa9e70cf4d05?pvs=4#3bc9bbba903c49b5bd9b27bfa9d290d3) unless you think there's no good replacement. Submitted and clicked should cover most cases. 

eg.
/**
 * User clicked add to custom report button in the SQL block in the assistant sidebar
 *
 * @group Events
 * @source studio
 * @page /dashboard/project/{ref}/reports
 */

SaxonF (approved):
Tests well. Once telemetry conflicts are resolved I think we can LGTM

pamelachia (approved):
events look good!

PR #32995, opened 2025-01-22
pg cron logs: fix chart query and add error condition

jordienr:

YES

- fix the chart query for pg cron logs
- adds the error condition for pg cron logs to show errors in the chart

- The chart wont load because its trying to query a non existing table instead of the postgres_logs table with the correct filter

## What is the new behavior?

CleanShot 2025-01-22 at 13 15 47@2x

## Additional context

To test this
- Create a cron that selects an existing table in the db
- Create a cron that selects a non existing table

You should see bars in the chart for errors (non existing table) and OKs

### DIFF
--- a/apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
+++ b/apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
@@ -326,7 +326,7 @@ export const genChartQuery = (
   // pg_cron logs are a subset of postgres logs
   // to calculate the chart, we need to query postgres logs
   if (table === LogsTableName.PG_CRON) {
-    table = LogsTableName.POSTGREST
+    table = LogsTableName.POSTGRES
     where = `where (parsed.application_name = 'pg_cron' OR event_message LIKE '%cron job%')`
   }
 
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
```diff
@@ -320,9 +320,16 @@ export const genChartQuery = (
...
   const [startOffset, trunc] = calcChartStart(params)
-  const where = genWhereStatement(table, filters)
+  let where = genWhereStatement(table, filters)
   const errorCondition = getErrorCondition(table)
 
+  // pg_cron logs are a subset of postgres logs
+  // to calculate the chart, we need to query postgres logs
+  if (table === LogsTableName.PG_CRON) {
+    table = 'postgres_logs'
+    where = `where (parsed.application_name = 'pg_cron' OR event_message LIKE '%cron job%')`
```
Ziinc, line 330: hmm what are the chances of non-pg_cron events ending up in the stream unexpectedly?

saltcod (approved):

PR #33010, opened 2025-01-22
fix:error filter for troubleshooting

charislam:
Fix error codes not showing up or not filtering properly in troubleshooting filter.

Also fix local dev when local infra is not available.

### DIFF
--- a/apps/docs/features/docs/Troubleshooting.ui.client.tsx
+++ b/apps/docs/features/docs/Troubleshooting.ui.client.tsx
@@ -2,7 +2,7 @@
 
 import { ChevronDown, RotateCw, Search, X } from 'lucide-react'
 import { useQueryStates } from 'nuqs'
-import { useEffect, useRef, useState, Suspense, useCallback } from 'react'
+import { useEffect, useRef, useState, Suspense, useCallback, useMemo } from 'react'
 
 import {
   Input_Shadcn_,
@@ -14,7 +14,6 @@ import {
 } from 'ui'
 import ShimmeringLoader from 'ui-patterns/ShimmeringLoader'
 
-import { MultiSelect } from '~/components/MultiSelect.client'
 import {
   MultiSelector,
   MultiSelectorContent,
@@ -107,7 +106,9 @@ function entryMatchesFilter(
     selectedTags.length === 0 || selectedTags.some((tag) => dataKeywords?.includes(tag))
   const errorsMatch =
     selectedErrorCodes.length === 0 ||
-    selectedErrorCodes.some((error) => dataErrors.includes(error))
+    selectedErrorCodes.some((error) =>
+      dataErrors.some((errorCode) => errorCode.includes(error.toString()))
+    )
   const searchMatch =
     searchState === '' || content.toLowerCase().includes(searchState.toLowerCase())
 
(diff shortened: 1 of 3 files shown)


apps/docs/features/docs/Troubleshooting.ui.client.tsx
```diff
@@ -200,6 +202,25 @@ function TroubleshootingFilterInternal({
...
+            }
+            if (error.code) {
+              result.push(error.code)
+            }
+            return result
+          }) ?? []
+        )
+      ),
+    [errors]
+  )
```
gregnr, line 222: Quick q - why create a `Set` then convert back to an array? Couldn't you return the result of the `flatMap` directly?

apps/docs/features/docs/Troubleshooting.utils.ts
```diff
@@ -27,7 +27,9 @@ export interface ITroubleshootingEntry {
 export const getArticleSlug = getArticleSlugInternal
 
 async function getAllTroubleshootingEntriesTyped() {
-  const result = (await getAllTroubleshootingEntriesInternal()) as ITroubleshootingEntry[]
+  const result = (
+    IS_PLATFORM ? await getAllTroubleshootingEntriesInternal() : []
+  ) as ITroubleshootingEntry[]
```
gregnr, line 32: Just a casting nit: try to avoid casting unless you have no other option, otherwise it defeats the purpose of type safety. I'm wondering if `getAllTroubleshootingEntriesInternal()` can explicitly return `ITroubleshootingEntry[]`, then you can just do:

 ```suggestion
  const result: ITroubleshootingEntry[] = (
    IS_PLATFORM ? await getAllTroubleshootingEntriesInternal() : []
  )
```

File: apps/docs/features/docs/Troubleshooting.ui.client.tsx [hunk 3]
```diff
@@ -220,9 +241,9 @@ function TroubleshootingFilterInternal({
           <MultiSelector.Trigger badgeLimit={1} className="w-48" label="Error codes" />
           <MultiSelector.Content>
             <MultiSelector.List>
-              {errors?.map((error) => (
-                <MultiSelector.Item key={`error-${error.code}`} value={error.code}>
-                  {error.code}
+              {allErrorCodes?.map((error) => (
+                <MultiSelector.Item key={`error-${error}`} value={error}>
+                  {error}
```
gregnr, line 246: Still need to optional chain now that there's the `[]` fallback?

```suggestion
              {allErrorCodes.map((error) => (
                <MultiSelector.Item key={`error-${error}`} value={error}>
                  {error}
```

gregnr (approved):

apps/docs/features/docs/Troubleshooting.ui.client.tsx
charislam: To remove duplicates: we could have multiple error entries that return the same HTTP status code for example

gregnr: Gotcha!

apps/docs/features/docs/Troubleshooting.ui.client.tsx
charislam: good catch!

apps/docs/features/docs/Troubleshooting.utils.ts
charislam: This problem occurs because we don't have `strictNullChecks` on (too many existing violations in the code base, need to find time to sit down and grind through all of them 😬), so Zod doesn't return the right required types even though the schema validation enforces it.

I either have to cast or deal with fake optional fields all the way down, so I figure casting is less annoying. Root fix is just to grit my teeth and get `strictNullChecks` done, I'll make a separate task for that!

gregnr: Pesky legacy code! Makes sense 😄

PR #33021, opened 2025-01-23
remove globe in onboarding

SaxonF:
This PR removes the globe from the new onboarding flow that's currently in trial. It looked nice, but was causing performance issues and was incompatible with certain browsers/devices.

All other behaviour remains the same.

image

### DIFF
--- a/apps/studio/pages/new/[slug].tsx
+++ b/apps/studio/pages/new/[slug].tsx
@@ -38,7 +38,7 @@ import {
   ProjectCreateVariables,
   useProjectCreateMutation,
 } from 'data/projects/project-create-mutation'
-import { TelemetryActions } from 'lib/constants/telemetry'
+import { TelemetryActions } from 'common/telemetry-constants'
 import { useProjectsQuery } from 'data/projects/projects-query'
 import { useOrgSubscriptionQuery } from 'data/subscriptions/org-subscription-query'
 import { useCheckPermissions } from 'hooks/misc/useCheckPermissions'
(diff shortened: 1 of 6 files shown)


packages/common/telemetry-constants.ts
```diff
@@ -302,6 +303,17 @@ export interface FeaturePreviewDisabledEvent {
...
 }
 
+/**
+ * Existing project creation form was submitted and the project was created.
+ *
+ * @group Events
+ * @source studio
+ * @page new/{slug}
+ */
+export interface ProjectCreationCurrentSubmittedEvent {
```
pamelachia, line 313: can we change the name to not use Current? Maybe ProjectCreationSimpleVersionSubmittedEvent or ProjectCreationSimpleSubmittedEvent. What is 'current' or what is 'existing' can change and we might lose track of what it means.

joshenlim (approved):
LGTM! Tested creating a project on preview

apps/studio/pages/new/[slug].tsx
```diff
@@ -194,6 +198,9 @@ const Wizard: NextPageWithLayout = () => {
     isSuccess: isSuccessNewProject,
   } = useProjectCreateMutation({
     onSuccess: (res) => {
+      sendEvent({
+        action: TelemetryActions.PROJECT_CREATION_CURRENT_SUBMITTED,
```
Comment (loong) [inline apps/studio/pages/new/[slug].tsx:202]: I think this hasn't been renamed

loong (approved):

PR #33026, opened 2025-01-23
Fix SQL block not rendering if its the only block in the report

joshenlim:


alaister (approved):
LGTM!

- Tested on preview

PR #33064, opened 2025-01-24
Update SQL editor search to return a flat list

joshenlim:
## Opting to show a flat list of snippets instead
- IMO faster searching UX for finding what you're looking for
  - Exposes queries that are within folders (easier finding)
    - Currently when searching, we still render the folders and you'd need to open the folders to check if the snippets are in them or not
  - Hides the 3 separate sections, makes it easier to scan for what you're looking for
    - Am also thinking when you're searching for something, you don't really care which section they're in
- Probably more importantly, relieves some load off the API
  - We were making 3 requests each time we trigger a search since each section (shared, favorites and private) was controlled by their own react query state
  - This change ensures that we only fire one

https://github.com/user-attachments/assets/3cd07693-882d-486b-b848-842d67d0316a

## Misc fixes:
- Resolves a common bug "Node ID doesn't exist in tree" -> happens when opening a snippet that's within a folder on a fresh browser session
- Makes the sort icon in the search field higher contrast (feedback was that users kept missing this action)

### DIFF
--- a/apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
+++ b/apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
@@ -1,10 +1,12 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { FilePlus, FolderPlus, Plus } from 'lucide-react'
 import { useRouter } from 'next/router'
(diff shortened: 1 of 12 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #33073, opened 2025-01-24
Billing docs revamp - Credits

ecktoteckto:
- ignore all files other than `credits.mdx`

- I tried out a "new" approach with pictures. I added the pictures not because they are needed because things are super complicated. I rather thought it might help the customer associate what they read in the docs with the Dashboard. Let me know what you think

- I added a FAQ section. Although we already have the big "Billing FAQ" I thought it might be useful to have a dedicated section on some pages. If we keep going with that approach I would refactor the respective FAQs into reusable partials later

- regarding partials in general: I'm not gonna refactor everything into partials right away. There's a lot of moving parts right now and it might not be worth the effort. Gonna do that in the end.

- I decided to not use a "see here for more info" link everywhere and instead write things out. I think that might be a better experience for the customer in some cases. E.g. the "Credits upon downgrade" section on page "Change your subscription plan"


apps/docs/components/Navigation/NavigationMenu/NavigationMenu.constants.ts
```diff
@@ -2101,6 +2101,10 @@ export const platform: NavMenuConstant = {
             },
           ],
         },
+        {
```
charislam, line 2104: We should probably revisit the order of these at the end, I feel like "Your monthly invoice" and "How billing works" are placed a bit too low in comparison to other sections

apps/docs/content/guides/platform/credits.mdx
```diff
@@ -0,0 +1,60 @@
...
+    dark: '/docs/img/guides/platform/credit-balance--dark.png',
+  }}
+  zoomable
+/>
+
+You can find the credit balance on the [organization's billing page](https://supabase.com/dashboard/org/_/billing).
+
+### What causes the credit balance to change?
+**Subscription plan downgrades:** Upon subscription downgrade, any prepaid subscription fee will be credited back to your organization for unused time in the billing cycle.\
+As an example, if you start a Pro Plan subscription on January 1 and downgrade to the Free Plan on January 15, your organization will receive approximately $12.50 in credits for the unused time between January 15 and January 31.
```
charislam, line 22: Maybe we should leave out the concrete number here, seems like a place that would be extremely easy to overlook when updating. Or we can calculate it from a centralized value.

apps/docs/content/guides/platform/credits.mdx
```diff
@@ -0,0 +1,60 @@
...
+  zoomable
+  className="max-w-[500px]"
+/>
+
+## Credit FAQ
+### Can I transfer credits to another organization?
+Yes, you can transfer credits to another organization. Please submit a [support ticket](https://supabase.help).
+
+### Can I get a refund of my unused credits?
+No, we do not provide refunds. Please refer to our [Terms of Service](https://supabase.com/terms#:~:text=5.-,Fees%20and%20Taxes.,-1.%20Fees.).
```
charislam, line 57: Would be good to put anchor links in the ToS for use cases like this. Links to highlights only work in Chrome, and only as long as we don't change the target text.

charislam (approved):

apps/docs/components/Navigation/NavigationMenu/NavigationMenu.constants.ts
ecktoteckto: Yes 👍 All "old" pages except from the FAQs will be deleted though, I guess.

apps/docs/content/guides/platform/credits.mdx
ecktoteckto: Good point. Changed it to 50%. This way the reader doesn't have to know that the Pro Plan is currently $25.

apps/docs/content/guides/platform/credits.mdx
ecktoteckto: Ah, wasn't aware of that. Just copied that over from the FAQs. Changed it to anchors in both places.

charislam (approved):

kevcodez (approved):

PR #33071, opened 2025-01-24
chore: Use `sql-formatter` for formatting all SQL in the `studio`

ivasilov:

PR #33075, opened 2025-01-24
chore: Fix Monaco editor in strict mode

ivasilov:
This PR fixes an issue in the SQL editor which causes the monaco instance to be initialized twice in React Strict mode which results in an console error. This happens every time when you refresh the /dashboard/project/_/sql/new page (but not if you're navigating to it from elsewhere). 

The issue is reported in https://github.com/suren-atoyan/monaco-react/issues/440.

https://github.com/supabase/supabase/pull/33071 should be merged first.


alaister (approved):
LGTM!

- Tested on preview

PR #33113, opened 2025-01-27
Support dupe query

joshenlim:
As per PR title


apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SQLEditorTreeViewItem.tsx
```diff
@@ -294,6 +294,16 @@ export const SQLEditorTreeViewItem = ({
                   Move query
                 </ContextMenuItem_Shadcn_>
               )}
+              {onSelectDuplicate !== undefined && isOwner && (
```
alaister, line 297: Just to check, is there a reason the user needs to be an owner to duplicate a query? I think it would be useful to be able to duplicate other team members queries to change variables in them etc

alaister (approved):
LGTM!

- Tested on preview

alaister (approved):
Double approve

PR #33128, opened 2025-01-27
chore: add auto-troubleshooting-template to local dev

charislam:
Small DX improvement for writing troubleshooting guides. When a new troubleshooting guide is created, it is auto-populated with the template (as long as docs dev is running).


apps/docs/scripts/troubleshooting/watch.mjs
```diff
@@ -0,0 +1,68 @@
...
+    .catch((err) => {
+      console.error('Error closing watcher:', err)
+      process.exit(1)
+    })
+}
+
+process.on('SIGINT', () => shutdown('SIGINT'))
+process.on('SIGTERM', () => shutdown('SIGTERM'))
+process.on('SIGHUP', () => shutdown('SIGHUP'))
+process.on('SIGQUIT', () => shutdown('SIGQUIT'))
```
gregnr, line 68: Do you think it's worth just doing:

```suggestion
process.on('SIGINT', shutdown)
process.on('SIGTERM', shutdown)
process.on('SIGHUP', shutdown)
process.on('SIGQUIT', shutdown)
```

apps/docs/scripts/troubleshooting/watch.mjs
```diff
@@ -0,0 +1,68 @@
...
+      const fd = openSync(path, 'w')
+      writeSync(fd, getTemplate())
+      closeSync(fd)
+    } catch (err) {
+      console.error(`Error writing template to ${path}:`, err)
+    }
+  })
+
+// Keep the process alive
+const interval = setInterval(() => {}, 1000)
```
gregnr, line 44: Strange you need this since chokidar's [persistence flag](https://github.com/paulmillr/chokidar?tab=readme-ov-file#persistence) should have handled this.

apps/docs/scripts/troubleshooting/watch.mjs
```diff
@@ -0,0 +1,68 @@
+// @ts-check
+
+import chokidar from 'chokidar'
+import { closeSync, openSync, readFileSync, writeSync } from 'node:fs'
+import { basename, dirname, join } from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = dirname(__filename)
+const TROUBLESHOOTING_DIRECTORY = join(__dirname, '../../content/troubleshooting')
```
gregnr, line 10: You can also just do:

```suggestion
const TROUBLESHOOTING_DIRECTORY = join(import.meta.dirname, '../../content/troubleshooting')
```

gregnr (approved):
Nice 😄

apps/docs/scripts/troubleshooting/watch.mjs
charislam: Yeah I thought it would automatically work too, no idea what went wrong

PR #33141, opened 2025-01-28
Support peeking referencing record from table editor

joshenlim:
image


alaister (approved):
LGTM!

- Tested on preview

Two possible improvements that could be down as follow ups:
1. Show the column info vertically. There's no real need to show a horizontal row here – we've got more vertical room available.
2. If it's the auth users schema being referenced, show the user somehow instead:
Screenshot 2025-01-28 at 19 33 10

WoodyWoodster:
That was quick! Looks great!

joshenlim:
> Show the column info vertically. There's no real need to show a horizontal row here – we've got more vertical room available.

gotcha on this! will think about this one - main consideration here was to keep a similar UI for the peek as per the Table Editor grid so that it feels relatively "seamless", but I do agree that vertically is a better use of the space since its only one row

> If it's the auth users schema being referenced, show the user somehow instead:

will need to double check what you might mean for this one? Not sure I follow just yet 🙏 

will get this through first! can iterate as we get more feedback along too 🙂

PR #33161, opened 2025-01-28
Billing docs revamp - Compute usage

ecktoteckto:
- decided to build things around "Usage" instead of "Metrics". I think that gives us a bit more flexibility regarding what content to put on these pages and we can abstract things away that might be a bit too "technical" otherwise
- we can probably extract a few things that repeat into the top level "Manage your usage" page. I'm not doing this right now though. Want to see what patterns we have first. Might also be better to have some repetition and work with partials instead of having to jump back and forth between pages too much
- we need to have a look regarding the terminology again. We are not consistent with things like "Compute size", "Compute instance", "Server", with/without capital letters etc. Not in the docs and not on the Dashboard. We have to agree on things first though
- there'll be separate pages for Branching Compute and Read Replica Compute
- to fetch prices from a central source is done at the end


apps/docs/content/guides/platform/manage-your-usage.mdx
```diff
@@ -0,0 +1,4 @@
+---
+id: 'manage-your-usage'
+title: 'Manage your usage'
+---
```
charislam, line 4: We'll need some kind of text here ,even if it's only a link to subpages, it looks too weird without any content

charislam (approved):

apps/docs/content/guides/platform/manage-your-usage/compute.mdx
```diff
@@ -0,0 +1,117 @@
...
+  src={{
+    light: '/docs/img/guides/platform/usage-compute--light.png',
+    dark: '/docs/img/guides/platform/usage-compute--dark.png',
+  }}
+  zoomable
+/>
+
+## Optimize Compute usage
+- load test your application in staging to understand your Compute requirements
+- start out on a smaller Compute size, [create a report](https://supabase.com/dashboard/project/_/reports) in the Dashboard to monitor your CPU and memory utilization, and upgrade the Compute size as needed
```
kevcodez, line 113: Is this supposed to be the section on how to reduce usage?

If yes, let's add something along the lines of
- Delete projects or transfer a paid plan project using self-serve project transfers to reduce compute usage

kevcodez (approved):

apps/docs/content/guides/platform/manage-your-usage.mdx
ecktoteckto: Yes. Either the link, kind of an explanation text what the usage pages are about and/or info that is relevant for all the subpages. E.g. "what does 'GB-Hrs' mean?"

Gonna add links for now.

apps/docs/content/guides/platform/manage-your-usage/compute.mdx
ecktoteckto: Done

PR #33164, opened 2025-01-29
FE-1338: New UI for LogsDatePickers (FE-1512)

jordienr:

YES

- Updates the UI for LogsDatePickers
- Uses less space in trigger
- Allows copypasting date ranges 
- Allows pasting dates into TimeSplitInput
- Adds more tests to DatePicker

CleanShot 2025-02-03 at 22 41 21@2x

CleanShot 2025-02-03 at 22 41 36@2x

## What is the new behavior?

CleanShot 2025-02-03 at 22 40 53@2x


joshenlim:
Wondering if this is a bug - From what I can tell the default option is last hour but
- The bottom time range seems to round off to the nearest hour? would expect 14:15 to 15:15 (instead of 14:00 to 15:15)
image
- Should be there some selected state for the presets? (e.g lighter background color)
image

Minor feedback: wondering if the badge here should just be a tooltip so it shows up without having to click
image

apps/studio/components/interfaces/Settings/Logs/Logs.DatePickers.tsx
```diff
@@ -1,116 +1,357 @@
...
+interface Props {
+  value: DatePickerValue
+  onSubmit: (args: DatePickerValue) => void
   helpers: DatetimeHelper[]
 }
-const DatePickers = ({ to, from, onChange, helpers }: PropsWithChildren<Props>) => {
-  const defaultHelper = getDefaultHelper(helpers)
-  const [helperValue, setHelperValue] = useState<string>(to || from ? '' : defaultHelper.text)
+
+export const LogsDatePicker: React.FC<Props> = ({ onSubmit, helpers, value }) => {
```
joshenlim, line 27: generally avoid using `FC`, can use `PropsWithChildren` instead

`PropsWithChildren<{onSubmit, helpers, value}>`

apps/studio/components/ui/DatePicker/TimeSplitInput.tsx
```diff
@@ -3,6 +3,11 @@ import { Clock } from 'lucide-react'
 import React, { useEffect, useState } from 'react'
```
joshenlim, line 3: unrelated but can remove `React`? import selectively in `{ ... } from 'react'`

saltcod (commented):
Looks good! 
Tested on preview, pasting date ranges works great!

joshenlim (approved):
just ran through the changes again and tested in Edge Functions logs, Logs Explorer, and Postgres Logs

all issues identified now addressed, LGTM 🙂

joshenlim:
Wait sorry found one bug - if i have a custom date range applied, and i hit refresh, the end date range updates to the current time, which is an unexpected side effect. The UI doesn't update, but the URL does

https://github.com/user-attachments/assets/8f232e1b-ab1f-47ba-b703-0fd08ca1a633

PR #33172, opened 2025-01-29
db: allow public read access to troubleshooting entries table

charislam:


gregnr (approved):
Nice 🙌 worth updating [this code](https://github.com/supabase/supabase/blob/03515ee277ec2bff1a684e4ee35393ce59258e44/apps/docs/features/docs/Troubleshooting.utils.ts#L102-L105) to use a non-admin client? (eg. `NEXT_PUBLIC_SUPABASE_ANON_KEY`)

PR #33174, opened 2025-01-29
fix: add mobile sheet header

bytaesu:

YES

Improving mobile user experience

There is no close button, and the distinction between the Sheet and the outer area is unclear, making it very inconvenient to use.

## What is the new behavior?

I’ve added a sticky header and a close button without compromising the Supabase ui design. The code has been modified to ensure it doesn’t affect other design systems.

### Before:
before-gif

### After:
after-gif

dark-gif


bytaesu:
> Thanks @bytaesu for the suggestion!
> 
> 
> 
> We'd prefer to keep the sheet without the header as it adds quite a bit of vertical white (or black) empty space.
> 
> That said, we understand that it might feel weird to not have a close button when you close the sheet the first time, but from the second time it should become a pretty straight forward gesture to just tap the background to close it.
> 
> 
> 
> The thing that we _could_ indeed do is to add the border so that the sheet stands out a bit more, as you also suggested in this pr.
> 
> Would you mind just updating the pr to just keep the `border` addition? 
> 
> But also add `border-b-0` so that the bottom of the sheet remains flush to the bottom of the screen.



Got it.

For a clean commit history, I’ll submit a new PR either today or tomorrow. Once that PR receives a review, I will close this one!

PR #33204, opened 2025-01-30
Billing docs revamp - Branching usage

ecktoteckto:
- maybe we have to move this page somewhere else since it's not a usage item / metric on its own but actually a combination of different items / metrics
- decided to use the "Usage Summary" instead of the "Compute Hours" section for "View Branching usage". "Compute Hours" only distinguishes between Micro<->Small etc. but not between Not-Branch<->Branch.
- we might want to change the tooltip text of item "Branches Compute Hours" in the "Usage Summary". In my opinion a bit confusing for the user. Just removed the text entirely locally for now for the screenshot. Have to change the screenshot again then.


charislam (commented):
Hmm I think these changes were also in the replicas PR? So I reviewed them there. Lemme know if I'm mistaken and they're different

apps/docs/content/guides/platform/manage-your-usage/branching.mdx
```diff
@@ -0,0 +1,67 @@
...
+/>
+
+## Optimize Branching usage
+- merge Preview branches as soon as they are ready
+- delete Preview branches that are no longer in use
+- check whether your [persistent branches](/docs/guides/deployment/branching#persistent-branches) need to be defined as persistent, or if they can be ephemeral instead. Persistent branches will remain active even after the underlying PR is closed.
+
+## Branching FAQ
+### Do Compute Credits apply to Branching Compute?
+No, Compute Credits do not apply to Branching Compute
```
kevcodez, line 67:
```suggestion
No, Compute Credits do not apply to Branching Compute.
```

kevcodez (approved):
Assuming the comments from Charis in the other PR get addresses

charislam (approved):

PR #33210, opened 2025-01-31
Warn for certain symbols in database passwords

joshenlim:
Fixes FE-484

Callout when `@`, `:`, and `/` in database passwords that users need to percent-encode them (as they are used as delimiters in postgres connection strings and are a source of confusion where it appears the customer's client side tooling can't connect.)

Generate password function doesnt need updating since it's only alphanumerical

image

### DIFF
--- a/apps/studio/components/interfaces/ProjectCreation/SpecialSymbolsCallout.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/SpecialSymbolsCallout.tsx
@@ -3,11 +3,11 @@ import { InlineLink } from 'components/ui/InlineLink'
 export const SpecialSymbolsCallout = () => {
   return (
     <p className="mb-2">
-      Note: If using a connection string, you will need to{' '}
+      Note: If using the Postgres connection string, you will need to{' '}
       <InlineLink href="https://supabase.com/docs/guides/database/postgres/roles#special-symbols-in-passwords">
         percent-encode
       </InlineLink>{' '}
-      the special symbols in the password
+      your password
     </p>
   )
 }
--- a/apps/studio/components/interfaces/ProjectCreation/SpecialSymbolsCallout.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/SpecialSymbolsCallout.tsx
@@ -7,7 +7,7 @@ export const SpecialSymbolsCallout = () => {
       <InlineLink href="https://supabase.com/docs/guides/database/postgres/roles#special-symbols-in-passwords">
         percent-encode
       </InlineLink>{' '}
-      your password
+      the password
     </p>
   )
 }
--- a/apps/studio/components/interfaces/ProjectCreation/SpecialSymbolsCallout.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/SpecialSymbolsCallout.tsx
@@ -1,0 +1,13 @@
+import { InlineLink } from 'components/ui/InlineLink'
+
+export const SpecialSymbolsCallout = () => {
+  return (
+    <p className="mb-2">
+      Note: If using a connection string, you will need to{' '}
+      <InlineLink href="https://supabase.com/docs/guides/database/postgres/roles#special-symbols-in-passwords">
+        percent-encode
+      </InlineLink>{' '}
+      the special symbols in the password
+    </p>
+  )
+}
--- a/apps/studio/pages/new/[slug].tsx
+++ b/apps/studio/pages/new/[slug].tsx
@@ -12,6 +12,7 @@ import { z } from 'zod'
 import { PopoverSeparator } from '@ui/components/shadcn/ui/popover'
 import { components } from 'api-types'
 import { useParams } from 'common'
+import { TelemetryActions } from 'common/telemetry-constants'
 import {
   FreeProjectLimitWarning,
   NotOrganizationOwnerWarning,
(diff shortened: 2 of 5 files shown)


ivasilov (approved):

MildTomato (changes requested):
Left comment in linear.

joshenlim:
Great catch - will need to update the PR
It shouldn't be restricting from creating passwords with those characters, but add a callout that the user will need to URL encode them 

should also link to our docs:
image

apps/studio/pages/new/[slug].tsx
```diff
@@ -779,41 +778,58 @@ const Wizard: NextPageWithLayout = () => {
...
-                            description={
-                              <PasswordStrengthBar
-                                passwordStrengthScore={form.getValues('dbPassStrength')}
-                                password={field.value}
-                                passwordStrengthMessage={passwordStrengthMessage}
-                                generateStrongPassword={generatePassword}
-                              />
-                            }
-                          >
-                            <FormControl_Shadcn_>
```
Comment (MildTomato) [inline apps/studio/pages/new/[slug].tsx:795]: `If using a connection string, you will need to precent-encoded password`

frontload the important part, if they are using a connection string

apps/studio/pages/new/[slug].tsx
```diff
@@ -779,41 +778,58 @@ const Wizard: NextPageWithLayout = () => {
...
-                                  } else {
-                                    await delayedCheckPasswordStrength(value)
-                                  }
-                                }}
-                              />
-                            </FormControl_Shadcn_>
-                          </FormItemLayout>
-                        )}
+                        render={({ field }) => {
+                          const regex = /^[^@:\/]*$/
```
Comment (MildTomato) [inline apps/studio/pages/new/[slug].tsx:782]: it's a bit weird the regex is stored here.
also feels like we should be testing it.

apps/studio/pages/new/[slug].tsx
```diff
@@ -779,41 +778,58 @@ const Wizard: NextPageWithLayout = () => {
...
+                                  placeholder="Type in a strong password"
+                                  {...field}
+                                  autoComplete="off"
+                                  onChange={async (event) => {
+                                    field.onChange(event)
+                                    form.trigger('dbPassStrength')
+                                    const value = event.target.value
+                                    if (event.target.value === '') {
+                                      await form.setValue('dbPassStrength', 0)
+                                      await form.trigger('dbPass')
```
Comment (MildTomato) [inline apps/studio/pages/new/[slug].tsx:823]: i know this wasn't in the PR, but this is strange.
seems like we should just form.trigger('dbPassStrength') regardless, and not check for a value to be defined for this input

apps/studio/pages/new/v2/[slug].tsx
```diff
@@ -716,40 +713,60 @@ const WizardForm = () => {
...
-                                              } else {
-                                                await delayedCheckPasswordStrength(value)
-                                              }
-                                            }}
-                                          />
-                                        </FormControl_Shadcn_>
-                                      </FormItemLayout>
-                                    )}
+                                    render={({ field }) => {
+                                      const regex = /^[^@:\/]*$/
```
Comment (MildTomato) [inline apps/studio/pages/new/v2/[slug].tsx:717]: yep. defo should have stored this somewhere else, and once!

apps/studio/pages/new/v2/[slug].tsx
```diff
@@ -716,40 +713,60 @@ const WizardForm = () => {
...
+                                            <>
+                                              {hasSpecialCharacters && (
+                                                <p className="mb-2">
+                                                  Note: Special symbols need to be{' '}
+                                                  <InlineLink href="https://supabase.com/docs/guides/database/postgres/roles#special-symbols-in-passwords">
+                                                    percent-encoded
+                                                  </InlineLink>{' '}
+                                                  when using your password later in Postgres
+                                                  connection strings
+                                                </p>
```
Comment (MildTomato) [inline apps/studio/pages/new/v2/[slug].tsx]: I would think we should be storing these messages more globally. 
We're making it easy for 1 of these to change in another PR by someone else and they won't realize the same logic and content is replicated elsewhere

apps/studio/pages/new/v2/[slug].tsx
```diff
@@ -716,40 +713,60 @@ const WizardForm = () => {
...
+                                              copy={field.value.length > 0}
+                                              type="password"
+                                              placeholder="Type in a strong password"
+                                              {...field}
+                                              autoComplete="off"
+                                              onChange={async (event) => {
+                                                field.onChange(event)
+                                                form.trigger('dbPassStrength')
+                                                const value = event.target.value
+                                                if (event.target.value === '') {
```
Comment (MildTomato) [inline apps/studio/pages/new/v2/[slug].tsx:758]: not applicable to this PR, but it's weird we are checking it's defined

apps/studio/pages/new/[slug].tsx
Comment (joshenlim) [reply apps/studio/pages/new/[slug].tsx:823]: this is to immediately set the password strength bar to 0 if the password value is empty
otherwise the checking of password strength is on a debounce

apps/studio/components/interfaces/ProjectCreation/SpecialSymbolsCallout.tsx
```diff
@@ -0,0 +1,13 @@
+import { InlineLink } from 'components/ui/InlineLink'
+
+export const SpecialSymbolsCallout = () => {
+  return (
+    <p className="mb-2">
+      Note: If using the Postgres connection string, you will need to{' '}
+      <InlineLink href="https://supabase.com/docs/guides/database/postgres/roles#special-symbols-in-passwords">
+        percent-encode
+      </InlineLink>{' '}
+      your password
```
MildTomato, line 10:
```suggestion
      the password
```

MildTomato: belongs to database, not the user

MildTomato (approved):

MildTomato:
checked preview, only checked input onChange

joshenlim:
verified project creation! 🙏

PR #33211, opened 2025-01-31
Add refresh button to reports

joshenlim:
Quick heads up for Custom Reports and Database reports:
- These trigger the refreshes asynchronously so the "loading" state of the refresh button is just 1 second and not dependent on the actual loading state of the charts, because the parent component is unable to track the loading state
- Verified locally that each refresh button triggers the corresponding RQs to invalidate

Custom reports:
image

API report:
image

Storage report:
image

Database report:
image


saltcod (approved):
Checked preview and looks good

PR #33215, opened 2025-01-31
Chore/fix cancel button in data api settings

joshenlim:
Fixes FE-317

Hitting cancel after making a change in Data API settings toggles Enable Data API to false and closes the UI which is 
incorrect - this PR fixes it

### DIFF
--- a/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
+++ b/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
@@ -109,15 +109,18 @@ export const PostgrestConfig = () => {
   const isGraphqlExtensionEnabled =
     (extensions ?? []).find((ext) => ext.name === 'pg_graphql')?.installed_version !== null
 
+  const dbSchema = config?.db_schema ? config?.db_schema.replace(/ /g, '').split(',') : []
+  const defaultValues = {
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -109,15 +109,18 @@ export const PostgrestConfig = () => {
   const isGraphqlExtensionEnabled =
     (extensions ?? []).find((ext) => ext.name === 'pg_graphql')?.installed_version !== null
 
+  const dbSchema = config?.db_schema ? config?.db_schema.replace(/ /g, '').split(',') : []
+  const defaultValues = {
+    dbSchema,
+    maxRows: config?.max_rows,
+    dbExtraSearchPath: config?.db_extra_search_path,
+    dbPool: config?.db_pool,
+  }
```
ivasilov, line 118: Is `enableDataApi` missing here?

joshenlim: it gets set in a useEffect on L158 below cause its a "computed" property on

alaister (approved):
LGTM!

- Tested on preview

ivasilov (approved):

PR #33219, opened 2025-01-31
Chore/esc cron names

saltcod:
Allow apostrophes in cron job names

Update: should also allow other special characters, since you can add whatever you want via sql 

Example to reproduce: 
- try creating this job via sql 
- SELECT cron.schedule('Backup /logs/ + daily', '0 3 * * *', 'VACUUM'); 
- the History view for this job should work 

screenshot-2025-01-31-at-12 44 32


jordienr (approved):
tested on staging!

PR #33230, opened 2025-02-01
Chore/connect UI update for session pool mode

joshenlim:
Fixes FE-1356
Changes are only if pooling configuration is currently set to Session mode:

Show a callout under Connect UI -> Transaction Pooling instead of the connection string
image

Add a callout under Connect UI -> Session Pooling RE Port 6543 deprecation
image

Add a callout under Database Settings -> Pooling Configuration RE Port 6543 deprecation
image

Update callout under Database Settings -> Pooling Configuration to change to port 5432 if using 6543 when switching to Transaction mode (This callout was already implemented, just changing the language a bit and using Admonition warning instead)
image


alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/ui/Panel.tsx
```diff
@@ -62,14 +62,16 @@ const PanelNotice = forwardRef<
     description?: string
     href?: string
     buttonText?: string
+    layout?: 'horizontal' | 'vertical'
   }
->(({ className, title, description, href, buttonText, ...props }, ref) => {
```
MildTomato: @joshenlim just make sure this is ok I think

PR #33232, opened 2025-02-01
fix: hide pnpm update message, update to pnpm 9.15.5, specify pnpm >=9 to unbreak non-corepack pnpm 10, remove corepack in Dockerfile

thebengeu:
pnpm 10 upgrade should be handled separately, since it has breaking changes.

- **fix: install pnpm in base stage using npm, remove corepack**
  There's little point using corepack in the Docker image, now that pnpm
10 supports automatically downloading and running the version of pnpm
specified in the `packageManager` field of `package.json` by default:
https://pnpm.io/npmrc#manage-package-manager-versions.

  Having the pnpm version specified in the Dockerfiles instead of having
`packageManager` as the source of truth makes it slightly more likely
that the versions would diverge if we miss updating the version in some
places, but `manage-package-manager-versions` would take care of that
with pnpm 10, and we're on the latest version of pnpm 9 already.

  I think getting rid of corepack is worth the duplication of the pnpm
version within the Dockerfiles. Corepack recently caused issues with
https://github.com/nodejs/corepack/issues/612.

- **fix: remove `version` from `pnpm/action-setup@4`, redundant since it'll be picked up from `packageManager`**
  
- **fix: specify engine as >=9 instead of ^9, which breaks `packageManager` auto-install with pnpm 10 if not using corepack**

```
supabase/apps/studio 
❯ trash /Users/beng/.local/share/pnpm/.tools/pnpm/9.10.0

supabase/apps/studio 
❯ pnpm install
 ERROR  Unsupported engine for /Users/beng/supabase/supabase: wanted: {"pnpm":"^9.0.0"} (current: {"node":"v20.11.1","pnpm":"10.2.0"})
For help, run: pnpm help install

supabase/apps/studio 
❯ pnpm install
 ERROR  Failed to switch pnpm to v9.10.0. Looks like pnpm CLI is missing at "/Users/beng/.local/share/pnpm/.tools/pnpm/9.10.0/bin" or is incorrect
spawnSync /Users/beng/.local/share/pnpm/.tools/pnpm/9.10.0/bin/pnpm ENOENT
```
  
- **fix: install corepack v0.31.0 as workaround for nodejs/corepack#612**
  
- **chore: update to pnpm 9.15.5**
  
- **fix: hide pnpm update message, it's confusing since the pnpm version is locked**

### DIFF
--- a/.github/workflows/ai-tests.yml
+++ b/.github/workflows/ai-tests.yml
@@ -36,7 +36,7 @@ jobs:
       - uses: pnpm/action-setup@v4
         name: Install pnpm
         with:
-          version: 9.10.0
+          version: 9.15.5
           run_install: false
       - name: Use Node.js
         uses: actions/setup-node@v4
--- a/.github/workflows/ai-tests.yml
+++ b/.github/workflows/ai-tests.yml
@@ -36,7 +36,6 @@ jobs:
       - uses: pnpm/action-setup@v4
         name: Install pnpm
         with:
-          version: 9.15.5
           run_install: false
       - name: Use Node.js
         uses: actions/setup-node@v4
--- a/.github/workflows/autofix_linters.yml
+++ b/.github/workflows/autofix_linters.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: pnpm/action-setup@v4
         name: Install pnpm
         with:
-          version: 9.10.0
+          version: 9.15.5
           run_install: false
 
       - uses: actions/setup-node@v4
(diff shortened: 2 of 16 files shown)


kamilogorek (approved):

joshenlim (approved):
LGTM! Minimally verified that dash, www, and docs all render okay
Although not sure what else I should be looking out for 🙏

thebengeu:
These changes are very likely to be safe, pnpm 10 is where there are breaking changes.

PR #33250, opened 2025-02-03
fix success message in query block

SaxonF:
Fixes FE-1362

image

re-adds adding a message to QueryBlock (assistant and reports) when a query is run but no results returned e.g. a mutation


apps/studio/components/interfaces/SQLEditor/UtilityPanel/Results.tsx
```diff
@@ -127,19 +127,27 @@ const Results = ({ rows }: { rows: readonly any[] }) => {
...
-        column={cellPosition?.column.name ?? ''}
-        value={cellPosition?.row?.[cellPosition.column.name]}
-        visible={expandCell}
-        onClose={() => setExpandCell(false)}
-      />
+      {rows.length === 0 ? (
+        <div className="bg-table-header-light [[data-theme*=dark]_&]:bg-table-header-dark">
+          <p className="m-0 border-0 px-6 py-4 font-mono text-sm">Success. No rows returned</p>
+        </div>
+      ) : (
```
joshenlim, line 134: this shouldn't be required as it's already handled in the parent component

joshenlim (approved):
LGTM! Verified locally

PR #33255, opened 2025-02-03
Add default array for cron http headers

joshenlim:
Client error occurs as `values.httpHeaders` from `buildCommand` is undefined, so the filter errors out


alaister (approved):
LGTM!

- Tested on preview

PR #33256, opened 2025-02-03
Fix database tables not showing FKs on first panel open

joshenlim:
To reproduce:
- Go to Database -> Tables (It's not happening in Table Editor)
- Select "Edit table" on a table that has foreign keys
- Scroll down to the Foreign Keys section of the side panel, notice that's 0 keys
- Close the panel, and open the panel again via "Edit table"
- Foreign keys are there now


alaister (approved):
LGTM!

- Tested on preview

saltcod (approved):
double approved as I just did the exact same PR :^)

PR #33258, opened 2025-02-03
Refactor auth users the way we retrieve providers

joshenlim:
`raw_app_meta_data` can be anything and can be editable, so we're using the identities table instead to retrieve the providers information


alaister (approved):
LGTM!

- Tested on preview

PR #33262, opened 2025-02-03
fix: users fetch more on scroll

alaister:
Bug fix

https://github.com/user-attachments/assets/9ff7e668-62f7-42b6-91ab-9dd4b295df98

## What is the new behavior?

We now only fetch when there's more pages to fetch. I also have tweaked the close to bottom logic.


alaister:
Closed in favour of https://github.com/supabase/supabase/pull/33269

PR #33265, opened 2025-02-03
docs: Regenerate specs for all API docs references

kamilogorek:
Follow-up to https://github.com/supabase/supabase/pull/33263

ref: https://supabase.slack.com/archives/C0161K73J1J/p1738579423921249


charislam (approved):
Pushed two tiny commits to get rid of the odd `Array <>` when the inner type is not given , as well as this weirdness: 
image

PR #33269, opened 2025-02-03
Fix unnecessary requests in auth users

joshenlim:
Noticed that once you scroll to the bottom of the page in auth/users
- If you scroll horizontally while at the bottom, each scroll triggers a fetchNextPage() call due to handleScroll
- If you scroll a little up, then back to the bottom, there's one unnecessary fetchNextPage() call due to handleScroll

Separately, not yet addressed: Noticed that after loading all the pages, when switching tabs back to the page after a while, the data refreshes but the query is seemingly getting called multiple times unnecessarily

### DIFF
--- a/apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
+++ b/apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
@@ -110,6 +110,7 @@ export const UsersV2 = () => {
     isError,
     isFetchingNextPage,
     refetch,
+    hasNextPage,
     fetchNextPage,
   } = useUsersInfiniteQuery(
     {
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -484,7 +498,6 @@ export const UsersV2 = () => {
                     return (
                       <Row
                         {...props}
-                        key={props.row.id}
```
joshenlim, line 487: this key prop was unnecessary as its handled in ...props already
in fact it was causing some rendering issues when there's a lot of data loaded, and you search in the input field as such:

image

alaister (approved):
LGTM!

- Tested locally

PR #33275, opened 2025-02-03
docs(billing) - add egress usage page

ecktoteckto:


apps/docs/content/guides/platform/manage-your-usage/egress.mdx
```diff
@@ -0,0 +1,114 @@
+---
+id: 'manage-usage-egress'
+title: 'Manage Egress usage'
+---
+
+## What you are charged for
+You are charged for the network data transmitted out of the system to a connected client. Egress is incurred by all services - Database, Authentication, Storage, Edge Functions, Realtime, and Log Drains.
+
+## How charges are calculated
+Egress is charged by byte. Charges apply only for usage exceeding the subscription plan's free quota. This quota can be used across all services (Database, Authentication, Storage etc.), also referred to as Unified Egress.
```
charislam, line 10: The wording of this is a bit confusing. Looking at the examples, I get it, but it's unfortunate that we have a subscription plan called Free because I got confused between "the Free subscription plan's quota" and "the free quota of _this_ subscription plan".

What about:

```suggestion
Egress is charged by byte. Charges apply only for usage exceeding your subscription plan's base quota. This quota is called the Unified Egress Quota because it can be used across all services (Database, Authentication, Storage etc.).
```

Though I don't know if we want to introduce this new terminology or if it clashes with anything else.

charislam (approved):

apps/docs/content/guides/platform/manage-your-usage/egress.mdx
ecktoteckto: Yeah, I went back and forth here many times 😅 

"free" VS "base". We don't have a consistent naming. We use things like "free quota", "quota", "free quota allowance" ... During the revamp I started using "free quota". Fine with "everything" as long as its consistent.

"Unified Egress": we use "Unified Egress" in a few places, but also in an inconsistent way (e.g. in the [FAQs](https://supabase.com/docs/guides/platform/billing-faq#what-is-unified-egress-and-how-is-it-billed))

Will have a look again tomorrow.

kevcodez, line 10: I think what we also need to explain here are the different types of Egress: Database (i.e. Pooler), Storage, Realtime, Auth, API (PostgREST) and give some examples of what contributes to each type

kevcodez (changes requested):
Let's add a breakdown of the different types of egress and give examples of what contributes to it

kevcodez (approved):

PR #33277, opened 2025-02-03
Fix scrolling results in sql editor

saltcod:
When you get a long error in the SQL Editor, you can't scroll the pane its in.

To reproduce: 

1. run this script https://gist.github.com/saltcod/65fc8007f70497e5e6c5fc2993edd15a
2. try and scroll the results pane 

screenshot-2025-02-03-at-15 56 44


jordienr (approved):
tested in staging

PR #33293, opened 2025-02-04
Deploy edge function from Assistant

SaxonF:
Adds a tool to the Assistant that provides knowledge around how to write edge functions for Supabase (based on [this](https://supabase.com/docs/guides/getting-started/ai-prompts/edge-functions)) .

This also adds an edge function block that the Assistant can use to display the edge function. This block contains a deploy action which makes it possible for the user to generate a function and deploy it all from the Assistant.

**Note that this requires https://github.com/supabase/supabase/pull/33371 to be merged first.**

https://github.com/user-attachments/assets/c567ecc2-8eed-4589-b077-f578ceaeb562


apps/studio/pages/api/ai/sql/generate-v3.ts
```diff
@@ -116,6 +116,13 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse) {
       - Default to create or replace whenever possible for updating an existing function, otherwise use the alter function statement
       Please make sure that all queries are valid Postgres SQL queries
 
+      # You write edge functions
+      Your purpose is to generate a edge function with the constraints given by the user.
+      - First, always use the getEdgeFunctionKnowledge tool to get knowledge about how to write edge functions for Supabase
+      - When writing edge functions, always ensure that they are written in TypeScript and Deno JavaScript runtime.
+      - When writing edge functions, always ensure that they are written in a way that is compatible with the database schema.
+      - When suggesting edge functions, dont suggest generic edge functions as shown in the getEdgeFunctionKnowledge tool, create personalised edge functions based on the database schema.
```
laktek, line 124: @SaxonF Can you update this to say "When suggesting edge functions, follow the guidelines in getEdgeFunctionKnowledge tool. Always create personalised edge functions based on the database schema." 

Noticed the function suggested in the screenshot has the supabase dependency wrong - missing `npm:` prefix.

apps/studio/pages/api/ai/sql/tools.ts
```diff
@@ -387,5 +387,179 @@ export const getTools = ({
         }
       },
     }),
+    getEdgeFunctionKnowledge: tool({
+      description: 'Get knowledge about how to write edge functions for Supabase',
+      parameters: z.object({}),
+      execute: async ({}) => {
+        console.log('schemas', 'edge function called...')
```
joshenlim, line 394: remove console.log

apps/studio/components/ui/EdgeFunctionBlock/EdgeFunctionBlock.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  return (
+    <ReportBlockContainer
+      tooltip={tooltip}
+      icon={
+        <Code
+          size={18}
+          strokeWidth={1.5}
+          className={cn(
+            'transition-colors fill-foreground-muted group-aria-selected:fill-foreground',
```
joshenlim, line 76: lucide ones just need `text-n` for the colors - don't need fill

image

apps/studio/components/ui/EdgeFunctionBlock/EdgeFunctionBlock.tsx
```diff
@@ -0,0 +1,135 @@
...
+      </div>
+
+      {(isDeploying || isDeployed) && (
+        <div className="p-4 w-full border-t bg-surface-75 text-xs">
+          {isDeploying ? (
+            <p className="text-foreground-light">Deploying function...</p>
+          ) : (
+            <>
+              <p className="text-foreground-light mb-2">
+                Function deployed successfully! You can now invoke it at:
```
joshenlim, line 118: just a feedback - wondering to include CLI commands, so that users can directly copy paste into the terminal to run it - from the POV of someone not familiar with edge functions im not too sure what to do with this URL

image

apps/studio/components/ui/EdgeFunctionBlock/EdgeFunctionBlock.tsx
```diff
@@ -0,0 +1,135 @@
...
+        files: [{ name: 'index.ts', content: code }],
+      })
+    } catch (error) {
+      toast.error(
+        `Failed to deploy function: ${error instanceof Error ? error.message : 'Unknown error'}`
+      )
+    }
+  }
+
+  const functionUrl = ref ? `https://${ref}.supabase.red/functions/v1/${functionName}` : ''
```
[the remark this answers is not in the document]
joshenlim: okay yeap can verify that its `.co` on prod
image

joshenlim:
thought here: the "Deploy" button reverts back to the original state if i refresh the page - it doesn't have context that it's a linked to a function that's been deployed, just wondering if it might be a source of confusion?
image

apps/studio/components/ui/EdgeFunctionBlock/EdgeFunctionBlock.tsx
joshenlim: thinking also maybe provide a link to the edge function on the dashboard?

apps/studio/components/ui/EdgeFunctionBlock/EdgeFunctionBlock.tsx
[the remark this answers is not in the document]
joshenlim: For reference we do it here:
https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Functions/TerminalInstructions.tsx#L48

apps/studio/pages/api/ai/sql/tools.ts
```diff
@@ -387,5 +387,179 @@ export const getTools = ({
...
+        return stripIndent`
+        # Writing Supabase Edge Functions
+
+        You're an expert in writing TypeScript and Deno JavaScript runtime. Generate **high-quality Supabase Edge Functions** that adhere to the following best practices:
+
+        ## Guidelines
+
+        1. Try to use Web APIs and Deno's core APIs instead of external dependencies (eg: use fetch instead of Axios, use WebSockets API instead of node-ws)
+        2. If you are reusing utility methods between Edge Functions, add them to \`supabase/functions/_shared\` and import using a relative path. Do NOT have cross dependencies between Edge Functions.
+        3. Do NOT use bare specifiers when importing dependecnies. If you need to use an external dependency, make sure it's prefixed with either \`npm:\` or \`jsr:\`. For example, \`@supabase/supabase-js\` should be written as \`npm:@supabase/supabase-js\`.
```
gregnr, line 404:
```suggestion
        3. Do NOT use bare specifiers when importing dependencies. If you need to use an external dependency, make sure it's prefixed with either \`npm:\` or \`jsr:\`. For example, \`@supabase/supabase-js\` should be written as \`npm:@supabase/supabase-js\`.
```

apps/studio/pages/api/ai/sql/tools.ts
```diff
@@ -387,5 +387,179 @@ export const getTools = ({
...
+        console.log('schemas', 'edge function called...')
+        return stripIndent`
+        # Writing Supabase Edge Functions
+
+        You're an expert in writing TypeScript and Deno JavaScript runtime. Generate **high-quality Supabase Edge Functions** that adhere to the following best practices:
+
+        ## Guidelines
+
+        1. Try to use Web APIs and Deno's core APIs instead of external dependencies (eg: use fetch instead of Axios, use WebSockets API instead of node-ws)
+        2. If you are reusing utility methods between Edge Functions, add them to \`supabase/functions/_shared\` and import using a relative path. Do NOT have cross dependencies between Edge Functions.
```
gregnr, line 403: Might be worth removing for this version to avoid confusing the model.

apps/studio/pages/api/ai/sql/tools.ts
```diff
@@ -387,5 +387,179 @@ export const getTools = ({
...
+        3. Do NOT use bare specifiers when importing dependecnies. If you need to use an external dependency, make sure it's prefixed with either \`npm:\` or \`jsr:\`. For example, \`@supabase/supabase-js\` should be written as \`npm:@supabase/supabase-js\`.
+        4. For external imports, always define a version. For example, \`npm:@express\` should be written as \`npm:express@4.18.2\`.
```
gregnr, line 405: This is good practice, though I wonder how well the model will get these versions correct along with the features in that version. At the very least we might want to scope to just major version.

cc @laktek

apps/studio/pages/api/ai/sql/tools.ts
```diff
@@ -387,5 +387,179 @@ export const getTools = ({
...
+        5. For external dependencies, importing via \`npm:\` and \`jsr:\` is preferred. Minimize the use of imports from @\`deno.land/x\` , \`esm.sh\` and @\`unpkg.com\` . If you have a package from one of those CDNs, you can replace the CDN hostname with \`npm:\` specifier.
+        6. You can also use Node built-in APIs. You will need to import them using \`node:\` specifier. For example, to import Node process: \`import process from "node:process"\`. Use Node APIs when you find gaps in Deno APIs.
+        7. Do NOT use \`import { serve } from "https://deno.land/std@0.168.0/http/server.ts"\`. Instead use the built-in \`Deno.serve\`.
+        8. Following environment variables (ie. secrets) are pre-populated in both local and hosted Supabase environments. Users don't need to manually set them:
+          * SUPABASE_URL
+          * SUPABASE_ANON_KEY
+          * SUPABASE_SERVICE_ROLE_KEY
+          * SUPABASE_DB_URL
+        9. To set other environment variables (ie. secrets) users can put them in a env file and run the \`supabase secrets set --env-file path/to/env-file\`
```
gregnr, line 414: Do we have any strategies around this via UI? Link them to the secrets settings page? Even better: inline chat component to set secrets?

apps/studio/pages/api/ai/sql/tools.ts
```diff
@@ -387,5 +387,179 @@ export const getTools = ({
...
+          );
+        });
+        \`\`\`
+
+        ### Example Function using Node built-in API
+
+        \`\`\`edge
+        import { randomBytes } from "node:crypto";
+        import { createServer } from "node:http";
+        import process from "node:process";
```
gregnr, line 448: @laktek should we suggest importing edge runtime types at the top?

```
// Setup type definitions for built-in Supabase Runtime APIs
import "jsr:@supabase/functions-js/edge-runtime.d.ts";
```

apps/studio/pages/api/ai/sql/tools.ts
```diff
@@ -387,5 +387,179 @@ export const getTools = ({
...
+              return new Response(\\'ok\\', { headers: corsHeaders })
+            }
+
+            try {
+              // Create a Supabase client with the Auth context of the logged in user.
+              const supabaseClient = createClient(
+                // Supabase API URL - env var exported by default.
+                Deno.env.get(\\'SUPABASE_URL\\') ?? \\'\\',
+                // Supabase API ANON KEY - env var exported by default.
+                Deno.env.get(\\'SUPABASE_ANON_KEY\\') ?? \\'\\',
```
gregnr, line 521: Why fall back to empty string? Better to surface missing vars as soon as possible right?

apps/studio/components/ui/EdgeFunctionBlock/EdgeFunctionBlock.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const handleDeploy = async () => {
+    if (!code || isDeploying || !ref) return
+
+    try {
+      const deployResult = await deployFunction({
+        projectRef: ref,
+        metadata: {
+          entrypoint_path: 'index.ts',
+          name: functionName,
+          verify_jwt: false,
```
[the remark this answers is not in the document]
laktek: yep, let's keep it as true (our instructions to invoke a function will include the the Authorization header)

apps/studio/pages/api/ai/sql/tools.ts
laktek: In my tests (openai, supbase-js and express), the model ended up using a major version or a more common one. As long it appends a version it's okay, at least it gives the user a more definite path for debugging issues (rather not having a version declared and using unspecified API methods)

apps/studio/pages/api/ai/sql/tools.ts
laktek: This is copied from the general AI prompt (which may not always have access to Studio UI). But I suggested to @SaxonF in the Studio specific prompt which should suggest to use the Secrets UI over this (can we override this rule in Studio assistant?).

apps/studio/pages/api/ai/sql/tools.ts
laktek: yep good call!

apps/studio/pages/api/ai/sql/tools.ts
laktek: @SaxonF is this an example you added? Maybe the original example had empty strings, but I agree better to let it leave as undefined so it throws an error in case the env vars are missing.

apps/studio/pages/api/ai/sql/tools.ts
laktek: @SaxonF Can you add this line to all examples here?

joshenlim (approved):

PR #33294, opened 2025-02-04
Use portal for reference record peek

joshenlim:
As per PR title, solves an issue whereby clicking on "View referencing record" somehow scrolls the table editor to the left most (For reference - we had to use portal for tooltips in the table editor too cause of how react data grid is)


alaister (approved):
LGTM!

- Tested on preview

PR #33304, opened 2025-02-04
chore: static assets cdn

alaister:
This change introduces a new way to deal with static build assets and public files for our hosted frontends. In the past, we have had static assets on Vercel directly (with high charges) or enabled Cloudflare proxying, which added ~20ms of latency for each request due to double SSL termination and even worse, keeps leading to super random timeouts. These are most likely Orange-To-Orange issues (Cloudflare fighting itself), which we have seen many times in the past.

With this change, we are going for the least invasive solution and only using a CDN when deploying to production. Local dev & preview environments are not changing in any way.

We're using Cloudflare R2 as CDN. By using a CDN, we can serve static assets extremely fast while saving big time on egress costs. The `build` command has been modified to not only build our app, but also upload static assets to the CDN.

By using a dynamic path including the env, app and commit hash, we can ensure that there are no conflicts.
Static assets from previous deployments stick around for a while to ensure there are no "downtimes".

**Advantages of the CDN approach we're using**:

- Get rid of egress costs for static assets across our apps on Vercel
- Disable CF proxying and get around these odd timeouts issues
- Save ~20ms or so for asset requests, as there is no additional CF proxying and we avoid terminating SSL twice
- Always hits the CDN, gonna be super quick
- Does not run on local or preview environments, only on staging/prod deployments
- There are no other disadvantages - you don't have to consider it when developing locally, previews still work, everything on Vercel works as we're used to
- We don't lose any advantages of the Vercel platform here


scripts/upload-static-assets.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# Check for force env var or production environment
+if [[ "$FORCE_ASSET_CDN" != "1" ]] && [[ "$VERCEL_ENV" != "production" ]]; then
```
ivasilov, line 4: This script will need to finish before Vercel prod build for studio, right? When restoring from cache, the Vercel build can take ~1 minute.

Should we bother syncronizing them?

alaister: I think the deployment waits on the upload to finish before it switches things over. Check out the build logs here: https://vercel.com/supabase/studio-staging/9dJWq5UAFM6Eg8Yp8xTNvCq791j4

Is that what you mean by synchronising them?

ivasilov: Ah, the upload happens by a Vercel worker, I hallucinated that it's by a Github worker, sorry.

scripts/upload-static-assets.sh
```diff
@@ -0,0 +1,100 @@
...
+if [[ "$FORCE_ASSET_CDN" != "-1" ]]; then
+    echo "Skipping asset upload. Set FORCE_ASSET_CDN=1 or VERCEL_ENV=production to execute."
+    exit 0
+fi
+
+
+if [[ "$FORCE_ASSET_CDN" != "1" ]] && [[ "$VERCEL_ENV" != "production" ]]; then
+    echo "Skipping asset upload. Set FORCE_ASSET_CDN=1 or VERCEL_ENV=production to execute."
+    exit 0
+fi
```
ivasilov, line 35: You have this check twice?

kevcodez: whoops

kevcodez: Fixed

apps/studio/next.config.js
```diff
@@ -1,4 +1,5 @@
 const { withSentryConfig } = require('@sentry/nextjs')
+const { processDataProtocolResponse } = require('ai')
```
ivasilov, line 2: Remove the import?

kevcodez: Done

ivasilov (approved):

apps/studio/next.config.js
```diff
@@ -51,16 +51,20 @@ const GOOGLE_USER_AVATAR_URL = 'https://lh3.googleusercontent.com'
...
 const PUSHER_URL_WS = 'wss://*.pusher.com'
 
-const DEFAULT_SRC_URLS = `${API_URL} ${SUPABASE_URL} ${GOTRUE_URL} ${SUPABASE_LOCAL_PROJECTS_URL_WS} ${SUPABASE_PROJECTS_URL} ${SUPABASE_PROJECTS_URL_WS} ${HCAPTCHA_SUBDOMAINS_URL} ${CONFIGCAT_URL} ${STRIPE_SUBDOMAINS_URL} ${STRIPE_NETWORK_URL} ${CLOUDFLARE_URL} ${ONE_ONE_ONE_ONE_URL} ${VERCEL_INSIGHTS_URL} ${GITHUB_API_URL} ${GITHUB_USER_CONTENT_URL}`
-const SCRIPT_SRC_URLS = `${CLOUDFLARE_CDN_URL} ${HCAPTCHA_JS_URL} ${STRIPE_JS_URL}`
+const DEFAULT_SRC_URLS = `${API_URL} ${SUPABASE_URL} ${GOTRUE_URL} ${SUPABASE_LOCAL_PROJECTS_URL_WS} ${SUPABASE_PROJECTS_URL} ${SUPABASE_PROJECTS_URL_WS} ${HCAPTCHA_SUBDOMAINS_URL} ${CONFIGCAT_URL} ${STRIPE_SUBDOMAINS_URL} ${STRIPE_NETWORK_URL} ${CLOUDFLARE_URL} ${ONE_ONE_ONE_ONE_URL} ${VERCEL_INSIGHTS_URL} ${GITHUB_API_URL} ${GITHUB_USER_CONTENT_URL} ${SUPABASE_ASSETS_URL}`
+const SCRIPT_SRC_URLS = `${CLOUDFLARE_CDN_URL} ${HCAPTCHA_JS_URL} ${STRIPE_JS_URL} ${SUPABASE_ASSETS_URL}`
 const FRAME_SRC_URLS = `${HCAPTCHA_ASSET_URL} ${STRIPE_JS_URL}`
 const IMG_SRC_URLS = `${SUPABASE_URL} ${SUPABASE_COM_URL} ${SUPABASE_PROJECTS_URL} ${GITHUB_USER_AVATAR_URL} ${GOOGLE_USER_AVATAR_URL}`
```
staaldraad, line 66: to check, we aren't putting images into the cdn?

staaldraad (approved):
one double-check, but otherwise lgtm

PR #33322, opened 2025-02-04
feat: AI Edge function template

ivasilov:
Fixes FE-1376.

### DIFF
--- a/apps/studio/data/fetchers.ts
+++ b/apps/studio/data/fetchers.ts
@@ -7,7 +7,6 @@ import { ResponseError } from 'types'
 import type { paths } from './api' // generated from openapi-typescript
 
 const DEFAULT_HEADERS = {
-  'Content-Type': 'application/json',
   Accept: 'application/json',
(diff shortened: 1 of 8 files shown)


apps/studio/data/fetchers.ts
```diff
@@ -7,7 +7,6 @@ import { ResponseError } from 'types'
 import type { paths } from './api' // generated from openapi-typescript
 
 const DEFAULT_HEADERS = {
-  'Content-Type': 'application/json',
```
laktek, line 10: will this cause issues for other fetch requests? Maybe better to use a regular `fetch` for deployFunction so it doesn't use DEFAULT_HEADERS.

ivasilov: It's not an issue, `openapi-fetch` sends `application/json` by default when `bodySerializer` is not used. [Docs](https://openapi-ts.dev/openapi-fetch/api#bodyserializer)

PR #33332, opened 2025-02-04
fix(ref docs): make unions more intuitive

charislam:
Before:
image

After:
image


hf (approved):
Charis to save the day!

PR #33337, opened 2025-02-05
fix: use vault functions for creating/updating secrets

soedirgo:

YES

Bug fix

We're inserting/updating vault secrets directly into the table with `INSERT`/`UPDATE`. This is not part of the Vault API because some of the columns are internal implementation.

## What is the new behavior?

Use Vault functions instead


alaister (approved):
LGTM!

- Tested on preview

PR #33342, opened 2025-02-05
Fix download CSV and copy as markdown for JSON data

joshenlim:
Fixes FE-1375


alaister (approved):
LGTM!

- Tested on preview

PR #33346, opened 2025-02-05
Support resetting query performance report on read replica

joshenlim:
If a read replica is selected as such:
image

Hitting "Reset report" will show the Confirmation modal as such
image

Likewise, if its the primary database selected:
image


alaister (approved):
LGTM!

- Tested on preview

PR #33352, opened 2025-02-05
Only send required data for reset password, magic link and send OTP

joshenlim:
Auth/users: Hitting "Send password recovery", "Send Magic link" (and Send OTP probably) shows an error "property providers should not exist"
image

We recently added a new property when fetching users to display on that page, but because we're sending the whole user JSON to those calls, the API is hence throwing an error. We don't need to send the whole user JSON actually, just the required data


alaister (approved):
LGTM!

- Tested on preview

PR #33355, opened 2025-02-05
Check for value before converting bytea to hex in ForeignKeyFormatter

joshenlim:
As per title - causes an issue if the column is bytea data type, has a foreign key, and is NULL


saltcod (approved):

saltcod:
checked preview, looks good!

PR #33369, opened 2025-02-05
fix(admonitions): underline links

charislam:
Before:
image

After:
image


apps/docs/features/docs/MdxBase.shared.tsx
```diff
@@ -46,6 +46,7 @@ const components = {
   Accordion,
   AccordionItem,
   Admonition,
+  // Admonition: (props: AdmonitionProps) => <Admonition className="[&_a]:underline" {...props} />,
```
charislam, line 49: There seemed to be some caching problem when I updated ui-patterns without also updating docs, so I'm leaving this non-change in here for now. Can delete it pretty much as soon as the PR is merged.

ivasilov (approved):

PR #33371, opened 2025-02-06
deploy edge function mutaton

SaxonF:
This PR adds a deploy edge function mutation to go alongside update and delete. It can be used to deploy an edge function directly from the dashboard. 

It opts for a custom fetch instead of data/fetchers post in order to remove `'Content-Type': 'application/json',` without affecting other requests.

### DIFF
--- a/apps/studio/data/edge-functions/edge-functions-deploy-mutation.ts
+++ b/apps/studio/data/edge-functions/edge-functions-deploy-mutation.ts
@@ -1,8 +1,7 @@
 import { useMutation, UseMutationOptions, useQueryClient } from '@tanstack/react-query'
 import { toast } from 'sonner'
 
-import { handleError, constructHeaders } from 'data/fetchers'
-import { API_URL } from 'lib/constants'
+import { handleError, post } from 'data/fetchers'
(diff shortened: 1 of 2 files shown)


apps/studio/data/edge-functions/edge-functions-deploy-mutation.ts
```diff
@@ -0,0 +1,87 @@
...
+  const formData = new FormData()
+  formData.append('metadata', JSON.stringify(metadata))
+
+  files.forEach((file) => {
+    const blob = new Blob([file.content], { type: 'text/plain' })
+    formData.append('file', blob, file.name)
+  })
+
+  const response = await fetch(
+    `${API_URL?.replace('/platform', '')}/v1/projects/${projectRef}/functions/deploy?slug=${metadata.name}`,
```
joshenlim, line 37: do we need this part tho? `API_URL?.replace('/platform', '')`

we can't just do `/v1`? Similar to https://github.com/supabase/supabase/blob/master/apps/studio/data/edge-functions/edge-functions-query.ts#L18

ivasilov (approved):

PR #33375, opened 2025-02-06
Create MultiSelector with creatable flag

joshenlim:
Adds a `creatable` flag to `MultiSelectorList` component

https://github.com/user-attachments/assets/501169de-3cd2-47cc-9bb0-d3baf88015e3

JFYI there's an existing bug with the `MultiSelector` component that I can't seem to figure out - if the options are not sorted alphabetically, and you enter a value into the search input then clear it, the options somehow keep ordering themselves alphabetically - it's also messing with the Create n CTA if the `creatable` flag is true too

https://github.com/user-attachments/assets/8c8f129c-43df-41e2-a8d3-582594ecb0a1

### DIFF
--- a/apps/design-system/__registry__/index.tsx
+++ b/apps/design-system/__registry__/index.tsx
@@ -1919,6 +1919,17 @@ export const Index: Record<string, any> = {
       subcategory: "undefined",
       chunks: []
     },
+    "multi-select-combobox-creatable": {
+      name: "multi-select-combobox-creatable",
+      type: "components:example",
+      registryDependencies: undefined,
+      component: React.lazy(() => import("@/registry/default/example/multi-select-combobox-creatable")),
+      source: "",
+      files: ["registry/default/example/multi-select-combobox.tsx"],
+      category: "undefined",
+      subcategory: "undefined",
+      chunks: []
+    },
     "multi-select-deletable-badge": {
       name: "multi-select-deletable-badge",
       type: "components:example",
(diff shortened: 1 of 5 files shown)


fsansalvadore:
@joshenlim I can't replicate the ordering bug you mention. On which browser do you see it?
(I can look at that bug separately)

Just some feedback:
- the "create abc" command item should probably always be first (or last) and not ordered with the other items?
- we should probably provide a way to forward an "onCreate" function that you can pass as a prop?
- the new item should probably also appear as an item in the dropdown, once added?

joshenlim:
> the "create abc" command item should probably always be first (or last) and not ordered with the other items?

Yup i think it should be the last! but i think its currently slotted in the middle sometimes because of the reordering bug that's happening

> we should probably provide a way to forward an "onCreate" function that you can pass as a prop?

was thinking about this actually, but maybe let's hold this one off for now as I can't think of a use case where i'd want the new option to be created once i select it. chances are i might want to only create it when i save the form in a parent component

> the new item should probably also appear as an item in the dropdown, once added?

am thinking we don't add it as a new item actually - am referring to this as a prior art: https://shadcnui-expansions.typeart.cc/docs/multiple-selector#Creatable%20Selector

So we let the parent component decide whether to add the new item as a new MultiSelectorItem instead

fsansalvadore:
> chances are i might want to only create it when i save the form in a parent component

Got it, then we'll consider adding it if we ever see actual the need for it 👍

fsansalvadore:
@joshenlim btw the shadcn prior art you shared has a similar ordering bug, I wonder if it has something to do with the underlying radix dialog component..

fsansalvadore (approved):
Lgtm

PR #33405, opened 2025-02-07
track source when viewing plans

SaxonF:
Adds an extra source property to STUDIO_PRICING_SIDE_PANEL_OPENED event which we use to track pricing plan views. This will let us analyse and optimise which upgrade prompts are driving people to view plans and upgrade.


pamelachia:
How do you plan to handle or say the source if user just clicks the button to open at billing page or when user types ?panel=subscriptionPlan directly? 

And add a description for what source means or rename it to something else in telemetry-constants because we use source to describe whether this is an event from www, studio or docs

packages/common/telemetry-constants.ts
```diff
@@ -1069,7 +1069,7 @@ export interface StudioPricingPlanCtaClickedEvent {
  */
 export interface StudioPricingSidePanelOpenedEvent {
   action: TelemetryActions.STUDIO_PRICING_SIDE_PANEL_OPENED
-  properties: { currentPlan?: string }
+  properties: { currentPlan?: string; source?: string }
```
loong:
```suggestion
  properties: { 
    currentPlan?: string,
    
    /**
     * Tracks how user landed on the Pricing side panel, e.g. computeDiskManagement, restoreToNewProject
     */
    source?: string 
  }
```

apps/studio/components/interfaces/Organization/BillingSettings/Restriction.tsx
```diff
@@ -97,7 +97,11 @@ export const Restriction = () => {
             </p>
             <div className="flex items-center gap-x-2 mt-3">
               <Button asChild key="upgrade-button" type="default">
-                <Link href={`/org/${org?.slug}/billing?panel=subscriptionPlan`}>Upgrade plan</Link>
+                <Link
+                  href={`/org/${org?.slug}/billing?panel=subscriptionPlan&source=gracePeriodStarted`}
```
loong, line 101:
```suggestion
                  href={`/org/${org?.slug}/billing?panel=subscriptionPlan&source=fairUseGracePeriodStarted`}
```

apps/studio/components/interfaces/Organization/BillingSettings/Restriction.tsx
```diff
@@ -124,7 +128,11 @@ export const Restriction = () => {
             </p>
             <div className="flex items-center gap-x-2 mt-3">
               <Button key="upgrade-button" asChild type="default">
-                <Link href={`/org/${org?.slug}/billing?panel=subscriptionPlan`}>Upgrade plan</Link>
+                <Link
+                  href={`/org/${org?.slug}/billing?panel=subscriptionPlan&source=gracePeriodOver`}
```
loong, line 132:
```suggestion
                  href={`/org/${org?.slug}/billing?panel=subscriptionPlan&source=fairUseGracePeriodOver`}
```

apps/studio/components/interfaces/Organization/BillingSettings/Restriction.tsx
```diff
@@ -151,7 +159,9 @@ export const Restriction = () => {
             </p>
             <div className="flex items-center gap-x-2 mt-3">
               <Button key="upgrade-button" asChild type="default">
-                <Link href={`/org/${org?.slug}/billing?panel=subscriptionPlan`}>Upgrade plan</Link>
+                <Link href={`/org/${org?.slug}/billing?panel=subscriptionPlan&source=restricted`}>
```
loong, line 162:
```suggestion
                <Link href={`/org/${org?.slug}/billing?panel=subscriptionPlan&source=fairUseRestricted`}>
```

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/BillingMetric.tsx
```diff
@@ -143,7 +143,11 @@ const BillingMetric = ({
           ) : (
             <div>
               <Button type="default" asChild>
-                <Link href={`/org/${slug}/billing?panel=subscriptionPlan`}>Upgrade</Link>
+                <Link
+                  href={`/org/${slug}/billing?panel=subscriptionPlan&source=usage${metric.anchor}`}
```
loong, line 147:
```suggestion
                  href={`/org/${slug}/billing?panel=subscriptionPlan&source=billingBreakdownUsage${metric.anchor}`}
```

SaxonF:
@pamelachia I renamed to `upgradeOrigin` and only added source if it exists. If no source exists we can assume coming to the page directly or via billing page. I can add tracking to billing page CTA as a follow up.

loong (commented):
Would be good to make some of the source names more descriptive and add documentation.

packages/common/telemetry-constants.ts
loong: @SaxonF Ah started review before lunch and submitted after, so some of it is out of date now, but you get the gist!

apps/studio/components/interfaces/Organization/BillingSettings/Restriction.tsx
loong: Adding fair use here, just to make clear these are from the fair use / service restriction notes. Feel free to change if you can think of a better name

SaxonF:
@loong committed your suggestions and added comment to event definition

staaldraad (approved):
approved for changes to security owned pages

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
```diff
@@ -133,7 +133,7 @@ const PlanUpdateSidePanel = () => {
         currentPlan: subscription?.plan?.name,
       }
       if (source) {
-        properties.upgradeOrigin = source
+        properties.origin = source
```
loong, line 136: nitpick, not calling it upgrade, since users technically haven't upgraded yet, this just leads to the plan update panel

loong (approved):
Pushed a commit with some changes to this branch, so make sure to pull/rebase if you want to make any more changes!

Did some more renaming of the sources/origin just to give more context. I often use the filename to give more hints where these origins are from. Not super opinionated, feel free to change it as you see fit.

There are some `UpgradeToPro` buttons that don't have the `source` field set. Not a big deal if we don't cover everything just making sure in case you are not aware.

Really cool we have this info now. Curious what we can learn from it 🔥

joshenlim (approved):
LGTM - tested on staging preview on a free org for database backups - can verify that additional properties "origin" is appended to sendEvent call

PR #33407, opened 2025-02-07
Hide org members pending invite acceptance in org audit logs users filter

joshenlim:
Invited members who have yet to accept the invite would show up as a single alphabet here
image


alaister (approved):
LGTM!

- Tested on preview

staaldraad (approved):
LGTM! super fast turn around, thanks!🥇

PR #33408, opened 2025-02-07
feat: additional billing email addresses

kevcodez:
Customers can now set up additional billing email addresses themselves without having to reach out to us. This covers the following emails thus far:
- Invoice emails
- Payment failure emails

Dunning emails will still go out to all members to ensure visibility. The only thing this does not cover is payment receipts from Stripe, as Stripe still provides no API for setting additional emails (dashboard only for many years now)

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/BillingEmail.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/BillingEmail.tsx
@@ -160,7 +160,9 @@ const BillingEmail = () => {
                             <div className="flex items-center gap-x-1">
                               <span>Additional emails</span>
                               <InfoTooltip side="bottom">
-                                These email addresses will be CC'd in automated invoice emails.
+                                These email addresses will be CC'd in automated invoice or payment
+                                failure emails. Payment receipts will still only go to the primary
+                                billing address.
                               </InfoTooltip>
                             </div>
                           }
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Organization/BillingSettings/BillingEmail.tsx
```diff
@@ -70,51 +103,82 @@ const BillingEmail = () => {
...
+                      </FormItem>
+                    )}
+                  />
+
+                  <FormField
+                    control={form.control}
+                    name="additionalBillingEmails"
+                    render={({ field }) => (
+                      <FormItem>
+                        <FormLabel>Additional emails</FormLabel>
```
joshenlim, line 156: probably good to add a description what additional emails do? will they be CCed in the same email or all of these emails will also receive the same emails as the above email address? (mainly to address potential confusion between how the addresses here differ from the above one)

kevcodez: planning on adding a tooltip

staaldraad (approved):

joshenlim:
hold this PR for now - FYI there was an issue with the MultiSelect component that broke the Data API settings page, will be looking into it

alaister (approved):
LGTM!

- Tested on preview

PR #33411, opened 2025-02-07
Fix move query to new folder

joshenlim:
Client error occurs when moving a query to a new folder via the modal:
image


alaister (approved):
LGTM!

- Tested on preview

PR #33422, opened 2025-02-07
Chore/assistant in self hosted

saltcod:
fixes https://linear.app/supabase/issue/FE-1357/make-the-ai-assistant-available-in-local-environment


joshenlim (approved):
LGTM! Verified that assistant can be used on
- studio-self-hosted
- studio-staging preview
- local self-hosted

PR #33428, opened 2025-02-07
Billing docs revamp - Third-Party MAU usage page

ecktoteckto:
Left out the "Optimize usage" section on purpose. The user opts into this feature, so it doesn't make sense to tell them to save costs by not using the feature and instead migrate Users to Supabase Auth or translate JWTs.


charislam (approved):

apps/docs/content/guides/platform/manage-your-usage/monthly-active-users-third-party.mdx
```diff
@@ -0,0 +1,63 @@
+---
+id: 'manage-usage-monthly-active-users-third-party'
+title: 'Manage Monthly Active Third-Party Users usage'
+---
+
+## What you are charged for
+You are charged for the number of distinct users who log in or refresh their token during the billing cycle using a third-party authentication provider. Each unique user is counted only once per billing cycle, regardless of how many times they authenticate. These users are referred to as "Third-Party MAUs".
+
+## How charges are calculated
+You are charged by Third-Party MAU.
```
kevcodez, line 10: This should contain some example requests to explain what counts as a third-party auth activity and that the same user only counts once - otherwise this section is a bit useless

Issue #30587, opened 2024-11-21
Spaces are squeezed from column names when calling .select()


rickwargo:
I am unable to specify column names that have spaces (an possibly other things) in a fetch statement. I do have a workaround, and that is to literally quote the column name in the args.

In the following example (yes, I know it just returns a bunch of nulls, if it worked), the generated API call fails when `column` contains a space.

```Javascript
      const {data, error} = await supabase
        .from(table_name)
        .select(column)
        .is(column, null);
```

For example, assuming the database had a table named "My Table" with a nullable column named "My Column", and the query specified table_name = "My Table" and column = "My Column", the resulting REST query is

```
https://xxxxxx.supabase.co/rest/v1/My%20Table?select=MyColumn&My+Column=is.null
```

The table name is correctly interpreted as case-sensitive with spaces, however, for the returned data, the column name is stripped of its space and it returns a **400 Bad Request** HTTP error.

The error is:
```
{
  "code": "42703",
  "details": null,
  "hint": 'Perhaps you meant to reference the column "My Table.My Column".',
  "message": "column My Table.MyColumn does not exist"
}
```

I did not find in the API documentation where this behavior is specified.

tengweiherr:
I faced a similar issue. In my case, I named my table "Base Salary", and when fetching data with the Supabase JS client, it removed the space in the table name in the query, resulting in an error.

Here is my code:
```
const { data, error } = await supabase
  .from('User Career')
  .select(`
      id, 
      jobTitle, 
      Base Salary (
        amount
      )
  `)
```

The error response when I performed the SELECT query:
```
{
    "code": "PGRST200",
    "details": "Searched for a foreign key relationship between 'User Career' and 'BaseSalary' in the schema 'public', but no matches were found.",
    "hint": "Perhaps you meant 'Base Salary' instead of 'BaseSalary'.",
    "message": "Could not find a relationship between 'User Career' and 'BaseSalary' in the schema cache"
}
```

After digging into the source code, I found that in the `PostgrestQueryBuilder` of [postgrest.js](https://github.com/supabase/postgrest-js/tree/master), it actually [removes whitespaces unless quoted](https://github.com/supabase/postgrest-js/blob/master/src/PostgrestQueryBuilder.ts#L71-L82). Hence, to keep the space in the table or the column name, you need to add quotes around it. For example:

``` 
const { data, error } = await supabase
  .from('User Career')
  .select(`
      id, 
      jobTitle, 
      "Base Salary" (
        amount
[...]

PR #33443, opened 2025-02-08
docs(supabase-js): add example to clarify how to fetch data with spaces in table name

tengweiherr:

YES

docs update

to avoid confusion like https://github.com/supabase/supabase/issues/30587

Some users (including myself) ran into issues when our tables were named with spaces, such as "Base Salary". When fetching data using the Supabase JS client, the space in the table name was removed from the `select` query, causing an error as it is now "BaseSalary" in the query output.

After digging into the source code, I found that the `PostgrestQueryBuilder` in [postgrest.js](https://github.com/supabase/postgrest-js/tree/master) actually [removes whitespaces unless quoted](https://github.com/supabase/postgrest-js/blob/master/src/PostgrestQueryBuilder.ts#L71-L82). Therefore, to keep spaces in table or column names in the `select` statement, we need to enclose them explicitly in **double quotes**.

## What is the new behavior?

Added an example to clarify how to fetch data when there are spaces in table names

## Additional context

*Video example*

https://github.com/user-attachments/assets/3870efac-3313-46cb-986c-23336dd55d44


charislam (approved):
Thanks @tengweiherr!

tengweiherr:
@charislam thank you for approving the PR. However, seems like the workflow [Preview – zone-www-dot-com](https://zone-www-dot-347bm1fa1-supabase.vercel.app/) failed for some reasons. Let me know how can I fix it :)

PR #33472, opened 2025-02-10
Remove 2dp for ms values in query performance advisor

joshenlim:
Fixes FE-1321

As per PR title

Before:
image

After:
image


alaister (approved):
LGTM!

- Tested on preview

PR #33473, opened 2025-02-10
Fix text overlapping in PublicSchemaNotEnabledAlert

joshenlim:
Fixes FE-1297

As per PR title

Before:
image

After:
image

### DIFF
--- a/apps/studio/components/interfaces/Database/Schemas/SchemaGraph.tsx
+++ b/apps/studio/components/interfaces/Database/Schemas/SchemaGraph.tsx
@@ -157,7 +157,7 @@ export const SchemaGraph = () => {
               deletable: false,
               style: {
                 stroke: 'hsl(var(--border-stronger))',
-                strokeWidth: 0.5,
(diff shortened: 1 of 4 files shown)


apps/studio/components/interfaces/Database/Schemas/SchemaGraph.tsx
```diff
@@ -157,7 +157,7 @@ export const SchemaGraph = () => {
               deletable: false,
               style: {
                 stroke: 'hsl(var(--border-stronger))',
-                strokeWidth: 0.5,
+                strokeWidth: 1,
```
joshenlim, line 160: unrelated change here but its just to bring up the contrast of the schema graph lines for foreign keys (there were some feedback about this)

alaister (approved):
LGTM!

- Tested on preview

Small nit: I did notice this box flashes up for a second if you do have the public schema exposed. Since the default is having the public schema exposed it would be nice if this didn't flash.

PR #33479, opened 2025-02-10
Edge function redirect if slug doesnt exist

joshenlim:
As per PR title, can test by landing the URL: supabase.green/project/_/functions/blah


jordienr (approved):
tested on preview

PR #33480, opened 2025-02-10
chore(pg-meta): port functions tests

avallete:

Continue the effort to migrate pg-meta into a self serving query lib

This PR:

1. Port over the tests for the "functions" query

### DIFF
--- a/packages/pg-meta/src/pg-meta-functions.ts
+++ b/packages/pg-meta/src/pg-meta-functions.ts
@@ -4,30 +4,6 @@ import { DEFAULT_SYSTEM_SCHEMAS } from './constants'
 import { filterByList } from './helpers'
 import { FUNCTIONS_SQL } from './sql/functions'
 
-export type PGFunction = {
-  id: number
-  schema: string
-  name: string
-  language: string
-  definition: string
-  complete_statement: string
-  args: Array<{
-    mode: 'in' | 'out' | 'inout' | 'variadic' | 'table'
-    name: string
-    type_id: number
-    has_default: boolean
-  }>
-  argument_types: string
-  identity_argument_types: string
-  return_type_id: number
-  return_type: string
-  return_type_relation_id: number | null
-  is_set_returning_function: boolean
-  behavior: 'IMMUTABLE' | 'STABLE' | 'VOLATILE'
-  security_definer: boolean
-  config_params: Record<string, string> | null
-}
-
 export const pgFunctionZod = z.object({
   id: z.number(),
   schema: z.string(),
@@ -58,7 +34,9 @@ export const pgFunctionZod = z.object({
   behavior: z.union([z.literal('IMMUTABLE'), z.literal('STABLE'), z.literal('VOLATILE')]),
   security_definer: z.boolean(),
   config_params: z.union([z.record(z.string(), z.string()), z.null()]),
-}) satisfies z.ZodType<PGFunction>
+})
+
+export type PGFunction = z.infer<typeof pgFunctionZod>
 
 export const pgFunctionArrayZod = z.array(pgFunctionZod)
 export const pgFunctionOptionalZod = z.optional(pgFunctionZod)
--- a/packages/pg-meta/src/pg-meta-functions.ts
+++ b/packages/pg-meta/src/pg-meta-functions.ts
@@ -4,6 +4,30 @@ import { DEFAULT_SYSTEM_SCHEMAS } from './constants'
 import { filterByList } from './helpers'
 import { FUNCTIONS_SQL } from './sql/functions'
 
+export type PGFunction = {
+  id: number
+  schema: string
+  name: string
+  language: string
+  definition: string
+  complete_statement: string
+  args: Array<{
+    mode: 'in' | 'out' | 'inout' | 'variadic' | 'table'
+    name: string
+    type_id: number
+    has_default: boolean
+  }>
+  argument_types: string
+  identity_argument_types: string
+  return_type_id: number
+  return_type: string
+  return_type_relation_id: number | null
+  is_set_returning_function: boolean
+  behavior: 'IMMUTABLE' | 'STABLE' | 'VOLATILE'
+  security_definer: boolean
+  config_params: Record<string, string> | null
+}
+
 export const pgFunctionZod = z.object({
   id: z.number(),
   schema: z.string(),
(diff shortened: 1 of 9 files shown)


packages/pg-meta/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
-  "extends": "tsconfig/base.json",
+  "extends": "../tsconfig/base.json",
   "compilerOptions": {
-    "module": "ESNext"
+    "module": "ESNext",
+    "moduleResolution": "bundler"
```
avallete, line 5: **question**
Without this, I had types errors constantly showing up. Not sure if this is the best way to resolve this tough.

packages/pg-meta/test/db/utils.ts
```diff
@@ -25,7 +25,8 @@ export async function createTestDatabase() {
       dbName,
       executeQuery: async <T = any>(query: string): Promise<T> => {
         try {
-          const result = (await testDb.unsafe(query)).map((v) => v)
+          const execResult = await testDb.unsafe(query)
+          const result = execResult.map((v) => v)
```
avallete, line 29: Split it in two, making it easier to step in with the debugger if necessary

packages/pg-meta/test/db/00-init.sql
```diff
@@ -1,4 +1,5 @@
-
+-- We apply our test seedings to template1 so every new created db will have the same structure
+\c template1
```
avallete, line 2: This is a little tricks to avoid the need to "seed" each new "created" database at jest/typescript level. That way each newly created database for each test will have our tables initialized.

packages/pg-meta/src/pg-meta-functions.ts
```diff
@@ -212,8 +230,8 @@ function _generateCreateFunctionSql(
       config_params
         ? Object.entries(config_params)
             .map(
-              ([param, value]: string[]) =>
-                `SET ${param} ${value[0] === 'FROM CURRENT' ? 'FROM CURRENT' : 'TO ' + (value === '""' ? "''" : value)}`
+              ([param, value]) =>
+                `SET ${param} ${value === 'FROM CURRENT' ? 'FROM CURRENT' : 'TO ' + (value === '""' ? "''" : value)}`
```
avallete, line 234: **question**

Had a type error here, either the `config_params` type is wrong but doesn't seems so. I've added a test to cover this case but might want another look.

packages/pg-meta/test/functions.test.ts
```diff
@@ -0,0 +1,341 @@
...
+  expect(result1!.config_params).toEqual({
+    search_path: '""',
+    application_name: 'postgres.js',
+    work_mem: '8MB',
+  })
+
+  // Clean up
+  const { sql: removeSql1 } = await pgMeta.functions.remove(result1!)
+  await executeQuery(removeSql1)
+})
```
avallete, line 341: New test to cover the FROM CURRENT case.

packages/pg-meta/src/pg-meta-functions.ts
```diff
@@ -34,13 +58,11 @@ export const pgFunctionZod = z.object({
...
   security_definer: z.boolean(),
   config_params: z.union([z.record(z.string(), z.string()), z.null()]),
-})
+}) satisfies z.ZodType<PGFunction>
 
 export const pgFunctionArrayZod = z.array(pgFunctionZod)
 export const pgFunctionOptionalZod = z.optional(pgFunctionZod)
 
-export type PGFunction = z.infer<typeof pgFunctionZod>
-
```
soedirgo, line 43: Pourquoi ?

packages/pg-meta/tsconfig.json
soedirgo: Seems fine, let's go with it

packages/pg-meta/src/pg-meta-functions.ts
avallete: Good catch, that's a remaining artifact of me trying out some things to debug the types I had in my tests 😅 

Reverted this back to being infered by zod schema.

packages/pg-meta/src/sql/functions.ts
```diff
@@ -80,15 +80,16 @@ from
...
+        'has_default', COALESCE(has_default, false)
       )) as args
     from
       (
         select
           oid,
           unnest(arg_modes) as mode,
           unnest(arg_names) as name,
-          unnest(arg_types) as type_id,
+          unnest(arg_types)::int8 as type_id,
```
soedirgo, line 92: Let's keep this as `oid` instead of `int8`

packages/pg-meta/src/sql/functions.ts
```diff
@@ -80,15 +80,16 @@ from
         'mode', t2.mode,
         'name', name,
         'type_id', type_id,
-        'has_default', has_default
+        -- Cast null into false boolean
+        'has_default', COALESCE(has_default, false)
```
soedirgo, line 84: Is there ever a case where `has_default` is `null`?

soedirgo: Would've expected there to be none according to this https://github.com/supabase/supabase/blob/d78768cec464f420b765290208fd60ed98b1085d/packages/pg-meta/src/sql/functions.ts#L20

packages/pg-meta/src/pg-meta-functions.ts
soedirgo: Yeah I think the `value[0]` was wrong before

packages/pg-meta/src/sql/functions.ts
avallete: Can do but will have to convert the number in zod type to a string, because oid get out as a string with `postgres.`

packages/pg-meta/src/sql/functions.ts
avallete: Yup some of them get null values without that.

Screenshot 2025-02-11 at 19 32 06

avallete:
Updated things out to use `pg` instead of `postgres` since that's what's will be executed on the `postgres-meta` side.

That come from a discussion oustide of github on slack, here is a summary for context:

Refactor to use of the `pg` client for testing while preparing for potential future client changes. The main reasons are:  

1. **Consistency and Maintainability**: Using `pg` minimizes diffs and aligns with query execution in `postgres-meta` with the `executeQuery` endpoint
2. **Type Handling Improvements**: Discrepancies arose from custom type parsing (e.g., `INT8` as `number` or `string`), leading to inconsistent test results. To address this, `pg` is configured with type overrides similar to those in `postgres-meta`, ensuring consistent behavior.  
3. **Zod Integration**: Zod is kept to validate runtime values, catching edge cases like `null` defaults and `bigint > safeInteger`. This allows safe future client switches by flagging type differences during tests.  
4. **Testing Strategy**: Tests continue to use `toMatchInlineSnapshot` since we’re only using the `pg` client for now. However, the refactor prepares the codebase to easily switch clients later and catch any differences in runtime behavior.  

This approach keeps the current tests reliable while ensuring flexibility for future changes.

soedirgo (approved):

PR #33485, opened 2025-02-10
Billing docs revamp - Dashboard links, Redirects, high-level billing page

ecktoteckto:
This PR includes a few different things
- adds a page "About billing on Supabase"
  - supposed to be high-level. That's why there's no examples for billing or anything like that. Might be changed later (e.g. converted into a "Get started" page or somehting) but is fine for now I guess.
- adjusts links on the Dashboard to new billing docs pages
  - so far I've only adjusted existing links. Now that we have dedicated pages for different topics/usage items we could theoretically add more links to specific sections. But not part of this PR
- adds redirects from current/old billing docs pages to new pages
  - went for non-permanent links in case we change the pages of the new docs again. Unless I got something wrong regarding the redirect logic

### DIFF
--- a/apps/www/lib/redirects.js
+++ b/apps/www/lib/redirects.js
@@ -2851,6 +2851,86 @@ module.exports = [
   },
   {
     permanent: false,
+    source: '/docs/guides/platform/org-based-billing#how-billing-is-organized',
+    destination: '/docs/guides/platform/billing-on-supabase#organization-based-billing',
+  },
+  {
+    permanent: false,
+    source: '/docs/guides/platform/org-based-billing#quotas-and-features',
+    destination: '/docs/guides/platform/billing-on-supabase#usage-quota',
+  },
+  {
+    permanent: false,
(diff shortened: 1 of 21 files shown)


apps/www/lib/redirects.js
```diff
@@ -2844,6 +2844,96 @@ module.exports = [
...
+    destination: '/docs/guides/platform/manage-your-usage/compute#compute-credits',
+  },
+  {
+    permanent: false,
+    source: '/docs/guides/platform/org-based-billing#read-replicas',
+    destination: '/docs/guides/platform/manage-your-usage/read-replicas',
+  },
+  {
+    permanent: false,
+    source: '/docs/docs/guides/platform/org-based-billing#project-add-ons',
```
kevcodez, line 2889:
```suggestion
    source: '/docs/guides/platform/org-based-billing#project-add-ons',
```

kevcodez (approved):

PR #33512, opened 2025-02-11
docs: fix typographical error

clarencepenz:


charislam:
Thanks @clarencepenz! If I'm not mistaken, the name of that column in `pg_stat_activity` is in fact `usename`, right?

I just checked with:

```
select *
from information_schema.columns
where table_schema = 'pg_catalog'
and table_name = 'pg_stat_activity';
```

clarencepenz:
@charislam, I checked and noticed that as well. Correct me if I am wrong. Is `usename` a typo that needs to be corrected as `username`? As I see it appears in other files?

charislam:
Sorry, could you explain again what you mean? I'm not quite catching on 😅 

In general, I think username is the more commonly seen term, but there are cases where the underlying program does in fact call it `usename`, so we have to follow along 😄

clarencepenz:
Ah ha! I see. So in this case there won't be any need to use the standard naming convention `username` and leave it as `usename` correct?

charislam:
@clarencepenz yep! I'll close the PR now to clean up, but thank you for your contribution! We always welcome PRs and have a bunch of open issues if you'd like to keep contributing! ❤️

PR #33524, opened 2025-02-11
Add context menu to auth users

joshenlim:
Small QoL improvement to auth users grid - add context menu with actions for copying cell content and deleting user
- Reduces the number of clicks required to do these actions
- Copy cell content label is contextualized to the column name (e.g Copy email, Copy UID, etc)
image


saltcod (approved):
checked preview,  works great!

PR #33527, opened 2025-02-11
Chore/support downloading current view as png in schema visualizer

joshenlim:
Fixes FE-69

image


saltcod (approved):
looks great! checked on staging and 👍

PR #33540, opened 2025-02-12
docs: clarify requirement for combining restrictive policy with permi…

molti-tasking:
It took me many hours to find this issue. If there would have been a little more information in one particular place of the docs about anonymous users, it would have been easier. I added just one sentence about restrictive RLS policies and that they need another policy that returns true, ensuring the combined OR condition is satisfied.

YES

Just a little clarification for the docs as a hint for developers.

No issues, the docs are fine. But I feel like they are missing a little hint. 

## What is the new behavior?

I added the hint.

## Additional context

Not really much more context, but just extending the docs.


apps/docs/content/guides/auth/auth-anonymous.mdx
```diff
@@ -253,6 +253,7 @@ using ( true );
 <Admonition type="note" label="Use restrictive policies">
 
 RLS policies are permissive by default, which means that they are combined using an "OR" operator when multiple policies are applied. It is important to construct restrictive policies to ensure that the checks for an anonymous user are always enforced when combined with other policies.
+Be aware that a single 'restrictive' RLS policy alone will fail unless combined with another policy that returns true, ensuring the combined OR condition is met.
```
charislam, line 256:
```suggestion
Be aware that a single 'restrictive' RLS policy alone will fail unless combined with another policy that returns true, ensuring the combined condition is met.
```

charislam (approved):
Thanks @molti-tasking!

PR #33541, opened 2025-02-12
Inline editor feature preview

SaxonF:

image

This introduces a new inline editor behind a feature preview. The inline editor can be used to run SQL wherever you are in the dashboard without taking you away from what you doing. If the feature preview is enabled, creating and editing policies and functions is all done through the inline editor vs the current ui.

No inline editor exists. Editing policies and functions is done through a form based approach.

## What is the new behavior?

When creating or editing a policy, it will pass the policy definition in for editing along with templates. When creating it will also pass in a starting prompt so a user can lean on AI to generate a policy. The same behaviour exists for functions

The inline editor also connects with the chat based Assistant so you can pass queries back and forth. When you want to save a query for later editing or re-running, you can click save which saves it as a snippet to be accessed in the full SQL Editor.


kiwicopple (approved):

joshenlim:
## Bugs identified:
- Using inline assistant, hitting cmd enter to accept generated output doesn't accept the whole output. It does work properly if I click "Accept changes" instead of hitting cmd enter

https://github.com/user-attachments/assets/57b5bf52-a15e-49de-b221-6479c38b4c10

## Feedback for changes in the PR:
- Wondering if we should have a "toggle state" for the buttons? including assistant, if they're open (e.g lighter bg color)
image
- Probably should hook up keyboard shortcut (cmd + enter) to run the query? The enter icon here suggests that might be possible. Also should include the corresponding meta key too (mac -> cmd,  else ctrl)
image
- Thinking to support minimizing results so we get more real estate for writing queries after running one
image

## Just some ideas after testing around:
- Should we have another keyboard shortcut to toggle the inline editor pane? I think this might be a commonly used functionality, and having a shortcut will make the UX more seamless
- Wondering if having a CTA to open in SQL Editor might be helpful? (e.g "upgrading" to a better env for querying)

## Feedback not directly related to this PR
- Not related to this PR, but thinking might be helpful to highlight where the "write" operations are when the warning shows up, in this example i'm wondering where the "write" operations are in the query, but can't seem to find
image

apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/PolicyTableRow.utils.ts
```diff
@@ -12,7 +12,7 @@ export const generatePolicyCreateSQL = (policy: PostgresPolicy) => {
   }
 
   return `
-CREATE POLICY "${policy.name}" 
+CREATE OR ALTER POLICY "${policy.name}"
```
joshenlim, line 15: this is not valid SQL iinw - its either "create policy" or "alter policy", no "create or alter"

apps/studio/components/ui/AIEditor/index.tsx
```diff
@@ -0,0 +1,332 @@
...
+    includeSchemaMetadata?: boolean
+  }
+  initialPrompt?: string
+  readOnly?: boolean
+  className?: string
+  options?: monacoEditor.IStandaloneEditorConstructionOptions
+}
+
+const AIEditor = ({
+  id,
```
joshenlim, line 31: unused param

apps/studio/components/ui/AIEditor/index.tsx
joshenlim, line 1: just wondering if this is planned to be used elsewhere in the dashboard? otherwise can probably shift this and ResizableAIWidget into EditorPanel folder?

apps/studio/components/ui/EditorPanel/EditorPanel.tsx
```diff
@@ -0,0 +1,420 @@
...
+  const { profile } = useProfile()
+  const snapV2 = useSqlEditorV2StateSnapshot()
+  const { mutateAsync: generateSqlTitle } = useSqlTitleGenerateMutation()
+  const isOptedInToAI = useOrgOptedIntoAi()
+  const includeSchemaMetadata = isOptedInToAI || !IS_PLATFORM
+
+  const [isSaving, setIsSaving] = useState(false)
+  const [error, setError] = useState<QueryResponseError>()
+  const [results, setResults] = useState<undefined | any[]>(undefined)
+  const [showResults, setShowResults] = useState(false)
```
joshenlim, line 44: unused param -  showResults

apps/studio/components/ui/AIEditor/index.tsx
SaxonF: using it in edge functions and my hope is we can use it anywhere code editors are used

apps/studio/state/app-state.ts
```diff
@@ -47,6 +65,19 @@ const INITIAL_AI_ASSISTANT: AiAssistantPanelType = {
   tables: [],
 }
 
+const INITIAL_EDITOR_PANEL: EditorPanelType = {
+  open: false,
+  initialValue: '',
+  label: '',
+  saveLabel: '',
+  initialPrompt: '',
+  templates: SQL_TEMPLATES.filter((template) => template.type === 'template').map((template) => ({
```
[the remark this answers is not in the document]
SaxonF: made it so you can pass templates in externally e.g. policies, triggers and functions will all have different templates

SaxonF:
> When editing a database entity (policy, triggers, etc), im thinking if we should show the SQL statements as placeholders instead of filling up the SQL in the editor directly. We did that in the previous iteration of an inline editor and am thinking it might be better? The SQL statements for updating entities is sometimes different from the SQL statement for creating so will need some updating there

@joshenlim can you expand on this a little bit? When I used the previous policy editor it wasn't immediately clear I could "accept" the placeholder which is important especially when editing a policy.

joshenlim:
> @joshenlim can you expand on this a little bit? When I used the previous policy editor it wasn't immediately clear I could "accept" the placeholder which is important especially when editing a policy.

@SaxonF ahh good point actually RE no indication to "accept" the placeholder - hmm tbh thinking about this again maybe ignore this one for now.

separately, one point of friction i can foresee with the new UX (specifically for auth policies) - if say i just want to rename my policy, its not immediately clear how to do so i think (cause the SQL to rename a policy is different from updating the definition of a policy). Maybe we could include both SQL statements and add comments?
image

joshenlim (approved):
LGTM! Verified on staging preview only (Doesn't affect local/self-host since this is behind a feature preview)

Some tweaks probably needed to make the UX around managing policies / functions / triggers easier but can address in subsequent follow up PRs 🙂

PR #33542, opened 2025-02-12
Small refactor to use a separate confirmation modal for right click delete user

joshenlim:
Just so that we don't trigger the user's checkbox when we right click to delete a user (odd UX esp if we hit cancel and the checkbox stays checked)

I had this set up previously actually, but forgot to hook it up 😓


alaister (approved):
LGTM!

- Tested on preview

PR #33553, opened 2025-02-12
Chore/more refactoring of data fetchers methods

joshenlim:
React queries updated in this PR:

- bucket-object-download-mutation 
  - JFYI for this one we can't use openapi-fetch's method as they don't support receiving octet-stream responses, left comments for context
  - [ ] Platform + Local: Check downloading an object
- postgres-service-status-query
  - [ ] Platform: Check postgres service status in a project home page 
- password-strength
  - [ ] Platform: Test password strength bar when creating new project
- free-project-member-limit-check-query
  - [ ] Platform: Test free project member limit when creating a new project in a org which has a member hitting said limit
- exit-survey-send
  - [ ] Platform: Test downgrading project to Free and submitting an exit survey
- project-disk-resize-mutation
  - [ ] Platform: Test updating disk size in project's settings -> compute and disk
- project-update-mutation
  - [ ] Platform: Test updating project name
- project-postgrest-config-query
  - [ ] Platform: Test viewing Project's Data API Settings
- project-postgrest-config-update-mutation
  - [ ] Platform: Test updating Project's Data API Settings 
- organization-audit-logs-query
  - [ ] Platform: Test viewing and filtering on organization audit logs 
- organization-member-delete-invitation
  - [ ] Platform: Test removing an organization member

### DIFF
--- a/apps/studio/data/storage/bucket-object-download-mutation.ts
+++ b/apps/studio/data/storage/bucket-object-download-mutation.ts
@@ -3,7 +3,7 @@ import { toast } from 'sonner'
 
 import { components } from 'data/api'
 import { post } from 'lib/common/fetch'
-import { API_URL } from 'lib/constants'
+import { API_URL, IS_PLATFORM } from 'lib/constants'
 import { ResponseError } from 'types'
 
 type DownloadBucketObjectParams = {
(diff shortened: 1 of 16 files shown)


apps/studio/data/storage/bucket-object-download-mutation.ts
```diff
@@ -18,9 +18,10 @@ export const downloadBucketObject = async (
 ) => {
   if (!bucketId) throw new Error('bucketId is required')
 
-  // has to use lib/common/fetch post because the other post doesn't support wrapping blobs
+  // [Joshen] JFYI we have to use lib/common/fetch post as post from openapi-fetch doesn't support receiving octet-streams
+  // Opting to hard code /platform for non platform just for this particular mutation, so that it's clear what's happening
   const response = await post(
```
alaister, line 23: Just for future reference we can also just opt for regular `fetch()` instead of these old methods.

alaister (approved):
LGTM!

- Tested on preview

apps/studio/data/storage/bucket-object-download-mutation.ts
alaister:
```
const headers = await constructHeaders()
const response = await fetch(`${BASE_PATH}/path`, { headers, signal })
```

joshenlim: Ahh yeah good point actually, was wondering about that cause end goal i guess is to just deprecate the whole lib/common/fetch folder

Will look into that separately when i clean up the old methods! 🙏

PR #33560, opened 2025-02-12
fix:  clarify peak connections

pamelachia:

YES

- clarifying that realtime peak connections means concurrent peak connections. To follow other parts of billing where we say concurrent peak connections.
- Also there is a concern that [here](https://supabase.com/docs/guides/realtime/quotas#quotas-by-plan) we describe it as concurrent clients instead of concurrent connections but we use them interchangeably in this page. 
- resolves GROWTH-282


kevcodez (approved):

staaldraad (approved):

apps/studio/components/interfaces/Organization/BillingSettings/BillingBreakdown/BillingBreakdown.constants.ts
```diff
@@ -74,7 +74,7 @@ export const BILLING_BREAKDOWN_METRICS: Metric[] = [
   },
   {
     key: PricingMetric.REALTIME_PEAK_CONNECTIONS,
-    name: 'Realtime Peak Connections',
+    name: 'Realtime Peak Concurrent Connections',
```
charislam, line 77: this one says "peak concurrent" and the one below says "concurrent peak", let's standardize on one

charislam (approved):

PR #33571, opened 2025-02-12
Added hsts header via vercel.json

doublethink:
Changes to Cloudflare have necessitated migrating some header configurations to the Vercel layer.
The headers are necessary on all apps, only the subdomain and preload directive are needed on the www, but this will come after this change has been deployed.

Part of SEC-120


alaister (changes requested):
hey @doublethink,

Would you mind moving these to the `next.config.js` files as we already configure headers there, that way we don't have it in two places :)

doublethink:
> hey @doublethink,
> 
> Would you mind moving these to the `next.config.js` files as we already configure headers there, that way we don't have it in two places :)

Sure thing, was intended to avoid adding the header to localhost for development but I see we have a pattern to prevent that

alaister (approved):
LGTM!

I added `process.env.VERCEL === '1'` to your check since without it we would have HSTS on in local development (NEXT_PUBLIC_IS_PLATFORM === 'true' on local dev). Now it only runs on our hosted studio running on Vercel.

I also wanted to check if you want these headers on the root paths (`/`)? The `/:path*` check actually only applies to subpaths.

dshukertjr (approved):

staaldraad (approved):

PR #33573, opened 2025-02-12
add realtime toggle tracking

SaxonF:
Adds tracking for when a user toggles realtime on or off for a particular table. The goal is track _intention_ to enable and identify drop off between intention to use and successfully activating.


joshenlim (approved):
LGTM from FE end - verified on preview that the events are sent correctly when toggling on/off realtime from Table Editor

pamelachia (approved):
looks good to me too

PR #33578, opened 2025-02-13
fix(studio): correct bold font weight + csp

alaister:


joshenlim (approved):
minimally tested logging in

kevcodez (approved):

staaldraad (approved):

PR #33592, opened 2025-02-13
Add E2E tests for local studio logs

jordienr:

YES

Adds local studio E2E tests for logs that will
- navigate to logs pages
- expect an error NOT to show
- click in a log
- expect the panel to open up and show the same timestamp as the log clicked

- We dont have any e2e tests for logs in local studio and if it breaks it can be painful to update and wait for the release

## What is the new behavior?

- hopefully the tests will prevent anything from breaking in local studio


joshenlim (approved):
LGTM!

PR #33612, opened 2025-02-14
Feat/sidebar studio

MildTomato:

YES

### What kind of change does this PR introduce?

- Sidebar refactored to use shadcn
- Now features 'expanded', 'closed', and 'expand on hover' options.
- UserDropdown moved it its own file. 
  - We will probably use this later in the LayoutHeader.
- Old states removed from app-state
- User preferences updated to control sidebar behaviour.
- cmdk icon in sidebar changed to immediately open cmdk.

### Misc
- Old components for <NavigationBar /> can be removed in a follow up

https://github.com/user-attachments/assets/e0548883-c785-4a11-aa34-dd2f9b8dba30


apps/studio/components/interfaces/UserDropdown.tsx
```diff
@@ -0,0 +1,146 @@
+import { ProfileImage } from 'components/ui/ProfileImage'
```
MildTomato, line 1: old file can be removed

apps/studio/components/interfaces/UserDropdown.tsx
```diff
@@ -0,0 +1,146 @@
...
+              <ProfileImage
+                alt={profile?.username}
+                src={profile?.profileImageUrl}
+                className="w-7 h-7 rounded-md"
+              />
+            </div>
+            <span className="flex flex-col gap-0">
+              {profile?.username}
+              <span className="text-foreground-lighter text-xs">{profile?.primary_email}</span>
+            </span>
```
joshenlim, line 61: thinking we should use a different UI here for local set up?
image

joshenlim (approved):
LGTM! Verified on both hosted + local studio

apps/studio/components/interfaces/UserDropdown.tsx
MildTomato: oh, we could just hide it I suppose

MildTomato: should fix it?
https://github.com/supabase/supabase/commit/843f7437f263115fd57a80defe4feb3b2186316a

MildTomato: Fixed now, we have an alternative dropdown for local and self hosted.
thanks @joshenlim for flaggin!

saltcod (approved):
Checked hosted and local, looks great!

PR #33624, opened 2025-02-14
Fix height of storage panel

saltcod:
Just wraps the <autosizer> component in the flex-grow div. 
Problem seemed to be introduced with the layouts PR last week

The top storage item is always hidden behind the header:
https://share.cleanshot.com/6NDTxKXb

Note:  I checked the other instances of this InfiniteList component and all seem good to me.


joshenlim (approved):
Verified on staging preview where InfiniteList is being used:
- StorageExplorer List view
- TableEditorMenu
- NotificationsPopover
- SQL Editor search list

PR #33643, opened 2025-02-17
chore(pg-meta): migrate column-privileges

avallete:

Migrate code and tests for: https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaColumnPrivileges.ts


packages/pg-meta/src/pg-meta-column-privileges.ts
```diff
@@ -0,0 +1,160 @@
...
+  includedSchemas,
+  excludedSchemas,
+  columnIds,
+  limit,
+  offset,
+}: {
+  includeSystemSchemas?: boolean
+  includedSchemas?: string[]
+  excludedSchemas?: string[]
+  columnIds?: string[]
```
avallete, line 51: **remark**

I introduced this new params for the columns privilege listing to easily reproduce the behavior of getting back the results after modifications of a column originally here:

https://github.com/supabase/postgres-meta/blob/8418eaab8d5f6d5b3d32cbc63c9d671376845c08/src/lib/PostgresMetaColumnPrivileges.ts#L88-L95

I did wonder if it would be better within a `retrieve` method. But since this can be an array of columnIds, `retrieve` seemed like a poor fit. `list` with a filter seems more appropriate. But happy to change and open to discussion.

packages/pg-meta/test/column-privileges.test.ts
```diff
@@ -0,0 +1,356 @@
...
+      "relation_schema": "public",
+    }
+  `
+  )
+})
+
+withTestDatabase('revoke & grant column privileges', async ({ executeQuery }) => {
+  const testRole = `test_role_${Date.now()}`
+  // Create test role
+  await executeQuery(`create role ${testRole};`)
```
avallete, line 79: **remark**

Instead of using a static `r` role, use a dynamic changing role to reduce the chances of conflicts when we'll parallelize testing.

packages/pg-meta/src/pg-meta-column-privileges.ts
soedirgo: Sgtm 👍

soedirgo (approved):

PR #33645, opened 2025-02-17
chore(pg-meta): migrate indexes

avallete:

Migrate code from: https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaIndexes.ts


packages/pg-meta/src/pg-meta-indexes.ts
```diff
@@ -0,0 +1,101 @@
...
+  is_clustered: z.boolean(),
+  is_valid: z.boolean(),
+  check_xmin: z.boolean(),
+  is_ready: z.boolean(),
+  is_live: z.boolean(),
+  is_replica_identity: z.boolean(),
+  key_attributes: z.string(),
+  collation: z.string(),
+  class: z.string(),
+  options: z.string(),
```
avallete, line 26: **issue**

During migration I noticed that those fields type definitions were wrong in our `postgres-meta` repository.

Those are defined as being array of numbers here: https://github.com/supabase/postgres-meta/blob/ca06061b4708971628134f95e49f254c2dfdfa7d/src/lib/types.ts#L199-L202

But at execution time those are converted in plain strings:
https://github.com/supabase/postgres-meta/blob/ca06061b4708971628134f95e49f254c2dfdfa7d/test/server/indexes.ts#L40

Their actual types are postgres `int2vector` and `oidvector` when the vector have more than one attribute, the result will be a concatenated string like so:

```
    key_attributes: '2 3',
    collation: '0 950',
    class: '1981 10028',
    options: '0 0',
 ```

In order to limit the breaking changes that would be introduced by the migration to the new repo, I've just changed the type to match the actual runtime result.

The alternative would be to add a `parseArray` for such types at the pg override level here:
https://github.com/supabase/postgres-meta/blob/ca06061b4708971628134f95e49f254c2dfdfa7d/src/lib/db.ts#L18

Since I'm unsure of the potential side effect of this (it could affect user running custom queries via `postgres-meta` ?) I could use a second opinion on this.


**Note:** If we choose to keep those as plain string, I'll make a PR on `postgres-meta` to reflect the actual type returned at runtime.

avallete: After discussion with @soedirgo decided to fix the query to return an array of integer: https://github.com/supabase/supabase/pull/33645/commits/34b014c61828caf0bc9ecc148c750f773b760852

Will port the fix to postgres-meta as it only require the proper casting at query level.

soedirgo (approved):

PR #33647, opened 2025-02-17
Chore/refactor e2e tests

joshenlim:
This PR sets up running E2E tests on our staging dashboard instead of only local - I didn't want to overwrite any of our existing set up for local test (which are running on a per PR basis) so have created a new folder `studio-tests`
- Running tests on local: `pnpm run test:local`
- Running tests on staging: `pnpm run test:staging`

Changes broadly include:
- Spec test files are split into “common” and “production” functionality so that we can reuse test files across environments
  - e.g common: table editor, logs explorer (what we currently have)
  - e.g production: org settings, project settings, etc
- The auth step to log into staging has also been set up although currently requires a manual intervention to solve the hcaptcha for now (Will need to address this eventually for us to automate E2E tests on staging)
- Separately, I also made some minor changes to the logs spec tests (didn’t update the one in local-studio-tests) to address some flakiness + redundancy
  - Navigate directly to the logs pages since the first test already covers navigating to logs pages from home page
  - Add some waits for requests in favor of timeouts

The potential intended strategy from here is that we:
- Still run tests on each PR (on local env dashboard)
- Turn off auto promotion of vercel deploys on our prod app
- Run E2E tests on staging after the PR has been merged
- Deploy app to prod after E2E tests passes

Just need to make sure that all these + the set up here makes sense 🙏


.github/workflows/local-studio-tests.yml
jordienr, line 1: Should we rename the file to studio-tests too?

jordienr (approved):
Code LGTM and everythings working locally 🔥

.github/workflows/local-studio-tests.yml
joshenlim: oh yup! can do :)

PR #33660, opened 2025-02-17
feat: Add Login event to account audit logs

kamilogorek:
Fixes https://linear.app/supabase/issue/API-23/add-login-event-to-account-audit-logs
Depends on https://github.com/supabase/infrastructure/pull/21599

Also added missing `TelemetryActions.SIGN_IN` to "Login with GitHub" flow cc @loong 

image


loong (approved):

apps/studio/components/interfaces/SignIn/SignInSSOForm.tsx
```diff
@@ -53,7 +49,6 @@ const SignInSSOForm = () => {
     })
 
     if (!error) {
-      sendEvent({ action: TelemetryActions.SIGN_IN, properties: { category: 'account' } })
```
joshenlim, line 56: am removing as successful call here will redirect to /sign-in-mfa, which already sends a sign in telemetry event
https://github.com/supabase/supabase/blob/login-audit-entry/apps/studio/pages/sign-in-mfa.tsx#L49

joshenlim (approved):
LGTM - verified on staging preview that the following scenarios trigger the audit login endpoint
- Sign in with email/password
- Sign in with GH
- Sign in via SSO

PR #33672, opened 2025-02-18
fix: indexes in use

alaister:

Bug fix

- fixes indexes in use query – previously it would only work with queries with zero arguments or a single text argument.
- fixes index name extraction
- fixes index name argument passing to second query
- adds support for `with pgrst_payload` type queries.
- adds an error state for retrieve indexes in use.


joshenlim (approved):
Verified in query performance report that indexes are showing up correctly by running a select query on a table, and checking that the PK index shows up in query performance

Verified on
- Staging preview
- Self-host preview

PR #33679, opened 2025-02-18
chore(pg-meta): migrate tests for roles

avallete:

Migrate roles and tests: https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaRoles.ts


packages/pg-meta/src/pg-meta-roles.ts
```diff
@@ -65,24 +67,25 @@ where
...
+type RoleIdentifier = Pick<PGRole, 'id'> | Pick<PGRole, 'name'>
+
+function getIdentifierWhereClause(identifier: RoleIdentifier) {
+  if ('id' in identifier && identifier.id) {
+    return `${ident('id')} = ${literal(identifier.id)}`
+  } else if ('name' in identifier && identifier.name) {
+    return `${ident('name')} = ${literal(identifier.name)}`
+  }
+  throw new Error('Must provide either id or name')
+}
```
avallete, line 79: **remark**

Will be using this pattern to centralize the place where we create the condition to retrieve an unique element across all the `pg-meta-*` functions. So it can be shared between update/remove/retrieve without duplication.

packages/pg-meta/src/pg-meta-roles.ts
```diff
@@ -199,18 +192,18 @@ begin
     ${canLogin === undefined ? '' : canLogin ? 'login' : 'nologin'}
     ${isReplicationRole === undefined ? '' : isReplicationRole ? 'replication' : 'noreplication'}
     ${canBypassRls === undefined ? '' : canBypassRls ? 'bypassrls' : 'nobypassrls'}
-    ${connectionLimit === undefined ? '' : `connection limit ${literal(connectionLimit)}`}
+    ${connectionLimit === undefined ? '' : `connection limit ${connectionLimit}`}
     ${password === undefined ? '' : `password ${literal(password)}`}
-    ${validUntil === undefined ? '' : `valid until ${literal(validUntil)}`}
-  ', old.rolname));
+    ${validUntil === undefined ? '' : `valid until %L`}
+  ', old.name${validUntil === undefined ? '' : `, ${literal(validUntil)}`}));
```
avallete, line 198: **remark**

Here had to escape the date value that way to avoid escaping clash between JS and sql within the format string.

soedirgo (approved):

soedirgo (approved):

PR #33682, opened 2025-02-18
feat: Add supavisor connections to infra report

kamilogorek:
Depends on https://github.com/supabase/infrastructure/pull/21615
Closes https://linear.app/supabase/issue/API-454/infra-monitoring-endpoint-to-return-number-of-supavisor-connections


joshenlim (approved):
Changes here look good - minimally verified that the infra-monitoring call for supavisor_connections_active is working fine

I don't think the BE changes are deployed yet so will mark as draft for now (will merge tomorrow)

PR #33683, opened 2025-02-18
fix: disk config layout gap

fsansalvadore:
## Before
- Missing gap
Screenshot 2025-02-18 at 15 56 48

- Content overflowing for insufficient space
Screenshot 2025-02-18 at 16 00 38

## After

Screenshot 2025-02-18 at 15 57 09

Screenshot 2025-02-18 at 16 01 25


saltcod (approved):
checked on staging, looks good!

PR #33344, opened 2025-02-05
chore(www): static assets cdn

alaister:
Follow up to https://github.com/supabase/supabase/pull/33304

PR #33693, opened 2025-02-19
fix: run `next build` directly instead of `build` task to avoid uploading static assets when building image

thebengeu:
Was broken by #33344, which changed `build` task from `next build` to
`next build && ./../../scripts/upload-static-assets.sh`.


kevcodez (approved):
Not tested locally but looks sane

PR #33695, opened 2025-02-19
Update org roles confirmation to fire requests sequentially

joshenlim:
Changes do not affect local/self-host set up

Updates assigning, removing and updating organization roles to be fully sequential instead of parallel


alaister (commented):
I think there's a small issue with the invalidation... perhaps we should invalidate just once after performing all of these?

https://github.com/user-attachments/assets/bb25a8e8-403a-47d9-bc41-ee218a0b6941

joshenlim:
ahhh good catch! i'll add some skipInvalidation flags

alaister (approved):
LGTM!

- Tested on preview

PR #33710, opened 2025-02-19
Decouple updating disk size from retrieving disk util

joshenlim:
Mainly to support updating disk size, despite disk util failing

Just need to verify on hosted:
- Create a new project
- Jump straight to compute and disk
- Try to update the disk size
- Verify that only the disk size is being updated from the network requests

### DIFF
--- a/apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
+++ b/apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
@@ -292,10 +292,10 @@ export function DiskManagementForm() {
 
   useEffect(() => {
     // Initialize field values properly when data has been loaded, preserving any user changes
-    if (isSuccess) {
+    if (isDiskAttributesSuccess || isSuccess) {
       form.reset(defaultValues, {})
     }
-  }, [isSuccess])
+  }, [isSuccess, isDiskAttributesSuccess])
 
   // Redirect logic incase disk and compute feature is not live yet
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
```diff
@@ -292,10 +292,10 @@ export function DiskManagementForm() {
 
   useEffect(() => {
     // Initialize field values properly when data has been loaded, preserving any user changes
-    if (isSuccess) {
+    if (isDiskAttributesSuccess || isSuccess) {
```
joshenlim, line 295: so that the form can load initially first when the disk size is ready

apps/studio/components/interfaces/DiskManagement/fields/DiskSizeField.tsx
```diff
@@ -39,28 +37,26 @@ export function DiskSizeField({
   const org = useSelectedOrganization()
 
   const {
+    data,
```
delgado3d, line 40: I dont understand enough of this - just checking - is this `data` actually being used anywhere?

jordienr (approved):
tested in preview

saltcod (approved):
Checked preview, works as expected on a brand new project.

apps/studio/components/interfaces/DiskManagement/fields/DiskSizeField.tsx
joshenlim: oh yeah good catch - will remove!

PR #33723, opened 2025-02-20
Improve edge functions layout

SaxonF:
image

This updates brings the edge functions secrets page into the edge functions area of the dashboard. It adds a sidebar to navigate between functions list and secrets list. It also makes use of new PageLayout components for each page.

## How to test
1. Go into a project and go to [ref]/functions
2. See the updated functions page and the new left sidebar with secrets page
3. Click add a new function to make sure cli dialog works
4. Clicks secrets and add a new secret


joshenlim:
we can probably route this to the new secrets page with a link icon like what we do for vault
image
image

staaldraad (approved):

joshenlim (approved):
LGTM! Verified changes on staging

PR #33724, opened 2025-02-20
chore(pg-meta): migrate columns

avallete:
Migrate columns logic and tests from: https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaColumns.ts

Add coverage to the tests


soedirgo:
Maybe we can reuse the column zod type in tables/views/etc.?

avallete:
Good catch ! Yes indeed !

avallete:
Rebased the other PR's to re-use the columns zod definition

soedirgo (approved):

PR #33726, opened 2025-02-20
Create edge function

SaxonF:
image

This brings the ability to create multi-file edge functions via the dashboard. When creating an edge function you have an option to do it via an editor which takes you to the /new page. From here you can modify files using the AIEditor and an edge function focused complete endpoint. It includes the ability to view and select a template as a starting point. 

## Steps to test
1. Visit [projectRef]/functions
2. Click deploy a new function
3. Ensure only cli option is visible
4. Click via Cli and ensure dialog appears
5. Enable the `edgeFunctionCreate` feature flag
6. Click deploy new function and check two options now exist
7. Click via Editor and get redirected to /new
8. Give it a name and add some files
9. Click templates , hover templates and select a template
10. Deploy and get redirected to function


apps/studio/components/ui/FileExplorerAndEditor/FileExplorerAndEditor.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    isExpanded={isExpanded}
+                    isBranch={isBranch}
+                    isSelected={files.find((f) => f.id === originalId)?.selected}
+                    level={level}
+                    xPadding={16}
+                    name={element.name}
+                    icon={<File size={14} className="text-foreground-light" />}
+                    isEditing={Boolean(element.metadata?.isEditing)}
+                    onEditSubmit={(value) => {
+                      if (originalId !== null) handleFileNameChange(originalId, value)
```
joshenlim, line 156: Could we use a context menu to select and rename? Similar to SQL Editor for consistency (Clicking to rename was not intuitive)
there's also no way to delete the file atm i think

apps/studio/components/ui/FileExplorerAndEditor/FileExplorerAndEditor.tsx
```diff
@@ -0,0 +1,191 @@
...
+  return (
+    <div className="flex-1 overflow-hidden flex h-full">
+      <div className="w-64 border-r bg-surface-200 flex flex-col">
+        <div className="py-4 px-6 border-b flex items-center justify-between">
+          <h3 className="text-sm font-medium">Files</h3>
+          <Button size="tiny" type="default" icon={<Plus size={14} />} onClick={addNewFile}>
+            Add File
+          </Button>
+        </div>
+        <div className="flex-1 overflow-y-auto p-2">
```
joshenlim, line 127: i reckon theres no need to add padding here? let it take the full width and height 
image

apps/studio/pages/project/[ref]/functions/new.tsx
```diff
@@ -0,0 +1,286 @@
...
+            isDeploying ? (
+              <Loader2 className="animate-spin" size={10} strokeWidth={1.5} />
+            ) : (
+              <div className="flex items-center space-x-1">
+                <CornerDownLeft size={10} strokeWidth={1.5} />
+              </div>
+            )
+          }
+        >
+          Deploy function
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/functions/new.tsx:271]: shall we hook up a keyboard shortcut? the enter icon makes it seem like there's a keyboard shortcut to it
let's make it consistent across too like SQL Editor, Assistant, etc -> add the Cmd icon

image

apps/studio/pages/project/[ref]/functions/new.tsx
```diff
@@ -0,0 +1,286 @@
...
+          label: 'Edge Functions',
+          href: `/project/${ref}/functions`,
+        },
+        {
+          element: (
+            <div className="flex items-center gap-2">
+              <Label_Shadcn_ htmlFor="function-name" className="text-foreground-light sr-only">
+                Function name
+              </Label_Shadcn_>
+              <Input_Shadcn_
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/functions/new.tsx:168]: feedback: this wasn't very clear imo to name the function from here
I was confused for a bit when the deploy function button was disabled, but I couldn't figure out why
image

apps/studio/pages/project/[ref]/functions/new.tsx
```diff
@@ -0,0 +1,286 @@
...
+        router.push(`/project/${ref}/functions/${functionName}/details`)
+      }
+    },
+  })
+
+  const onDeploy = async () => {
+    if (isDeploying || !ref || !functionName) return
+
+    try {
+      await deployFunction({
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/functions/new.tsx:90]: consistency: use `mutate` instead of `mutateAsync`, then add the onError handlers to the useEdgeFunctionDeployMutation. We typically only use `mutateAsync` when we're firing multiple mutations that need to wait on each other

the mutation also has its own error handler by default so the toast in the catch error is rendering an extra error toast
image

Comment (joshenlim) [reply apps/studio/pages/project/[ref]/functions/new.tsx:90]: ^ on that note we probably need to add some client side validations for function names?
I ran into that error while setting the name as "Test Function" -> am guessing no spaces allowed

apps/studio/components/interfaces/Functions/Functions.templates.ts
```diff
@@ -0,0 +1,98 @@
...
+
+Deno.serve(async (req) => {
+  try {
+    const supabase = createClient(
+      Deno.env.get('SUPABASE_URL') ?? '',
+      Deno.env.get('SUPABASE_ANON_KEY') ?? '',
+      { global: { headers: { Authorization: req.headers.get('Authorization')! } } }
+    )
+
+    const { data, error } = await supabase.from('countries').select('*')
```
joshenlim, line 42: we've replaced the countries table recently with a colors table so maybe lets use that instead?

apps/studio/components/interfaces/Functions/Functions.templates.ts
joshenlim, line 1: tested all the examples here - looking great!

apps/studio/components/ui/FileExplorerAndEditor/FileExplorerAndEditor.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    }}
+                  />
+                </div>
+              )
+            }}
+          />
+        </div>
+      </div>
+      <div className="flex-1 min-h-0 relative px-3 bg-surface-200">
+        <AIEditor
```
joshenlim, line 169: hmm is the inline assistant hooked up to work with edge functions atm?

image

apps/studio/pages/api/ai/edge-function/complete.ts
```diff
@@ -0,0 +1,288 @@
...
+import { getTools } from '../sql/tools'
+import { executeSql } from 'data/sql/execute-sql-query'
+import pgMeta from '@supabase/pg-meta'
+
+export const maxDuration = 30
+const openAiKey = process.env.OPENAI_API_KEY
+const pgMetaSchemasList = pgMeta.schemas.list()
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  if (!openAiKey) {
```
joshenlim, line 13: could we write this in a consistent way as per `generate-v3` for example?
- What status to return when OpenAI key is not set
- Write as `switch(method)` etc
- Have a `handlePost` function

apps/studio/pages/api/ai/edge-function/complete.ts
```diff
@@ -0,0 +1,288 @@
...
+    return new Response(
+      JSON.stringify({
+        error: 'No OPENAI_API_KEY set. Create this environment variable to use AI features.',
+      }),
+      {
+        status: 500,
```
joshenlim, line 19: we'll need to be consistent on what status to return for this - generate-v3 returns 400, here and generate-v2 returns 500

we probably should set all of them to 400?

apps/studio/pages/project/[ref]/functions/new.tsx
```diff
@@ -0,0 +1,286 @@
...
+Deno.serve(async (req: Request) => {
+  const data = {
+    message: 'Hello from Supabase Edge Functions!',
+  };
+
+  return new Response(
+    JSON.stringify(data),
+    { headers: { 'Content-Type': 'application/json', 'Connection': 'keep-alive' }}
+  );
+});`,
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/functions/new.tsx:69]: this seems duplicate - can we retrieve this from Functions.templates.ts?

apps/studio/pages/api/ai/edge-function/complete.ts
```diff
@@ -0,0 +1,288 @@
...
+        ### Example Function using Node built-in API
+
+        \`\`\`edge
+        // Setup type definitions for built-in Supabase Runtime APIs
+        import "jsr:@supabase/functions-js/edge-runtime.d.ts";
+        import { randomBytes } from "node:crypto";
+        import { createServer } from "node:http";
+        import process from "node:process";
+
+        const generateRandomString = (length) => {
```
joshenlim, line 128: all these too - can they be retrieved from Functions.templates.ts?

apps/studio/pages/project/[ref]/functions/new.tsx
```diff
@@ -0,0 +1,286 @@
...
+    },
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/functions/new.tsx:70]: just wondering should we just use the hello world example from Functions.templates.ts here?

apps/studio/pages/project/[ref]/functions/new.tsx
```diff
@@ -0,0 +1,286 @@
...
+      await deployFunction({
+        projectRef: ref,
+        metadata: {
+          entrypoint_path: 'index.ts',
+          name: functionName,
+          verify_jwt: true,
+        },
+        files: files.map(({ name, content }) => ({ name, content })),
+      })
+    } catch (error) {
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/functions/new.tsx]: we probably need to add some client side validations for function names?
image
I ran into that error while setting the name as "Test Function" -> am guessing no spaces allowed

apps/studio/components/interfaces/Functions/Functions.templates.ts
SaxonF: @laktek mind updating this list of templates?

apps/studio/components/ui/FileExplorerAndEditor/FileExplorerAndEditor.tsx
SaxonF: actually, is that a legit function if you're just logging? If you type "write me a demo edge function that just returns hello world" it should write a function

apps/studio/pages/api/ai/edge-function/complete.ts
SaxonF: Reluctant to change this as what is there is based off prompt example we give people. We might have a lot of templates that may add context duplication.

apps/studio/components/ui/FileExplorerAndEditor/FileExplorerAndEditor.tsx
joshenlim: ahhh okay yeah that prompt works - might still need some tweaking but we'll learn more when users start using
image

apps/studio/components/interfaces/Functions/Functions.templates.ts
joshenlim: will replace this with something more generic like "table_name"

joshenlim (approved):
LGTM! Verified changes on staging + ensured that changes are not visible if feature flag is off

PR #33727, opened 2025-02-20
Edit edge functions

SaxonF:
image

This introduces a "code" tab the single edge function page which makes it possible to view and edit an edge functions source code then deploy the changes.

## How it works
1. New studio api `body` endpoint added
2. Endpoint first retrieves eszip bundle
3. Uses new parseEszip function to parse bundle into readable files. This uses the @deno/eszip package which has to be run in node env (not browser)
4. Endpoint returns files
5. Front-end uses existing file editor component (used in creating functions) 

## How to test
- Go to edge functions in dashboard
- Click on an edge function
- Notice no code tab
- Enable the feature flag
- Notice code tab
- Click code tab and confirm files are loaded properly
- Make a change to a file and deploy
- Use the "test" panel to confirm changes are live
- Make a change to a file name and confirm entry point is updated with new file name (ie function still works)


Krobys:
Hi, that sounds really cool! I can't wait for this change, especially if it works on the self-hosted version! 

Just checking in— is this PR still being worked on?

apps/studio/lib/eszip-parser.ts
```diff
@@ -0,0 +1,154 @@
...
+    try {
+      // Try to get the module source
+      const moduleSource = await parser.getModuleSource(specifier)
+
+      // Get the file path
+      const filePath = url2path(specifier)
+
+      // Create a file object
+      const file = new File([moduleSource], filePath, {
+        type: 'text/typescript',
```
joshenlim, line 144: does it matter if this is hard coded?
e.g what about `deno.json`?

laktek: AFAIK, this doesn't get sent to the server. I assumed it's needed on client side for syntax highlighting. If not we can probably remove it?

joshenlim (approved):
LGTM! Verified on staging, for edge functions that were deployed from the API (or dashboard)
- Can update code changes in edge function + verified with tester
- Can rename edge function, then update edge function code
- Can add files to edge function code + save and persist
- Can remove files from edge function code + save and persist

JFYI there's a known issue atm with updating edge functions that were deployed via the CLI, but we're investigating that separately

PR #33728, opened 2025-02-20
Test edge functions

SaxonF:
image

This introduces an edge function tester which is basically a simplified version of Postman that can be used to send requests to an edge function for testing. It adds a "test" button in an edge function page header which triggers a sheet that contains request configuration and response.

## Steps to test
1. Go to an edge function that is testable (will return a response)
2. Notice in top right there is no test button
3. Enable the `edgeFunctionCreate`flag
4. Notice there is now a test button
5. Click button and make sure it shows tester sheet
6. Add request details
7. send request and make sure response makes sense


joshenlim:
found some issues that i can help fix
- New function page clicking "Edge Function" leads to a 404, and is causing a full page refresh (Due to not using the Link component from next/link)
- Clicking on the example doesnt render the code in the editor (I don't think Local Storage is the right solution here tbh)

joshenlim (approved):
LGTM! Tested on staging preview :)

PR #33730, opened 2025-02-20
chore: remove event enums

loong:

YES

Removing enums inspired by @kamilogorek 

## Spot checks

Checked manually a few events locally just making sure they all send as usual

* `organization_created`
* `project_creation_simple_version_submitted `
* `project_created `
* `assistant_prompt_submitted`
* `project_removed`
* `realtime_inspector_filters_applied` (on an existing project instead of the newly created one)
* `organization_removed` (on a different org than the newly created one)


pamelachia (approved):
tested in preview that events get to posthog fine. not sure how else to further test. Also saw that if I rename the actions it would not be an acceptable type if they're not in telemetry-constants

staaldraad (approved):

saltcod (approved):
Tested a bunch of events on staging and lgtm!

PR #33733, opened 2025-02-20
Disable selection of io2 volume type if project region is not supported

joshenlim:
Fixes FE-1413

As per PR title - although we should eventually shift the newly added constants into [shared-types](https://github.com/supabase/shared-types) perhaps?

> [!NOTE]
> Does not affect local/self-host set up

image

### DIFF
--- a/apps/studio/components/interfaces/DiskManagement/fields/StorageTypeField.tsx
+++ b/apps/studio/components/interfaces/DiskManagement/fields/StorageTypeField.tsx
@@ -1,7 +1,9 @@
 import { UseFormReturn } from 'react-hook-form'
 
 import { useParams } from 'common'
+import { InlineLink } from 'components/ui/InlineLink'
 import { useDiskAttributesQuery } from 'data/config/disk-attributes-query'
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/DiskManagement/fields/StorageTypeField.tsx
```diff
@@ -26,8 +31,11 @@ type StorageTypeFieldProps = {
 }
 
 export function StorageTypeField({ form, disableInput }: StorageTypeFieldProps) {
-  const { ref: projectRef } = useParams()
   const { control, trigger } = form
+  const project = useSelectedProject()
+  const { ref: projectRef } = useParams()
+
+  const isIo2Supported = false // IO2_AVAILABLE_REGIONS.includes(project?.region ?? '')
```
alaister, line 38:
```suggestion
  const isIo2Supported = IO2_AVAILABLE_REGIONS.includes(project?.region ?? '')
```

😉

alaister (approved):
LGTM!

- Tested on preview

PR #33746, opened 2025-02-20
Resend customer story

fsansalvadore:


pamelachia (approved):
Read the post and it looks good. lgtm

PR #33751, opened 2025-02-20
feat: clarify impact of `--no-restart` flag when changing custom PG config

darora:


tomashley (approved):

apps/docs/content/guides/database/custom-postgres-config.mdx
```diff
@@ -159,7 +159,16 @@ supabase --experimental \
 postgres-config delete --config shared_buffers,work_mem
 ```
 
-By default, this might restart the database after deleting the configurations. You can use the `--no-restart` flag to prevent the restart:
+By default, changing the configuration, whether by updating or deleting, causes the database and all associated read replicas to restart. You can use the `--no-restart` flag to prevent this behavior, and attempt to reload the updated configuration without a restart. Please refer to the Postgres documentation to determine if a given parameter can be reloaded without a restart.
+
+<Admonition type="note" label="Read Replicas and Custom Config">
+
+Postgres requires several parameters to be synchronized between the Primary cluster and [Read Replicas](./docs/guides/platform/read-replicas).
```
charislam, line 166:
```suggestion
Postgres requires several parameters to be synchronized between the Primary cluster and [Read Replicas](/docs/guides/platform/read-replicas).
```

charislam (approved):

PR #33753, opened 2025-02-21
chore: only include platform/v1 routes in api types

kevcodez:
Reduces api types to just platform/v1 routes and strips v0 and system routes - based on https://github.com/supabase/infrastructure/pull/21676

To keep it clean, we generate two separate type files for v1 and platform. Fixed a couple of types as well.


kamilogorek (approved):

jordienr (approved):
lgtm

PR #33486, opened 2025-02-10
Feat: remember last visited route in logs FE-1418

jordienr:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES/NO

## What kind of change does this PR introduce?

When clicking the logs icon in the sidebar, it will remember which route you were last on and send you there instead of always sending you to /explorer

There's also a bug where using the project switcher breaks when users have query params in the URL, this fixes it.

## What is the current behavior?

Always sends users to /explorer

## To test

- Open project
- Click logs icon (goes to explorer)
- Click API/Storage whatever logs
- Click another link in the sidebar like Table Editor
- Click Logs again
- You should be in the page you visited in step 3 instead of explorer
- Enjoy the time you saved

PR #33756, opened 2025-02-21
Fix flakiness in logs test

joshenlim:
Noticed that playwright E2E tests are redirecting to /logs/logs for the logs test after [this PR](https://github.com/supabase/supabase/pull/33486). This PR attempts to fix that issue

Also removing the pause in the auth set up - so now we can look into running it through CI/CD

### DIFF
--- a/apps/studio/components/layouts/LogsLayout/LogsLayout.tsx
+++ b/apps/studio/components/layouts/LogsLayout/LogsLayout.tsx
@@ -15,36 +15,35 @@ interface LogsLayoutProps {
 }
 
 const LogsLayout = ({ title, children }: PropsWithChildren<LogsLayoutProps>) => {
-  const { isLoading, can: canUseLogsExplorer } = useAsyncCheckProjectPermissions(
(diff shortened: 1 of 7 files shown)


apps/studio/components/layouts/LogsLayout/LogsLayout.tsx
```diff
@@ -15,36 +15,35 @@ interface LogsLayoutProps {
...
+  const {
+    isLoading,
+    isSuccess,
+    can: canUseLogsExplorer,
+  } = useAsyncCheckProjectPermissions(PermissionAction.ANALYTICS_READ, 'logflare')
 
   const router = useRouter()
   const [_, setLastLogsPage] = useLocalStorageQuery(
     LOCAL_STORAGE_KEYS.LAST_VISITED_LOGS_PAGE,
-    router.pathname.split('/').pop()
```
joshenlim, line 26: reckon this should be the same behaviour as logs/index.tsx - default to /explorer

PR #33757, opened 2025-02-21
chore: add signin method to tracking

pamelachia:

YES

add signin method in tracking. for sso and github we pass url param during redirect. 

Tested locally for github and email and it shows on posthog
Tested in preview for sso


4L3k51 (approved):

apps/studio/pages/sign-in-mfa.tsx
```diff
@@ -46,7 +48,15 @@ const SignInMfaPage: NextPageWithLayout = () => {
           }
 
           if (data.currentLevel === data.nextLevel) {
-            sendEvent({ action: TelemetryActions.SIGN_IN, properties: { category: 'account' } })
+            if (signInMethod === 'github' || signInMethod === 'sso') {
```
loong, line 51: so if for any reason signInMethod is neither github nor sso, then no event gets emitted? maybe if signInMethod is invalid we should still emit the event but set method to "unknown" or "invalid"

pamelachia: idk if signin method can be invalid. It seems like the only way to login is github, sso and email so we covered all the bases. I was thinking it'll be better to not have unknown so we are stricter with the standards, but I can add unknown too if you think that's better.

loong: just worried if there are any changes later the events will not be emitted; they'll be silently missing. If we add an unknown at least there's a hint that something is wrong

loong (approved):
giving +1 ahead. still think it's better for event to always emit, avoiding silently missing events, hope we can do that before merging 👍

alaister (approved):
LGTM!

- Tested on preview with all available sign in methods

PR #33759, opened 2025-02-21
move templates in inline editor

SaxonF:
image

This moves the template selector from the bottom of the inline editor to the top. It uses a hover card for peeking at template code rather than replacing editor code. We aren't affecting editor state that way and can compare to our existing code too.

This PR also makes it so you can change prompt externally and have it show up.

## Steps to test

1. Enable the inline editor feature preview
2. Open editor
3. Click templates in top right
4. See list of templates
5. Hover templates to see preview
6. Click to insert


jordienr (approved):
Tested in preview all bueno

PR #33769, opened 2025-02-21
Add studio version in header for local self host

joshenlim:
Strategy is that Studio gets the version its on by an environment variable - the UI changes here is set up such that is the env var is present, we'll then render the Version popover

We'll do this by specifying a BUILD_VERSION env var when building the docker image - so this also needs to be updated in [here](https://github.com/supabase/supabase/blob/master/.github/workflows/publish_image.yml) eventually too which I might need @sweatybridge's help on this one as my Docker knowledge is not the best 😅 (I might be missing a detail or two too while i'm trying to piece what needs to be done in my mind 😬)

Nonetheless, verified working locally by building docker with the Dockerfile in studio via
`docker build . -f apps/studio/Dockerfile --target production -t studio:latest --build-arg BUILD_VERSION="20250219-7101314"`

Changes should not affect the hosted version.

On latest version:
image

Not on latest version:
image

Release schedule modal:
Approximate release date is based on last image release + 14 days (2 weeks)
image

Addresses https://linear.app/supabase/issue/FE-390/show-warning-on-localself-hosted-when-cli-version-is-out-of-date


apps/studio/Dockerfile
```diff
@@ -41,6 +41,9 @@ COPY --from=turbo /app/out/pnpm-lock.yaml ./
 # No need to clean cache because production uses standalone build
 RUN pnpm install --frozen-lockfile
 
+ARG BUILD_VERSION
+ENV STUDIO_VERSION="$BUILD_VERSION"
+
```
sweatybridge, line 46: You want to move this to line 29 so it's available on both dev and production images.

joshenlim: pushed an update :)

sweatybridge (approved):
Docker changes look good!

PR #33772, opened 2025-02-21
add pgbouncer config and status queries

jordienr:
adds 2 new queries for pgbouncer endpoints


jordienr:
> i'd opt to put all these under data/database - beside pooling-configuration-query

done 🫡

apps/studio/data/config/keys.ts
```diff
@@ -1,6 +1,4 @@
 export const configKeys = {
-  pgBouncerStatus: (projectRef: string | undefined) =>
```
jordienr, line 2: not used

joshenlim (approved):

PR #33775, opened 2025-02-21
allow nested routes for last visited route in logs

jordienr:


saltcod (approved):
Checked staging:
- going to logs → select a log → home → back to logs 
- logs → templates → home → logs → a log 
All looks good

PR #33780, opened 2025-02-21
feat: llms.txt

charislam:
https://zone-www-dot-com-git-feat-llmstxt-supabase.vercel.app/llms.txt

Closes DOCS-198


apps/docs/scripts/search/sources/github-discussion.ts
```diff
@@ -25,16 +25,22 @@ export type DiscussionsResponse = {
   }
 }
 
+const appId = process.env.SEARCH_GITHUB_APP_ID ?? process.env.DOCS_GITHUB_APP_ID
```
charislam, line 28: There are a bunch of environment variables that are differently named between Vercel and GitHub Actions. We should consolidate these later but I'm leaving them alone for now.

kiwicopple (approved):

PR #33795, opened 2025-02-23
docs: unnecessary duplicate safeGetSession() call in sveltekit.mdx

petrhora:
There is an unnecessary double call of safeGetSession(). Locals are already typed and assigned in authGuard() function in  hooks.server.ts. It's much simpler and more efficient to just reuse those locals in +layout.server.ts file. Please, if you disagree, let me know why.

## I have read the [CONTRIBUTING.md]
YES

This is a docs update.

Double call of safeGetSession despite locals being setup and assigned.

## What is the new behavior?
The session is now propagated through the app using the local.

## Additional context
No additional context needed.


j4w8n:
@BoogeyMan24 they may already have a plan on when they'll update the guides to coincide with asym jwts being released.

Also, I believe they'll be adjusting the timeline; so, the date for having asym jwts be the default will likely change.

BoogeyMan24:
Alright, since it will change anyways there is not point in updating it now.

PR #33801, opened 2025-02-24
fix: destructive query warning on run button

alaister:

Bug fix

Clicking the run button does not open the destructive query warning modal because the event parameter of the button is being passed to the `executeQuery` method, which is acting as a truthy value for the `force` parameter.

## What is the new behavior?

The event parameter is no longer passed so the query will not be force ran.


saltcod (approved):
Tested on preview and shows modal when deleting a table

PR #33802, opened 2025-02-24
Edge functions empty state

SaxonF:
image

Updates the empty state of the edge functions list to support the new ways we can create edge functions now. This also includes the ability to pass template code in to the web editor. 

## To test
1. Go to functions
2. If you have functions, set `hasFunctions` to false in functions/index.tsx.
3. Without the `edgeFunctionCreate` flag, notice only two options to create
4. Enable the `edgeFunctionCreate` flag, and notice three options and "start with a template" section
5. Test each of the options
6. Click a template and notice code from template is copied over to the web editor for editing

## Requires https://github.com/supabase/supabase/pull/33726


joshenlim:
found some issues that i can help fix
- New function page clicking "Edge Function" leads to a 404, and is causing a full page refresh (Due to not using the Link component from next/link)
- Clicking on the example doesnt render the code in the editor (I don't think Local Storage is the right solution here tbh - actually i realised that its not hooked up yet)
- Side menu doesnt select "Functions" when in /new

joshenlim (approved):
LGTM - verified changes on staging preview and pushed up fixes for issues identified
Small issue still whereby the list of edge functions isn't invalidating somehow when we deploy the function via the dashboard editor, left a comment to investigate separately

PR #33817, opened 2025-02-24
Support for Dedicated Pooler in Connection Pooling

joshenlim:
This is the first part for supporting swapping to PgBouncer as the pooler in the Database Settings -> Connection Pooling Configuration section

> [!NOTE]
> Changes are only applicable for hosted dashboard, so no need to test for local self-host

## Changes involved:
- Users can now select the pooler type (either Supavisor or PgBouncer)
image
- When swapping pooler types, we show a warning to change all connection strings to the new pooler within 2 hours
image
- For PgBouncer specifically: 
  - Can only be selected if the project's on a Micro Compute and above (Texts PgBouncer have been updated to dedicated pooler)
    - Tooltip if the org is on a free plan: 
image
    - Tooltip if the org is already on a paid plan:
image
  - Max client connections can be edited (For Supavisor, it remains fixed and disabled)
  - Statement mode is not supported (so status quo)
  - Notice for IPv4 add on will be shown if the project does not have the add on yet when swapping to PgBouncer
image
- Only when changing pooler types will we show a confirmation modal when saving the config (otherwise changes will save once the save button is hit) 
image
- Changes are feature flagged
  - Pooler type input field is hidden unless the flag is enabled
  - Supavisor configuration will be rendered unless the flag is enabled (then it'll depend on whether PgBouncer is enabled or not)

## What to test for:
- [ ] Verify that the UI changes are present and correct (including restrictions in place)
- [ ] Verify that you can update Supavisor / PgBouncer configurations with swapping pooler type
- [ ] Verify that you can update Supavisor / PgBouncer configurations without swapping pooler type
- [ ] Verify that with the feature flag off, the behaviour is status quo as per staging / production

## Remaining to-dos that will be in a separate PR:
- Connect UI's connection strings needs to be updated to be dependent on the pooler type
  - Depending on this i might update the warning RE using the right connection strings, might add a link to Connect UI
- Update Project Addons IPv4 panel -> "You do not need this" language, you'll now need it depending on the pooler type
- Nice to have: Show 2 hour countdown (Although this data is not available on the BE, so might need to deal with local storage)


alaister (approved):
LGTM!

- Tested on preview as directed

PR #33833, opened 2025-02-25
docs contributing layout

fsansalvadore:


charislam (approved):
The mobile menu expands into the Contributing menu but it's titled "Home", should we change this as well?

image

PR #33840, opened 2025-02-25
Feedback component polish

fsansalvadore:
Polish up the Feedback component styling.

Screenshot 2025-02-25 at 19 01 24


apps/docs/components/Feedback/Feedback.tsx
```diff
@@ -141,72 +141,83 @@ function Feedback({ className }: { className?: string }) {
   }
 
   return (
-    <section className={cn('@container px-5 mb-6', className)} aria-labelledby="feedback-title">
-      <h3 id="feedback-title" className="block font-mono text-xs uppercase text-foreground mb-4">
+    <section className={cn('@container', className)} aria-labelledby="feedback-title">
+      <span id="feedback-title" className="block font-mono text-xs text-foreground-light mb-3">
```
charislam, line 145: I think the h3 here works better for accessibility, if we change the selector in the other PR then it won't show up in the Table of Contents which I'm guessing is the main problem with the h3?

charislam (approved):

apps/docs/components/Feedback/Feedback.tsx
fsansalvadore: yeah that was a weird bug I was trying to avoid, but no need for this change now

PR #33857, opened 2025-02-26
Second phase of HSTS header settings

doublethink:
First phase of this rollout added the HSTS header via vercel config. This setting was less strict than the original Cloudflare sourced directive that was lost when the proxy was disabled.
That setting was maxage=31536000; includeSubdomains

This phase is temporary but takes a cautious approach to applying the includeSubdomains directive. The max age is set at 30 minutes so that we can back out if necessary. 

This setting is not intended to be in place for any longer than a month. Phase three will restore the longer maxAge directive as well as including the preload directive.

Part of SEC-120

### DIFF
--- a/apps/docs/next.config.mjs
+++ b/apps/docs/next.config.mjs
@@ -74,7 +74,7 @@ const nextConfig = {
         headers: [
           {
             key: 'Strict-Transport-Security',
-            value: process.env.VERCEL === '1' ? 'max-age=31536000' : '',
+            value: process.env.VERCEL === '1' ? 'max-age=1200; includeSubDomains' : '',
           },
           {
             key: 'X-Robots-Tag',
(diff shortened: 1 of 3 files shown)


alaister (commented):
This looks good!

One quick check before approving:
`/:path*` doesn't seem to apply to the base `/` path – do you want these headers to apply there too?

staaldraad (approved):

doublethink:
@alaister its not necessary for docs provided they stay in a subpath of the www.
And after trying various permutations of the root path that works in the other configs I figured it wasn't worth my mental health getting it to work :)

alaister:
@doublethink makes sense re the subpaths.

So perhaps only the root path on www needs it then? Since if I visit supabase.com that initial request isn't showing any HSTS header?
Screenshot 2025-02-27 at 13 19 44

doublethink:
@alaister perhaps disable cache to see the header on live?
the header throughout the app is still needed as we can't assume everyone visiting the site starts there
image

alaister (approved):
Ahh caching what a fun thing 😂

pamelachia (approved):

PR #33875, opened 2025-02-26
Add support for configuring realtime settings

joshenlim:
Fixes FE-1392

As per PR title - this should only apply for hosted (hidden on local/self-host), and is currently feature flagged

image

## What to test for
- [ ] No changes if feature flag is off
- [ ] Can render realtime settings if feature flag is on
- [ ] Can update realtime settings if feature flag is on

### DIFF
--- a/apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
+++ b/apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
@@ -2,14 +2,19 @@ import { zodResolver } from '@hookform/resolvers/zod'
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { useEffect } from 'react'
 import { SubmitHandler, useForm } from 'react-hook-form'
+import { toast } from 'sonner'
 import * as z from 'zod'
 
 import { useParams } from 'common'
 import { ScaffoldSection } from 'components/layouts/Scaffold'
 import { convertFromBytes } from 'components/to-be-cleaned/Storage/StorageSettings/StorageSettings.utils'
+import AlertError from 'components/ui/AlertError'
 import { FormSection, FormSectionContent, FormSectionLabel } from 'components/ui/Forms/FormSection'
 import { useRealtimeConfigurationUpdateMutation } from 'data/realtime/realtime-config-mutation'
-import { useRealtimeConfigurationQuery } from 'data/realtime/realtime-config-query'
+import {
+  REALTIME_DEFAULT_CONFIG,
+  useRealtimeConfigurationQuery,
+} from 'data/realtime/realtime-config-query'
 import { useCheckPermissions } from 'hooks/misc/useCheckPermissions'
 import {
   Button,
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -60,6 +60,17 @@
...
+    "postgres",
+    "firebase",
+    "storage",
+    "functions",
+    "database",
+    "auth"
+  ],
+  "packageManager": "pnpm@9.15.5",
+  "dependencies": {
+    "@supabase/shared-types": "0.1.80",
```
ivasilov, line 73: These imports are probably wrong?

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
```diff
@@ -0,0 +1,345 @@
...
+  const { ref: projectRef } = useParams()
+  const canUpdateConfig = useCheckPermissions(PermissionAction.REALTIME_ADMIN_READ, '*')
+
+  const { data, error, isLoading, isSuccess, isError } = useRealtimeConfigurationQuery({
+    projectRef,
+  })
+  const { mutate: updateRealtimeConfig, isLoading: isUpdatingConfig } =
+    useRealtimeConfigurationUpdateMutation({
+      onSuccess: () => {
+        form.reset(form.getValues())
```
ivasilov, line 54: Why is this `reset` needed?

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
```diff
@@ -0,0 +1,345 @@
...
+                      }
+                    >
+                      <FormSectionContent loading={isLoading}>
+                        <FormControl_Shadcn_>
+                          <Input_Shadcn_
+                            {...field}
+                            type="number"
+                            disabled={!canUpdateConfig}
+                            value={field.value || ''}
+                            {...form.register('connection_pool', { valueAsNumber: true })}
```
ivasilov, line 138: Are these `register` calls really needed?

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
```diff
@@ -0,0 +1,345 @@
...
+                    </FormSection>
+                  )}
+                />
+              </CardContent>
+              <CardContent>
+                <FormField_Shadcn_
+                  control={form.control}
+                  name="max_bytes_per_second"
+                  render={({ field }) => {
+                    const { value, unit } = convertFromBytes(max_bytes_per_second ?? 0)
```
ivasilov, line 219: you can use `field.value` here instead of `max_bytes_per_second`. The current approach with `form.watch` will cause entire form to rerender.

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
joshenlim: yup this is so that the form's default values get updated (and the cancel button goes away + save button disables)

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
joshenlim: yeah this one's a weird one that i've been observing with react hook form
so without this (given that we've correctly declared the form schema as `xxx: z.number()`, and assigned type `number` to the input field)

if we clear the input, then type in a number again, the value somehow gets saved in the form state as a string type thereafter
image

which causes the zod validation to fail when trying to save the form - so this register enforces that the input value _always_ gets parses as a number

although do let me know if there's a better way though 🙏

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
joshenlim: oooh good catch 🙏

package.json
joshenlim: you're right - these shouldn't be here sorry

alaister (approved):
LGTM!

PR #33877, opened 2025-02-26
Add pgbouncer logs FE-1420

jordienr:

YES

- Checks if the user has pgbouncer enabled and if so queries the pgbouncer_logs table for pooler logs

- always queries supavisor_logs

## What is the new behavior?

- queries correct logs table


apps/studio/pages/project/[ref]/logs/pooler-logs.tsx
```diff
@@ -1,24 +1,41 @@
...
+  if (pgBouncerLoading || isLoading) return <Loading />
+
+  function getPoolerTable() {
+    if (!IS_PLATFORM) {
+      return LogsTableName.PGBOUNCER
+    }
+
+    if (pgBouncerStatus?.active) {
+      return LogsTableName.PGBOUNCER
+    }
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/logs/pooler-logs.tsx:29]: hmm JFYI the connection pooling and connect ui changes don't use this as an indication on whether pgbouncer is active

we're currently using pgbouncer config -> pgbouncer_enabled instead, so am thinking we keep consistent in using that

Comment (jordienr) [reply apps/studio/pages/project/[ref]/logs/pooler-logs.tsx]: done thanks!

saltcod (approved):
Checked preview and works as expected.

PR #33883, opened 2025-02-26
Fix normalize strings to use function properly

akshayravikumar:


akshayravikumar:
Oops, made this on accident, my bad

PR #33893, opened 2025-02-27
Update metrics documentation

encima:

YES

Docs update

Metrics discusses a few deployment options and mentions the use of unofficial integrations

## What is the new behavior?

Make more genera deployment instructions for both grafana and prometheus, provide links to managed options and highlight unsupported integrations

### DIFF
--- a/apps/docs/content/guides/telemetry/metrics.mdx
+++ b/apps/docs/content/guides/telemetry/metrics.mdx
@@ -28,13 +28,39 @@ Your project's metrics endpoint is accessible at `https://<project-ref>.supabase
 > curl https://<project-ref>.supabase.co/customer/v1/privileged/metrics --user 'service_role:<service-role-jwt>'
 ```
 
-## Configuring Prometheus
+## Supabase Grafana
 
-For self-hosted Prometheus, modify your `prometheus.yaml` file to add a Supabase job, and set the `metrics_path`, `scheme`, `basic_auth` and `targets` parameters. For example:
+The pre-configured Supabase Grafana Dashboard is an advanced version of the [Dashboard's Database Reports](https://supabase.com/dashboard/project/_/reports/database). It visualizes over 200 database performance and health metrics and is updated every minute.
+
+![Supabase Grafana](/docs/img/guides/platform/supabase-grafana-prometheus.png)
+
+Instructions are included in the README for deploying the repository using docker.
+
+
+## Using the metrics endpoint in production
+
+To set up monitoring for your project, you will need two things:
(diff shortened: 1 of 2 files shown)


apps/docs/content/guides/telemetry/metrics.mdx
```diff
@@ -28,13 +28,39 @@ Your project's metrics endpoint is accessible at `https://<project-ref>.supabase
 > curl https://<project-ref>.supabase.co/customer/v1/privileged/metrics --user 'service_role:<service-role-jwt>'
 ```
 
-## Configuring Prometheus
+## Supabase Grafana
 
-For self-hosted Prometheus, modify your `prometheus.yaml` file to add a Supabase job, and set the `metrics_path`, `scheme`, `basic_auth` and `targets` parameters. For example:
+The pre-configured Supabase Grafana Dashboard is an advanced version of the [Dashboard's Database Reports](https://supabase.com/dashboard/project/_/reports/database). It visualizes over 200 database performance and health metrics and is updated every minute.
```
z0r1k, line 33: > ...updated every minute.

I don't know how to phrase it better, but one might think new metrics will be coming every minute, but I think that depends on prometheus scraping interval and what is meant here is that dashboard is refreshed every minute (also could be changed by a user).
Or did I get it wrong?

apps/docs/content/guides/telemetry/metrics.mdx
```diff
@@ -48,198 +74,42 @@ scrape_configs:
...
-        git clone https://github.com/supabase/supabase-grafana.git
-        ```
-    </StepHikeCompact.Code>
+In the left-hand menu, select `Data sources` and click `Add new data source`.
 
-  </StepHikeCompact.Step>
+Select `Prometheus` and enter the connection details for the Prometheus instance you have set up.
 
-  <StepHikeCompact.Step step={2} >
+Under `Interval behaviour`, set the `scraping interval` to 60s and test the data source. Once it has passed, save it.
```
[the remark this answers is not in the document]
z0r1k: haha, that misspelling is older than a country which made you, @github-actions bot.

z0r1k (approved):
Nice one 👍🏼

apps/docs/content/guides/telemetry/metrics.mdx
encima: You are right, those things should be separated!
- It does mean the metrics endpoint is refreshed every minute (so the scrape job should be 60s)
- The dashboard probably should not mention the refresh time as it is not configured to auto refresh by default

PR #33899, opened 2025-02-27
docs: add more filtering examples to the API docs

laurenceisla:
- Adds the logical operators `not` and `or`
- All Javascript examples are now present for Bash. Used `curl --get` to present each query parameter in a different line.

YES

Docs.

- There are no "Filtering" examples for the logical operators `not` and `or`.
- There are many Javascript examples for filters, but Bash only shows `id=eq.1`

## What is the new behavior?

image

## Additional context

Old behavior:

image


saltcod (approved):

saltcod:
Tested on preview, looks good!

apps/studio/components/interfaces/Docs/Snippets.ts
```diff
@@ -365,9 +387,14 @@ let { data: ${resourceId}, error } = await supabase
...
   .in('column', ['Array', 'Values'])
   .neq('column', 'Not equal to')
+
   // Arrays
   .contains('array_column', ['array', 'contains'])
   .containedBy('array_column', ['contained', 'by'])
+
+  // Logical operators
+  .not('column', 'like', 'Negate filter')
+  .or('some_column.eq.Some value, other_column.eq.Other value')
```
steve-chavez, line 397: This right now needs raw PostgREST syntax as mentioned on https://supabase.com/docs/reference/javascript/or, which will look mysterious to users (all other methods are somewhat intuitive, except this one).

Maybe link to the above doc and additionally link to https://docs.postgrest.org/en/latest/references/api/tables_views.html#logical-operators on https://supabase.com/docs/reference/javascript/or.

laurenceisla: > Maybe link to the above doc

I think it would look a bit bloated with links in the examples, besides there's already a "Documentation" button (API docs in the Table editor) and a "Learn more" (API Docs from the sidebar) that link to the client docs. I think the examples should be a "quick reference", so if more info is needed then they can access those links.

> additionally link to https://docs.postgrest.org/en/latest/references/api/tables_views.html#logical-operators on https://supabase.com/docs/reference/javascript/or.

Good idea, can't find a link to the docs for PostgREST operators anywhere (the ["Match the filter"](https://supabase.com/docs/reference/javascript/filter) section should definitely have it).

apps/studio/components/interfaces/Docs/Snippets.ts
```diff
@@ -365,9 +387,14 @@ let { data: ${resourceId}, error } = await supabase
   .is('column', null)
   .in('column', ['Array', 'Values'])
   .neq('column', 'Not equal to')
+
   // Arrays
   .contains('array_column', ['array', 'contains'])
   .containedBy('array_column', ['contained', 'by'])
+
+  // Logical operators
+  .not('column', 'like', 'Negate filter')
```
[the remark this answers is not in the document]
laurenceisla: I agree that it's more complex but not so much to have a separate subheading, I think. Besides, apart from being logical operators, they are similar in that they both use "raw PostgREST syntax".

steve-chavez (approved):

PR #33906, opened 2025-02-27
Fix/table editor export

saltcod:
exporting large numbers of table rows isn't working properly

To test, create a table with 300k rows, try and select all, then export it: 

CREATE TABLE users (
    id SERIAL PRIMARY KEY,
    name TEXT NOT NULL,
    email TEXT UNIQUE NOT NULL,
    created_at TIMESTAMP DEFAULT now()
);

INSERT INTO users (name, email)
SELECT 
    'User ' || generate_series(1, 300000),
    'user' || generate_series(1, 300000) || '@example.com';


SaxonF:
Tested locally and on staging. How long is it meant to take to export? It's been about 10 minutes so far. Smaller exports are almost immediate so it at least doesn't degrade current experience. 

In future, if we think this is a common action vs database backup, we should move to back-end with a download link.

saltcod:
Yeah it takes a long time for giant exports, part of the built-in rate limiting.  
Good idea on the separate download process. I can't think of another ticket like this one in the past while all the same.

kamilogorek:
@saltcod can you port this functionality to uploading as well please?

alaister (commented):
The requests worked, but the output csv was 0 bytes?

saltcod:
hummmm mine output as expected. Using the query for the users table above. 20mb, 300k rows exported after ~ 10mins


screenshot-2025-03-13-at-15 50 49

joshenlim:
@saltcod can verify 0 bytes here on my end - wondering if it might be cause the data is not being put together properly
image

alaister (approved):
LGTM!

- Tested on preview

PR #33911, opened 2025-02-28
chore: increase stale time for project addons / org subscription

kevcodez:
Occasionally results in a 429 in API - there are plans to fix this on the API level, but it's not a quick change, so hoping to reduce the likeliness of 429 with a higher stale time. Subscriptions and project addons rarely change and if they are changed, it's invalidated on the client anyway.


alaister (approved):
LGTM!

- Tested on preview

PR #33914, opened 2025-02-28
Fix connection pooling feature flag

joshenlim:
- Use connection_string parameter from API if available
- Fix loading state of Connection pooling if feature flag is off

### DIFF
--- a/apps/studio/components/interfaces/Connect/DatabaseConnectionString.tsx
+++ b/apps/studio/components/interfaces/Connect/DatabaseConnectionString.tsx
@@ -135,7 +135,10 @@ export const DatabaseConnectionString = () => {
       ? getConnectionStrings({
           connectionInfo,
           poolingInfo: {
-            connectionString: poolingConfiguration.connectionString,
+            connectionString:
+              'connection_string' in poolingConfiguration
+                ? poolingConfiguration.connection_string
+                : poolingConfiguration.connectionString,
             db_host: poolingConfiguration.db_host,
             db_name: poolingConfiguration.db_name,
             db_port: poolingConfiguration.db_port,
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/Connect/DatabaseConnectionString.tsx
```diff
@@ -135,7 +135,10 @@ export const DatabaseConnectionString = () => {
       ? getConnectionStrings({
```
ivasilov, line 135: Should this fix also be applied in `apps/studio/components/interfaces/Connect/ConnectTabContent.tsx`?

joshenlim: oh yes! thank you so much 🙏

apps/studio/components/interfaces/Settings/Database/ConnectionPooling/ConnectionPooling.tsx
```diff
@@ -169,22 +169,32 @@ export const ConnectionPooling = () => {
...
     },
   })
-  const { type, default_pool_size, max_client_conn } = form.watch()
+  const { type, default_pool_size } = form.watch()
   const error = useMemo(
-    () => (type === 'PgBouncer' ? pgbouncerConfigError : supavisorConfigError),
-    [type]
+    () =>
+      allowPgBouncerSelection && type === 'PgBouncer' ? pgbouncerConfigError : supavisorConfigError,
+    [type, allowPgBouncerSelection]
```
ivasilov, line 176: You need to add `pgbouncerConfigError` and `supavisorConfigError` as deps here, otherwise the value won't change when an error happens.

Same for the rest of the useMemos.

joshenlim: ahh okay gotcha on this! 🙏

ivasilov (approved):

alaister (approved):

PR #33951, opened 2025-03-03
Link to auth provider

SaxonF:
This changes behaviour of providers list so that on clicking a provider it adds a provider query param. This query param is passed in to providerform which has an added useEffect to determine whether to show or hide the provider sheet. Sheet on change now just updates query param.

## To test
1. Go to auth/providers
2. Click a provider
3. Notice query param addition
4. Refresh the page and notice provider is open
5. Close sheet and notice provider query param goes away

### DIFF
--- a/apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersForm.tsx
+++ b/apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersForm.tsx
@@ -1,5 +1,6 @@
 import { ExternalLink } from 'lucide-react'
 import Link from 'next/link'
+import { useRouter } from 'next/router'
 
 import { useParams } from 'common'
 import {
@@ -23,6 +24,8 @@ import { ProviderForm } from './ProviderForm'
 
 export const AuthProvidersForm = () => {
   const { ref: projectRef } = useParams()
+  const router = useRouter()
+  const { provider: queryProvider } = router.query
   const {
     isLoading,
     error: authConfigError,
@@ -95,6 +98,7 @@ export const AuthProvidersForm = () => {
                   config={authConfig!}
                   provider={providerSchema as unknown as Provider}
                   isActive={isActive}
+                  queryProvider={queryProvider as string}
                 />
               )
             })}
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersForm.tsx
```diff
@@ -23,6 +24,8 @@ import { ProviderForm } from './ProviderForm'
 
 export const AuthProvidersForm = () => {
   const { ref: projectRef } = useParams()
+  const router = useRouter()
+  const { provider: queryProvider } = router.query
```
alaister, line 28: You can use `useParams()` here to avoid the `as string` below

alaister (approved):
LGTM!

One small possible improvement:

apps/studio/components/interfaces/Auth/AuthProvidersForm/ProviderForm.tsx
```diff
@@ -113,10 +116,35 @@ export const ProviderForm = ({ config, provider, isActive }: ProviderFormProps)
...
+    }
+  }
+
+  // Handle clicking on a provider in the list
+  const handleProviderClick = () => {
+    // Update URL with provider query param
+    const query = { ...router.query, provider: provider.title }
+    router.replace({ pathname: router.pathname, query }, undefined, { shallow: true })
+  }
+
```
joshenlim, line 143: i'd opt to use `useUrlState` here actually, then it'll remove the need for all the manual router updates with its inbuilt `setParams`

alaister: Good shout – I forgot about that hook!

apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersForm.tsx
joshenlim: JFYI we can probably just call useParams in ProviderForm as well, no need to pass as a prop below

joshenlim (approved):
LGTM! Verified changes in staging preview auth providers

joshenlim (approved):
LGTM! Verified changes on preview

PR #33962, opened 2025-03-03
Hide pool mode field in connection pooling

joshenlim:
Fixes FE-1445

As per PR title - also removes `pool_mode` as a param in supavisor-configuration-update-mutation
but we're manually setting pool_mode as 'transaction' in the mutation itself as I noticed a bug in the API end for updating pgbouncer or supavisor's configuratiojn

We can remove that eventually tho once the API is fixed

Also removes the session mode warning banner in Connect UI if dedicated pooler is selected 

## To test:
- Ensure `dualPoolerSupport` flag is off (so as to ensure that we're testing what's seen on production)
- Pool mode field no longer present in connection pooling under settings/database
- Can still update pool_size for connection pooling


jordienr (approved):
LGTM Tested in staging preview.

PR #33966, opened 2025-03-03
Chore/support updating email and unlinking identities

joshenlim:
Fixes FE-1438

> [!WARNING]
> Dependant on API PR: https://github.com/supabase/infrastructure/pull/21806
> Ready on staging - ensure that changes are through to prod prior to merging

Changes are mainly in Account Preferences (/account/me) - ignore the TS issues for now, will fix again once the API PR goes through cause I think there's some discrepancies with the API typing

## Some context
- An account can only have GH, or Email, or both - SSO identities cannot be associated with any other identity (So SSO accounts are the only identity for that account)
- To create an account with Email + GH identities, create an Email + Password account first with an email that matches the GH account's email. Thereafter, log out, and log in via GH with that same email - accounts will then be linked afterwards
- Note that if you unlink the identity that you're currently logged in with, we currently don't log users out immediately because there's no way atm to identity which identity the user is currently logged in with
  - Auth team is aware of this and will work on the required changes to support this

## Changes involved
- Account Information is now replaced with Account Identities
- Unlink CTA only shows up if there's more than 1 identity
- Reset password CTA only shows up for the email identity
- Update email address modal content will vary based on the provider

## Things to test
- [ ] Can view all identities of the account
- [ ] Can change email for email password identity (And verify that the user flow is clear)
  - I won't explain how the flow works in this PR, but if you have trouble figuring that you - means that we need to improve the UX here
- [ ] Update email address modal contents are contextualized to the provider
- [ ] Can unlink an identity if there's more than 1 identity tied to the account


apps/studio/components/interfaces/Account/Preferences/ChangeEmailAddress.tsx
```diff
@@ -0,0 +1,133 @@
...
+  return (
+    <DialogSection className="flex flex-col gap-y-2">
+      <p className="text-sm">
+        Email addresses for SSO should be updated through your identity provider
+      </p>
+      <ol className="flex flex-col gap-y-0.5 text-sm ml-4 pl-2 list-decimal text-foreground-light">
+        <li>Log out of Supabase</li>
+        <li>Change your Primary Email in your identity provider</li>
+        <li>Log back into Supabase via SSO with the new, desired email</li>
+      </ol>
```
kangmingtay, line 54:
```suggestion
      <p className="text-sm">
        Email addresses for SSO should be updated through your identity provider
      </p>
      <ol className="flex flex-col gap-y-0.5 text-sm ml-4 pl-2 list-decimal text-foreground-light">
        <li>Contact the owner / admin for your team to change your email</li>
      </ol>
```

kangmingtay: SSO emails are managed through the company's IT admin

kangmingtay (commented):
@joshenlim is it also possible to show the new email as "pending" after initiating the update? 
image

kangmingtay:
@joshenlim hmm i think from a user's POV, unlinking an identity isn't super clear. I took a look at sentry and i like how they phrase it as "disconnecting". Maybe we can also add something like: "You will no longer be able to use <identity> to sign-in".

image

kangmingtay:
@joshenlim was testing this on staging with a sso user (yes the api changes are out already :))

the identity displayed is kinda weird - i think we don't have to show the `provider` field here for sso identities because it's not useful information
image

i think we shouldn't even allow the user to update their email for sso identities:
image

joshenlim:
addressed all the other feedbacks otherwise 😄

apps/studio/components/interfaces/Account/Preferences/AccountIdentities.tsx
```diff
@@ -0,0 +1,192 @@
...
+      </Dialog>
+
+      <ConfirmationModal
+        variant="warning"
+        size="small"
+        loading={isUnlinking}
+        visible={!!selectedProviderUnlink}
+        title={`Unlink ${getProviderName(selectedProviderUnlink ?? '')} identity`}
+        onCancel={() => setSelectedProviderUnlink(undefined)}
+        onConfirm={() => onConfirmUnlinkIdentity()}
```
SimonGodefroid, line 181:
```suggestion
        onConfirm={onConfirmUnlinkIdentity}
```

apps/studio/components/interfaces/Account/Preferences/ChangeEmailAddress.tsx
```diff
@@ -0,0 +1,134 @@
...
+                <FormControl_Shadcn_>
+                  <Input_Shadcn_ {...field} placeholder="example@email.com" />
+                </FormControl_Shadcn_>
+              </FormItemLayout>
+            )}
+          />
+        </DialogSection>
+
+        <DialogFooter>
+          <Button type="default" disabled={isLoading} onClick={() => onClose()}>
```
SimonGodefroid, line 124:
```suggestion
          <Button type="default" disabled={isLoading} onClick={onClose}>
```

alaister (commented):
small light mode issue:

Screenshot 2025-03-20 at 12 27 15

alaister (approved):
LGTM!

- Tested as per instructions

A couple of small nits:

staaldraad (approved):

PR #34011, opened 2025-03-05
Multiple assistant chats

SaxonF:
image

Adds the ability to create and store multiple chats per project. 
- Create, rename, switch to and delete chats
- Chats are scoped per project so switching project also switches chat
- Stored in local storage

This PR also introduces a useAssistant hook which is the beginnings of a slight refactoring taking logic out of app-state and moving in to the hook. Eventually all interactions should be done through this hook. As part of this refactoring we are also removing the AIAssistantPanel as a general clean up.


joshenlim (approved):
Checked on preview:
- Assistant chats are scoped to project
- Switching chats within a project
- Deleting chats
- RLS policies via Assistant

Minor tweaks added:
- Prevent deleting a chat if it's the only one left
- Use ButtonTooltip for Assistant header buttons
- Add ScrollArea to chat selector popover
- Condensed new chat CTA into view chats popover
- Added chat name to assistant header
- Added `name` parameter to `newChat` -> noticed that it could create a lot of "New Chat" easily when editing / creating entities via the assistant
- Updated each call of `newChat` to include a name (e.g Create policy -> "Create new policy")

PR #34014, opened 2025-03-05
fix: bytea null values

alaister:

Bug fix

When user has `table_a` with a `bytea` column, and `table_b` with a reference to `table_a`, the foreign row selector will error when `table_a` has a `null` value in the `bytea` column.


saltcod (approved):
Confirmed fix on staging!

joshenlim (approved):
LGTM! Verified code changes only, although I saw in real life Alaister verifying the bug fix working

PR #34015, opened 2025-03-05
Cleanup reports

joshenlim:
The following reports are planned to be deprecated as the data returned from these charts are not intrinsically helpful of a daily aggregate basis:
- total_realtime_ingress
- total_rest_options_requests
- total_auth_ingress
- total_auth_get_requests
- total_auth_post_requests
- total_auth_patch_requests
- total_auth_options_requests
- total_storage_options_requests
- total_storage_patch_requests
- total_options_requests

For those charts that are affected, they'll no longer be selectable in the Add block dropdown, and we'll be rendering the following empty state instead if they're currently used in custom reports:
image


jordienr (approved):
LGTM
Tested creating a report with deprecated charts and opening it with this branch

alaister (approved):
Joshen says it's good

PR #34016, opened 2025-03-05
fix: stop sending view definition query all the time

alaister:
It seems the view definition query was sending for any entity type. This limits it to views only and also only when the modal is open


Hardanish-Singh (approved):

joshenlim (approved):
Verified that view definitions query is only triggered when the autofix modal is open

PR #34019, opened 2025-03-05
Refactor ChartConfig.tsx

joshenlim:
Slight refactor of ChartConfig but also solves a re-rendering issue in the SQLEditor charts which you can reproduce by updating any chart config like checking "Cummulative"

Before:

https://github.com/user-attachments/assets/56bf09c6-92e0-433b-8f7b-05643a7d861f

After:

https://github.com/user-attachments/assets/f372ccb8-de09-4b55-870b-ee9df1e7e004


alaister (approved):
LGTM!

- Tested on preview

PR #34023, opened 2025-03-05
fix: double ? on sign in urls

alaister:

Bug fix

Screenshot 2025-03-05 at 19 07 33

## What is the new behavior?

The URL gets created properly.


ivasilov (approved):

joshenlim (approved):
LGTM - Verified on prod that logging in works fine, and can verify that the query params are formed correctly with only one `?`

PR #34046, opened 2025-03-06
Skip JSON stringify cell values on string types to remove unnecessary double quotes

joshenlim:
Before:
image

After:
image


saltcod (approved):
Checked using the colors quickstart, works as expected. Nice one!

PR #34093, opened 2025-03-10
fix: allow null values for update bucket

alaister:

Bug fix

These values cannot be reset/removed:
Screenshot 2025-03-10 at 10 43 48

## What is the new behavior?

They can be reset


joshenlim (approved):
LGTM! Verified changes on staging preview

PR #34094, opened 2025-03-10
Fix uploading files to wrong bucket when navigating to another bucket while files were being uploaded

joshenlim:
As per PR title

### To reproduce:
- Upload a bunch of files (I did 70)
- While the upload is going on, switch to another bucket or create a new one (which will redirect to the new bucket)
- Stay on the bucket until the upload completes
- Notice that the files get uploaded to the bucket that you're currently on the dashboard


alaister (approved):
LGTM!

- Tested on preview

PR #34099, opened 2025-03-10
Table editor fix is extra options state for PK columns

joshenlim:
### To reproduce
- Table editor -> New table
- Uncheck the default id column as primary key
- set id column to is nullable
- check if back as primary key
- Notice that the number here is 2, despite is nullable no longer showing up as an option in extra options
image


alaister (approved):
LGTM!

- Tested on preview

PR #34100, opened 2025-03-10
docs: clarify pool size and session mode relationship

charislam:


charislam:
@TheOtherBrian1 this is to clarify some customer questions around session mode connections counting against the underlying pool, but I'm not sure that I've explained this the best way, and you have more expertise in this area. Feel free to suggest different wording or explanations if it would help.

soedirgo (approved):

charislam:
closing after discussion with @TheOtherBrian1 since the entire issue of pool size configuration is overly complicated and deserves a more thorough approach

PR #34102, opened 2025-03-10
Fix: Fix the set null button when the value is empty

ivasilov:
To replicate the issue:
1. Go to table editor
2. Try to edit a text value (via the inline edit)
3. Set the value to empty string and save it.
4. The cell should say `EMPTY`
5. Expand the inline edit again on that cell.
6. Click Set to NULL, click confirm on the modal.
7. The modal never closes.


jordienr (approved):
LGTM

joshenlim (approved):
LGTM

PR #34104, opened 2025-03-10
ci(docs): improve linting of external PRs

charislam:
On internal PRs, lint results are displayed as PR comments, but due to reduced permissions, this doesn't happen on external PRs. This makes the linting experience for external contributors not so great: linting results are hidden in actions check logs, in a hard-to-read JSON format.

This PR alters the behaviour for external PRs so that the results are instead written as a comment instead:

image


gregnr (approved):
Nice 🎉

PR #34109, opened 2025-03-10
fix: join org url search params bug

alaister:

Bug fix

Since https://github.com/supabase/supabase/pull/34023, all the query params now correctly get sent over to the join page, which surfaced another bug with multiple query params on the join page.

## What is the new behavior?

The join page now works with any number of query params correctly.

## Additional context

This only occurred when the user was signed out, then signed in with github and got redirected to the join org page. They could still join the org but would have to click the link in their email again.


joshenlim (approved):
Verified changes on staging preview by inviting a GH account, and accepting the invite while first being logged out, then log in with a GH account

PR #34120, opened 2025-03-11
Fix openai api key empty state not showing in local

joshenlim:
Realised that the OPENAI_API_KEY empty state was not showing up, because we were sending a 404 error for check-api-key endpoint which in turn the react query was treating it as an error. Updated the check-api-key endpoint to return a 200 instead (which i think makes sense - its still a correct response if the API key is not set)

image


alaister (approved):
LGTM!

DoiiarX:
Do you know how to use other non-openai llm api?

PR #34121, opened 2025-03-11
fix: update column type and add foreign key

alaister:

Bug fix

If you update a foreign key type and try to add a foreign key at the same time, you'll get a type mismatch error involving the old column type.


jordienr (approved):
Probado en el entorno de previsualización, gracias.

joshenlim (approved):
LGTM - verified changes on staging preview table editor

PR #34146, opened 2025-03-12
Add backup type label

joshenlim:
There's a confusion regarding downloading of backups from the scheduled backups page in the database section of a project. 
- Logical backups can be downloaded, but not physical backups
- The scheduled backups page can show physical  backups too depending on its set up
- But the download button is hidden if the backup is physical, which is the source of the confusion.
- Adding a label to indicate what type of backup it is + tooltip with disabled state for download button instead of hiding it

image

RestoringState also has a CTA for "Download backup" which could be confusing if you land in this state from restoring a backup.
- Updated CTA to indicate "latest" backup
- Updated logic to download only latest logical backup (physical will fail)

Note to self:
https://supabase.com/blog/postgresql-physical-logical-backups
https://supabase.com/docs/guides/troubleshooting/download-logical-backups

### DIFF
--- a/apps/studio/components/interfaces/Database/Backups/BackupItem.tsx
+++ b/apps/studio/components/interfaces/Database/Backups/BackupItem.tsx
@@ -3,10 +3,11 @@ import { Download } from 'lucide-react'
 
 import { useParams } from 'common'
 import { ButtonTooltip } from 'components/ui/ButtonTooltip'
+import { InlineLink } from 'components/ui/InlineLink'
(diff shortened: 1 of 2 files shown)


saltcod:
getting strong "what on earth does Logical mean?" vibes. 
Since the distinction between these is logical vs physical, could we instead do an `i` info icon to explain (with something like "this is a physical backup, and only logical backups can be downloaded...read more...")

screenshot-2025-03-12-at-14 50 54

alaister (approved):
LGTM!

- Tested on preview

PR #34152, opened 2025-03-12
improve postgres logs rows and detail

jordienr:
- Adds hint and detail to postgres logs rows
- Adds hint query and query to the first level of the detail panel

## To test this:
- Run some sql that generates logs with hints or detail (you can ask assistant)
I used this
```sql
BEGIN;

INSERT INTO public.customers (id, first_name, email) VALUES (1, 'John', 'john@example.com');

-- Attempt to insert the same id again to trigger a unique constraint violation
INSERT INTO public.customers (id, first_name, email) VALUES (1, 'Jane', 'jane@example.com');

ROLLBACK;
```
- Check that the `detail` or `hint` field shows up in the results 

CleanShot 2025-03-12 at 17 11 35@2x


joshenlim (approved):
LGTM! Verified changes on preview

PR #34020, opened 2025-03-05
chore: remove react-tracked part 1

alaister:
## What kind of change does this PR introduce?

chore/fix

- Removes rows part of react-tracked store
- Fixes cmd+arrow navigation in the table editor

PR #34153, opened 2025-03-12
chore: remove react-tracked part 2

alaister:
Follow up to #34020


apps/studio/components/grid/SupabaseGrid.utils.ts
```diff
@@ -119,36 +116,113 @@ export function parseSupaTable(table: Entity): SupaTable {
...
 }
 
-export function getStorageKey(prefix: string, ref: string) {
-  return `${prefix}_${ref}`
+export const saveTableEditorStateToLocalStorageDebounced = AwesomeDebouncePromise(
+  saveTableEditorStateToLocalStorage,
+  500
+)
+
+export function useLoadTableEditorStateFromLocalStorage({
```
joshenlim, line 177: this function doesn't seem to be used? can remove probably

apps/studio/components/grid/components/formatter/ForeignKeyFormatter.tsx
```diff
@@ -14,19 +14,18 @@ import { NullValue } from '../common/NullValue'
...
   const { project } = useProjectContext()
 
   const { tableId, row, column } = props
-  const id = tableId ? Number(tableId) : undefined
 
   const { data } = useTableEditorQuery({
     projectRef: project?.ref,
     connectionString: project?.connectionString,
-    id,
+    id: tableId,
```
joshenlim, line 28: sorry if i might be overlooking something - but wondering if we could just use the table id from the URL via useParams actually? or is the tableId here potentially not the table that the user is looking at?

apps/studio/components/grid/utils/gridColumns.tsx
joshenlim, line 1: just to double check - getCellRenderer, for case 'foreign_key' -> we don't need to pass projectRef as prop anymore right?
image

apps/studio/components/grid/components/header/RefreshButton.tsx
```diff
@@ -6,14 +6,14 @@ import { tableRowKeys } from 'data/table-rows/keys'
...
+  tableId?: number
   isRefetching?: boolean
 }
 
-const RefreshButton = ({ table, isRefetching }: RefreshButtonProps) => {
+const RefreshButton = ({ tableId, isRefetching }: RefreshButtonProps) => {
   const { ref } = useParams()
   const queryClient = useQueryClient()
-  const queryKey = tableRowKeys.tableRowsAndCount(ref, table.id)
+  const queryKey = tableRowKeys.tableRowsAndCount(ref, tableId)
```
joshenlim, line 16: likewise here - wondering if we can just use id from `useParams`

apps/studio/pages/project/[ref]/editor/[id].tsx
```diff
@@ -8,12 +7,11 @@ import DefaultLayout from 'components/layouts/DefaultLayout'
 import { useProjectContext } from 'components/layouts/ProjectLayout/ProjectContext'
 import TableEditorLayout from 'components/layouts/TableEditorLayout/TableEditorLayout'
 import { useTableEditorQuery } from 'data/table-editor/table-editor-query'
-import { TableEditorTableStateContextProvider } from 'state/table-editor-table'
+import { TablesData } from 'data/tables/tables-query'
+import { useCallback } from 'react'
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/editor/[id].tsx:11]: tiny nit - can shift up to the first section of import statements

joshenlim (commented):
Checked the following on the Table Editor:
- [x] Adding a row
- [x] Updating a row
- [x] Deleting a row via context menu
- [x] Deleting a row via checkbox
- [x] Update row value via grid
- [x] Update row via side panel editor
- [x] Add a column
- [x] Update a column
- [x] Delete a column
- [x] Reorder a column + refresh to check persist
- [x] Add multiple sorts
- [x] Add multiple filters

❌ Add multiple sorts + filters (Recording below)

https://github.com/user-attachments/assets/6a9304a0-95be-46d6-81d2-5f9cbab24f9b

lemme know if i should be testing anything else!

apps/studio/components/grid/components/header/RefreshButton.tsx
alaister: this one seems to be passed in from a foreign key:

Screenshot 2025-03-21 at 15 16 56

apps/studio/components/grid/components/formatter/ForeignKeyFormatter.tsx
alaister: I think this would probably be alright to switch, but I'm gonna err on the side of caution and leave it as a prop – since it's just the one prop!

joshenlim (approved):
Verified that identified issues have been fixes - also checked switching schemas ✅

saltcod (approved):

PR #34175, opened 2025-03-13
Advisor Rules (Part 1)

joshenlim:
*Changes for this feature will come in several PRs to keep things smaller*

## Adds a UI to add rules for advisor lints 
image
image

This is just part 1 and is not fully complete just yet (hence feature flagged, and still okay to go through to prod)
Actual filtering of lints is still not ready on the API end yet (API PR [here](https://github.com/supabase/infrastructure/pull/22005))

## Features to test
- [ ] Can add a rule
- [ ] Can remove a rule (Currently cannot edit a rule)
- [ ] UI is hidden behind feature flag

## Next part will cover
- [ ] Can edit a rule
- [ ] Add CTA in Advisor reports to ignore a specific lint
- [ ] Rule can filter down to Entity/Item


apps/studio/components/layouts/AdvisorsLayout/AdvisorsMenu.utils.ts
```diff
@@ -28,5 +32,20 @@ export const generateAdvisorsMenu = (project?: Project): ProductMenuGroup[] => {
...
       ],
     },
+    ...(IS_PLATFORM && features?.advisorRules
+      ? [
+          {
+            title: 'Configuration',
+            items: [
+              {
+                name: 'Advisor Rules',
+                key: 'advisor-rules',
```
alaister, line 42: nit: this menu item doesn't show as active:

Screenshot 2025-03-25 at 15 17 12

apps/studio/components/interfaces/Advisors/CreateRuleSheet.tsx
```diff
@@ -0,0 +1,272 @@
...
+                control={form.control}
+                render={({ field }) => (
+                  <FormItemLayout layout="vertical" className="px-5" label="Lint name">
+                    <FormControl_Shadcn_>
+                      <Select_Shadcn_ onValueChange={field.onChange} defaultValue={field.value}>
+                        <SelectTrigger_Shadcn_ className="col-span-8">
+                          <SelectValue_Shadcn_ placeholder="Select a lint name" />
+                        </SelectTrigger_Shadcn_>
+                        <SelectContent_Shadcn_>
+                          {lintInfoMap.map((x) => (
```
alaister, line 150: nit: I can see both performance and security lints in this list... kinda strange having this at all tbh 😅
 
Screenshot 2025-03-25 at 15 20 43

alaister (approved):
LGTM!

- Followed testing instructions on preview

Couple of small nits:

SaxonF:
My general feedback is purely from a usability standpoint and that I think we could still simplify this. For example, in the below, does this lint still apply to the whole team or just me?

image

I could be missing a specific requirement from a specific customer but an alternative approach might be to:
- List all advisor lints 
- Each lint has:
  - enabled for everyone (green light on list item)
  - if true, show field for "hide for team members" which is a list of team members
  - if false, show field for "show for specific team members" which is a list of team members

This is opposed to creating individual rules for each lint item.

apps/studio/components/interfaces/Advisors/CreateRuleSheet.tsx
joshenlim: yeah agreed - i was wondering whether to hide this field tbh, think i will hahaha given that you also find this weird

joshenlim: will remove as such, and also beefed up the description under "What this rule means" a bit more
image

Giant-Space-Bee:
Where is Part 2? :) I'd love to supress the 34 warnings I have about allowing anon users...

PR #34182, opened 2025-03-13
Fix in returnTo param

doublethink:

YES

Security fix

Due to nextjs dynamic routing behavior, the regex to filter out non-paths was not working as intended

## What is the new behavior?

The new regular expression will check for more than the first word character within the pattern instead of just the first in detecting a valid path

## Additional context

Fixes SEC-230


joshenlim (approved):
LGTM - tested logging in via GH, logging in via Email Password + logging in with redirects

PR #34216, opened 2025-03-16
chore: inform about non-encrypted instance risk in doc

### LINKED ISSUE #34215: Deceptive self-hosting doc luring user into exposing non-encrypted instance
# Improve documentation

## Link 

https://supabase.com/docs/guides/self-hosting/docker

## Describe the problem

The video showing the user how to setup a self hosted instance on DigitalOcean never informs that the resulting instance is not SSL protected.

## Describe the improvement

Warnings are added under the heading **`Securing your services`** and above the video itself.

## Additional context

I don't believe Supabase should assume that a user playing around with self-hosting fully understand the implication of deploying an instance publicly available on the Internet.

kler:
- fixes #34215

YES

Docs update

The video showing the user how to setup a self hosted instance on DigitalOcean never informs that the resulting instance is not SSL protected.

## What is the new behavior?

Warnings are added under the heading Securing your services and above the video itself.

## Additional context

I don't believe Supabase should assume that a user playing around with self-hosting fully understand the implication of deploying an instance publicly available on the Internet.

### DIFF
--- a/apps/docs/content/guides/self-hosting/docker.mdx
+++ b/apps/docs/content/guides/self-hosting/docker.mdx
@@ -159,6 +159,13 @@ You'll want to update the Studio(Dashboard) frequently to get the latest feature
 
 While we provided you with some example secrets for getting started, you should NEVER deploy your Supabase setup using the defaults we have provided. Follow all of the steps in this section to ensure you have a secure setup, and then [restart all services](#restarting-all-services) to pick up the changes.
 
+<Admonition type="danger">
+
+Your instance is running without any type of encryption.
+This guide does not walk you through how to secure your instance using SSL (`https://...`). Before exposing your instance on the Internet, make sure all traffic is secure. This is outside of the scope for this guide for the time being.
+</Admonition>
+
+
 ### Generate API keys
 
 We need to generate secure keys for accessing your services. We'll use the `JWT Secret` to generate `anon` and `service` API keys using the form below.
@@ -437,6 +444,9 @@ docker rm supabase-analytics
 ## Demo
 
 A minimal setup working on Ubuntu, hosted on DigitalOcean.
+
+After completing the guide your DigitalOcean instance will be running without any type of encryption.
+This means that anything you communicate with your instance could be intercepted. This guide does not walk you through how to secure your instance using SSL (`https://...`).
 
 <div className="video-container">
   <iframe


apps/docs/content/guides/self-hosting/docker.mdx
```diff
@@ -159,6 +159,13 @@ You'll want to update the Studio(Dashboard) frequently to get the latest feature
 
 While we provided you with some example secrets for getting started, you should NEVER deploy your Supabase setup using the defaults we have provided. Follow all of the steps in this section to ensure you have a secure setup, and then [restart all services](#restarting-all-services) to pick up the changes.
 
+<Admonition type="danger">
+
+Your instance is running without any type of encryption.
+This guide does not walk you through how to secure your instance using SSL (`https://...`). Before exposing your instance on the Internet, make sure all traffic is secure. This is outside of the scope for this guide for the time being.
+</Admonition>
+
+
```
staaldraad, line 168: this would fit better under the existing warning about securing services: https://github.com/supabase/supabase/pull/34216/files#diff-3552d44e92efffffff97b64076ea318ab06f07980d335d620f2bfe5447167135R90-R94

```
Your app is now running with default credentials and http endpoints do not have TLS configured.
```

We are discussing adding TLS configuration guidance to the `Securing your services` section.

apps/docs/content/guides/self-hosting/docker.mdx
```diff
@@ -438,6 +445,9 @@ docker rm supabase-analytics
 
 A minimal setup working on Ubuntu, hosted on DigitalOcean.
 
+After completing the guide your DigitalOcean instance will be running without any type of encryption.
+This means that anything you communicate with your instance could be intercepted. This guide does not walk you through how to secure your instance using SSL (`https://...`).
+
```
staaldraad, line 450: The section is unnecessary given the warning above about securing services. `Demo` and `minimal setup` fully implies that this video is not showing the full setup nor purports to represent a production ready deployment

staaldraad (changes requested):
Thanks for the PR @kler. Suggested changes in-line.

There is an existing warning where we can add this additional note.

apps/docs/content/guides/self-hosting/docker.mdx
kler: I don't quite follow - this warning *is* under the existing warning about securing services, isn't it?

staaldraad: Rewording, "under" here refers to being within the same warning block. 

The warning is about the risk of deploying directly with defaults and directs you to a section about securing your setup. Thus it makes sense adding additional security+defaults language in the same location.

kler: The current warning block is under the heading **Installing and running Supabase**.
I don't think the warning I've proposed to add belongs under that heading, but rather under the heading **Securing your services**.

It's misleading to have a complete section called **Securing your services** and bluntly talk about changing the default credentials, while leaving it to the user to communicate the very same keys in plain text... Why even go trough the trouble of changing the defaults if you're gonna show them to world anyway.

I still believe my edit is more appropriately placed, including the repetition.

apps/docs/content/guides/self-hosting/docker.mdx
kler: > The section is unnecessary given the warning above about securing services. `Demo` and `minimal setup` fully implies that this video is not showing the full setup nor purports to represent a production ready deployment

The problem is not that the use wouldn't understand that it's a demo and that it's not production ready. The problem is the the user is lured into fully exposing everything he does on the Internet. That's just plain irresponsible by the video author, and in turn this guide.

I am not concerned that the user would create his production setup using this guide, but _I am_ concerned that a user would for instance use his own email, reuse passwords, possibly putting sensitive data like his address in his cloud demo. _Hopefully_ the user understands the implications of this, but I could almost guarantee that not all would realise what it would mean to put the instance fully exposed on the Internet.

I don't see why you can't just accept this PR and make it clear to the user. It's ironic just ironic disputing three sentences that are in fact accurate, with the only possible drawback of being repetitive. Again, the very good reason to repeat this is that many users will see the video and jump straight to it.

staaldraad: Apologies for the delay.

> I don't see why you can't just accept this PR and make it clear to the user. It's ironic just ironic disputing three sentences that are in fact accurate,

There is no dispute, docs should be consistent and clear however. Putting the warning front-and-center with the other warning makes sense. Otherwise, the other warning should also be duplicated in this section to maintain consistency. Because the video also doesn't cover changing default credentials, right?

> After completing the guide your DigitalOcean instance will be running without any type of encryption.

This can be better phrased to clarify that "without any type of encryption" refers to the network end-point. Encryption is too generic of a term in this case, `Supabase Studio will be accessible over clear-text HTTP and will not have the HTTPS end-point configured` would for example be accurate

PR #34228, opened 2025-03-17
fix crashes in useLogsPreview

jordienr:

YES

pages cound be undefined so it can crash the app.

## REPRO CRASH

In prod or staging:
- Go to a logs page
- Open a log
- Copy the event message
- Paste it in the search and press enter
- Click "Open in Logs Explorer" button in top right
- Run query
- Press back button in browser (two times probably)
- Logs Previewer component crashes


alaister (approved):
LGTM!

- Tested on preview

PR #34232, opened 2025-03-17
chore(studio): move Query to pgMeta add tests

avallete:
- Move the Query builder from studio to pgMeta
- Add e2e tests over the generated sql to ensure syntax and runtime result over pg database
- fix bug with "order by" for table with undefined column

The idea is to move as much as possible SQL generation logic to `pg-meta` package, where it can be tested e2e against a postgres database. Ensuring that the generated sql is valid, and yield the expected results for a given database setup. So we can both improve coverage over those parts and reproduce bugs (such as one mentioned in the linked issue) more easily.

This PR focus on only moving the `Query` builder from studio to `pg-meta` package, and add tests for it.
A follow-up PR will focus on moving to pgMeta and fixing the behavior of `table-row-query`, and adding tests for it.

Since this following PR will introduce some behavioral changes, I'll rather have it discussed and tested on it own.

See: https://github.com/supabase/supabase/pull/34237


alaister (approved):
LGTM!

- Tested on preview
  - Tested table row actions
  - Tested pg vault actions
  - Generally clicked around to ensure existing functionality

Thanks for doing this! 🔥💯

PR #34237, opened 2025-03-17
fix(studio): update table-query-rows generated sql for truncation

avallete:
- Move the sql creation logic for the table editor to pg-meta package, add e2e tests to ensure resulting query meet requirements in different scenarios
- Fix the field truncation query bug leading to failure on `postgres-meta` container by specifying every columns instead of `select *, ${truncatedColumns.join(',')}`
- Add truncation to large array by checking the array field global size, and truncating to the X first elements if it's too large
- Use `octet_length` instead of `length` function which help speed up the query on the server side

Closes: https://linear.app/supabase/issue/PGMETA-41/improve-studio-table-rows-query


apps/studio/data/table-rows/table-rows-query.ts
```diff
@@ -264,7 +199,7 @@ export async function getTableRows(
   const table = parseSupaTable(entity)
 
   const sql = wrapWithRoleImpersonation(
-    getTableRowsSql({ table, filters, sorts, limit, page, impersonatedRole }),
+    getTableRowsSql({ table: entity, filters, sorts, limit, page }),
```
avallete, line 202: **note**

Since this is now all in pgMeta, I made it accept "entity like" object coming from it's introspection directly. No need to turn it into a different object.

Removed the "impersonatedRole" logic to keep the separation of concerns consistent. This give you the select query only for the table. You can then wrap this into a impersonated transaction to change what this query will be able to fetch.

packages/pg-meta/test/query/table-row-query.test.ts
```diff
@@ -0,0 +1,1111 @@
...
+
+        expect(testTable).toBeDefined()
+
+        // Generate SQL with lower maxCharacters and maxArraySize limits
+        const sql = getTableRowsSql({
+          table: testTable!,
+          page: 1,
+          limit: 10,
+          maxCharacters: 2048,
+          maxArraySize: 10,
```
avallete, line 359: **note**

Made the maximum values a customizable parameter for testing clarity sake (don't want to check a 10Kb in my tests). But that could also be used if studio want to increase the limit (eg: custom user based limit) in some cases without completely breaking free of all limits.

apps/studio/data/table-rows/table-rows-query.ts
alaister: Correct me if I'm wrong, but the `getTableRowsSql` wasn't actually using the `impersonatedRole` arg right?

alaister (approved):
LGTM!

- Tested on preview

🔥🔥🔥

apps/studio/data/table-rows/table-rows-query.ts
avallete: Nope you're right, I think it was just passed around to satisfies the `GetTableRowsArgs` type

PR #34268, opened 2025-03-18
Update realtime settings based on feedback

joshenlim:
Fixes FE-1475

As per PR title - just updating language + scoping down the input fields

image

## Changes involved:
- Renamed "Private channels only" to "Allow public channels" + negate field value
- Update copy for connection pool size + max concurrent clients
- Added "You need additional permissions" text beside save button if user doesn't have the perms to update realtime settings
- Added validation for connection pool size to be below max database connection size
- Added warning for connection pool size if 80% of max database connection size
image
- Disable max concurrent clients field if subscription doesn't have spend cap disabled
image

## To test
- [ ] Can update realtime settings on a project that doesn't have realtime settings yet (check for 404 on /settings endpoint - this is expected for projects that have yet to save their realtime settings at least once)
- [ ] Can update realtime settings on a project that has realtime settings (basically make more than 1 save changes to the project's realtime settings)

### DIFF
--- a/apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
+++ b/apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
@@ -85,6 +85,7 @@ export const RealtimeSettings = () => {
   })
 
   const onSubmit: SubmitHandler<z.infer<typeof FormSchema>> = (data) => {
+    console.log('FUNCTION ONSUBMIT')
     if (!projectRef) return console.error('Project ref is required')
     updateRealtimeConfig({
       ref: projectRef,
@@ -403,8 +404,10 @@ export const RealtimeSettings = () => {
                   <Button
                     type="primary"
                     htmlType="submit"
+                    form={formId}
                     disabled={!canUpdateConfig || isUpdatingConfig || !form.formState.isDirty}
                     loading={isUpdatingConfig}
+                    onClick={() => console.log('onCLICK SUBMIT')}
                   >
                     Save changes
                   </Button>
--- a/apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
+++ b/apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
@@ -122,7 +122,7 @@ export const RealtimeSettings = () => {
                       <FormSectionContent loading={isLoading} className="!gap-y-2">
                         <FormItemLayout
                           layout="flex"
-                          label="Allow public channels"
+                          label="Allow public access"
                           description="If disabled, only private channels will be allowed"
                         >
                           <FormControl_Shadcn_>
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
```diff
@@ -93,8 +120,8 @@ export const RealtimeSettings = () => {
                       <FormSectionContent loading={isLoading} className="!gap-y-2">
                         <FormItemLayout
                           layout="flex"
-                          label="Private channels only"
-                          description="If this is enabled, only private channels will be allowed"
+                          label="Allow public channels"
+                          description="If this is disabled, only private channels will be allowed"
```
saltcod, line 124: > If this is disabled, only private channels will be allowed

"If disabled, ..."  ?

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
```diff
@@ -120,11 +147,11 @@ export const RealtimeSettings = () => {
                         <FormSectionLabel
                           description={
                             <p className="text-foreground-lighter text-sm !mt-1">
-                              Sets connection pool size for Realtime Authorization
+                              Realtime Authorization uses this database pool to check client access
```
saltcod, line 150: Default here is 2, though I have no idea if that's good or not. Should it be 2? 20?  

Aside: clicked the docs link above and that isn't connected yet (docs prob not written yet?)   — just calling out in case it gets forgotten

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
```diff
@@ -138,6 +165,14 @@ export const RealtimeSettings = () => {
                           />
                         </FormControl_Shadcn_>
                         <FormMessage_Shadcn_ />
+                        {!!maxConn && field.value > maxConn.maxConnections * 0.5 && (
+                          <Admonition
```
saltcod, line 169: Maybe keep this admonition, but put a small note in the ui. That helps tie the two numbers together for users? 

screenshot-2025-03-18-at-21 03 48

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
joshenlim: ^ yup we'll update the docs link separately once its ready 🙏 

default is 2 yup! probably might have more context and details once the docs are out

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
joshenlim: hmm 50% is not a hard limit though - its only a recommendation so reckon we only show it if the users actually input a large value

saltcod (approved):
Checked preview and i can change all the various settings.

alaister (approved):
Double approve!

PR #34280, opened 2025-03-19
fix(studio): delete all branches disabling branching

avallete:
- Filter out the "default" branch that can't ever be deleted.


alaister (approved):
LGTM!

- Tested on staging. I still got a permissions error but I think this is due to my staging project being super old rather than this PR

avallete:
Issue is server side.

PR #34281, opened 2025-03-19
fix: edge function deno types

gregnr:
In the dashboard functions editor, edge-runtime-specific types are not available to the TS language service and results in bad DX:
image

This PR manually adds types. There are 2 types we need:
- `edge-runtime.d.ts`: this is available on jsr.io which supports CORS, so we make a direct fetch request to it at runtime and add to Monaco's typescript language server
- `lib.deno.d.ts`: this is only available on GitHub releases, which do not support CORS. To get these, I added some webpack rules to fetch these at build time and store under nextjs generated assets (`/static/media/...`). Then we can simply fetch them from that path and add to the TS language server in the same way

## Preview
https://studio-staging-git-fix-edge-function-deno-types-supabase.vercel.app/dashboard/project/_/functions/new?template=hello-world

## Todo
- [x] Supabase/edge runtime specific types
- [x] Deno types


staaldraad (approved):

joshenlim (approved):
Verified changes on staging :)

PR #34286, opened 2025-03-19
chore: remove nano_enabled flag usage

kevcodez:
It is true in all cases


alaister (approved):
LGTM!

- Tested on preview

joshenlim (approved):
LGTM - changes don't have visible differences on preview as per PR description, but checked the Downgrade + Exit survey modals nonetheless

staaldraad (approved):

PR #34302, opened 2025-03-20
Download Deno Types locally 

jordienr:
- Adds script to download deno types and save them in /public
- Loads deno types from same host (public folder) instead of remotely at build time with webpack

## To test:
- Check builds dont break
- Go to the function editor
- Deno should have correct types. (hover Deno and see info)
CleanShot 2025-03-20 at 13 25 59@2x


ivasilov (approved):

staaldraad (approved):

apps/studio/next.config.js
```diff
@@ -530,10 +524,6 @@ const nextConfig = {
         source: '/favicon/:slug*',
         headers: [{ key: 'cache-control', value: 'public, max-age=86400' }],
       },
-      {
-        source: '/(.*).ts',
-        headers: [{ key: 'content-type', value: 'text/typescript' }],
-      },
```
gregnr, line 536: You might still want this, Vercel defaults `*.ts` file content type to `video/mp2t`. Shouldn't affect functionality, but it will send with that content type.

PR #34307, opened 2025-03-20
feat: add mfa enforcement toggle

staaldraad:

YES

Feature

## What is the new behavior?

Allows toggling MFA enforcement on an organization level.

Depends on API change #22088

Screenshot 2025-04-30 at 15 20 11
Screenshot 2025-04-30 at 15 18 32
Screenshot 2025-04-30 at 15 19 17


doublethink (approved):

joshenlim:
Pushing up some changes while testing:

- Don't show toggle for org not on paid plan, the callout to upgrade is sufficient i think
image
- Adding inline links to when the toggle is disabled
  - For enabling MFA on own account -> account/security
image
- Added invalidation when adding / removing MFA factors on own account to invalidate organization members based on last visited organization
  - So that the toggle will disable / enable accordingly without having to refresh the page
- Add `toast.error` in `onError` for OrganizationMfaToggleMutation
  - Since we're passing an `onError` the default `onError` handler in the mutation itself will not get triggered then, as the error handling is handled by the provided parameter instead
- (Unrelated) Fix ConfirmationModal ignoring provided loading param, gets stuck in loading state if error happens onConfirm

joshenlim (approved):
Changes LGTM - added feature flag so this isn't user facing just yet
Still need to figure out a better UX on communicating to users if they need to enable MFA to access an organization

PR #34316, opened 2025-03-21
fix duplicate policy name

SaxonF:
Fixes duplicate policy name in policy list


joshenlim (approved):
Checked changes on staging!

PR #34317, opened 2025-03-21
Feat/realtime examples

SaxonF:
https://github.com/user-attachments/assets/fde699e6-a3be-41c6-ab21-18b9ba040056

Adds a realtime examples section to the realtime page. It replaces the old non interactive section. 

For this PR, lets focus on ensuring all examples work and fit nicely in the realtime page. Lets follow up with refinements to the actual realtime example code so that it fits best practices etc.


apps/www/components/Realtime/examples/chat-example.tsx
```diff
@@ -0,0 +1,191 @@
...
+import type { ExampleLayoutProps } from '../example-layout'
+
+const instanceId = Math.random().toString(36).substring(2, 9)
+
+const appJsCode = `import { useEffect, useState, useRef } from 'react';
+import './styles.css';
+import { createClient } from '@supabase/supabase-js';
+
+// Initialize Supabase client
+const supabaseUrl = '${process.env.NEXT_PUBLIC_SUPABASE_REALTIME_URL}';
```
joshenlim, line 13: all these likely needs to be changed to `NEXT_PUBLIC_SUPABASE_URL` and `NEXT_PUBLIC_SUPABASE_ANON_KEY` so that we consolidate all the www stuff into one project

joshenlim (commented):
Checked through all the examples locally - i only couldn't get Todos and Logs working

Tried to debug, but hitting a deadend here as the database broadcasts are not firing despite ensuring there's RLS, trigger and functions set up as per [here](https://supabase.com/docs/guides/realtime/broadcast#broadcast-from-the-database)

[Update] Resolved now

joshenlim (approved):
LGTM from FE end - verified that all the realtime examples are working (in particular the todo + logs one which involve a supabase project)

cc @supabase/marketing as we'll need a +1 too 🙏

craigcannon (approved):

PR #34321, opened 2025-03-21
Authorized apps in org settings add copy action for App ID

joshenlim:
Fixes FE-1478

As per PR title - also swapped out the time strings with TimestampInfo component
image


alaister (approved):
LGTM!

- Tested on preview

PR #34329, opened 2025-03-22
feat: bump auth-js to RC 2.69.0

kangmingtay:
Per new process, a RC for @supabase/auth-js needs to be deployed on supabase.com before it is declared an official release.

Tested in local studio and staging preview env


alaister (approved):
LGTM!

- Tested auth on studio/www/docs previews

PR #34332, opened 2025-03-22
ci: fix autofix

kevcodez:
Actually make the autofixing work - previously the GH action jobs (typecheck, ...) would stall, because the default GH token will never trigger follow-up actions after committing, leaving your PR in a weird unmergeable state that would only be solved by doing a manual commit.

- Using personal access token, otherwise follow-up actions do not trigger (even when requesting GH write permissions)
- Bumped up GH action versions
- Fixed prettier command failing on some files with invalid CSS/JSON
- Ignored generated files for prettier to speed it up

Adding `autofix` label is now enough to run prettier and automatically push up a fix.


.github/workflows/autofix_linters.yml
```diff
@@ -16,32 +13,39 @@ concurrency:
...
 
       - name: Run Prettier in fix mode
         run: pnpm run format
 
       - name: Commit changes and push to existing branch
         run: |
-          git config --global user.name 'github-tidy-bot'
-          git config --global user.email 'github-tidy-bot@supabase.com'
+          git config --global user.name 'kevcodez'
+          git config --global user.email 'k.grueneberg1994@gmail.com'
```
kevcodez, line 48: Needs to be some user that has write access to the repo, otherwise builds are not running one committed - we can use another one too

ivasilov (approved):

fsansalvadore (approved):

PR #34362, opened 2025-03-23
Fix owner_id type

Revadike:
If you don't type cast you'll encounter the following error:
```
Error: ERROR: 42883: operator does not exist: text = uuid HINT: No operator matches the given name and argument types. You might need to add explicit type casts. 
```
This should fix it.


Hallidayo:
Hi @Revadike - We’re doing a bit of housekeeping and plan to close out PRs that have been inactive for more than 3 months.

PR #34369, opened 2025-03-24
Update dropzone component for UI Library

joshenlim:
## Changes involved
- Added JSDocs to each parameter when initializing `useSupabaseUpload` hook
- Added `allowedMimeTypes` to be returned from `useSupabaseUpload` hook
- Added `cacheControl` and `upsert` to be optional params for `useSupabaseUpload` hook
- Updated `maxFiles` param from `useSupabaseHook` to be default 1 instead of 0
- Fixed `maxFileSize` message to be formatted based on bytes (not fixed at MB)
- Added client side validation for `maxFiles`
- Add UI truncate for long file names
- Allow `path` prop to be optional (Support uploading to root of bucket)
- Prevent removal of files on the UI while uploading
- Handle partial upload success (some files uploaded, some files didn't)
  - Updated error/success handling to be on a per file level
  - There's also a `isSuccess` state if all uploads end up successful
- Handle `onDrop` for preventing duplicate file names

## To test (and also tested on my end locally)
- [x] Success state when all files are uploaded successfully
- [x] Client side validations for max file size and max files
- [ ] Partial success state when uploading files but some files errored-out (e.g same name + upsert set to false)
  - I've made this to work such that you can swap out the files while in partial state, and the success state will show up once all files are uploaded successfully.
  - When hitting "Upload" while in a partial state, only files that errored-out, or new files that are added will be uploaded (Uploading is skipped for files that have already been successfully uploaded)
  - Uploaded files cannot be removed, max number of files is also still validated  

### Empty state
empty

### Attached files
upload

### Success state
success

### Partial success state
partial


apps/ui-library/registry/default/components/ui/button.tsx
```diff
@@ -1,15 +1,16 @@
-import * as React from 'react'
 import { Slot } from '@radix-ui/react-slot'
 import { cva, type VariantProps } from 'class-variance-authority'
+import * as React from 'react'
 
 import { cn } from '@/lib/utils'
 
 const buttonVariants = cva(
-  'inline-flex items-center justify-center gap-2 whitespace-nowrap rounded-md text-sm font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50 [&_svg]:pointer-events-none [&_svg]:size-4 [&_svg]:shrink-0',
+  'inline-flex items-center cursor-pointer justify-center gap-2 whitespace-nowrap rounded-md text-sm font-medium transition-colors focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring disabled:pointer-events-none disabled:opacity-50 [&_svg]:pointer-events-none [&_svg]:size-4 [&_svg]:shrink-0',
```
ivasilov, line 8: This component should be kept the same as shadcn button so that users know what are they getting.

ivasilov (approved):

PR #34375, opened 2025-03-24
[FE-1423] chore(studio): allow add-ons for branches

alaister:

chore

Users can't add:
- IPv4 Add on
- PITR
- Custom domains

To their branches.

## What is the new behavior?

They now can.

## To Test:

- [ ] Create a preview branch
- [ ] Add all these add-ons


joshenlim (approved):
Verified enabling + disabling addons on a branch + verify that they apply to the branch only

PR #34409, opened 2025-03-25
Fix Connect UI not showing the right connection string for poolers when a read replica is selected

joshenlim:
As per PR title - we needed to use the `db_host` for the replica in the connection string for the pooler

## To test
- [ ] For a project that has read replicas, verify that the connection string updates accordingly when switching to a read replica


alaister (approved):
LGTM!

- Tested on preview

PR #34421, opened 2025-03-26
Fix account audit logs date range selection

joshenlim:
Realised that we were sending the same start and end time when selecting a single date in account audit logs, hence why returning 0 results. This PR fixes that + use TimestampInfo for account audit logs date column

## To test:
- [ ] Can select a single date (the day before) in the date range selector for account audit logs
- [ ] Can select a single date (the current day) in the date range selector for account audit logs
- [ ] Can select a date range in the date range selector for account audit logs (default)


alaister (approved):
LGTM!

- Tested locally

PR #34444, opened 2025-03-26
Mamba version 2

goszczynskip:


joshenlim (approved):
Checked:
- www /launch-week -> can claim ticket + change to gold from sharing

Issues found (Can be fixed separately since this PR is already really huge):
- www -> is there supposed to be a LW banner? (seems like there should be from the code changes?)
- docs -> promo toast doesn't show up (seems like there should be from the code changes?)

fsansalvadore (approved):

PR #34463, opened 2025-03-27
auth-email-passwordless.mdx: use magiclink auth type in email

jeanlauliac:

YES

Docs update. 

I *think* the suggested template is not completely right, as it passes type `email` which is then used when calling `verifyOtp`, for example when integrating with next.js: https://supabase.com/docs/guides/auth/server-side/nextjs?queryGroups=router&router=app. Type `email` seems to be when login with a username/password rather than a magic link, and I can see the supabase type `EmailOtpType` has a possible value `magiclink`.

Or maybe it doesn't matter so much, the token is enough to recognize the type?

## What is the new behavior?

Passes `magiclink` as type instead.

## Additional context

n/a

### DIFF
--- a/apps/docs/content/guides/auth/auth-email-passwordless.mdx
+++ b/apps/docs/content/guides/auth/auth-email-passwordless.mdx
@@ -135,7 +135,7 @@ If you're using PKCE flow, edit the Magic Link [email template](/docs/guides/aut
 <h2>Magic Link</h2>
 
 <p>Follow this link to login:</p>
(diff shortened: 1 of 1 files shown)


Hallidayo:
Hi @jeanlauliac - We’re doing a bit of housekeeping and plan to close out PRs that have been inactive for more than 3 months.

silentworks:
This change is incorrect as `magiclink` has been deprecated and `email` is what should be used.

image

PR #34483, opened 2025-03-27
mamba fix og

goszczynskip:


fsansalvadore (approved):

thinkverse:
Shouldn't the secret OG be the golden one? When clicking "share on," it says "Found the secret golden ticket.", but the image is this one and not the golden one, the golden one seems to be the platinum which you get for sharing twice?

Screenshot 2025-03-27 at 17 57 14

PR #34491, opened 2025-03-27
Update tokenizer.ts

mst-rajatmishra:

YES

This PR is a refactor that streamlines and optimizes the token counting logic. While it doesn't add any new features or fix bugs, it makes the code easier to maintain and more efficient by simplifying how we handle model versions and token configurations.

Currently, the code handles token counting by looking at the model type (like gpt-3.5-turbo or gpt-4). If the model is one of these, it defaults to a specific version (e.g., gpt-3.5-turbo-0301) and gives a warning that the model version may change over time.

For each message, the code counts the tokens by checking different parts of the message, such as the role, content, and name. It uses a tokenizer to calculate how many tokens are in the message and adds extra tokens for specific elements, like the message’s name, depending on the model.

The code also calculates the maximum number of tokens a model can handle, with a set limit (e.g., 4097 tokens for certain models).

However, there is some redundancy in the logic, particularly when handling different model versions, which could be simplified.

## What is the new behavior?

The new behavior simplifies and improves how the code handles token counting and model versions:

**Easier Model Version Handling:**

Instead of checking and falling back to different versions for models like gpt-3.5-turbo or gpt-4, the model version is now determined up front using a map. This makes the code cleaner and easier to work with.

**Streamlined Token Counting:**

The process for counting tokens in each message is more efficient. It still looks at things like message content, role, and name, but model-specific adjustments are now handled in a simpler way.

**Clearer Error Handling:**

If an unsupported model is used, the code immediately throws an error, making it easier to identify and fix issues.

**Improved Performance and Maintenance:**

By reducing redundancy and simplifying the logic, the code is more efficient and easier to maintain or extend in the future.

### DIFF
--- a/packages/ai-commands/src/tokenizer.ts
+++ b/packages/ai-commands/src/tokenizer.ts
@@ -3,6 +3,15 @@ import type OpenAI from 'openai'
 
 export const tokenizer = getEncoding('cl100k_base')
 
+// Define constants for versions
+const MODEL_VERSIONS = {
+  'gpt-3.5-turbo': 'gpt-3.5-turbo-0301',
+  'gpt-4': 'gpt-4-0314',
+  'gpt-3.5-turbo-0301': 4097,
+  'gpt-4-0314': 4097,
+  'gpt-4o-mini-2024-07-18': 4097,
+}
+
 /**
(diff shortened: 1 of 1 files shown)


Hallidayo:

PR #34492, opened 2025-03-27
feat(docs): fetch conn string

charislam:
Add option to ProjectConfigVariables to fetch the Supavisor session mode connection string, for setting up MCP server

image

## Testing

There's no way to sign in from a docs preview site, so sign in via a studio preview site, then copy and paste your Auth Token from local storage.

### DIFF
--- a/apps/docs/components/ProjectConfigVariables/ProjectConfigVariables.tsx
+++ b/apps/docs/components/ProjectConfigVariables/ProjectConfigVariables.tsx
@@ -33,6 +33,7 @@ import {
 import { useCopy } from '~/hooks/useCopy'
 import { useBranchesQuery } from '~/lib/fetch/branches'
 import { useOrganizationsQuery } from '~/lib/fetch/organizations'
+import { type SupavisorConfigData, useSupavisorConfigQuery } from '~/lib/fetch/pooler'
 import { useProjectApiQuery } from '~/lib/fetch/projectApi'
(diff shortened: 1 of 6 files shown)


gregnr (approved):
Awesome 🤩 

I added [this commit](https://github.com/supabase/supabase/pull/34492/commits/e493afdf0465faef384c4ae97223fb79b47e960e) to remind folks to replace the `[YOUR-PASSWORD]` placeholder (should have had this before too).

One thing I noticed with `<ProjectConfigVariables>` is that if the project is paused, the session pooler URL doesn't show up. I wonder if we should either disable these projects in the dropdown, or (probably better) show a message saying the project is paused and they must restore to see the connection string?

charislam:
> Awesome 🤩
> 
> I added [this commit](https://github.com/supabase/supabase/pull/34492/commits/e493afdf0465faef384c4ae97223fb79b47e960e) to remind folks to replace the `[YOUR-PASSWORD]` placeholder (should have had this before too).
> 
> One thing I noticed with `<ProjectConfigVariables>` is that if the project is paused, the session pooler URL doesn't show up. I wonder if we should either disable these projects in the dropdown, or (probably better) show a message saying the project is paused and they must restore to see the connection string?

oh good catch, Stojan also pointed this out a few weeks back but I forgot about it

charislam:
Fixed the paused behavior 
image

PR #34493, opened 2025-03-27
fix: Add previews for the new components on the main page

ivasilov:


SaxonF (approved):
image


Realtime avatar stack is empty on the page but could be my setup. Lets merge, keep moving and come back to it if its still an issue.

PR #34501, opened 2025-03-28
Add download edge functions CTA to edge function details header

joshenlim:
image

### DIFF
--- a/apps/studio/components/layouts/EdgeFunctionsLayout/EdgeFunctionDetailsLayout.tsx
+++ b/apps/studio/components/layouts/EdgeFunctionsLayout/EdgeFunctionDetailsLayout.tsx
@@ -169,7 +169,7 @@ const EdgeFunctionDetailsLayout = ({
               </PopoverTrigger_Shadcn_>
               <PopoverContent_Shadcn_ align="end" className="p-0">
                 <div className="p-3 flex flex-col gap-y-2">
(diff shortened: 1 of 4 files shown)


apps/studio/components/layouts/EdgeFunctionsLayout/EdgeFunctionDetailsLayout.tsx
```diff
@@ -87,6 +102,26 @@ const EdgeFunctionDetailsLayout = ({
       ]
     : []
 
+  const downloadFunction = async () => {
+    if (filesError) return toast.error('Failed to retrieve edge function files')
+
+    const zipFileWriter = new BlobWriter('application/zip')
+    const zipWriter = new ZipWriter(zipFileWriter, { bufferedWrite: true })
+    functionFiles.forEach((file) => {
+      const fileName = file.name.split('/').pop() ?? file.name
```
laktek, line 111: zip archives support directories right? I think if we include the full path they will be organized into directories.

ie. `const fileName = file.name`

joshenlim: i'll try doing that actually

laktek: i think ok to go ahead with current logic

alaister (approved):
LGTM!

- Tested on preview

PR #34506, opened 2025-03-28
Fix edge functions secrets permissions state

joshenlim:
As per PR title

Case 1: Can read, cannot update
image

Case 2: Cannot read and update
image

Case 3 (Unlikely): Cannot read, can update
image


alaister (approved):
LGTM!

- Tested on preview

PR #34509, opened 2025-03-28
Cache results in assistant panel

joshenlim:
Changes here are mainly to prevent the Assistant from running SQL queries each time that it's opened as it might unexpectedly consume resources (in particular if the query is resource heavy)

## Changes involved
- SQL queries no longer run when the assistant is re-opened
- Results from SQL queries are cached, so that they persist when the assistant is re-opened (also across browser sessions)
- Unrelated bugfix: Ensure that either the Assistant Panel of the Inline SQL editor panel is opened, never both (was running into a bug whereby the panels were overlapping, so opening the Inline SQL editor while the Assistant was open never showed the Inline SQL editor)

## Things to test
- [ ] Ensure that `SELECT` SQL queries only run once when first loaded, and no longer run when the assistant is re-opened within the same browser session (and their results should be cached)
- [ ] Ensure that `SELECT` SQL queries only run once when first loaded, and no longer run when the assistant is re-opened after a browser refresh (and their results should be cached)
- [ ] Ensure that ^ such queries can be re-ran manually by clicking the run icon, and the results should be updated + cached too
- [ ] Ensure that ^ this behaviour is consistent across chats (since we support multiple chats now)
- [ ] Ensure that Inline Editor panel and AI Assistant panel can be swapped between each other properly (open and close)

### DIFF
--- a/apps/studio/components/ui/AIAssistantPanel/AIAssistant.tsx
+++ b/apps/studio/components/ui/AIAssistantPanel/AIAssistant.tsx
@@ -144,6 +144,9 @@ export const AIAssistant = ({ className }: AIAssistantProps) => {
     }
   }, [])
 
+  // TODO(refactor): This useChat hook should be moved down into each chat session.
+  // That way we won't have to disable switching chats while the chat is loading,
+  // and don't run the risk of messages getting mixed up between chats.
   const {
     messages: chatMessages,
     isLoading: isChatLoading,
--- a/apps/studio/components/ui/AIAssistantPanel/AIAssistant.tsx
+++ b/apps/studio/components/ui/AIAssistantPanel/AIAssistant.tsx
@@ -191,7 +191,6 @@ export const AIAssistant = ({ className }: AIAssistantProps) => {
       resultId?: string
       results: any[]
     }) => {
-      console.log('Update message', { id: messageId, resultId, results })
       snap.updateMessage({ id: messageId, resultId, results })
     },
     [snap]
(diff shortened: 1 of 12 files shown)


apps/studio/components/ui/AIAssistantPanel/AIAssistant.tsx
```diff
@@ -168,6 +181,7 @@ export const AIAssistant = ({ className }: AIAssistantProps) => {
             key={message.id}
             message={message}
             isLoading={isChatLoading && message.id === chatMessages[chatMessages.length - 1].id}
+            onResults={(props) => snap.updateMessage({ id: message.id, ...props })}
```
alaister, line 184: Since you're using a memoized component here, this should also be memo'd so it doesn't break the memoization – wrap it in `useCallback` :)

apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
```diff
@@ -211,8 +229,10 @@ export const MarkdownPre = ({ children }: { children: any }) => {
             isChart={isChart}
             isLoading={isLoading}
             isDraggable={isDraggableToReports}
-            runQuery={runQuery}
+            runQuery={!results && runQuery}
+            results={results}
             onRunQuery={onRunQuery}
+            onResults={(results) => onResults({ resultId: snippetProps.id, results })}
```
alaister, line 235: Same with this one!

alaister (approved):
LGTM!

PR #34513, opened 2025-03-28
Prevent disabling PITR when hipaa is enabled

joshenlim:
As per PR title

image

## Things to test:

For a Pro project with no HIPAA add-on
- [ ] Can enable PITR
- [ ] Can change PITR duration
- [ ] Can disable PITR

For a project with HIPAA add-on (probably easiest to test locally with the logic hardcoded (`hasHipaaAddon`)
- [ ] Can enable PITR (If somehow not enabled yet)
- [ ] Can change PITR duration
- [ ] Cannot disable PITR


saltcod (approved):
Tested these conditions, looks good.


For a Pro project with no HIPAA add-on

- Can enable PITR
- Can change PITR duration
- Can disable PITR
 

Can enable PITR (If somehow not enabled yet)
- Can change PITR duration
- Cannot disable PITR

apps/studio/components/interfaces/Settings/Addons/PITRSidePanel.tsx
```diff
@@ -71,7 +71,7 @@ const PITRSidePanel = () => {
   const { data: databases } = useReadReplicasQuery({ projectRef })
   const { data: addons, isLoading } = useProjectAddonsQuery({ projectRef })
   const { data: subscription } = useOrgSubscriptionQuery({ orgSlug: organization?.slug })
-  const hasHipaaAddon = subscriptionHasHipaaAddon(subscription)
+  const hasHipaaAddon = true
```
kevcodez, line 74: This change needs to be reverted

apps/studio/components/interfaces/Settings/Addons/PITRSidePanel.tsx
```diff
@@ -102,10 +102,12 @@ const PITRSidePanel = () => {
 
   const hasReadReplicas = (databases ?? []).length > 1
   const hasChanges = selectedOption !== (subscriptionPitr?.variant.identifier ?? 'pitr_0')
-  const selectedPitr = availableOptions.find((option) => option.identifier === selectedOption)
   const isFreePlan = subscription?.plan?.id === 'free'
+  const selectedPitr = availableOptions.find((option) => option.identifier === selectedOption)
   const blockDowngradeDueToReadReplicas =
     hasChanges && hasReadReplicas && selectedCategory === 'off' && selectedOption === 'pitr_0'
+  const blockDowngradeDueToHipaa =
```
kevcodez, line 109: Customers with HIPAA should only be able to use 28-days PITR - not less, lowering the PITR days should also not be possible

apps/studio/components/interfaces/Settings/Addons/PITRSidePanel.tsx
kevcodez: We should also check for project settings `is_sensitive` flag - if that's false, we can skip all HIPAA check and treat it as regular project

staaldraad (approved):
🔥

apps/studio/components/interfaces/Settings/Addons/PITRSidePanel.tsx
ivasilov: Fixed.

PR #34533, opened 2025-03-29
chore: rework disk validation and cleanup

joshenlim:
- Enforce large instance for io2 disk, not just gp3
- Do not forcefully use minimum IOPS when switching disk type, but stick with the previously configured IOPS
  - Only overwrite if the min IOPS of the disk is greater than what the user had previously configured
- Get rid of computeAndDisk feature flag and delete a bunch of duplicated code
- Get rid of compute instance side panel in favour of compute and disk page
- Fix a bug that would not display the change in disk size pricing if the disk size is unchanged, but disk type is changed 
- Change minimum IOPS for io2 disk to 1,500 to avoid obvious foot guns
- Get rid of duplicate constants in `DiskManagement.constants` to avoid diverging
- Fly support for compute and disk page

image


apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
```diff
@@ -357,7 +341,11 @@ export function DiskManagementForm() {
             type="default"
             visible={isFlyArchitecture}
             title="Disk configuration is not available on Fly Postgres"
-            description="Please contact Fly support if you need to update your disk configuration"
+            description={
+              isBranch
+                ? 'Please recreate your branch to configure disk size and ensure your branch gets launched on the latest branching infrastructure.'
```
MildTomato, line 353:
``` suggestion
                ? 'Delete and recreate your Preview Branch to configure disk size. It was deployed on an older branching infrastructure.'
```

staaldraad (approved):

joshenlim (commented):
Verified Kevin's changes as per PR description on the staging preview 🙂

alaister (approved):
LGTM!

- Approving on behalf of Joshen

PR #34541, opened 2025-03-29
ref: Remove redundant payload for backup download

kamilogorek:
This was always optional and deprecation was never followed-up upon https://github.com/supabase/infrastructure/blob/71c7c6829e90912b3c5f7ba2f4cacb6419cde533/api/src/routes/platform/database/ref/backups.dto.ts#L3-L31

Types are incorrect because Angelico forgot to mark the types are optional 2 years ago :P validation is indeed optional so I ignored TS for now and will follow-up once zod changes for that endpoint lands in https://github.com/supabase/infrastructure/pull/22278


joshenlim (approved):
Verified on preview by downloading a logical backup from the database/backups page on one of my staging projects

PR #34546, opened 2025-03-30
fix(studio): keep json formatting when truncating arrays

avallete:
- Update the table-editor query selection on array of json fields to keep the final truncated result as a json array.

Before:

Screenshot 2025-04-02 at 19 49 28

After:

Screenshot 2025-04-02 at 19 50 27


saltcod (approved):
Tested on preview and looks good with sample data

PR #34553, opened 2025-03-30
Use relative paths in File editor and support existing entrypoint, import map settings

laktek:

YES/NO

This PR updates the Functions editor to use relative file paths (eg: for existing deploys done via CLI). 

Also, it fixes failing to deploy functions with existing entrypoint and import maps.

One thing to note: This PR uses `@std/path` package from Deno JSR registry. Had to add it as a new npm registry + update pnpm lockfile.

### DIFF
--- a/apps/studio/data/edge-functions/edge-functions-deploy-mutation.ts
+++ b/apps/studio/data/edge-functions/edge-functions-deploy-mutation.ts
@@ -65,7 +65,7 @@ export const useEdgeFunctionDeployMutation = ({
         const { projectRef, slug } = variables
         await Promise.all([
           queryClient.invalidateQueries(edgeFunctionsKeys.list(projectRef)),
-          queryClient.invalidateQueries(edgeFunctionsKeys.body(projectRef, slug)),
+          queryClient.invalidateQueries(edgeFunctionsKeys.detail(projectRef, slug)),
         ])
         await onSuccess?.(data, variables, context)
       },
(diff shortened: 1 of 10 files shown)


apps/studio/data/edge-functions/edge-functions-deploy-mutation.ts
```diff
@@ -65,7 +65,7 @@ export const useEdgeFunctionDeployMutation = ({
         const { projectRef, slug } = variables
         await Promise.all([
           queryClient.invalidateQueries(edgeFunctionsKeys.list(projectRef)),
-          queryClient.invalidateQueries(edgeFunctionsKeys.body(projectRef, slug)),
+          queryClient.invalidateQueries(edgeFunctionsKeys.detail(projectRef, slug)),
```
laktek, line 68: Invalidating `detail` query here will ensure `Details` tab is updated with new deploy. Also, looks like invalidating `detail` query will also trigger a new request to body (I'm not sure exactly how that happens though) cc: @SaxonF

joshenlim: ^ at first glance - we probably need to keep the body invalidation so that the code content is updated, will look into why invalidating details also triggers the body to invalidate

joshenlim: we probably don't need the list invalidation actually

joshenlim (approved):
Verified the following
- Updating the function code for a function that was deployed via the API
- Updating the function code for a function that was deployed via the CLI
- Updating the function code doesnt change the file that is currently selected

PR #34611, opened 2025-04-01
ref: Use update-status endpoint for jwt secret queries

kamilogorek:
> [!WARNING]
> Do not merge until https://github.com/supabase/infrastructure/pull/22261 is live in prod.

A follow-up to my API update https://github.com/supabase/infrastructure/pull/22261 which aims to deprecate `/props` endpoints (which are "deprecated" for almost 3y.


alaister (approved):
LGTM!

- Tested on preview by updating a JWT secret

PR #34629, opened 2025-04-01
refactor: use database status from general health check endpoint

kevcodez:
Saves an unnecessary direct call to the database and uses the health-check in the mgmt-api we are already calling anyway


joshenlim (approved):
Verified on staging that service status checks are all good!

PR #34643, opened 2025-04-02
chore: increase stale times for a few common queries

kevcodez:


alaister (approved):
LGTM!

- Tested on preview

PR #34653, opened 2025-04-02
Dont re route after selecting schema in table editor

joshenlim:
As per PR title - we already use nuqs to adjust the URL query state, no need to re-route


alaister (approved):
LGTM!

- Tested changing tabs between schemas on preview

PR #34664, opened 2025-04-02
chore: Bump `vite` and `vitest`

ivasilov:
I've added
```
"vinxi>esbuild": "0.25.2",
"@tanstack/directive-functions-plugin>vite": "6.2.4",
"@tanstack/react-start-plugin>vite": "6.2.4"
```
to the overrides because they're safe (dependencies of code that's not used in production, only for building of the tanstack blocks).

This PR also removes `lodash.template` dependency from the `design-system`.


alaister (approved):
LGTM!

- Opened previews

PR #34671, opened 2025-04-02
fix: Add a catch block around a JSON.parse when rendering AI assistant messages

ivasilov:


joshenlim (approved):
Changes LGTM - although i'm not sure how to test this on the preview though, is there a specific prompt that will cause it to error out here?

PR #34674, opened 2025-04-02
Table editor filter convert empty string value to null if column is not text based type

joshenlim:
As per PR title, addresses a case whereby applying an empty filter results in a type mismatch for a column that's a numerical type for e.g
image

so e.g in the case of an `id` `int8` column, the filter will be `id = null` instead of `id == ''`
image

Also updated error state in table editor to add a hint if a filter was applied
image


saltcod (approved):
Tested on staging that I can filter by an empty string successfully

PR #34678, opened 2025-04-02
docs: add type hints

charislam:
Add type hints and restore filenames for code blocks across docs.


apps/docs/features/docs/__snapshots__/Reference.typeSpec.test.ts.snap
charislam, line 1: can ignore this, just updating snapshot since the spec was updated

apps/docs/content/guides/auth/auth-anonymous.mdx
```diff
@@ -56,6 +56,10 @@ For self-hosting, you can update your project configuration using the files and
 Call the [`signInAnonymously()`](/docs/reference/javascript/auth-signinanonymously) method:
 
 ```js
+const supabase = createClient('https://your-project.supabase.co', 'your-anon-key')
+
+// ---cut---
```
gregnr, line 62: What does this mean?

charislam: It hides everything above the cut (we need these statements in order to get the type hints, but we don't want to actually show them in the UI cause they're just clutter).

From the contributing guide:

image

mansueli (approved):

mansueli:
LGTM

PR #34689, opened 2025-04-03
Fix single date selection in organization audit logs

joshenlim:
Mainly just shifting the fix that I did in Account AuditLogs from this [PR](https://github.com/supabase/supabase/pull/34421) into a utils that is also shared with Organization AuditLogs

## To test
- [ ] Can filter single date on Account Audit Logs
- [ ] Can filter single date on Organization Audit Logs

### DIFF
--- a/apps/studio/components/interfaces/Account/AuditLogs.tsx
+++ b/apps/studio/components/interfaces/Account/AuditLogs.tsx
@@ -15,6 +15,7 @@ import { useProfileAuditLogsQuery } from 'data/profile/profile-audit-logs-query'
 import { useProjectsQuery } from 'data/projects/projects-query'
 import { Alert, Button } from 'ui'
 import { TimestampInfo } from 'ui-patterns'
+import { formatSelectedDateRange } from '../Organization/AuditLogs/AuditLogs.utils'
 
 const AuditLogs = () => {
   const currentTime = dayjs().utc().set('millisecond', 0)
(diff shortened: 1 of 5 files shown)


apps/studio/data/organizations/organization-audit-logs-query.ts
```diff
@@ -68,11 +67,12 @@ export const useOrganizationAuditLogsQuery = <TData = OrganizationAuditLogsData>
   }: UseQueryOptions<OrganizationAuditLogsData, OrganizationAuditLogsError, TData> = {}
 ) => {
   const { slug, iso_timestamp_start, iso_timestamp_end } = vars
-  const date_start = dayjs(iso_timestamp_start).utc().format('YYYY-MM-DD')
-  const date_end = dayjs(iso_timestamp_end).utc().format('YYYY-MM-DD')
```
joshenlim, line 72: this was an unnecessary formatting, and causing some invalidations to not occur
should just use raw strings from iso_timestamp as per account audit logs

apps/studio/components/interfaces/Account/AuditLogs.tsx
```diff
@@ -91,47 +95,13 @@ const AuditLogs = () => {
               triggerButtonTitle=""
               from={dateRange.from}
               to={dateRange.to}
-              minDate={dayjs().subtract(retentionPeriod, 'days').toDate()}
-              maxDate={dayjs().toDate()}
+              minDate={minDate.toDate()}
+              maxDate={maxDate.toDate()}
               onChange={(value) => {
+                console.log('onChange', value)
```
alaister, line 101:
```suggestion
```

alaister (approved):
LGTM!

- Tested on preview by viewing the sent dates in the network tab
- Observed that the time of the start date was the current time minus however many days ago you selected

PR #34691, opened 2025-04-03
Fix error handling for table side panel editor

joshenlim:
A bug was reported whereby when updating a table via the Table Editor through the side panel, if any error occurs when adding new columns, the error doesn't persist and we show a success state instead - which is misleading. We have a handler to show a toast if any error happened as such, but it wasn't getting triggered for errors on column creation:
image

So this PR fixes that - and also opts to show the ^ above warning toast as a separate toast message so users can still see the error toasts that occured:
image

## To reproduce
- Go to the Table Editor, and on any table, hit "Edit" and add a new column:
  - Name: `X`, type: `int8`, default value: `'0'::bigint`
  - The user who submitted this bug report also suggested to support choosing whether the default value is an expression or not, but reckon we look at that separately
- Then hit save, and the toast should show


alaister (approved):
LGTM!

- Tested on preview

saltcod:
aah I see — I can make changes (like renaming the table, AND add a column at the same time) — where one could be successful and one could fail. 

This initially looked like another bug to me. Wonder if the error could be slightly more explicit? 
>... has been updated, but there were some errors. Please heck these errors separately.
? 

screenshot-2025-04-03-at-10 57 55

joshenlim:
^ gotcha yeap! will update the copy

PR #34692, opened 2025-04-03
fix: disallow edit row on protected schemas

alaister:

Bug fix

Screenshot 2025-04-03 at 17 45 43

## What is the new behavior?

Screenshot 2025-04-03 at 17 44 50


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/InputField.tsx
```diff
@@ -223,14 +227,16 @@ const InputField = ({
...
         actions={
-          <Button
-            type="default"
-            htmlType="button"
-            onClick={() => onEditJson({ column: field.name, value: field.value })}
-            icon={<Edit2 />}
-          >
-            Edit JSON
-          </Button>
+          isEditable && (
```
joshenlim, line 230: nit: thinking it'll still be good to allow users to view JSON in the side panel if this might be a quick one

joshenlim (approved):
Verified changes on preview
- Can edit rows in non protected schemas
- Cannot edit rows in protected schemas

PR #34695, opened 2025-04-03
fix: Set the description field in column to empty string when cleared

ivasilov:


saltcod (approved):
Tested on staging. looks good.

PR #34700, opened 2025-04-03
Chore/add copy error button

saltcod:
Add a copy button when SQL Editor shows errors:

screenshot-2025-04-03-at-10 35 01


charislam (approved):
tested both erroring and non-erroring cases in SQL editor

PR #34724, opened 2025-04-04
container queries basics

SaxonF:
image

This introduces the container query Tailwind plugin which adds support for container queries. As a starting point I have made use of container queries in ScaffoldContainer which removes dependency on knowing whether sidebars (like Assistant) are open. I've also adjusted the compute and disk page to be a little more responsive.


joshenlim (approved):
LGTM! Verified changes on preview and resolved conflicts

PR #34725, opened 2025-04-04
handle string error

SaxonF:
New edge function body query uses handleError which doesn't "support" strings and instead logs via Sentry. This adds some extra checking and parses into format handleError supports.


alaister (approved):
LGTM!

- Tested on preview

PR #34729, opened 2025-04-04
QueryBlock whitelist now() and left() functions

joshenlim:
This is for SQL Blocks in Custom Reports

For context - our SQL blocks will only automatically run in custom reports if we decide that the SQL queries are "readonly", and the way we decide that is currently restrictive by default, in particular the functions that we allow in the SQL query (denoted by the `SAFE_FUNCTIONS` variable - these are mostly postgres native functions). We'll slowly expand this list, but opting to be more careful than flexible atm.

This PR just adds 2 functions to the whitelist `now()` and `left()`


alaister (approved):
LGTM!

PR #34733, opened 2025-04-04
Remove useKeyboardShortcuts in Results.tsx

joshenlim:
## Context on the bug
`Results.tsx` was using the `useKeyboardShortcuts` to handle `Cmd + C` to copy cell contents from the `DataGrid`, but this hook overrides the keyboard shortcut globally across the page. Because `copyCellValue` specifically checks for `cellPosition`, this resulted in hitting `Cmd + C` on anywhere else but the `DataGrid` to run into `event.stopPropagation()` and the default browser copying text behaviour is ignored.

## Changes involved
- Use `useCellKeydown` event handler from `DataGrid` thats scoped to just the DataGrid
- Remove `Cmd + C` overrides in Table Editor as well, use `useCellKeydown` instead
- Remove `useKeyboardShortcuts` hook from `deprecated.ts` -> there's a dupe in `grid/components/common/Hooks.tsx`

## To reproduce
- Open the Assistant panel and ask it to generate some query that returns some results (e.g "How many colors are there in the colors table")
- Once the Assistant replies with the SQL query with results, try copying any text on the page that you're on in the dashboard via `Cmd + C`. Notice that you won't be able to paste it subsequently
- This bug applies also to the SQL Editor when you run a select query and the results get returned (since they use the same `Results` component
- This bug applies also to the Table Editor when rendering a table, you won't be able to Cmd + C any text in the page

### DIFF
--- a/apps/studio/components/grid/SupabaseGrid.utils.ts
+++ b/apps/studio/components/grid/SupabaseGrid.utils.ts
@@ -1,14 +1,16 @@
 import AwesomeDebouncePromise from 'awesome-debounce-promise'
 import { compact } from 'lodash'
 import { useEffect } from 'react'
-import { CalculatedColumn } from 'react-data-grid'
+import { CalculatedColumn, CellKeyboardEvent } from 'react-data-grid'
 
 import type { Filter, SavedState } from 'components/grid/types'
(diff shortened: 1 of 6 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #34734, opened 2025-04-04
Update project usage chart date picker

jordienr:
- default to last hour


joshenlim:
nit: language here - looking into it now
image

joshenlim (approved):
Changes LGTM! Verified on preview

PR #34735, opened 2025-04-04
Remove safe parse checking in NewOrgForm updateForm preventing completely removing organization name in input field

joshenlim:
As per PR title - the safe parse checking was preventing clearing the input field of the organization name, but we still have a validation either way when hitting the submit button if the name input field is empty.

Ideal refactor here is use a proper form handler but will do for now


saltcod (approved):
tested on staging, looks good!

PR #34739, opened 2025-04-04
fix: stop needlessly saving snippets

alaister:

Bug fix

When a user switches snippets, we immediately send a PUT request to update it.

## What is the new behavior?

Nothing has been updated, so nothing needs to be saved.


joshenlim (approved):
Can verify on preview that switching snippets no longer sends a PUT request to update the snippet. Also verified that updating the snippet content still triggers a PUT request

PR #34754, opened 2025-04-04
chore: project transfer preview supports messages of level INFO

ecktoteckto:
Besides errors and warnings also messages of level INFO are supported. Introduced for info that project on NANO doesn't automatically get upgraded to MICRO. A warning wouldn't be appropriate here.

warning+info+error
warning+info


kevcodez (approved):
Tested on staging

PR #34777, opened 2025-04-07
move create edge function via assistant

SaxonF:
image

Moves the button to create an edge function via Assistant in to the create dropdown alongside other options


joshenlim (approved):
Verified changes on preview 🙂

PR #34781, opened 2025-04-07
Add meters table and swap listbox to combobox for WrapperTableEditor target selector

joshenlim:
As per PR title - this table's available via the stripe fdw, just wasn't available for selection from the dashboard
https://supabase.com/docs/guides/database/extensions/wrappers/stripe#meters

Also swapped out Listbox to Combobox to support searching since the list for Stripe FDW is quite big

image


alaister (approved):
LGTM!

- Tested on preview

PR #34788, opened 2025-04-07
Set URL provider value to null on auth providers on cancel and save

joshenlim:
As per PR title - addresses an issue whereby if you open a provider in the side panel, then click "Cancel" or "Save", the URL value of `provider` doesnt go away, and you can't open the same provider again without having to click on another provider, then close the panel click on the previous provider 

We have that logic, but its only called in Sheet's `onOpenChange` which i realise only gets triggered when clicking outside of the Sheet 😓


alaister (approved):
LGTM!

- Tested on preview

PR #34789, opened 2025-04-07
Fix enabling captcha link to protection page, and add docs to captcha providers

joshenlim:
As per PR title:

"Enabling captcha" text here shows up when anonymous sign ins are enabled + captcha is not enabled. It's behaviour previously was to scroll to the captcha section back when all the settings were on one single page. But now that captcha settings are on a different page, hence the change to link to auth/protection instead
image

Also added links to set up captcha provider here "How to set up ___"
image


alaister (approved):
LGTM!

- Tested on preview

PR #34796, opened 2025-04-07
Table Editor filters: Add backwards compatible filter parser

jordienr:
# Changes

- Adds new hook for handling table editor filters
- Adds tests for the new hook
- Adds `@testing-library/react-hooks` to make testing hooks easier

# Context

When the table editor loads its only grabbing the first filter.  
Using Nuqs for filters isn't a good idea because it uses a different syntax which would break old URLs.  
The new hook supports old syntax.

# To test

1. Open the table editor
2. Add 2/3 filters
3. Reload page
4. All filters you added are there and table is filtered correctly.


apps/studio/hooks/misc/use-table-editor-filters.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  function getFilters(): string[] {
+    if (isNewSyntax) {
+      return urlParams.getAll('filter')[0].split(',')
+    }
+
+    return urlParams.getAll('filter')
+  }
+
+  const setParams = (
```
ivasilov, line 18: Add `useCallback` here.

apps/studio/hooks/misc/use-table-editor-filters.ts
```diff
@@ -0,0 +1,41 @@
...
+      ...router,
+      query: {
+        ...router.query,
+        filter: newParams.filter,
+        sort: newParams.sort,
+      },
+    })
+  }
+
+  const filters = getFilters()
```
ivasilov, line 33: These should be `useMemo`ed on `router.asPath`.

joshenlim (approved):
Pushed up some changes to address Ivan's feedback, and also renamed `use-table-editor-filters` to `useTableEditorFilters`. Tested locally - all looks good 🙂

PR #34799, opened 2025-04-07
chore: fix quoting for json column export

staaldraad:

YES

Bug fix

JSON and JSONB columns aren't exported in a manner that allows re-import when a single quote is stored in a json string.

## What is the new behavior?

Single quotes don't break generated SQL used for imports.


saltcod (approved):

saltcod:
Tested on staging and exports json with quotes successfully

PR #34815, opened 2025-04-08
Fix version selector if only one result returned, and default to first result if no GA version found

joshenlim:
As per PR title - 2 changes here on the creation project page
- The postgres version dropdown was getting disabled as there was a logic for the `disabled` property  `availableVersion.length <= 1`, seems like a mistake so swapped that out to `.length === 0`
- If there's no versions with `release_channel` as `ga`, we currently don't select anything by default, so updating here to at least select the first result from the available_versions endpoint
- (Unrelated) one oddity i noticed in `PostgresVersionSelector` - `field.value` somehow was not reflecting the updated value when we call `setValue` in the `useEffect`, causing the UI dropdown to seemingly have no default selected, but I checked the form state that the value _did_ get updated, and that `onSubmit` also had the updated value. Swapped it to use the value from `form.watch()` to ensure the UI is correct
  - This behaviour is also happening on prod if you hit refresh while on the new project page

Note that this shouldn't have any user-facing changes, the main bug that this PR is fixing is just the first pointer

### DIFF
--- a/apps/studio/components/interfaces/ProjectCreation/PostgresVersionSelector.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/PostgresVersionSelector.tsx
@@ -70,17 +70,22 @@ export const PostgresVersionSelector = ({
   )
 
   useEffect(() => {
-    const gaVersion = availableVersions.find((x) => x.release_channel === 'ga')
-    const defaultValue = gaVersion ? formatValue(gaVersion) : undefined
-    form.setValue('postgresVersionSelection', defaultValue)
+    if (availableVersions.length > 0) {
+      const gaVersion = availableVersions.find((x) => x.release_channel === 'ga')
+      const defaultValue = gaVersion ? formatValue(gaVersion) : formatValue(availableVersions[0])
+      console.log({ defaultValue })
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/ProjectCreation/PostgresVersionSelector.tsx
```diff
@@ -70,17 +70,22 @@ export const PostgresVersionSelector = ({
   )
 
   useEffect(() => {
-    const gaVersion = availableVersions.find((x) => x.release_channel === 'ga')
-    const defaultValue = gaVersion ? formatValue(gaVersion) : undefined
-    form.setValue('postgresVersionSelection', defaultValue)
+    if (availableVersions.length > 0) {
+      const gaVersion = availableVersions.find((x) => x.release_channel === 'ga')
+      const defaultValue = gaVersion ? formatValue(gaVersion) : formatValue(availableVersions[0])
+      console.log({ defaultValue })
```
alaister, line 76:
```suggestion
```

alaister (approved):
LGTM!

- Tested on preview

PR #34820, opened 2025-04-08
fix: org restriction banner links

alaister:


joshenlim (approved):
Code changes LGTM - can verify that [slug]/settings route doesn't exist

PR #34827, opened 2025-04-08
chore: refactor role impersonation

alaister:
- Refactors the claims to be stored inside the state (generated when the user is selected) instead of generated per request.

To test:
- Ensure all existing impersonation functionality works as before


joshenlim (approved):
I'm not super familiar with the role impersonation implementation, but checked the preview LGTM!

Tested that current functionality didn't change:
- Created a policy on `countries` table: allow select for authenticated for id <= 10
- In SQL Editor / Table Editor, run `select * colors`
  - No impersonation, all rows returned
  - anon, no rows returned
  - authenticated, rows with id <= 10 returned

PR #34828, opened 2025-04-08
set default value if empty array

SaxonF:
Resolves `Multiple TreeView nodes have the same ID (-Infinity). IDs must be unique.` error that a couple of people have run in to. This most likely triggeres when there is an empty file list (somehow).


saltcod (approved):
Tested on staging, looks good.

PR #34846, opened 2025-04-09
DropdownMenu component modal prop default to false, but allow passing prop value

joshenlim:
## Context
The removal of `modal=false` for **all** dropdowns broke some interactions with dropdowns and dialogs/sheets, so while it fixed DropdownMenu for the RowEditor in the Table Editor, it caused problems in other parts of the dashboard

## Changes involved
- Keep allowing `modal` prop to be passed in DropdownMenu, but default the value to `false`
- Reduce the blast radius for the change in this [PR](https://github.com/supabase/supabase/pull/34829)
  - So only usage of DropdownMenu in InputField is updated, everything else across the dashboard remains status quo
- Revert the changes in this [PR](https://github.com/supabase/supabase/pull/34829/files)


alaister (approved):
LGTM!

- Tested on preview

PR #34848, opened 2025-04-09
Move assistant stored state in to IndexedDB

SaxonF:
We currently use local storage to store Assistant history but some users are hitting quota limits which also affects other parts of the dashboard. This PR moves state storage into an IndexedDB store with keys for each project instead.

It also attempts to migrate from old local storage into new IndexedDB if it doesn't yet exist and old local storage keys exist.

**How to test**
1. Open chat in one project, make some requests
2. Create a separate chat in same project, make some requests
3. Switch project and notice new project list
4. Make some requests on new project
5. Switch back to old project to see other projects chats
6. Rename, delete a chat


alaister (approved):
LGTM!

- Tested on preview

PR #34849, opened 2025-04-09
remove flag for just accessing inline editor

SaxonF:
image

Removes the feature preview check for showing inline editor trigger in header. With this merged, anyone will be able to access the inline editor. The feature preview still exists for using the inline editor when creating/modifying policies/functions/triggers.


joshenlim (approved):
Verified changes on preview 🙂

PR #34854, opened 2025-04-09
Show IOPS price change in DiskManagementReviewAndSubmitDialog when switching storage type but no changes to IOPS value

joshenlim:
Fixes FE-1527, as per PR title

When storage type changes, the price of IOPS is also different, hence why the need to show the IOPS price change row even if IOPS value didn't change
image


alaister (approved):
LGTM!

- Tested on preview

PR #34855, opened 2025-04-09
Adjust AI assistant panel styles

joshenlim:
## Changes
- On desktop:
image

- At `md` (1280px): Opting for 75% width (not adjustable), overlay dashboard contents including banners
  - Fixes the odd `top` placement + width of panel changing when the panel starts from empty state to having some content:
image
image

- Any smaller: No change, panel takes up full screen width
image


alaister (approved):
LGTM!

- Tested on preview

PR #34858, opened 2025-04-09
Fix compute size check in PITR side panel

joshenlim:
## Context
- Enabling PITR requires project to be on a minimum Compute Size of Small
- Before the introduction of Nano size, Micro was the smallest, so we used to check for minimum compute size by `subscriptionCompute !== undefined`
- However, currently if the project is on Micro size, `subscriptionCompute` is no longer undefined, and hence the check is now incorrect

## Changes
- Fix check for minimum compute size before enabling PITR by checking for undefined, as well as checking that current compute its not micro
- Prevent downgrading to Micro if PITR is enabled

## To test
- For a project that's never changed its compute size, verify that you cannot enable PITR while you're on a Micro compute
  - e.g New project (Pro -> defaults to Micro, Free -> defaults to Nano)
- Change the project to a small, verify that you _can_ enable PITR
- Then back to micro and verify the same that you cannot enable PITR thereafter

### DIFF
--- a/apps/studio/components/interfaces/DiskManagement/fields/ComputeSizeField.tsx
+++ b/apps/studio/components/interfaces/DiskManagement/fields/ComputeSizeField.tsx
@@ -274,7 +274,7 @@ export function ComputeSizeField({ form, disabled }: ComputeSizeFieldProps) {
                             {lockedMicroDueToPITR && (
                               <TooltipContent side="bottom" className="w-64 text-center">
                                 Project has PITR enabled which requires a minimum of Small compute.
-                                Please <InlineLink href="/">disable PITR</InlineLink> first before
+                                Please <InlineLink href="/project/_/settings/addons?panel=pitr">disable PITR</InlineLink> first before
                                 selecting Micro
                               </TooltipContent>
                             )}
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/DiskManagement/fields/ComputeSizeField.tsx
```diff
@@ -195,79 +192,93 @@ export function ComputeSizeField({ form, disabled }: ComputeSizeFieldProps) {
...
                                   </div>
                                 </div>
                               </div>
-                            </div>
-                          </>
+                            </TooltipTrigger>
+                            {lockedMicroDueToPITR && (
+                              <TooltipContent side="bottom" className="w-64 text-center">
+                                Project has PITR enabled which requires a minimum of Small compute.
+                                Please <InlineLink href="/">disable PITR</InlineLink> first before
```
alaister, line 277:
```suggestion
                                Please <InlineLink href="/dashboard/project/_/settings/addons?panel=pitr">disable PITR</InlineLink> first before
```

apps/studio/components/interfaces/DiskManagement/fields/ComputeSizeField.tsx
```diff
@@ -195,79 +192,96 @@ export function ComputeSizeField({ form, disabled }: ComputeSizeFieldProps) {
...
                                 </div>
                               </div>
-                            </div>
-                          </>
+                            </TooltipTrigger>
+                            {lockedMicroDueToPITR && (
+                              <TooltipContent side="bottom" className="w-64 text-center">
+                                Project has PITR enabled which requires a minimum of Small compute.
+                                Please{' '}
+                                <InlineLink href="/project/_/settings/addons?panel=pitr">
```
alaister, line 278: sorry realised we have the project ref!
```suggestion
                                <InlineLink href={`/project/${project?.ref}/settings/addons?panel=pitr`}>
```

alaister (approved):
LGTM!

- Tested on preview

PR #34862, opened 2025-04-09
Fix filters and sort not added to query in table editor

jordienr:
# To test
- add some filters or sort
- reload page
- filters are there and the results are filtered as expected.


alaister (approved):
LGTM!

- Tested on preview

PR #34865, opened 2025-04-09
fix(studio): Handle query timeout error

avallete:
There is a global timeout of about 50s on /query endpoint This would fail the request independently from the user statement timeout setting add a case to handle this error and add a CTA to suggest the user of direct connection for long running queries

Screenshot 2025-04-09 at 15 27 07


apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityTabResults.tsx
```diff
@@ -81,7 +83,16 @@ const UtilityTabResults = forwardRef<HTMLDivElement, UtilityTabResultsProps>(
                   >
                     increase the statement timeout
                   </a>
-                  .
+                  , if your query take more than 50s to run, please{' '}
+                  <Link
```
joshenlim, line 87: we can use the InlineLink component for this

apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityTabResults.tsx
```diff
@@ -81,7 +83,16 @@ const UtilityTabResults = forwardRef<HTMLDivElement, UtilityTabResultsProps>(
                   >
                     increase the statement timeout
                   </a>
-                  .
+                  , if your query take more than 50s to run, please{' '}
```
joshenlim, line 86:
```suggestion
                  . If your query take more than 50s to run, please{' '}
```

joshenlim: hmm it feels odd to specify the 50s here - for e.g from a user's POV I wouldn't know how long my query will take

avallete: Yeah, I just wanted to make it clear that if the query take more than 50s it'll timeout no matter what.

avallete:
Screenshot 2025-04-17 at 22 10 33

Applied the comments, turned the rest of the links into InlineLinks as well for consistency

joshenlim (approved):
All LGTM! just pushed up a tiny refactor to open the Connect modal directly instead of redirecting to the database settings for the 'connect to your database directly' CTA 🙂

PR #34871, opened 2025-04-09
Fix toggling of roles apply to all projects

saltcod:
[Joshen] Have edited @saltcod's changes to remove `initialConfiguration`, and just lean on `originalConfiguration` instead (no need for additional state)

## Context:
- Only affecting project-scope roles which is behind Enterprise organizations only
- UI allows saving of roles when the role ID is not properly assigned, example for when that happens is when the target role is missing in the confirmation dialog: (Notice From X to Y, Y is missing)
image
- So the changes in this PR is meant to prevent that from happening, along with a couple of other minor improvements

## Changes involved:
_Note the term "set" just means setting the input field, not actually "assigning" the role_
- Main fix: When toggling "Apply roles to all projects":
  - If toggling OFF and user originally has project-scope OR toggling ON and user originally has org-scope -> reset field to original role state
  - If toggling OFF and user originally has org-scope, set roles on all projects to "Developer"
  - If toggling ON and user originally has project-scope, set role on organization to "Developer"
  - Based on this: Should _never_ run into a situation where by the role ID is incorrectly assigned
- Disable save button if no changes are made (clearer UX) 
  - Previously we had this check when clicking "Confirm", which then just closed the panel with no changes 
- Fix hover states for select item not changing bg color + cursor pointer
- Add tooltip to "X" button "Remove access to project"

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
@@ -1,34 +1,36 @@
 import { useQueryClient } from '@tanstack/react-query'
-import { InfoIcon, Check, ExternalLink } from 'lucide-react'
+import { Check, ExternalLink, InfoIcon } from 'lucide-react'
 import Link from 'next/link'
-import { useState, useEffect } from 'react'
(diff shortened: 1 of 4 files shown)


alaister (approved):
LGTM!

- Tested on preview

staaldraad (approved):

saltcod (commented):
Tested locally and looks good!

PR #34884, opened 2025-04-10
Chore/org nav layout fixes 01

joshenlim:
When preview is OFF
- Fixes odd padding in Organization Settings tabs
image

When preview is ON
- Update redirect when deleting an organization to /organizations instead of /projects


apps/studio/components/interface/Organization/exhaustion.constants.ts
joshenlim, line 1: this is an unused file

alaister (approved):
LGTM!

- Tested on preview

PR #34887, opened 2025-04-10
remove restriction banner component

joshenlim:
RestrictionBanner component is now a duplicate of OrganizationResourceBanner, both of which were rendered in AppBannerWrapper

Since `OrganizationResourceBanner` also handles grace period stuff, is safe to delete RestrictionBanner (otherwise both banners will show up at the same time
image


alaister (approved):
LGTM!

PR #34890, opened 2025-04-10
fix: Update the function code template for OpenAI completion API to work with current packages

dshukertjr:

YES

I couldn't get the current function template for `openai-completion` to work and discovered that the openai npm package is quite outdated, and the model `text-davinci-003` is now deprecated. This PR fixes the function to work with the current version.


dshukertjr:
@SaxonF By the way, where was this code taken from? We might want to update that one as well.

laktek (approved):

laktek:
@dshukertjr I think these were AI-generated. I went through them quickly and updated the obvious issues. Can you look at other examples as well?

Ideally, we should update examples in https://github.com/supabase/supabase/tree/master/examples/edge-functions/supabase/functions and load them as templates in future.

SaxonF (approved):

PR #34922, opened 2025-04-11
feat: billing address, name and tax id for org creation/upgrade

SaxonF:
We currently get support requests about adding or amending invoices due to no billing address being set. Changes in this PR:
- Unify tax id and billing address into a single form
- Add ability to set billing name that diverges from org name
- Add ability to set billing address & tax id during org creation (only paid plans)
- Add ability to set billing address & tax id during plan upgrade

New org:

Screenshot 2025-04-21 at 21 28 39

Screenshot 2025-04-21 at 21 29 01

Plan change:

Screenshot 2025-04-21 at 21 29 31

Billing Settings

Screenshot 2025-04-21 at 21 29 45


apps/studio/components/interfaces/Organization/BillingSettings/BillingAddress/BillingAddressDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+    PermissionAction.BILLING_READ,
+    'stripe.customer'
+  )
+
+  const canUpdateBillingAddress = useCheckPermissions(
+    PermissionAction.BILLING_WRITE,
+    'stripe.customer'
+  )
+
+  const {
```
kevcodez, line 31: This wouldn't work for new organization creation, as there is no organization yet - it would be great to have this dialog not do any actual API calls and just be functional (data-in as props, callback for saving)

kevcodez, line 31: This wouldn't work for new organization creation, if we wanna reuse, as there is no organization yet - it would be great to have this dialog not do any actual API calls and just be functional (data-in as props, callback for saving)

apps/studio/components/interfaces/Organization/BillingSettings/BillingAddress/BillingAddressDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+      </div>
+
+      <Dialog
+        open={open}
+        onOpenChange={(value) => {
+          if (!value) onClose()
+          else setOpen(true)
+        }}
+      >
+        <DialogContent size={'large'} className="border-none">
```
kevcodez, line 103: This needs to cover tax id as well

apps/studio/components/interfaces/Organization/BillingSettings/BillingAddress/BillingAddressDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { PermissionAction } from '@supabase/shared-types/out/constants'
+import NoPermission from 'components/ui/NoPermission'
+import ShimmeringLoader from 'components/ui/ShimmeringLoader'
+import AlertError from 'components/ui/AlertError'
+
+interface BillingAddressDialogProps {
+  slug: string
+}
+
+const BillingAddressDialog = ({ slug }: BillingAddressDialogProps) => {
```
kevcodez, line 18: might be better to rename this modal to something like BillingCustomerDataDialog or so, as it will cover address + tax id

apps/studio/components/interfaces/Organization/BillingSettings/BillingAddress/BillingAddressForm.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+const BillingAddressForm = ({
+  address,
+  onClose,
+  insideDialog = false,
+  disabled = false,
+  formId = 'billing-address-form',
+}: BillingAddressFormProps) => {
+  const { slug } = useParams()
+  const { mutate: updateCustomerProfile, isLoading: isUpdating } =
```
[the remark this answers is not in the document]
kevcodez: i.e. on org creation it would not issue any API calls, but just change form state and then that form state would be used when confirming the org creation

staaldraad (approved):

joshenlim:
Loading state here positioning is off - can look into it in a bit
image

joshenlim:
Bug found: 
- Adding address when creating a paid org, once the org is created, checked the billing section of the org, only the name was added (Address line 1, 2 country didn't save)
image

apps/studio/components/interfaces/Organization/BillingSettings/BillingCustomerData/BillingCustomerDataForm.tsx
```diff
@@ -0,0 +1,338 @@
...
+                      }
+                    >
+                      {field.value
+                        ? COUNTRIES.find((country) => country.code === field.value)?.name ||
+                          'Select country'
+                        : 'Select country'}
+                    </Button>
+                  </FormControl>
+                </PopoverTrigger>
+                <PopoverContent className="w-full p-0" align="start">
```
ivasilov, line 169: You can use `sameWidthAsTrigger` prop here to match the trigger width.

joshenlim:
JFYI pushed up 2 changes
- Swapped Tax ID to use Command + Popover instead of Select, so that we get search + solves the hover state issue when Select is used in the Dialog
- Have each section in the org billing page manage its own loading state to mitigate having one loader blocking the page at the beginning (should improve time to FCP)

kevcodez (approved):

PR #34929, opened 2025-04-11
fix: handle pgsodium not being created by default

soedirgo:
pgsodium isn't created by default anymore for new projects


ivasilov (approved):

soedirgo:
Hold off on merging this for now; there's a bigger issue re: Vault & Wrappers that we need to resolve first

soedirgo:
Handled in https://github.com/supabase/supabase/pull/35012

PR #34930, opened 2025-04-11
Support drag drop SQL block from assistant to empty custom report

joshenlim:
As per PR title - previously wasn't able to, could only drag a SQL block from the AI Assistant into a custom report which already had some charts

https://github.com/user-attachments/assets/fb57cb64-c35c-486d-b2f5-b347b2fca216


jordienr (approved):
tested in preview ✅✅✅

PR #34934, opened 2025-04-11
fix deleting a column not removing it from filters/sorts

jordienr:
# To test:
- create column
- sort by that column
- delete column
- it dont break
- try reloading the page
- it dont break again


saltcod (approved):
confirmed working on staging

PR #34943, opened 2025-04-11
chore: add warning if upgrade without projects

kevcodez:
Happens quite regularly. Doesn't have to be super pretty, as it is a rather rare case that this is a conscious upgrade. We get a few support tickets a week with accidental upgrades.

Screenshot 2025-04-12 at 01 17 01

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
@@ -31,6 +31,8 @@ import {
 } from 'ui'
 import { InfoTooltip } from 'ui-patterns/info-tooltip'
 import PaymentMethodSelection from './PaymentMethodSelection'
+import { ProjectInfo } from 'data/projects/projects-query'
+import { Admonition } from 'ui-patterns'
 
 const getRandomTweet = () => {
   const filteredTweets = tweets.filter((it) => it.text.length < 180)
@@ -71,6 +73,7 @@ interface Props {
   subscription: any
   slug?: string
   currentPlanMeta: any
+  projects: ProjectInfo[]
 }
 
 const SubscriptionPlanUpdateDialog = ({
@@ -88,6 +91,7 @@ const SubscriptionPlanUpdateDialog = ({
   subscription,
   slug,
   currentPlanMeta,
+  projects,
 }: Props) => {
   const queryClient = useQueryClient()
   const [selectedPaymentMethod, setSelectedPaymentMethod] = useState<string>()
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
```diff
@@ -568,6 +572,16 @@ const SubscriptionPlanUpdateDialog = ({
                     )}
                 </div>
               )}
+
+              {projects.length === 0 && (
+                <div className="pb-2">
+                  <Admonition title="Empty organization" type="warning">
+                    This organization has no active projects. Did you select the correct
```
ecktoteckto, line 579: Do we wanna show the message in case there's no projects at all or when there's no **active** projects? Right now the list of projects passed into this component also includes inactive projects.

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
```diff
@@ -568,6 +572,16 @@ const SubscriptionPlanUpdateDialog = ({
                     )}
                 </div>
               )}
+
+              {projects.length === 0 && (
+                <div className="pb-2">
```
ecktoteckto, line 577: Screenshot 2025-04-11 at 22 42 27

Maybe we can reduce the blank space underneath the estimation a bit? Seems like this more in space is caused by this change.

kevcodez: I don't think it's caused by this change, but rather because the right-side is so long and it just spaces it out. We're going to add billing address in a separate PR too (collaposed to one/two lines) that would fill the gap a bit

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
kevcodez: Good point, filtering now

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
ecktoteckto: oh yeah, you're right 👍

ecktoteckto (approved):

ecktoteckto:
Checked on staging

doublethink (approved):

PR #34951, opened 2025-04-12
chore: route credit package requests to sales instead of billing

kevcodez:


joshenlim (approved):
Verified that the 2 links updated leads to the support form with the sales category pre-selected

doublethink (approved):

PR #34973, opened 2025-04-13
feat: confirmation before creating paid org

kevcodez:
It is common for customers to not understand the concepts of projects and organizations. Customers think creating a new organization on the paid plan would upgrade their whole account/all projects.

PR adds an extra confirmation step, in case the customer creates a new paid org and they are in other non-empty organizations.

Customer can then directly jump into upgrading an existing organization or create a new one.

image

Screenshot 2025-04-13 at 23 30 16

I have replaced the deprecated components with the Shadcn equivalents, but not refactoring the entire form (it's using some odd self-wired validation). Also shortened some texts to be less overwhelming.

Screenshot 2025-04-13 at 23 33 02

Screenshot 2025-04-13 at 23 33 07

### DIFF
--- a/apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
+++ b/apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
@@ -312,7 +312,7 @@ const NewOrgForm = ({ onPaymentMethodReset }: NewOrgFormProps) => {
                 onValueChange={(value) => updateForm('kind', value)}
               >
                 <SelectTrigger_Shadcn_ id="kind" className="w-full">
-                  <SelectValue_Shadcn_ placeholder="Select a plan" />
+                  <SelectValue_Shadcn_ />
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
```diff
@@ -158,17 +183,26 @@ const NewOrgForm = ({ onPaymentMethodReset }: NewOrgFormProps) => {
...
     }
 
+    if (projects && projects.length > 1 && formState.plan !== 'FREE') {
+      setIsOrgCreationConfirmationModalVisible(true)
+    } else {
+      await handleSubmit()
+    }
+  }
+
+  const handleSubmit = async () => {
```
alaister, line 201: nit: move this above `onSubmitWithOrgCreation`

alaister (approved):
LGTM!

- Tested on preview

One small one:

PR #34986, opened 2025-04-14
Obfuscate org invite name and slug if not logged in

joshenlim:
Opting to only show the invite's organization name and slug only when the user is logged in to mitigate any potential phishing opportunities. Also fixes some padding issues and hides the error alert component if the error is related to the user not being signed in, as there's already a UI to handle that

Before:
image

After:
image

Part of SEC-281


alaister (approved):
LGTM!

- Tested on preview

PR #34990, opened 2025-04-14
Fix mobile nav menu for org nav layout

joshenlim:
Fixes FE-1573

Addresses the following issues that were identified
- The mobile navigation menu was using the new org layout one despite the feature preview being toggled off. This resulted in incorrect links to org settings pages as the org slug in the URL was undefined. Fix is to reinstate the old mobile nav menu when feature preview is off
image

- Fixes some padding issues in a  when nav layout preview is on
  - Before:
image
image
  - After:
image
image


alaister (approved):
LGTM!

- Tested on mobile preview with flag both off and on

PR #34994, opened 2025-04-14
Fix re-rendering of Account AuditLogs

joshenlim:
There's some unnecessary re-renders happening on the account audit logs pages, mainly caused by the layout components - this resulted in many unnecessary API requests getting triggered to fetch the audit logs but ultimately getting cancelled.

The fixes here adjusts 2 layouts - AccountLayout and OrganizationLayout, with the main fix being in the latter by pulling out `<Content>` outside of the `OrganizationLayout`. This brings the re-render down from 10-ish to 2 - the 2 re-renders are unavoidable atm due to the org nav layout feature preview (it'll go away once the org nav layout feature preview becomes the default)

## To reproduce
- Go to Account Audit Logs page, then refresh - you'll notice that the browser fires several endpoints to the audit logs endpoint, which get cancelled, one of them resulting in a 500, and the last one coming back successfully

## To test
- Verify that there's at most max 2 requests only when refreshing the browser while on the account audit logs page

### DIFF
--- a/apps/studio/components/interfaces/Account/AuditLogs.tsx
+++ b/apps/studio/components/interfaces/Account/AuditLogs.tsx
@@ -72,7 +72,7 @@ const AuditLogs = () => {
     }, 5 * 60000)
 
     return () => clearInterval(interval)
-  }, [])
+  }, [dateRange.from, dateRange.to])
 
   return (
     <>
--- a/apps/studio/components/interfaces/Account/AuditLogs.tsx
+++ b/apps/studio/components/interfaces/Account/AuditLogs.tsx
@@ -57,10 +57,6 @@ const AuditLogs = () => {
   const minDate = dayjs().subtract(retentionPeriod, 'days')
   const maxDate = dayjs()
 
-  useEffect(() => {
-    console.log('Check')
-  }, [])
-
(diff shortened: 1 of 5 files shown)


apps/studio/components/interfaces/Account/AuditLogs.tsx
```diff
@@ -74,6 +57,23 @@ const AuditLogs = () => {
...
+    const interval = setInterval(() => {
+      const currentTime = dayjs().utc().set('millisecond', 0)
+      setDateRange({
+        from: currentTime.add(duration).toISOString(),
+        to: currentTime.toISOString(),
+      })
+    }, 5 * 60000)
+
+    return () => clearInterval(interval)
+  }, [])
```
alaister, line 75: `[dateRange.from, dateRange.to]` intentionally removed?

joshenlim: oh nope! removed it while debugging 🤦

joshenlim: reinstated! 🙏  thanks for the catch!

alaister (approved):
LGTM!

- Tested on preview

PR #35000, opened 2025-04-14
fix: ensure MFA SMS pricing is displayed when toggling

kevcodez:
Value can be null or false, not just false


ecktoteckto (approved):

ecktoteckto:
Tested on staging.

PR #35012, opened 2025-04-14
fix: wrappers sql

soedirgo:
Make Wrappers queries handle old and new Vault/Wrappers. On new Vault (0.3.0+) and Wrappers (0.4.6+), pgsodium is not used, so queries for mutations need to be adjusted.

YES

Bug fix

Wrappers queries always assume usage of pgsodium

## What is the new behavior?

- Don't assume usage of pgsodium (i.e. create if it doesn't exist when doing FDW creation) on Wrappers <0.4.6
- Make Wrappers queries conditional on Wrappers version; if >=0.4.6, use Vault interfaces directly and skip key management APIs

## Additional context

https://supabase.slack.com/archives/C08MKCA50HL


apps/studio/data/fdw/fdw-create-mutation.ts
```diff
@@ -49,15 +49,56 @@ export function getCreateFDWSql({
...
+          '0.2.0',
+          '0.3.0',
+          '0.3.1',
+          '0.4.0',
+          '0.4.1',
+          '0.4.2',
+          '0.4.3',
+          '0.4.4',
+          '0.4.5'
+        ) then
```
soedirgo, line 83: not sure if/how to best decouple these if/else thingy but feel free to do so!

imor (approved):

alaister (approved):
LGTM!

- Tested creating a new wrapper on 0.4.5

PR #35024, opened 2025-04-15
Fix/safari browser rendering Popover in grid header

MildTomato:

YES

- adds `portal` to force Popover to render in body element


joshenlim (approved):
Verified on Safari that the updated Popovers render correctly - also double checked on Chrome just to be sure

PR #35027, opened 2025-04-15
Downgrade and fix radix dialog version to 1.0.5, and remove unnecessary modal props from all usages of DropdownMenu

joshenlim:
Radix Dialog versions 1.1.6 have been causing quite a number of [issues](https://github.com/radix-ui/primitives/issues?q=is%3Aissue%20state%3Aopen%20dialog), notably the ones that have been affecting the dashboard were
- DropdownMenus immediately closing when rendered + opened within a Sheet
- Inputs not being editable when rendered in a Dialog based component

Am fixing the version to 1.0.5 which was what we were on before, and that seemed to have fixed all the issues we've been running into. Also reverted all the injection of `modal` params we've been putting in to fix the above issues, and also removed some unnecessary `modal=false` params as well, have verified locally that all the updated files are working correctly 

Also fixes FilterPopover and SortPopover when they're rendered in ForeignRowSelector being unclickable due to portal set to true - the fix [here](https://github.com/supabase/supabase/pull/35024) addressed the Popovers rendering incorrectly in Safari in the TableEditor, but broke the Popovers in the ForeignRowSelector - so making the `portal` param passable as a prop instead. Also broke the RoleImpersonationPopover used in EdgeFunctionTesterSheet

## To test
- In the Table Editor, verify that you can open these sort of dropdown menus in the RowEditor
image
- For a table that has a foreign key, verify that you can hit "Select a record" and then filter + sort in the ForeignRowSelector
- In an Edge Function, verify that you can select a role when testing an edge function
- In realtime inspector, verify that you can select a role
- Verify all the above in Safari

### DIFF
--- a/apps/studio/components/grid/components/header/Header.tsx
+++ b/apps/studio/components/grid/components/header/Header.tsx
@@ -161,8 +161,8 @@ const DefaultHeader = () => {
   return (
     <div className="flex items-center gap-4">
       <div className="flex items-center gap-2">
-        <FilterPopover portal filters={filters} onApplyFilters={onApplyFilters} />
-        <SortPopover portal sorts={sorts} onApplySorts={onApplySorts} />
+        <FilterPopover filters={filters} onApplyFilters={onApplyFilters} />
+        <SortPopover sorts={sorts} onApplySorts={onApplySorts} />
       </div>
       {canAddNew && (
         <>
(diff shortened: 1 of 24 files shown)


alaister (approved):
LGTM!

- Tested as instructed on preview (chrome + safari)

PR #35031, opened 2025-04-15
Table Editor: Shift loading, error and empty states outside of DataGrid

joshenlim:
The `noRowsFallback` prop from the DataGrid confines the rendering of those UI to be with the columns of the DataGrid, which means that the UI will shift if the DataGrid is scrolling (e.g in the case of many columns). Shifting the loading, empty and error states outside of the `noRowsFallback` prop will allow these UIs to stay in place even while scrolling the DataGrid.

For context - this caused some confusion when the user, after scrolling to the right end of the table, applied an incorrect filter which returned an error, but the error UI only showed up after the user scrolled back to the left end of the table

## Before

https://github.com/user-attachments/assets/a912a36f-8cb6-427a-964f-9f85295425bc

## After

https://github.com/user-attachments/assets/6eee07e2-6296-46bd-a512-116fd6281dc1

## To Test
- Check in the Table Editor, for a table that has many columns and is horizontally scrollable in the DataGrid, that the following render alright: 
 - Loading state (when opening a table)
 - Empty state (No rows, no filters applied)
 - Error state (Apply an incorrect filter to simulate)


alaister (approved):
LGTM!

- Tested on preview

PR #35032, opened 2025-04-15
Add error handling UI for QueryBlock

joshenlim:
QueryBlock component was lacking an error handling UI, so if the query that it ran fails, it fails silently - this PR resolves that

In Custom Report:
image

In AI Assistant:
JFYI we won't save the error into indexDB (unlike the query results) as I don't think there's a need to (but please let me know if I might be overlooking something)
image


alaister (approved):
LGTM!

- Tested on preview

PR #35034, opened 2025-04-15
Fix back to dashboard link on account page

joshenlim:
Fixes FE-1586

For new layout: Adjusts URL for "Back to dashboard" CTA link on the account pages
- Currently when landing on an account page directly from a fresh browser session, the "Back to dashboard" link just redirects back to the same page that you're on (since there's no last visited route available)
- Am opting to set the link as
  - If there's a last visited route, then direct back to there
  - Otherwise, if there's a last visited organization in local storage, then direct back to the home page of the organization which lists all the org's projects
  - Otherwise direct back to /organizations

## To test
- Land on /account/me directly (Refresh)
- Verify that the "Back to dashboard" link doesn't direct back to the same page that the user is on


saltcod (approved):
Tested per steps above, works as expected!

PR #35035, opened 2025-04-15
Add org plan and num projects to organization action card

joshenlim:
Fixes FE-1585

Just adds a bit more details to the organization action card, specifically the org's plan and number of projects
image

Also added in a small refactor for the feature preview modal by removing an unnecessary state


saltcod (approved):

saltcod:
Looks good on staging!

Issue #18815, opened 2023-11-08
Supabase on Oracle ARM64


sneakyfluffy:
# Bug report

- [*] I confirm this is a bug with Supabase, not with my own application. 
- [*] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When I want to self host Supabase with docker on Oracle VM ARM64 I get an error on supabase-vector container. 

Logs of supabase-vector:
<jemalloc>: Unsupported system page size
memory allocation of 5 bytes failed

After that nothing happens and starting of other containers is interrupted. 
## To Reproduce

1. Oracle OCI instance with ARM64 OS. (Rocky linux in my case)
2. Install docker
3. git clone supabase form github
4. copy .env file
5. docker compose up -d

## Expected behavior

Supabase instance should start in docker

## Screenshots

![image](https://github.com/supabase/supabase/assets/18366933/8d2e5055-a976-42a2-bc1e-74bd336b506f)

![image](https://github.com/supabase/supabase/assets/18366933/03cfd447-7803-41e1-8d56-ede560e36d74)

## System information

- OS: Rocky Linux 8.8 ARM64
- Oracle Cloud free tier

kunasun:
Also having the exact same issue using Oracle Linux 8.8 ARM64, same error log message.

mattyx97:
> Also having the exact same issue using Oracle Linux 8.8 ARM64, same error log message.
> 
> EDIT: I don't need the vector capabilities of Postgres for my current project so I removed the vector image from docker-compose and the vector attribute in the depends-on field under db in the docker-compose.yml file. This allows me to docker compose supabase without supabase-vector running as a stopgap fix for now

Please can you send me the fix?

nicufarmache:
I'm having the same issue on Raspberry Pi OS 64bit

ivasilov:
@Jayllyz @nicufarmache @mattyx97 This is probably due to `timberio/vector` not supporting your CPU architecture. 

To disable the vector container, just remove or comment the following lines https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml#L348-L350 and https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml#L384-L405.

nns52k:
> I'm having the same issue on Raspberry Pi OS 64bit

same here

rjbrown99:
[...]

PR #35042, opened 2025-04-15
Fix: supabase vector on system using larger pages (16 KB, 64 KB)

Avie238:

YES

Bug fix

The current timberio vector container can't run when the page size is set to something different then 4 KB
This was mentioned in this [issue](https://github.com/supabase/supabase/issues/18815) however the fix - change the page size to 4 KB - does not work on systems where this is not possible (e.g. Asahi Linux)

## What is the new behavior?

Update to a newer version (0.28.1 -> 0.33.0) to include [this](https://github.com/vectordotdev/vector/pull/18481) change which compiles the container using a larger page size.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -443,7 +443,7 @@ services:
 
   vector:
     container_name: supabase-vector
-    image: timberio/vector:0.28.1-alpine
+    image: timberio/vector:0.33.0-alpine
     restart: unless-stopped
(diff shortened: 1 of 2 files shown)


Siilwyn:
Please merge this :pray: @joshenlim

aantti:
Updated via https://github.com/supabase/supabase/pull/42525

Cc @singh-inder

PR #35046, opened 2025-04-15
Prevent setting Authorization header to empty string by default in AI Assistant component

idpbond:
Currently on self-hosted supabase, which uses HTTP Basic Auth on the dashboard/API by default, the AI Assistant component of the dashboard will set the `Authorization` header to be an empty string even when `headerData` is empty. 

This can override the existing basic auth credentials and cause auth errors on the API.

The change fixes this by first checking whether `headerData` contains an `Authorization` value and only appending it to the payload when it does, allowing `Authorization` to fall back to the browser sessions' existing basic auth credentials.


Hallidayo:

PR #35053, opened 2025-04-16
feat: slack fdw

alaister:

feature

## To test

Verify that all options on https://fdw.dev/catalog/slack/ are available in the dashboard and create a test wrapper


joshenlim (commented):
Verified tables against the docs + Verified creating a wrapper in the staging preview

Just needs an alert as the minimum wrappers ext version to support this FDW seems to be 0.4.5 (had a project that wrappers was at 0.4.0, creating the wrapper failed but there was no alert)

joshenlim (approved):

PR #35060, opened 2025-04-16
Make feature preview available + Fix up Connect UI for local self-host

joshenlim:
### Context
Feature previews have always been available only for the hosted version of the dashboard, but there's not really much reason why it shouldn't be available for local/self-host. (e.g We've had a couple of users reach out on how they can enable previews like the Table Editor tabs which should be perfectly fine to use)

### Changes involved
- Renamed `useNewLayout` to `useIsNewLayoutEnabled` for consistency
  - It's why there's so many file changes sorry, but its just renaming the variable and its usage
- Updated the "Theme" dropdown to be a user icon (similar to prod), and include feature previews + command menu
image
- Some feature previews are hosted specific, so local / self-host won't have access to them (in this case specifically just the nav layout update + SQL Editor tabs)
image
- (Unrelated) Remove unnecessary prop `showProductMenu` from `DefaultLayout` (Was not being used)

### To test
- [ ] Verify that you can toggle feature previews in local / self-host dashboard
- [ ] Verify that no changes on hosted dashboard

### Context
The Connect UI shows up for local/self-host dashboard, but clicking on it shows some loaders that stay indefinitely - it also makes some API calls to fetch the pooling configurations but these endpoints do not exist for local/self-host and hence are unnecessary

### Changes involved
- Only direct connection will be visible in the Connect UI for local/self-host
image
- All the other tabs should have the relevant values shown (ORMs, App Frameworks, etc)

### To test
- [ ] Verify that what's shown in the Connect UI + the values in it looks correct for local / self-hosted dashboard
- [ ] Verify that the Connect UI still looks okay for hosted dashboard


alaister (approved):
LGTM!

- Tested on hosted and self-hosted previews

PR #35065, opened 2025-04-16
FIx layout in support form

joshenlim:
Fixes FE-1593 - Only applies to new layout feature preview

## Context
Support form with the new layout feature preview had the side bar and project / org picker which shouldn't be there. This is mainly because the logic to show the pickers + side bars were (largely) checking for `ref` in the URL. However, we support `/support/new?ref=xxx` to allow users to prefill the project selection when rendering the support form, but it sort of muddles the logic with the LayoutHeader and Sidebar then.

## Changes involved
- Swapped the query param to prefill the selected project from `ref` to `projectRef` to mitigate this overlap
- Updated `useHideSidebar` to account for the support form route

## Before:
image

## After:
image


alaister (approved):
LGTM!

- Tested on preview

PR #35075, opened 2025-04-16
fix: handle new Wrappers in create server sql

soedirgo:
FLUP to https://github.com/supabase/supabase/pull/35012


joshenlim (approved):
Tested on a new project in the staging preview (which is on ver xxx.073)
- Able to create, update, delete, rename wrapper
- JFYI that i'm running into issues with updating a wrapper by adding foreign tables on ver xxx.0.7.0 (although it does seem unrelated to this PR)

PR #35086, opened 2025-04-17
Fix undo behaviour in logs explorer

joshenlim:
## Context
Realized that undoing changes via `cmd+z` doesn't seem to work in the logs explorer when doing either of the following actions:
- Inserting a source
- Selecting a template
- Running the query

This is because we've been manually setting the `editorId` and `editorValue` state which resets the history of the monaco editor. Monaco has its own internal state when making edits in the code editor, which is why we've had these 2 states to manually update the code editor's value programatically - `editorValue` to store the content, and `editorId` to completely refresh the monaco editor entirely. (Setting `editorValue` along will not update the value in the code editor due to Monaco's internal state)

To support a proper editing history, we'd need to use the editor's `executeEdit` methods when making changes programatically

## Changes involved
- Update `handleInsertSource` to insert the selected source with `executeEdit` instead of setting `editorValue` and resetting `editorId`
- Update `onSelectTemplate` to insert the template with `executeEdit` instead of setting `editorValue` and resetting `editorId`
- Update the `useEffect` on `q` to just call `setEditorValue` instead of `onSelectTemplate`
- (Unrelated) Removed `changeQuery` and `router.push` call in `onSelectTemplate` for consistency in behaviour
  - I don't think we should be automatically running the template when it's selected (like how the SQL Editor behaves, the SQL snippet isn't immediately run when selected from templates or quickstarts)
  - I don't think we should be changing the URL route when selecting a template - we can just update the route when after the query is run (we don't update the URL when we "insert source" either)
- The changes also actually fixes FE-1565

## To test
- In the logs explorer, insert multiple sources, then hit `cmd+z` to undo - ensure that each source added is undone individually
- Change the template, then hit `cmd+z` to undo to the previous state
- Change the template, run the query, then hit `cmd+z` to undo to the previous state
- Make a change in the query manually by typing in the code editor, then hit `cmd+z` to undo to the previous state


alaister (approved):
LGTM!

- Tested on preview

PR #35088, opened 2025-04-17
Clear last visited organization when deleteing or leaving org

joshenlim:
Fixes FE-1587, as per PR title (For new nav layout)

## Changes involved
- Clear `LAST_VISITED_ORGANIZATION` local storage value when deleting or leaving an organization
- Unify behaviour for deleting and leaving organization (Re-route logic + clearing local storage value)
- Refetch organizations when leaving organization
- Shift `DEFAULT_HOME` out from lib/constants to `RouteValidationWrapper` (since it's only used there) and make it dynamic based on whether the new layout preview is enabled or not (e.g if not enabled, default home should be `/projects` not `/organizations`)

## To test
- Just need to verify that when deleting / leaving an organization, user is redirected back to /organizations correctly, no unexpected error toasts
- Double check for when the feature preview is off too, that the user is redirected back to /projects in this case, no unexpected error toasts


alaister (approved):
LGTM!

- Tested on preview

PR #35092, opened 2025-04-17
Fix SQL editor styling when multiple snippets are selected

joshenlim:
Fixes FE-1596, Bug was probably introduced with the tabs changes

Before:
image

After:
image

## To test
- [ ] SQL Editor -> select a query, then shift+click on another query to select all the queries between them, ensure that there's a UI indication to show which queries are selected


alaister (approved):
LGTM!

- Tested on preview

PR #35094, opened 2025-04-17
Support persisting scroll position for SQL editor tabs

joshenlim:
Fixes FE-1590

While the non-tabbed SQL Editor UI didn't support persisting scroll position either, the UX feels a bit more un-intuitive with the new tabs interface if we're switching between tabs and the scroll position isn't persisted. Changes here implements that behaviour (so it's even more similar to VS code now):

https://github.com/user-attachments/assets/bd40d172-84b9-4241-a2aa-8577ab47a589

## To test: (Ensure SQL Editor tabs feature preview is toggled on)
- Open a SQL snippet that can be scrolled (e.g the Colors quickstart), and scroll down somewhere
- Open another SQL snippet that can be scrolled (e.g the Slack clone template) and scroll down somewhere
- Switch back to the first snippet, verify that the scroll position is persisted
- Switch back to the other snippet, verify that the scroll position is persisted
- Refresh the page, verify that the scroll position is persisted


alaister (approved):
LGTM!

- Tested on preview

PR #35097, opened 2025-04-17
Tiny fix for copy client id

joshenlim:
This is for Organization Settings when creating a new OAuth app - the copy button for client ID was accidentally copying the secret instead


saltcod (approved):

saltcod:
looks good on staging!

PR #35112, opened 2025-04-18
Remove isValidating disabled check for CustomDomainVerify cancel action

joshenlim:
For this part of the Custom Domains UI
image

Realised that we shouldn't need to disable the "Cancel" button while the user is re-verifying the custom domain at this step, as they should be allowed to cancel the custom domain set up any time


alaister (approved):
LGTM!

PR #35114, opened 2025-04-18
Increase padding for org and project dropdown in new layout

joshenlim:
Fixes FE-1584

Addressing some internal feedback that it's confusing + hard to hit the dropdown button in the layout to switch organization / projects

It was a deliberate decision to have it such that clicking the org name will redirect users to /org/slug route, and clicking the project name will redirect users to /project/ref route now instead of opening the dropdown - so only the chevron button will open the dropdown instead. But agree that the hit box for that dropdown trigger is quite tiny - so the changes in this PR just expands it a little

## Before
image

image

## After
image

image


alaister (approved):
LGTM!

- Tested on preview

PR #35119, opened 2025-04-18
Fix CodeBlock component allowing users to backspace content when it should be readonly

joshenlim:
## Content

We had a bug report that users could actually backspace in `CodeBlock` components, although not being able to input any characters. Our `CodeBlock` component has a prop `contentEditable` that we're flagging by default as `true`, mainly to support users being able to focus on the code content, and conveniently `cmd+a` to copy the content.

## Changes involved

`onBeforeInput` prop blocks against entering additional characters but we needed an `onKeyDown` check as well to check for the backspace event

## To test

- [ ] Staging: Go into any project, open the Connect modal, notice you can backspace in the direct connection input field
image
- [ ] Preview: Do the same thing, but verify that you can no longer hit backspace in the field


alaister (approved):
LGTM!

- Tested on preview

PR #35140, opened 2025-04-19
fix: convert booleans to strings and remove invalid root name key in docker-compose

arakashnawin:
### Summary

This PR fixes the following issues in the `docker-compose.yml`:
- Replaces raw boolean values (`true`) with strings (`"true"`) which are the correct type for environment variables in Compose.
- Removes the invalid `name:` key at the root level which causes `docker-compose` to throw a parsing error.

### Why?

Trying to run `docker compose up` on a fresh clone fails due to:
- Invalid Compose file structure
- Improper data types that are not compatible with environment variables

### What?

- Converted boolean environment values to strings
- Removed root-level `name:` key that does not follow Compose syntax

### Additional Notes

Tested with Docker Compose v2.22 and confirmed to boot successfully.

Happy to make further tweaks if required!

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -5,7 +5,6 @@
 #   Destroy:            docker compose -f docker-compose.yml -f ./dev/docker-compose.dev.yml down -v --remove-orphans
 #   Reset everything:  ./reset.sh
 
-name: supabase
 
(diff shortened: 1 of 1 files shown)


Hallidayo:

Issue #31014, opened 2024-12-09
Add server-side rendering example with SolidStart


Eliot00:
# Improve documentation

## Link 

[Server-side doc](https://supabase.com/docs/guides/auth/server-side)

## Describe the problem

Now we have server-side document for nextjs and svelte, but missing SolidStart ssr guide.

charislam:
Definitely! Opening this to the community for anyone who wants to submit a PR!

demoncoder-crypto:
#35160 I have tried an implementation here. Let me know if i am in the right track, i will iteratively fix any issues if my logic is correct. thanks.

abdulbaqui17:
Opened PR #43943 for this.

What was added:
- SolidStart env vars in the server-side auth guide.
- SolidStart server API route example using createServerClient and cookie adapters.
- SolidStart browser component example using createBrowserClient.

PR: https://github.com/supabase/supabase/pull/43943

LeonxLJX:
I'd like to take this one.

Reading the issue, this looks like a documentation-clarity gap rather than a code defect: the current page doesn't give readers enough to self-diagnose, which is why it surfaces here as an issue instead of being answered by the docs.

Plan: locate the relevant section, confirm what the code actually does today (so the docs match behavior, not intent), tighten the wording, and scan neighboring pages for the same inconsistency — otherwise the fix just relocates the confusion.

One thing to confirm before writing: minimal correction scoped to this page, or also cross-link from related guides if the same gap shows up there?

(claiming via @LeonxLJX)

PR #35160, opened 2025-04-20
Feat/docs solidstart ssr

demoncoder-crypto:
#31014 An attempt to solve this


Hallidayo:
Hi @demoncoder-crypto - Thanks for this one, Their seems to be merge conflicts which need to be looked at.

PR #35167, opened 2025-04-21
API Docs side panel preview, only show keys button in introduction section

joshenlim:
RE the API docs side panel feature preview - realised that the "Show/Hide keys" button is only relevant in the introduction page when you're setting up your Supabase client, so opting to only show that button while on that page to reduce confusion (since toggling that button on any other page has no visual change)

image


alaister (approved):
LGTM!

- Tested on preview

PR #35168, opened 2025-04-21
fix: remove refs to pgsodium keys from vault & wrappers

soedirgo:

YES

Bug fix

Vault & Wrappers use references to pgsodium keys and `key_id` in `vault.secrets`

## What is the new behavior?

Remove these references (some are still left in in Wrappers for back compat; these have and if/else block around them to differentiate the behavior from new Vault)

Tests done on old Vault (15.8.1.054) and new Vault (15.8.1.073):
- navigate to the Vault page
- create a secret (use dummy values for name, desc, and value)
- edit the secret (edit the value)
- delete the secret


jordienr (approved):
lgtm, no issues found in preview

PR #35170, opened 2025-04-21
Fix product menu for mobile not showing in LayoutHeader

joshenlim:
Fixes an issue where the product menu here isn't showing on mobile. Should only show up if in a project route, and not on the project overview
image


alaister (approved):
LGTM!

- Tested on preview

PR #35171, opened 2025-04-21
Prevent sidebar jitter for expand on hover

joshenlim:
Fixes FE-1604

If the sidebar behaviour is expand on hover, fix the padding of the side bar buttons so that there's no shift in layout (currently the padding x increases from 1.5 to 2)

However, if the sidebar behaviour is to stay expanded, then padding x can be fixed at px-2


alaister (approved):
LGTM!

- Ahh so much better!

PR #35175, opened 2025-04-21
Chore/prepare org nav rollout

joshenlim:
Fixes FE-1605
Fixes FE-1606

## Context

We're preparing to do an incremental rollout for the dashboard organization layout changes after a week of internal testing and feedback iteration:
- Initial PR: https://github.com/supabase/supabase/pull/33150
- GH Discussion: https://github.com/orgs/supabase/discussions/33670

The layout changes will still be a feature preview, but we'll be opting users into the preview by default based on a percentage-based random sampling. We'll observe any error rates or feedback as we slowly increase the percentage value till all our users are using the new layout. 🙂🙏 

> [!NOTE]  
> While the behaviour is opt in by default, if any users might want to opt out of the changes, this is still in the user's control by disabling the preview within the Feature Previews section

## Changes involved
- Make org layout feature preview available for all users
  - Users can still opt in if they want to, although I reckon chances are no one really checks out the feature previews normally
- Add logic to support default opt in feature preview
  - If the user doesn't have the corresponding key in local storage, will default to the boolean value set by our feature flag
- Add a dismissible notice banner to inform users of the layout change to mitigate any potential surprise
  - Mainly on the /org/[slug] page and /organizations page
image
image
- Fix feedback link for nav layout feature preview to the correct GH discussion for users to leave feedback on
- (Unrelated) Clean up some `LOCAL_STORAGE_KEYS` that are no longer in use

## To test
- Ensure that your local storage doesn't have the new layout preview key (you shouldn't have it if testing on the staging preview)
- Since staging is fully opted in, you should be redirected to /organizations after signing in
- [ ] Ensure that you can see the notice banner in /organizations
- [ ] Ensure that you can see the notice banner in /org/[slug] (By clicking into an organization)
- [ ] Ensure that the layout feature preview is enabled in the Feature Preview modal
- [ ] Ensure that you can dismiss the notice banner (and it should no longer show up across browser sessions)
- [ ] (Reset your local storage new layout key here) Ensure that if you disable, then enable the nav layout feature preview and refresh, you shouldn't see the notice banner again (since it's no longer opt in "by default")
- Apart from these ^ not sure if there might be any other test cases here that I might be overlooking

### DIFF
--- a/apps/studio/components/interfaces/App/FeaturePreview/FeaturePreviewContext.tsx
+++ b/apps/studio/components/interfaces/App/FeaturePreview/FeaturePreviewContext.tsx
@@ -17,7 +17,7 @@ export const FEATURE_PREVIEWS = [
     key: LOCAL_STORAGE_KEYS.UI_NEW_LAYOUT_PREVIEW,
     name: 'Layout Update for Organizations',
     content: <LayoutUpdatePreview />,
-    discussionsUrl: 'https://github.com/orgs/supabase/discussions/18038',
+    discussionsUrl: 'https://github.com/orgs/supabase/discussions/33670',
     isNew: false,
     isPlatformOnly: true,
   },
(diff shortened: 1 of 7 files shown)


alaister (approved):
LGTM!

PR #35177, opened 2025-04-21
feat (docs): on-demand revalidation for graphql and wrappers

charislam:
Before:

Revalidation was turned off entirely for wrappers pages, and done daily for GraphQL pages.

After:

Revalidation is via tag. There is an /api/revalidate endpoint that is used by GitHub Actions on the federated repos to trigger the revalidation. This decreases the number of unnecessary page generations by Vercel, and gives the maintainers of federated repos more fine-grained control over when to redeploy their docs (see the docs deployment playbook for more info).

### DIFF
--- a/apps/docs/app/api/revalidate/route.ts
+++ b/apps/docs/app/api/revalidate/route.ts
@@ -1,10 +1,10 @@
 import { createClient } from '@supabase/supabase-js'
+import { type Database } from 'common'
 import { revalidateTag } from 'next/cache'
 import { headers } from 'next/headers'
 import { type NextRequest } from 'next/server'
 import { z } from 'zod'
-
-import { type Database } from 'common'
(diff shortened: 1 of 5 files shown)


apps/docs/app/api/revalidate/route.ts
```diff
@@ -35,14 +35,12 @@ export async function _handleRevalidateRequest(request: NextRequest) {
   }
 
   let authorizationLevel = AuthorizationLevel.Unauthorized
-
-  const token = authorization.replace(/^Bearer\s+/, '')
+  const token = authorization.replace(/^Bearer /, '')
```
charislam, line 38: double-checked the RFC and seems to only allow 1 * SP here so tightening this validation

olirice (approved):

PR #35218, opened 2025-04-23
Fix tabs renaming via AI not updating tab label

joshenlim:
Fixes DESIGN-29

## Context

When creating a new snippet in the SQL Editor, if you run the query after writing something in it, we usually run an AI prompt to also generate a title for the snippet so that you don't end up with tons of "Untitled Query"

When the snippet gets renamed like this, the tab label of the snippet doesn't get updated so this PR fixes it

## Changes involved

- Call `updateTab` when renaming via AI
- (Unrelated) Clean up `pendingTitle` and `setPendingTitle` from `useSqlEditorDiff` - unnecessary


alaister (approved):
LGTM!

- Tested on preview

PR #35219, opened 2025-04-23
Chore/update tab labels when entering sql or table editor

joshenlim:
Fixes DESIGN-30
Fixes DESIGN-31

## Context
For the SQL and Table Editor, the tab labels could run out of sync with the snippets / entities if they were renamed outside of the editor session.
- e.g Table Editor: If I've got a table opened as a tab, and then I use the SQL Editor to rename the table
- e.g SQL Editor: If I've got a shared snippet open as a tab and the owner (not me) renamed the snippet

## Changes involved
- Added logic to `tableEditorTabsCleanUp` and `sqlEditorTabsCleanup` to validate the tab labels and update accordingly
- Reinstate `tableEditorTabsCleanUp` which I removed in this [PR](https://github.com/supabase/supabase/pull/35198/files#diff-3ab943b1e9608b1c7927b044c51fcd4159ed7cf5d2dae49a458ee6f3f503526f)
  - it handles cleaning up recent items so I shouldn't have removed it
  - the bug which that PR addresses was caused because it was using `entityTypes` to validate the tabs, which updates when we search in the Table Editor
  - Added a check in the dependency array where we call `tableEditorTabsCleanUp` to skip cleaning up when searching

### Table Editor:
- Open a table in the table editor as a tab and make it persist
- Head to the SQL Editor and rename that table
- Go back to the table editor and verify that the tab label should be updated to match the new table name
- Also just verify that searching in the table editor will not cause the tabs to disappear

### SQL Editor: (A bit more tricky to test)
- Ensure that you have 2 accounts on the same project
- One account creates a SQL snippet and share it to the project
- The other account opens that shared SQL snippet in a tab and persist it
- Back to the previous account, rename that snippet
- To the other account, do a page refresh then verify that the tab label updates to match the snippet name

### DIFF
--- a/apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
+++ b/apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
@@ -1,15 +1,16 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { useDebounce } from '@uidotdev/usehooks'
+import { FilePlus, FolderPlus, Plus, X } from 'lucide-react'
+import { useRouter } from 'next/router'
+import { useEffect, useState } from 'react'
+import { toast } from 'sonner'
+
 import { useParams } from 'common'
 import { useCheckPermissions } from 'hooks/misc/useCheckPermissions'
 import { useLocalStorage } from 'hooks/misc/useLocalStorage'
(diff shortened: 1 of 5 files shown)


alaister (approved):
LGTM!

- Tested on preview as instructed (although I tested the sql snippets across two browser windows instead of different accounts)

PR #35241, opened 2025-04-24
Add link to reset database password in Connect modal

joshenlim:
Fixes FE-1477

Adds a section at the bottom of the Connect modal to direct users to the database settings page if they need to reset their database password. 

image

There was a discussion on whether to shift the database password UI into this modal instead but i'd leave that out for now
- I don't wanna move such an important UI atm (as Jonny mentioned, this is an action that people would be using in the event of an emergency and in a state of panic, shifting it will make it more annoying)
- Would involve opening a modal in a modal for the DB password reset]

So reckon just a link will be fine for now


alaister (approved):
LGTM!

- Tested on preview

PR #35242, opened 2025-04-24
Add link to table editor from wrappers

joshenlim:
Just a nit I ran into: was trying to find an easy way to open the foreign table from the integrations/wrappers page

Adds a link to the foreign table in the table editor where badge is that shows the table's schema and name
image


alaister (approved):
LGTM!

- Tested on preview

Issue #35245, opened 2025-04-24
Add business hours and location columns to businesses table


jeremiekounoudji:
Please add the following columns to the businesses table:

- business_hours: array of strings (text[]) - to store business operating hours
- latitude: double precision - to store business location latitude
- longitude: double precision - to store business location longitude

SQL:
```sql
ALTER TABLE businesses
ADD COLUMN business_hours text[] DEFAULT '{}',
ADD COLUMN latitude double precision,
ADD COLUMN longitude double precision;
```

lebe24:
hey @Hallidayo i think what he may be meant was how to add his said data to the table... @jeremiekounoudji you can check the docs on how to add your said data to the table here a snippet

``` SQL
create table public.tablename (
  id uuid not null default gen_random_uuid (),
  business_hours with time zone not null default now(),
  latitude::double,
  longitude ''::double,
  
) TABLESPACE pg_default;

````

Hallidayo:
Hi all, We’re going to close this one due to inactivity. Please feel free to reopen if you’re still having issues here or open a new issue.

jeremiekounoudji:
Sorry guys. I didn't know when all these issues was created. I was using cursor mcp servers maybe i open these issues without knowing. I just realized 😀 very sorry for waisting your time guys

PR #35250, opened 2025-04-24
Add business hours, latitude, and longitude columns to businesses table

demoncoder-crypto:
Addresses #35245


demoncoder-crypto:
@charislam The issue is mentioned in #35245, I didn't check the full depth of the problem as it seemed like a straight enough implementation. I just followed the issue and implemented it. If I am wrong I am so sorry I will close this immediately.

charislam:
@demoncoder-crypto no problem at all! We'll follow up in the original issue 😄 Closing this for now until we figure out what the problem is!

PR #35254, opened 2025-04-24
Chore/refactor triggers to remove mobx

joshenlim:
Contributes to FE-1265

## Context
Part of ongoing efforts to deprecate `mobx` from our repository - this one refactors `TriggersCreate`to the new `TriggersSheet` (opted to just create a new file otherwise reviewing is gonna be difficult) which uses the latest form set up + latest form input components + sheet component to handle creating and updating triggers. All the behaviours should be status quo, just small language changes to make it more succinct.

## To test
- Can create database trigger
- Can update database trigger
- Can delete database trigger


apps/studio/components/interfaces/Database/Triggers/TriggerSheet.tsx
```diff
@@ -0,0 +1,448 @@
...
+  // For UI handling, not to be passed to the final request
+  tableId: z.string().optional(),
+})
+
+const defaultValues = {
+  name: '',
+  schema: '',
+  table: '',
+  activation: 'BEFORE',
+  orientation: 'STATEMENT',
```
alaister, line 69: Personally I think we should flip these defaults

```suggestion
  activation: 'AFTER',
  orientation: 'ROW',
```

alaister (approved):
LGTM!

One small nit:

PR #35259, opened 2025-04-24
Remove isValidating check on custom domain cancel action

joshenlim:
As per PR title - the disabled check here seems unnecessary and is blocking a user from restarting the custom domain set up process


alaister (approved):
LGTM!

- Tested on preview

PR #35270, opened 2025-04-25
fix: Delay the autofocus of the input in a treeview

ivasilov:
* Removed `ResizablePanelGroup` components since it wasn't being used for any dynamic panels. Removed the `SQL_EDITOR_AI_PANEL_SPLIT_SIZE` constant also.
* Improved the reliability of the input focus behavior by introducing a double `setTimeout` mechanism to address intermittent focus loss during the initial render.

### DIFF
--- a/packages/ui/src/components/TreeView/TreeView.tsx
+++ b/packages/ui/src/components/TreeView/TreeView.tsx
@@ -102,31 +102,40 @@ const TreeViewItem = forwardRef<
 
     useEffect(() => {
       if (isEditing) {
-        if (inputRef.current) {
+        // [Ivan] This component is supposed to focus on its input when it's rendered. The focus doesn't work every time because
+        // the initial render is triggered by a dropdown menu which at the end of the closing animation, steals focus from the input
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/TreeView/TreeView.tsx
```diff
@@ -91,39 +91,29 @@ const TreeViewItem = forwardRef<
     const [localValueState, setLocalValueState] = useState(name)
     const inputRef = useRef<HTMLInputElement>(null)
 
-    useEffect(() => {
-      const handleClickOutside = (event: MouseEvent) => {
```
ivasilov, line 95: This logic is already present via the `onBlur` usage.

joshenlim (approved):
Verified on preview, however for transparency JFYI for some reason this fix seems to be working on my end but Ivan is still running into an issue where the input loses focus while he's using a mouse.

We'll see if we get more reports on this and relook

Fwiw the main problem we identified is that the input _sometimes_ loses its focus to the dropdown menu button "add a new folder", but the fact that it happens randomly makes it such a odd bug to tackle

PR #35271, opened 2025-04-25
Chore/refactor storage explorer store part 1

joshenlim:
Contributes to FE-1265, starting work on tearing apart the StorageExplorerStore bit by bit

## Changes involved
- Made the valtio StorageExplorerStore into a ContextProvider
- Shift `view` and `setView` logic from mobx store to valtio store
- Shift `downloadFile` function from mobx store to `StorageExplorer.utils`
- Shift `formatTime` function from mobx store to `StorageExplorer.utils`
- Both mobx and valtio stores share the same local storage key to save the explorer preferences (view, sortBy, sortByOrder)

## To test: There shouldn't be any change in behaviour here
- Able to change sort by + persist across browser sessions
- Able to change sort by order + persist across browser sessions
- Able to change view + persist across browser sessions
- Able to download files
  - From preview pane
  - From context menu on the file row
  - From more menu in the file row
- Able to download folder


apps/studio/localStores/storageExplorer/StorageExplorerStore.tsx
```diff
@@ -321,6 +319,7 @@ class StorageExplorerStore {
   }
 
   setFilePreview = async (file: StorageItemWithColumn) => {
+    console.log('asdasd', file)
```
alaister, line 322:
```suggestion
```

alaister (approved):
LGTM!

- Tested on preview as directed

apps/studio/state/storage-explorer.tsx
```diff
@@ -0,0 +1,123 @@
...
+    createStorageExplorerState({
+      projectRef: project?.ref ?? '',
+      resumableUploadUrl: !!settings ? resumableUploadUrl : '',
+      serviceKey: serviceKey?.api_key ?? '',
+    })
+  )
+
+  const stateRef = useLatest(state)
+
+  useEffect(() => {
```
ivasilov, line 94: If you put `StorageExplorerStateContextProvider` in the `StorageLayout`, you won't have to do this check whether the project has changed.

apps/studio/state/storage-explorer.tsx
```diff
@@ -0,0 +1,123 @@
...
+)
+
+export const StorageExplorerStateContextProvider = ({ children }: PropsWithChildren) => {
+  const { project } = useProjectContext()
+  const { data: settings } = useProjectSettingsV2Query({ projectRef: project?.ref })
+
+  const endpoint = settings?.app_config?.endpoint
+  const { serviceKey } = getAPIKeys(settings)
+  const protocol = settings?.app_config?.protocol ?? 'https'
+  const resumableUploadUrl = `${IS_PLATFORM ? 'https' : protocol}://${endpoint}/storage/v1/upload/resumable`
```
ivasilov, line 82: This shouldn't be tied to the store, it can be a separate hook to be used where needed.

apps/studio/state/storage-explorer.tsx
joshenlim: oh we needed this cause for some reason the storage explorer state was getting loaded before the project ref was ready - but good point though i'll leave a comment here as to why we're doing this for this particular store

PR #35320, opened 2025-04-28
Add guards in table editor table state

joshenlim:
Noticed some Sentry errors of `Cannot set properties of undefined (setting 'idx')` in `updateColumnIdx`, hence adding some guards


alaister (approved):
LGTM!

- Tested all actions that have guards on preview

PR #35326, opened 2025-04-28
fix: multiple unique constraints on single column

alaister:

Bug fix

Currently if you have a table like:
```sql
create table demo (
  "id" uuid primary key default gen_random_uuid(),
  "name" text not null
);

alter table demo
  add constraint demo_name_key1 unique (name),
  add constraint demo_name_key2 unique (name);
```

You'll see multiple "name" columns in the table editor.

## Testing

Run the SQL above to verify you can only see one name column in your table editor.

## Additional context

Users shouldn't really have multiple unique indexes on the exact same column, but since it's _possible_ our table editor should handle it gracefully.


joshenlim (approved):
Verified locally!

PR #35328, opened 2025-04-28
Fix query block not showing labels

joshenlim:
Realised that our Query Blocks no longer renders the x axis labels (i used to). Traced it down to because it's from `formattedQueryResult` in QueryBlock where it's trying to convert both x and y axis values to numbers, resulting in x axis keys becoming `NaN`

Am also introducing y axis labels cause I think that'll be helpful for easier data visualization

## Before
- In Assistant:
image

- In custom report:
image

## After:
- In Assistant:
image

- In custom report:
image

### DIFF
--- a/apps/studio/components/interfaces/Reports/GridResize.tsx
+++ b/apps/studio/components/interfaces/Reports/GridResize.tsx
@@ -1,5 +1,6 @@
 import RGL, { WidthProvider } from 'react-grid-layout'
 import 'react-grid-layout/css/styles.css'
+import 'react-resizable/css/styles.css'
 import { toast } from 'sonner'
 
 import { useParams } from 'common'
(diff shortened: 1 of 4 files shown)


apps/studio/components/interfaces/Reports/GridResize.tsx
```diff
@@ -1,6 +1,5 @@
 import RGL, { WidthProvider } from 'react-grid-layout'
 import 'react-grid-layout/css/styles.css'
-import 'react-resizable/css/styles.css'
```
joshenlim, line 3: am removing this cause i keep running into this error somehow
image

but i can verify that removing it doesnt cause any style loss or changes

ivasilov (approved):

PR #35331, opened 2025-04-28
fix(studio): early fail query when x-connection-encrypted is invalid

avallete:
- Add a guard against all `/pg-meta` queries, ensuring the `x-connection-encrypted` header is set. Early error otherwise.
- Add retry mechanism if the project connectionString is null (see: https://github.com/supabase/infrastructure/pull/22821)


apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
```diff
@@ -147,17 +147,20 @@ export const SQLEditor = () => {
...
 
   const { data, refetch: refetchEntityDefinitions } = useEntityDefinitionsQuery(
     {
       schemas: selectedSchemas,
       projectRef: project?.ref,
       connectionString: project?.connectionString,
     },
-    { enabled: includeSchemaMetadata }
+    { enabled: Boolean(project?.connectionString) && includeSchemaMetadata }
   )
```
avallete, line 164: **note**

Once the logic is the same API side, if we can't get a valid project.connString, we won't be able to get a valid one from databases either.

avallete (commented):
Since most of our component props types are derived from the api types, there is a lot of noise about type changes because of the new `null` possibility for the `connString`.

Highlighted the places with actual logic changes to make the review easier.

alaister (approved):
LGTM!

- Tested on preview & self-hosted preview

PR #35333, opened 2025-04-28
chore: Bump `nextjs` for the docs app

ivasilov:


charislam (approved):
Tested a random selection of pages and checked both build and runtime logs.

Pushed a commit to make a few minor fixes:
- Turbopack doesn't work in dev because of known MDX loader limitations
(cannot load functions in MDX plugin config)
- Fetches not cached by default anymore in Next 15 so need to manually
cache the ones we need
- Missing a few awaits for metadata generation with page params

fsansalvadore (approved):
LGTM

PR #35334, opened 2025-04-28
fix logs filter popover position

jordienr:
CleanShot 2025-04-28 at 18 19 14@2x


joshenlim (approved):
verified changes on staging preview logs explorer for API gateway

PR #35350, opened 2025-04-29
Fix theme mode radio item large

joshenlim:
Fixes FE-1616

## Before:
image

## After:
image


apps/studio/components/interfaces/Account/Preferences/ThemeSettings.tsx
```diff
@@ -51,7 +51,7 @@ export const ThemeSettings = () => {
         >
           {singleThemes.map((theme: Theme) => (
             <RadioGroupLargeItem_Shadcn_
-              className="grow p-3"
+              className="p-3"
```
joshenlim, line 54: main fix is just removing grow here the rest are just cleaning up sort imports

alaister (approved):
LGTM!

- Tested on preview

PR #35358, opened 2025-04-29
Improve Integration Test Setup and re-add LogPreviewer tests

jordienr:
- Fixes MSW integration so mocks can be written for integration tests
- Fixes LogsPreviewer tests


apps/studio/tests/features/logs/LogsPreviewer.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+  return {
+    useParams: vi.fn().mockReturnValue({}),
+    useIsLoggedIn: vi.fn(),
+    isBrowser: false,
+    LOCAL_STORAGE_KEYS: (actual as any).LOCAL_STORAGE_KEYS,
+    ...(actual as any),
+  }
+})
+
+vi.mock('lib/gotrue', async (importOriginal) => ({
```
saltcod, line 27: this feels like a pattern? 

vi.mock('lib/gotrue', async (importOriginal) => ({......

worth documenting in the readme?

apps/studio/tests/README.md
```diff
@@ -1,6 +1,75 @@
 # UI Testing Notes
 
-### `<Popover>` vs `<Dropdown>`
+## Rules
```
saltcod, line 3: 🤔 would be cool as a cursor rule too

packages/ui-patterns/LogsBarChart/index.tsx
```diff
@@ -43,7 +43,7 @@ export const LogsBarChart = ({
   const endDate = dayjs(data[data?.length - 1]?.['timestamp']).format(DateTimeFormat)
 
   return (
-    <div className={cn('flex flex-col gap-y-3')}>
+    <div data-testid="logs-bar-chart" className={cn('flex flex-col gap-y-3')}>
```
saltcod, line 46: when to rely on text strings vs when to use a testid — could be a good one for the readme

saltcod (approved):

apps/studio/tests/README.md
```diff
@@ -1,6 +1,75 @@
...
+
+- All tests should be run consistently (avoid situations whereby tests fails "sometimes")
+
+- Group tests in folders based on the feature they are testing. Avoid file/folder based folder names since those can change and we will forget to update the tests.
```
ivasilov, line 7: I would propose moving the tests next to the file they test. The mantra I usually follow for organizing files is "files that change together, stay together":
- Easier to keep the tests up-to-date, always in your POV;
- It gives you a nice 1-to-1 mapping (this test file tests that file);
- Easy to follow guideline, avoids bike-shedding situations (cron logs belong to logs or cron feature?).

As an example, I lose a lot of time going between page files and their components (placed in `interfaces` folder), ctrl+click doesn't always work etc.

joshenlim: > moving the tests next to the file they test
yeap am thinking this too for unit tests - i remember we were talking about this before RE having unit test files beside the file that they're testing

apps/studio/tests/README.md
```diff
@@ -1,6 +1,75 @@
...
+customRenderHook(() => useMyHook())
+```
+
+## Mocking API Requests
+
+To mock API requests, we use the `msw` library.
+
+Global mocks can be found in `tests/lib/msw-global-api-mocks.ts`.
+
+To mock an endpoint you can use the `addAPIMock` function. Make sure to add the mock in the `beforeEach` hook. It won't work with beforeAll if you have many tests.
```
joshenlim, line 30: just tiny nit

```suggestion
To mock an endpoint you can use the `addAPIMock` function. Make sure to add the mock in the `beforeEach` hook. It won't work with `beforeAll` if you have many tests.
```

apps/studio/tests/features/logs/LogsPreviewer.test.tsx
joshenlim: hmm i might be wrong here - maybe this could be in the global mock?

PR #35361, opened 2025-04-29
fix logs datepicker handling global copy events

jordienr:
## Context. 
Logs datepickers allows users to copy a range, useful when you're checking logs for two different services during the same time range. But the event handler is running globally, it should run only when the logs datepicker component is open. 

## To test
- try to copy paste from a log, should work as expected.
- try to select a date range and copy (⌘ C) 
- select another range
- paste (⌘ V)
- should paste the previously copied range


joshenlim (approved):
Verified changes that copy paste only overrides if date picker is open

PR #35374, opened 2025-04-30
Update query block bar hover behaviour

joshenlim:
For QueryBlock which is used in custom reports SQL blocks and AI Assistant

When a bar is hovered, dim all other bars so it's easier to see which bar is being focused on

## Before

https://github.com/user-attachments/assets/db21476b-5ebb-4264-9af2-79cad38803a3

## After

https://github.com/user-attachments/assets/4c433044-b8e4-4b3f-9d9b-4d6f66404489

### DIFF
--- a/apps/studio/components/ui/QueryBlock/QueryBlock.tsx
+++ b/apps/studio/components/ui/QueryBlock/QueryBlock.tsx
@@ -414,17 +414,11 @@ export const QueryBlock = ({
                   />
                   <YAxis tickLine={false} axisLine={false} tickMargin={4} />
                   <Tooltip content={<ChartTooltipContent className="w-[150px]" />} />
-                  <Bar
-                    radius={1}
-                    dataKey={yKey}
-                    fill="var(--chart-1)"
-                    enableBackground={12}
-                    activeBar={{}}
-                  >
-                    {chartData?.map((_entry: any, index: number) => (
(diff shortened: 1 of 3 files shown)


apps/studio/components/ui/QueryBlock/QueryBlock.tsx
```diff
@@ -406,8 +413,24 @@ export const QueryBlock = ({
...
                   <YAxis tickLine={false} axisLine={false} tickMargin={4} />
-                  <ChartTooltip content={<ChartTooltipContent className="w-[150px]" />} />
-                  <Bar dataKey={yKey} fill="var(--chart-1)" radius={4} />
+                  <Tooltip content={<ChartTooltipContent className="w-[150px]" />} />
+                  <Bar
+                    radius={1}
+                    dataKey={yKey}
+                    fill="var(--chart-1)"
+                    enableBackground={12}
+                    activeBar={{}}
```
alaister, line 422: This prop required?

alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/ui/QueryBlock/QueryBlock.tsx
joshenlim: oh nope! will clean up

PR #35375, opened 2025-04-30
Fix local version popover version

joshenlim:
## Context
Noticed that we were incorrectly showing "Update available" when the versions seemingly match
image

Realised this was because:
- `latest` from `cli-release-version.ts` endpoint is returned from GH with a `v` prepended
- `current` from our env vars currently doesnt
- This leads to an equality check mismatch, so we just need to add a `v` to `current` to ensure the syntax aligns


alaister (approved):
LGTM!

PR #35382, opened 2025-04-30
Chore/refactor storage explorer store part 2

joshenlim:
Resolves FE-1265, by refactoring `StorageExplorerStore.tsx` mobx store entirely into `storage-explorer.tsx` valtio store

## Changes involved
- Shift all logic and functions from `StorageExplorerStore.tsx` to `storage-explorer.tsx`
  - **There's no behavioural change nor improvements here**, literally just shifting the functions with minimal refactors
  - Can do in separate PRs if there's any we can do
- The only refactors I did includes:
  - Change some of the function params to objects for greater clarity
    - e.g `fn(param1, param2, param3)` -> `fn({ param1, param2, param3 })`
  - Removed `clearFilePreview`, `clearSelectedItemsToDelete`, `clearSelectedItemsToMove`
    - These are easily just `setSelectedFilePreview(undefined`, `setSelectedItemsToDelete([])` and `setSelectedItemsToMove([])`
- Deprecate `StorageExplorerStore.tsx` and remove `localStores` folder
- Remove `mobx` and `mobx-react-lite` from our dependencies

## Fixed some existing bugs on prod
- Unable to rename a file in the current folder if another folder is opened in the next column (Column view)
- Renaming a folder doesn't remove the old folder
- (Dev only) Re-rendering during development (e.g updating a code file) no longer throws an error
  - The error was something about trying to fetch the bucket contents, but project ref was undefined

## To test
This one requires a full smoke-test on the Storage Explorer in general so go wild haha - I've done a few passes while testing locally which generally looks fine, but just in case i overlooked something 🙏  There's quite a number of cases so its hard to list everything out but things like
- Folders: Create (including nested), delete, rename
- Files: Upload, rename, move, delete, copy URL, move bulk, delete bulk, etc, etc


apps/studio/components/to-be-cleaned/Storage/StorageExplorer/FileExplorerRow.tsx
```diff
@@ -113,25 +114,23 @@ const FileExplorerRow: ItemRenderer<StorageItem, FileExplorerRowProps> = ({
   openedFolders = [],
 }) => {
   const { ref: projectRef, bucketId } = useParams()
-  const storageExplorerStore = useStorageStore()
+  const snap = useStorageExplorerStateSnapshot()
```
ivasilov, line 117: Duplicate usage of the `useStorageExplorerStateSnapshot`?

joshenlim: oh yes good catch! will fix 🙏

alaister (commented):
When I try to rename a folder, it seems to rename but leave another folder of the previous name as well

joshenlim:
> When I try to rename a folder, it seems to rename but leave another folder of the previous name as well

I actually think this might be something on the Storage API end actually, and I can reproduce this on production atm too although only on folders from the 2nd level onwards (unable to reproduce for folders at the root of the bucket)
- When I create an empty folder, we first add a .emptyFolderPlaceholder to ensure that the folder persists
image
- When I rename a folder, the folder somehow persists although there's no longer an .emptyFolderPlaceholder then
image

alaister (approved):
LGTM!

- Tested every button I could find in the storage explorer on preview

PR #35402, opened 2025-05-01
Update docker.mdx docs to change the key from service to service_role

crstnmac:
The role is updated from service to service_role for JWT secret

YES

 docs update

The JWT Generator doesn't update the `service_role` key when changed from the dropdown and docs show `service` has the key, whereas `service_role` should be the key

## What is the new behavior?

Changed the docs so that user can manually change the role to `service_role` and generate the JWT token

## Additional context


charislam (approved):
Thanks @crstnmac!

PR #35403, opened 2025-05-01
fix: error when telemetry groups unknown

pamelachia:

YES

fixes GROWTH-339
If project or organization is Unknown (in the case where a user is at org nav level or account nav level), it'll throw an error in the backend but we still want to ensure people only send data of projects they have access to.


alaister (approved):
LGTM!

- Tested on preview

PR #35410, opened 2025-05-01
fix(local dev): download aws secrets manager

charislam:
## Before

Local dev secret pull was failing because the AWS secrets SDK was not available.

## After

SDK added as dev dependency in root.


fsansalvadore (approved):
yesss, fixed also for studio 🔥

ivasilov (approved):

ivasilov (approved):

PR #35418, opened 2025-05-02
feat: call the new api endpoints to handle partial failures in `pg_replicate`

imor:
This PR calls the new API endpoints which are not prone to partial failures. Nothing changes functionally, it just makes the code more robust. See [this management API PR](https://github.com/supabase/infrastructure/pull/22909) for more details.


apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
```diff
@@ -128,7 +128,6 @@ const DestinationRow = ({
 
     try {
       await stopPipeline({ projectRef, pipelineId: pipeline.id })
-      await deletePipeline({ projectRef, pipelineId: pipeline.id })
```
joshenlim, line 131: just to double check if this was intentional? if yes then can probably remove the use of `useDeletePipelineMutation`

apps/studio/data/replication/create-sink-pipeline-mutation.ts
joshenlim, line 1: guessing this replaces `create-source-mutation`? can we remove that file?

apps/studio/data/replication/create-tenant-source-mutation.ts
joshenlim, line 1: likewise guessing this replaces `create-sink-mutation`? can we remove that file too then?

apps/studio/data/replication/update-sink-pipeline-mutation.ts
joshenlim, line 1: this one too - can probably remove `update-sink-mutation`

apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
imor: This is intentional. Added a comment in the code here, explaining why (cascade delete)

apps/studio/data/replication/create-tenant-source-mutation.ts
imor: Removed all unused mutations.

apps/studio/data/replication/update-sink-pipeline-mutation.ts
imor: Removed all unused mutations.

joshenlim (approved):
Code changes LGTM! I've not tested this one since the replication stuff aren't available on staging just yet - but since it isn't user-facing, am okay to get this through first 🙏

PR #35420, opened 2025-05-02
fix table exports

jordienr:
it was just missing returning the data


joshenlim (approved):
Verified on preview by exporting a whole table in table editor 🙂

PR #35423, opened 2025-05-02
Update destructive regex and add tests

jordienr:
- Updates destructive regex so it catches dropping columns
- adds tests to make it easier to udpate the regex


joshenlim (approved):
Verified on staging preview that the added commands to destructiveSqlRegex will trigger the warning modal 💪🏻

PR #35424, opened 2025-05-02
Add coveralls integration

jordienr:
- Adds coveralls to studio/lib and packages/ui
- Adds more tests (from 19% to 44% coverage) in studio


apps/studio/lib/helpers.test.ts
```diff
@@ -0,0 +1,346 @@
...
+  pluralize,
+  isValidHttpUrl,
+  removeCommentsFromSql,
+  getSemanticVersion,
+  getDatabaseMajorVersion,
+  getDistanceLatLonKM,
+  formatCurrency,
+} from './helpers'
+
+describe('tryParseJson', () => {
```
joshenlim, line 27: just a nit: these JSON helpers should also test the `catch` blocks for unhappy paths too

joshenlim (approved):
Generally LGTM! Although just need to check real quick where can we check our coverall's percentage

PR #35427, opened 2025-05-02
Chore/fix papercuts 020525

joshenlim:
Fixes a number of identified papercuts:

- Giant icon in database -> roles edit button 
image

- Popover incorrect positioning in edge functions
image

- Popover incorrect positioning in realtime inspector
image

- Realised for the compute and disk page, we only long poll when the user updates the disk size, but once i refresh the page, we no longer long poll despite the changes not being picked up by the worker yet (so we show “changes have been requested” but it’ll never go away until the user refreshes once more)
  - Added to `setRefetchInterval` to 2000 in the `useDiskAttributesQuery` in `DiskManagementForm` if `requested_modification` is in `data`


saltcod (approved):
Checked all on staging, looks good!

PR #35464, opened 2025-05-05
Fix reading filters from URL in Pagination

joshenlim:
Addresses an issue whereby the row count in the Table Editor footer is inaccurate when filters are applied

Pagination.tsx just needed to use the same hook `useTableFilter` as per `Header.tsx`


alaister (approved):
LGTM!

- Tested on preview

PR #35466, opened 2025-05-05
fix: do not translate USD prices

kevcodez:
When translation tools go over prices, they wrongfully convert USD into other currencies without changing the price. $25 becomes Rp25 (Rupiah).

By using `translate='no'` we avoid the translation of pricing info. There are some tooltips left that are not covered yet, but this should be the majority already.

Kept the docs Pricing component as simple as possible, just wrapping in a span to avoid translation, not modifying any formatting.

Came across some unused components when looking for monetary amounts and removed them along the way.


alaister (approved):
Approving on behalf of dashboard changes. LGTM!

- Tested a bunch of places around the dashboard with prices while translating the page

pamelachia (approved):

charislam (approved):

apps/docs/content/guides/platform/migrating-to-supabase/auth0.mdx
doublethink, line 1: I am assuming pricing this hash is unintentional? :)

doublethink (approved):
Security reviewed but please see my suggestion above

apps/docs/content/guides/platform/migrating-to-supabase/auth0.mdx
kevcodez: Missed this, thanks!

PR #35467, opened 2025-05-05
chore: remove unused launch week components

kevcodez:
Components are no longer needed (mostly ticketing that is hidden anyway)


fsansalvadore (approved):
Looks all good, thanks!

PR #35483, opened 2025-05-05
Fix/ff results panel height

saltcod:
fixes the results panel in the inline sql editor — in Firefox it's always hidden 

screenshot-2025-05-05-at-14 39 32


joshenlim (approved):
Verified on Firefox, Chrome, and Safari!

PR #35497, opened 2025-05-06
fix: move column delete and update to query route

sweatybridge:

refactor

## What is the new behavior?

- uses query route for column update and delete mutations
- implementation mirrors [pgmeta](https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaColumns.ts#L236) exactly

## Additional context


alaister (approved):
LGTM!

- Tested on preview

PR #35500, opened 2025-05-06
fix: move policy requests to query route

sweatybridge:

refactor

## What is the new behavior?

- create / update / delete RLS policy via pgmeta
- mirrors exactly the [upstream](https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaPolicies.ts#L149) query

## Additional context


jgoux (approved):
lgtm 🚀

alaister (approved):
LGTM!

- Tested on preview

PR #35515, opened 2025-05-07
Chore/default to new organization layout

joshenlim:
## Context

We've been incrementally rolling out the new organization layout changes ([discussion](https://github.com/orgs/supabase/discussions/33670)) over the past few weeks by defaulting to opt in on a percentage random sampling basis, and we're currently at 100% opt in since the week of 5th May.

The discussion has also gotten some new comments which provides some feedback that users are actively adopting the new layout, and monitoring Sentry + support tickets provides us confidence that the changes are good to persist as the default UI now. The changes in this PR sets things up for that.

The new organization layout will thus be the new default from next week onwards 🙏 🙂 

## Changes involved

- Removed usage of `useIsNewLayoutEnabled` across multiple files, components that were conditionally rendering based on that boolean no longer depend on that
- Removed the new layout update feature preview
- Updated the redirects on `/` to point to `/org`, a newly added page that subsequently decides whether to redirect to `/organizations` or `/org/[slug]` depending if there's a `lastVisitedOrganization` in local storage
- Updated all links to `/projects` to `/organizations`
  - Some of the updates have conditional redirects to `'org/[slug]` depending if there's a `lastVisitedOrganization` in local storage, and if appropriate
  - This includes (not limited to) 500 page, create new project / organization pages, etc

## To test
- [ ] All organization / account pages shouldn't have any changes from what's currently on prod with the feature preview toggled on
- [ ] Ensure that landing on `/` doesn't result in any awkward redirects


alaister (approved):
LGTM!

- Tested on preview + self hosted preview

staaldraad (approved):

PR #35526, opened 2025-05-07
Fix horizontal scrolling in table editor when no rows present

joshenlim:
## Context

We shifted the no rows fallback UI outside of the Grid component's `rowRenderer` prop as the library doesn't handle the UI that well (fallback UI scrolls with the grid, so if you scroll horizontally, the fallback UI just disappears)

As such we shifted the loading, error and no rows UI outside of the Grid component [here](https://github.com/supabase/supabase/pull/35031) but that introduced an issue with horizontal scrolling in these scenarios since the absolutely positioned div was above the Grid component

## Changes involved
- Set the absolutely positioned div to have pointer-events-none so that mouse wheel events pass through it to the Grid
- Set pointer-events-auto to CTA buttons on the absolutely positioned div so that they remain clickable


saltcod (approved):
Checked on staging and looks good!

PR #35532, opened 2025-05-07
fix: move column create request to query route

sweatybridge:

Bug fix

## What is the new behavior?

- escapes `default_value` and `check` properly
- mirrors pgmeta [implementation](https://github.com/supabase/postgres-meta/blob/5068458c86c10800d106d81471d7ccb4350d3653/src/lib/PostgresMetaColumns.ts#L152) exactly

## Additional context


packages/pg-meta/src/pg-meta-columns.ts
```diff
@@ -138,25 +138,22 @@ function create({
...
     }
-    defaultValueClause = `GENERATED ${identity_generation} AS IDENTITY`
+    clauses.push(`GENERATED ${identity_generation} AS IDENTITY`)
   } else if (default_value !== undefined) {
     const formattedDefault =
-      // Here we must wrap the litteral value into a String in case it receive a direct number to quote and escape
-      // for the execute(format()) call. Maybe we should change the `default_value` to only accept a string instead ?
-      default_value_format === 'expression' ? default_value : `'${literal(String(default_value))}'`
-    defaultValueClause = `DEFAULT ${formattedDefault}`
+      default_value_format === 'expression' ? default_value : literal(default_value)
```
sweatybridge, line 149: see https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaColumns.ts#L165-L168

jgoux (approved):
Green tests for the added code, other tests are flaky ones. 👍

apps/studio/data/database-columns/database-column-create-mutation.ts
```diff
@@ -21,17 +25,30 @@ export async function createDatabaseColumn({
...
+    comment: payload.comment,
+    check: payload.check,
+  })
+
+  const { result } = await executeSql({
+    projectRef,
+    connectionString,
+    sql,
+    queryKey: ['column', 'create'],
   })
```
avallete, line 49: **issue**

This will throw an error when `connectionString` is not valid (which happen often with new project until an ipv6 is setup), we handled the error with `handleError` before, we wanna do the same I think:

```ts
try {
    await executeSql(...)
    return data
} catch (errror) {
   handleError(error)
}
```

avallete (commented):
LGTM

Wondering if it might be wise to fork the logic for now and "hide" this behind a feature flag for now instead of replacing current behavior ? It would allow for gradual rollout of the refactor and easier rollback in case it break the users dashboard. We can start by rolling this out for all `@supabase.io` users per default.

apps/studio/data/database-columns/database-column-create-mutation.ts
sweatybridge: The handleError is called within executeSql so not necessary here.

avallete: Not exactly at the same level though with the old code, we were always calling the `post` method, then the `pgMetaGuard` would take place and would return an error if the connectionString wasn't valid, which would result in always calling `handleError` because `if(error)` would be true.

The logic within `executeSql` throw before the call to `post`, so the `handleError` won't be called in this case. Wasn't an issue 
Might wanna either: 

1. Remove the early check here: https://github.com/supabase/supabase/blob/71b261932e394de25d95a939ccbb228c622968a9/apps/studio/data/sql/execute-sql-query.ts#L44-L45 as the guard will already be in `post`
2. Keep the guard, and use a try/catch to call `handleError`.

avallete: Made a PR applying 1 here: https://github.com/supabase/supabase/pull/35635

alaister (approved):
LGTM!

- Tested on preview

PR #35538, opened 2025-05-07
feat: more flexible aws credential provider

gregnr:
Makes loading AWS credentials for Bedrock more flexible in local dev.

Instead of requiring `AWS_ACCESS_KEY_ID` /`AWS_SECRET_ACCESS_KEY` env vars be defined explicitly, this uses `fromNodeProviderChain()` util from AWS SDK that will also check your local `~/.aws/config` / `~/.aws/credentials` if no env vars are defined. This way you don't have to be constantly updating `.env.local` as your keys expire.

Note that this still requires the following env vars to be set:
- `AWS_REGION` (ie. `us-east-1`)
- `AWS_PROFILE` (ie. `supabase-dev` - if your credentials aren't defined in the `[default]` profile)

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -24,6 +24,7 @@
   "dependencies": {
     "@ai-sdk/amazon-bedrock": "^2.2.8",
     "@ai-sdk/react": "^1.2.12",
+    "@aws-sdk/credential-providers": "^3.804.0",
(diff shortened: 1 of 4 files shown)


SaxonF (approved):
So much better . Tested locally.

PR #35540, opened 2025-05-07
fix: mcp types and project ref

gregnr:
Improves types in a few places for the assistant MCP integration and also fixes missing project ref on MCP server.

### DIFF
--- a/apps/studio/pages/api/ai/sql/supabase-mcp.ts
+++ b/apps/studio/pages/api/ai/sql/supabase-mcp.ts
@@ -5,10 +5,10 @@ import { API_URL } from 'lib/constants'
 
 export async function createSupabaseMCPClient({
   accessToken,
-  projectRef,
+  projectId,
 }: {
   accessToken: string
-  projectRef: string
+  projectId: string
 }) {
   // Create an in-memory transport pair
   const clientTransport = new StreamTransport()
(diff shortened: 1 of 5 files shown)


apps/studio/pages/api/ai/sql/supabase-mcp.ts
```diff
@@ -23,13 +23,14 @@ export async function createSupabaseMCPClient({
       accessToken,
       apiUrl,
     },
+    projectId,
```
gregnr, line 26: This option was missing (I think we just forgot to pass it on). Without it the MCP server would be scoped to the account-level.

apps/studio/pages/api/ai/sql/supabase-mcp.ts
```diff
@@ -23,13 +23,14 @@ export async function createSupabaseMCPClient({
...
     readOnly: true,
   })
   await server.connect(serverTransport)
 
   // Create the MCP client and connect to its transport
   const client = await createMCPClient({
-    name: 'Studio AI Assistant', // Give a more specific name
+    name: 'supabase-studio',
```
gregnr, line 33: This my proposed client name based on names I've seen elsewhere

SaxonF (approved):

PR #35541, opened 2025-05-07
fix: ai assistant markdown formatting

gregnr:
Switches the rendered LLM content in `<ReactMarkdown>` to `prose` which renders lists and inline code better:

| Before | After |
|--------|-------|
|image|image|
|image|image|

I also added back left margin to `<QueryBlock>` to make it align better in the chat.

| Before | After |
|--------|-------|
|image|image|

I realizes this was probably intentional (maybe to maximize real estate) so let me know if you disagree. My personal opinion is that alignment is more important than the small amount of extra width, but happy to be wrong.


SaxonF (approved):
Alignment was intention but not strongly opinionated about it. Lets go with aligning to text.

PR #35543, opened 2025-05-08
Fix query block read only catch

joshenlim:
## Context

This is related to a [previous change](https://github.com/supabase/supabase/pull/34892) whereby the QueryBlock (which is used in the AI Assistant) now uses the read only connection string first when executing a query in order to ensure that no mutation related queries are ran without a warning first (previously we were depending on manually checking on the client side which wasn't a scalable way)

To check if the query didn't run because it was mutation related, we only checked the error message for "permission denied", but in this case "must be owner ..." is also a valid error that's related, so adding that check as well as the fix

image

The expected UI response will thus be this to show the warning:
image

### DIFF
--- a/apps/studio/components/ui/QueryBlock/QueryBlock.tsx
+++ b/apps/studio/components/ui/QueryBlock/QueryBlock.tsx
@@ -194,8 +194,6 @@ export const QueryBlock = ({
   const handleExecute = () => {
     if (!sql || isLoading) return
 
-    console.log('Handle execute')
-
     if (readOnlyError) {
       return setShowWarning('hasWriteOperation')
     }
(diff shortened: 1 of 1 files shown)


apps/studio/components/ui/QueryBlock/QueryBlock.tsx
```diff
@@ -189,6 +194,8 @@ export const QueryBlock = ({
   const handleExecute = () => {
     if (!sql || isLoading) return
 
+    console.log('Handle execute')
+
```
alaister, line 198:
```suggestion
```
😉

alaister (approved):
LGTM!

- Tested on preview

PR #35545, opened 2025-05-08
chore: Only fetch overdue invoices if the user is logged in

ivasilov:
This PR changes the `useOverdueInvoicesQuery` to be run only if a user profile is present (the user is logged in).

I saw the API call being made while on the `/sign-in` page.


kevcodez (approved):

alaister (approved):
LGTM!

- Tested on preview

PR #35546, opened 2025-05-08
chore: add signed out redirect timeout

alaister:
If `isFinishedLoading` gets stuck for an unknown reason, the user will be stuck seeing a loading state.

- Shows a modal prompting them to login again after 10 seconds if the loading of auth state hasn't finished
- finished = false is hardcoded on staging, so you should be able to see this after 10s of waiting 

screenshot-2025-06-20-at-10 27 28


apps/studio/hooks/misc/withAuth.tsx
```diff
@@ -46,30 +49,83 @@ export function withAuth<T>(
...
     })
 
+    const signOut = useSignOut()
+
     const isLoggedIn = Boolean(session)
-    const isFinishedLoading = !isLoading && !isAALLoading
+    //const isFinishedLoading = !isLoading && !isAALLoading
+    // for testing on staging
+    // remove before merging
+    const isFinishedLoading = false
```
saltcod: Reminder to remove this before merging!

joshenlim (approved):
Verified changes and have reverted the staging testing specific logic

@saltcod will let you double check once more the code changes, then feel free to merge 🙂 🙏

joshenlim (approved):
Verified changes locally by swapping selected keys in API docs page

[Edit] Ignore this 🤦  i left a comment on the wrong PR hahaha

saltcod (changes requested):
Requesting changes as a reminder to remove the hardcoded:

`const isFinishedLoading = false`

saltcod (approved):

PR #31071, opened 2024-12-11
Feat/tabs

MildTomato:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## All Tabs functionality is behind a team feature flag.

There are some issues outstanding that need to be addressed before moving to feature preview release

### Tab preview

Tabs are currently a opt in feature, if opted in via Feature Preview.

- Tabs
  - tabs opened are stored in localstorage with valtio
  - tabs become 'permanent' when the tab is interacted with
  - tabs are replaceable before they are dirty/touched (like vscode)
- empty state tab
  - includes 'recent items' list
  - includes easy actions to 'create table' and 'create query' 
    - we can add more actions later
  - plus button action: 
    - clicking plus button/new tab in table editor also opens this view
    - clicking plus button/new tab in SQL editor opens a new SQL editor directly
- Reorder
  - Tabs can be re-ordered. drag and drop

PR #35549, opened 2025-05-08
Set up to default opt in tabs interface

joshenlim:
## Context

We've previously introduced tab interfaces to the Table Editor and SQL Editor behind a feature preview [here](https://github.com/supabase/supabase/pull/31071), and have been monitoring and patching any issues reported since then. There hasn't been any new bug reports since the latest fixes last week, so reckon it's time that we incrementally increase the adoption of this UI via a default opt-in behaviour, and eventually make this the default interface.

This will happen from next week onwards as we do a percentage-based random sampling, and we'll slowly increase the percentage value depending on the severity + frequency of issues reported 🙏 

## Changes involved
- Set up feature preview to be default opt in
- Add a UI indication on both SQL and Table Editors to inform the user of the new interface + add a link for them to find where to disable the tabs interface if they might want to
image


alaister (approved):
LGTM!

- Tested on preview

joshenlim:
Adding a callout in the NewTab UI
image

PR #35550, opened 2025-05-08
Add integrations to command k

joshenlim:
## Context

Add Integrations links to our Command K - we've seen a couple of feedbacks about users not knowing that we've got a dedicated UI for managing Cron (and on that note, not realising that Cron is under the integrations section), so hopefully this helps to ease finding things a bit more

image


jordienr (approved):
LGTM

PR #35551, opened 2025-05-08
chore: refactor tabs store

alaister:
## Changes
- refactors the tabs + recent items stores to use the context pattern
- combines the two stores to prevent circular imports


apps/studio/components/interfaces/App/FeaturePreview/FeaturePreviewModal.tsx
```diff
@@ -47,13 +45,6 @@ const FeaturePreviewModal = () => {
       properties: { feature: selectedFeatureKey },
       groups: { project: ref ?? 'Unknown', organization: org?.slug ?? 'Unknown' },
     })
-
-    if (selectedFeatureKey === LOCAL_STORAGE_KEYS.UI_TABLE_EDITOR_TABS) {
-      removeTabsByEditor(ref as string | undefined, 'table')
-    }
-    if (selectedFeatureKey === LOCAL_STORAGE_KEYS.UI_SQL_EDITOR_TABS) {
-      removeTabsByEditor(ref as string | undefined, 'sql')
-    }
```
alaister, line 56: I've removed this as we're removing the feature flag anyway and it would be a pain to implement

joshenlim: hmm JFYI we aren't removing the feature preview at least not within the next few weeks though as we're doing an incremental default opt in (which means users can still disable the tabs interface if they choose to)

as long as we consider this use case when users are disabling the tabs feature preview

joshenlim (approved):
Generally LGTM! did a smoke test and all functionalities are still the same, just a question in the comment above

apps/studio/components/interfaces/App/FeaturePreview/FeaturePreviewModal.tsx
joshenlim: > remove tabs when they disable feature flag

yup! that works

joshenlim (approved):
Checked through changes again to verify comments, LGTM!

PR #35562, opened 2025-05-08
Add custom error handler around fetch method

joshenlim:
## Context

This happens rarely, but in the event that for some reason the `fetch` call fails (likely due to a network issue on the user's end), the default error that comes back from the `fetch` API is just "Failed to fetch" which is often confusing from the user's POV especially when it floats up as a toast message as such:

image

## Changes involved

Opting to override the default error message with the following message that's a bit more user friendly and active

image

- Added `fetchHandler` in `data/fetchers` to handle this custom error message
- Use `fetchHandler` in most instances where we're calling `fetch` directly
  - This is more important for mutation queries since we usually float the errors up as toast messages, the others based on observation are relatively safe to fail silently
- (Unrelated) Refactors EdgeFunctionTesterSheet to call POST /test via react query than manually calling `fetch`

### DIFF
--- a/apps/studio/data/ai/check-api-key-query.ts
+++ b/apps/studio/data/ai/check-api-key-query.ts
@@ -1,7 +1,7 @@
 import { useQuery, UseQueryOptions } from '@tanstack/react-query'
 
-import { constructHeaders } from 'data/fetchers'
-import { BASE_PATH, IS_PLATFORM } from 'lib/constants'
+import { constructHeaders, fetchHandler } from 'data/fetchers'
+import { BASE_PATH } from 'lib/constants'
 import { ResponseError } from 'types'
 import { aiKeys } from './keys'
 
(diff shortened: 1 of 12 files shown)


apps/studio/data/ai/check-api-key-query.ts
```diff
@@ -34,5 +37,5 @@ export const useCheckOpenAIKeyQuery = <TData = ResourceData>({
   useQuery<ResourceData, ResourceError, TData>(
     aiKeys.apiKey(),
     ({ signal }) => checkOpenAIKey(signal),
-    { enabled: !IS_PLATFORM && enabled, ...options }
+    { enabled: enabled, ...options } // !IS_PLATFORM &&
```
ivasilov, line 40: The `!IS_PLATFORM` is intentional to guide the user into adding their own OpenAI API key when selfhosting/CLI.

joshenlim: oh crap - yup! need to revert this one 🙏 🙏

alaister (approved):
LGTM!

- Tested on preview

PR #35566, opened 2025-05-08
Chore/replace auth policies rls warnings with security lints

joshenlim:
## Context

There was a confusion raised with the warning label shown in the Auth -> Policies page when a table has RLS disabled, that the table was publicly readable and writable. This warning persists even if the schema of the table is not being exposed through the API (Under Settings -> Data API), which understandably causes unnecessary concern since the table wouldn't be publicly readable or writable in this case specifically
image

## Changes involved

- The RLS warning in the Auth -> Policies page was showing as long as the table had RLS disabled, so we're going to move to using the lints from the security advisor as the source of truth
image
- That gives us a few other benefits
  - We don't need to manually check if the schema is exposed through the API since that logic is in the lint itself (e.g below, schema joshen_test is not exposed)
image
  - We can also show the new "Info" label for when RLS is enabled but there's no policies (which i reckon is a valid callout)

## To test
- [ ] Verify that the warning label shows if a table under the public schema has RLS disabled
- [ ] Verify that the info label shows if a table has RLS enabled but no policies
- [ ] Verify that the warning label updates to info if you enable RLS on a table in the public schema with no policies
- [ ] Verify that the info label goes away if you create a policy for a table that had RLS enabled and had no policies
- [ ] Verify that the info label shows up if you delete the only policy on a table that has RLS enabled

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/Policies.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/Policies.tsx
@@ -1,4 +1,5 @@
 import type { PostgresPolicy } from '@supabase/postgres-meta'
+import { useQueryClient } from '@tanstack/react-query'
 import { isEmpty } from 'lodash'
 import { HelpCircle } from 'lucide-react'
 import { useRouter } from 'next/router'
@@ -16,6 +17,7 @@ import NoSearchResults from 'components/to-be-cleaned/NoSearchResults'
 import ProductEmptyState from 'components/to-be-cleaned/ProductEmptyState'
 import InformationBox from 'components/ui/InformationBox'
 import { useDatabasePolicyDeleteMutation } from 'data/database-policies/database-policy-delete-mutation'
+import { lintKeys } from 'data/lint/keys'
 import { useTableUpdateMutation } from 'data/tables/table-update-mutation'
 import ConfirmModal from 'ui-patterns/Dialogs/ConfirmDialog'
 
@@ -38,6 +40,7 @@ const Policies = ({
 }: PoliciesProps) => {
   const router = useRouter()
   const { ref } = useParams()
+  const queryClient = useQueryClient()
   const { project } = useProjectContext()
 
   const [selectedTableToToggleRLS, setSelectedTableToToggleRLS] = useState<{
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Auth/Policies/Policies.tsx
```diff
@@ -57,8 +60,9 @@ const Policies = ({
     },
   })
   const { mutate: deleteDatabasePolicy } = useDatabasePolicyDeleteMutation({
-    onSuccess: () => {
+    onSuccess: async () => {
       toast.success('Successfully deleted policy!')
+      await queryClient.invalidateQueries(lintKeys.lint(ref))
```
ivasilov, line 65: Should we move this to `useDatabasePolicyDeleteMutation` definition? It should always invalidate the lint rules, right?

apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/index.tsx
```diff
@@ -47,6 +50,15 @@ export const PolicyTableRow = ({
     .filter((policy) => policy.schema === table.schema && policy.table === table.name)
     .sort((a, b) => a.name.localeCompare(b.name))
 
+  const { data: lints = [] } = useProjectLintsQuery({ projectRef: project?.ref })
+  const [rlsLint] = lints.filter(
+    (lint) =>
+      lint.categories.includes('SECURITY') &&
+      lint.metadata?.name === table.name &&
+      lint.metadata?.schema === table.schema &&
+      (lint.name === 'rls_enabled_no_policy' || lint.name.includes('rls_disabled'))
```
ivasilov, line 59: `lint.name.includes('rls_disabled')` may not be forward-compatible. We could add a new rule with `rls_disabled` which means something else.

apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/index.tsx
```diff
@@ -59,27 +71,37 @@ export const PolicyTableRow = ({
...
+              'w-1.5 h-1.5 rounded-full',
+              rlsLint.level === 'ERROR' ? 'bg-warning-600' : 'bg-selection'
+            )}
+          />
+          <span className={cn('font-bold', rlsLint.level === 'ERROR' ? 'text-warning-600' : '')}>
+            {rlsLint.level === 'ERROR' ? 'Warning' : 'Note'}:
+          </span>{' '}
+          <Markdown
+            className="text-xs text-foreground-light !max-w-full"
+            content={`${rlsLint.detail}${rlsLint.level === 'ERROR' ? ' Your table is publicly readable and writable.' : '. No data will be selectable via Supabase APIs.'}`}
```
ivasilov, line 92: the if condition `rlsLint.level === 'ERROR'` should be about the specific rls rule, not the level.

joshenlim:
Closing in favor of https://github.com/supabase/supabase/pull/35579

PR #35579, opened 2025-05-09
Improve auth policies RLS warnings granularity

joshenlim:
Replaces this [PR](https://github.com/supabase/supabase/pull/35566), will shift the context over

## Context
There was a confusion raised with the warning label shown in the Auth -> Policies page when a table has RLS disabled, that the table was publicly readable and writable. This warning persists even if the schema of the table is not being exposed through the API (Under Settings -> Data API), which understandably causes unnecessary concern since the table wouldn't be publicly readable or writable in this case specifically. Also persists if the `anon` and `authenticated` roles have their privileges revoked on that table
image

## Changes involved

- The RLS warning in the Auth -> Policies page was showing as long as the table had RLS disabled, so we're improving the granularity of it
image
- Only show warning that table is publicly readable / writable if RLS is disabled and
  - Table's schema is exposed via the API
  - `anon` or `authenticated` roles have access to that table
- Show "Note" label if RLS is enabled but no policies have been created
- I left a comment, but highlighting here as well - ideal scenario is that we can use the security lints so that the logic is all consistent

## To test
- [ ] Verify that the warning label shows if a table under the public schema has RLS disabled
- [ ] Verify that the info label shows if a table has RLS enabled but no policies
- [ ] Verify that the warning label updates to info if you enable RLS on a table in the public schema with no policies
- [ ] Verify that the info label goes away if you create a policy for a table that had RLS enabled and had no policies
- [ ] Verify that the info label shows up if you delete the only policy on a table that has RLS enabled

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/index.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/index.tsx
@@ -5,7 +5,9 @@ import { Info } from 'lucide-react'
 import { useProjectContext } from 'components/layouts/ProjectLayout/ProjectContext'
 import AlertError from 'components/ui/AlertError'
 import Panel from 'components/ui/Panel'
+import { useProjectPostgrestConfigQuery } from 'data/config/project-postgrest-config-query'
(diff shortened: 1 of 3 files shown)


alaister (approved):
LGTM!

PR #35580, opened 2025-05-09
add advisor and stats to home

SaxonF:
https://github.com/user-attachments/assets/c564510c-75d7-4ce9-886d-a46d77606bdb

- Adds an advisor widget to the homepage that includes performance, security and slow queries
- Adds table, function and replication count to home


joshenlim:
Will push up changes for the following
- "Replicas 1" -> this shows up when I dont have any replicas, so I'd expect it to be 0. The read replicas query also includes the primary database, so need to discount that haha
- Can probably deprecate `<SecurityStatus>` component in favor of AdvisorWidget
- (Unrelated) Noticed that we're still showing the NewProjectPanel even if the project is older than 2 days
  - Fixing by removing the check for `hasProjectData`

joshenlim:
Adding a link to the advisor widget to link to the corresponding advisors
image

apps/studio/components/interfaces/Home/AdvisorWidget.tsx
```diff
@@ -0,0 +1,328 @@
...
+                <TableBody>
+                  {/* Added explicit types for map parameters */}
+                  {top5SlowestQueries.map((query: SlowQuery, i: number) => (
+                    <TableRow key={i}>
+                      <TableCell className="font-mono truncate max-w-xs py-2">
+                        {query.query}
+                      </TableCell>
+
+                      <TableCell className="font-mono truncate max-w-xs py-2">
+                        {typeof query.avg_time === 'number'
```
joshenlim, line 311: there's no `avg_time` parameter btw - will swap this with `mean_time`

joshenlim, line 311: there's no `avg_time` parameter btw - will swap this with `mean_time`

apps/studio/components/interfaces/Home/AdvisorWidget.tsx
```diff
@@ -0,0 +1,328 @@
...
+                          ? `${query.avg_time.toFixed(2)}ms`
```
joshenlim, line 312: opting to change this to seconds actually - there was some feedback previously that it was easier to glance in seconds as the numbers for ms were too large haha

apps/studio/components/interfaces/Home/AdvisorWidget.tsx
```diff
@@ -0,0 +1,328 @@
...
+    title: string,
+    lints: Lint[],
+    errorCount: number,
+    warningCount: number,
+    isLoading: boolean
+  ) => {
+    const topIssues = lints
+      .filter((lint) => lint.level === LINTER_LEVELS.ERROR || lint.level === LINTER_LEVELS.WARN)
+      .sort((a, b) => (a.level === LINTER_LEVELS.ERROR ? -1 : 1))
+      .slice(0, 5)
```
joshenlim, line 115: hmm any reason why we're only choosing the first 5? reckon we can just show all tbh

apps/studio/components/interfaces/Home/AdvisorWidget.tsx
```diff
@@ -0,0 +1,328 @@
...
+        {!isLoading && (errorCount > 0 || warningCount > 0) && (
+          <ul>
+            {topIssues.map((lint) => (
+              <li
+                key={lint.cache_key}
+                className="text-sm px-4 py-3 w-full border-b my-0 last:border-b-0 group"
+              >
+                <div className="flex items-center justify-between w-full">
+                  <Link
+                    href={`/project/${projectRef}/advisors/${title.toLowerCase()}?id=${lint.cache_key}`}
```
joshenlim, line 135: should add `preset` to the URL here so it opens the right tab when navigating

joshenlim:
ahh one sec - needs some fixes for paused state

joshenlim (approved):
All LGTM now! Verified on preview 🙂

PR #35583, opened 2025-05-09
fix: move extension requests to query route

sweatybridge:

refactor

## What is the new behavior?

- supports cascade param when deleting extensions
- combine schema creation in one query
- mirrors exactly the [upstream](https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaExtensions.ts#L52) implementation

## Additional context


jgoux (approved):
lgtm 🚀

alaister (approved):
LGTM!

- Tested on preview

PR #35589, opened 2025-05-09
Split index advisor hook into installed and enabled instead of just a…

saltcod:
We need to know if index_advisor (and its dependency, hypopg) are installed. 
But we also need to know if they are _available_ to be installed. 

This PR renames useIsIndexAdvisorAvailable -> useIsIndexAdvisorAvailable which exposes these two states separately 

`{ isIndexAdvisorAvailable, isIndexAdvisorEnabled }`

- hook is renamed accordingly 
- affected files are updated to use the new name
- now, the EnableIndexAdvisorButton  will only render after checking both states (not already enabled, and can be enabled)


apps/studio/components/interfaces/QueryPerformance/hooks/useIsIndexAdvisorStatus.ts
```diff
@@ -3,16 +3,27 @@ import { useDatabaseExtensionsQuery } from 'data/database-extensions/database-ex
...
   })
 
   const { hypopg, indexAdvisor } = getIndexAdvisorExtensions(extensions ?? [])
-  return hypopg?.installed_version !== null && indexAdvisor?.installed_version !== null
+
+  const isIndexAdvisorAvailable = !!hypopg && !!indexAdvisor
+
+  const isIndexAdvisorEnabled =
+    !!hypopg &&
+    !!indexAdvisor &&
```
joshenlim, line 24: probably can just be `isIndexAdvisorAvailable`?

apps/studio/components/interfaces/QueryPerformance/index-advisor.utils.ts
```diff
@@ -7,12 +7,13 @@ import { executeSql } from 'data/sql/execute-sql-query'
...
  * Gets the required extensions for index advisor
  * @param extensions Array of database extensions
- * @returns Object containing hypopg and index_advisor extensions if they exist
+ * @returns Object containing hypopg, index_advisor, and test_extension extensions if they exist
  */
 export function getIndexAdvisorExtensions(extensions: DatabaseExtension[] = []) {
   const hypopg = extensions.find((ext) => ext.name === 'hypopg')
   const indexAdvisor = extensions.find((ext) => ext.name === 'index_advisor')
-  return { hypopg, indexAdvisor }
+  const test_extension = extensions.find((ext) => ext.name === 'test_extension')
```
joshenlim, line 15: is there a new extension called test_extension?

joshenlim: this looks unintentional as there's no usage of `test_extension` when calling this function, will remove

apps/studio/pages/project/[ref]/advisors/query-performance.tsx
```diff
@@ -36,7 +36,7 @@ const QueryPerformanceReport: NextPageWithLayout = () => {
   })
 
   const config = PRESET_CONFIG[Presets.QUERY_PERFORMANCE]
-  const hooks = queriesFactory(config.queries, projectRef ?? 'default')
+  const hooks = queriesFactory(config.queries, project?.ref ?? 'default')
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/advisors/query-performance.tsx:39]: any reason why we can't use ref from params? its faster than depending on useProjectContext

Comment (joshenlim) [reply apps/studio/pages/project/[ref]/advisors/query-performance.tsx:39]: i'll revert for now but lemme know if i might be overlooking something here

apps/studio/components/interfaces/QueryPerformance/EnableIndexAdvisorButton.tsx
```diff
@@ -70,7 +72,8 @@ export const EnableIndexAdvisorButton = () => {
     }
   }
 
-  if (isAdvisorAvailable) return null
+  // if index_advisor is already enabled or not available to install, show nothing
+  if (isIndexAdvisorEnabled || !isIndexAdvisorAvailable) return null
```
joshenlim, line 76: hmm technically don't need to check for `isIndexAdvisorAvailable`

since `isIndexAdvisorEnabled` already innately includes the check for `isIndexAdvisorAvailable`

apps/studio/components/interfaces/QueryPerformance/hooks/useIsIndexAdvisorStatus.ts
```diff
@@ -3,16 +3,27 @@ import { useDatabaseExtensionsQuery } from 'data/database-extensions/database-ex
...
+    hypopg.installed_version !== null &&
+    indexAdvisor.installed_version !== null
+
+  return { isIndexAdvisorAvailable, isIndexAdvisorEnabled }
```
joshenlim, line 28: nit but we probably don't need to return `isIndexAdvisorAvailable` actually since we're not using it atm, and `isIndexAdvisorEnabled` already considers it too

joshenlim: unless we'll be rendering a specific UI with the `isIndexAdvisorAvailable` param

joshenlim (approved):
Changes LGTM! Can verify no user-facing changes

PR #35617, opened 2025-05-12
chore(refacto): throw error instances

avallete:
- Throw Error instances to keep stacktrace when error occurs

### DIFF
--- a/apps/studio/data/fetchers.ts
+++ b/apps/studio/data/fetchers.ts
@@ -63,13 +63,11 @@ function pgMetaGuard(request: Request) {
     // If there is no valid `x-connection-encrypted`, pg-meta will necesseraly fail to connect to the target database
     // in such case, we save the hops and throw a 421 response instead
     if (!isValidConnString(request.headers.get('x-connection-encrypted'))) {
-      // TODO: Maybe here add a sentry warning to monitor how often this happen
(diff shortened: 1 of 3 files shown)


apps/studio/data/fetchers.ts
```diff
@@ -63,13 +63,11 @@ function pgMetaGuard(request: Request) {
...
-      // TODO: Maybe here add a sentry warning to monitor how often this happen
-      // Simulate a 421 response by throwing an error
-      throw {
-        code: 421,
-        message: 'x-connection-encrypted header is required for /platform/pg-meta/ requests',
-        requestId: request.headers.get('X-Request-Id'),
-      }
+      throw new ResponseError(
+        'x-connection-encrypted header is required for /platform/pg-meta/ requests',
+        400,
```
avallete, line 68: **note**

Turned 421 to 400, 421 is misredirect which doesn't seems appropriate. Use the generic "Bad Request" instead.

kamilogorek (approved):

alaister (approved):

apps/studio/data/fetchers.ts
sweatybridge: The previous code threw an object but the new code threw an exception. This might be causing elevated error due to different error handling path?

PR #35629, opened 2025-05-12
fix(docs): crawlers page generation for pages with mdx components

charislam:
# Before

Crawlers pages that used MDX components with defined attributes errored. Example: /docs/reference/kotlin/installing

# After

The pages do not error.

### DIFF
--- a/apps/docs/app/api/crawlers/route.ts
+++ b/apps/docs/app/api/crawlers/route.ts
@@ -157,9 +157,7 @@ async function functionDetails(
   return fullDescription + parameters + examples
 }
 
-function mdxToHtml(markdownUnescaped: string): string {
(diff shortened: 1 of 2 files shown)


apps/docs/app/api/crawlers/route.ts
```diff
@@ -157,9 +157,7 @@ async function functionDetails(
   return fullDescription + parameters + examples
 }
 
-function mdxToHtml(markdownUnescaped: string): string {
-  const markdown = markdownUnescaped.replace(/(?<!\\)\{/g, '\\{').replace(/(?<!\\)\}/g, '\\}')
```
charislam, line 161: This was a hack for some previous reason where we had Markdown files with braces that couldn't be removed. I can't remember why they couldn't be removed at the time, but I searched through the JS reference page and couldn't find those exceptions anymore, so we should be safe to remove this, since it causes a different problem with MDX parsing.

saltcod (approved):

PR #35633, opened 2025-05-13
fix: move publication create and update to query route

sweatybridge:

refactor

## What is the new behavior?

- Use the query route for creating and updating publications
- Mirrors exactly the implementation of [pgmeta](https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaPublications.ts#L78) route

## Additional context


alaister (approved):
LGTM!

- Tested on preview

PR #35643, opened 2025-05-13
fix(docs): update callback url to use token_hash instead of token

### LINKED ISSUE #35592: Incorrect code in email auth hook docs
# Improve documentation

## Link 

https://supabase.com/docs/guides/auth/auth-hooks/send-email-hook?queryGroups=language&language=http

## Describe the problem

In the i81n example, the magiclink that's generated is using a query string of `?token=${email_data.token}`, when it should be `${email_data.token_hash}`. The existing example in the docs does not work.

## Additional context

See the `generateConfirmationURL` function in the `Add Internationalization for Email Templates` tab.

silentworks:

YES

Docs update

The generated callback URL is using `token` which is incorrect.

## What is the new behavior?

The generated callback URL is using `token_hash` which is correct.

## Additional context

Fix for #35592

### DIFF
--- a/apps/docs/content/guides/auth/auth-hooks/send-email-hook.mdx
+++ b/apps/docs/content/guides/auth/auth-hooks/send-email-hook.mdx
@@ -643,7 +643,7 @@ const templates = {
 
 function generateConfirmationURL(email_data) {
   // TODO: replace the ref with your project ref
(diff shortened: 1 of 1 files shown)


charislam (approved):
Thanks @silentworks!

PR #35662, opened 2025-05-14
fix: move table create update delete to query route

sweatybridge:

refactor

- mirrors implementation of [pgmeta](https://github.com/supabase/postgres-meta/blob/master/src/lib/PostgresMetaTables.ts#L122) exactly
- adjusts update query to fetch primary keys implicitly

## What is the new behavior?

## Additional context


packages/pg-meta/src/pg-meta-tables.ts
```diff
@@ -196,103 +178,76 @@ function update(
...
-  INTO v_table
-  FROM tables
-  WHERE ${getIdentifierWhereClause(identifier)};
-
-  IF v_table IS NULL THEN
-    ${exceptionIdentifierNotFound('table', getIdentifierWhereClause(identifier))}
+  SELECT conname
+    INTO r
+    FROM pg_constraint
+    WHERE contype = 'p' AND conrelid = ${literal(old.id)};
```
avallete, line 220: **question**

If we wanna get rid of the `id` dependencies for migrations tracking should this be replaced with a select subquery to get the table id from `name+schema` ?

avallete (approved):

packages/pg-meta/src/pg-meta-tables.ts
sweatybridge: Currently the UI relies on table id in path param for navigation. It will be quite complicated to move away from it.

alaister (commented):
I get the following error when trying to add a column in the table edit ui:
Screenshot 2025-05-14 at 15 55 58

I also got an error the first time I tried creating a table, but I can't seem to replicate that again :/

saltcod:
renaming a table gives me this:
screenshot-2025-05-14-at-11 04 48

saltcod:
screenshot-2025-05-14-at-11 06 32

saltcod:
disabling RLS fails with the same error as the rest 

screenshot-2025-05-14-at-11 07 21

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
```diff
@@ -670,14 +685,28 @@ export const updateTable = async ({
...
 
+  const updatedTable = await queryClient.fetchQuery({
+    queryKey: tableKeys.retrieve(
+      projectRef,
+      payload.name ?? table.name,
+      payload.schema ?? table.schema
+    ),
+    queryFn: ({ signal }) =>
+      getTable(
+        { projectRef, connectionString, name: payload.name, schema: payload.schema },
```
sweatybridge, line 705: This needs a ternary as well

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
```diff
@@ -126,7 +127,7 @@ export const generateUpdateColumnPayload = (
   const type = field.isArray ? `${field.format}[]` : field.format
   const comment = ((field.comment?.length ?? '') === 0 ? '' : field.comment)?.trim()
 
-  const check = field.check?.trim()
+  const check = field.check?.trim() ? field.check?.trim() : undefined
```
sweatybridge, line 130: what is this supposed to do?

sweatybridge: Based on https://github.com/supabase/supabase/blob/master/packages/pg-meta/src/pg-meta-columns.ts#L341, we should set empty string to null.

```
const check = field.check?.trim() || null
```

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
```diff
@@ -126,7 +127,7 @@ export const generateUpdateColumnPayload = (
   const type = field.isArray ? `${field.format}[]` : field.format
   const comment = ((field.comment?.length ?? '') === 0 ? '' : field.comment)?.trim()
```
sweatybridge, line 128:
```suggestion
  const comment = field.comment?.trim() || null
```

ditto

saltcod:
Tested every table action I could think of and looks good on staging!

saltcod (approved):

PR #35666, opened 2025-05-14
Minor fix database trigger header

joshenlim:
## Fixes involved
- Database Trigger sheet: Update to say "Edit database trigger" instead of "Create a new database trigger" when editing a trigger
- Add guard for TriggerSheet if tables.length is 0, and also add tooltip if user has no tables in non-protected schemas
- (Unrelated) Fix alignment of badges in project home page


jordienr (approved):
tested in preview lgtm!

PR #35670, opened 2025-05-14
Remove unnecessary includeSystemSchemas, stay consistent with hosted — fixes custom schema enums

joshenlim:
Just needed removing `includeSystemSchemas` from the local API endpoint, this part was added 3 years ago hahaha
Removing it to stay consistent with the hosted, which we don't use `includeSystemSchemas` anywhere anymore

## To test (Must be tested via self-host/local)
- Create enum type in custom schema
- Can view enum in database/enums
- Can view enum in table editor when creating a column -> column type
- (Unrelated) Adds error handling for database trigger sheet when a function is not selected

Fixes FE-1643


saltcod (approved):
Created a custom schema in self-hosted and enums appear! great success!

PR #35674, opened 2025-05-14
fix: Minor fixes for the Cron integration

ivasilov:
This PR fixes 3 bugs in the Cron integration:
- If the cron job calls an Edge Function with a query param or hash, the command wasn't processed properly and the Edge Function was shown in the `CreateCronJobSheet`.
- The cron job name is optional in `pg_cron`. Cron jobs without names cannot be edited in SQL, now they can't be edited in UI also.
- Previous runs for cron jobs without names weren't working at all, this is now fixed.
Screenshot 2025-05-14 at 15 42 02
Screenshot 2025-05-14 at 15 42 15

### DIFF
--- a/apps/studio/components/interfaces/Integrations/CronJobs/CronJobCard.tsx
+++ b/apps/studio/components/interfaces/Integrations/CronJobs/CronJobCard.tsx
@@ -22,6 +22,9 @@ import {
   DropdownMenuTrigger,
   Label_Shadcn_,
   Switch,
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
 } from 'ui'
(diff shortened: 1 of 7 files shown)


joshenlim:
i created a cron job with no name but i'm not seeing the placeholder i the cron job card
image

```
select cron.schedule('', '*/5 * * * *', ...);
```

joshenlim:
Delete modal for no name cron job could probably be simplified to a ConfirmationModal without the text validation
image

joshenlim:
Running into a client error when opening HIstory for a cron with no name
image
image

joshenlim (approved):
Verified that identified issues are now resolved 🙂 🙏

PR #35686, opened 2025-05-15
Add regex guard for edge functions test endpoint

joshenlim:
Fixes SEC-365

Validates that provided URL parameter is only a Supabase edge function URL

- Tested working normally on the dashboard edge functions page
- Tested curling directly with an invalid URL returns a 400

### DIFF
--- a/apps/studio/pages/api/edge-functions/test.ts
+++ b/apps/studio/pages/api/edge-functions/test.ts
@@ -19,7 +19,21 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
 
 async function handlePost(req: NextApiRequest, res: NextApiResponse) {
   try {
+    console.log('post', req.body)
     const { url, method, body: requestBody, headers: customHeaders } = req.body
+
+    const regexValidEdgeFunctionURL = new RegExp(
+      'https://[a-z]*.supabase.(red|co)/functions/v[0-9]{1}/S*',
+      'g'
+    )
+    const validEdgeFunctionURL = regexValidEdgeFunctionURL.test(url)
+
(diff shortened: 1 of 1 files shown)


apps/studio/pages/api/edge-functions/test.ts
```diff
@@ -21,6 +21,19 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse) {
   try {
     const { url, method, body: requestBody, headers: customHeaders } = req.body
 
+    const regexValidEdgeFunctionURL = new RegExp(
+      'https://[a-z]*.supabase.(red|co)/functions/v[0-9]{1}/S*',
```
staaldraad, line 25:
```suggestion
      '^https://[a-z]*.supabase.(red|co)/functions/v[0-9]{1}/\S*$',
```

Anchoring the regex to the start of the string, otherwise it leaves room for something like `https://localhost/?https://aaaa.supabase.co/functions/v1/xxx`

And I think the `/S*` should be `/\S*` to capture the intent Any non-whitespace character, otherwise it is matching the character `S` zero or more times 

For good measure we can have the end anchor in there as well (RegExp probably also has match the whole string, rather than test - can't recall)

staaldraad (changes requested):

apps/studio/pages/api/edge-functions/test.ts
joshenlim: awesome thank you so much for checking!! 🙏 🙏

joshenlim:
Closing in favor of https://github.com/supabase/supabase/pull/35688/files

PR #35687, opened 2025-05-15
Chore/refactor derivate more types from queries

avallete:
- It allows to decouple more the frontend logic and the pg-meta/sql-query logic allowing to reduce the number of cast
and get closer types between what we do fetch and what we expect in our components

The idea is to "cast" and "parse / validate" the result of the 3rd party (pg-meta call) in a single point into the frontend app (after the fetching). Then, all the frontend types within component should be derivated from this function fetcher call. It will allows to reduce casting, get closer match between data retrieved and data expected in components, and will show what exactly will be in error within the frontend if the result type of the query fetch were to be changed.

Basically:

3rd parties (sql queries) -> React app fetching (parse/validate result type) -> ResultType - | -> Frontend-components

### DIFF
--- a/packages/pg-meta/src/pg-meta-tables.ts
+++ b/packages/pg-meta/src/pg-meta-tables.ts
@@ -1,5 +1,5 @@
 import { z } from 'zod'
-import { coalesceRowsToArray, exceptionIdentifierNotFound, filterByList } from './helpers'
+import { coalesceRowsToArray, filterByList } from './helpers'
 import { TABLES_SQL } from './sql/tables'
 import { COLUMNS_SQL } from './sql/columns'
 import { DEFAULT_SYSTEM_SCHEMAS } from './constants'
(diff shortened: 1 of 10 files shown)


packages/pg-meta/src/pg-meta-tables.ts
```diff
@@ -107,14 +107,14 @@ function list<T extends boolean | undefined = true>(
...
-  zod: z.ZodType<PGTable | undefined>
+  zod: z.ZodType<TableWithColumns>
 } {
   let whereClause = getIdentifierWhereClause(identifier)
 
   const sql = `${generateEnrichedTablesSql({ includeColumns: true })} where ${whereClause};`
   return {
     sql,
-    zod: pgTableZod.optional(),
+    zod: pgTableZod,
```
sweatybridge, line 117:
```suggestion
    zod: pgTableZod.optional(),
```

this is still optional because result can be empty array right?

sweatybridge (approved):

PR #35688, opened 2025-05-15
Add regex guard for edge functions test endpoint

jordienr:
Fixes SEC-365

- moves edge fn validation outside the api for testing
- adds tests

### DIFF
--- a/apps/studio/lib/api/edgeFunctions.ts
+++ b/apps/studio/lib/api/edgeFunctions.ts
@@ -1,0 +1,7 @@
+export const isValidEdgeFunctionURL = (url: string) => {
+  const regexValidEdgeFunctionURL = new RegExp(
+    '^https://[a-z]*.supabase.(red|co)/functions/v[0-9]{1}/.*$'
+  )
+
+  return regexValidEdgeFunctionURL.test(url)
+}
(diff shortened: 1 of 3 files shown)


joshenlim (approved):
LGTM!

apps/studio/lib/api/edgeFunctions.ts
```diff
@@ -0,0 +1,7 @@
+export const isValidEdgeFunctionURL = (url: string) => {
+  const regexValidEdgeFunctionURL = new RegExp(
+    '^https://[a-z]*.supabase.(red|co)/functions/v[0-9]{1}/.*$'
```
joshenlim, line 3: just noticed this is slightly different from what @staaldraad added in the previous PR if there's a difference functionally?

'^https://[a-z]*.supabase.(red|co)/functions/v[0-9]{1}/\S*$',

```suggestion
    '^https://[a-z]*.supabase.(red|co)/functions/v[0-9]{1}/.*$'
```

joshenlim: i think they should be the same yeah - tests are passing as expected so no issue

staaldraad (approved):

PR #35690, opened 2025-05-15
fix: edgeFunctions test should not follow redirects

staaldraad:

YES

Bug fix

Follows a redirect sent by the edge function

## What is the new behavior?

Return the edge function response without following any redirects.


joshenlim (approved):
Tested on preview that testing edge function still works as expected 🙂

PR #35698, opened 2025-05-15
fix: allow PITR removal with active read replicas

kevcodez:


alaister (approved):
LGTM!

- Tested on preview

PR #35705, opened 2025-05-16
Fix service checks coming up state

joshenlim:
## Context

Noticed that for new projects, Edge Functions under "Project Status" was stuck in COMING UP, despite the corresponding network request which controls its status returning `{ healthy: true }`

Realised that this was caused by a couple of things:
- `services` [here](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Home/ServiceStatus.tsx#L122) for Edge Functions didn't have a `status` property as per other services in the array which was the main reason why it happened
- `StatusMessage` ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Home/ServiceStatus.tsx#L25)) and `StatusIcon` ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Home/ServiceStatus.tsx#L53)) has checks for both `isSuccess` and `status`, but because `isSuccess` is the last conditional in the list, and Edge Functions has no `status` being passed to these 2 components, it defaulted to `isProjectNew`, which meant that it'll stay as "Coming Up" for the duration of the `SERVICE_STATUS_THRESHOLD` ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Home/ServiceStatus.tsx#L23)) that was set at 5 minutes
image
  - This threshold was just a buffer for if the project is less than 5 minutes old, and status is not operational, then it's likely the service is still starting up (instead of showing that it's unhealthy)

## Changes involved
- Shift check for `isSuccess` to the top for `StatusMessage` and `StatusIcon` - should short circuit as long as the service status is healthy
- Add `status` for edge functions in `services`
- For the Project Status popover button trigger, only show loader if `(!allServicesOperational && isProjectNew)` instead of just if `isProjectNew`
- (Unrelated) Remove unnecessary padding in Delete project section
image


apps/studio/components/interfaces/Home/ServiceStatus.tsx
```diff
@@ -193,6 +193,11 @@ export const ServiceStatus = () => {
             docsUrl: 'https://supabase.com/docs/guides/functions/troubleshooting',
             isLoading,
             isSuccess: edgeFunctionsStatus?.healthy,
+            status: edgeFunctionsStatus?.healthy
```
darora, line 196: any reason not to make `status`, `isSuccess` etc required fields? That would've prevented this afaict

If a service doesn't have data yet, it can fall back to 'unhealthy' / coming up

joshenlim: yup makes sense! will update

joshenlim: pushed an update!

apps/studio/components/interfaces/Home/ServiceStatus.tsx
```diff
@@ -33,12 +33,12 @@ const StatusMessage = ({
   isProjectNew: boolean
   status?: ProjectServiceStatus
 }) => {
+  if (isSuccess) return 'Healthy'
```
ivasilov, line 36: Shouldn't this be
```
  if (isLoading) return 'Checking status'
  if (isSuccess) {
    if (status === 'UNHEALTHY') return 'Unhealthy'
    if (status === 'COMING_UP') return 'Coming up...'
    if (status === 'ACTIVE_HEALTHY') return 'Healthy'
  }
  if (isProjectNew) return 'Coming up...'
  return 'Unable to connect'
```

joshenlim: isSuccess shouldn't be mashed with status imo

i realised this is confusing but `isSuccess` here isn't from react-query, so its not indicative or a successful network request

image

joshenlim: lemme refactor this to change `isSuccess` to `isHealthy`, its very easily confused haha

joshenlim: okay just pushed!

alaister (approved):
LGTM!

- Tested on preview

PR #35709, opened 2025-05-16
Fix unnecessary GET content request when writing a new SQL query

joshenlim:
Fixes FE-1646

## Context

I noticed that whenever we're typing a new query in the SQL Editor on the `/sql/new` route (e.g clicking on the new snippet button), we'd make an unnecessary GET content request which naturally results in a 404 as the snippet is not created just yet

- Upon typing in the code editor in `/sql/new`, we'd redirect to `/sql/[id]` immediately ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/SQLEditor/MonacoEditor.tsx#L174))
- Thereafter, once landing on `/sql/[id]`, the `useContentIdQuery` gets called [here](https://github.com/supabase/supabase/blob/master/apps/studio/pages/project/%5Bref%5D/sql/%5Bid%5D.tsx#L35) which results in the 404

## Changes involved

- Adding a new flag `isNotSavedInDatabaseYet` that defaults to false unless specified otherwise
  - This is specifically set to `true` within `createSqlSnippetSkeletonV2` and will be set to false after the content is saved in `sql-editor-v2` -> `upsertSnippet`
- Pass this flag into the `enabled` param of `useContentIdQuery`
- I've also added some better error handling that's similar to the Table Editor when landing on a snippet ID that doesn't exist
image

## To test

- First verify on staging/prod that there's an unnecessary 404 request once you type in something while on `/sql/new`
- Then verify on this preview that it no longer happens + doesn't have any unintended side effects

### DIFF
--- a/apps/studio/components/interfaces/SQLEditor/SQLEditor.utils.ts
+++ b/apps/studio/components/interfaces/SQLEditor/SQLEditor.utils.ts
@@ -68,6 +68,7 @@ export const createSqlSnippetSkeletonV2 = ({
       content_id: id ?? '',
       sql: sql ?? '',
     } as any,
+    isNotSavedInDatabaseYet: true,
   }
 }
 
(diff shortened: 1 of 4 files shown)


jordienr (approved):
confirmed I get the 404 in prod but not in this branch.

PR #35710, opened 2025-05-16
Fix Table Editor non-existing table state

joshenlim:
Fixes FE-1645

## Context

A couple of quirks when opening the table editor with an ID that doesn't exist in the database (e.g if deleting the table via the SQL editor from another browser session)

- In the event whereby you have no other tables in the database + schema, clicking "Close tab" does nothing and hitting
- Clicking the "+" new tab opens the new tab state, but if you close it, it leads to a 404

## Changes involved

- For a non-existent table, CTA button to handle other scenarios
  - If no other tables, the button will just redirect to `/editor` instead
  - If there are other tabs opened, the button will redirect to the first opened tab
- Don't show tabs UI if there's no opened tabs for table editor
  - Similar handling as per landing on `/editor`

image

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/TableGridEditor.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/TableGridEditor.tsx
@@ -134,6 +134,17 @@ export const TableGridEditor = ({
                   >
                     Close tab
                   </Button>
+                ) : tabs.openTabs.length > 0 ? (
+                  <Button
+                    asChild
+                    type="default"
+                    className="mt-2"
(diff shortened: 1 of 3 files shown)


jordienr:
If you don't have any tables and click "close tab" you get stuck in loading and clicking "Create table" doesn't work. 

Also navigating somewhere else and back to Table Editor doesn't fix it meaning the user is stuck, unless they create a table with the SQL Editor.

CleanShot 2025-05-16 at 11 54 39@2x

jordienr (approved):

PR #35717, opened 2025-05-16
Update user impersonation search

saltcod:
1. allow search by id, email, phone # or email 
2. return email, display name (if exists) and user id in the results
3. store recent selections for easy re-selection

screenshot-2025-05-16-at-10 16 29


alaister (approved):
LGTM!

- Tested on preview

One tiny nit: the spacing before the line looks a little odd... not sure if it's actually an easy fix though because it might look worse without the recents being there 😅 

Screenshot 2025-05-21 at 19 54 28

PR #35726, opened 2025-05-17
fix: fix clipboard copy not working on Safari due to async issue

bytaesu:

YES

Bug fix

before

The text copy function intermittently fails to work on the Safari browser.

## What is the new behavior?

after

Tested on a local server, and the copy function works on the Safari browser.

## Additional context

The currently deployed version works fine on Chrome, but not on Safari, so I made this fix as a Safari user.


safaiyeh:
nice I came here exactly for this!

safaiyeh (approved):

PR #35769, opened 2025-05-20
fix: edge function auto refetching

alaister:

Bug fix

- Prevents `useEdgeFunctionBodyQuery()` from refetching data while the user is editing a function


saltcod (approved):
Tested on staging and looks good!

PR #35774, opened 2025-05-20
fix(studio): fallback to count estimate if exact count query timeout

avallete:

YES/NO

## What is the new behavior?

## Additional context


apps/studio/components/grid/components/footer/pagination/Pagination.tsx
```diff
@@ -133,6 +133,14 @@ const Pagination = () => {
     }
   }, [id])
 
+  useEffect(() => {
+    // If the count query encountered a timeout error with exact count
+    // turn off the exact count to rely on approximate
+    if (isError && snap.enforceExactCount && error.code === 408) {
```
alaister, line 139: If you don't want to include `snap.enforceExactCount` and `error.code` in the dependency array, you should wrap them in `useLatest()` outside the `useEffect` so we don't accidentally read stale values here!

alaister (approved):
LGTM!

- Tested with a partitioned table on preview. Note I couldn't trigger a timeout, so just testing the happy path.

One small code note:

apps/studio/components/grid/components/footer/pagination/Pagination.tsx
avallete: Oh ! I was wondering since `snap.setPage` wasn't a deps here: https://github.com/supabase/supabase/pull/35774/files#diff-357ec6b1c5b04e3a99a116ebe222a9682dc4cbb1ab1bc175202db3a6fb6193c6R124-R128

I wondered if maybe there was some case where we don't wanna mention all dependencies (methods, derived properties, ...).

But I don't mind declaring them all in the array of dependencies. Good spot.

PR #35799, opened 2025-05-21
default to last 24h in free, 7 days in paid

jordienr:
Free plan will default to last 24h instead of last hour.
Paid plans default to last 7 days.


alaister (approved):
LGTM!

- Tested on preview

PR #35808, opened 2025-05-21
fix new replica click

jordienr:
# problem
clicking new replica in the infrastructure settings doesnt do anything

now it will open the panel


saltcod (approved):
aah it's this one! 
looks good on staging
screenshot-2025-05-23-at-17 01 57

PR #35809, opened 2025-05-21
feat: add last used timestamp in the last 24h to legacy api keys

hf:
Queries the API Gateway logs to find the last time `anon` or `service_role` were seen in the past 24 hours.

image

### DIFF
--- a/apps/studio/components/ui/ProjectSettings/DisplayApiSettings.tsx
+++ b/apps/studio/components/ui/ProjectSettings/DisplayApiSettings.tsx
@@ -1,3 +1,7 @@
+import dayjs from 'dayjs'
+import duration from 'dayjs/plugin/duration'
+import relativeTime from 'dayjs/plugin/relativeTime'
+import { useMemo } from 'react'
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { JwtSecretUpdateStatus } from '@supabase/shared-types/out/events'
 import { useParams } from 'common'
@@ -8,6 +12,10 @@ import { useAsyncCheckProjectPermissions } from 'hooks/misc/useCheckPermissions'
 import { useFlag } from 'hooks/ui/useFlag'
 import { AlertCircle, Loader2 } from 'lucide-react'
 import { Input } from 'ui'
+import useLogsQuery from 'hooks/analytics/useLogsQuery'
+
+dayjs.extend(duration)
+dayjs.extend(relativeTime)
 
 const DisplayApiSettings = ({
   legacy,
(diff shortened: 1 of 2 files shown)


apps/studio/components/ui/ProjectSettings/DisplayApiSettings.tsx
```diff
@@ -143,6 +173,9 @@ const DisplayApiSettings = ({
                       (legacy ? 'Prefer using Secret API keys instead.' : '')
                 }
               />
+
+              {(!isLoadingLastUsed && lastUsedAPIKeys[x.api_key]?.toISOString()) ??
+                'Not used in the past 24 hours.'}
```
hf, line 178: Hocus pocus make this pretty

apps/studio/components/ui/ProjectSettings/DisplayApiSettings.tsx
```diff
@@ -42,6 +50,36 @@ const DisplayApiSettings = ({
   // api keys should not be empty. However it can be populated with a delay on project creation
   const isApiKeysEmpty = apiKeys.length === 0
 
+  const { isLoading: isLoadingLastUsed, logData: lastUsedLogData } = useLogsQuery(projectRef, {
+    iso_timestamp_start: new Date(Date.now() - 24 * 60 * 60 * 1000).toISOString(),
+    iso_timestamp_end: new Date().toISOString(),
+    sql: "-- last-used-anon--service_role-api-keys\nSELECT unix_millis(max(timestamp)) as timestamp, payload.role, payload.signature_prefix FROM edge_logs cross join unnest(metadata) as m cross join unnest(m.request) as request cross join unnest(request.sb) as sb cross join unnest(sb.jwt) as jwt cross join unnest(jwt.apikey) as apikey cross join unnest(apikey.payload) as payload WHERE apikey.invalid is null and payload.issuer = 'supabase' and payload.algorithm = 'HS256' and payload.role in ('anon', 'service_role') GROUP BY payload.role, payload.signature_prefix",
+  })
+
+  const lastUsedAPIKeys = useMemo(() => {
```
jordienr, line 59: Can you turn this into a function and wrap it in a trycatch? I'm worried it could cause runtime issues since we dont validate the data coming from the LF query

function getLastUsedAPIKeys(lastUsedLogData)...

const lastUsedAPIKeys = getLastUsedAPIKeys()...

Also this will make it easier to add a test if we have to update this logic.

jordienr (changes requested):
just one change

jordienr (approved):
tested locally, lgtm!

PR #35804, opened 2025-05-21
fix(studio): self-hosted assistant authorization header

alaister:
## What kind of change does this PR introduce?

Bug fix

Fixes https://github.com/supabase/supabase/issues/34955, fixes https://github.com/supabase/supabase/issues/33777

The issue is when the user is using basic auth, we accidentally overrode the browser provided basic auth header with an empty string.

PR #35812, opened 2025-05-21
fix(studio): bypass pg meta route on self-hosted for ai endpoints

alaister:
Follow up to #35804


bradennapier:
I can confirm this fixes it! 

By the way it would be nice to have a SUPABASE_AI_MODEL for self hosting so we can select our own.

jordienr (approved):

PR #35813, opened 2025-05-21
fix(troubleshooting sync): use proper graphql variables

charislam:
Don't know why I was using string interpolation instead of GraphQL variables for the GitHub API calls, leading to all kinds of escaping problems. Changed to using GraphQL variables properly.


fsansalvadore (approved):
nice! :)

PR #35831, opened 2025-05-22
feat: Snippets

ivasilov:
This PR adds snippets for the selfhosted/CLI version of Studio. The snippets are saved in the `supabase/snippets` folder. This may be configurable to match the Docker image structure. The snippets are saved as SQL files and may be edited with code editor outside Studio.

All features (create, save, rename, folders, search, sorting, pagination) are the same as on hosted version except:
- Shared snippets are removed (the selfhosted Studio is single user so snippets are already shared in a way)
- Favorite snippets are removed
- Snippet name is not AI generated on first save (You can still use AI when renaming a snippet)


apps/studio/pages/project/[ref]/index.tsx
```diff
@@ -119,7 +119,7 @@ const Home: NextPageWithLayout = () => {
                     </Link>
                     <p className="text-2xl tabular-nums">
                       {isLoadingTables ? (
-                        <ShimmeringLoader className="w-full h-[32px] w-6 p-0" />
+                        <ShimmeringLoader className="h-[32px] w-6 p-0" />
```
Comment (ivasilov) [inline apps/studio/pages/project/[ref]/index.tsx:150]: Duplicate classes `w-full` and `w-6`.

supabase/config.toml
```diff
@@ -73,14 +73,14 @@ enable_confirmations = false
 # Uncomment the following to use gh oAuth app locally with your own test app
 # to use env vars locally, run > `source ./supabase/.env && supabase [command...]`
 [auth.external.github]
-enabled = true
+enabled = false
```
ivasilov, line 76: Needs to be reverted.

joshenlim:
Doing a first pass smoke test specifically for self-hosted (yet to test hosted)

## Bugs identified
- Moving a snippet to a folder, seems to have a duplicate row in the interface
  - Seems like a UI issue, rather than a duplicate file - once I open that file in the editor, the duplicate row goes away
  - Moving back to the root seems okay

  https://github.com/user-attachments/assets/5a0a3223-d471-40c3-b9a7-9fd65c3809c2

- Missing validation when moving a snippet to another folder which has another snippet of the same name
  - The operation completes successfully from the API request, but one of the snippet then goes missing
  - Reckon we should add both client and server side validation

  https://github.com/user-attachments/assets/4d4c5f4f-7a42-4887-8d7f-83f5ba6c683e

- There's some issue with moving multiple snippets that I'm not able to properly pinpoint atm sorry, the modal doesn't close and I saw in the request network tab that there's a 404 (Skip to 00:20 in the video for the demo)

  https://github.com/user-attachments/assets/91b46374-5aca-4ba8-b743-7ea319172f71

  - Notice in the video at the end, the files all did get moves but Untitled Query 405 there's a duplicate row at the root and in the folder. When I refresh the browser, Untitled Query 405 in the root goes appear
  image

- Searching doesn’t return the correct count number, and also seems to return unrelated results
  - I think it's just returning the list of snippets at the root (ignores the ones in folders)
  Screenshot 2025-08-07 at 14 21 44

- Unable to delete folder (vague error from API)
Screenshot 2025-08-07 at 14 23 12

- (Unrelated) Double clicking on either “Templates” or “Quickstarts” from the product menu doesn’t make its tab permanent - this might also be on hosted probably

- Should remove “Shared” and “Favorites” sections
  image

- Should remove share snippet and add to favourite context menu options
  image

- Should have some UI indication to provide OpenAI key if needed for Assistant related functionality in SQL Editor
  - e.g Edit with Assistant via CMD K, Rename with Supabase AI  
  image


## Feedback
- Thinking that duplicating a snippet should perhaps also add a random integer to the name, otherwise won’t be able to duplicate multiple without changing the name of the duplicated snippet
  - Likewise for when selecting a template or quickstart, perhaps the first copy doesn’t need the integer, only subsequent ones
  - Storage explorer does something similar when uploading a file of the same name ([ref](https://github.com/supabase/supabase/blob/feat/local-snippets/apps/studio/state/storage-explorer.tsx#L1638))

apps/studio/lib/api/snippets.browser.ts
```diff
@@ -0,0 +1,37 @@
...
+      // Simple LCG (Linear Congruential Generator) for deterministic randomness
+      seed = (seed * 1103515245 + 12345) & 0x7fffffff
+      bytes[i] = (seed >>> 16) & 0xff
+    }
+    return Array.from(bytes)
+  }
+
+  // Generate UUID v4 using the deterministic RNG
+  return uuidv4({ rng })
+}
```
gregnr, line 37: Any reason for this over uuidv5?

apps/studio/lib/api/snippets.utils.ts
```diff
@@ -0,0 +1,451 @@
+import fs from 'fs/promises'
```
gregnr, line 1: Unless we had a good reason not to:

```suggestion
import fs from 'node:fs/promises'
```

apps/studio/lib/api/snippets.utils.ts
```diff
@@ -0,0 +1,451 @@
...
+import { compact, sortBy } from 'lodash'
+import path from 'path'
```
gregnr, line 3:
```suggestion
import path from 'node:path'
```

apps/studio/lib/api/snippets.utils.ts
```diff
@@ -0,0 +1,451 @@
...
+  const entries = await getFilesystemEntries()
+  const foundSnippet = entries.find((e) => e.type === 'file' && e.id === snippetId)
+
+  if (!foundSnippet) {
+    throw new Error(`Snippet with id ${snippetId} not found`)
+  }
+
+  return buildSnippet(
+    foundSnippet.name,
+    foundSnippet.content || '',
```
gregnr, line 192:
```suggestion
    foundSnippet.content ?? '',
```

apps/studio/lib/api/snippets.utils.ts
```diff
@@ -0,0 +1,451 @@
...
+
+  if (snippetAtTargetLocation && snippetAtTargetLocation.id !== foundSnippet.id) {
+    throw new Error(
+      `Snippet named "${updates.name ?? foundSnippet.name}" already exists in the specified folder`
+    )
+  }
+
+  const snippet = buildSnippet(
+    foundSnippet.name,
+    foundSnippet.content || '',
```
gregnr, line 381:
```suggestion
    foundSnippet.content ?? '',
```

apps/studio/lib/api/snippets.utils.ts
```diff
@@ -0,0 +1,451 @@
...
+
+  // it's easier to delete the old file first and then recreate a new one
+  await deleteSnippet(snippet.id)
+
+  const updatedSnippet = await saveSnippet({
+    name: updates.name ?? snippet.name,
+    content: updates.content ?? snippet.content,
+    // folder_id can be null
+    folder_id: updates.folder_id !== undefined ? updates.folder_id : snippet.folder_id,
+  } as Snippet)
```
gregnr, line 394: Why do we need casting here?

apps/studio/lib/api/snippets.utils.ts
```diff
@@ -0,0 +1,451 @@
...
+
+/**
+ * Updates a snippet in the filesystem. It also handles renaming and moving.
+ */
+export async function updateSnippet(id: string, updates: DeepPartial<Snippet>): Promise<Snippet> {
+  const entries = await getFilesystemEntries()
+  const foundSnippet = entries
+    .filter(
+      (entry): entry is FilesystemEntry & { type: 'file'; content: string } => entry.type === 'file'
+    )
```
gregnr, line 357: Discriminated unions work nicely for this use case. Instead of:

```ts
export type FilesystemEntry = {
  id: string
  name: string
  type: 'file' | 'folder'
  folderId: string | null
  content?: string // Only for files
  createdAt: Date
}
```

Do:

```ts
export type BaseEntry = {
  id: string
  name: string
  folderId?: string
  createdAt: Date
}

export type FileEntry = BaseEntry & {
  type: 'file',
  content?: string
}

export type FolderEntry = BaseEntry & {
  type: 'folder'
}

export type FilesystemEntry = FileEntry | FolderEntry
```

Then TS will auto narrow:

```ts
entries.filter((entry) => entry.type === 'file') // narrows to `FileEntry[]`
```

awaseem:

PR #35961, opened 2025-05-27
fix: select newer postgres version as default

delgado3d:
Reverse sort available versions so that newer version is selected by default. Version selector is only available in staging/local.


alaister (approved):
LGTM!

- Tested on preview

PR #35972, opened 2025-05-27
fix(ui): show confirmLabelLoading in ConfirmationModal when loading is true

ffuyu:
Previously, the confirm button always rendered `confirmLabel`, even when loading. This change ensures that `confirmLabelLoading` is displayed when loading is true.

YES

Bug fix

The confirm button always shows `confirmLabel`, even when `loading` is `true`. The `confirmLabelLoading` prop was accepted but not used.

## What is the new behavior?

When `loading` is `true`, the confirm button now displays `confirmLabelLoading` if provided.

## Additional context

This improves the UX of destructive actions by aligning the label with the loading state. It also unblocks use cases relying on `confirmLabelLoading`.


Hallidayo:

PR #35977, opened 2025-05-27
seo(docs): canonical links for reference pages

charislam:
To avoid content duplication warnings, direct all canonical links for reference pages to the newest library version.


jordienr (approved):
lgtm
CleanShot 2025-05-28 at 10 31 33@2x

PR #35983, opened 2025-05-28
Create branch without GitHub connection

SaxonF:
image

This PR makes so you can create branches even if there is no GitHub connection.
- Adds GitHub connection to account/me settings
- Enable branching without GitHub connection including renaming main
- Creating a new branch without GitHub connection
- Editing a branch

The ability to enable and create branching is behind a feature flag (`gitlessBranching`). There is currently no way to merge these branches (coming up) and we have some work to do so that new branches more accurately reflect main vs just migrations.

Once we have these two pieces in places we can replace the feature flag with a feature preview.

## To test:
1. Open a Supabase project in dashboard
2. Enable branching in project without GitHub connection
3. Create a new branch
4. Enable GitHub connection on project
5. Create a new branch and choose GitHub branch


jgoux (commented):
I'm just reviewing from a usage and DX perspective. 👍 

Not sure if it's in the scope of this PR but when a GitHub repo isn't connected, maybe we should hide the Pull Request tab and the "GitHub branch workflow" info box. Also it seems like there is an empty button in this info box on the right of the "Settings" button (I think it's the one linking to the linked repository normally).

image

When I click on the "Settings" button, I'm in the project's integration page, and I'm unable to enable GitHub connection:

image

There is the info box saying I need to go to my main branch to do any modification to the project's settings, it's helping, but I struggled a good minute before actually reading it and understanding why I can't edit the settings from my dev branch. Maybe we could adapt the disaled buttons messages to reflect this info?
Or we could maybe add an `Edit` button only showing up in preview branches on every settings page which would just be a link to the same page from the main branch because here I clicked on the `main` link in the info box and get redirected to `http://localhost:8082/project/{ref}/settings/general`, ideally I would be redirected to the same page.

When the project is connected to GitHub, the dialog says that the link to the Git branch is optional while it's in fact required:

image

I assume there is no possibility to create non-connected Supabase branches for now due to this dialog's restriction?

SaxonF:
@jgoux 

> Not sure if it's in the scope of this PR but when a GitHub repo isn't connected, maybe we should hide the Pull Request tab and the "GitHub branch workflow" info box. Also it seems like there is an empty button in this info box on the right of the "Settings" button (I think it's the one linking to the linked repository normally).

I've updated to add a message about connecting GitHub instead of hiding. We can use this pull requests page in future to show merge requests from gitless branches too.

> When I click on the "Settings" button, I'm in the project's integration page, and I'm unable to enable GitHub connection:

We can definitely improve this and just the general connection between branch <> project. We have some IA work to do later which clarifies that relationship so can take this on at same time if needed.

> I assume there is no possibility to create non-connected Supabase branches for now due to this dialog's restriction?

You'll need to enable feature flag to test this

joshenlim:
Generally looks good to me from a smoke test! Just some stuff:

## Issues identified

- nit: account preferences -> connections, GH icon probably doesn't need a bg and border? can refer to account identities. Also - should we just mention branching here? In my mind i'm thinking "why would I want to sync GH repos to supabase projects?", so figure we can load the value prop up front
image

- nit: close button in enable branching modal overlaps the docs button
image

- Edit branch modal, the validation seems incorrect? my branch name seems to acceptable based on the error message
image

## Feedback

- Wondering what's the motivation for the "Connections" section under account preferences? We have this under org/integrations too where the user can add a GH connection. Just wondering if it might be confusing, but no strong opinion on this

- Wondering if we should redirect the user to the branch management page after enabling branching?

- Thinking should add a tooltip here "Switch to branch", to alleviate guessing what clicking this button does
image

- Is this CTA "Switch to persistent" still relevant for Gitless branching?
image

SaxonF:
@joshenlim good feedback, thanks! I've resolved most of it.

> Wondering what's the motivation for the "Connections" section under account preferences? We have this under org/integrations too where the user can add a GH connection. Just wondering if it might be confusing, but no strong opinion on this

The reasoning was to just decouple account connection from repo connection. A project may be connected to GH through accounts other than your own.

joshenlim:
Minor issue: I can help push up the fix in a bit once i get through the smoke test (Edit: pushed a fix up!)
Enable branching modal on a free plan padding + button should be disabled
image

joshenlim (approved):
LGTM now - verified changes on preview

PR #35992, opened 2025-05-28
feat: Project claim

ivasilov:
This PR adds a page where a user can claim an already chosen project from an oAuth app.

To setup a project to be claimed, you need to do the following steps:
1. Create an oAuth app
2. Go to the homepage of the project-to-be-claimed, there's a "Generate Claim URL"
3. Open browser console, copy the token id and go to the logged oauth link
4. Copy the `token` and the `auth id` and use them in the claim-project URL so it looks like this 
```
http://localhost:8082/claim-project?auth_id=<auth_id>&token=<token>
```

If you're testing this on previews, the oAuth link would take to the staging URL, you'll need to update it to `https://studio-staging-git-feat-project-claim-supabase.vercel.app/dashboard`

Auth ids expire in 5 minutes, so you need to repeat the 2-4 steps.


joshenlim:
## Minor issues:

- This page needs to vertically center align the panel
image

- This page too - vertically center align? Also the layout of this page feels odd - most of the UI is horizontally centered except the list?
image

- Nit: Also i reckon we can increase the content width here - maybe `max-w-lg` instead of `max-w-md`
Otherwise i'd presume an average length project name will get wrapped here
image

- Nit: niche case, my oauth app has no permissions - so the dropdown shows nothing when expanded. Probably needs some empty state?
image

- Stress testing came in handy 😂 Probably needs a truncate
image

## Feedback:

- After creating a new org - default the search input to the new org name? chances are i'm looking for the newly created org

- I feel like the token validation should happen on the first step? Only seeing this error after clicking on an organization (2nd step)
image

## Questions:

- I don't quite understand what this permissions here mean or imply? I understand that these are the OAuth App permissions - but what does this mean for my project after claiming (or what does "apply" mean)?
[Update] I think i understand now, so the app that you're claiming the project from, they will still have these permissions on your project after you've claimed right? Perhaps the copy writing needs a bit more clarity then ("List of permissions which will be applied...")
image

- Is it expected that my project goes straight into resizing after getting claimed? Feels kinda odd from a UX POV - expected that I'd be able to immediately look into my project 
image

joshenlim:
### Non-blocking feedback
- Still think this could be clarified as “List of permissions that {app name} will have for the selected organization and all of its projects” (But is my understanding here correct though?)
image

- Could also clarify at the end “The access that {app name} has is described below” or something similar
image

- Reckon this could be an Admonition component perhaps
image

- Choose another organization CTA here should remove the `org` URL param to clear search results
image

- I think it might be better if we can show this full header from step 1 when choosing an org if possible
image

- Probably needs a invalidation here? I got re-directed to my org page after claiming but the project doesn't show up
image

apps/studio/components/interfaces/Organization/ProjectClaim/layout.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { UserDropdown } from 'components/interfaces/UserDropdown'
+import { FeedbackDropdown } from 'components/layouts/ProjectLayout/LayoutHeader/FeedbackDropdown'
+import { BASE_PATH } from 'lib/constants'
+import { Separator } from 'ui'
+
+export const ProjectClaimLayout = ({
+  children,
+  title,
+}: PropsWithChildren<{
+  title: React.ReactNode
```
joshenlim: nit: import ReactNode from react

apps/studio/data/projects/project-claim-token.ts
joshenlim, line 1: guessing this needs to be removed?

apps/studio/pages/project/[ref]/index.tsx
```diff
@@ -106,6 +141,7 @@ const Home: NextPageWithLayout = () => {
                   infra_compute_size: project?.infra_compute_size,
                 }}
               />
+              <Button onClick={buildOAuthUrl}>Generate URL for claiming</Button>
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/index.tsx]: guessing this needs to be removed?

joshenlim (approved):
Tested on staging preview the whole claim project flow, just a couple of non-blocking feedback and clean up to do before we merge this in 🙂 🙏

PR #36010, opened 2025-05-29
fix: Sandbox email preview frame

doublethink:
This change will allow scripts to continue running in the Preview pane of the Email Template view, but will run as a cross origin context, namely calls and scripts will not inherit the users cookies.

Could not be tested locally, but have POC'ed this iframe config
image
Can add additional sandbox attributes as needed (just not allow-same-origin)

Fixes SEC-348


staaldraad (approved):
tested locally and in staging. Works as expected. 
Has the nice side-effect of blocking `alert` too, due to 
```
Ignored call to 'alert()'. The document is sandboxed, and the 'allow-modals' keyword is not set.
```

I don't think we need `allow-modals` since emails shouldn't be loading modals..

SaxonF (approved):
Tested can view emails in staging.

PR #36012, opened 2025-05-29
text and heading components

SaxonF:
image

This adds two separate approaches to standardizing typography across the dashboard.
- A Text and Heading component with variants
- [Base CSS styles](https://github.com/supabase/supabase/pull/36012/files#diff-a2ba3886fb7933475fd9003b26cab438275202399837eb513d3652dcbcbd60fc) applied to typographic elements along with classes

A decision needs to be made on which (if any) approach is a suitable _first_ step for us to take.


awaseem:

PR #36022, opened 2025-05-29
perf: speed up local dev

kevcodez:
* Migrate from SCSS -> CSS to boost performance and avoid unnecessary Sass parsing/compilation
* Bumped Turbo, PostCSS and Next.js
* Use `incremental` option for TypeScript compilation


ivasilov (approved):

CoolAssPuppy (approved):

ivasilov (changes requested):
The SQL editor colors are not initialized.

Screenshot 2025-06-09 at 12 35 17

PR #36049, opened 2025-05-30
Assistant MCP tests

gregnr:
Refactors AI model loading and MCP logic so that we can unit test them.

Specifically we test that:
1. The correct model loads (bedrock vs openai) based on environment. We reuse this `getModel()` logic everywhere we use AI (cron, completion, etc).
2. MCP tools are correctly filtered and stubbed based on opt-in level


SaxonF (approved):
Tested locally the different permission levels and all working as expected. We have some changes from generate-v3 and complete endpoints on main branch for self hosted support. Are we able to pull in resolve conflicts before merging?

PR #36104, opened 2025-06-02
fix: delete row loading spinner

alaister:

Bug fix; when deleting a row, no loading spinner is shown. This PR fixes that.

Screenshot 2025-06-02 at 17 31 32


joshenlim (approved):
Tested deleting a row via checkbox, deleting all rows with filters applied (to trigger isDeletingAllRows), and deleting all rows with no filters applied (to trigger isTruncatingRows), verified loading state is now present

PR #36105, opened 2025-06-02
Configure restricted region pool from configcat

joshenlim:
`RESTRICTED_POOL` within `get-default-region-query` is currently hard-coded within the FE repo, but we're gonna move this to be configurable through ConfigCat instead - primarily to support percentage-based sampling with different values


delgado3d (approved):

ivasilov (approved):
LGTM

PR #36108, opened 2025-06-02
fix(table-editor): handle multi-dimentional array truncation and preview

avallete:

- Add a case over truncation arrays queries to handle multi-dimentionals arrays. Which fix the error: `Arrays with differing dimensions are not compatible for concatenation.` When trying to preview multi-dimentional arrays.
- Add a test case that reproduce the issue without the change to avoid regression.
- Update and unify the `isTruncatedValue` logic on the TableEditor side to handle this case.

- Tested over the `stress-table-editor` project on preview with the `monitor_data` table.


joshenlim:
JFYI found an issue - since we now have the truncation logic in place for even numbers (apart from text and JSON)
reckon we then need to do the “load full content” logic in the Row Editor Side Panel too then, as currently if i load the full content in the cell, its not the same value as per the row editor side panel

For context:
image

I think we need to address this first before this PR can go through

So array fields will need the edit button to open the data in a side panel + load content logic

joshenlim (approved):
Verified that the monitor_data table sample cannot be opened in the table editor on prod due to a 400 when retrieving the rows, and can be opened in the staging preview

also tested, within the grid and row editor side panel
- able to load + view the data on demand
- able to update the data

PR #36113, opened 2025-06-02
Update global error boundary component FE-1662

jordienr:
- Rename component to GlobalErrorBoundary to differentiate it from the ErrorBoundary component easily
- Adds some instructions on how to fix common errors
- Adds buttons to perform those instructions 
- Adds button to copy the error

## Before
image

## After
CleanShot 2025-06-02 at 15 30 10@2x

### DIFF
--- a/apps/studio/components/ui/GlobalErrorBoundaryState.tsx
+++ b/apps/studio/components/ui/GlobalErrorBoundaryState.tsx
@@ -18,7 +18,7 @@ export type FallbackProps = {
   resetErrorBoundary: (...args: any[]) => void
 }
 
-export const ErrorBoundaryState = ({ error, resetErrorBoundary }: FallbackProps) => {
+export const GlobalErrorBoundaryState = ({ error, resetErrorBoundary }: FallbackProps) => {
   const router = useRouter()
   const checkIsError = isError(error)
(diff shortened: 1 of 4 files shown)


joshenlim:
will see if we can streamline this a bit - feels like a lot of buttons up front, esp in the context that its an error and users might already be annoyed/frustrated

cause am thinking things like "Try accessing the dashboard in an incognito/private window." or "Try a different browser to see if the issue persists." might not be steps to *resolve* this error - as from the user's POV they aren't gonna always be using incognito, or swap to a different browser from their preference

^these are perhaps debug steps that are useful to us as the maintainers of the dashboard, but not so much for the users

joshenlim:
okay think we might go with this perhaps?
- Added header with logo so that the page shift isn't too jarring when error boundary is triggered (since its similar to most pages of the dashboard)
- Swapped for inline CTAs instead of separate buttons (quicker to find while reading)
- Added a title + shifted the copy button to be inline with the title
- Updated recommended steps title

Lemme know what you think!

image

joshenlim (approved):
Verified changes locally since we can't really trigger on prod (we shouldn't be able to 😬 😂  )

PR #36122, opened 2025-06-02
Use email instead of username in org member list

saltcod:
Swap username for email in org member list

screenshot-2025-06-02-at-16 07 27


joshenlim (approved):
Verified on staging preview

PR #36130, opened 2025-06-03
fix: use configcat proxy

kangmingtay:

YES

* Use configcat proxy

## To test
- [x] Verify that toggling on/off a feature flag works as expected

## Additional context

- [x] Requires adding `NEXT_PUBLIC_CONFIGCAT_PROXY_URL` to local / staging / prod (configcat.supabase.green for local / staging and configcat.supabase.com for prod)
  - [Joshen] Added the env var for `studio-staging` and `studio` vercel apps
  - [Joshen] Added env var in BitWarden for local


staaldraad (approved):

joshenlim (approved):
Verified on staging by toggling the incident banner on/off

Once this is deployed to prod, will be good to immediately check thereafter that feature flags are also working there
- We can do this by toggling the incident banner but only for internal staff, and also inform the wider team about what we're doing so that we don't jumpscare our team 😬

PR #36131, opened 2025-06-03
Tiny patch table editor

joshenlim:
Mistake in TableGridEditor where we're checking through _all_ open tabs for the CTA when landing on a table ID that doesn't exist - resulting in an incorrect URL if there were opened SQL tabs as well

We already had declared `openedTabs` that filters out SQL Editor tabs, just that we weren't using it 😓


alaister (approved):
LGTM!

- Tested on preview

PR #36140, opened 2025-06-03
Log Explorer improvements

jordienr:
- make warnings tooltip instead of popover so it doesnt have to be clicked
- improve spacing, padding
- fix LIMIT warning not being caught when written in uppercase
- fix bug where log explorer doenst send query with correct time range

## To test:  
- write a query in log explorer
`select
 COUNT(path),
 path
from edge_logs
cross join unnest(metadata) as metadata
cross join unnest(response) AS response
cross join unnest(request) AS request
where
  regexp_contains(path, '^/rest/v1/')
group by path` 

- set date range to 7 days
- you should see 2 warnings

CleanShot 2025-06-03 at 12 27 18@2x


saltcod (approved):
Checked staging and works as expected.

PR #36141, opened 2025-06-03
fix schema selector in create fn sheet

jordienr:


saltcod (approved):
Checked every instance of SchemaSelector, all work as expected.

PR #36147, opened 2025-06-03
Chore/feedback classifier

saltcod:
Detect feedback | support in the FeedbackDropdown and add an alert if looks like feedback

screenshot-2025-06-05-at-17 12 34

 screenshot-2025-06-05-at-17 12 15


apps/studio/pages/api/ai/feedback/classify.ts
```diff
@@ -0,0 +1,58 @@
...
+          error:
+            'Your feedback prompt is too large for Supabase AI to ingest. Try splitting it into smaller prompts.',
+        })
+      }
+    } else {
+      console.error(`Unknown error: ${error}`)
+    }
+
+    return res.status(500).json({
+      error: 'There was an unkgenerwtor classcron syntgxhe feedback. Please try again.',
```
gregnr: 😵‍💫

joshenlim (approved):
Giving this a pre-emptive ✅ as I've checked the code + smoke test locally

But i'm running into weird errors when testing sending feedback on staging / local
image

Prod seems okay though and the changes in this PR doesn't change the feedback sending behaviour, so reckon should be okay - although we should immediately check on prod once this gets deployed to be sure

PR #36157, opened 2025-06-04
fix home counts

SaxonF:
Fixes issue where replica count on project home can be below 0


apps/studio/pages/project/[ref]/index.tsx
```diff
@@ -70,9 +70,9 @@ const Home: NextPageWithLayout = () => {
     projectRef: project?.ref,
   })
 
-  const tablesCount = tablesData?.length ?? 0
-  const functionsCount = functionsData?.length ?? 0
-  const replicasCount = (replicasData?.length ?? 1) - 1
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/index.tsx:75]: hmm iirc we need to minus one here, cause replicasData also includes the primary database (which shouldn't be considered as a replica) and we'd be showing 2 incorrectly then - i'll double check on this branch

joshenlim (approved):

PR #36160, opened 2025-06-04
Fixing The unnecessacity of calling auth.getUser inside an edge function.(Issue : #35777)

### LINKED ISSUE #35777: Calling auth.getUser inside an edge function is not necessary
# Improve documentation

## Link 

[https://supabase.com/docs/guides/functions/auth](https://supabase.com/docs/guides/functions/auth)

## Describe the problem

In this page in the docs, it's suggested that you call `auth.getUser` with the JWT in order to get the user object. This will do a network request to the auth server to verify the JWT then return the decoded User object. However, unless `verify_jwt` is set to false for the edge function, the JWT signature is already verified. There's no need to do a network request to check it again. You can simply decode the JWT to get the user object.

## Describe the improvement

You can instead just decode the JWT with any JWT library like [https://deno.land/x/djwt@v3.0.2](https://deno.land/x/djwt@v3.0.2) for example:

```javascript
import { decode } from 'jsr:@zaubrik/djwt';

Deno.serve(async (req) => {
  // we know this header must exist and be valid
  const jwt= req.headers.get('Authorization')?.replace('Bearer ', '')!;
  // this doesn't check the JWT signature, but it was already checked before the request reached us
  const [_header, user, _signature] = decode(jwt); 
  return new Response(JSON.stringify(user), { status: 200 });
});
```

For creating requests with the RLS policies of that user, you can just provide the JWT instead of the `SUPABASE_ANON_KEY` when you create the client. So instead of 

```javascript
  const supabaseClient = createClient(
    Deno.env.get('SUPABASE_URL') ?? '',
    Deno.env.get('SUPABASE_ANON_KEY') ?? '',
  );
```

you would do

```javascript
  const supabaseClient = createClient(
    Deno.env.get('SUPABASE_URL') ?? '',
    jwt
  );
```

with the `jwt` from the authorization header.

seika-afk:
In the Page : "supabase.com/docs/guides/functions/auth":
""
 it's suggested that you call auth.getUser with the JWT in order to get the user object. This will do a network request to the auth server to verify the JWT then return the decoded User object. However, unless verify_jwt is set to false for the edge function, the JWT signature is already verified. There's no need to do a network request to check it again. You can simply decode the JWT to get the user object.
""
## Solution : 

Decoded the JWT ,and directly provided the user. Without the unnecessarily need of network request to check again.

Before :
sss
 

After :  
ss

### DIFF
--- a/examples/edge-functions/supabase/functions/select-from-table-with-auth-rls/index.ts
+++ b/examples/edge-functions/supabase/functions/select-from-table-with-auth-rls/index.ts
@@ -1,6 +1,7 @@
 // Follow this setup guide to integrate the Deno language server with your editor:
 // https://deno.land/manual/getting_started/setup_your_environment
 // This enables autocomplete, go to definition, etc.
+
 import {decode} from "jsr:@zaubrik/djwt"
 import { createClient } from 'jsr:@supabase/supabase-js@2'
 import { corsHeaders } from '../_shared/cors.ts'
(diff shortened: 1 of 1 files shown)


Hallidayo:

PR #36162, opened 2025-06-04
allow select and copy text, fix context menu

jordienr:
- allows users to select part of the row text to easily copy
- fixes right click context menu


joshenlim (approved):
Verified in logs explorer
- Can select partial text of a row and copy
- Context menu is working on each row with copy event message as the main CTA

PR #36163, opened 2025-06-04
Adjust open new tab SQL Editor CTA

joshenlim:
"Open in new tab" button here opens the new in a new _browser_ tab which doesn't really make sense if the Tabs Interface is enabled. This was brought up in the GH discussion [here](https://github.com/orgs/supabase/discussions/35636#discussioncomment-13235674), so updating the button to just open a new tab in the tabs interface, instead of browser tab if the tabs interface feature preview is enabled

image


jordienr (approved):
tested in preview

PR #36164, opened 2025-06-04
Fix cmd i not closing AI assistant

joshenlim:
As per PR title - were just using `openAssistant` incorrectly


jordienr (approved):
tested in local dev

PR #36169, opened 2025-06-04
fix: use LOGFLARE_PRIVATE_ACCESS_TOKEN for querying, LOGFLARE_PUBLIC_…

Ziinc:
This PR fixes querying for studio for latest v1.14+ logflare.
Querying should be done via the private access token env var, `LOGFLARE_PRIVATE_ACCESS_TOKEN`.
Ingestion should be done via the public access token env var, `LOGFLARE_PUBLIC_ACCESS_TOKEN`.

This PR also does the following:
- shifts vector ingestion API key from the query params to the header so that it does not get logged in api gateway logs.
- removes all usage of `LOGFLARE_API_KEY` env var
- removes redundant documentation from spec files, updates related spec files.
- removes deprecated `get()` usage from studio analytics endpoint proxying logic, switches to direct `fetch` usage
 

post merge:
- need to bump studio image version in compose file

Screenshot 2025-06-04 at 7 27 23 PM

### DIFF
--- a/apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts
+++ b/apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts
@@ -36,14 +36,14 @@ const proxyRequest = async (req: NextApiRequest) => {
   const apiKey = process.env.LOGFLARE_PRIVATE_ACCESS_TOKEN
   const url = `${PROJECT_ANALYTICS_URL}endpoints/query/${name}${search}`
   console.log('url', url)
-  const result =  await fetch(url, {
+  const result = await fetch(url, {
     method: 'GET',
     headers: {
       'x-api-key': apiKey as string,
       'Content-Type': 'application/json',
       Accept: 'application/json',
     },
-  }).then(res => res.json())
+  }).then((res) => res.json())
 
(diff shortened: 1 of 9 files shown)


apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts
```diff
@@ -33,14 +33,17 @@ const proxyRequest = async (req: NextApiRequest) => {
   const { name, ...toForward } = req.query
   const payload = { ...toForward, project_tier: 'ENTERPRISE' }
   const search = '?' + new URLSearchParams(payload as any).toString()
-  const apiKey = process.env.LOGFLARE_API_KEY
+  const apiKey = process.env.LOGFLARE_PRIVATE_ACCESS_TOKEN
   const url = `${PROJECT_ANALYTICS_URL}endpoints/query/${name}${search}`
-  const result = await get(url, {
+  console.log('url', url)
```
Comment (sweatybridge) [inline apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts]:
```suggestion
```

sweatybridge (approved):
just a nitpick

joshenlim (approved):
minimally verified that staging preview logs explorer is all still working as per status quo

PR #36175, opened 2025-06-04
fix infinite loop in api keys

jordienr:
### Problem. 

- The code used new Date() as a default argument in a hook, which created a new value on every render.
- This caused the query parameters to change every time the component rendered.
- React Query saw the parameters as different and kept refetching data.
- This led to an infinite loop of fetching and re-rendering.
---
### fix:
- Makes sure the "now" timestamp used for fetching API key usage logs is stable and does not change on every render.
- Uses useRef(new Date()).current so the query only uses the time from when the component first mounts.


saltcod (approved):
Checked project settings → api keys and looks good

PR #36184, opened 2025-06-05
assistant button in help popover

SaxonF:
image

Adds a Supabase Assistant option to HelpPopover which opens Assistant along with some predefined suggestions.

### DIFF
--- a/apps/studio/components/layouts/ProjectLayout/LayoutHeader/HelpPopover.tsx
+++ b/apps/studio/components/layouts/ProjectLayout/LayoutHeader/HelpPopover.tsx
@@ -60,7 +60,7 @@ export const HelpPopover = () => {
             {projectRef && (
               <ButtonGroupItem
                 size="tiny"
-                icon={<AiIconAnimation allowHoverEffect size={16} />}
+                icon={<AiIconAnimation allowHoverEffect size={14} />}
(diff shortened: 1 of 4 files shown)


apps/studio/components/layouts/ProjectLayout/LayoutHeader/HelpPopover.tsx
```diff
@@ -44,42 +48,70 @@ export const HelpPopover = () => {
...
+            )}
+            <ButtonGroupItem
+              size="tiny"
+              icon={<Wrench strokeWidth={1.5} size={14} />}
+              onClick={() =>
+                window.open('https://supabase.com/docs/guides/platform/troubleshooting', '_blank')
+              }
+            >
+              Troubleshooting
+            </ButtonGroupItem>
```
joshenlim: nit: could we use `<a>` tags here instead of manually calling window.open? from an accessibility POV

joshenlim: so perhaps ButtonGroupItem needs to support an `asChild` param

joshenlim (approved):
Verified changes on preview

MildTomato (approved):

PR #36185, opened 2025-06-05
MCP Self Hosted Check

SaxonF:
For now, if environment is self hosted, we will need to use old tools for accessing schema information rather than MCP. Self-hosted level defaults to 'schema'. This may change in future.


joshenlim:
JFYI not able to test this atm as we've got some issues with running the self-host/non-hosted set up of the dashboard locally, will do a smoke test once thats resolved 🙏

PR #36192, opened 2025-06-05
Add context menu to Tabs Interface for closing tabs

joshenlim:
As per suggestion from Tabs Interface GH Discussion [here](https://github.com/orgs/supabase/discussions/35636#discussioncomment-13351469): 
Adds a context menu to the tabs interface to support a few tab closing actions
image

## Things to test
- Each context menu action works as expected, in particular the post-closing behaviour
  - e.g close all -> should redirect to base page
  - e.g close others -> should redirect to the tab which you hit close others on, if not currently on that tab
- Each context menu action is isolated to just the editor (e.g Hitting "Close to the right" in the table editor, should not affect any tabs in the SQL editor)


saltcod (approved):
nice! tested and works well!

PR #36214, opened 2025-06-06
feat: add back sql-to-rest AI tool

gregnr:
## Background

We previously had to remove the `convertSqlToSupabaseJs()` assistant tool because of an underlying dependency that broke builds on some versions of macOS (and Windows). The dependency was `libpg-query` which is a Postgres SQL parser that works by creating native Node bindings to Postgres C code. This native C++ build process is what broke a lot of dev environments. It also included a WASM version, but it would still attempt to build the native bindings on install.

## Change
We've since replaced `libpg-query` with `@supabase/pg-parser` which exclusively compiles the Postgres C code to WASM at the time of publish without requiring any native bindings. This PR updates `@supabase/sql-to-rest` to the latest version which uses the new parser.

We also used libpg-query for tests in `ai-commands`. To keep things consistent this PR also replaces that with `@supabase/pg-parser`, which is mostly a drop-in replacement.


Hallidayo:

PR #36218, opened 2025-06-06
fix: Add admonition about pitr to new project for vercel marketplace

kamilogorek:
This is currently not available and once someone does restore it like that, they get locked out of possibility to remove the project, as it's treated as managed by Vercel.


joshenlim (approved):
Verified locally since I don't have a staging project that's managed by vercel marketplace

PR #36229, opened 2025-06-06
fix: bump auth-js to 2.71.0-rc.4

hf:
Fixes a bug.


ivasilov (changes requested):
You need to also change the version in `pnpm-workspace.yml`.

cemalkilic (approved):

ivasilov (approved):

PR #36315, opened 2025-06-10
feat(api gateway logs): add error code explanation

charislam:
Add the ability to look up error code explanation in API Gateway logs. Also a bunch of GraphQL-related utilities and generated types for calling the Content API (since we hope to expand the ability to pull content from dashboard).

image
image


apps/docs/app/api/graphql/route.ts
```diff
@@ -12,6 +12,31 @@ export const runtime = 'edge'
...
 
+function isAllowedCorsOrigin(origin: string): boolean {
+  const exactMatches = IS_DEV
+    ? ['http://localhost:8082', 'https://supabase.com']
+    : ['https://supabase.com']
+  if (exactMatches.includes(origin)) {
+    return true
+  }
+
+  return /^https:\/\/[\w-]+\w-supabase.vercel.app$/.test(origin)
```
doublethink, line 48: is "-supabase" a reserved subdomain in vercel?
Would test true if someone outside our org created an app with the same pattern>

charislam: hmm good point, I don't think there's anything preventing someone from registering the `my-supabase` org with Vercel and matching it... I don't think there's any match we could do that someone couldn't imitate, so we'd need a custom domain on previews to filter by

or, given that this is an endpoint with zero authentication anyway, I just allow '*' and don't bother pretending to secure it?

apps/studio/data/graphql/fragment-masking.ts
charislam, line 1: auto-generated by graphql-codegen

apps/studio/data/graphql/gql.ts
charislam, line 1: autogenerated by graphql-codegen

apps/studio/data/graphql/graphql.ts
charislam, line 1: autogenerated by graphql-codegen

apps/studio/data/graphql/index.ts
charislam, line 1: autogenerated by graphql-codegen

apps/studio/scripts/schema.graphql
charislam, line 1: Automatically pulled from remote endpoint schema (like api-types)

apps/docs/app/api/graphql/route.ts
doublethink: If those vercel hostnames has IS_DEV we could put it in there?
If not possible, at least the CORS is better than nothing. Not a big deal

charislam: can't filter by IS_DEV unfortunately, because preview studio needs to connect to prod docs (we have no persistent staging version of docs)

doublethink (approved):

jordienr (approved):

jordienr (approved):

PR #36322, opened 2025-06-11
Add edge functions empty state for local/self-host

joshenlim:
## Context

Management of edge functions in the dashboard for local CLI or self-host is currently not available and we hide the Edge Functions button in the side bar navigation in those environments.

However, this has led to some confusions about the parity difference between the hosted and non-hosted "versions" of the dashboard, so we're bridging this gap a little by minimally not hiding the Edge Functions button + adding an empty state UI for managing edge functions with local CLI or self-host

image
image

Templates are still available for reference and clicking on them will open the template snippet in a modal
image
image

## What to test / verify

- [ ] Content for the added empty state makes sense from a dev's POV and are factually correct
- [ ] Changes should not affect the hosted dashboard's Edge Functions pages


saltcod:
screenshot-2025-06-12-at-15 48 24
capitals here?

alaister (approved):
LGTM!

- Tested on hosted and self hosted previews

PR #36323, opened 2025-06-11
feat: Add a toast to refresh the page if there's a new version of Studio available

ivasilov:
This PR will show a toast to users who are on older version of Studio to refresh the page. 

The toast is only shown if:
- there's a newer version of Studio deployed and
- the version that's currenly shown is more than 24 hours old

If the user dismisses it won't show up again (until the next refresh).


apps/studio/hooks/use-check-latest-deploy.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <div className="flex gap-3 flex-col w-full">
+      <div className="flex gap-3 flex-row">
+        <StatusIcon variant="default" className="mt-0.5" />
+        <div className="flex w-full justify-between flex-col text-sm">
+          <p className="text-foreground">A new version of this page is available</p>
+          <p className="text-foreground-light">Refresh to see the latest changes.</p>
+        </div>
+      </div>
+
+      <div className="flex gap-5 justify-end">
```
joshenlim, line 24: reckon gap-x-2 is sufficient - we usually use 2 for spacing between buttons

joshenlim (approved):
Verified toast showing up with the test set up on preview

kevcodez:
@ivasilov While the default 24h staleness seems fine, what would be very useful is being able to define a minimum commit/version that forces refreshes more or less immediately (within a few minutes).

Sometimes we may roll out a fix or the likes and it requires a dashboard refresh - it would be great if these type of fixes can be forced onto the user quicker to avoid more users running into problems

apps/studio/hooks/use-check-latest-deploy.tsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    // prevent showing the toast again if user has already seen and dismissed it
+    if (isToastShown) {
+      return
+    }
+
+    // check if the time difference between commits is more than 24 hours
+    const hourDiff = dayjs(commit.commitTime).diff(dayjs(currentCommitTime), 'hour')
+    // if (hourDiff < 24) {
```
ivasilov: This needs to be uncommented.

apps/studio/pages/api/get-deployment-commit.ts
```diff
@@ -0,0 +1,40 @@
...
+    return 'unknown'
+  }
+}
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse<{ commitSha: string; commitTime: string }>
+) {
+  // Set cache control headers for 10 minutes so that we don't get banned by Github API
+  // res.setHeader('Cache-Control', 's-maxage=600')
```
ivasilov: This also needs to be uncommented.

PR #36324, opened 2025-06-11
fix: include primary database identifier in API report filtering

jose-ledesma:

YES

Bug fix

API reports are filtering by secondary identifiers but are missing the primary database identifier in the filter criteria. This means the primary database identifier is not being included when generating filtered reports, creating inconsistent filtering behavior across different identifier types.

## What is the new behavior?
API reports now include the primary database identifier in the filtering logic alongside the existing secondary identifiers. This ensures consistent and complete filtering behavior across all identifier types.


joshenlim (approved):
Verified on staging preview on a project with no replicas (only primary)
the API reports page is the same

PR #36325, opened 2025-06-11
fix: managed by header

kevcodez:
When pushing the org into the org list, there is a banner showing "managed by ", because the "managed_by" property is missing, which is only set when retrieving data on organization listing.


joshenlim (approved):

PR #36333, opened 2025-06-11
feat: support pending changes

kevcodez:
Adds support for the new Orb pending subscription change flow that has been added to the backend.

By default, the new feature is disabled (enabled after merge on staging). If disabled, this should work just like before with the regular `confirmPayment` function that triggers 3DS if needed and properly lets a user confirm their additional factor.

With the new flag, we only create a payment method, as we will do another payment intent that is set up for future usage and that may require 3DS - so we avoid the possible double confirmation (setup confirmation + payment intent). The organization creation endpoint can either return a full organization (i.e. on Free Plan or immediate payment success) or a payment_intent_secret. In case of the secret being returned, we need to confirm the secret with the Stripe SDK.

Moved the Stripe Elements further down to avoid the entire form reloading (especially billing address) in case of a payment failure.

----

**Tests**
_With Pending Changes (New)_
- [x] Create Free Plan org
- [x] Create Pro Plan org without 3DS
- [x] Create Pro Plan org with 3DS
- [x] Fail if credit card declined, ability to re-attempt
- [x] Fail if 3DS is denied
- [x] Org creation fails unrelated to payment and resets payment successfully
- [x] Indian Credit Card with mandate

_Without Pending Changes (Old)_
- [x] Create Free Plan org
- [x] Create Pro Plan org without 3DS
- [x] Create Pro Plan org with 3DS
- [x] Fail if credit card declined, ability to re-attempt
- [x] Fail if 3DS is denied
- [x] Org creation fails unrelated to payment and resets payment successfully
- [x] Indian Credit Card with mandate

### DIFF
--- a/apps/studio/components/interfaces/Billing/Payment/PaymentConfirmation.tsx
+++ b/apps/studio/components/interfaces/Billing/Payment/PaymentConfirmation.tsx
@@ -1,0 +1,36 @@
+import { useStripe } from '@stripe/react-stripe-js'
+import { PaymentIntentResult } from '@stripe/stripe-js'
+import { useEffect } from 'react'
+import { LoadingLine } from 'ui'
+
+export const PaymentConfirmation = ({
+  paymentIntentSecret,
+  onPaymentIntentConfirm,
+  onLoadingChange,
+  paymentMethodId,
+}: {
+  paymentIntentSecret: string
+  paymentMethodId: string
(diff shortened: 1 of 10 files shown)


ecktoteckto:
Looks good. Tested locally

ecktoteckto (approved):

staaldraad (approved):

PR #36337, opened 2025-06-11
fix: trim confirmation input and comparison

kevcodez:
We regularly have customer support tickets that cannot confirm because of a mismatching space at the end (and they cannot for the life of them figure this one out).

Trimming both input and comparison string will prevent unnecessary support tickets and still guarantees enough safety.


joshenlim (approved):
LGTM! Verified on preview 🙏

PR #36067, opened 2025-05-31
style: add comma to the Applied to target roles

Mohammed-Yasin-Mulla:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

Issue Link : #36068 

## What kind of change does this PR introduce?

Style changes

## What is the current behavior?
Screenshot 2025-05-31 at 1 31 44 PM

The target roles do not have any space in between them, making it difficult to read
## What is the new behavior?
image

PR #36338, opened 2025-06-11
Restore spaces

saltcod:
fixes spacing issue added in https://github.com/supabase/supabase/pull/36067/files

Before: 

screenshot-2025-06-11-at-14 53 20

After: 
screenshot-2025-06-11-at-14 52 00

Imo the tooltip if roles > 3 isn't really necessary here


joshenlim (approved):
Verified on preview

PR #36349, opened 2025-06-12
Update new project flow to add a warning before redirecting

saltcod:
Issue:

- users land at supabase.com/new/abc
- if they don't have access to that org, they get automatically redirected to organizations[0].slug
- it isn't obvious that this has happened, and users end up creating projects in the wrong org 

This PR make is explicit: they're shown that they don't have access to their referring org and given a list of orgs that they are part of. 

screenshot-2025-06-11-at-23 08 37

### DIFF
--- a/apps/studio/components/interfaces/Organization/NewProject/NotOrganizationOwnerWarning.tsx
+++ b/apps/studio/components/interfaces/Organization/NewProject/NotOrganizationOwnerWarning.tsx
@@ -1,19 +1,30 @@
 import InformationBox from 'components/ui/InformationBox'
 import { AlertCircle } from 'lucide-react'
 
+interface NotOrganizationOwnerWarningProps {
+  slug?: string
+}
+
 // [Joshen] This can just use NoPermission component i think
-const NotOrganizationOwnerWarning = () => {
+const NotOrganizationOwnerWarning = ({ slug }: NotOrganizationOwnerWarningProps) => {
   return (
     <div className="mt-4">
       <InformationBox
-        icon={<AlertCircle className="text-white" size="20" strokeWidth={1.5} />}
+        icon={<AlertCircle size="20" strokeWidth={1.5} />}
         defaultVisibility={true}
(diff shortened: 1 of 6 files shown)


ivasilov (changes requested):
I think the PR is in right direction, but the `You do not have permission to create a project` message may not be clear enough that you reached an org that may not exist or you don't have access to.
Screenshot 2025-06-12 at 23 24 24

the logic for whether you can create a project in a free org is flawed. I have 5 orgs, 2 of them are listed as eligible, but when I select one of them, I'm hit with the limit reached message. We could also decide that this is an edge case and just list all orgs.
Another issue connected to this is what happens when all orgs are filitered through the current flawed logic (`if (numProjects) return null`)? No filtering would be best, I think. 

Screenshot 2025-06-12 at 23 18 51
Screenshot 2025-06-12 at 23 19 01

joshenlim:
## Feedback

- Agree with Ivan on the following notes:
  - "You do not have permission to create a project" is misleading, communicates a different problem and should be hidden if the org slug in the route doesn't correspond to any org that the user is a part of
  - Should not filter orgs with no projects - we should be listing all organizations unless i might be overlooking something here

> I have 5 orgs, 2 of them are listed as eligible, but when I select one of them, I'm hit with the limit reached message

- RE ^ this, i think that's okay, the main aim (imo) is for us to facilitate getting the user into the right UI flow which is the create project form. From there they can re-select the organization via the drop down. 
- Based on this as well, I don't think we need any warning or callout about "org doesn't exist" if they land on /new/slug where the slug doesn't correspond to any org that the user is a part of

- Organization dropdown in LayoutHeader needs to be handled properly 
image

## Changes made
- Hide "You do not have permissions" callout if the slug doesn't corresponding to any orgs that the user is a member of
- Show _all_ organizations, instead of hiding those with no projects
- Added empty state for organization dropdown in LayoutHeader
image

image

ivasilov (approved):

PR #36351, opened 2025-06-12
fix: redirect to Vercel marketplace for plan change

kevcodez:
Disable plan change buttons for Vercel marketplace managed orgs and rather redirect to Vercel marketplace


joshenlim (approved):
Verified changes locally by hardcoding the variables 🙏

staaldraad (approved):

PR #36355, opened 2025-06-12
fix log explorer date state on first load

jordienr:
fixes 2 issues:
- log explorer date state gets reseted between running queries
- log explorer date state is incorrect on first load


awaseem:

PR #36363, opened 2025-06-12
Mitigate race condition when fetching folder contents storage

joshenlim:
Resolves FE-1677

## Context

We've got a report internally that the storage explorer occasionally shows an error "Failed to retrieve folder contents from ...: bucketId is required". This error stems from `fetchFolderContents` in the storage explorer store, and `fetchFolderContents` uses the internal state `state.selectedBucket.id` to call the request to fetch the contents

`StorageExplorer.tsx` has a a couple of `useEffect`s - of which two of them have a tendency for a race condition, specifically:
- The `useEffect` for `openBucket` with `bucket` in the dependency array
- The `useEffect` for `fetchFolderContents` with `itemSearchString` in the array

Because of this - there's technically a chance of `fetchFolderContents` being triggered before `openBucket` was called, in which case `state.selectedBucket` wouldn't have been set just yet, and hence cause the error

## Changes involved
- `fetchFolderContents` now accepts the an additional parameter `bucketId` rather than relying on the internal state of `selectedBucket`
- Within the `useEffect` for `fetchFolderContents` in the if block for `STORAGE_VIEWS.COLUMNS`, conditionally call `fetchFolderContents` if `openedFolders.length` is 0
- Added a hook `useSelectedBucket` so that the currently selected bucket can be accessed easily

## To test
- [ ] Mainly just do a brief smoke test of the storage explorer, ensure that everything is behaving as per status quo (open bucket, open folders, etc)

### DIFF
--- a/apps/studio/components/to-be-cleaned/Storage/EmptyBucketModal.tsx
+++ b/apps/studio/components/to-be-cleaned/Storage/EmptyBucketModal.tsx
@@ -19,7 +19,12 @@ export const EmptyBucketModal = ({ visible = false, bucket, onClose }: EmptyBuck
   const { mutate: emptyBucket, isLoading } = useBucketEmptyMutation({
     onSuccess: async () => {
       if (bucket === undefined) return
-      await fetchFolderContents({ folderId: bucket.id, folderName: bucket.name, index: -1 })
+      await fetchFolderContents({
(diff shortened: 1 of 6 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #36365, opened 2025-06-12
docs: fix wrappers assets (images)

tomaspozo:

YES

This adds a small function that updates the assets urls imported by the wrappers md files to use proper external urls from the wrappers repo (just like we load the md files).

The images are not loading in the supabase/docs/wrappers pages because they are referenced as local resources, but they are actually external resources.

image

## What is the new behavior?

With this fix, now they load properly:

image

## Additional context

None.

### DIFF
--- a/apps/docs/app/guides/database/extensions/wrappers/[[...slug]]/page.tsx
+++ b/apps/docs/app/guides/database/extensions/wrappers/[[...slug]]/page.tsx
@@ -96,6 +96,10 @@ async function getLatestRelease(after: string | null = null) {
   }
 }
 
+function rewriteAssetUrls(mdContent: string, baseUrl: string): string {
+  return mdContent.replace(/(\.\.\/)+assets\//g, baseUrl)
+}
+
 // Each external docs page is mapped to a local page
 const pageMap = [
   {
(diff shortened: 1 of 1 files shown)


charislam (commented):
Thanks @tomaspozo! This is a great idea but string replacement is kind of brittle. I would prefer this to be done with a rehype plugin, similar to how we handle linkTransform.

Is this something you want to take on? I would be happy to review an updated PR, or if you don't have that much time to invest, you could open an issue instead!

tomaspozo:
Hi @charislam! thank you for your review and feedback! Yes, I would like to tackle this and get more familiar with the code base and best practices, I expect to keep contributing to the docs 😃 

Will make the fix and update the PR!

Thank you!

tomaspozo:
Hi @charislam, I just pushed some [changes](https://github.com/supabase/supabase/pull/36365/commits/ef7a7157c9e7f54cdac6f8be4c2b645629a4dcc2) based on you feedback.

Let me know if this is what is wanted, or if there is any additional change I need to make. 

Thank you!

charislam (approved):

PR #36367, opened 2025-06-12
fix(SQLEditor): disable transactional mode for manual queries

avallete:
Statement timeout implicitely create a transaction which can cause issue while running some queries (eg: `CREATE INDEX CONCURENTLY`).

While we want this to be enabled to ensure we don't run long running queries over the database by mistake, in the case of SQLEditor, it's an user manually ran query, we want to allow anything to run.


joshenlim (approved):
Code changes LGTM - although not tested on preview yet as corresponding infra [change](https://github.com/supabase/infrastructure/pull/23566) is not merged just yet

avallete:
Screenshot 2025-06-12 at 22 35 56

Just tested over the preview now that staging is released. Ensured the query run didn't had the statement_timeout.

Still need the change to be deployed on prod before merging.

PR #36382, opened 2025-06-13
Update setup-replication-external.mdx

Alexkane:
Clearer directions around logical replication in PG.

YES

Small documentation update

Which database (primary vs replica) is unclear.

## What is the new behavior?

Clarify which database to execute the command on.

## Additional context

I think this would be helpful for others to setup logical replication without having to dig in to the PG docs.


charislam (approved):
Thanks @Alexkane!

PR #36386, opened 2025-06-13
Shift positioning of "Leave team" button in Org Team Settings

joshenlim:
## Context

We had a feedback that the "Leave team" button in the organization team settings was a little easy to miss. Understandably so as the member actions are typically found in the row of each member, whereas the "Leave team" button was above the table.

## Changes involved

- "Leave team" button is now shifted to the same row as the user
image
- Refactored "Leave team" button into its own component
- Fix sorting of org members to always have your own user as the first in the list

## To test
- [ ] Verify that you can leave an organization

### DIFF
--- a/apps/studio/components/interfaces/Organization/TeamSettings/LeaveTeamButton.tsx
+++ b/apps/studio/components/interfaces/Organization/TeamSettings/LeaveTeamButton.tsx
@@ -1,0 +1,121 @@
+import { useRouter } from 'next/router'
+import { useState } from 'react'
+import { toast } from 'sonner'
+
+import { LOCAL_STORAGE_KEYS, useParams } from 'common'
+import { ButtonTooltip } from 'components/ui/ButtonTooltip'
+import { useOrganizationRolesV2Query } from 'data/organization-members/organization-roles-query'
+import { useOrganizationMemberDeleteMutation } from 'data/organizations/organization-member-delete-mutation'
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Organization/TeamSettings/LeaveTeamButton.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const { slug } = useParams()
+  const { profile } = useProfile()
+  const selectedOrganization = useSelectedOrganization()
+
+  // if organizationMembersDeletionEnabled is false, you also can't delete yourself
+  const { organizationMembersDelete: organizationMembersDeletionEnabled } = useIsFeatureEnabled([
+    'organization_members:delete',
+  ])
+
+  const [isLeaving, setIsLeaving] = useState(false)
```
ivasilov, line 29: Shouldn't this `isLoading` on the mutation?

ivasilov (approved):

apps/studio/components/interfaces/Organization/TeamSettings/LeaveTeamButton.tsx
joshenlim: ahh yes edge case here: we're using a useState here as i'd want the button to still be disabled while we're re-refetching the organizations + re-routing the user

if we were to use `isLoading` from the mutation, the button will undo the disabled state too early

saltcod (approved):
Great change!

PR #36430, opened 2025-06-16
feat: support pending changes for plan upgrades

kevcodez:
This PR implements the new flow to confirm subscription upgrades using Orb pending changes. This is backwards compatible and based on a flag exposed by the backend (`subscriptionPreview.pending_subscription_flow`).

Just like the organization creation, the entire flow is slightly different - instead of creating a payment method separately, the payment method is added inline while doing the upgrade and then attached to the customer. If payment fails, the upgrade will not go through. If payment requires additional action, the user needs to confirm the payment before allowing the upgrade.

For testing the new flow locally, toggle the flag in `flags.ts` on the backend.

Changes include
- No longer rely on the `changeType` from the plans endpoint as this is regularly out-of-sync and displays wrong up/downgrade info due to race conditions
- `readOnly` mode for Stripe elements if anything is loading/submitting
- Reduced prop drilling for some components
- Hide payment method and address selection on downgrade

----

Tested

Old flow (feature toggle off)
- [x] Upgrade from free -> Pro
- [x] Upgrade from Pro -> Team
- [x] Downgrade from Team -> Pro
- [x] Downgrade from Pro -> Free
- [x] Use already present payment method
- [x] Put down new payment method in the modal
- [x] 3DS confirmation on setup works

New flow (feature toggle on)
- [x] Upgrade from free -> Pro
- [x] Upgrade from Pro -> Team
- [x] Downgrade from Team -> Pro
- [x] Downgrade from Pro -> Free
- [x] Use already present payment method
- [x] Put down new payment method inline
- [x] Should fail upgrade if the payment fails
- [x] 3DS confirmation on payment works
- [x] If 3DS confirmation fails, user gets an appropriate message and can try again


ecktoteckto (approved):

ecktoteckto:
Looks good. Tested locally

doublethink (approved):

PR #35925, opened 2025-05-26
chore: move HIPAA checks from org to project level

staaldraad:
HIPAA orgs can mark specific projects as sensitive. Those projects will have HIPAA restrictions applied to them. This opens non-HIPAA projects in the same Organization up to use previously disabled features.

## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

Feature update

## What is the current behavior?

HIPAA restrictions apply at the Org level for some features, but at the project level for others.

## What is the new behavior?

HIPAA restrictions are now at the project level

## Additional context

https://supabase.com/docs/guides/platform/hipaa-projects

PR #35318, opened 2025-04-28
Feature/assistant mcp

SaxonF:
## What kind of change does this PR introduce?

- Updates the Assistant to use the [Supabase MCP](https://github.com/supabase-community/supabase-mcp) as a library
- Updates AI SDK 
- Updates to use Bedrock with Sonnet 4
- Updates to system prompt to temporarily include condensed version of old RLS and Edge Function knowledge. In future, we may re-introduce knowledge retrieval tools to MCP

There is a side effect of this change. The MCP supports running queries directly then reading and interpreting the results. This is different to the current behaviour where all queries are run outside the context of the Assistant (QueryBlock runs the query and renders results). This creates a couple of challenges:
- Ideally the results that are run do not need to be output by the model and can instead just be referenced. This is why we are wrapping the execute_sql tool to include a unique id with results that the model can include when calling displayBlock tool. We use this ID to retrieve results on the front-end.
- We are now sending not just schema information to OpenAI but also actual data. We are currently looking in to the best way to handle this from a privacy perspective to make sure the user has full control and visibility over what happens with their data. **We cannot merge this PR until we have a solution.**

## What is the current behavior?

Uses tools built in in Studio

## What is the new behavior?

Uses Supabase MCP

## How to test
1. Set up AWS ENV variables

    ```bash
    AWS_REGION=us-east-1
    AWS_PROFILE=supabase-dev
    ```
3. Switch to [infra PR](https://github.com/supabase/infrastructure/pull/22828)
4. Open Assistant and test:
    - Make sure old chats still work and charts are still visible
    - Listing tables
    - Performing a read query and notice results shown in chat
    - Ask the Assistant to interpret the results
    - Ask to create a write query (e.g. create a table) and notice query needs to be run manually
    - Ask to create an edge function
    - Ask to create RLS policies
    - Open SQL editor and use the inline assistant

## Other questions
- Do we need to remove OpenAI from title generation and old sql debugging?

## Infrastructure
This work requires adding a tag to API via https://github.com/supabase/infrastructure/pull/22828

PR #36436, opened 2025-06-16
Fix incorrect logic for isHipaaProjectDisallowed

joshenlim:
## Context

We've received some reports of the SQL snippets not automatically renaming by AI, causing a list of "Untitled query" to snowball. Realised that the AI renaming of snippets wasn't getting triggered because of a newly added `isHipaaProjectDisallowed` check that was introduced [here](https://github.com/supabase/supabase/pull/35925).

The `useDisallowHipaa` hook in itself is quite confusing tbh, and we've looked into refactoring the logic in this [PR](https://github.com/supabase/supabase/pull/35318) - so am bringing the changes from that PR out to fix this issue

## To test
- [ ] Verify that running a new query in the SQL editor will rename "Untitled query" into something relevant to the query's contents


staaldraad (approved):
Tested in staging ✅

alaister (approved):
LGTM!

- Tested on preview

PR #36439, opened 2025-06-16
feat: add UI to show MFA enforcement on org

staaldraad:

Feature

When MFA is enforced on an organization and the user does not have MFA enabled, the organization is not visible in the dashboard.

## What is the new behavior?

Disables organization selection if the org requires MFA but the user does not have MFA enabled.

Add badges / hints to show that user has access to an Organization but access is disabled because of MFA access requirement

## Additional context

MFA enforcement is still behind a feature flag

User has MFA enabled, can select all orgs and has a badge to know about enforcement:

Screenshot 2025-06-16 at 13 16 12

User does not have MFA enabled, org is visible but not selectable:
Screenshot 2025-06-16 at 13 17 25
Screenshot 2025-06-16 at 13 17 42


joshenlim:
Pushing up some UI tweaks

- Reduced UI indication for MFA enabled to just an icon and tooltip
image

- Allow switching to the organization that has MFA enforced but user hasn't set up MFA + show an indicator to set up MFA prior to accessing organization
image

- Add redirect logic, when entering a page that's scoped to an organization (includes project pages), if org has MFA enabled but user does not have MFA set up, redirect back to org home page

- (Unrelated) Fix OrganizationCard linking to /new/[slug], support taking `href` as a param instead

alaister (approved):
LGTM!

- Tested on preview

inian (approved):

PR #36443, opened 2025-06-16
fix: Refactor the careers page to handle a rate-limit issue from Github.

ivasilov:
This PR fixes an issue where the Careers page would break if it hits the Github rate limit (~60 API calls per hour).

This shouldn't usually happen because we're caching the response (maybe Vercel colocates functions from other projects under the same IP?).


fsansalvadore (approved):
Looks good!

PR #36458, opened 2025-06-17
Remove enable branching modal

SaxonF:
image

image

The point of this PR is to abstract away the concept of enabling/disabling branches. The enabling process happens as part of branch creation if coming from a gitless workflow OR as part of the GitHub connection workflow. Setting production git branch now happens within the GitHub connection panel.

To test:
- Enable `gitlessBranching` feature flag
- Create or view project with branching disabled
- Notice branch button with dropdown in header instead of enable branching button
- Visit /branches
- Click create new branch
- Notice create branch modal
- Create a new branch without GitHub and see new branch
- Connect GitHub to project
- Create a new branch with GH connected
- Disable feature flag and try and create a gitless branch (gh disabled)


joshenlim:
## Bugs identified

- For a repo that's connected to GH, Is this supposed to always show a check despite the branch name not corresponding to any branch in the connected repo? The branch checking doesn't seem to be getting triggered either - i don't see a call to `/branches/branchName` in the network tab
image

- (Non-FE) Migration on branches seems to be failing for a project thats on PG15
image

## Feedback

- Thinking the requirement for GH connection while the `gitlessBranching` flag is off might be easily missed, users might be wondering why the create branch button is disabled - thinking whether we should add a tooltip of sorts, or make the requirement more pronounced
image

- Hmm layout wise this looks a bit awkward atm
image

- The `GithubIntegration/GithubSection.tsx` is also used in the organizations settings page so the language here is a bit off atm. From "Add project connection" to "Automatic branching"

https://github.com/user-attachments/assets/9132342c-c8bc-4363-af90-27295ac55d70

apps/studio/components/layouts/ProjectLayout/ProjectLayout.tsx
```diff
@@ -257,7 +261,11 @@ const ProjectLayout = forwardRef<HTMLDivElement, PropsWithChildren<ProjectLayout
             </ResizablePanel>
           </ResizablePanelGroup>
         </div>
-        <EnableBranchingModal />
+        <CreateBranchModal
+          visible={showCreateBranchModal}
+          onClose={() => setShowCreateBranchModal(false)}
```
joshenlim: the component could just get these from the snapshot directly, instead of passing in a params

apps/studio/components/layouts/AppLayout/BranchDropdown.tsx
```diff
@@ -113,78 +226,30 @@ export const BranchDropdown = () => {
...
-                  </CommandGroup_Shadcn_>
-                </CommandList_Shadcn_>
-              </Command_Shadcn_>
+              <MenuContent branchList={sortedBranches ?? []} showSearch />
             </PopoverContent_Shadcn_>
           </Popover_Shadcn_>
         </>
       )}
+
+      {isSuccess && (!isBranchingEnabled || branches.length === 0) && (
```
joshenlim: nit: could probably remove the branches.length === 0 check since it seems unnecessary

once branching is enabled, and even after you've deleted all preview branches, isBranchingEnabled will stay as true

apps/studio/components/interfaces/Integrations/VercelGithub/IntegrationConnection.tsx
```diff
@@ -167,4 +167,4 @@ const IntegrationConnectionItem = forwardRef<HTMLLIElement, IntegrationConnectio
 
 IntegrationConnectionItem.displayName = 'IntegrationConnectionItem'
 
-export { IntegrationConnectionItem }
+export { IntegrationConnectionItem, IntegrationConnection }
```
joshenlim: nit: seems unnecessary

alaister (approved):
LGTM!

- Tested as described on preview

PR #36466, opened 2025-06-17
Add feature flag for opt in tags

joshenlim:
## Context

Prior to the release of the new Assistant MCP, we're going to have to reset our users' opt in preference for the AI Assistant through a migration on the BE. While the migration is going on, we'll be defaulting the opt in level for all users to `disabled` and also preventing users from changing their opt in levels, in which these 2 behaviours will be controlled by a feature flag

image
image

Once the migration is completed, we can then toggle the feature flag to be on, in which users can thereafter change their opt in levels again, and their opt in levels will be respected in `useOrgAiOptInLevel`

## Changes involved

While feature flag is OFF
- Users will not be able to update their Assistant Opt In levels (via org settings or opt in modal)
- Assistant Opt In levels will be defaulted to `disabled`


SaxonF (approved):
Looks good. Tested on staging and confirmed inability to change permission and level set to disabled.

PR #36477, opened 2025-06-17
Fix loading in confirmation modal

joshenlim:


saltcod (approved):
yay! reviewed on staging, works!

PR #36480, opened 2025-06-17
Fix leave team button visibility

joshenlim:
## Context

The "Leave team" button was recently shifted [here](https://github.com/supabase/supabase/pull/36386) to be consistent with the member actions, but we noticed that it doesn't show up if you're a "Developer" role.

This is due to a wrong ordering of conditionals when rendering the UI in `MemberActions.tsx`.

(Unrelated) Also gonna opt to always show the "Invite member" button, but disable it if the permissions are insufficient (avoid hidden UI)

## To test

- [ ] Ensure that as a member in the organization, you're able to leave the organization irregardless of role
- [ ] Ensure that as a Developer or lower role, you can see the invite member button, but it should be disabled with a tooltip


saltcod (approved):
Verified leaving an org as a developer on staging — looks good!

PR #36503, opened 2025-06-18
Chore/deprecate lib common fetch

joshenlim:
Related to FE-1638 (Housekeeping)

## Context

Will be looking into cleaning up `lib/common/fetch` files by either removing unused ones, or consolidating them into `data/fetchers`. With our current set up though, we probably cannot completely remove the methods in `lib/common/fetch` as they're still used in our dashboard internal API routes - so we'll be looking into slowly consolidating them into `data/fetchers`, and removing any duplicated methods (e.g `constructHeaders`)

## Changes involved in this PR

- Refactored `BuildingState` and `RestoringState` to use RQ instead of manually long polling the project status endpoint (Also removes the use of `getWithTimeout` so that's deprecated too)
- Refactor `profile-create-mutation` to use RQ's `post`
- Shift `post` from `lib/common/fetch` to `data/fetchers` + refactor `bucket-object-download-mutation` to use that new method

I'll be doing the others in another PR, as the remaining changes are solely affecting dashboard in CLI / self-hosted 

## Things to test
- [ ] Able to create a new account on the dashboard (to test `profile-create-mutation`)
- [ ] Able to spin up a project (Just verify that the project should lead to the home page once its ready)
- [ ] Able to restore a paused project (Just verify that the confirmation modal shows up when project is back to healthy from restored state)
- [ ] Can retrieve edge function status health check on project home page (to test `edge-function-status-query`, just verify the request in the network tab)
- [ ] Can download files from storage explorer (to test `bucket-object-download-mutation`)

PR #36506, opened 2025-06-18
Refactor post calls from lib/common/fetch in auth pages to data/fetchers

joshenlim:
Related to FE-1638 (Housekeeping)

## Context

Continuing from this [PR](https://github.com/supabase/supabase/pull/36503), this PR focuses more on refactoring `post` calls in the Project Auth pages including a couple of clean ups

## Changes involved
- Use `fetchPost` in `queryPgMetaSelfHosted` in `lib/self-hosted`
- Add check for `IS_PLATFORM` for `auth-config-query`
  - We don't support managing auth configs in CLI/self-host dashboard, so this endpoint should not be called
- Updated `api/platform/auth/[ref]/...` routes to use `fetchPost` instead of `post` from `lib/common/fetch`
  - Invite, magiclink, OTP, recover
- Generally standardise error response from all dashboard auth endpoints
- Fix deleting a user via context menu while that user is selected to close the user overview panel
- Removed `api/platform/auth/[ref]/users.ts`
  - We use `pg-meta/query` to list the users in the dashboard
- Removed `api/platform/auth/[ref]/config.ts`
  - We don't support managing auth configs in CLI/self-host dashboard, so this endpoint isn't used
- Removed `POST`, `PATCH` and `DELETE` methods in `api/platform/pg-meta/[ref]/policies.ts`
  - These methods are not used in the dashboard anymore since management of policies (except GET) are going through `pg-meta/query`
- Add missing `api/platform/auth/[ref]/users/index.ts` to support creating users
- Fix `api/platform/pg-meta/[ref]/query/index.ts` to return the right error response
  - Was missing `formattedError` property as per hosted, resulting in some UI not showing the error message (e.g in Auth policies when creating a new policy)

## To test
I need helping testing this one as I'm not sure why the following actions are just throwing 500 for me on their respective endpoints within `pnpm run dev:studio-local` (unless this is expected)
- [ ] Invite users
- [ ] Send magic link
- [ ] Send password recovery

These ones I've verified myself locally
- [ ] Create, delete user
- [ ] Listing users
- [ ] Creating, updating, delete policies
- [ ] Error handling when creating a policy of the same name


alaister (approved):
LGTM!

- Tested on staging and in local dev setup

PR #36525, opened 2025-06-19
Fix edge functions headers span

joshenlim:
Addresses a tiny misalignment issue in edge functions secrets page
image


alaister (approved):
LGTM!

- Tested on preview

Nice fix!

PR #36527, opened 2025-06-19
Add custom link component to assistant message block

joshenlim:
## Context

In efforts to mitigate any potential risks in the hyperlinks that the AI Assistant generates, we're adding a custom `a` component that'll render in the Assistant's messages, such that clicking on any hyperlinks will open a confirmation dialog before navigating to the actual link

image

## To test

Can be tested with the following prompt which should get the assistant to generate some links:
```hey there, can you give me links to supabase docs?```

- [ ] Verify that rendered links open a confirmation dialog instead of directly opening the links

### DIFF
--- a/apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
+++ b/apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
@@ -1,3 +1,4 @@
+import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { useRouter } from 'next/router'
 import {
   DragEvent,
@@ -10,7 +11,6 @@ import {
   useRef,
 } from 'react'
 
-import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { ChartConfig } from 'components/interfaces/SQLEditor/UtilityPanel/ChartConfig'
 import { useSendEventMutation } from 'data/telemetry/send-event-mutation'
 import { useCheckPermissions } from 'hooks/misc/useCheckPermissions'
@@ -19,7 +19,21 @@ import { useSelectedProject } from 'hooks/misc/useSelectedProject'
 import { useProfile } from 'lib/profile'
 import { useAiAssistantStateSnapshot } from 'state/ai-assistant-state'
 import { Dashboards } from 'types'
-import { Badge, cn, CodeBlock, CodeBlockLang } from 'ui'
+import {
+  Badge,
+  Button,
+  cn,
+  CodeBlock,
+  CodeBlockLang,
+  Dialog,
+  DialogClose,
+  DialogContent,
+  DialogFooter,
+  DialogHeader,
+  DialogSection,
+  DialogTitle,
+  DialogTrigger,
+} from 'ui'
 import { DebouncedComponent } from '../DebouncedComponent'
 import { EdgeFunctionBlock } from '../EdgeFunctionBlock/EdgeFunctionBlock'
 import { QueryBlock } from '../QueryBlock/QueryBlock'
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx [hunk 1]
```diff
@@ -50,6 +64,58 @@ export const InlineCode = memo(
...
+
+        <DialogFooter>
+          <DialogClose asChild>
+            <Button type="default" className="opacity-100">
+              Cancel
+            </Button>
+          </DialogClose>
+          <DialogClose asChild>
+            <Button asChild type="primary" className="opacity-100">
+              <a
```
SaxonF: Should we use NextJS Link here to speed up internal dashboard linking?

apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
```diff
@@ -50,6 +64,58 @@ export const InlineCode = memo(
...
+                href={href}
```
SaxonF: should we do same safety transformation as Markdown React? https://github.com/remarkjs/react-markdown/blob/fda7fa560bec901a6103e195f9b1979dab543b17/lib/index.js#L425

SaxonF (changes requested):

apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
joshenlim: oh good point - will do a conditional render

apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
joshenlim: i'll add an additional state here if an unsafe URL is detected based on ^ this safety transformation
image

apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
SaxonF: we can still use Nextjs Link for external links?

apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
SaxonF: I think we should just not render the link of its deemed unsafe, and instead just render raw text. I doubt there will be a case where we need to render a link that fails

apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
joshenlim: hmm im trying to find the article i came across long ago but from what i understand:
for best practice - `<a>` for external links, `<Link>` for internal links

something like [this](https://deadsimplechat.com/blog/links-in-next-js-a-detailed-guide/#how-to-handle-external-links-in-nextjs)

although i can't seem to find any official documentation on this best practice though (several reddit threads do talk about this nonetheless)

it may very well no longer be the best practice actually if next has done some updates

apps/studio/components/ui/AIAssistantPanel/MessageMarkdown.tsx
joshenlim: okay yeap makes sense - will render raw text if the URL is not safe then

joshenlim: just pushed up the changes! 🙏

SaxonF (approved):

PR #36547, opened 2025-06-20
Update warnings in bucket modals

joshenlim:
## Changes involved

- Update Alerts in bucket modals to use Admonition for consistency
image
image

- Update warning copy when toggling a bucket from public to private to add instructions about purging cache
image


alaister (approved):
LGTM!

- Tested on preview

PR #36549, opened 2025-06-20
Header icons refine

SaxonF:
image

Simplifies inbox icon


joshenlim (approved):
Verified locally (no notifications on staging preview) although i do have a sneaky notification on prod haha will double check there once this is through 🙂

PR #36554, opened 2025-06-20
Minor update to new API keys modal for disabling / reenabling legacy keys

joshenlim:
Just updating the UI and copy in the modals from a UX pov - the warning in the panel feels too upfront, and i was worried whether clicking the button would open a modal, or immediately run the action

(Unrelated) I also pushed up an unrelated bug fix in the auth users management page haha - clicking refresh should refresh both the users list and the users count (currently only does the former)

## Before

image

image

image

## After

image

image

image


hf (approved):

alaister (approved):
LGTM!

- Tested on preview

PR #36555, opened 2025-06-20
feat: add auth reports FE-1569

jordienr:
Adds new auth report page.


apps/studio/components/ui/Charts/ComposedChart.tsx
```diff
@@ -149,13 +163,45 @@ export default function ComposedChart({
     color: CHART_COLORS.REFERENCE_LINE,
   }
 
+  const chartData =
```
saltcod, line 166: This is quite a thing to behold!

apps/studio/data/reports/auth-report-query.ts
```diff
@@ -0,0 +1,351 @@
...
+import { get } from 'data/fetchers'
+import { AnalyticsInterval } from 'data/analytics/constants'
+import type { MultiAttribute } from 'components/ui/Charts/ComposedChart.utils'
+import { getHttpStatusCodeInfo } from 'lib/http-status-codes'
+
+/**
+ * UTILS.
+ */
+
+type Granularity = 'second' | 'minute' | 'hour' | 'day' | 'week'
```
saltcod: utils file?

saltcod (approved):
Looks good!

apps/studio/data/reports/auth-report-query.ts
fsansalvadore: btw I don't think we ever use "second" granularity, so we could probably not include it? @jordienr

jordienr: 🫡

PR #36561, opened 2025-06-20
Fix storage explorer search

saltcod:
Fixes search

screenshot-2025-06-20-at-15 27 10


joshenlim (approved):
LGTM! Verified searching
- In column view with no folders opened
- In column view with folders opened
- In list view

Also pushed up a nit to simplify the logic

PR #36593, opened 2025-06-23
fix mcp tools and parse data

SaxonF:
Fixes issues with recent MCP update.
- We now have to parse execute_sql results to get the actual data
- We have to use `createSupabaseApiPlatform` 
- Re-added id generation to user messages to save correctly


joshenlim (approved):
Verified on preview that assistant is able to run a query

kimsoo527 (approved):

kimsoo527 (approved):

kimsoo527 (approved):

PR #36598, opened 2025-06-23
Chore/refactor unified logs infinite queries

joshenlim:
## Context

This PR is related to the Unified Logs interface, specifically refactoring the hooks in `QueryOptions` by shifting them into their respective react queries within the  `data` folder for consistency.

## Changes involved

- Refactor `dataOptions` originally in `UnifiedLogs/QueryOptions.ts`
  - Split fetching count + logs data into 2 separate queries: `unified-logs-infinite-query` and `unified-logs-count-query`, this should speed up the UI loading speed as the original InfiniteQuery was waiting on 2 calls. This will allow us to fetch both data in parallel
  - `unified-logs-infinite-query` returns `{ data, nextCursor, prevCursor }`
  - `unified-logs-count-query` returns `{ totalRowCount, facets }`
- Remove unused variables which `dataOptions` was returning (i.e. did not bring them over to the new react queries)
  - `filterRowCount` - this could be obtained directly from the RQ's data parameter
  - `chartData` - this seems to be a duplicate as we already have another hook to fetch the chart data
  - `metadata` - none of the parameters in here seems to be used in the UI
- Refactor `useChartData` into `unified-logs-chart-query`
  - Also removed `totalCount` that was getting returned here - unnecessary as we'd already have the count from the count query
- Deprecate `UnifiedLogs/QueryOptions.ts`

## To test

There shouldn't be any functionality changes to this so just need to check that the Unified Logs interface renders as status quo with what's currently on staging, no errors


apps/studio/data/logs/unified-logs-count-query.ts
```diff
@@ -0,0 +1,102 @@
...
+    params: {
+      path: { ref: projectRef },
+      query: {
+        iso_timestamp_start: isoTimestampStart,
+        iso_timestamp_end: isoTimestampEnd,
+        project: projectRef,
+        sql,
+      },
+    },
+    // body: { sql },
```
alaister, line 35: Checking this isn't needed?

apps/studio/data/logs/unified-logs-infinite-query.ts
```diff
@@ -0,0 +1,182 @@
...
+    params: {
+      path: { ref: projectRef },
+      query: {
+        iso_timestamp_start: timestampStart,
+        iso_timestamp_end: timestampEnd,
+        project: projectRef,
+        sql,
+      },
+    },
+    // body: { sql },
```
alaister, line 90: Same with this one

alaister (approved):
LGTM!

- Tested on preview

apps/studio/data/logs/unified-logs-count-query.ts
joshenlim: correct - for context sorry, we're gonna shift the logs.all requests to a POST endpoint for unified logs
just that its dependent on this API PR: https://github.com/supabase/infrastructure/pull/23013

PR #36599, opened 2025-06-23
chore: disable role impersonation for edge functions

alaister:


joshenlim (approved):
Verified that role impersonation button for edge function test panel is no longer present

joshenlim (approved):
Updated changes LGTM! Verified that Role Impersonation resets to "Service role" just for edge functions tester (and only service/anon roles are selectable) and does not follow the globally set role when using in other areas of the dashboard (e.g table editor / sql editor etc)

PR #36600, opened 2025-06-23
fix: 3ds confirmation / downgrades

kevcodez:
- Remove captcha for downgrades to Free, it is not needed
- Properly handle 3DS if customer has multiple payment methods
- Move payment method invalidation to react-query


inian (approved):

ecktoteckto (approved):

ecktoteckto:
Tested locally

PR #36601, opened 2025-06-23
Show credit balance also for partner billed organizations

ecktoteckto:
When an organization that is billed through Vercel is downgraded in the middle of the billing cycle, the organization gets credits (same behavior as with Supabase platform billing). To make that more obvious/reduce confusion, we show the credit balance on the organization's biling page. That becomes more relevant now that we automatically sync the Vercel invoice when an invoice gets voided or a credit note gets issued, resulting in fewer refunds to the credit card.

Shows the balance for all partner billed organizations. That's fine for now. Will have to do some adjustments to the Dashboard in the context of the AWS Marketplace revamp anyway.

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/BillingSettings.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/BillingSettings.tsx
@@ -82,7 +82,7 @@ export const BillingSettings = () => {
         </>
       )}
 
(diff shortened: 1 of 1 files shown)


kevcodez (approved):

doublethink:
The E2E actions runners appear to be functional?

kevcodez:
@doublethink e2e tests are flaky/broken and unit tests have been broken for a while - all independent of this PR

doublethink (approved):

PR #36614, opened 2025-06-23
Use a badge for a more explicit last used indicator

saltcod:
Folks don't seem to see the dot, let's switch for an explicit badge:

screenshot-2025-06-23-at-17 02 33

screenshot-2025-06-23-at-17 01 54


joshenlim (approved):
Verified changes on preview with GH log in and email/password login - looks good!

gianpaj:
Is this a functionality that's enabled for Supabase Auth or just for the Studio?

It would be great if we could display this easily in my Supabase-powered web app.

Clerk seems to have this https://clerk.com/changelog/2025-09-12-last-used-sign-in

Thanks

PR #36629, opened 2025-06-24
ref: Use organization_slug instead of org_id for project creation

kamilogorek:
It's been due for almost a year now since Hieu updated the endpoint. It was never followed up on the FE side. Tested locally by creating new project.


joshenlim (approved):
Verified on staging preview with a new project creation 🙏

PR #36634, opened 2025-06-24
Docs/self hosted jwt setup

Durganjali-sidda:

YES

Docs update – clarifies JWT configuration steps for self-hosted Supabase environments.

Docs mention JWTs but lack complete explanation for self-hosted setups, which can lead to confusion or token verification failures.

## What is the new behavior?

This PR:
- Adds a dedicated section explaining required JWT claims (`aud`, `sub`, `exp`, etc.)
- Includes `.env` configuration (`JWT_SECRET`)
- Provides developer-friendly context and helpful links (e.g., [jwt.io](https://jwt.io))

## Additional context

Fixes a common issue seen by self-hosting users getting "JWT verification failed" errors.

File updated: `apps/docs/pages/docs/guides/auth/auth-jwts.mdx`

### DIFF
--- a/apps/docs/content/guides/auth/jwts.mdx
+++ b/apps/docs/content/guides/auth/jwts.mdx
@@ -192,15 +192,18 @@ Your tokens should include the following claims:
   "role": "authenticated"
 }
 
-Explanation of claims:
-- `aud`: Audience – typically set to `"authenticated"` for Supabase projects
-- `sub`: Subject – the user's unique ID
-- `iss`: Issuer – identifies who created the token (your app name)
-- `exp`: Expiration time in UNIX timestamp
-- `email`: Optional – used for user identification
-- `role`: Supabase uses this for role-based access
-
 ```
+
+**Explanation of claims:**
+
+- `aud`: Audience – typically set to `"authenticated"` for Supabase projects  
+- `sub`: Subject – the user's unique ID  
+- `iss`: Issuer – identifies who created the token (your app name)  
+- `exp`: Expiration time in UNIX timestamp  
+- `email`: Optional – used for user identification  
+- `role`: Supabase uses this for role-based access  
+
+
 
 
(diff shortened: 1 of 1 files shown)


charislam (commented):
Thanks for the PR @Durganjali-sidda!

Just wondering under what situations you might want to manually build an authenticated JWT, rather than requesting one from the auth service?

For anon and authenticated keys (JWTs), we have the [self-hosted instructions](https://supabase.com/docs/guides/self-hosting/docker#securing-your-services) for generating and I'm wary of duplicating it in another location for maintenance reasons, but maybe I don't quite understand the use case.

Durganjali-sidda:
Hi, thanks for the thoughtful feedback! 🙌

You’re absolutely right that Supabase Auth usually handles issuing JWTs automatically, and in most cases developers shouldn’t need to generate them manually. The intent behind this addition was not to duplicate the self-hosted docs, but to highlight when manual JWTs can be useful and then point developers to the existing guide.

Some example situations where manually creating a JWT is helpful:

Testing & Debugging → quickly minting a token with custom claims to test RLS policies or API calls.

Service-to-Service communication → backend scripts or server-to-server requests that need a token but don’t go through Supabase Auth.

Custom/Third-Party Auth → integrating an existing identity system or migration where JWTs are minted outside Supabase.

I agree we should avoid duplicating the detailed “how to” steps — instead, we could just keep a short note here explaining why you’d want to build your own JWT and then link directly to the self-hosted docs for the actual generation instructions. 
That way:
Developers get immediate clarity about the use case.
We don’t need to maintain multiple sources of truth

aantti:
Not going to merge this one for now. This is an excessive amount of implementation details that a) could be changing, b) fairly straightforward to decipher for someone who really needs it. Overall, user session JWTs should be issued by the Auth service.

PR #36638, opened 2025-06-24
feat: add publishable key in connect dialog

hf:
Adds support for the publishable key (when available) in the Connect dialog. For mobile / native libs it does not show the anon key any more and tells developers to prefer a publishable key (due to the inability to rotate their secret without weekslong downtime).


grdsdev (approved):
Approved mobile related content

charislam (approved):

PR #36646, opened 2025-06-25
fix: warning message on upgrade guards

soedirgo:

YES

Bug fix

When users are blocked from an upgrade because they have objects that need to be dropped, the message in the warning is incorrect. It asks users to "drop these extensions". But they're not extensions, they're user-defined tables, foreign keys, etc. This will cause confusion.

Screenshot 2025-06-20 at 4 36 32 PM

## What is the new behavior?

Here we use a different messaging when users need to drop their objects.

Screenshot 2025-06-25 at 12 42 17 PM

## Additional context

Backend changes are on https://github.com/supabase/infrastructure/pull/23802; this PR needs to be deployed first


joshenlim (approved):
code changes LGTM - unable to properly test on preview as i don't have a project that matches the conditions, but i do have a project with `unsupported_extensions` - verified that it all looks okay

PR #36650, opened 2025-06-25
Chore/fix unified logs

joshenlim:
## Context

Couple of fixes + tweaks to Unified Logs interface

- Fix infinite scrolling logic to fetch next page when scrolled to the bottom (Fixes FE-1671)
- Clean up pagination logic + add de-duping logic for log query results (Fixes FE-1672)
- Fixes toggling column visibility not hiding columns in individual rows (Fixes FE-1689)
- Decouple count loading states from table data loading states
- Fix z-index issue between row and table header
- Deprecate `uuid` param of each row to just be `id`
- Standardise react-query options for all unified logs RQ
- Add missing `OPTIONS` method
- Adjust column widths to better fit a macbook viewport better
- Clean up debug logs + misc other clean up

## To test
Can use Jordi's [supamocka](https://supamocka.vercel.app/) to help generate some logs under "Polling"

- Just need to verify that unified logs fetches more data as you scroll
- Also that the number should tally when scrolled to the bottom
  - Although even if it doesn't it's quite tricky tbh to debug this, more information in `unified-logs-infinite-query.ts`
  - Very open to any suggestions on improving the pagination logic 🙏 
image
- Check that you can toggle column visibility
- Generally doesn't look weird

## What's next

There's still a couple more parts of the interface that needs going through, at the top of my head:

- Facets stuff, checking that all of these work
- Search data table bar
- Verify that "Live mode" works
- Refresh button doesn't seem to be working i think
- Hiding columns isn't currently hiding the column values of each row, only the column header

Note there's one potential problem that I noticed with this data table that we're using and that it's not virtualized. So it's not meant to handle large amounts of data despite having infinite loading logic


apps/studio/components/ui/DataTable/DataTableInfinite.tsx
```diff
@@ -69,118 +71,103 @@ export function DataTableInfinite<TData, TValue, TMeta>({
   }, 'u')
 
   return (
-    <>
```
MildTomato, line 72: could we not do fragment removals. makes the diffs impossible to review

apps/studio/components/interfaces/UnifiedLogs/components/FunctionLogsTab.tsx
```diff
@@ -40,9 +39,8 @@ export const FunctionLogsTab = ({ logs = [] }: FunctionLogsTabProps) => {
                     level={log.level}
                     className="min-w-20"
                   />
-                  {/* <span className="text-foreground">{log.event_type || 'Log'}</span> */}
                 </div>
-                <div className="mt-1 whitespace-pre-wrap break-all pl-2 text-[0.8rem]">
```
MildTomato, line 45: this was deliberate

MildTomato: can we please revert

apps/studio/components/interfaces/UnifiedLogs/components/Columns.tsx
```diff
@@ -163,7 +163,7 @@ export const COLUMNS: ColumnDef<ColumnSchema>[] = [
           {logCount && (
             <Tooltip>
               <TooltipTrigger>
-                <span className="text-foreground-lighter bg-surface-400 px-[6px] py-[2px] h-fit rounded-md text-[0.8rem] leading-none">
```
MildTomato: revert please

MildTomato (changes requested):

apps/studio/components/ui/DataTable/DataTableInfinite.tsx
joshenlim: this is all part of clean up - the fragments are unnecessary
for this part nothing changed in the fragment

apps/studio/components/interfaces/UnifiedLogs/components/FunctionLogsTab.tsx
joshenlim: done!

apps/studio/components/interfaces/UnifiedLogs/components/Columns.tsx
joshenlim: done too 🙏

MildTomato (approved):
lgtm

apps/studio/components/ui/DataTable/DataTableSheetRowAction.tsx
```diff
@@ -55,7 +55,6 @@ export function DataTableSheetRowAction<TData, TFields extends DataTableFilterFi
         return (
           <DropdownMenuItem
             onClick={() => {
-              // FIXME:
```
MildTomato, line 58: lol

apps/studio/components/interfaces/UnifiedLogs/components/Columns.tsx
```diff
@@ -39,9 +39,9 @@ export const COLUMNS: ColumnDef<ColumnSchema>[] = [
     },
     filterFn: (row, columnId, filterValue) => true,
     enableResizing: false,
-    size: 190,
-    minSize: 190,
-    maxSize: 190,
+    size: 130,
```
MildTomato, line 42: hang on, need to check

MildTomato (changes requested):
lgtm

apps/studio/components/interfaces/UnifiedLogs/components/Columns.tsx
joshenlim: the column size changes were mainly to have the columns a bit more fitted for a laptop viewport
otherwise, the data table would be horizontally scrollable from the start for a laptop viewport and the "load more" button + Empty state texts like "No rows" or "Fetching rows" will be off center

apps/studio/components/interfaces/UnifiedLogs/components/Columns.tsx
```diff
@@ -55,16 +55,16 @@ export const COLUMNS: ColumnDef<ColumnSchema>[] = [
...
       return (
-        <div className="absolute right-0 top-1/2 -translate-y-1/2 text-right flex items-center gap-1">
+        <div className="flex items-center justify-end gap-1">
           <LogTypeIcon type={logType} size={16} className="text-foreground/70" />
         </div>
       )
     },
     enableHiding: false,
     filterFn: (row, columnId, filterValue) => true,
-    size: 48,
```
MildTomato, line 65: ?

apps/studio/components/interfaces/UnifiedLogs/components/Columns.tsx
MildTomato: nice. all good, when i was testing on my laptop it looked different initially.
all good. thanks!

MildTomato (approved):
also tested preview

PR #36663, opened 2025-06-25
Update Admonition for AI Assistant for when opt in is re-enabled

joshenlim:
Mainly to raise awareness that we've updated the Assistant (after we've reset their opt in tags), rather than just saying that they're not sending any project metadata

While flag is OFF:
image

While flag is ON:
Note that this admonition texts will switch back to the original once the user has updated the opt in level as least once 
image


saltcod (approved):
Checked locally and looks good!

PR #36668, opened 2025-06-25
feat: temp test api endpoint for bedrock

gregnr:
Adds a temporary API endpoint to test Bedrock in production. This is a pre-req to merging https://github.com/supabase/supabase/pull/35318. We'll revert this after confirming it is working e2e in production.

To test on staging:

```shell
curl 'https://studio-staging-git-feat-test-bedrock-supabase.vercel.app/dashboard/api/ai/temp/bedrock-test' \
-H 'Content-Type: application/json' \
-d '{
    "messages": [
        {
            "role": "user",
            "content": "What is 2+2?"
        }
    ]
}'
```

To test on prod (once this is merged):

```shell
curl 'https://supabase.com/dashboard/api/ai/temp/bedrock-test' \
-H 'Content-Type: application/json' \
-d '{
    "messages": [
        {
            "role": "user",
            "content": "What is 2+2?"
        }
    ]
}'
```

PR #36669, opened 2025-06-25
feat: revert temp bedrock test api

gregnr:
Reverts https://github.com/supabase/supabase/pull/36668 as it was meant to be temporary.


saltcod (approved):
Looks good!

PR #36699, opened 2025-06-26
add cursor rule

jordienr:
Source: https://docs.sentry.io/platforms/javascript/guides/nextjs/

### DIFF
--- a/.cursor/rules/sentry.mdc
+++ b/.cursor/rules/sentry.mdc
@@ -1,0 +1,133 @@
+---
+description: Use this rule for integrating and configuring Sentry error tracking in Next.js projects. It helps set up Sentry correctly, capture server/client errors, and follow best practices for monitoring and debugging in production.
+globs: 
+alwaysApply: false
+---
+These examples should be used as guidance when configuring Sentry functionality within a project.
+
+# Exception Catching
+
+- Use `Sentry.captureException(error)` to capture an exception and log the error in Sentry.
+- Use this in try catch blocks or areas where exceptions are expected
+
+# Tracing Examples
+
+- Spans should be created for meaningful actions within an applications like button clicks, API calls, and function calls
+- Use the `Sentry.startSpan` function to create a span
+- Child spans can exist within a parent span
+
+## Custom Span instrumentation in component actions
+
+- The `name` and `op` properties should be meaningful for the activities in the call.
+- Attach attributes based on relevant information and metrics from the request
+
+```javascript
+function TestComponent() {
+  const handleTestButtonClick = () => {
+    // Create a transaction/span to measure performance
+    Sentry.startSpan(
+      {
+        op: "ui.click",
+        name: "Test Button Click",
+      },
+      (span) => {
+        const value = "some config";
+        const metric = "some metric";
+
+        // Metrics can be added to the span
+        span.setAttribute("config", value);
+        span.setAttribute("metric", metric);
+
+        doSomething();
+      },
+    );
+  };
+
+  return (
+    <button type="button" onClick={handleTestButtonClick}>
+      Test Sentry
+    </button>
+  );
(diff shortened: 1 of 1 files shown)


saltcod (approved):
looks good!

.cursor/rules/sentry.mdc
```diff
@@ -0,0 +1,133 @@
...
+description: Use this rule for integrating and configuring Sentry error tracking in Next.js projects. It helps set up Sentry correctly, capture server/client errors, and follow best practices for monitoring and debugging in production.
+globs: 
+alwaysApply: false
+---
+These examples should be used as guidance when configuring Sentry functionality within a project.
+
+# Exception Catching
+
+- Use `Sentry.captureException(error)` to capture an exception and log the error in Sentry.
+- Use this in try catch blocks or areas where exceptions are expected
```
kamilogorek:
```suggestion
- Use this in try catch blocks or areas where exceptions are expected and not rethrown
```

Otherwise you'll get double-reporting as errors will be handled at the top level as well (or at least should be setup like that)

.cursor/rules/sentry.mdc
```diff
@@ -0,0 +1,133 @@
...
+        name: "Test Button Click",
+      },
+      (span) => {
+        const value = "some config";
+        const metric = "some metric";
+
+        // Metrics can be added to the span
+        span.setAttribute("config", value);
+        span.setAttribute("metric", metric);
+
```
kamilogorek, line 40: You can set those directly in `startSpan` first arg

.cursor/rules/sentry.mdc
```diff
@@ -0,0 +1,133 @@
...
+  );
+}
+```
+
+## Custom span instrumentation in API calls
+
+- The `name` and `op` properties should be meaningful for the activities in the call.
+- Attach attributes based on relevant information and metrics from the request
+
+```javascript
```
kamilogorek, line 59: Useless as fetch calls should be automatically instrumented https://docs.sentry.io/platforms/javascript/guides/nextjs/tracing/instrumentation/automatic-instrumentation/

.cursor/rules/sentry.mdc
```diff
@@ -0,0 +1,133 @@
...
+      return data;
+    },
+  );
+}
+```
+
+# Logs
+
+- Where logs are used, ensure Sentry is imported using `import * as Sentry from "@sentry/nextjs"`
+- Enable logging in Sentry using `Sentry.init({ _experiments: { enableLogs: true } })`
```
kamilogorek, line 78: Just FYI I dont know how logs are billed as they are new feature

.cursor/rules/sentry.mdc
```diff
@@ -0,0 +1,133 @@
...
+- Reference the logger using `const { logger } = Sentry`
+- Sentry offers a consoleLoggingIntegration that can be used to log specific console error types automatically without instrumenting the individual logger calls
```
kamilogorek, line 80: Not sure if you need that rule. Either instrument it now or not.

.cursor/rules/sentry.mdc
```diff
@@ -0,0 +1,133 @@
...
+
+## Configuration
+
+- In NextJS the client side Sentry initialization is in `instrumentation-client.ts`, the server initialization is in `sentry.edge.config.ts` and the edge initialization is in `sentry.server.config.ts`
+- Initialization does not need to be repeated in other files, it only needs to happen the files mentioned above. You should use `import * as Sentry from "@sentry/nextjs"` to reference Sentry functionality
```
kamilogorek, line 85: Why do you need those initialization rules at all? Its done once. Or are those rules for our users?

jordienr:
looking back i think this is unnecessary

PR #36703, opened 2025-06-26
chore: add docs for new db reports charts

fsansalvadore:


Ziinc:
As discussed, we can have an overview page in to Telemetry with a detailed page for each product.
Thoughts? @charislam

charislam (approved):

PR #36710, opened 2025-06-26
Chore/unified logs fixes 02

joshenlim:
Couple more fixes from testing for Unified Logs interface, this round was more focused on the facets part of the UI

## Changes involved

- Temporarily disable Cmd+K shortcut for unified logs top search bar as it conflicts with our existing Cmd+K behaviour (Fixes FE-1688)
- Fix `getLogsCountQuery` WHERE param in the query where methods are being fetched (Fixes FE-1667)
- Fix loading state for refresh button (Fixes FE-1692)
- Update placeholder for top search bar to be "Search logs..."
- Fix Log Type facet labels to replace `_` with a space
- Fix searching "Pathname" (there was an inconsistency between "path" in the SQL query and "pathname" in the UI)
- Searching host and pathname to use `LIKE` instead of `=` to be less strict

## What has been tested so far (Can use this list to test this PR too)
- [x] Levels: Results + count behaviour working as expected by adding / removing filters
- [x] Log type: Results + count behaviour working as expected by adding / removing filters
- [x] Method: Results + count behaviour working as expected by adding / removing filters
- [x] Status Code: Results + count behaviour working as expected by adding / removing filters
- [x] Host: Can search results by host
- [x] Pathname: Can search results by host

## What's remaining to be tested / Will be looking into next

- Auth user filters
- Filtering by time range
- Live mode
- Searching via top search bar
- Wondering what "Info" level is searching for? Do we need it?

### DIFF
--- a/apps/studio/components/interfaces/UnifiedLogs/UnifiedLogs.fields.tsx
+++ b/apps/studio/components/interfaces/UnifiedLogs/UnifiedLogs.fields.tsx
@@ -52,9 +52,10 @@ export const filterFields = [
       return (
         <div className="flex w-full items-center justify-between gap-2 font-mono">
           <span className="capitalize text-foreground/70 group-hover:text-accent-foreground text-xs">
-            {props.label}
+            {props.label.replace('_', ' ')}
           </span>
-          <span className="text-xs text-muted-foreground/70">{props.value}</span>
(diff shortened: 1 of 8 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #36714, opened 2025-06-26
docs: fix RPC example for parameterless calls by adding undefined

### LINKED ISSUE #36713: Correct usage of RPC with no parameters (and options)
# Improve documentation

## Link

[https://supabase.com/docs/reference/javascript/rpc](https://supabase.com/docs/reference/javascript/rpc)

Image

## Describe the problem

The RPC docs don’t clearly state that when calling a function with 'no parameters' and using options (like `{ get: true }`), you must pass 'undefined' as the second argument. Otherwise, the options are treated as a parameter, causing errors.

## Describe the improvement

```js
// Incorrect
const { data, error } = await supabase.rpc('hello_world', { get: true }); // Will error

// Correct
const { data, error } = await supabase.rpc('hello_world', undefined, { get: true });
```

## Additional context

I tested both cases and confirmed that the incorrect version fails with an error, while the correct version returns a successful result.

Image

GAURAV-DEEP01:

YES

docs update

The RPC example doesn’t show that `undefined` must be passed as the second argument when using options (`get: true`) for parameterless RPCs. 
From issue **#36713**

## What is the new behavior?

Updated the example to include `undefined` as the second parameter.

```js
// Incorrect
const { data, error } = await supabase.rpc('hello_world', { get: true }); // Will error

// Correct
const { data, error } = await supabase.rpc('hello_world', undefined, { get: true });
```
---

### DIFF
--- a/apps/docs/spec/supabase_js_v2.yml
+++ b/apps/docs/spec/supabase_js_v2.yml
@@ -4406,7 +4406,7 @@ functions:
         name: Call a read-only Postgres function
         code: |
           ```ts
-          const { data, error } = await supabase.rpc('hello_world', { get: true })
+          const { data, error } = await supabase.rpc('hello_world', undefined, { get: true })
           ```
         data:
           sql: |


charislam (approved):
Thanks @GAURAV-DEEP01!

PR #36723, opened 2025-06-27
Fix database selector not selectable in connect modal

joshenlim:
As per PR title - was a portal issue 😕 

## To test
- [ ] Can click "Create a new read replica" in Connect modal under "Source" dropdown


saltcod (approved):
looks good!

PR #36730, opened 2025-06-27
feat: jwt signing key dialog supports bring your own key

hf:
Adds the ability to choose what JWT signing algorithm and bring your own key.

image

image


ivasilov (approved):

kimsoo527 (commented):
Nice

PR #36732, opened 2025-06-27
feat: enable Self-Service Single Sign-On (SSO) for orgs

saltcod:
## Context

[Joshen] Currently blocked on an identified issue found on staging BE

Introduces the UI for organizations to set up SSO for the organization. BE changes are already on prod so this can go through whenever ready

image


cstockton:
This is ready for final review, summary of changes:
* Expanded default attribute mapping preset for `Okta` to include `user_name`, `first_name`, and `last_name` to match our docs.
* Normalized role casing in the "Join Organization on Signup" dropdown to match the roles expected by the backend (`Owner`, `Administrator`, `Developer`).
* Added the role (`Read-only`) to the selectable roles for auto-join.
* Call update mutation when an `ssoConfig` has been retrieved.
* Treats `404` as a valid "create" flow state rather than an error.
* Remove zod URL restriction on domain validation.
* Conditionally renders the SSO config form when config is successfully loaded *or* when the provider is not found, allowing users to onboard from scratch and disable the provider.

joshenlim (approved):
Generally LGTM - JFYI i've yet to test actually creating an SSO config as I'm not sure how to do so

PR #36759, opened 2025-06-30
Webhooks automatically append authorization header if edge function with verify JWT enabled is selected

joshenlim:
## Context

We've seen a number of tickets for Database Webhooks whereby users tend to forget adding an Authorization header to their webhook if they've selected an edge function that has "Verify JWT" enabled. 

## Changes involved

**_There's a couple of clean ups i've done as well, but the main changes to look at is `FormContents.tsx` from L83 in the useEffect block - that's where the main fix for this PR is_**

Changes in this PR will make it such that the Authorization header will be automatically added to the form if an edge function with "Verify JWT" enabled is selected. On the flip side, if an edge function with "Verify JWT" disabled is selected, we'll filter out any existing Authorization header. In this case we'll be adding the service key for consistency with the existing dropdown action here

image

image

## To test

- [ ] Verify that when selecting an edge function with "Verify JWT" enabled, the authorization header is automatically added
- [ ] Verify that when selecting an edge function with "Verify JWT" disabled, any authorization header is automatically removed


apps/studio/components/interfaces/Database/Hooks/FormContents.tsx
joshenlim, line 1: this is mainly clean up - bringing this out into another file from EditHookPanel

**main changes** however to look at is from L83 in the useEffect block - that's where the main fix for this PR is

alaister (approved):
LGTM!

- Tested on preview

PR #36762, opened 2025-06-30
chore/unified-logs-fixes-04

joshenlim:
## Context

Couple more fixes from testing for Unified Logs interface, this round testing the following:

- Filter by Time range
- Filter Command Bar
- Self-hosted
  - JFYI UI is controlled via feature flag which means that it won't be available on self-hosted just yet until we clean up the feature flag
- RE Auth users: I'm unable to test this on my end as I'm not sure how to generate logs that'll include auth users

## Changes involved

- Fix missing colors for calendar component when a range is selected (Fixes FE-1694)
image
- Prevent selecting future dates from current date in unified logs date picker (Fixes FE-1695)
  - e.g Current date is 30th June, all dates in July are disabled
image
- Small cleanups for DataTableFilterCommand
  - Only show "Query" section if there are any queries
  - Only show "Suggestions" section if there are any suggestions
image
- Add missing POST method to dashboard analytics API to support Unified Logs for self-hosted (Fixes FE-1696)
  - I'm not able to test this locally by the way, running into this error for any pings to the logs pages for self-hosted run via `pnpm run dev:studio-local`
image


packages/ui/src/components/shadcn/ui/calendar.tsx
```diff
@@ -10,6 +10,11 @@ import { buttonVariants } from './button'
 export type CalendarProps = React.ComponentProps<typeof DayPicker>
 
 function Calendar({ className, classNames, showOutsideDays = true, ...props }: CalendarProps) {
+  // [Joshen] Wondering if this is a bug with react-day-picker v8 whereby a single selected date
+  // has styles applied for both day_range_start and day_range_end
+  const fullDateRangeSelected =
+    props.mode === 'range' && !!props.selected?.from && !!props.selected?.to
```
MildTomato, line 16: this doesn't seem right. 
need to investigate

apps/studio/components/ui/DataTable/DataTableFilters/DataTableFilterCommand.tsx
```diff
@@ -227,104 +223,110 @@ export function DataTableFilterCommand({
...
-                      {field.value}
+                      {field.label}
                       <CommandItemSuggestions field={field} />
                     </CommandItem>
                   )
                 })}
               </CommandGroup>
               <CommandSeparator />
-              <CommandGroup heading="Query">
-                {filterFields?.map((field) => {
```
MildTomato, line 238: lets change this to have an empty state instead of hiding entirely

apps/studio/components/ui/DataTable/DataTableFilters/DataTableFilterCommand.tsx
```diff
@@ -370,10 +372,12 @@ export function DataTableFilterCommand({
 function CommandItemSuggestions<TData>({ field }: { field: DataTableFilterField<TData> }) {
   const { table, getFacetedMinMaxValues, getFacetedUniqueValues } = useDataTable()
   const value = field.value as string
+  const className = 'ml-2 hidden truncate text-foreground-lighter group-aria-[selected=true]:block'
+
   switch (field.type) {
     case 'checkbox': {
       return (
-        <span className="ml-1 hidden truncate text-muted-foreground/80 group-aria-[selected=true]:block">
+        <span className={cn(className)}>
```
MildTomato, line 380: need to check

apps/studio/components/ui/DataTable/DatePickerWithRange.tsx
```diff
@@ -98,6 +98,7 @@ export function DatePickerWithRange({
               selected={date}
               onSelect={setDate}
               numberOfMonths={1}
+              disabled={{ after: new Date() }}
```
MildTomato, line 101: seems weird this is hardcoded into component

apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts
```diff
@@ -23,15 +24,26 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+      res.setHeader('Allow', ['GET', 'POST'])
       res.status(405).json({ data: null, error: { message: `Method ${method} Not Allowed` } })
   }
 }
 
 const proxyRequest = async (req: NextApiRequest) => {
   const { name, ...toForward } = req.query
-  const payload = { ...toForward, project_tier: 'ENTERPRISE' }
-  const search = '?' + new URLSearchParams(payload as any).toString()
+  const project_tier = 'ENTERPRISE'
```
Comment (MildTomato) [inline apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts:34]: need const for plan name

apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts
```diff
@@ -23,15 +24,26 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+
+  if (req.method === 'GET') {
+    const payload = { ...toForward, project_tier }
+    return retrieveAnalyticsData(name as string, payload)
+  } else if (req.method === 'POST') {
+    const payload = { ...req.body, project_tier }
+    return retrieveAnalyticsData(name as string, payload)
+  }
+}
```
Comment (MildTomato) [inline apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts:43]: huh

apps/studio/components/ui/DataTable/DataTableFilters/DataTableFilterCommand.tsx
```diff
@@ -85,14 +89,6 @@ export function DataTableFilterCommand({
...
     })
 
-    const commandDisabledFilterKeys = currentDisabledFilters.reduce(
-      (prev, curr) => {
-        prev[curr.id] = curr.value
-        return prev
-      },
-      {} as Record<string, unknown>
-    )
-
```
MildTomato, line 95: wondering why this was removed

apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts
Comment (joshenlim) [reply apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts:43]: This is similar to how we did the logs.all endpoint in the API where we supported both GET and POST endpoints - in which we're just parsing the variables into a payload to retrieve the data from analytics

apps/studio/components/ui/DataTable/DatePickerWithRange.tsx
joshenlim: Hmm good point this since this is a reusable component - could be passed in as props instead from the parent

apps/studio/components/ui/DataTable/DataTableFilters/DataTableFilterCommand.tsx
joshenlim: this isn't being used atm

apps/studio/components/ui/DataTable/DataTableFilters/DataTableFilterCommand.tsx
joshenlim: this is just removing duplication of class names by shifting them into a const, so that it can be reused in each instance below

apps/studio/components/ui/DataTable/DataTableFilters/DataTableFilterCommand.tsx
joshenlim: was thinking to hide this since it only has values after a filter is selected?

apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts
Comment (joshenlim) [reply apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts:34]: this is just a hard code for local/self-host studio so not super important tbh

joshenlim:
Closing in favor of https://github.com/supabase/supabase/pull/37150

PR #36763, opened 2025-06-30
Edge Function Report

jordienr:
Adds new edge function report.

## To test
- go to reports
- go to edge functions
- things make sense


saltcod (approved):
Looks great on staging!

PR #36785, opened 2025-07-01
Change path to pathname in storage unified logs query

joshenlim:
## Context

I've recently updated all `path` properties for the Unified Logs queries to `pathname` for consistency + fixes searching pathname through the DataTableFilterSheet back in this [PR](https://github.com/supabase/supabase/pull/36710). This PR just updates `path` to `pathname` for the newly added storage logs query 

## To test

- [ ] Verify that storage logs can still be viewed and can be searched via the Unified Logs interface
  - I'm not exactly sure how to generate storage logs though, might need to check with Jonny for this. Although can probably try doing some stuff in the storage explorer, see if that generates anything


saltcod (approved):
Works as described on staging

PR #36788, opened 2025-07-01
fix: disk attributes on compute down size

kevcodez:
When downgrading from a larger compute size to <large while having provisioned IOPS or throughput, we would leave the disk as-is, even though the configuration is unsupported and then block any disk changes because the instance size is too small, essentially dead-locking the customer on the disk attributes, unless they bump back up to a higher instance size.

PR addresses multiple issues:
- Automatically reset disk attributes to default when customers downgrade to <large instances
- Do not use the baseline compute IOPS/throughput for the fields, as this leads to accidentally over-provisioning by customers


ecktoteckto (approved):

ecktoteckto:
Looks good to me. Tested on Preview.

PR #36793, opened 2025-07-01
feat: Add duplicate key check with confirmation dialog for Edge Function secrets

laktek:
- Created DuplicateSecretWarningModal component following the pattern of DeployEdgeFunctionWarningModal
- Added duplicate key detection logic to AddNewSecretForm before secret creation
- Shows confirmation dialog when attempting to create a secret with an existing name
- Allows users to proceed with replacing the existing secret or cancel the operation

🤖 Generated with [Claude Code](https://claude.ai/code)

https://github.com/user-attachments/assets/611d2a74-2f1b-4812-b6d8-9f9be131ffaf


jordienr (approved):
lgtm tested in preview

PR #36795, opened 2025-07-01
Review and merge branch

SaxonF:
image

Adds the ability to review and merge preview branches. 

To test:
- Create new project
- Enable gitlessBranching flag (should be enabled by default)
- Create two new gitless branches
- Switch to first branch and add a bunch of tables
- Add an edge function with multiple files
- Click branch dropdown and "review changes"
- Make sure changes are visible in database diff
- Merge and see merge status/logs
- On completion make sure there is no difference between branch and prod
- Switch to the other preview branch and add a table column
- Visit review changes page and notice you need to pull latest changes
- Pull changes and then merge
- Switch back to the other preview branch and add the same column then attempt to pull latest changes via review page and notice failed merge
- Make some more edge function changes
- Disable gitlessBranching flag and make sure merge page is not accessible


alaister (approved):
LGTM!

PR #36836, opened 2025-07-02
format data to ms in execution time - edge fn report

jordienr:
## To test
- go to edge functions report
- go to execution time
- hover the chart
- shows numbers as ms (ex: 320ms)


joshenlim (approved):
Verified changes locally - got a cron job that triggers edge function running perpetually 😬 😄 
image

joshenlim:
Small feedback here: There's a bit of duplicate information in the tooltip
image

PR #36837, opened 2025-07-02
Distributes bedrock requests evenly across regions

gregnr:
Bedrock has per-region inference limits which we are hitting. This distributes requests evenly across regions by hashing the project ref.


saltcod (approved):
Checked each region and works!

PR #36851, opened 2025-07-03
remove results from message

SaxonF:
Removes results property if it exists before messages are sent to server


alaister (approved):
LGTM!

- Tested on preview

PR #36866, opened 2025-07-03
feat(replication): Update the API calls for replication endpoints

iambriccardo:
This PR updates the API endpoints calls for the new replication feature.


apps/studio/pages/project/[ref]/database/replication.tsx
```diff
@@ -25,16 +24,18 @@ const DatabaseReplicationPage: NextPageWithLayout = () => {
...
-          </ScaffoldSection>
-        </ScaffoldContainer>
+        <>
+          <ScaffoldContainer>
+            <ScaffoldHeader>
+              <ScaffoldTitle>Replication</ScaffoldTitle>
+              <ScaffoldDescription>Send data to other destinations</ScaffoldDescription>
+            </ScaffoldHeader>
+          </ScaffoldContainer>
+          <ReplicationComingSoon />
```
Comment (iambriccardo) [inline apps/studio/pages/project/[ref]/database/replication.tsx]: We keep the replication coming soon only when the flag is not set.

apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
```diff
@@ -115,10 +115,11 @@ const DestinationPanel = ({
       name: destinationData?.name ?? '',
       projectId: destinationData?.config?.big_query?.project_id ?? '',
       datasetId: destinationData?.config?.big_query?.dataset_id ?? '',
-      serviceAccountKey: destinationData?.config?.big_query?.service_account_key ?? '',
-      publicationName: pipelineData?.publication_name ?? '',
-      maxSize: pipelineData?.config?.config?.max_size ?? 1000,
-      maxFillSecs: pipelineData?.config?.config?.max_fill_secs ?? 10,
+      // For now, the password will always be set as empty for security reasons.
+      serviceAccountKey: '',
```
imor: What happens if the user edits the destination? This won't overwrite the serviceAccountKey in the API to an empty string when the user saves right?

apps/studio/pages/project/[ref]/database/replication.tsx
```diff
@@ -35,7 +33,7 @@ const DatabaseReplicationPage: NextPageWithLayout = () => {
               <ScaffoldDescription>Send data to other destinations</ScaffoldDescription>
             </ScaffoldHeader>
           </ScaffoldContainer>
-          <ReplicationComingSoon />
+          <Destinations />
```
Comment (imor) [inline apps/studio/pages/project/[ref]/database/replication.tsx]: `ReplicationComingSoon` should stay.

imor (dismissed):
Left a couple of comments, otherwise LGTM.

apps/studio/pages/project/[ref]/database/replication.tsx
Comment (iambriccardo) [reply apps/studio/pages/project/[ref]/database/replication.tsx]: Good catch, I screwed this up during merge.

apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
iambriccardo: Yeah, this is a problem indeed, I initially thought that we would always have to re-submit the key for security reasons but it is not a good UX. Either we return it from the endpoint and pre-fill it like before or we allow the key to be optional.

apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
```diff
@@ -404,7 +405,7 @@ const DestinationPanel = ({
                                     <Input_Shadcn_
                                       {...field}
                                       type="number"
-                                      {...form.register('maxFillSecs', {
+                                      {...form.register('maxFillMs', {
```
ivasilov, line 408: small nit: `form.register('maxFillMs')` is the same as `field` in these props so it's not needed.

ivasilov (approved):

apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
imor: Worth checking what UX patterns do we follow elsewhere in the UI. E.g. how do we handle the database password, IMO we can use similar pattern here as well.

PR #36868, opened 2025-07-03
feat: add billing information to the leave org modal

ignaciodob:
This PR improves the confirmation modal shown when an Owner or Admin leaves an organization.
The existing modal warns about the deletion of user content but does not clearly explain that the user’s credit card may remain active or continue to be charged.

### Why
This aims to reduce billing-related support tickets from users who assume that leaving an organization removes their credit card or stops charges. It will also hopefully protects users from surprise charges when they no longer have access to the org.

### Changes
- Added a role check for the Administrator role to only show the billing information to Owners and Administrators.
- Added the relevant information to the confirmation modal.

| Before  | After |
| Screenshot 2025-07-03 at 3 35 05 PM  | Screenshot 2025-07-03 at 3 34 30 PM |

### Testing
- Head over to `http://localhost:8082/org/_/team`
- Pick an org with two or more users
- With an user with the `Owner` or `Admin` role, click on `Leave team`
- You should see the new billing information in the confirmation modal
- Repeat the same process with a user with the `Developer` role
- You should **not** see the billing information in the confirmation modal

### DIFF
--- a/apps/studio/components/interfaces/Organization/TeamSettings/LeaveTeamButton.tsx
+++ b/apps/studio/components/interfaces/Organization/TeamSettings/LeaveTeamButton.tsx
@@ -37,8 +37,13 @@ export const LeaveTeamButton = () => {
   const { data: members } = useOrganizationMembersQuery({ slug })
   const { data: allRoles } = useOrganizationRolesV2Query({ slug })
 
-  const isOwner = selectedOrganization?.is_owner
   const roles = allRoles?.org_scoped_roles ?? []
+  const currentUserMember = members?.find((member) => member.gotrue_id === profile?.gotrue_id)
+  const currentUserRoleId = currentUserMember?.role_ids?.[0]
+  const currentUserRole = roles.find((role) => role.id === currentUserRoleId)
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Organization/TeamSettings/LeaveTeamButton.tsx
```diff
@@ -106,6 +111,23 @@ export const LeaveTeamButton = () => {
                 <li>Custom reports</li>
                 <li>Log Explorer queries</li>
               </ul>
+              {(isOwner || isAdmin) && (
+                <>
+                  <br />
```
kevcodez: Rather than a break, can wrap it in a div (replace <> with <div>) and add some margin to the top i.e. `mt-2` css class

kevcodez (approved):

apps/studio/components/interfaces/Organization/TeamSettings/LeaveTeamButton.tsx
ignaciodob: Much better! Thank you

PR #36881, opened 2025-07-04
fix: remove apac bedrock region and add us region

delgado3d:


gregnr (approved):
LGTM pending test fix

alaister (approved):

PR #36907, opened 2025-07-07
Assistant error state

SaxonF:
Gives the user a way to retry a message if it fails.

To test, fake an error response from generate-v4


alaister (approved):
LGTM!

- Tested on preview by blocking the `/generate-v4` endpoint to cause an error

PR #36908, opened 2025-07-07
Revert "fix: remove apac bedrock region and add us region"

delgado3d:
Reverts supabase/supabase#36881


darora:
Am assuming this is not stale, please re-open if required

PR #36917, opened 2025-07-07
chore: add final e2e tests for table-editor and sql-editor

Michaelomh:

YES/NO

add more e2e test for table editor and sql editor. 
image

image

**Tests all passing for staging**
image

## What is the new behavior?

## Additional context


e2e/studio/utils/is-local.ts
```diff
@@ -0,0 +1,9 @@
+import { env } from '../env.config'
+
+export function isLocal(url?: string): boolean {
```
jordienr: Studio can be run in "hosted" mode locally too, so I think this should check the value of IS_PLATFORM instead.

IS_PLATFORM=true // hosted mode
IS_PLATFORM=false // cli/selfhosted mode

Think you could rename this to "isCLI" too.

jordienr (changes requested):

jordienr (approved):

PR #36919, opened 2025-07-07
fix(branches): update branch name with git_branch changes

avallete:

YES/NO

## What is the new behavior?

## Additional context


sweatybridge (changes requested):
I don't think this is the right change. We should keep the branch name decoupled from git branch. The current behaviour is correct as our convention for the production branch name is always `main`, regardless of how user changes their git branch.

PR #36944, opened 2025-07-07
Update `@deprecated` suggestions for `ui` package  components

Saeris:

YES

This updates the deprecation notices for components in the `ui` package to follow this format:

```
/** @deprecated Use `import { <Component_Name> } from "ui"` instead */
```

The intent here is to be more explicit about what import a developer should be using instead to replace the deprecated component. Most of these notices were using a confusing relative path segment, which made it unclear as to which library the suggested replacement was coming from.

## Additional context

The idea for this came up while preparing to do some refactoring tasks assigned to me by @saltcod

Figured what would help me would help others


jordienr (approved):

MildTomato (approved):

Saeris:
since opening this I think there's more places where I can improve the import hints but I'll follow up with another PR somewhere down the road if necessary

thanks for the approvals!

PR #36948, opened 2025-07-08
Updates Assistant snippet handling

SaxonF:
image

Updates the styling of snippets and moves combining of message into the chat form component.


apps/studio/components/ui/AIAssistantPanel/AssistantChatForm.tsx
```diff
@@ -0,0 +1,174 @@
...
+      <div className="w-full">
+        <form
+          id="assistant-chat"
+          ref={formRef}
+          {...props}
+          onSubmit={handleSubmit}
+          className={cn('relative overflow-hidden', className)}
+        >
+          {/* Snippet badges row */}
+          {sqlSnippets && sqlSnippets.length > 0 && (
```
ivasilov, line 96: This `snippet badges row` should be extracted into a component and used where needed. There's no other changes to `AssistantChatFormComponent` when compared to `AssistantChatForm`, right?

apps/studio/components/ui/AIAssistantPanel/AIOnboarding.tsx
```diff
@@ -33,51 +32,42 @@ export const AIOnboarding = ({
       title: 'Create a back-end',
```
ivasilov: Take the `defaultPrompts` out of the component, they're causing rerenders because a new array is created each render.

apps/studio/components/ui/AIAssistantPanel/AssistantChatForm.tsx
SaxonF: You mean the old ui-patterns form @ivasilov ? No. i've moved it out of that package because it's not used anywhere else outside of Assistant. Will remove ui-patterns component in follow up.

alaister (approved):
LGTM!

- Tested on preview

PR #36949, opened 2025-07-08
fix: show run logs for gitless branches

sweatybridge:

Bug fix and removal of old GitHub side panel which is no longer in use

## What is the new behavior?

Regardless of whether there's a connected github repo, we should always allow viewing the branch merge logs.
Screenshot 2025-07-08 at 1 54 37 PM

## Additional context

supersedes https://github.com/supabase/supabase/pull/36919


apps/studio/components/interfaces/BranchManagement/BranchPanels.tsx
```diff
@@ -210,32 +210,51 @@ export const BranchRow = ({
...
+            <DropdownMenu>
+              <DropdownMenuTrigger asChild>
+                <Button type="text" icon={<MoreVertical />} className="px-1" />
+              </DropdownMenuTrigger>
+              <DropdownMenuContent className="p-0 w-56" side="bottom" align="end">
+                {repo ? (
+                  <Link passHref href={`/project/${projectRef}/settings/integrations`}>
+                    <DropdownMenuItem asChild className="gap-x-2">
+                      <a>Change production branch</a>
+                    </DropdownMenuItem>
```
sweatybridge, line 229: @SaxonF may be we can remove this eventually to use the same tooltip for editing branch name, with a big warning when user wants to change the associated git branch.

alaister (approved):
LGTM!

- Tested on preview

PR #36654, opened 2025-06-25
feat: show upgrade warning if user objects exist in internal schemas

soedirgo:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

feature

## What is the current behavior?

n/a

## What is the new behavior?

Show upgrade warning if user-defined objects exist in auth/realtime/storage. This is done to nudge users to follow the changes in https://github.com/orgs/supabase/discussions/34270

Screenshot 2025-06-25 at 5 59 42 PM

PR #36954, opened 2025-07-08
chore: clean up removed field handling

soedirgo:
Follow up to https://github.com/supabase/supabase/pull/36646 and https://github.com/supabase/supabase/pull/36654


CoderNasim2023 (changes requested):
I think its okk and will be execute  smoothly

CoderNasim2023 (commented):
I think its okk and will be execute  smoothly

ivasilov (approved):

PR #36972, opened 2025-07-09
Refactor `NewAccessTokenButton` to use shadcn components

Saeris:

YES

This brings `NewAccessTokenButton` into conformance with the latest UI patterns of using shadcn based components, zod schemas, and react-hook-form for form submission.

I based this refactor on the implementation of `CreateSecretAPIKeyDialog`, which already made use of the above.

I'll be leaving this as a draft until I've added some associated tests and have gotten feedback on slight differences in appearance and whether it should include success/failure toasts.

Here is a comparison of existing (prod) vs new (sandboxed):

image

## Additional context


Saeris:
Alright, in working on the tests for this I came across a bug in the testing setup! This one took way longer than I would like to admit to figure out, but fortunately there was a simple fix. Here's the gist:

Because `restoreMocks: true` was set in the studio's vitest configuration, this meant that the mock implementation for `window.matchMedia` defined in the `tests/vitestSetup.ts` file was being **reset** to an empty function before each test suite was run.

While this didn't appear to be a problem before, that appears to be because none of the existing tests had `motion.div` in their component trees.

Since `motion.div` calls `addListener` on a media query as part of it's lifecycle, that meant that any test which included it would fail as `addListener` would be undefined in this situation. Removing `restoreMocks: true` from the config results in the desired mocking behavior.

This change is necessary in order to test any component that has the `FormItemLayout` component, as it uses `framer-motion` to animate in error messages for input elements.

Saeris:
Added some more tests after evaluating whether we could use [`msw-auto-mock`](https://github.com/zoubingwu/msw-auto-mock) (short answer: no).

Going to mark this ready for review, but wanted to add a few more notes on small changes:

This fixes an issue with the polyfills for the testing environment, where a call to `useMutation` would fail as a result of `TransformStream` being set to `null`.

Basic tests for access token creation and form resetting added. Adds `@faker-js/faker` to the studio app devDependencies to generate mock data for `msw` endpoint response.

Adds [`shx`](https://github.com/shelljs/shx) to the docs app devDependencies to ensure that the `codegen:examples` script runs cross-platform. I develop on Windows, and ran into errors running `pnpm dev`, so this is a fix for that.

apps/studio/vitest.config.ts
```diff
@@ -24,8 +24,6 @@ export default defineConfig({
   test: {
     globals: true,
     environment: 'jsdom', // TODO(kamil): This should be set per test via header in .tsx files only
-    include: [resolve(dirname, './**/*.test.{ts,tsx}')],
```
Saeris: context for removing this line is that I use Wallabyjs to run tests while I'm writing them, and for whatever reason this was causing both Wallaby and Vitest to fail to find any tests in studio on Windows. appears unnecessary to me

apps/studio/vitest.config.ts
```diff
@@ -24,8 +24,6 @@ export default defineConfig({
...
-    restoreMocks: true,
```
Saeris: this causes polyfills for matchMedia to be reset before tests are even run, so it has to be left to the default `false` for components with framer-motion to render

apps/studio/tests/lib/msw.ts
```diff
@@ -1,25 +1,66 @@
...
   method,
   path,
   response,
 }: {
-  method: keyof typeof http
-  path: string
-  response: any
+  method: M
+  path: P
+  response: SuccessResponse<TrimQueryParams<P>, M> | HttpResponseResolver
```
Saeris: the added types here cause addAPIMock to behave like the fetch client, which gets type information generated from the openapi schemas. so the idea here is that we can surface the valid return types for a given endpoint and request method to aid in mocking responses and highlight when a change has been made that might break a test

apps/docs/package.json
```diff
@@ -149,6 +149,7 @@
     "openapi-types": "^12.1.3",
     "postcss": "^8.5.3",
     "shiki": "^3.2.1",
+    "shx": "^0.4.0",
```
Saeris, line 152: this was added to work around `codegen-examples` not running on Windows

apps/studio/package.json
```diff
@@ -189,6 +190,7 @@
     "eslint-plugin-barrel-files": "^2.0.7",
     "graphql-ws": "5.14.1",
     "import-in-the-middle": "^1.13.1",
+    "jsdom-testing-mocks": "^1.13.1",
```
Saeris: this was added to polyfill missing web animation APIs needed for tests in another PR. I rebased all of my in-flight PRs off of a branch containing just the testing setup changes, so that's why this appears here

apps/studio/tests/features/logs/LogsPreviewer.test.tsx
```diff
@@ -32,7 +32,7 @@ vi.mock('lib/gotrue', async (importOriginal) => ({
 beforeEach(() => {
   addAPIMock({
     method: 'get',
-    path: '/platform/projects/default/analytics/endpoints/logs.all',
+    path: '/platform/projects/:ref/analytics/endpoints/logs.all',
```
Saeris: with the type changes to `addAPIMock` the path should always use the path parameter syntax rather than the literal in order to get the response type. while this won't technically have a runtime error, it will raise a type error

apps/studio/components/interfaces/Account/__tests__/NewAccessTokenButton.test.tsx
joshenlim, line 1: generally don't need to create a separate folder for tests - we typically just keep test files beside where their component is

so having this in components/interfaces/Account/NewAccessTokenButton.test.tsx is okay

apps/studio/components/interfaces/Account/__tests__/NewAccessTokenButton.test.tsx
```diff
@@ -0,0 +1,96 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest'
+import { screen, waitFor } from '@testing-library/dom'
+import userEvent from '@testing-library/user-event'
+import { faker } from '@faker-js/faker'
```
joshenlim: reckon we don't need to introduce a new library - we can just hard code the API response mock

joshenlim: actually no worries - should be fine since its just a dev dependency

apps/docs/package.json
joshenlim: gotcha - tested working alright on macos

joshenlim:
Found an issue here: will push up a fix
image

apps/studio/components/interfaces/Account/NewAccessTokenButton.tsx
```diff
@@ -1,158 +1,185 @@
...
-                  <Button type="default" disabled={isLoading} onClick={() => setIsOpen(false)}>
-                    Cancel
-                  </Button>
-                  <Button htmlType="submit" loading={isLoading} disabled={isLoading}>
-                    Generate token
-                  </Button>
+        setVisible(open)
+      }}
+    >
+      <DialogTrigger asChild>
```
joshenlim: Hmm in this case we probably don't need DialogTrigger i reckon

joshenlim: Yeah we if we're controlling the Dialog with an external state this way, i reckon the buttons could be rendered outside the Dialog component actually

staaldraad (approved):

pamelachia (approved):

joshenlim (approved):
Verified creating tokens + experimental tokens on staging preview

PR #36976, opened 2025-07-09
format edgefn report exec time chart

jordienr:
the execution time chart will now show average time for all functions unless the user has selected one or more functions in the filter.


joshenlim (approved):
Verified changes locally - LGTM 🙂

PR #36992, opened 2025-07-09
reports pricing

jordienr:
Testing:
- try with different orgs (free, pro, team)
- try to select time ranges higher than current org limit → shows upselling dialog
- metrics are disabled for your tier and show upselling dialog

CleanShot 2025-07-14 at 12 20 00@2x
CleanShot 2025-07-14 at 12 20 50@2x
CleanShot 2025-07-14 at 12 21 47@2x


apps/studio/data/reports/database-charts.ts
```diff
@@ -148,7 +148,7 @@ export const getReportAttributesV2: (org: Organization, project: Project) => Rep
       id: 'disk-iops',
       label: 'Disk Input/Output operations per second (IOPS)',
       syncId: 'database-reports',
-      hide: false,
+      hide: isFreePlan,
```
kevcodez: Are these actually hidden or upsell? We should not hide any elements that are gated behind a plan but rather show what a free plan customer is missing out

jordienr: Hides it.

For cases where the metric is only available in paid tiers, we show the upselling state. But in some cases we need to hide it to replace it with the same metric slightly different (ex: showing totals in free, showing grouped by in paid)

fsansalvadore (approved):

ivasilov (approved):
LGTM

PR #36997, opened 2025-07-09
Add missing permissions text for AI opt in modal

joshenlim:
## Context

Realized that for the AI Opt-in modal specifically that's toggled from the SQL Editor, there's no helper text indicating while the options are disabled if the user doesn't have the necessary permissions for it (based on the role that the user has in the organization)

## Changes involved

- Add missing helper text to indicate disabled reason
image


alaister (approved):
LGTM!

PR #37010, opened 2025-07-10
Add migration status to branch

SaxonF:
image

Adds status on branches that shows whether migrations and functions have been deployed. When you create a branch, it takes a while for migrations and functions to deploy and we currently don't indicate that they are yet to run, so the project looks healthy.


alaister (approved):
LGTM!

- Tested on preview

PR #37018, opened 2025-07-10
Merge complete loop

SaxonF:
Currently when a branch merge completes we update the branch and set the `review_requested_at` field to null via `requestReview`. This causes re-renders which recreates callback and creates a loop. This change uses useCallback and adds review_requested_at check before updating branch.

To test:
1. Make sure gitlessBranching flag is enabled
2. Create branch
3. Make a change on branch
4. Create a branch review and visit /merge
5. Click merge and see workflow logs
6. Wait until completion and check if can see completion state without error


alaister (approved):
LGTM!

- Tested on preview

PR #37024, opened 2025-07-10
feat(replication): Improve status display of the pipeline

iambriccardo:
This PR improves the pipeline status display in the following ways:
- More statuses are displayed with animations.
- The `RequestStatus` is now an enum.
- The pipeline status refreshes every 5 seconds continuously, as errors may occur after some time, even after the pipeline has started.

Currently, the statuses still require backend improvements, but this is a solid starting point.


imor (approved):
LGTM

ivasilov (approved):
LGTM!

PR #37026, opened 2025-07-10
feat: add Realtime private only feature flag

w3b6x9:
[Joshen] Copy and pasting over the description from the [previous PR](https://github.com/supabase/supabase/pull/37040) into here since its all the same changes

## Context
Some updates to what Wen Bo has already done in this [PR](https://github.com/supabase/supabase/pull/37026), branching out so that the changes are easier to check.

Main thing that's changing here is that
- Visibility of realtime settings page is now controlled via feature preview, instead of feature flag
- Feature preview of realtime settings is controlled via the feature flag

image

## Changes involved
- Refactored realtime settings to be available via feature preview
  - Visiblity of this feature preview is controlled via feature flag
  - Only available for hosted (via `isPlatformOnly` property in `FeaturePreview.constants.ts` - afaik realtime settings aren't available yet for self-hosted or CLI, so this feature preview shouldn't be visible in that env too)
- Refactored feature flag for showing realtime settings into a separate hook
  - Considers both `isRealtimeSettingsEnabledOnProjects` and `enableRealtimeSettings` feature flags
- Added image and updated title + description for realtime settings feature preview
- Clean up some old feature preview images

## To test
- [ ] Verify that before enabling the feature flag, you shouldn't be able to see realtime settings in the dashboard
- [ ] Verify that you can see the feature preview for realtime settings (feature flag is currently toggled on for staging)
- [ ] Verify that you cannot see the feature preview for realtime settings if the feature flag is off
- [ ] Verify that you can access realtime settings after enabling the feature preview


apps/studio/components/interfaces/App/FeaturePreview/FeaturePreview.constants.tsx
```diff
@@ -1,6 +1,13 @@
 import { LOCAL_STORAGE_KEYS } from 'common'
 
 export const FEATURE_PREVIEWS = [
+  {
+    key: LOCAL_STORAGE_KEYS.REALTIME_PRIVATE_ONLY_MODE,
+    name: 'Enable Realtime private only mode',
+    discussionsUrl: '',
```
w3b6x9: @filipecabaco

apps/studio/components/interfaces/App/FeaturePreview/RealtimePrivateOnlyModePreview.tsx
```diff
@@ -0,0 +1,19 @@
+import Image from 'next/image'
+
+import { BASE_PATH } from 'lib/constants'
+
+export const RealtimePrivateOnlyModePreview = () => {
```
w3b6x9: @filipecabaco

joshenlim (approved):
Verified changes locally (for hard coding toggling off of feature flag) and staging (for toggling visibility of realtime settings via feature preview)

PR #37042, opened 2025-07-11
fix(docs): move prerequisite on top

avallete:

Move pre-requisites instructions before the actual guide rather than after.


ChrisChinchilla (approved):
Generally I'd like to improve some of the wording in general here, but this is a good fix to start :)

PR #37048, opened 2025-07-11
fix(api): set x-pg-application-name for dashboard

avallete:
- Add a  `x-pg-application-name` header to postgres-meta calls to identify the origin on the query on database

Should be merged after: https://github.com/supabase/infrastructure/pull/24189


alaister (approved):
LGTM!

- Tested on preview

PR #37049, opened 2025-07-11
feat: distribute bedrock requests per regional limits

delgado3d:
Currently we distribute Bedrock requests across multiple regions equally, but in reality each region has different rate limits. This updates the routing logic to account for weights when distributing requests.


SaxonF (approved):
Tested locally and on staging across assistant, inline assistant and cron.

PR #37092, opened 2025-07-14
download as migration

SaxonF:
image

Adds a download as migration button to schema diff.


joshenlim (approved):
Verified changes on staging preview by downloading the migration file

PR #37093, opened 2025-07-14
remove duplicate download

SaxonF:
removes duplicate download as png


joshenlim (approved):

PR #37094, opened 2025-07-14
chore: hint for compute costs in upgrade screen

kevcodez:
- Remove random tweet to focus the user on what's important in the upgrade modal (already lots of info)
- Add compute cost hint if upgrading to Pro Plan with one project (common refund scenario)

Screenshot 2025-07-14 at 13 53 06


staaldraad (approved):

ecktoteckto (approved):
Looks good. Tested on staging.

PR #37095, opened 2025-07-14
Dynamic model setting

SaxonF:
Updates to Haiku 3.5 and makes use of env variables for model setting. Feature flags should be used once we support server-side.

### DIFF
--- a/apps/studio/lib/ai/model.ts
+++ b/apps/studio/lib/ai/model.ts
@@ -16,7 +16,7 @@ export const regionMap = {
 
 const IS_THROTTLED = process.env.IS_THROTTLED !== 'false'
 const PRO_MODEL = process.env.AI_PRO_MODEL ?? 'anthropic.claude-3-7-sonnet-20250219-v1:0'
-const NORMAL_MODEL = process.env.AI_NORMAL_MODEL ?? 'meta.llama4-maverick-17b-instruct-v1:0'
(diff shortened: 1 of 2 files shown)


apps/studio/lib/ai/model.ts
```diff
@@ -14,8 +14,9 @@ export const regionMap = {
   eu: 'eu',
 }
 
-const SONNET_MODEL = 'anthropic.claude-3-7-sonnet-20250219-v1:0'
-const HAIKU_MODEL = 'anthropic.claude-3-haiku-20240307-v1:0'
+const IS_THROTTLED = process.env.IS_THROTTLED !== 'false'
```
joshenlim, line 18: to make it more clear hahaha

```suggestion
const IS_THROTTLED = process.env.IS_THROTTLED === 'true'
```

SaxonF: @joshenlim this was intentional as a safety precaution so that it defaults to false always

joshenlim: Ohhh okay I think I understand - its intended that the default behaviour is _not throttled_ yeah?

SaxonF: default behaviour should be throttled

SaxonF: this is really just a temporary solution. throttling should by dynamic and we have the isThrottled prop for this, its just not in use right now as we experiment with different models

joshenlim (approved):
Verified changes locally

PR #37101, opened 2025-07-14
fix find by id in vault

jordienr:
CleanShot 2025-07-14 at 12 48 02@2x

- try to find a secret by id
- it should work


joshenlim (approved):
Verified changes in staging preview vault interface

PR #37105, opened 2025-07-14
feat(replication): Implement new error UI when the pipeline fails

iambriccardo:
This PR implements a new error UI for when the pipeline fails unexpectedly. This change is connected to backend changes which expose this new error information, for this reason, this PR should be merged **after** those PRs are in.

### DIFF
--- a/apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
+++ b/apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
@@ -6,7 +6,10 @@ import ShimmeringLoader from 'ui-patterns/ShimmeringLoader'
 import RowMenu from './RowMenu'
 import PipelineStatus, { PipelineStatusRequestStatus, PipelineStatusName } from './PipelineStatus'
 import { useParams } from 'common'
-import { useReplicationPipelineStatusQuery } from 'data/replication/pipeline-status-query'
+import {
+  ReplicationPipelineStatusData,
+  useReplicationPipelineStatusQuery,
+} from 'data/replication/pipeline-status-query'
 import { useState } from 'react'
 import { toast } from 'sonner'
 import { useStartPipelineMutation } from 'data/replication/start-pipeline-mutation'
@@ -65,11 +68,14 @@ const DestinationRow = ({
   const { mutateAsync: startPipeline } = useStartPipelineMutation()
   const { mutateAsync: stopPipeline } = useStopPipelineMutation()
   const pipelineStatus = pipelineStatusData?.status
-  const getStatusName = (status: any) => {
+  const getStatusName = (
+    status: ReplicationPipelineStatusData['status'] | undefined
+  ): string | undefined => {
     if (status && typeof status === 'object' && 'name' in status) {
       return status.name
     }
-    return status
+
+    return undefined
   }
 
   const statusName = getStatusName(pipelineStatus)
--- a/apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
+++ b/apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
@@ -4,7 +4,7 @@ import { ReplicationPipelinesData } from 'data/replication/pipelines-query'
 import { ResponseError } from 'types'
 import ShimmeringLoader from 'ui-patterns/ShimmeringLoader'
 import RowMenu from './RowMenu'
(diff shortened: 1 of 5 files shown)


apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
```diff
@@ -65,10 +65,19 @@ const DestinationRow = ({
...
+  }
+
+  const statusName = getStatusName(pipelineStatus)
   if (
     (requestStatus === PipelineStatusRequestStatus.EnableRequested &&
-      pipelineStatus === 'Started') ||
-    (requestStatus === PipelineStatusRequestStatus.DisableRequested && pipelineStatus === 'Stopped')
+      (statusName === PipelineStatusName.STARTED || statusName === PipelineStatusName.FAILED)) ||
+    (requestStatus === PipelineStatusRequestStatus.DisableRequested &&
+      (statusName === PipelineStatusName.STOPPED || statusName === PipelineStatusName.FAILED))
```
iambriccardo, line 86: When we reach a terminal state, we want to show it.

apps/studio/components/interfaces/Database/Replication/PipelineStatus.tsx
```diff
@@ -28,6 +52,175 @@ const PipelineStatus = ({
...
+      status.name === PipelineStatusName.FAILED
+    )
+  }
+
+  const isLogLoadingState = (failedStatus: FailedStatus): boolean => {
+    // Right now we hardcode the error message which is returned when k8s is not able to find logs, which seems
+    // to be a transient error. In case we find a way to properly handle this in the backend, this hack will not
+    // be needed anymore.
+    return (
+      failedStatus.message?.startsWith('unable to retrieve container logs for containerd://') ===
```
imor, line 99: We do not want to expose the fact that we are running containers. Users might not even care, so the error could be more general "failed to fetch logs" or something.

apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
```diff
@@ -65,10 +65,19 @@ const DestinationRow = ({
...
   const { mutateAsync: stopPipeline } = useStopPipelineMutation()
   const pipelineStatus = pipelineStatusData?.status
+  const getStatusName = (status: any) => {
+    if (status && typeof status === 'object' && 'name' in status) {
+      return status.name
+    }
+    return status
+  }
+
+  const statusName = getStatusName(pipelineStatus)
```
imor, line 81: Why has `pipelineStatus` a type of `any`?

imor (approved):
Error messages can avoid mentioning k8s specific terms like containers. Otherwise LGTM for now, but we'd need to take input from the design team how to best expose these errors in the UI (might have to redirect users to the logs). Also, in general it helps the frontend team to review if we add screenshots of the changes to the PR because just looking at the code doesn't give an indication of the visual changes.

apps/studio/components/interfaces/Database/Replication/PipelineStatus.tsx
iambriccardo: This is a limitation since this error is thrown by the K8S client itself, what we can do to hide this is to actually return `loading` from the backend when the error is this one and the frontend just spins when loading is true.

imor: I don't think this is about where we put this logic, it more generally sounds like a losing battle to rely on error message contents. This whole approach about how to show the status needs to be rethought/revisited.

I just checked how the project status button (see below screenshot) shows status for other services and it's a healthcheck call. We'll mostly likely have to expose a health check endpoint from inside the replicator.

image

iambriccardo: Happy to discuss any improvements. If you want we can drop entirely the error message and just show that it failed and maybe prompt to check logflare? Wdyt?

imor: We can merge this PR, since this will be useful for debugging on staging, but iterate on the design in new PRs.

apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
iambriccardo: Fixed.

ivasilov (approved):

PR #37109, opened 2025-07-14
Improve loading and success handlers for JWT keys

joshenlim:
## Context

Slight confusion for JWT keys when shifting standby keys between previously used and back to standby and there's no loading indicator. 

## Changes involved
- Add loading indicator to SigningKeyRow when shifting from standby to previously used and vice versa
- Add success toasts when the key is successfully shifted from standby to previously used and vice versa
- Add tooltip for disabled dropdown item when trying to shift a key from previously used to standby there's already a standby key
image


alaister (approved):
LGTM!

- Tested on preview

PR #37127, opened 2025-07-14
feat: hybrid search

charislam:
Implement hybrid search for the /docs/api/graphql searchDocs endpoint. Prepend a more descriptive title and introduction to database advisor docs so they rank more highly when directly searched for.


gregnr (approved):

gregnr:
Do we need an index for FTS, or are we too small for that to be worthwhile?

charislam:
> Do we need an index for FTS, or are we too small for that to be worthwhile?

We're reusing the tsvector columns from the Docs UI hybrid search, which are already indexed

PR #37128, opened 2025-07-14
Remove stop database admonition when performing db diff

ebauman:
Docs currently admonish the user to stop their local db instance before performing db diff. This is no longer necessary, see https://github.com/supabase/cli/pull/3829

YES

Docs update

n/a

## What is the new behavior?

n/a

## Additional context

### DIFF
--- a/apps/docs/content/guides/local-development/declarative-database-schemas.mdx
+++ b/apps/docs/content/guides/local-development/declarative-database-schemas.mdx
@@ -42,12 +42,6 @@ create table "employees" (
   </StepHikeCompact.Step>
 </StepHikeCompact>
 
-<Admonition type="note">
-
-Make sure your local database is stopped before diffing your schema.
-
-</Admonition>
-
 <StepHikeCompact>
 
   <StepHikeCompact.Step step={2}>


rlirli:
Came here to report the same documentation inconsistency.

The [`supabase db diff` CLI reference](https://supabase.com/docs/reference/cli/supabase-db-diff) correctly states:
> "Requires the local development stack to be running when diffing against the local database."

However, the [Declarative Database Schemas guide](https://supabase.com/docs/guides/local-development/declarative-database-schemas) still contains the contradictory instruction:
> "Make sure your local database is stopped before diffing your schema."

### Actual behavior:
When the database is stopped, `supabase db diff` fails with a connection error:
```
failed to connect to postgres: failed to connect to `host=127.0.0.1 user=postgres database=postgres`: dial error (dial tcp 127.0.0.1:54322: connect: connection refused)
```
(Although this might be cause I'm still running CLI on 2.30.4 instead of most recent 2.31.4.)


### Suggested fix:
Instead of removing the admonition, I'd suggest updating it to provide the correct instruction (if the error persists on 2.31.4):
```diff
- Make sure your local database is stopped before diffing your schema.
+ Make sure your local database is running before diffing your schema.
```

charislam (approved):
Thanks @ebauman!

PR #37134, opened 2025-07-15
add default main branch if none exists

SaxonF:
image

Adds default main branch as a temporary ux fix until main branch is created on project creation.


joshenlim (approved):
Verified changes on staging preview for a project that doesn't have branching enabled

PR #37137, opened 2025-07-15
Edit function refetch fix

SaxonF:
Fixes an issue where function body/files were re-fetched due to the parent detail layout. This was overwriting or removing files when editing a function.


joshenlim (approved):
Verified on staging preview that body request no longer gets called after refocusing

PR #37138, opened 2025-07-15
Remove review button

SaxonF:
We previously had a button to mark a branch as ready for review, but repositioning as merge requests so removing this button for now. A user could technically access the merge page and merge without a merge request, but requires a direct link.


joshenlim (approved):
Pushed up a small fix that i found with the tooltips - but otherwise LGTM! 🙂🙏

PR #37140, opened 2025-07-15
feat: bedrock prompt caching

gregnr:
Adds prompt caching to the AI assistant.

Cache works on input prefixes in the order: `tools` -> `system` -> ` messages`. So as long as tool schema and system prompt remain fixed between requests, all of their input tokens will be cached and reused, which results in faster inference and cheaper costs (1/10 the regular input cost). Any input token that is sent after the system prompt will be charged at the regular price.

If the input prefix changes between requests though, cache will be missed. System I've adjusted to be completely static and tools should only change based on opt-in level. So we should expect 4 separate caches active at any given time - one for each opt-in level, since these alter the tools provided.

See:
- https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html
- https://ai-sdk.dev/providers/ai-sdk-providers/amazon-bedrock#cache-points

Confirmed this is actually getting cached by adding the following during development:
```ts
result.providerMetadata.then((metadata) => {
  if (metadata?.bedrock) {
    console.log('Bedrock usage:', metadata.bedrock.usage)
  }
})
```

which produced:
```
Bedrock usage: { cacheReadInputTokens: 7854, cacheWriteInputTokens: 0 }
```

So we are saving on ~8k tokens every request. On 3.7 Sonnet and 3.5 Haiku, cache reads are 1/10 the cost of regular input tokens. We can add more cache points in the future if we want to - the thing to remember is that cache writes have a cost too (1.25x that of regular input token), so you want to make sure you will get at least one cache read from it within the next 5 min (cache has 5 min lifetime). System prompt is guaranteed to be reused across tenants, but beyond that the cache would be unique to each chat, so by adding additional cache points we would be betting whether or not the user decides to send more messages. Bedrock also only supports 4 cache checkpoints per request, and each cache point needs at least 1024 tokens.


SaxonF (approved):
Confirmed working locally and getting cache hit. Confirmed working on staging.

Image

PR #37143, opened 2025-07-15
Icons and close branch button changes

SaxonF:
Uses icons for edge function files diff and removes close branch button if pull updates vs merging updates.


alaister (approved):
LGTM!

- Tested on preview

PR #37146, opened 2025-07-15
fix: Fix link pointing to new API keys

lforst:

YES

Fixes a link that 404s by changing an `<a>` tag to a Next.js `<Link>`. I think the problem is that the `<a>` tag doesn't pick up the basePath that is configured in the Next.js config and doesn't prefix the link with `/dashboard`.


alaister (approved):
LGTM!

- Tested on preview

PR #37147, opened 2025-07-15
Some fixes for local studio

joshenlim:
## Changes involved

- Hide `BranchDropdown` behind `IS_PLATFORM`
- Add `IS_PLATFORM` checks to the following RQ queries
  - api-keys-query
  - branches-query
  - branch-query
  - branch-diff-query
  - legacy-api-keys-status-query.ts
- Update `cli-release-version` endpoint to return `current` as `undefined` if no `CURRENT_CLI_VERSION` env var is specified
  - This fixes the `LocalVersionPopover` as it's incorrectly showing up for self-hosted atm, showing version as undefined
  - image


alaister (approved):
LGTM!

- Tested on self-hosted preview

PR #37150, opened 2025-07-15
chore/unified-logs-fixes-04b

joshenlim:
Shifting changes over from [here](https://github.com/supabase/supabase/pull/36762) cause I messed up that PR 🥲 
Just copy pasting the PR description over since its the same, and also have addressed the feedback from that PR too

## Context

Couple more fixes from testing for Unified Logs interface, this round testing the following:

- Filter by Time range
- Filter Command Bar
- Self-hosted
  - JFYI UI is controlled via feature flag which means that it won't be available on self-hosted just yet until we clean up the feature flag
- RE Auth users: I'm unable to test this on my end as I'm not sure how to generate logs that'll include auth users

## Changes involved

- Fix missing colors for calendar component when a range is selected (Fixes FE-1694)
image
- Prevent selecting future dates from current date in unified logs date picker (Fixes FE-1695)
  - e.g Current date is 30th June, all dates in July are disabled
image
- Small cleanups for DataTableFilterCommand
  - Only show "Query" section if there are any queries
  - Only show "Suggestions" section if there are any suggestions
image
- Add missing POST method to dashboard analytics API to support Unified Logs for self-hosted (Fixes FE-1696)
  - I'm not able to test this locally by the way, running into this error for any pings to the logs pages for self-hosted run via `pnpm run dev:studio-local`
image

### DIFF
--- a/packages/ui/src/components/shadcn/ui/calendar.tsx
+++ b/packages/ui/src/components/shadcn/ui/calendar.tsx
@@ -10,6 +10,11 @@ import { buttonVariants } from './button'
 export type CalendarProps = React.ComponentProps<typeof DayPicker>
 
 function Calendar({ className, classNames, showOutsideDays = true, ...props }: CalendarProps) {
+  // [Joshen] Wondering if this is a bug with react-day-picker v8 whereby a single selected date
+  // has styles applied for both day_range_start and day_range_end
+  const fullDateRangeSelected =
+    props.mode === 'range' && !!props.selected?.from && !!props.selected?.to
+
   return (
     <DayPicker
       showOutsideDays={showOutsideDays}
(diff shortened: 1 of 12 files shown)


alaister (approved):
LGTM!

- Tested on preview

packages/ui/src/components/shadcn/ui/calendar.tsx
```diff
@@ -30,23 +35,27 @@ function Calendar({ className, classNames, showOutsideDays = true, ...props }: C
         head_row: 'flex',
         head_cell: 'text-foreground-muted rounded-md w-9 font-normal text-[0.8rem]',
         row: 'flex w-full mt-2',
-        cell: 'text-center text-sm p-0 relative [&:has([aria-selected])]:bg-accent first:[&:has([aria-selected])]:rounded-l-md last:[&:has([aria-selected])]:rounded-r-md focus-within:relative focus-within:z-20',
+        cell: cn(
```
MildTomato, line 39: Maybe no need for it here

packages/ui/src/components/shadcn/ui/calendar.tsx
```diff
@@ -10,6 +10,11 @@ import { buttonVariants } from './button'
 export type CalendarProps = React.ComponentProps<typeof DayPicker>
 
 function Calendar({ className, classNames, showOutsideDays = true, ...props }: CalendarProps) {
+  // [Joshen] Wondering if this is a bug with react-day-picker v8 whereby a single selected date
+  // has styles applied for both day_range_start and day_range_end
+  const fullDateRangeSelected =
```
MildTomato, line 16: Need to check if we actually need this, there might be a data attr we could target tailwind with

MildTomato (approved):

packages/ui/src/components/shadcn/ui/calendar.tsx
joshenlim: i'll add a comment about this before merging so that we don't forget to check

PR #37160, opened 2025-07-15
fix filltimeseries for reports

jordienr:
uses fillTimestampSorted in new reports so the charts dont look bad when there isnt a lot of data.

## BEFORE
CleanShot 2025-07-15 at 17 01 01@2x

## AFTER
CleanShot 2025-07-15 at 17 01 17@2x


joshenlim (approved):
Verified changes on staging preview - although noticed some odd behaviour with the charts that's not related to the PR

PR #37174, opened 2025-07-15
Fix planet return example

hakro:
The example provided does not work. 
It ends up with a `ERROR:  query has no destination for result data`, because we need to return the actual set and mark it as a query.

YES

docs example fix

N/A

## What is the new behavior?
N/A


charislam:
Thanks for the PR @hakro! The `return query` is not needed in this case because the example is explicitly written as a plain SQL function, not a plpgsql one. I'm going to close but feel free to reopen if you think there are outstanding issues.

PR #37183, opened 2025-07-15
Upgrades pro users and above

SaxonF:
If an organization is on a paid plan, we upgrade the model. Note that this will have no effect until we set IS_THROTTLED to "false".

This requires https://github.com/supabase/supabase/pull/37140 to be merged first


joshenlim (approved):
Minimally tested that the assistant is still working as expected on staging preview 🙂

PR #37191, opened 2025-07-16
chore: deep link feature preview modal

alaister:
- Adding `?featurePreviewModal` to the URL will open the modal and select the first item
- Specify a feature to deep link to via `?featurePreviewModal=supabase-ui-branching-2-0` etc


apps/studio/components/interfaces/App/FeaturePreview/FeaturePreviewContext.tsx
```diff
@@ -89,3 +98,68 @@ export const useIsBranching2Enabled = () => {
...
+    ? router.query.featurePreviewModal[0]
+    : router.query.featurePreviewModal
+
+  const selectedFeatureKey = featurePreviewModal?.trim()
+  const showFeaturePreviewModal = selectedFeatureKey !== undefined
+
+  const openFeaturePreviewModal = useCallback(() => {
+    router.replace({
+      pathname: router.pathname,
+      query: { ...router.query, featurePreviewModal: '' },
```
joshenlim: could default `featurePreviewModal` to the first one in the list?
so that users can share the link for the first feature preview when they initially open it

joshenlim (approved):

PR #37192, opened 2025-07-16
Branch refinements

SaxonF:
- PITR copy
- Project name in header and home page should not include branch
- Reset branch requires adding text
- Persistent branch copy
- Branch price copy
- Migration status link (in future can open logs dialog)
- Branch badge preview vs persistent in header
- Create branch permission check


alaister (approved):
LGTM!

- Tested locally

PR #37193, opened 2025-07-16
fix: disable usercentrics for local dev

alaister:
In local dev there's a giant error caused by usercentrics being unable to load on local host. This PR treats consent as accepted while in local development.


joshenlim (approved):
Verified locally - unable to test on staging preview as i'm not able to opt in (staging works tho)

PR #37200, opened 2025-07-16
Chore/async filters for unified logs

joshenlim:
## Context

This is for Unified Logs, specifically the filter side bar in which some of the facets can have dynamically rendered values such as host, pathname, and auth user.

## Changes involved

- Facet options will be limited to up to 20 results (this is arbitary - happy to change), mainly for performance (configured from the SQL query)
- Host, pathname, and auth user now have search input, in which facet options will be dynamically retrieved after a debounce input of 700ms (arbitary value too - happy to change)
- Refactors to make the code a bit more readable with the dynamic data loading of facets
  - Added `isDynamicOptions` and `hasAsyncSearch` properties in `UnifiedLogs.fields.ts` -> more declarative, less inferring

https://github.com/user-attachments/assets/08ce24e2-b0a6-4a13-8f79-b14c8ab57acc


alaister (approved):
LGTM!

- Tested on preview

Works beautifully!

PR #37205, opened 2025-07-16
feat: Create new schema when creating a FDW with the schema option

ivasilov:
When creating a new foreign data wrapper, if you select the option to target a specific schema, the schema will be created for you, you can't use existing (pre-populated) schema.


apps/studio/components/interfaces/Integrations/Wrappers/CreateIcebergWrapperSheet.tsx
```diff
@@ -292,18 +318,21 @@ export const CreateIcebergWrapperSheet = ({
...
-                        <Label_Shadcn_ className="text-foreground-light">
-                          Target Schema
-                        </Label_Shadcn_>
-                        <SchemaSelector
-                          portal={false}
-                          size="small"
-                          selectedSchemaName={values.target_schema}
-                          onSelectSchema={(schema) => setFieldValue('target_schema', schema)}
-                          onSelectCreateSchema={() => setCreateSchemaSheetOpen(true)}
+                        <InputField
```
joshenlim, line 326: for iceberg specifically - does this mean each wrapper needs to be in its own schema?

joshenlim:
Pushed up some minor fixes

- Z index issue here
image

- Alignment issue here
image

joshenlim:
Unrelated but i'm getting 2 error toasts when creating a analytics bucket
[edit] pushed up a fix for this
image

joshenlim:
Going through this while focusing on the UX side of things this time,  some thoughts/ feedback - wondering what you might think. Trying to reduce any ambiguity for a user who might not understand what foreign tables / foreign schemas are

- Thinking we change the caption here to explain the outcome of choosing this action
  - Tables: "Create foreign tables to query data from [wrapper name e.g Stripe]"
  - Schema: "Create all foreign tables from [wrapper name e.g Stripe] in a specified schema"
  image

- Thinking we can hide the "Source schema" input field here since its read only and showing this might not have any benefit? (so we can reduce unnecessary information)
  - Change "Target schema"  label to "Specify a new schema to create all wrapper tables in"
  - Description could explain (if possible) why existing schemas cannot be used? e.g "For security purposes, the wrapper tables from the foreign schema cannot be created within an existing schema"
  image
  - These feedback would also apply to the import foreign schema dialog RE label and description
  image


- Can change this description to be more similar to the Foreign Tables one
  - You can query your data from the foreign tables in the specified schema after the wrapper is created
  image

apps/studio/data/storage/iceberg-namespace-create-mutation.ts
joshenlim, line 1: i might be overlooking something but i realised that this RQ mutation isn't being used?

apps/studio/data/storage/iceberg-namespace-exists-query.ts
joshenlim, line 1: this one as well - not being used atm?

apps/studio/data/storage/iceberg-wrapper-create-mutation.ts
joshenlim, line 1: unrelated but noticed that `disableCreation` on L40 isn't being used atm - should it be used somewhere here?

joshenlim:
UX feedback - thinking we could add links to vault secrets for these
Don't have to tackle in this PR btw, just leaving feedback as I come across any
[edit] Added in this commit [here](https://github.com/supabase/supabase/pull/37205/commits/1965df5bf3a1c42c46aa1257571a11c49349cae0)

image

apps/studio/data/storage/iceberg-namespace-create-mutation.ts
ivasilov: Yes, I used them in prototyping. They'll probably be used in future iterations of the UI so I left them.

joshenlim: gotcha on this all good then! 🙏

apps/studio/data/storage/iceberg-wrapper-create-mutation.ts
ivasilov: Removed.

ivasilov:
> Going through this while focusing on the UX side of things this time, some thoughts/ feedback - wondering what you might think. Trying to reduce any ambiguity for a user who might not understand what foreign tables / foreign schemas are
> ....

Agree with all copy changes, I've applied them.

One small remark on hiding the source schema input: I've hidden it (if it's read-only) in all wrappers except in wrappers where it's configurable (Iceberg for example).

joshenlim (approved):
All good now! 🙂 🙏

PR #37210, opened 2025-07-16
fix: update endpoint for cname check

kangmingtay:
* some ISPs block requests to one.one.one.one, which results in the custom domain check to fail and the user not being able to activate it

## Changes involved
- Use `cloudflare-dns.com` instead of `one.one.one.one`
- Shift CNAME checking to dashboard API route
- Remove `one.one.one.one` from CSP rules

## To test
- [ ] Can set up custom domain on project


delgado3d (approved):

apps/studio/data/custom-domains/check-cname-mutation.ts
```diff
@@ -22,19 +23,24 @@ export type CheckCNAMERecordResponse = {
...
-  const res = await fetchHandler(`https://one.one.one.one/dns-query?name=${domain}&type=CNAME`, {
-    method: 'GET',
-    headers: { accept: 'application/dns-json' },
-  })
-  const verification = (await res.json()) as CheckCNAMERecordResponse
+  try {
+    const headers = await constructHeaders({ 'Content-Type': 'application/json' })
+    const res: CheckCNAMERecordResponse = await fetchHandler(`${BASE_PATH}/api/check-cname`, {
+      headers,
+      method: 'POST',
```
kangmingtay, line 30: nit: since the underlying endpoint is just a `GET` request, maybe it's worth making this a `GET` and sending the domain as a query param?

ivasilov (approved):
LGTM.

staaldraad (approved):

PR #37231, opened 2025-07-17
fix: plan redirect from pricing page

kevcodez:
- Fixes the broken preselection of the Plan when going through pricing page while not signed in
- Adds proper plan selection when navigating through pricing page while being logged in and selecting "New Organization" in the redirect modal


ecktoteckto (approved):
Tested on Preview

fsansalvadore (approved):

PR #37242, opened 2025-07-17
fix timestamp missing

jordienr:
- adds back missing timestamp


alaister (approved):
LGTM!

- Tested on preview

PR #37243, opened 2025-07-17
Feat/logs date refactor

MildTomato:
- adds reports date picker to new logs page


joshenlim:
Small issue found - it says there's 2 filters applied when i selected "last 30 minutes" -> should be 1?
Hmm actually no worries - noticed that this is same as per current behaviour

joshenlim (approved):

PR #37246, opened 2025-07-17
fix: full refresh on logs page

alaister:
Refreshing on the `/logs` page would cause the redirect to be triggered back to the old logging interface (regardless that the feature preview was on), and because the redirect triggered too early there would be no project ref.

To test:
- Navigate to the new logs and refresh the page


joshenlim (approved):
Verified changes on staging preview that refreshing doesnt prematurely route back to the old logs UI

PR #37247, opened 2025-07-17
fix: enrichment logs queries

MildTomato:
- enrichment logs (serviceflow.sql) was targeting wrong node in logs. caused bigquery error on prod and staging


joshenlim (approved):
Minimally checked the Unified Logs on staging preview as I wasn't able to reproduce the bug that this PR fixes

PR #37248, opened 2025-07-17
Reports clean ups

jordienr:
- use similar styles for Shared API Reports (uses old components) and new ReportChart
- Fix loading state not triggering no refreshes in auth report
- Remove unnecessary code in useChartData


alaister (approved):
LGTM!

- Tested on preview

MildTomato (approved):
approved /packages/ui/* changes

PR #37249, opened 2025-07-17
Chore/add instructions for psql export table editor

joshenlim:
## Context
2 problems that this is trying to address
- We've had issues of users trying to figure out the best way to export their table data. While the dashboard does currently provide a way to export as CSV / SQL, its not really the best way as it involves paginating via limit and offset with `pg-meta` and involves several calls to `pg-meta` (we fetch 500 rows per page, so it could involve hundreds of calls if the table has > 100k rows). Not to mention this involves heavy load on the browser as we're storing all the rows in memory before downloading it
- There's been some inconsistencies with exporting all rows via this method as well, most commonly reported issue is with duplicated rows in the results. Likely related to the limit offset logic.

Reckon the best solution here is to export rows via the CLI and using either psql or pg_dump export data from the database directly, so this PR tries to make it easier for users to do so by having the necessary commands easily copyable

## Changes involved
- Add "Export via CLI" dropdown option under Export in Table Editor if _all_ rows are selected
image

- Add instructions for exporting table data to CSV via psql (tested command locally)
image

- Add instructions for exporting table data to SQL via pg_dump (tested command locally)
image

## To test
- [ ] Verify the UX, if it makes sense from a user's POV
- [ ] Test that the command to export to CSV works
- [ ] Test that the command to export to SQL works

### DIFF
--- a/apps/studio/components/grid/components/header/Header.tsx
+++ b/apps/studio/components/grid/components/header/Header.tsx
@@ -484,6 +484,8 @@ const RowHeader = () => {
           <DropdownMenuContent className={snap.allRowsSelected ? 'w-52' : 'w-40'}>
             <DropdownMenuItem onClick={onRowsExportCSV}>Export as CSV</DropdownMenuItem>
             <DropdownMenuItem onClick={onRowsExportSQL}>Export as SQL</DropdownMenuItem>
+            {/* [Joshen] Should make this available for all cases, but that'll involve updating
+            the Dialog's SQL output to be dynamic based on any filters applied */}
             {snap.allRowsSelected && (
               <DropdownMenuItem className="group" onClick={() => setShowExportModal(true)}>
                 <div>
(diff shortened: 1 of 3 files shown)


apps/studio/components/grid/components/header/Header.tsx
```diff
@@ -441,61 +443,71 @@ const RowHeader = () => {
...
+              loading={isExporting}
+              disabled={isExporting}
+            >
+              Export
+            </Button>
+          </DropdownMenuTrigger>
+          <DropdownMenuContent className={snap.allRowsSelected ? 'w-52' : 'w-40'}>
+            <DropdownMenuItem onClick={onRowsExportCSV}>Export as CSV</DropdownMenuItem>
+            <DropdownMenuItem onClick={onRowsExportSQL}>Export as SQL</DropdownMenuItem>
+            {snap.allRowsSelected && (
```
ivasilov, line 489: I think we should always show this option and push the user towards using it since it's the most reliable.

ivasilov (approved):
Tested, both commands work. Left a small nit comment.

apps/studio/components/grid/components/header/Header.tsx
joshenlim: agree on this one! i'll leave a comment for now in the code to support this - cause we'll need to update the SQL in the dialog as well to be dynamic based on any filters applied

can look into it next 🙂

PR #37250, opened 2025-07-17
fix: Prevent arrow key propagation in filter command

MildTomato:


joshenlim (approved):

PR #37252, opened 2025-07-17
Chore/logs reloading state

MildTomato:
- Logs table now has a more subtle reloading state wen refetching new logs


joshenlim (approved):
Verified locally!

PR #37258, opened 2025-07-17
Refactor `EditSecretModal` to use shadcn components

Saeris:

YES

This primarily refactors `EditSecretModal` for the vault secrets page such that it uses shadcn components instead of older deprecated ones. It follows the same implementation pattern as #36972

Additionally it adds some basic UI tests for this component as well as some fixes and improvements to `studio`'s testing setup and utilities.

Noteworthy is that this introduces an extra layer of type safety to `addAPIMock()` inspired by the pattern used with `createClient()` from `openapi-fetch`. It uses the same generated type definitions to provide type information to the response type for each endpoint you may want to mock. Not only will this provide autocomplete and type info for the response objects, it can also potentially highlight changes to the API which would affect existing tests.

This also adds `jsdom-testing-mocks` as a new development dependency to polyfill web animation APIs necessary for the shadcn Dropdown components to be rendered in tests.

Because this is branched off of work on #36972, I'll be leaving this as a draft until that PR is reviewed and merged.

## Additional context


apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretRow.tsx
```diff
@@ -46,6 +46,15 @@ const SecretRow = ({ secret, onSelectEdit, onSelectRemove }: SecretRowProps) =>
...
   )
+  const renderModal = () => {
+    const onClose = () => setModal(null)
+    switch (modal) {
+      case `edit`:
+        return <EditSecretModal secret={secret} onClose={onClose} />
+      default:
+        return null
+    }
+  }
```
Saeris, line 50: I end up moving the delete modal into `SecretRow` in a subsequent PR, so that's the reason for this block here

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretRow.tsx
```diff
@@ -128,15 +130,16 @@ const SecretRow = ({ secret, onSelectEdit, onSelectRemove }: SecretRowProps) =>
...
-        </DropdownMenu>
+              </DropdownMenuContent>
+            </DropdownMenu>
+          </TooltipTrigger>
+          {!canManageSecrets && (
+            <TooltipContent side="bottom">
+              You need additional permissions to manage secrets
+            </TooltipContent>
+          )}
+        </Tooltip>
```
Saeris: rather than having a repetitive message for each menu item, we can just disable the whole menu since all the items require permission to manage the secret

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretsManagement.tsx
```diff
@@ -9,22 +9,29 @@ import { ButtonTooltip } from 'components/ui/ButtonTooltip'
...
 export const SecretsManagement = () => {
   const { search } = useParams()
   const { project } = useProjectContext()
 
   const [searchValue, setSearchValue] = useState<string>('')
-  const [selectedSort, setSelectedSort] = useState<'updated_at' | 'name'>('updated_at')
   const [showAddSecretModal, setShowAddSecretModal] = useState(false)
-  const [selectedSecretToEdit, setSelectedSecretToEdit] = useState<VaultSecret>()
   const [selectedSecretToRemove, setSelectedSecretToRemove] = useState<VaultSecret>()
+  const [selectedSort, setSelectedSort] = useState('updated_at')
```
Saeris, line 34: removing the string literal union here was necessary because of a type error on `Select_Shadcn_`

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretsManagement.tsx
```diff
@@ -81,26 +88,21 @@ export const SecretsManagement = () => {
...
-                >
+                </SelectItem_Shadcn_>
+                <SelectItem_Shadcn_ value="name" className="text-xs">
                   Name
-                </Listbox.Option>
-              </Listbox>
-            </div>
+                </SelectItem_Shadcn_>
+              </SelectContent_Shadcn_>
+            </Select_Shadcn_>
```
Saeris, line 105: I'm not sure it matters but I do want to point out that the select dropdown icon is different between these two implementations

Saeris: additionally, I noticed that when TanStack dev tools are open, opening this select causes a layout shift, so if you ever see that and wonder why, it's not something you need to worry about debugging

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretsManagement.tsx
```diff
@@ -167,10 +165,6 @@ export const SecretsManagement = () => {
         </div>
       </div>
 
-      <EditSecretModal
-        selectedSecret={selectedSecretToEdit}
-        onClose={() => setSelectedSecretToEdit(undefined)}
-      />
       <DeleteSecretModal
         selectedSecret={selectedSecretToRemove}
         onClose={() => setSelectedSecretToRemove(undefined)}
```
Saeris, line 176: both of these get moved into `SecretRow` because it doesn't really make sense to have this awkward sibling relationship where the dialog trigger is nested, causing state management in the parent. I think it's the kind of wonky coupling of components that we'd want to avoid

apps/studio/components/interfaces/Integrations/Vault/Secrets/EditSecretModal.tsx
```diff
@@ -1,158 +1,214 @@
...
+    defaultValues: {
+      name: secret?.name ?? '',
+      description: secret?.description ?? '',
+      secret: secret?.decryptedSecret ?? data ?? '',
+    },
+    values: {
+      name: secret?.name ?? '',
+      description: secret?.description ?? '',
+      secret: secret?.decryptedSecret ?? data ?? '',
+    },
```
Saeris: this might seem like unnecessary duplication but I found that this was what was needed to fill in the values that are loaded in asynchronously from `useVaultSecretDecryptedValueQuery`. it's better than the odd hook inside of render function that was here previously

apps/studio/components/interfaces/Integrations/Vault/Secrets/EditSecretModal.tsx
```diff
@@ -1,158 +1,214 @@
...
 
   return (
-    <Modal
-      hideFooter
-      size="medium"
-      visible={selectedSecret !== undefined}
-      onCancel={onClose}
-      header={<h5 className="text-sm text-foreground">Edit secret</h5>}
+    <Dialog
+      open
```
Saeris: because open state is effectively being controlled by whether the component is mounted or not in `SecretRow`, we just set this dialog to be default open. that's one way to get around the decoupled dialog trigger within a dropdown menu pattern we've got going on here

apps/studio/components/interfaces/Integrations/Vault/Secrets/__tests__/EditSecretModal.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+    expect(descriptionInput).toHaveValue(secret.description)
+    expect(valueInput).toHaveAttribute(`type`, `password`)
+    await userEvent.click(togglePasswordButton)
+    expect(valueInput).toHaveAttribute(`type`, `text`)
+
+    await userEvent.type(nameInput, `updated-name`)
+    await userEvent.clear(descriptionInput)
+    await userEvent.type(valueInput, `qux`)
+
+    fireEvent.click(submitButton)
```
Saeris, line 99: something to note here, which I went into more detail on Slack, is that I had a heck of a time getting form submission to play nice and for this particular component `fireEvent` was the only method I could get to work *in test*

the form works just as you would expect in the browser, but for whatever reason in a testing environment it doesn't like either `userEvent.click` or an `{enter}` on a text input. I wouldn't waste more time trying to debug why this is the case, but make a mental note of this for writing future tests as it may crop up as a problem you'll encounter

and importantly since `fireEvent` isn't something you can `await` on, you will need to have something like the following line to wait for a change in the UI state when the event in question triggers something asynchronous like a fetch call

apps/studio/components/interfaces/Integrations/Vault/Secrets/__tests__/EditSecretModal.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <DropdownMenuContent side="bottom" align="end" className="w-32">
+          <DropdownMenuItem className="space-x-2" onClick={() => setModal(`edit`)}>
+            <Edit3 size="14" />
+            <p>Edit</p>
+          </DropdownMenuItem>
+        </DropdownMenuContent>
+      </DropdownMenu>
+
+      {renderModal()}
+    </ProjectContextProvider>
```
Saeris: this much boilerplate may not be necessary, but I wanted to ensure that we're getting the right behavior integrating with `DropdownMenu`, where the two popovers can have some complicated mounting/unmounting behaviors if not done correctly

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretRow.tsx
saltcod: this is valid, but not common I think for this codebase. Could we use inline conditionals or extract to a new component completely?

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretsManagement.tsx
saltcod: the selected check icon does look big. I think this needs `portal=false` to fix the layout issue?

screenshot-2025-07-28-at-17 13 11

saltcod (commented):
Left a couple of nits, but tested on staging and adding/changing/deleting secrets works as expected.

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretsManagement.tsx
Saeris: hmm, that appears to be off. this is what I am seeing on my end:

image

that layout issue would appear to be consistent with the old vault sort select

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretRow.tsx
Saeris: yeah I mean I suppose we could inline this in the JSX with conditionals instead. I'm using the switch statement pattern here based on a solution I came across researching how to combine a radix dialog with a dropdown menu. personally I like that it keeps this whole conditional UI in one place, close handler and all

Saeris:
> Left a couple of nits, but tested on staging and adding/changing/deleting secrets works as expected.

did you check the related PRs for adding/deleting? because this one only handles edits

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretRow.tsx
saltcod: yeah it's tidy, but a bit of an inconsistent pattern for here is all. The more verbose inline conditionals would be more consistent

Saeris: alright then, how about I implement that change in #37386 then? just to avoid needing to repeat the change across multiple PRs, since we're needing to review and merge them in order anyways

joshenlim: chiming in on the consistency part here 🙏  just to share a bit of context:

we're prioritising consistency to mitigate the problem of "n ways to do something" - which can be confusing for developers contributing to the codebase as they'd be confused on which style to follow (we've had several feedbacks about this internally 😅) There's currently a number of other stuffs that are inconsistent in the dashboard too, so trying to avoid widening the gap as well

I'd also prefer we update this within this PR if that's alright? We may not be able to get to your other PRs immediately, so wanna avoid a limbo of this inconsistency 🙏

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretRow.tsx
```diff
@@ -95,31 +104,24 @@ const SecretRow = ({ secret, onSelectEdit, onSelectRemove }: SecretRowProps) =>
...
-          <DropdownMenuTrigger asChild>
-            <Button type="text" className="px-1" icon={<MoreVertical />} />
-          </DropdownMenuTrigger>
-          <DropdownMenuContent side="bottom" align="end" className="w-32">
-            <Tooltip>
-              <TooltipTrigger asChild>
-                <DropdownMenuItem
-                  className="space-x-2"
+
+        <Tooltip>
```
joshenlim, line 108: same thing RE consistency - we aren't currently following a pattern of disabling the dropdown menu button here

i'd stick to keeping the dropdown menu clickable, but disabling the menu items with DropdownMenuItemTooltip instead ([example](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/PolicyRow.tsx#L152))

this is probably worth relooking as a pattern across the dashboard, but i'd say we do that separately

apps/studio/components/interfaces/Integrations/Vault/Secrets/EditSecretModal.tsx
joshenlim: this will be related to this thread [here](https://github.com/supabase/supabase/pull/37258/files#r2225321499) RE consistency - i'd prefer to have the open param controlled by a provided param for now

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretRow.tsx
joshenlim: same thing RE consistency - we aren't currently following a pattern of disabling the dropdown menu button here

i'd stick to keeping the dropdown menu clickable, but disabling the menu items with DropdownMenuItemTooltip instead ([example](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/PolicyRow.tsx#L152)).

It's another pattern that we've been following - avoid hiding UI, make it visible but indicate why certain actions aren't allowed. So by showing disabled dropdown menu items, users will still be able to see what actions are available

apps/studio/components/interfaces/Integrations/Vault/Secrets/EditSecretModal.tsx
joshenlim: one thing i noticed as well if we do this way, is that we won't have the animation of the modal when its closing (since the component is being unmounted immediately)

apps/studio/components/interfaces/Integrations/Vault/Secrets/EditSecretModal.tsx
joshenlim: TIL! Wonder if we can declare a const here to reduce duplication

```
const values =  {
  name: secret?.name ?? '',
  description: secret?.description ?? '',
  secret: secret?.decryptedSecret ?? data ?? '',
}
```

then 
```
{ defaultValues: values, values } 
```

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretsManagement.tsx
joshenlim: yup can verify looks okay on my end too:
image

Saeris:
@joshenlim @saltcod I've implemented all of the requested changes I think

simplified the test to remove the extra boilerplate, as after writing a handful more like it this week it doesn't seem to be necessary now that I'm more familiar with the dialog triggering patterns we're using

apps/studio/components/interfaces/Integrations/Vault/Secrets/EditSecretModal.tsx
Saeris: probably that would work! I'm pretty sure we can make the `secret` prop non-optional too, since `SecretRow` wouldn't be rendered without one being defined

apps/studio/components/interfaces/Integrations/Vault/Secrets/EditSecretModal.tsx
```diff
@@ -1,158 +1,215 @@
...
+    },
+  })
 
   useEffect(() => {
-    if (selectedSecret !== undefined) {
+    if (secret !== undefined) {
       setShowSecretValue(false)
     }
-  }, [selectedSecret])
+  }, [secret])
```
Saeris: if I implement the change suggested above, then I think this `useEffect` would become unnecessary

apps/studio/components/interfaces/Integrations/Vault/Secrets/SecretRow.tsx
Saeris: that's fair. my general line of thinking here is to reduce unnecessary complexity and fewer props and state are some that I'm rather staunch about. sometimes that's necessarily going to result in some inconsistency but I agree that we should avoid that where possible

apps/studio/components/interfaces/Integrations/Vault/Secrets/EditSecretModal.tsx
Saeris: looks like that is the case, so this and other unnecessary conditional logic was cleaned up

Saeris:
merged up and implemented the requested changes into the downstream PRs

Saeris:
synced with master and did some minor cleanup. pretty sure all feedback has been addressed and changes implemented. let's merge?

apps/studio/components/interfaces/Integrations/Vault/Secrets/__tests__/EditSecretModal.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+  )
+}
+
+mockAnimationsApi()
+
+describe(`EditSecretModal`, () => {
+  beforeEach(() => {
+    // useSelectedProjectQuery -> useParams
+    routerMock.setCurrentUrl(`/project/default/integrations/vault/secrets`)
```
Saeris, line 45: this one gets me *all the time*

useParams pulls the project ref from the current route, so without mocking this, it will return undefined and cause the whole component to be stuck in a perpetual loading state

apps/studio/tests/lib/route-mock.ts
```diff
@@ -1,11 +1,26 @@
...
+// remove the file extension or index segment
+const removeExt = (str: string) => str.replace(/.tsx|\/index.tsx/, ``)
+
+// Glob the `pages/` directory for all dynamic route segments
+const paths = [
+  ...(await Array.fromAsync(glob(`${pagesRoot}/**/*\].tsx`))),
+  ...(await Array.fromAsync(glob(`${pagesRoot}/**/**\]/**/index.tsx`))),
+].map(pipe(normalize, toPosix, removePrefix, removeExt))
+
+routerMock.useParser(createDynamicRouteParser(paths))
```
Saeris, line 26: I thought this had already been merged in with another PR, but it must still be under review

without providing a list of our dynamic routes to parse, we can't extract a value like `projectRef` when `useParams` is called

since we don't throw any errors on this, many of our hooks will just accept an `undefined` value for proectRef and get stuck in a perpetual loading state, causing a test to fail due to a UI not updating or a by timing out

saltcod (approved):
Looks good on staging!

PR #37262, opened 2025-07-17
Refactor unified logs count query logic

MildTomato:
This pull request improves UnifiedLogs.queries.ts by:
	•	Removing unused fields and joins (host, api_role, auth_user, and related LEFT JOINs) to reduce query complexity and improve performance.
	•	Rewritten the union all query to only query whats needed for count query


joshenlim (approved):
Code changes LGTM - minimally did some basic smoke testing on staging preview (although my logs on my staging project don't really vary)

Double checked the logs count query that it's returning the right data, and UI is displaying respectively

MildTomato:
Also smoke tested. ran a log cannon on a project.

PR #37263, opened 2025-07-17
fix: remove async filter search

MildTomato:
- Disabling async search on pathname
- Removing the “Auth User” filter entirely


joshenlim (approved):
Verified on staging preview that no longer able to search on pathname, and no longer able to filter on auth users

MildTomato:
auto merge on

PR #37265, opened 2025-07-17
fix: restrict log inspection panel tab focus refetch

MildTomato:
- stops inspection panel running an enrichment query on each tab refocus
  - we don't need this as log data doesn't change over time

### DIFF
--- a/apps/studio/data/logs/unified-log-inspection-query.ts
+++ b/apps/studio/data/logs/unified-log-inspection-query.ts
@@ -182,6 +182,7 @@ export const useUnifiedLogInspectionQuery = <TData = UnifiedLogInspectionData>(
     ({ signal }) => getUnifiedLogInspection({ projectRef, logId, type, search }, signal),
     {
       enabled: enabled && typeof projectRef !== 'undefined',
+      refetchOnWindowFocus: false,
       ...options,
     }
   )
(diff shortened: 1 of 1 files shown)


apps/studio/data/logs/unified-log-inspection-query.ts
```diff
@@ -182,6 +182,7 @@ export const useUnifiedLogInspectionQuery = <TData = UnifiedLogInspectionData>(
     ({ signal }) => getUnifiedLogInspection({ projectRef, logId, type, search }, signal),
     {
       enabled: enabled && typeof projectRef !== 'undefined',
+      refetchOnWindowFocus: false,
```
joshenlim: reckon we can just use `UNIFIED_LOGS_QUERY_OPTIONS` here which does all this
image

we use it in the facet-count-query and the infinite-query too

joshenlim (approved):
Code changes LGTM

apps/studio/data/logs/unified-log-inspection-query.ts
MildTomato: nice

MildTomato: updating

PR #37281, opened 2025-07-18
remove dynamic chat name from prompt for caching

SaxonF:
Removes chat name from system prompt so that the entire system prompt can be cached.


gregnr (approved):

gregnr:
Smoke test on staging confirming LLM still knows chat name and other functionality still works.

PR #37290, opened 2025-07-18
feat: Make the protected schemas dynamic, namespace schemas are now protected

ivasilov:
This PR refactors the `PROTECTED_SCHEMAS` array to be dynamic. The dynamic schemas are coming from the Iceberg FDWs.

This is to make the Iceberg-connected schemas read-only.


apps/studio/hooks/useProtectedSchemas.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Returns a list of schemas that are protected by Supabase (internal schemas or schemas used by Iceberg FDWs).
+ */
+export const useProtectedSchemas = ({
+  excludeSchemas = [],
+}: { excludeSchemas?: string[] } = {}) => {
+  // Stabilize the excludeSchemas array to prevent unnecessary re-computations
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const stableexcludeSchemas = useMemo(() => excludeSchemas, [JSON.stringify(excludeSchemas)])
```
joshenlim: tinnyyyy nit

```suggestion
  const stableExcludeSchemas = useMemo(() => excludeSchemas, [JSON.stringify(excludeSchemas)])
```

apps/studio/hooks/useProtectedSchemas.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const result = useIcebergFdwSchemasQuery()
+
+  const schemas = useMemo<{ name: string; type: 'fdw' | 'internal' }[]>(() => {
+    const internalSchemas = INTERNAL_SCHEMAS.map((s) => ({ name: s, type: 'internal' as const }))
+    const fdwSchemas = result.data?.map((s) => ({ name: s, type: 'fdw' as const }))
```
joshenlim: tiny nit just to be clearer that this is only FWDs for iceberg (although lemme know if im wrong though)

```suggestion
    const icebergFdwSchemas = result.data?.map((s) => ({ name: s, type: 'fdw' as const }))
```

apps/studio/components/interfaces/Database/ProtectedSchemaWarning.tsx
```diff
@@ -45,30 +45,58 @@ export const ProtectedSchemaModal = ({
...
 }
 
 const ProtectedSchemaWarning = ({ schema, entity }: { schema: string; entity: string }) => {
   const [showModal, setShowModal] = useState(false)
+  const { isSchemaLocked, reason } = useIsProtectedSchema({ schema })
+
+  if (!isSchemaLocked) return null
 
   return (
     <>
```
joshenlim: probably can remove unnecessary fragment

apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
```diff
@@ -164,12 +155,30 @@ const TableEditorMenu = () => {
                   Viewing protected schema
                 </AlertTitle_Shadcn_>
                 <AlertDescription_Shadcn_ className="text-xs">
-                  <p className="mb-2">
```
joshenlim, line 169: this Alert_Shadcn can be replaced by ProtectedSchemaWarning?

apps/studio/components/interfaces/Database/ProtectedSchemaWarning.tsx
```diff
@@ -1,37 +1,37 @@
...
+      <DialogHeader>
+        <DialogTitle>Schemas managed by Supabase</DialogTitle>
+      </DialogHeader>
+      <DialogSectionSeparator />
+      <DialogSection className="space-y-2">
         <p className="text-sm">
           The following schemas are managed by Supabase and are currently protected from write
           access through the dashboard.
         </p>
         <div className="flex flex-wrap gap-1">
```
joshenlim, line 31: probably needs `prose` class so that the `code` elements render nicely, currently looks like this
image

although separately, i'm wondering what's the intended styles of the `code` elements - i feel like this shouldn't depend on `prose` being passed from the parent?

```suggestion
        <div className="flex flex-wrap gap-1 prose">
```

joshenlim:
Not related but reckon we show a warning if a schema exposed via API is selected?
image

joshenlim:
Some UI issues observed:

- Tab for foreign table doesn't have schema name if you have tables across schemas opened as tabs
image

- Was expecting to see "Viewing protected schema" warning here in table editor, schema has iceberg fdw
image

- (Unrelated) I'm not sure when this broke, but opened tabs are no longer getting persisted across browser refreshes

apps/studio/hooks/useProtectedSchemas.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const schemas = useMemo(() => {
+    const icebergFDWs = result.data
+      ?.filter((wrapper) =>
+        wrapperMetaComparator(
+          { handlerName: WRAPPER_HANDLERS.ICEBERG, server: { options: [] } },
+          wrapper
+        )
+      )
+      .filter((fdw) => fdw.name.endsWith('_fdw'))
```
joshenlim: Was checking locally and realised `icebergFDWs` returns an empty array here, despite me having one iceberg FDW
likely because of the `fdw.name` comparison here - `name` is dynamically input by the user so perhaps we should be checking on another property? or is this additional filter required actually

ivasilov: You're right, it's supposed to filter all FDWs that have a connected schema. Due to how `import foreign schema` works, the FDW doesn't save the schema in its state. I think I've figured out a way to save it, I'll try and add it to https://github.com/supabase/supabase/pull/37205.

apps/studio/components/interfaces/Database/ProtectedSchemaWarning.tsx
ivasilov: Fixed.

apps/studio/hooks/useProtectedSchemas.ts
ivasilov: Fixed.

apps/studio/data/fdw/fdw-delete-mutation.ts
```diff
@@ -73,7 +73,7 @@ export const getDeleteFDWSql = ({
   const deleteEncryptedSecretsSql = deleteEncryptedSecretsSqlArray.join('\n')
 
   const sql = /* SQL */ `
-    drop foreign data wrapper if exists ${wrapper.name} cascade;
+    drop foreign data wrapper if exists "${wrapper.name}" cascade;
```
ivasilov, line 76: This fixes an issue when trying to delete a FDW with a number in its name.

apps/studio/hooks/misc/useSchemaQueryState.ts
```diff
@@ -12,7 +12,12 @@ const useIsomorphicUseQueryState = (defaultSchema: string) => {
     return [defaultSchema, () => {}] as const
   } else {
     // eslint-disable-next-line react-hooks/rules-of-hooks
-    return useQueryState('schema', parseAsString.withDefault(defaultSchema))
+    return useQueryState(
```
ivasilov, line 15: Previously, if you chose a schema in the schema selector which matches the original schema when the page loaded, it would not show up in the query param. This change fixes that.

apps/studio/components/layouts/Tabs/SortableTab.tsx
```diff
@@ -28,8 +28,7 @@ export const SortableTab = ({
   openTabs: Tab[]
   onClose: (id: string) => void
 }) => {
-  const router = useRouter()
-  const currentSchema = (router.query.schema as string) || 'public'
+  const { selectedSchema: currentSchema } = useQuerySchemaState()
```
ivasilov, line 31: `useQuerySchemaState` is the best source for selected schema, whether in local storage or query param.

apps/studio/components/layouts/Tabs/SortableTab.tsx
```diff
@@ -45,16 +44,8 @@ export const SortableTab = ({
   const shouldShowSchema = useMemo(() => {
     // For both table and schema tabs, show schema if:
     // Any tab has a different schema than the current schema parameter
-    if (tab.type === 'r') {
```
ivasilov, line 48: Show schema names on all type of tables.

ivasilov:
>   * Tab for foreign table doesn't have schema name if you have tables across schemas opened as tabs

Fixed.
>   * Was expecting to see "Viewing protected schema" warning here in table editor, schema has iceberg fdw

Fixed.
>   * (Unrelated) I'm not sure when this broke, but opened tabs are no longer getting persisted across browser refreshes

Left for another PR.

joshenlim (approved):
Verified changes locally! 🙂

PR #37300, opened 2025-07-18
Deprecate getAPIKeys from project settings v2, use new api keys endpoint instead

joshenlim:
## Changes involved
- Deprecate `getAPIKeys` from project-settings-v2-query
  - Replace all usage of this with `getKeys` from `api-keys-query`

- Add dashboard API endpoint `api-keys` which only returns `anon` and `service_role` keys -> just so that the changes do not break self-hosted

- Some features will not be available if legacy API keys have been disabled
  - Storage Explorer to use only service role key, if legacy API keys have been disabled, show a note instead to re-enable legacy keys for now
  - image
  - Realtime Inspector to disable role impersonation and default to anon role if legacy keys are disabled
  - image

- Cmd+K to show option to copy new keys from "Get API keys"
- image

- Integrations -> Webhooks and Cron, when adding new header for edge function, if using new secret key, also add apikey header
- image
- image

- Edge Functions -> Deploy a new function -> Via CLI, include `apikey` header in instructions if using publishable key
- image

- Edge Functions Details -> Invoke function via curl, include `apikey` header in instructions if using publishable key
- image

## Other notes

- JFYI, for Edge Functions Tester, if the edge function has verify jwt enabled + disabled legacy keys, testing will return a 401

- Have also deprecated all usage of API keys (`service_api_keys` property) from project settings v2, EXCEPT `DisplayApiSettings.tsx` -> can't replace with new API keys query as there's a piece of logic that retrieves when the legacy keys are last used, in which that logic depends on the `tag` property of the keys that is no longer present in the new api keys endpoint ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/components/ui/ProjectSettings/DisplayApiSettings.utils.ts#L45))

## To test
- [ ] Can upload files to Storage Explorer
- [ ] Can copy publishable key from Cmd+K
- [ ] Integrations > Webhooks / Cron -> Edge Functions, Add header -> Authorization
  - If using secret key, then also add a `apikey` header
- [ ] Edge Functions terminal instructions include `apikey` header if using publishable key
- [ ] Edge Functions details include `apikey` header if using publishable key
- [ ] Test self-hosted


hf:
Thank you for all of this!

ivasilov (approved):

PR #37308, opened 2025-07-18
Refactor vault `AddNewSecretModal` and add associated UI test

Saeris:

YES

Updates `AddNewSecretModal` for the Vault Secrets management screen to use shadcn based components and form management with react-hook-form + zod.

Adds an associated UI test for this modal and updates `SecretsManagement` since it no longer needs to manage visibility for this modal.

Branched from #37258 which has associated changes to testing setup and dependencies, so it should be reviewed and merged after that PR is accepted.

## Additional context


jordienr (changes requested):
Hey Drake. just one nitpick from me:

Think we could reduce some duplicate code by moving the form to its own `SecretForm` component with the zod schema, layout, etc. 

Tests look great. Ran them and tried breaking the component and they caught it. Thanks!

Saeris:
> Think we could reduce some duplicate code by moving the form to its own `SecretForm` component with the zod schema, layout, etc.

Figure you mean like with the Webhook Create/Edit form yeah? I'm open to it but would prefer to do that as a follow-up PR if you don't mind.

PR #37310, opened 2025-07-18
chore(docs): add logflare logging

charislam:


apps/docs/lib/logger.ts
```diff
@@ -0,0 +1,76 @@
...
+    transmit: {
+      level: 'info',
+      send: function (_level, logEvent) {
+        // [Charis 2025-07-21] Surprisingly not filtered by enabled
+        if (IS_CI || IS_DEV) return
+
+        const { messages } = logEvent
+        const remappedMessages: Partial<LogflareMessage>[] = []
+
+        let currentMessage: Record<string, unknown> | null = null
```
Ziinc: > The messages array is all arguments passed to logger method, (for instance logger.info('a', 'b', 'c') would result in messages array ['a', 'b', 'c']).

https://getpino.io/#/docs/browser?id=transmit-object

from the added route logging, the first object in the array would be the object, the second would be the enum.

you might want to swap the logger args such that:
- first arg is the event message/enum
- second arg is the metadata to merge into the payload

Ziinc: you probably don't want them to be separate events, since this function is called on each `logger.info(...)` call.

note that this won't do batching for you, in case there are performance concerns, as it would be in the same event loop as the server.

Ziinc (approved):

PR #37341, opened 2025-07-21
Table editor cast column type to text if filtering with like based filter

joshenlim:
## Context

Currently if you try to filter via `LIKE` on a column that's not a string type (e.g UUID), you get this Postgres error:
image

While it's technically correct, we could make this UX a little better by type casting the column to text whenever filtering using a LIKE, or ILIKE filter

## Changes involved
- Cast column to text when filtering on LIKE or ILIKE filter in table editor

## To test
- [ ] Can filter by LIKE or ILIKE on a UUID column (or any non text column)


alaister (approved):
LGTM!

- Tested on preview

PR #37349, opened 2025-07-21
Refactor custom domain verification

joshenlim:
## Context

We've seen a couple of tickets whereby users hit "Cancel" on the 2nd step of setting up custom domains seems to trigger an odd unrelated error as such:
image

2 observations here:
- There's a `useInterval` in `CustomDomainVerify.tsx` [here](https://github.com/supabase/supabase/blob/dbb413beebcb07532d5d0ebbf4fbd837652293b4/apps/studio/components/interfaces/Settings/General/CustomDomainConfig/CustomDomainVerify.tsx#L53), in which the polling doesn't get cancelled if the user has clicked to delete the custom domain. Hence there's a chance that the /reverify request subsequently errors out, leading to a seemingly random error message
- Deleting a custom domain, based on the API implementation, is an async process that depends on a queue. As such the custom domain of a project may not be immediately updated upon clicking delete.

## Changes involved
- [ ] Set the delay for the `useInterval` in `CustomDomainVerify.tsx` to be false if user is deleting custom domain
- [ ] Add + Update toast messages for deleting custom domain that it may take a few seconds before changes are reflected
- [ ] Also small refactor to `CustomDomainVerify` to just use the custom domain RQ, instead depending on a prop from parent

## To test
- [ ] Test that you can configure a custom domain from start to end

### DIFF
--- a/apps/studio/components/interfaces/Settings/General/CustomDomainConfig/CustomDomainActivate.tsx
+++ b/apps/studio/components/interfaces/Settings/General/CustomDomainConfig/CustomDomainActivate.tsx
@@ -33,7 +33,7 @@ const CustomDomainActivate = ({ projectRef, customDomain }: CustomDomainActivate
   const { mutate: deleteCustomDomain, isLoading: isDeleting } = useCustomDomainDeleteMutation({
     onSuccess: () => {
       toast.success(
-        'Cancelled setting up custom domain. It may take a few seconds before your custom domain is fully removed, hence you may need to refresh your browser.'
+        'Custom domain setup cancelled successfully. It may take a few seconds before your custom domain is fully removed, so you may need to refresh your browser.'
       )
     },
   })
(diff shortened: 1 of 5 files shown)


apps/studio/components/interfaces/Settings/General/CustomDomainConfig/CustomDomainActivate.tsx
```diff
@@ -30,7 +30,13 @@ const CustomDomainActivate = ({ projectRef, customDomain }: CustomDomainActivate
       },
     }
   )
-  const { mutate: deleteCustomDomain, isLoading: isDeleting } = useCustomDomainDeleteMutation()
+  const { mutate: deleteCustomDomain, isLoading: isDeleting } = useCustomDomainDeleteMutation({
+    onSuccess: () => {
+      toast.success(
+        'Cancelled setting up custom domain. It may take a few seconds before your custom domain is fully removed, hence you may need to refresh your browser.'
```
alaister: wdyt?
```suggestion
        'Custom domain setup cancelled successfully. It may take a few seconds before your custom domain is fully removed, so you may need to refresh your browser.'
```

alaister (approved):
LGTM!

- Tested on preview

kimsoo527 (approved):

PR #37370, opened 2025-07-22
fix: Make the database.new redirect work

ivasilov:
This PR adds a special case for the `/new/last-visited-org` URL which will redirect the user to create a project in the last visited org.

### DIFF
--- a/apps/studio/pages/new/[slug].tsx
+++ b/apps/studio/pages/new/[slug].tsx
@@ -148,7 +148,11 @@ const Wizard: NextPageWithLayout = () => {
 
   // This is to make the database.new redirect work correctly. The database.new redirect should be set to supabase.com/dashboard/new/last-visited-org
   if (slug === 'last-visited-org') {
-    router.replace(`/new/${lastVisitedOrganization}`, undefined, { shallow: true })
+    if (lastVisitedOrganization) {
+      router.replace(`/new/${lastVisitedOrganization}`, undefined, { shallow: true })
+    } else {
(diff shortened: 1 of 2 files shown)


staaldraad (approved):

apps/studio/pages/new/[slug].tsx
```diff
@@ -146,6 +146,11 @@ const Wizard: NextPageWithLayout = () => {
     ''
   )
 
+  // This is to make the database.new redirect work correctly. The database.new redirect should be set to supabase.com/dashboard/new/last-visited-org
+  if (slug === 'last-visited-org') {
+    router.replace(`/new/${lastVisitedOrganization}`, undefined, { shallow: true })
+  }
```
Comment (joshenlim) [inline apps/studio/pages/new/[slug].tsx]: should also consider if `lastVisitedOrganization` is undefined here

```suggestion
  if (slug === 'last-visited-org' && !!lastVisitedOrganization) {
    router.replace(`/new/${lastVisitedOrganization}`, undefined, { shallow: true })
  }
```

Comment (joshenlim) [reply apps/studio/pages/new/[slug].tsx]: otherwise i'd reckon redirect then to `/new/_` otherwise if `lastVisitedOrganization` is undefined, and we should then hide this warning if `slug === '_'`

main thing we need to aware of is to keep the users on this page, as the rule with `.new` domains is that they need to lead to a "creation" flow

image

Comment (ivasilov) [reply apps/studio/pages/new/[slug].tsx]: Fixed both cases, ty for catching!

joshenlim (approved):
All LGTM! Just pushed up a bit more nit improvements 🙂

PR #37376, opened 2025-07-22
feat(replication): Implement UI to render pipeline replication status

iambriccardo:
# Context
This PR implements a new UI to render the pipeline replication status which is needed to show per-table status of replication.

# Changes Involved
* There is a new `info` button that opens up a new screen which makes an API request to `replication-status` endpoint which returns per table the status of the replication.
  image
* There is a new screen which shows the replication state for each table of the publication of the pipeline.
  image
* The handling of the disabling and enabling state have been added within the new screen.
  image
* The failed status now doesn't have anymore a loading state and it just shows failed.
  image
* Error messages have been unified to be shared across the main table containing all the destinations and the screen to show the replication status of a specific pipeline.
* A new context have been added to share state effectively between the components.


imor (approved):

joshenlim:
## Couple of notes

- We've got a component `ButtonTooltip` that composes of the Tooltip and Button components - can probably use that to prevent duplicate redeclaration of Tooltip and Button. Also don't need to redeclare TooltipProvider, we only need to declare this once in the app (we've done it in `_app.tsx`). There's a couple of existing places where we declared the TooltipProvider but those need cleaning up 😅 
  - Have fixed this in this [commit](https://github.com/supabase/supabase/pull/37376/commits/5270243f6e0505018c1f6f2a41889533514cd282)

- We've moved away from doing `export default` for components, can just `export const Component`

## Some UI feedback

- There might have been a bug introduced here - early access state no longer using the full width. Noticed that we're using the full width on staging.
  image

- Could add "Enable CTA" here - as a new user i wouldn't know where to enable the pipeline from. Added this in this [commit](https://github.com/supabase/supabase/pull/37376/commits/5270243f6e0505018c1f6f2a41889533514cd282) although will double check if I did it correctly once the preview is built hahaha was coding a little blind locally
  image

- New pipeline status UI should ideally be on a route of its own but okay to address this in a subsequent PR - left a comment in this [commit](https://github.com/supabase/supabase/pull/37376/commits/5270243f6e0505018c1f6f2a41889533514cd282)

joshenlim (approved):
Verified changes on staging preview, code changes LGTM as well

PR #37382, opened 2025-07-22
Fix variable name in self-send messages

Plaetorius:
Fix name: FROM "channelC" (never declared) to "myChannel", the name of the channel in the example

YES

Variable name fix

The code in the example is wrong

## What is the new behavior?

The naming of the variable is correct


charislam (approved):
Thanks @Plaetorius!

PR #37394, opened 2025-07-23
Fix unified logs stuff on self-hosted

joshenlim:
For local/self-hosted
- Hides the early access call out
- Fix redirect when landing on /logs to /logs/explorer


alaister (approved):
LGTM!

- Tested on self-hosted preview

PR #37399, opened 2025-07-23
Factor in filters and sorts for export via CLI option

joshenlim:
## Context

Related to the recent addition of this option: Export via CLI in the table editor
image

We currently only show this option if all rows have been selected, but it didn't consider any filters or sorts that were applied (including role impersonation). This PR hence addresses that for the Export to CSV via psql option, and also adds a note for Export to SQL via pg_dump option that filters cannot be considered.
image

Note that we're only showing this open when _all rows are selected_ for now as 
- It's tricky to generate the SQL to select specific rows from the table since it'll require the table to have a primary key, and we'd need to consider composite primary keys as well. We're using a query builder for this which doesn't really support composite primary keys for filters very well
- It's pretty straight forward to export them via the browser since they're already stored in memory, reckon supporting this might be overkill unless we have requests for them

[Update] With the changes in this [PR](https://github.com/supabase/supabase/pull/37513), we've also introduced the Export via CLI option from the Table Menu as well as such, but this option will export the table as a whole (will not factor in sorts, filters and role impersonation, as per the other export options here):
image

## Changes involved
- Export to CSV via psql option, command will consider any filters or sorts applied (this will align with the `fetchAllTableRows` method, both are using the same function to generate the SQL)
- Export to SQL via pg_dump option, show warning that filters cannot be considered

## To test
- [ ] Verify that "Export to CLI" option only shows up if all rows are selected
- [ ] Verify that the command for the "Export as CSV" works and data exported is correct


alaister (approved):
LGTM!

- Tested on preview

One small note:
Screenshot 2025-08-04 at 20 13 43
clicking these three dots clears any sorts/filters currently applied since I think the click propagates through to the table name

joshenlim:
great catch will push up a fix! 🙏

kimsoo527 (commented):
Npm run

kimsoo527 (commented):
@kimsoo527

PR #37401, opened 2025-07-23
Fix visibility of edge fn charts for enterprise

joshenlim:
## Context

Noticed that edge function reports charts were incorrectly showing a disabled state for an organization on the enterprise plan as such:
image

Was just missing the values in `availableIn` under `edgefn-charts.ts`

## Changes involved

- Add `enterprise` to all charts within `edgefn-charts.ts`
- Refactor `ReportChart.tsx`
  - Removed param `availableIn`, seems duplicated and unnecessary as the chart already has that parameter
    - There was also some inconsistency of using `availableIn` from the parent props, and `chart.availableIn`
  - Removed param `orgPlanId`, can be obtained via the `useCurrentOrgPlan` hook, doesn't need to be passed as a prop
- Removed unused `isFreePlan` param from `getAuthReportAttributes` in `auth-charts.ts`
- The other changes are all import statement sorting by prettier

### DIFF
--- a/apps/studio/components/interfaces/Reports/ReportChart.tsx
+++ b/apps/studio/components/interfaces/Reports/ReportChart.tsx
@@ -7,16 +7,19 @@
  * This component acts as a bridge between the data-fetching logic and the
  * presentational chart component.
  */
-import { useFillTimeseriesSorted } from 'hooks/analytics/useFillTimeseriesSorted'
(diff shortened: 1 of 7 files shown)


alaister (approved):
LGTM!

- Tested with an enterprise org on preview

PR #37409, opened 2025-07-23
feat(auth): add audit logs configuration page

cemalkilic:

Adding the configuration page for auth audit logs.

[Joshen] JFYI dependant on a BE change, verify that changes have gone live prior to merging
We'll also need the docs URL to update on the FE if possible prior to merging too

## What is the new behavior?

[Joshen] Updating screenshots here based on latest commit

image

If configuration is currently toggled off, and the user is toggling it on, the following warning will show up
image

## Additional context

After updating the API, gonna run `codegen` to get the updated api schema.


joshenlim (approved):
LGTM 🙂 Feel free to merge whenever ready - i'll leave it in drafts still for now

joshenlim (approved):
Double approve! Just note the following
- Let's merge this after the [docs PR](https://github.com/supabase/supabase/pull/37548) goes through so we won't have a broken link
- ✅ We should update the "Field reference" for Auth Audit Logs as well in the Logs Explorer
  - This same information needs to be updated in the docs as well ([ref comment](https://github.com/supabase/supabase/pull/37548#issuecomment-3269648743))

PR #37418, opened 2025-07-23
fix: use gotrue_id for member search and partition

cstockton:
It is possible to have multiple emails from different providers, a common example being SSO users. The user flow being:

1) Sign up with email using name@company.local
2) Upgrade to teams or enterprise
3) Enable SSO for the Org using domain company.local
4) Sign in with name@company.local
5) The SSO provider now handles the login, assigning new gotrue_id but uses the same primary_email
6) View the teams members and one will be omitted

This change fixes the case above so that both users are displayed. It also uses the gotrue_id to ensure that the "You" badge displays the correct user.


ivasilov (approved):
LGTM.

PR #37422, opened 2025-07-24
Table definition, actions and query

SaxonF:
https://github.com/user-attachments/assets/590c2277-6c81-40df-91fe-5514d0cf0f12

- Moves table definition from table editor, to table page
- Table name now clickable
- Toggle to show list or definition in table page
- Refresh button moved to header
- ~~Toggle added to table editor toolbar to show query behind the results (e.g. filtering etc)~~ Removed definition from table editor

Table editor is now focused on results, much like SQL Editor, and table details can be found in table list.


awaseem:

PR #37426, opened 2025-07-24
Table editor test mode

SaxonF:
## Context

This is related to our RLS testing policies project, specifically on the manual testing side of things

image

This adds a "test mode" to table editor which wraps any table row mutations in a transaction with a rollback. We link to this mode from policies page via a "test" button on each table. This prompts user to set impersonation and enables test mode by default.

Requires enabling `enableTestMode` configCat flag


joshenlim (commented):
Pushed up some refactors and fixes, some comments below:

## Potential issue

- There's a problem with this strategy as despite doing rollbacks, we're still actually firing the query which may introduce potential unwanted side effects 
  - e.g For a table that has an integer column defined as `generated by default as identity`, when adding a new row even if we're rolling back the addition, the identity count increases and doesn't go back. 
    - So e.g if I had rows ID 1, 2, 3, and I add a row in test mode, now if i exit test mode and add another row, my rows will then be ID 1, 2, 3, 5 (instead of 1, 2, 3, 4)
  - e.g Users may have certain hooks or triggers on the table that might get triggered depending on the hook / trigger conditions (e.g inserting, updating or deleting rows) 

## Bugs

- Deleting a row by selecting via checkbox seems to open the role impersonation popover

## Feedback

- Wondering for the header here RE the ordering of the buttons
  - There's a mix of icon buttons and text buttons, and this [PR](https://github.com/supabase/supabase/pull/37422) will also be adding a refresh icon button at the right end of the header which makes it a little messy. No strong opinion though 🙏 
    image

- Tried to add a row in test mode, but was a little confused at first why there was no visual indication
  - While we do show toasts, i think the bright amber border of the test mode draws all the attention away from the toasts so I missed it easily. Also thinking we could just show one toast? This is happening when I add a new row
    image
  - Maybe its just a matter of saying that the rollback is immediate here (I can add that in)
    image

- Was initially confused why the role impersonation popover kept opening whenever i hit "Test" from the RLS policies page. For users who aren't familiar with RLS and Postgres, i reckon this might be confusing without additional info RE "how to test"
  image

- On a similar note to the previous point, wondering if we could improve the toast messaging during test mode to make it more clear on what the user is testing
  - Just wondering, as the focus of the test mode is to validate certain operations, rather than the data changes
  - `Successfully deleted selected row(s) and rolled back` -> `Successfully deleted selected row(s) as [role].`, and `Changes have been immediately rolled back.` could be like a toast description in a `text-foreground-lighter` perhaps

awaseem:

PR #37431, opened 2025-07-24
Shift pipeline status page to its own route

joshenlim:
## Context

This is related to the Database Replication page, previous PR was [here](https://github.com/supabase/supabase/pull/37376) which was to introduce a new pipeline status page from the database replication page. But the navigation between replications to pipeline status page (and vice versa) was all controlled via state.

So this PR changes that to have pipeline status page on its own route

## Changes involved

- Shift pipeline status page to its own route and refactor page navigation UI logic accordingly
- Reverted clicking on row to navigate to pipeline status page - realised this was inconsistent, so following how we do it with database/tables by reinstantiating a button to navigate "View status"
  image
- Updated long poll frequency for replication + pipeline status from 2s to 5s, think 2s is a bit too frequent esp if at scale, mgmt API server might be hammered

### DIFF
--- a/apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
+++ b/apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
@@ -1,3 +1,4 @@
+import Link from 'next/link'
 import { useEffect, useState } from 'react'
 import { toast } from 'sonner'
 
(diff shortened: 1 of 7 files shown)


iambriccardo (approved):
UI changes look good to me!

Will let other frontend members approve.

ivasilov (approved):

PR #37442, opened 2025-07-24
feat: add client-side PostHog telemetry tracking

seanoliver:
## What does this PR do?

Adds client-side PostHog tracking to run in parallel with server-side telemetry across `studio`, `docs`, and `www`. This enables session replays and resolves a race condition where page views arrive before group assignments resulting in attribution errors.

Resolves GROWTH-438
Resolves GROWTH-271

## Changes

- Created PostHog client wrapper with consent-aware initialization in common package
- Integrated PostHog client calls into existing telemetry functions to send events to both PostHog (client) and backend (server)
- Updated CSP to allow connections to PostHog endpoints
- Added environment variable support for all apps

## Implementation Details

- PostHog client accepts consent as a parameter and respects user preferences
- Events can be distinguished in PostHog by `$lib` property (`posthog-js` vs `posthog-node`)
- PostHog URL configured based on environment (staging/local uses `ph.supabase.green`)
- Maintains full backward compatibility with existing telemetry system

During this migration, we should be seeing both client and server-side events in posthog logged across `studio`, `docs`, and `www`:
CleanShot 2025-07-31 at 16 26 50

Client-side events are now correctly grouped by org only if they have an `orgSlug` or a corresponding `projectRef` in the URL (some `posthog-node` PVs are still being incorrectly bucketed but these events are effectively deprecated with this change and will be removed in an upcoming PR):
CleanShot 2025-07-31 at 16 47 45

Client-side page views that aren't grouped to a specific project or org are still being tracked, but they no longer clutter the org-specific event feed:
CleanShot 2025-07-31 at 16 50 53

## Local Development Setup

To test this locally, add the following environment variables to each app's `.env.local` file:

```
NEXT_PUBLIC_IS_PLATFORM="true"
NEXT_PUBLIC_ENVIRONMENT="local"
NEXT_PUBLIC_POSTHOG_KEY="phc_ihjC3EeB2wXCt87yccX5idgIgeZsub7WG0XR5hGFhJz"
```

Note: `studio` already has the first two variables configured, so only `NEXT_PUBLIC_POSTHOG_KEY` needs to be added there.

## Production/Staging Deployment

The following has already been done in separate PRs/tasks:
- `NEXT_PUBLIC_POSTHOG_KEY` configured in all Vercel environments
- PostHog reverse proxy (ph.supabase.green and ph.supabase.com) configured to accept client-side requests

This should work in production and staging environments. Testing will be needed after merge to confirm.

## Next Steps

1. Test in staging/production environments after merge
2. Monitor both event streams during migration period
3. Eventually deprecate server-side page tracking

## Testing Instructions

1. **Run the Studio app locally.**
2. **In PostHog**, switch to the **staging environment**.
3. Navigate to **People → Organizations**, choose one of your local organizations, and open the **Events** tab.
4. In a separate browser tab, visit an **organization-specific page** (e.g., `/org/[orgSlug]` or `/project/[projectRef]`).
5. **Refresh the events list** in PostHog:

   * You should see events from both the **`web` library** (frontend) and **`posthog-node`** (backend).
   * It can take a few minutes for events to appear.
6. Visit an **unassociated page** (e.g., `/organizations` or `/new`):

   * **Expected:** No `web` library page view events.
   * **Note:** Some `posthog-node` events may still appear due to client-server race conditions.
7. **Optional:** Run `docs` and `www` locally:

   * Page views from these apps should **not** appear in the **organization-specific** event list.
   * However, if you click your **user ID** and view the **user-specific events** list, you should see **all** page views (including `docs` and `www`).

### DIFF
--- a/apps/studio/csp.js
+++ b/apps/studio/csp.js
@@ -26,6 +26,11 @@ const SUPABASE_CONTENT_API_URL = process.env.NEXT_PUBLIC_CONTENT_API_URL
   ? new URL(process.env.NEXT_PUBLIC_CONTENT_API_URL).origin
   : ''
 
+const isDevOrStaging =
+  process.env.NEXT_PUBLIC_VERCEL_ENV === 'preview' ||
+  process.env.NEXT_PUBLIC_ENVIRONMENT === 'local' ||
+  process.env.NEXT_PUBLIC_ENVIRONMENT === 'staging'
+
 const SUPABASE_STAGING_PROJECTS_URL = 'https://*.supabase.red'
 const SUPABASE_STAGING_PROJECTS_URL_WS = 'wss://*.supabase.red'
 const SUPABASE_COM_URL = 'https://supabase.com'
@@ -58,6 +63,7 @@ const SUPABASE_ASSETS_URL =
   process.env.NEXT_PUBLIC_ENVIRONMENT === 'staging'
     ? 'https://frontend-assets.supabase.green'
     : 'https://frontend-assets.supabase.com'
+const POSTHOG_URL = isDevOrStaging ? 'https://ph.supabase.green' : 'https://ph.supabase.com'
 
 const USERCENTRICS_URLS = 'https://*.usercentrics.eu'
 const USERCENTRICS_APP_URL = 'https://app.usercentrics.eu'
@@ -89,6 +95,7 @@ module.exports.getCSP = function getCSP() {
     USERCENTRICS_URLS,
     STAPE_URL,
     GOOGLE_MAPS_API_URL,
+    POSTHOG_URL,
   ]
   const SCRIPT_SRC_URLS = [
     CLOUDFLARE_CDN_URL,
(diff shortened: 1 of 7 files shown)


doublethink (approved):
No changes required to Privacy Policy, PostHog already in applicable sections

apps/studio/csp.js
```diff
@@ -26,6 +26,11 @@ const SUPABASE_CONTENT_API_URL = process.env.NEXT_PUBLIC_CONTENT_API_URL
   ? new URL(process.env.NEXT_PUBLIC_CONTENT_API_URL).origin
   : ''
 
+const isDevOrStaging =
+  process.env.NEXT_PUBLIC_VERCEL_ENV === 'preview' ||
+  process.env.NEXT_PUBLIC_ENVIRONMENT === 'local' ||
+  process.env.NEXT_PUBLIC_ENVIRONMENT === 'staging'
```
joshenlim, line 32: just a nit but wondering if it might be better to check for isProduction? so that all other values will default to use `ph.supabase.green` instead (if for any reason the env value messes up then we won't affect prod PostHog)

`const isProduction` = process.env.NEXT_PUBLIC_ENVIRONMENT === 'prod'`

pamelachia (approved):
As long as dashboard is fine w it then lgtm

joshenlim (approved):
All good from FE end - verified events are getting sent to PH, though only can verify "web" events

PR #37451, opened 2025-07-25
add with data field to branch

SaxonF:
image

Adds data branching switch behind feature flag. If this is true, data will be copied from production to the new branch.


sweatybridge (approved):

joshenlim (approved):
Tested data branching with a project that has PITR - can verify that data is loaded

PR #37495, opened 2025-07-28
Clear IndexedDB

SaxonF:
Clear IndexedDB on sign out


joshenlim (approved):
Verified on staging preview that index db clears when signing out

PR #37500, opened 2025-07-28
Prompt and tool refactoring

SaxonF:
This PR attempts to refactor our chat and completion based AI endpoints to better share prompts and tools.
- Moves prompts into a lib prompts file
- Moves tools into a tools folder and runs all tools through a single tool filtering function
- Moves organisation opt in level checks into file to be used across generate-v4 and complete
- Adds a generic `code/complete` endpoint which replaces the old `sql/complete-v2` and `edge-functions/complete-v2` endpoints. This make use of same prompts and tools and generate-v4 except with different output prompts.
- Adds more RLS context
- Adds list_policies tool

Requires https://github.com/supabase/supabase/pull/37656 to be merged first


gregnr (approved):

apps/studio/lib/ai/tools/schema-tools.ts
```diff
@@ -0,0 +1,49 @@
...
+export const getSchemaTools = ({
+  projectRef,
+  connectionString,
+  authorization,
+}: {
+  projectRef: string
+  connectionString: string
+  authorization?: string
+}) => ({
+  list_policies: tool({
```
joshenlim, line 14: just out of curiosity - wondering why this doesn't live in MCP tools like list_tables, list_extensions, etc

joshenlim:
## Feedback

Some oddity here in the SQL editor completion while testing on staging preview
- Response from assistant included my prompt (Started with a blank snippet), SQL output is also semantically incorrect
  image
- In this case the assistant returned an empty string  
  image
- Tried using inline completion for a snippet that already has some content but it's including non-SQL stuff too   
  image

For edge functions, a prompt like this also returns a semantically wrong response, am expecting `message: "Hello World"`, not just `"Hello World"`
image

joshenlim (approved):
Tested AI Assistant and inline completion for SQL + Edge Functions editors

The quality still needs work nonetheless but its better than what's currently on staging

PR #37505, opened 2025-07-28
Fix tabs valtio store initial render

joshenlim:
## Context

Noticed that our tabs layout (in SQL Editor and Table Editor) doesn't persist opened tabs across browser sessions (e.g via refresh). Realised that `createTabsState` from `tabs.tsx` is only initially triggered when `projectRef` is still an empty string (since `useSelectedProject` hasn't loaded), but doesn't trigger again when the `projectRef` is updated subsequently once `useSelectedProject` is loaded - hence causing this bug

There's also a similar issue with the `role-impersonation-state` since both of these states share a similar implementation (e.g the selected role doesn't persist across browser sessions) which we'll look into in another PR - I'd want to double check if the fix here looks correct first, before doing it for other states

## To test
- [ ] Verify that tabs persist across browser sessions
- [ ] Verify that opening new tabs, closing tabs, etc all are still working as expected


apps/studio/state/tabs.tsx
```diff
@@ -399,8 +400,16 @@ export type TabsState = ReturnType<typeof createTabsState>
 export const TabsStateContext = createContext<TabsState>(createTabsState(''))
 
 export const TabsStateContextProvider = ({ children }: PropsWithChildren) => {
-  const project = useSelectedProject()
-  const state = useConstant(() => createTabsState(project?.ref ?? ''))
+  const { data: project } = useSelectedProjectQuery()
+
+  const initState = useConstant(() => createTabsState(project?.ref ?? ''))
```
ivasilov: You don't need the `useConstant`, you can just instantiate the tabsState in the `useState`, right?

ivasilov (approved):

apps/studio/state/tabs.tsx
joshenlim: yeahhhh was thinking about that too - i'll remove it 🙏

PR #37506, opened 2025-07-28
Add export as JSON option for table editor

joshenlim:
## Context

Fixes FE-1636
From a user's request [here](https://x.com/maxrusakovic/status/1920111895807189006) 🙂 

Adds an export to JSON option in the Table Editor, but only when selecting rows within the same page (will not show up if all rows in the table are selected)
image

### DIFF
--- a/apps/studio/components/grid/components/header/Header.tsx
+++ b/apps/studio/components/grid/components/header/Header.tsx
@@ -277,7 +277,7 @@ const RowHeader = () => {
     })
   }
 
-  async function onRowsExportCSV() {
+  const onRowsExportCSV = async () => {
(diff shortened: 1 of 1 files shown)


ivasilov:
Should we have `export to JSON` in the table context menu also?
Screenshot 2025-07-28 at 15 52 52

ivasilov (approved):

joshenlim:
> Should we have export to JSON in the table context menu also?

ohhh good point - i don't think we should have export to JSON here, since i'm limiting export to JSON to just selected rows within a page (and not the whole table)

but, I should include the export via CLI option here actually

PR #37510, opened 2025-07-28
feat: Display OAuth client secret last used time

ivasilov:
Add display of last_used_at timestamp for OAuth client secrets in Studio. Shows relative time (e.g., "Last used 2 hours ago") when available, or "Never used" for unused secrets.

Screenshot 2025-07-28 at 14 47 03


joshenlim (approved):
LGTM! Verified locally

PR #37386, opened 2025-07-22
Refactor vault `DeleteSecretModal` and add associated UI test

Saeris:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

Continuing with the refactor work in  #37258 and #37308, this implements the same dialog pattern as `EditSecretModal`, adds a simple UI test, and updates the associated call sites in `SecretsManagement` and `SecretRow`.

I will leave this in draft until the above linked PRs have been reviewed and merged, as this is branched off of those.

## Additional context

Part of general deprecated component refactor work at the request of @saltcod

PR #37517, opened 2025-07-28
Refactor Storage Create/Edit/Empty/Delete Modals to use Shadcn components

Saeris:

YES

Following the same implementation patterns as in #37258, #37308, and #37386, this PR refactors `CreateBucketModal`, `EditBucketModal`, `EmptyBucketModal`, and `DeleteBucketModal` to use the shadcn based form controls and dialog patterns. It also cleans up `StorageMenu` and `BucketRow` to reduce the overall complexity involved in the usage of these modals. Finally, it adds some basic tests for each of these modals.

Some before/after screenshots:

`CreateBucketModal`
image

`EditBucketModal`
image

`EmptyBucketModal`
image

`DeleteBucketModal`
image


apps/studio/components/interfaces/Storage/__tests__/CreateBucketModal.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ProjectContextProvider } from 'components/layouts/ProjectLayout/ProjectContext'
+import { render } from 'tests/helpers'
+import { routerMock } from 'tests/lib/route-mock'
+import CreateBucketModal from '../CreateBucketModal'
+
+describe(`CreateBucketModal`, () => {
+  beforeEach(() => {
+    vi.mock(`hooks/misc/useCheckPermissions`, () => ({
+      useCheckPermissions: vi.fn().mockImplementation(() => true),
+    }))
```
Saeris: this seemed to be a better solution than trying to mock the auth context provider, but I'd be curious to know if there's an existing pattern I should follow instead for future reference

apps/studio/components/interfaces/Storage/__tests__/DeleteBucketModal.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+          <DropdownMenuItem onClick={() => setModal(`delete`)}>Delete</DropdownMenuItem>
+        </DropdownMenuContent>
+      </DropdownMenu>
+
+      {renderModal()}
+    </ProjectContextProvider>
+  )
+}
+
+mockAnimationsApi()
```
Saeris: note that this is needed anytime we've got `DropdownMenu` in the component tree

apps/studio/components/interfaces/Storage/__tests__/CreateBucketModal.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    expect(mimeTypeInput).toHaveValue(``)
+    await userEvent.type(mimeTypeInput, `image/jpeg, image/png`)
+
+    const submitButton = screen.getByRole(`button`, { name: `Create` })
+
+    fireEvent.click(submitButton)
+
+    await waitFor(() =>
+      expect(routerMock.asPath).toStrictEqual(`/project/default/storage/buckets/test`)
+    )
```
Saeris, line 98: because this component doesn't have an `onClose` callback like the other modals, we instead wait on the route navigation to validate that we've successfully created a new bucket

apps/studio/components/interfaces/Storage/BucketRow.tsx
```diff
@@ -23,20 +26,25 @@ export interface BucketRowProps {
...
+      case `edit`:
+        return <EditBucketModal bucket={bucket} onClose={onClose} />
+      case `empty`:
+        return <EmptyBucketModal bucket={bucket} onClose={onClose} />
+      case `delete`:
+        return <DeleteBucketModal bucket={bucket} onClose={onClose} />
+      default:
+        return null
+    }
+  }
```
Saeris: rather than using a visibility state, we can simply mount/unmount the active modal from the tree to achieve the same effect. the only reason why we need to separate the dialog content from its trigger is because we're placing the trigger inside of a `DropdownMenu`, wherein it's visibility state mounts/unmounts its options popover in the DOM

apps/studio/components/interfaces/Storage/DeleteBucketModal.tsx
```diff
@@ -77,27 +90,58 @@ const DeleteBucketModal = ({ visible = false, bucket, onClose }: DeleteBucketMod
...
+          <Label_Shadcn_ htmlFor="confirm">
+            Type <span className="font-bold text-foreground">{bucket.name}</span> to confirm.
+          </Label_Shadcn_>
+          <Input_Shadcn_
+            id="confirm"
+            value={value}
+            autoComplete="off"
+            onChange={(e) => setValue(e.target.value)}
+            placeholder="Type in name of bucket"
+          />
```
Saeris: it might be better to run validation on this input and provide an error message when it has an invalid value on submit rather than disabling the submit button, let me know what would be preferable

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -16,256 +42,332 @@ import { useBucketUpdateMutation } from 'data/storage/bucket-update-mutation'
...
+      name: bucket?.name ?? '',
+      public: bucket?.public,
+      has_file_size_limit: isNonNullable(bucket?.file_size_limit),
+      formatted_size_limit: fileSizeLimit ?? 0,
+      allowed_mime_types: (bucket?.allowed_mime_types ?? []).join(', '),
+    },
+    values: {
+      name: bucket?.name ?? '',
+      public: bucket?.public,
+      has_file_size_limit: isNonNullable(bucket?.file_size_limit),
```
Saeris, line 91: at least as far as the OpenAPI response type indicates, `file_size_limit` can be `number | undefined`, so I decided to pull in a handy `isNonNullable` utility here to guard against `null | undefined`. it's useful as a type guard in a variety of contexts (especially for narrowing types) so hopefully it can find broader use elsewhere too

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -16,256 +42,332 @@ import { useBucketUpdateMutation } from 'data/storage/bucket-update-mutation'
...
-    { projectRef: ref },
-    { enabled: IS_PLATFORM && visible }
-  )
+  const { mutate: updateBucket, isLoading: isUpdating } = useBucketUpdateMutation()
+  const { data } = useProjectStorageConfigQuery({ projectRef: ref }, { enabled: IS_PLATFORM })
   const { value, unit } = convertFromBytes(data?.fileSizeLimit ?? 0)
   const formattedGlobalUploadLimit = `${value} ${unit}`
 
-  const [selectedUnit, setSelectedUnit] = useState<StorageSizeUnits>(StorageSizeUnits.BYTES)
+  const [selectedUnit, setSelectedUnit] = useState<string>(StorageSizeUnits.BYTES)
```
Saeris, line 75: note that the type change here is necessary for `Select_shadcn` not to raise a type error

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -16,256 +42,332 @@ import { useBucketUpdateMutation } from 'data/storage/bucket-update-mutation'
...
-      hideFooter
-      visible={visible}
-      size="medium"
-      header={`Edit bucket "${bucket?.name}"`}
-      onCancel={onClose}
+    <Dialog
+      open
+      onOpenChange={(open) => {
+        if (!open) {
+          form.reset()
```
Saeris, line 136: thinking about it, `form.reset()` might not be necessary at all because the moment the dialog closes its unmounted from the component tree, which should reset its state implicitly

might come back and update this after testing that theory

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -16,256 +42,332 @@ import { useBucketUpdateMutation } from 'data/storage/bucket-update-mutation'
...
+                />
+              )}
+              <Collapsible_Shadcn_
                 open={showConfiguration}
                 onOpenChange={() => setShowConfiguration(!showConfiguration)}
               >
-                <Collapsible.Trigger asChild>
-                  <div className="w-full cursor-pointer py-3 px-5 flex items-center justify-between border-t border-default">
+                <CollapsibleTrigger_Shadcn_ asChild>
+                  <button className="w-full cursor-pointer py-3 flex items-center justify-between border-t border-default">
```
Saeris, line 233: for accessibility this trigger should be a button, not a div

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -16,256 +42,332 @@ import { useBucketUpdateMutation } from 'data/storage/bucket-update-mutation'
...
+                                name="formatted_size_limit"
+                                description={`Equivalent to ${convertToBytes(
+                                  formattedSizeLimit,
+                                  selectedUnit as StorageSizeUnits
+                                ).toLocaleString()} bytes.`}
+                              >
+                                <FormControl_Shadcn_>
+                                  <Input_Shadcn_
+                                    id="formatted_size_limit"
+                                    aria-label="File size limit"
```
Saeris, line 284: because we don't wrap this input with a label, we should at least give it an `aria-label` attribute describing what it is for screen readers. it also helps us to target the input with `screen.getByLabelText` in testing

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -16,256 +42,332 @@ import { useBucketUpdateMutation } from 'data/storage/bucket-update-mutation'
...
+                                    {...field}
+                                  />
+                                </FormControl_Shadcn_>
+                              </FormItemLayout>
+                            )}
+                          />
+                        </div>
+                        <Select_Shadcn_ value={selectedUnit} onValueChange={setSelectedUnit}>
+                          <SelectTrigger_Shadcn_
+                            aria-label="File size limit unit"
```
Saeris, line 296: same thing goes for select components that don't have a visible label

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -16,256 +42,332 @@ import { useBucketUpdateMutation } from 'data/storage/bucket-update-mutation'
...
+                    render={({ field }) => (
+                      <FormItemLayout
+                        name="allowed_mime_types"
+                        label="Allowed MIME types"
+                        labelOptional="Comma separated values"
+                        description="Wildcards are allowed, e.g. image/*. Leave blank to allow any MIME type."
+                      >
+                        <FormControl_Shadcn_>
+                          <Input_Shadcn_
+                            id="allowed_mime_types"
```
Saeris, line 342: general note to always provide an `id` attribute for an input, paired with a `name` attribute on the `FormItemLayout`, which will wire up the label with `htmlFor` to improve accessibility

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -16,256 +42,332 @@ import { useBucketUpdateMutation } from 'data/storage/bucket-update-mutation'
...
+            type="default"
+            disabled={isUpdating}
+            onClick={() => {
+              form.reset()
+              onClose()
+            }}
+          >
+            Cancel
+          </Button>
+          <Button form={formId} htmlType="submit" loading={isUpdating}>
```
Saeris, line 366: general note that we should always use the `form` attribute on submit buttons like this to wire up submit events when the button itself is not a child of a `form` element

apps/studio/tests/lib/route-mock.ts
```diff
@@ -1,11 +1,26 @@
...
+// remove the file extension or index segment
+const removeExt = (str: string) => str.replace(/.tsx|\/index.tsx/, ``)
+
+// Glob the `pages/` directory for all dynamic route segments
+const paths = [
+  ...(await Array.fromAsync(glob(`${pagesRoot}/**/*\].tsx`))),
+  ...(await Array.fromAsync(glob(`${pagesRoot}/**/**\]/**/index.tsx`))),
+].map(pipe(normalize, toPosix, removePrefix, removeExt))
+
+routerMock.useParser(createDynamicRouteParser(paths))
```
Saeris: this will make it easier to maintain our router mock by globbing for dynamic routes in our pages directory rather than having to manually maintain that list by hand

the `useParams` hook in particular relies on this list in order to know what route segments are parameters or not

apps/studio/tests/setup/polyfills.ts
```diff
@@ -35,3 +35,5 @@ Object.defineProperties(globalThis, {
   ReadableStream: { value: ReadableStream },
   TransformStream: { value: TransformStream },
 })
+
+window.HTMLElement.prototype.hasPointerCapture = vi.fn()
```
Saeris: a click event on `SelectTrigger_shadcn_` calls this browser API, so we need to mock it

apps/studio/components/interfaces/Storage/__tests__/CreateBucketModal.test.tsx
jordienr: We could improve the custom-render function to include some auth context but I think it's perfectly fine to mock the permission hook here.

apps/studio/components/interfaces/Storage/__tests__/CreateBucketModal.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    vi.mock(`hooks/misc/useCheckPermissions`, () => ({
+      useCheckPermissions: vi.fn().mockImplementation(() => true),
+    }))
+    // useParams
+    routerMock.setCurrentUrl(`/project/default/storage/buckets`)
+    // useSelectedProject -> Project
+    addAPIMock({
+      method: `get`,
+      path: `/platform/projects/:ref`,
+      // @ts-expect-error
```
jordienr, line 23: This is due to me making the addAPIMock type-safe forcing the response to be similar to the API response type generated by openapi lol.

But in reality it's just adding noise. Fine to expect error but think we could just allow any type to go into response in `addAPIMock` 🤔

apps/studio/components/interfaces/Storage/__tests__/DeleteBucketModal.test.tsx
jordienr: we should def add that to `apps/studio/tests/README.md`!

apps/studio/components/interfaces/Storage/DeleteBucketModal.tsx
jordienr: Yep +1 to error message.

jordienr (changes requested):
looking good, I think we could add some tests for error scenarios like network errors. 

One UI issue here in the padding and bottom border for the warning. Also similar issue happening in the edit dialog!

Image

prod for reference:

Image

apps/studio/components/interfaces/Storage/__tests__/DeleteBucketModal.test.tsx
Saeris: plan to write up some notes there as I discover more edge cases that need special treatment!

apps/studio/components/interfaces/Storage/__tests__/CreateBucketModal.test.tsx
Saeris: naw, you may have missed that I updated `addApiMock` in another PR to behave like the fetcher client, so this one is legit coming from the openapi response type. the data I'm using is a response I copied from my local, which is missing some required fields specified by the api spec

could make this more permissive in the types there with like `Partial<>`, but otherwise this is the desired behavior because for the most part we want to have errors crop up here as the API spec changes so we know when to expect different behavior to test for

there is also potential to improve this such that we can automatically mock the response data using a tool such as [`zod-mock`](https://github.com/anatine/zod-plugins/tree/main/packages/zod-mock) by leveraging generated response schemas, but that's work for another day

apps/studio/components/interfaces/Storage/__tests__/CreateBucketModal.test.tsx
Saeris: agree that it would be nice to have this in the custom render!

apps/studio/components/interfaces/Storage/BucketRow.tsx
```diff
@@ -113,6 +111,10 @@ const BucketRow = ({
       ) : (
         <div className="w-7 mr-1" />
       )}
+
+      <EditBucketModal visible={modal === `edit`} bucket={bucket} onClose={onClose} />
+      <EmptyBucketModal visible={modal === `empty`} bucket={bucket} onClose={onClose} />
+      <DeleteBucketModal visible={modal === `delete`} bucket={bucket} onClose={onClose} />
```
Saeris, line 117: changed this to follow the requested pattern from another set of PRs instead of using a switch block

apps/studio/components/interfaces/Storage/DeleteBucketModal.tsx
Saeris: Error message implemented!

Saeris:
> looking good, I think we could add some tests for error scenarios like network errors.
> 
> One UI issue here in the padding and bottom border for the warning. Also similar issue happening in the edit dialog!

Think I fixed this. Issue is that we switched from Modal to Dialog, and Dialog Content has a padding value on it, in addition to Admonition having a padding value of it's own.

jordienr (approved):

PR #37531, opened 2025-07-29
Add link to edge function logs for cron history if cron is using edge function

joshenlim:
Fixes FE-1340

RE Cron history page, if the cron job is using an edge function + the edge function is from the project
Adds a link to the respective edge function's logs page

image


apps/studio/components/interfaces/Integrations/CronJobs/PreviousRunsTab.tsx
```diff
@@ -177,8 +184,14 @@ export const PreviousRunsTab = () => {
     [fetchNextPage, isLoadingCronJobRuns]
   )
 
-  const currentJobState = cronJobs?.find((job) => job.jobid === jobId)
-  const cronJobRuns = useMemo(() => data?.pages.flatMap((p) => p) || [], [data?.pages])
+  useEffect(() => {
+    // Refetch only the first page
+    const timerId = setInterval(() => {
```
alaister: nit: could this be moved to the refetchInternal function of the query?

alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/interfaces/Integrations/CronJobs/PreviousRunsTab.tsx
joshenlim: opting to remove this instead actually - and letting `staleTime` from the query handle the refetching logic
cause this might be confusing if the user's not on the first page after the interval duration

PR #37537, opened 2025-07-29
fix(studio): empty team drop downs

kemaldotearth:
This fixes dropdowns returning empty on team invites when a user is signed in as a developer. For now it adds conditionals to show the option is not available for that role, but perhaps we can handle a little more gracefully.

YES

This fixes a small UI bug within the Teams area member list, where users signed in with the `Developer` role see empty drop down menus on pending invites.

Current behaviour returns as follows:
Screenshot 2025-07-29 at 13 19 18

## What is the new behavior?

For the time being I've updated the conditional checks on `canRevokeInvite` and `canSendInvite` to be `if` statements. This ensures we handle it somewhat gracefully for now. An open question is, can the `Developer` role do anything with the dropdown menu anyway? Might be worth hiding it from their scope?

Updated behaviour is like so:
Screenshot 2025-07-29 at 14 40 17

## Additional context

Please let me know if this is taken care of in another PR, if so, feel free to close this!

### DIFF
--- a/apps/studio/components/interfaces/Organization/TeamSettings/MemberActions.tsx
+++ b/apps/studio/components/interfaces/Organization/TeamSettings/MemberActions.tsx
@@ -27,6 +27,7 @@ import {
   DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from 'ui'
+import { DropdownMenuItemTooltip } from 'components/ui/DropdownMenuItemTooltip'
 import ConfirmationModal from 'ui-patterns/Dialogs/ConfirmationModal'
 import { LeaveTeamButton } from './LeaveTeamButton'
 import { useGetRolesManagementPermissions } from './TeamSettings.utils'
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Organization/TeamSettings/MemberActions.tsx
```diff
@@ -190,15 +190,22 @@ export const MemberActions = ({ member }: MemberActionsProps) => {
             <>
               {isPendingInviteAcceptance ? (
                 <>
-                  {canRevokeInvite && (
+                  {canRevokeInvite ? (
                     <DropdownMenuItem onClick={() => handleRevokeInvitation(member)}>
```
ivasilov: Merge the two `DropdownMenuItem`, then depending on `canRevokeInvite`: 
1. set the `disabled` prop 
2. render the correct paragraph

Same feedback for the other dropdown item.

joshenlim:
Nice! Likely overlooked this one RE the empty dropdown. Also 💯 great PR description - helps a lot to get through the changes quickly 🙂🙏 

For context we've been adopting a pattern of using Tooltips in this case for disabled actions (using `DropdownMenuItemTooltip`)([example](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/PolicyRow.tsx#L152)) - wondering what you reckon is a better UX? (As compared to textual descriptions within the `DropdownMenuItem` itself)
image

> Might be worth hiding it from their scope?

Guessing you're referring to hiding the dropdown menu? Ideally we'd avoid hiding UIs in these scenarios but offer some visual indication as to why the action isn't allowed (as a pattern). So e.g disable buttons, instead of hiding the button CTAs entirely

kemaldotearth:
> For context we've been adopting a pattern of using Tooltips in this case for disabled actions (using `DropdownMenuItemTooltip`)([example](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/PolicyRow.tsx#L152)) - wondering what you reckon is a better UX? (As compared to textual descriptions within the `DropdownMenuItem` itself) 

If we're using this pattern elsewhere, then I can mimic that. I quite like the inline one as it requires "less effort" to know what you're looking at. I guess that would require changes elsewhere? I'll update to use the tooltip, gets rid of the need for the conditionals in the current way. 👍 
 
> Guessing you're referring to hiding the dropdown menu? Ideally we'd avoid hiding UIs in these scenarios but offer some visual indication as to why the action isn't allowed (as a pattern). So e.g disable buttons, instead of hiding the button CTAs entirely

Good to know re: hiding the dropdown. Then let's forget that, was just thinking out loud :) 

Thanks @joshenlim @ivasilov !

joshenlim:
> i guess that would require changes elsewhere

Yup - whichever way we decide it'll be best to keep it as a consistent pattern across the dashboard. We can do the tooltips for now and revisit this as a whole again separately 🙂

joshenlim (approved):
LGTM! Verified changes locally 😄

PR #37547, opened 2025-07-29
docs: adjust pricing wording for log drains

Ziinc:
This adjusts the log drains pricing wording to be clearer on how it is billed, which is per hour.
https://supabase.slack.com/archives/C08D303DC57/p1753810942453589?thread_ts=1753720546.522159&cid=C08D303DC57

### DIFF
--- a/packages/shared-data/pricing.ts
+++ b/packages/shared-data/pricing.ts
@@ -590,7 +590,7 @@ export const pricing: Pricing = {
           free: false,
           pro: false,
           team: [
-            '$60 per drain per month',
+            '$0.0822 per drain, per hour',
             '+ $0.20 per million events',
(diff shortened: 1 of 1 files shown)


kevcodez (commented):
All other add-on prices in the pricing table are also communicated as "per month", so this would make that inconsistent.

A more thorough explanation for log drain billing is found here: https://supabase.com/docs/guides/platform/manage-your-usage/log-drains

I think for the pricing page, the monthly values make more sense to get a better understanding and not have the user calculate it. Log drain are also something people don't just run for two hours

charislam (approved):

Ziinc:
Closing in favour of Kevin's per month wording.

PR #37561, opened 2025-07-30
Refactor Storage Explorer `ConfirmDeleteModal`, `CustomExpiryModal`, and `MoveItemsModal` to use shadcn components

Saeris:
Following the same implementation patterns as in #37517 this PR refactors each of these modals to use our shadcn based components and react-hook-form patterns.

I eliminated the need to pass props to each of these by moving the logic associated with `useStorageExplorerStateSnapshot()` into the components themselves and out of `StorageExplorer.tsx`.

While I believe it would be ideal to have each of these modals be co-located with their dialog trigger buttons, due to them being located in multiple different places, each with slightly different visual representations, and the modal open state being tied to the internal state of `useStorageExplorerStateSnapshot()`, I've opted to keep them in `StorageExplorer.tsx` despite the fact that it does not contain any of the triggers directly.

For each modal I've added an associated UI test. This PR includes the changes to the testing setup that were added in #37517.

Note that currently there isn't strong validation on folder paths in `MoveItemsModal` and it may be worth considering updating the zod schema here to filter for invalid path segments. I'm not sure if we are already doing this sort of validation on the server, but it would be a better overall user experience if we can catch malformed paths early and alert the user to it before failing at the API call.


apps/studio/components/interfaces/Storage/StorageExplorer/CustomExpiryModal.tsx
```diff
@@ -10,94 +31,141 @@ const unitMap = {
...
+                  key="expiresIn"
+                  name="expiresIn"
+                  control={form.control}
+                  render={({ field }) => (
+                    <FormControl_Shadcn_>
+                      <Input_Shadcn_
+                        aria-label="Duration"
+                        id="expiresIn"
+                        type="number"
+                        min={0}
```
Saeris, line 106: note that the `min` attribute here eliminates the need for the previous validation logic, as the input itself will prevent values below 0 from being entered

apps/studio/components/interfaces/Storage/StorageExplorer/CustomExpiryModal.tsx
```diff
@@ -10,94 +31,141 @@ const unitMap = {
...
+                          <SelectValue_Shadcn_ asChild>
+                            <>{units}</>
+                          </SelectValue_Shadcn_>
+                        </SelectTrigger_Shadcn_>
+                        <SelectContent_Shadcn_>
+                          {Object.keys(unitMap).map((unit: string) => (
+                            <SelectItem_Shadcn_ key={unit} value={unit} className="text-xs">
+                              {unit}
+                            </SelectItem_Shadcn_>
+                          ))}
```
Saeris, line 136: note that this now ties the option values to the keys of `unitMap`, so adding a new key there will update this list

apps/studio/components/interfaces/Storage/StorageExplorer/MoveItemsModal.tsx
```diff
@@ -1,90 +1,127 @@
...
-  visible = false,
-  selectedItemsToMove = [],
-  onSelectCancel = noop,
-  onSelectMove = noop,
-}: MoveItemsModalProps) => {
-  const [moving, setMoving] = useState(false)
-  const [newPath, setNewPath] = useState('')
+const PathSchema = z.object({
+  path: z.string().min(0, 'Please provide a valid path name'),
+})
```
Saeris, line 29: we probably want this to validate against a regex or something similar, so that only valid directory paths can be entered

apps/studio/components/interfaces/Storage/StorageExplorer/MoveItemsModal.tsx
```diff
@@ -1,90 +1,127 @@
...
+                  <FormItemLayout
+                    name="path"
+                    label={`Path to new directory in ${selectedBucket.name}`}
+                    description="Leave blank to move items to the root of the bucket"
+                  >
+                    <FormControl_Shadcn_>
+                      <Input_Shadcn_
+                        id="path"
+                        {...field}
+                        autoFocus
```
Saeris, line 104: so far this is the only modal I've come across that had `autoFocus` set on an input. I'm not sure we want to be doing this as it could interfere with screen readers

apps/studio/components/interfaces/Storage/StorageExplorer/ConfirmDeleteModal.tsx
```diff
@@ -1,27 +1,49 @@
...
+          await deleteFolder(itemToDelete)
+        } else if (itemToDelete.type === STORAGE_ROW_TYPES.FILE) {
+          await deleteFiles({ files: [itemToDelete] })
+        }
+      } else {
+        await deleteFiles({ files: selectedItemsToDelete })
+      }
+      onClose?.()
+    })
+  }
```
Saeris, line 46: I might actually need to revisit this implementation. I thought it would be a good idea to use `useTransition` to tract the submitting state but I must be missing something since it appears to resolve immediately instead of when the API call promise returns

Hallidayo:

PR #37562, opened 2025-07-30
Fix cron panel if unable to parse command

joshenlim:
## Context

There's a bug with the Cron Job editor when editing an existing cron job of the type "http_request" that was created outside of the cron job interface.

For cron jobs created via this interface (specifically http request types, the declaration of the SQL for the job's command is technically consistent / deterministic, as the dashboard calls the `net.http_xxx` function using named notation. For example:
image
In this case, when editing such cron job, the dashboard extracts each parameter value through a series of regex [here](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Integrations/CronJobs/CronJobs.utils.ts#L41) (`parseCronJobCommand`), and we're subsequently able to render those values in the side panel editor.

However, cron jobs can also be created through direct SQL (e.g via the SQL editor), in which users may call the `net.http_xxx` function using positional notation (our pg net [docs](https://github.com/supabase/pg_net?tab=readme-ov-file#requests-api) also show examples with this notation) - in which case `parseCronJobCommand` will not be able to extract the parameters since it relies on the `net.http_xxx` call declared with named notation (and hence the side panel will not have any values when editing that cron job through the UI). For example:
image

## Changes involved

Since we shouldn't be expecting users to declare cron jobs a specific way, and its way too broad to be writing regex for positional notation, am opting to default back to "sql snippet" type if we derive that the cron job is of type "http_request" + we're unable to extract any information from the command. (and only doing this in this specific scenario)

## To test
- Create a cron job through the SQL editor using positional notation, e.g:
```
SELECT net.http_post(  
  'https://webhook.site/dacc2028-a588-462c-9597-c8968e61d0fa',
  '{"message":"Hello from Supabase"}'::jsonb,
  '{}'::jsonb,
  '{"Content-Type":"application/json"}'::jsonb
);
```
- Edit that cron job via the cron job interface, and verify the default type is SQL snippet, and the cron command is rendered correctly
- Verify as well that editing cron jobs created through the cron job UI are all still okay as well (http, edge function, etc)

### DIFF
--- a/apps/studio/components/interfaces/Account/Preferences/AnalyticsSettings.tsx
+++ b/apps/studio/components/interfaces/Account/Preferences/AnalyticsSettings.tsx
@@ -1,21 +1,21 @@
-import { Alert_Shadcn_, AlertDescription_Shadcn_, AlertTitle_Shadcn_, Badge, Toggle } from 'ui'
 import { ButtonTooltip } from 'components/ui/ButtonTooltip'
 import { X } from 'lucide-react'
 import { toast } from 'sonner'
+import { Alert_Shadcn_, AlertDescription_Shadcn_, AlertTitle_Shadcn_, Badge, Toggle } from 'ui'
 
 import { useConsentState } from 'common'
+import { LOCAL_STORAGE_KEYS } from 'common/constants/local-storage'
 import Panel from 'components/ui/Panel'
 import { useSendResetMutation } from 'data/telemetry/send-reset-mutation'
 import { useLocalStorageQuery } from 'hooks/misc/useLocalStorage'
-import { LOCAL_STORAGE_KEYS } from 'common/constants/local-storage'
 
 export const TermsUpdateBanner = () => {
-  const [termsUpdateAcknowledged, setTermsUpdateAcknowledged] = useLocalStorageQuery(
+  const [termsUpdateAcknowledged, setTermsUpdateAcknowledged, { isSuccess }] = useLocalStorageQuery(
     LOCAL_STORAGE_KEYS.TERMS_OF_SERVICE_ACKNOWLEDGED,
     false
   )
 
-  if (termsUpdateAcknowledged) return null
+  if (!isSuccess || termsUpdateAcknowledged) return null
 
   return (
     <Alert_Shadcn_ className="mb-4 relative">
--- a/apps/studio/components/interfaces/Integrations/CronJobs/CreateCronJobSheet.tsx
+++ b/apps/studio/components/interfaces/Integrations/CronJobs/CreateCronJobSheet.tsx
@@ -237,6 +237,8 @@ export const CreateCronJobSheet = ({
   })
 
   const isEdited = form.formState.isDirty
+  // if the form hasn't been touched and the user clicked esc or the backdrop, close the sheet
+  if (!isEdited && isClosing) onClose()
 
   const onClosePanel = () => {
     if (isEdited) {
@@ -270,11 +272,6 @@ export const CreateCronJobSheet = ({
       'values.functionName',
     ],
   })
-
-  useEffect(() => {
-    // if the form hasn't been touched and the user clicked esc or the backdrop, close the sheet
-    if (isClosing && !isEdited) onClose()
-  }, [isClosing, isEdited, onClose])
 
   // update the snippet field when the user changes the any values in the form
   useEffect(() => {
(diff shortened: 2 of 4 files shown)


apps/studio/components/interfaces/Account/Preferences/AnalyticsSettings.tsx
joshenlim, line 1: This is unrelated, but adding the `isSuccess` check to prevent layout shift on page render if the user already acknowledged

apps/studio/components/interfaces/Integrations/CronJobs/CreateCronJobSheet.tsx
```diff
@@ -219,23 +226,22 @@ export const CreateCronJobSheet = ({
 
   const cronJobValues = parseCronJobCommand(selectedCronJob?.command || '', project?.ref!)
 
+  // [Joshen] Keeping the var name here generic, in case there's other cases that might fail parsing
+  const unableToParseCronJob =
```
ivasilov: We should do this logic in `parseCronJobCommand` and add it to the tests. You can just do on line 116
```
if (endpoint !== '') {
  return {
        type: 'http_request',
...
}
```
and it will fall through the sql snippet case. Please add a test for this, it's really hard to test all cases.

joshenlim: great catch! simplifies things a lot haha

apps/studio/components/interfaces/Integrations/CronJobs/CronJobs.utils.ts
```diff
@@ -112,18 +112,20 @@ export const parseCronJobCommand = (originalCommand: string, projectRef: string)
...
+        httpHeaders: headersObjs,
+        httpBody: body,
+        timeoutMs: Number(timeout ?? 1000),
+        snippet: originalCommand,
+      }
     }
   }
 
-  const regexDBFunction = /select\s+[a-zA-Z-_]*\.?[a-zA-Z-_]*\s*\(.+/g
+  const regexDBFunction = /select\s+[a-zA-Z-_]*\.?[a-zA-Z-_]*\s*\(\)/g
```
joshenlim, line 128: Am updating the regex function here so only DB functions that don't take in any parameters will be considered as sql_function type

opting for this as our UI for DB functions only allows selecting the schema and function, no support for parameters here, so even if there were params the UI wouldn't be able to reflect them here which is inconsistent (noticed this as if we fail to parse a `select net.http_xxx` command, it'll currently fallback to here)
image

apps/studio/components/interfaces/Integrations/CronJobs/CronJobs.utils.test.ts
```diff
@@ -30,12 +30,10 @@ describe('parseCronJobCommand', () => {
     })
   })
 
-  it('should return a sql function command when the command is SELECT public.test_fn(1, 2)', () => {
+  it('should return a sql snippet command when the command is SELECT public.test_fn(1, 2)', () => {
```
joshenlim, line 43: related to the change [here](https://github.com/supabase/supabase/pull/37562/files#r2243175364)
so DB functions with parameters will be taken as SQL snippet type instead

Hallidayo:
Hi @joshenlim - Does this fix [this issue](https://github.com/supabase/supabase/issues/37528)?

apps/studio/components/interfaces/Integrations/CronJobs/CreateCronJobSheet.tsx
```diff
@@ -231,11 +238,6 @@ export const CreateCronJobSheet = ({
 
   const isEdited = form.formState.isDirty
 
-  // if the form hasn't been touched and the user clicked esc or the backdrop, close the sheet
-  if (!isEdited && isClosing) {
```
ivasilov, line 235: @joshenlim this change causes a bad render of the "are you sure you want to close?". To replicate, try opening the sheet and clicking on the backdrop. 

Calling functions in render is ok if they don't cause a render loop https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

ivasilov:
> Hi @joshenlim - Does this fix [this issue](https://github.com/supabase/supabase/issues/37528)?

I've pushed a fix and unit test for this issue.

ivasilov (approved):

joshenlim:
just taking a look at this [one](https://github.com/supabase/supabase/pull/37562#discussion_r2243856004) before getting this PR through 🙏

apps/studio/components/interfaces/Integrations/CronJobs/CreateCronJobSheet.tsx
joshenlim: gotcha on this! will revert 🙂

PR #37572, opened 2025-07-30
Add troubleshooting section for Docker password issues

gitarun009:

YES

Documentation update

No guidance currently exists for handling issues that arise when environment variables (like POSTGRES_PASSWORD) are changed. This may confuse users when Docker containers fail to restart properly due to persistent volumes or stale configurations.

## What is the new behavior?

Adds a new troubleshooting section in the docs that explains:
	•	Why Docker may fail after .env password changes.
	•	How to reset the environment cleanly using docker compose down -v and up --build.
	•	A warning about volume deletion and the need to back up data.

## Additional context

This update is meant to help users recover from common setup issues, especially after modifying sensitive values like database passwords in .env files.

### DIFF
--- a/apps/docs/content/guides/self-hosting/docker.mdx
+++ b/apps/docs/content/guides/self-hosting/docker.mdx
@@ -469,6 +469,24 @@ docker rm supabase-analytics
 
 ---
 
+### Troubleshooting: Docker fails after password changes
+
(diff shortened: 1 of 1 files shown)


aantti:
This has been addressed a while ago a bit differently. Refer to https://supabase.com/docs/guides/self-hosting/docker#changing-database-password

PR #37579, opened 2025-07-31
Chore/new sql

SaxonF:
The goal of this PR is to simplify the initial empty state of the SQL editor.

image

- Removes new tab state for sql editor. 
- Moves inline assistant instructions into placeholder. 
- Moves templates and quickstarts under a "community" section. The idea here is similar to Figma's community. Initially all written by Supabase but in future could open it up to community to share their own useful snippets. We've explored this idea with "cookbook" but that encompasses snippets, edge functions etc.


apps/studio/components/layouts/SQLEditorLayout/SqlEditorMenuStaticLinks.tsx
```diff
@@ -1,34 +1,58 @@
...
+  InnerSideMenuSeparator,
+} from 'ui-patterns'
 
 const OPTIONS = ['templates', 'quickstarts'] as const
 
 export function SqlEditorMenuStaticLinks() {
   const { ref } = useParams()
   const router = useRouter()
 
+  const [communitySectionOpen, setCommunitySectionOpen] = useLocalStorage<boolean>(
```
joshenlim: hmm ideally we consolidate this in SQLEditorNav `sectionVisibility`

joshenlim:
There's currently a flash of the old empty state when landing on the SQL Editor, wondering if we might be able to mitigate that [Edit: think i might know how to - will push up]

https://github.com/user-attachments/assets/db4c7154-5f30-456e-98c6-8b8326a6c6c8

apps/studio/components/layouts/SQLEditorLayout/SqlEditorMenuStaticLinks.tsx
SaxonF: ah sorry, I recon we could just useState for static links

joshenlim: also wondering any thoughts of shifting community section up top above shared?

joshenlim (approved):

PR #37585, opened 2025-07-31
fix: Use the new `envVars` property in the `shadcn` blocks

ivasilov:
This PR adds the new `envVars` property in `shadcn` blocks to add Supabse env vars to the `.local.env` file in the project.


saltcod (approved):
Looks great!

PR #37589, opened 2025-07-31
Invalidate project upgrade eligibility query when disabling database extension

joshenlim:
Fixes FE-1714

## Context:

Related to this particular warning in the project settings infrastructure page - this warning doesn't go away after disabling the extensions listed here as the RQ for the upgrade eligibility isn't getting invalidated, so this PR fixes that.

Also fixes small style issue in the extension card where the buttons should wrap.

Also added a small QoL improvement to support clicking on the extension name from the upgrade eligibility warning, which will direct users to the database extensions page with the search input filled with the extension name

Before:
image

After:
image

### DIFF
--- a/apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
+++ b/apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
@@ -1,3 +1,4 @@
+import { useParams } from 'common'
 import { Alert_Shadcn_, AlertDescription_Shadcn_, AlertTitle_Shadcn_, Button } from 'ui'
 
 export const ReadReplicasWarning = ({ latestPgVersion }: { latestPgVersion: string }) => {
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
```diff
@@ -67,12 +70,17 @@ export const UnsupportedExtensionsWarning = ({
...
       <AlertDescription_Shadcn_ className="flex flex-col gap-3">
         <div>
-          <p className="mb-1">You'll need to remove the following extensions before upgrading:</p>
+          <p className="mb-1">The following extensions have to be removed before upgrading:</p>
 
           <ul className="pl-4">
             {unsupportedExtensions.map((obj: string) => (
               <li className="list-disc" key={obj}>
-                {obj}
+                <a
```
ivasilov: Why are you using `<a>` here? Were you planning to add `target="_blank"`?

joshenlim: oh 🤦  should be using Link here sorry!

joshenlim: just pushed! 🙏

ivasilov (approved):

PR #37599, opened 2025-07-31
add syncing between charts

jordienr:
https://github.com/user-attachments/assets/fc2b7b69-663b-45af-955a-8cf2426d9047


joshenlim:
Unrelated feedback - wondering if might be worth making the width of the y axis labels consistent across charts? Just so the tooltip lines up nicely. But just a nit though cause the syncing of the tooltips make it easy to track the values across charts already, feel free to ignore haha 🙏 
image

joshenlim (approved):
Verified changes on staging preview with database reports and looks great! 😄 

Reckon this might also be useful with the home page charts 🙂 
image

PR #37600, opened 2025-07-31
Fix Webhook Form not rendering persisted HTTP Headers upon refresh

Saeris:
In #36759 a new `useEffect` was added to `FormContents.tsx` which introduced a race condition with another `useEffect` in `EditHookPanel.tsx`, whereby each was trying to call `setHttpHeaders`. This was resulting in some easy to miss behavior where the HTTP Headers list was not rendering as expected.

Currently, if you visit https://supabase.com/dashboard/project/_/integrations/webhooks/webhooks and try to edit an existing webhook, you'll see that there are no headers rendering in the form, even if those were specified upon creation of the webhook. Also, if you create a new webhook and then immediately edit it, you *will* see whichever headers you specified upon creation, *however* if you refresh the page they will not be present.

This PR implements a fix for this unwanted behavior by moving the header parsing logic down into `FormContents` so that `setHttpHeaders` is only called once by a `useEffect` in a given render.


saltcod (commented):
Tested on staging and entries now persist!

screenshot-2025-07-31-at-16 51 15

joshenlim (commented):
Found a bug here:

The http headers state doesn't persist while you're in the panel and switching between types (unlike http parameters). Any added http headers during the session gets cleared (reset) when you switch the type

FWIW - i reckon consolidating the setHttpHeaders logic into EditHookPanel might be better actually, since thats the parent component , it'll be clearer where the state changes are happening from

https://github.com/user-attachments/assets/279fd683-afb2-485d-aabc-ce72f6bd0b2b

Saeris:
hmm, alright will need to investigate this more then

> FWIW - i reckon consolidating the setHttpHeaders logic into EditHookPanel might be better actually, since thats the parent component , it'll be clearer where the state changes are happening from

I'm not sure that will be possible since the logic that was already in FormContext was dependent on the form `values` object, which can't be hoisted up and out unless I did a more comprehensive refactor like migrating this to `react-hook-form`.

Ideally we wouldn't be using `useEffect` at all, but for now I think the desire here is to keep the changes narrowed to just solving this bug

Saeris:
Alright I've opted to try for a different strategy here by moving the header and parameter initialization logic out of `useEffect` entirely by leveraging the `key` prop to force `useState` to re-initialize whenever `selectedHook` changes

this should be the desired behavior anyways because we want this form to be completely reset whenever we open it to create a new hook or select an existing hook to edit. `useState`'s value should never carry over when this changes. in this scenario, state would only be preserved when making an edit to a hook, closing the form, and editing it again

making this change also brought to my attention that not only were headers for an existing hook not being loaded upon editing, they also weren't being initialized to `content-type application/json` when creating a new hook

please give this a look over in staging for anything I might have missed 🙏

apps/studio/components/interfaces/Integrations/Webhooks/ListTab.tsx
```diff
@@ -38,6 +38,7 @@ export const WebhooksListTab = () => {
     <div className="p-10">
       <HooksList createHook={createHook} editHook={editHook} deleteHook={deleteHook} />
       <EditHookPanel
+        key={selectedHook?.id}
```
Saeris, line 41: whenever this changes, the internal state of `EditHookPanel` will be re-initialized

saltcod (approved):
Tested and looks good on staging!

krisbaum74:
this is still an issue. webhook heades are still not saved

PR #37601, opened 2025-07-31
Chore/convert cron jobs to data grid

joshenlim:
Fixes FE-1633

## Context

Our current cron jobs UI currently retrieves _every_ cron job within the project, which causes issues if the project has many many cron jobs (e.g 10k). The fix would be to introduce pagination to the UI but the current UI doesn't really allow for that in a nice UX. Hence opting to switch over to DataGrid for displaying data of the cron jobs, which will also be consistent to how we display data for Auth -> Users.

This is also technically a pattern that we've been looking into adopting RE using DataGrid to manage data, so this is also taking a step in that direction.

## Before
image

## After
image

## Changes involved
- Use DataGrid for integrations -> Cron Jobs, similar to Auth Users
  - Infinite scrolling + server side filtering
  - Clicking the row now goes to the jobs run page (Previously was clicking the "History" button)
- Update `database-cron-jobs-query.ts` to also fetch latest run information
  - We're previously calling `getDatabaseCronJobRun` per Cron Job card just to retrieve the latest run for each job
  - So this will mitigate `n` number of requests to pg-meta
- Deprecate `CronJobCard.tsx` and `database-cron-jobs-run-query.ts` (no longer used)
- Shifted over the nuance details for cron too (let me know if i might have missed any)
  - Shifted `sendEvent` calls for clicking edit, delete, and history of cron job accordingly from `CronJobCard.tsx`
  - "Edit job" CTA in context menu within the grid is disabled if job has no name

## To test
- [ ] Verify that the infinite scrolling works
- [ ] Verify that you can still create/edit/delete cron jobs
- [ ] Verify if any of the UX feels awkward
  - I won't go too much into detail on what to look out for, but as long as anything feels a tiny bit off, just let me know!


jordienr (approved):
tested in preview and all changes lgtm 

1 nitpick in slack but non blocking

PR #37604, opened 2025-07-31
fix: Make the padlock red if the entity is insecure in the Table Editor

ivasilov:


charislam (approved):

joshenlim:
JFYI updating the UI based on the feedback in the thread
- Use a big red badge instead of a lock icon
- Slightly more verbose tooltip message
image

joshenlim (approved):

PR #37613, opened 2025-08-01
basic typography classes

SaxonF:
Adds some basic typography classes we should start utilising to keep typography consistent across the dashboard, including links.


apps/studio/styles/typography.scss
```diff
@@ -0,0 +1,78 @@
...
+  .text-subTitle {
+    @apply text-lg;
+  }
+
+  .text-compact {
+    @apply text-xs;
+  }
+
+  /* Link */
+  .text-link {
```
joshenlim, line 75: wanna add `<a>` as well in @layer base?
if we do - we'd just need to clean up a bit in `main.scss` too ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/styles/main.scss#L184))

joshenlim:
Just pushing some minor clean ups on the text size classes

e.g for where we use `<h1>`, if there's a className assigned like `<h1 className="text-2xl" ... />`, i'm just removing `text-2xl` since its already declared in the new `typography.scss`

am leaving custom text sizes like e.g `<h1 className="text-3xl" ... />` as they are

joshenlim (approved):
Generally LGTM! Verified on local

staaldraad (approved):

apps/studio/styles/typography.scss
SaxonF: this one is a more risky change because we have so many variations, including block variations wrapping entire divs . I think we leave this for another PR.

PR #37573, opened 2025-07-30
fix: Move confirm email setting from email provider to basic auth settings

ivasilov:
- Remove MAILER_AUTOCONFIRM configuration from email provider form validation
- Add confirm email setting to BasicAuthSettingsForm with proper form validation and UI components
- Improve user experience by grouping the email confirmation setting with other basic authentication settings

PR #37615, opened 2025-08-01
fix: The mail confirm setting should be negated

ivasilov:
This PR fixes an issue caused by https://github.com/supabase/supabase/pull/37573 where the `MAILER_AUTOCONFIRM` setting should be negated because the UI shows an inverted state.


joshenlim (approved):
Verified on staging preview that the payload getting sent for mailer autoconfirm is flipped from the UI

Issue #37616, opened 2025-08-01
Cron edit no longer works


silentworks:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When trying to edit an existing cron job an error is being thrown stating `A cron job with this name already exists`, this shouldn't hapen.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to [https://supabase.com/dashboard/project/_/integrations/cron/jobs](https://supabase.com/dashboard/project/_/integrations/cron/jobs)
2. Right click on an existing cron job and edit it
3. Click **Save cron job**

## Expected behavior

There shouldn't be any error thrown about the cron job name as this is not editable.

## Screenshots

https://github.com/user-attachments/assets/28bd6726-0f1a-4f48-8f76-dc09982d5408

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

spormeon:
+1  seen this, stopping editing of cron jobs.  Oh and bring back the 3 dot editing button, its not intuative how you can edit anymore

silentworks:
This is kind of related to the issue above. When this issue was reported by a @spormeon on Discord it took me a while to figure out how to edit a cron as I went to the docs and found this guide page https://supabase.com/docs/guides/cron/quickstart?queryGroups=database-method&database-method=sql-schedule-job#edit-a-job and there is no mention of right clicking to see the edit menu. It was through trial and error that I found the edit menu. I think this should be made more prominent in the UI and the 3 dots editing button should return.

joshenlim:
Thanks so much everyone for the report! Fix should be live now 🙂 

Also got something to address the UX issue with the 3 dots menu 🙏 
https://github.com/supabase/supabase/pull/37618

PR #37617, opened 2025-08-01
Chore/fix cron job create

joshenlim:
## Context

Fixes an issue whereby the editing a cron job is blocked as we're checking for an existing cron job. This check should only be done for when creating a new cron job

Addresses #37616


alaister (approved):
LGTM!

- Tested on preview

ivasilov (approved):

PR #37618, opened 2025-08-01
Add more menu button cron

joshenlim:
## Context

Couple of feedback which I agree that with the new cron UI it's not intuitive how you'd edit or delete a cron job since those actions are hidden behind a context menu right click.

Adding back the more menu button to the UI to make it clearer
image


alaister (approved):
LGTM!

- Tested on preview

PR #37621, opened 2025-08-01
Add a CTA to run the command of the cron job

joshenlim:
Fixes FE-1733

Adds an option to run the command from a cron job manually - helpful for testing if the command of the cron job is working
image

If command is alright, shows a success toast
image

Otherwise will show the SQL error
image

JFYI ideally we'd be able to use something from pg_cron that will run the job manually and also reset the job's timer, but that isn't supported atm https://github.com/citusdata/pg_cron/issues/226


alaister (approved):
LGTM!

- Tested on preview

PR #37626, opened 2025-08-01
Chore/cron fixes

joshenlim:
## Context

Fixes 2 identified issues
- Cron job infinite query was returning a duplicate row for a job that previously succeeded, but latest run failed (if im not wrong). Bug was in the SQL query for fetching the cron jobs, so got that fixed in this PR (Although just to call out that I needed help from AI for this one, so if there's a way to optimize the query please let me know)
- Clicking on the child tab from the integrations page (in this case, e.g. "Test Ping") loses the label in the tabs, because its reading from the `child-label` param in the URL. This PR fixes that as well
  image


ivasilov (approved):

ivasilov (approved):

saltcod (approved):

saltcod:
Looks good on staging!

PR #37632, opened 2025-08-01
Allow protected schema names in table editor actions

saltcod:
A schema name with a reserved keyword causes issues with the table editor: 

To reproduce: 

1. Run this 
-- Create the schema
CREATE SCHEMA grouping;

-- Create a table in the schema
CREATE TABLE grouping.example_table (
  id SERIAL PRIMARY KEY,
  name TEXT NOT NULL,
  created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
  is_active BOOLEAN DEFAULT TRUE
);

2. Try and insert a row 
3. See error. 

This is a lot of code though. 
Wondering if it can be handled it pgmeta itself instead?


soedirgo:
@saltcod this should fix it: https://github.com/supabase/supabase/pull/37693

PR #37635, opened 2025-08-01
fix: Add handling for single quotes in option values when creating FDW.

ivasilov:
There's a bug in fdw creation API call where it doesn't handle options with `'` in them. I've added code to escape them. 

Because the option value is going through `EXECUTE FORMAT` it needed to be escaped twice (one `'` turns into `''''`).

To test this, try creating a wrapper with a `'` in one of its options.

Screenshot 2025-08-01 at 23 50 49


tylershibata:
LGTM!

tylershibata:
Thank you for digging into this Ivan!

tylershibata (approved):

joshenlim (approved):

PR #37653, opened 2025-08-04
Minor clean ups

joshenlim:
## Context

Trying to address a scrollbar issue that was reported internally (specifically on Firefox browser) - while I wasn't able to reproduce on my end, am removing an excess `overflow-y-auto` that I think might help (will still be a clean up either way if it didn't work)

## Changes involved
- Remove `overflow-y-auto` from `ProductMenu.tsx`
  - This component is typically rendered within `ProductMenuBar.tsx` and there's already an `overflow-y-auto` in `ProductMenuBar.tsx` that should cover the scrolling behaviour
- Pass in an `overflow-y-hidden` from the Table Editor pages into `EditorBaseLayout.tsx`
  - We don't need the product menu to be scrollable in the Table Editor since the list of tables is already scrollable
- Clean up props in `EditorBaseLayout.tsx`
  - The prop `hideTabs` wasn't being used since its being declared inside the component
  - Opting to declare the props getting passed into `ProjectLayoutWithAuth` instead of spreading props for readability
- Remove duplicate import declaration of `global.css` in `main.scss`
- Remove Column Privileges shifting notice from AuthLayout (Been around for 16 months already, reckon we can clean this up)

## To test
- [ ] Verify that the product menu can still be scrolled as expected across all the project pages in the dashboard


apps/studio/components/interfaces/Settings/API/ApiKeysMoved.tsx
joshenlim, line 1: Just cleaning up unused imports, and letting prettier sort them

alaister (approved):
LGTM!

- Tested on preview

PR #37654, opened 2025-08-04
feat: sticky storage config

dnywh:

Makes the configuration section for Storage (Policies, Settings) easier to find for those people with lots of buckets.

People with ~20 or so buckets on an average laptop/desktop display would not see the _Configuration_ section on their sidebar, since the list of buckets would push it down below-the-fold.

## What is the new behavior?

The _Configuration_ section is sticky, anchored to the bottom, so that **only** in these circumstances (overflowing content), the section becomes sticky.

https://github.com/user-attachments/assets/19b0b797-d5e6-4f9d-9865-15d32aa2c609

## Additional context

The parent `MobileSheetNav` component had some `py` hardocded in causing visual bleed on smaller screen sizes. I’ve made this conditional via an optional `stickySidebarBottom` prop (default `false`) so that we can handle this (and other) sticky cases.


apps/studio/components/interfaces/Storage/StorageMenu.tsx
```diff
@@ -180,9 +180,7 @@ const StorageMenu = () => {
             )}
           </div>
 
-          <div className="h-px w-full bg-border" />
-
-          <div className="mx-3">
+          <div className="w-full px-3 py-6 sticky bottom-0 bg-background border-t border-border">
```
joshenlim: probably dont need bg-background? otherwise there's a color diff here

image

apps/studio/components/layouts/StorageLayout/StorageLayout.tsx
```diff
@@ -7,11 +7,17 @@ import ProjectLayout from '../ProjectLayout/ProjectLayout'
 export interface StorageLayoutProps {
   title: string
   children: ReactNode
+  stickySidebarBottom?: boolean
```
joshenlim: seems unnecessary since we're passing a fixed value to ProjectLayout

joshenlim (approved):
Verified changes locally! Looks great 🙂

apps/studio/components/interfaces/Storage/StorageMenu.tsx
dnywh: I believe we do, otherwise the transparent background means the _Buckets_ section is exposed on scroll, like this:

see-through

Right?

Good catch on the the colour diff! I just changed it to `bg-dash-sidebar`. See https://github.com/supabase/supabase/pull/37654/commits/7680ad6997c05b79fd0c195113ac20b709f4ec22. Thanks!

apps/studio/components/layouts/StorageLayout/StorageLayout.tsx
dnywh: Duh, got carried away creating types. Thanks

apps/studio/components/interfaces/Storage/StorageMenu.tsx
joshenlim: nice yeap looks good!

PR #37659, opened 2025-08-04
Chore/address danny friction log

joshenlim:
## Context

Addressing a number of issues that @dnywh brought up in his friction log (thank you so much for that too! 🙏 )

## Changes involved

- Standardize AI perms button CTA copy
  image

- Standardize "More" icon button for Table Editor + Storage Explorer
  image
  image

- Improve toast error copy for invalid folder name (just to be more explicit about what special characters are allowed, since its only hyphen and underscore)
  image

- Storage explorer, open the new folder when it's being created

https://github.com/user-attachments/assets/c18fa790-4f93-49de-9354-9b29d1bda5c0

- Highlight file explorer row item when there's issues with the name for folders (make it more clearer that something went wrong)

https://github.com/user-attachments/assets/b3ff9f21-c6f9-4233-9ed5-16a783e45dc0

- File explorer row item, remove folder row if there's an error while trying to save on blur (specifically for this event only), so that you don't get stuck in an error state while trying to do other things

https://github.com/user-attachments/assets/0117213a-ce37-4fc4-84d8-82878e4f81e4

- Update Table Editor header action "Auth policy" -> "RLS policy"
image

- Add dialog if navigating away from custom reports (or closing tab) while there's unsaved changes

https://github.com/user-attachments/assets/09c16e7a-2ee2-49c4-99c1-2d964e79c393

- Add Cmd + S shortcut to save snippet in SQL Editor (which also silences the native browser save action)

- Add toggling of favourite to SQL Editor context menu
image


apps/studio/components/interfaces/SQLEditor/MonacoEditor.tsx
```diff
@@ -88,6 +88,17 @@ const MonacoEditor = ({
       },
     })
 
+    editor.addAction({
+      id: 'run-query',
+      label: 'Run Query',
+      keybindings: [monaco.KeyMod.CtrlCmd + monaco.KeyCode.KeyS],
```
alaister, line 94: Does this work for you? I didn't see anything happen...

joshenlim: ahh yup - it just triggers the save asynchronously by adding it to the queue. The UI indication is quite subtle though as its just the check icon in the SQL editor utility bar. I should fix the inaccurate id and label here though
image

alaister (approved):
LGTM!

- Tested on preview

PR #37662, opened 2025-08-04
feat(replication): Add new UI for new table state handling

iambriccardo:
This PR adds a new UI for handling the `error` state of a table within the pipeline replication status UI. An error is displayed differently based on the contents and retry policy. In addition, it removes the `copy` action in the table, which was not needed.

In addition, it removes the `failed` status parameters since now it's just a state like the others.

### No Retry
image

### Manual Retry
image

### Timed Retry
image

### How to test

The internal project which has ETL enabled is a good starting point to test this PR. For the sake of simplicity the database which contains the state could be directly modified to show the errors. Feel free to ask me on how to do this in detail.


apps/studio/components/interfaces/Database/Replication/ErroredTableDetails.tsx
```diff
@@ -0,0 +1,117 @@
...
+  state: Extract<TableState['state'], { name: 'error' }>
+  tableName: string
+  tableId: number
+  projectRef: string
+  pipelineId: number
+}
+
+const ContactSupportButton = ({ size = 'tiny' }: { size?: 'tiny' | 'small' | 'medium' }) => (
+  <Button asChild type="default" size={size} icon={<ExternalLink className="w-3 h-3" />}>
+    <Link href="https://supabase.com/support" target="_blank" rel="noopener noreferrer">
```
jordienr: Better to add some params to the support link so the form is auto populated as much as possible. Search other implementations of the contact support link for reference.

apps/studio/components/interfaces/Database/Replication/ReplicationPipelineStatus.tsx
```diff
@@ -198,25 +191,9 @@ export const ReplicationPipelineStatus = () => {
               <h4 className="font-medium text-destructive-900 mb-1">
                 {errorTables.length} table{errorTables.length > 1 ? 's' : ''} failed
               </h4>
-              <p className="text-sm text-destructive-700 mb-3">
-                Some tables encountered replication errors. Check the logs for detailed error
-                information.
+              <p className="text-sm text-destructive-700">
```
jordienr: Can swap this section for a destructive Alert to reduce code :) 

https://supabase-design-system.vercel.app/design-system/docs/components/alert

jordienr (changes requested):
Think this would look better within a `<pre>` tag with font-mono or some styling that lets the user know this is a system error.

Image

I see you've removed the link to logs, think that was useful for self debugging. Is there really nothing we can do here other than direct the user to contact support 🤔?

joshenlim:
Updating the different states, mainly to
- Reduce the colors that we're using (generally avoid colors for text, and in this case the status badge stands out enough too)
- Update the copy for the error messages to be more streamlined and clear

- Timed retry:
  image

- No retry:
  - Shifted support CTA to an inline link instead of a button
  image

- Manual retry:
  - Removed contact support CTA here - can leave this one out if user can self-remedy 
  image

  - Also adjusted the dropdown here
    image

joshenlim (approved):

jordienr (approved):

PR #37674, opened 2025-08-05
fix visualiser size

SaxonF:
image

image

Fixes an issue with font size in visualiser.


joshenlim (approved):

PR #37682, opened 2025-08-05
Pool size default

SaxonF:
image

Rather than showing default value as placeholder, we show it as an actual value like other places.


jordienr (approved):
lgtm!

PR #37683, opened 2025-08-05
Fix Table Editor not identifying truncated array column values properly

joshenlim:
Fixes FE-1731

## Context

Bug came from [here](https://github.com/supabase/supabase/pull/34546/files) RE the truncation logic for really large cell values, where for an array type column we're adding a `{"truncated":true}` object at the end of the array if the value needs to be truncated so to maintain the array structure

But `isValueTruncated` from `RowEditor.utils.ts` wasn't updated so the Table Editor wasn't able to identify such columns if the whole value needs to be fetched

## To test

There's a repro step in the attached linear task, but for convenience:
- Create a test table with a jsonb array column
```
create table public.jsonb_test (
  id bigint generated by default as identity not null,
  created_at timestamp with time zone not null default now(),
  test jsonb[] null
)
```

- Add row with a really long JSON array value into the column (grab the value from the linear ticket)
- Verify that the Table Editor shows this warning to fetch the whole cell value
  image


jordienr (approved):
pushed some tests. looks good on preview!

joshenlim:
thank you so much @jordienr you legend 💪 🙏 ❤️

PR #37684, opened 2025-08-05
Fe 1799/consolidate to useselectedprojectquery and

joshenlim:
Fixes FE-1799

## Context

This in efforts to consolidate how we retrieve projects and organizations in the dashboard, specifically to just use `useSelectedProjectQuery` and `useSelectedOrganizationQuery` hooks.

- `useProjectContext` and `useSelectedProject` are identical in what they return with `useSelectedOrganizationQuery`
- `useSelectedOrganization` and `useSelectedOrganizationQuery` are identical in what they return
- `useProjectByRef` and `useProjectByRefQuery` are identical in what they return
- Only difference is just that the new query hooks return the shape of `{ data }`, rather than directly `data`

Based on these ^ the replacement is pretty straight forward and there shouldn't be any issues as long as the typescript build check is passing, hence why doing everything in one PR. Manually went through everything as well just to be sure of the changes 🙏 

## Changes involved
- Replace all usage of `useProjectContext` with `useSelectedProjectQuery`
- Replace all usage of `useSelectedProject` with `useSelectedProjectQuery`
- Replace all usage of `useProjectByRef` with `useProjectByRefQuery`
- Deprecate `useSelectedProject`, `useSelectedOrganization`,  and `useProjectByRef`
- Deprecate `ProjectContext`
- Removal of some unused imports and variables, sort some import statements

## To test
- Just needs a rough smoke test on the staging preview


kimsoo527 (approved):

kimsoo527 (dismissed):

staaldraad (approved):
Staging worked for me, no errors in the console.

jordienr (approved):
lookin good on preview

PR #37693, opened 2025-08-05
fix(pg-meta): update reserved words

soedirgo:
Fixes the issue https://github.com/supabase/supabase/pull/37632 is meant to address. List of reserved words in pg-format just needs updating.

As per the [docs](https://www.postgresql.org/docs/current/sql-keywords-appendix.html), some keywords are non-reserved but cannot be used as a function/type name without quoting. `GROUPING` belongs to this group (pun not intended)


saltcod (approved):
Checked this on staging and works great!

-- Create the schema
CREATE SCHEMA grouping;

-- Create a table in the schema
CREATE TABLE grouping.example_table (
id SERIAL PRIMARY KEY,
name TEXT NOT NULL,
created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
is_active BOOLEAN DEFAULT TRUE
);

PR #37695, opened 2025-08-05
feat(docs): code samples with ts type stripping

charislam:
Introduce a new option to `$CodeSample`, `convertToJs`, which takes a code sample written in TypeScript and strips the types to produce a JavaScript version.

Adds tests for type stripping.


ChrisChinchilla (approved):
I made one small change to the instructions as while I think it all works fine, i wasn't completely sure what I was supposed to do once I had added `convertToJs`, I felt like the other code sample would just appear, when actually you need a pair of them.

PR #37696, opened 2025-08-05
feat: use temporary api keys for storage explorer

hf:
Instead of authorizing with the `service_role` key (discouraged) or a secret API key (might not exist, might be customized, etc.) use a special platform temporary API key that performs the same role.


joshenlim:
Pushed up the following changes
- Fix merge conflict with API types
- Remove service key entirely from storage explorer store
- Small UI fix on storage explorer padding
- Small UI fix to set duration of Uploading toast to Infinity

Tested uploading a 2.7GB file - all looks good! 🙂

joshenlim (approved):
Verified changes on staging preview - able to upload files via storage explorer

cemalkilic (approved):

PR #37699, opened 2025-08-05
feat: add sign-in-partner page

hf:
Adds a new page `/sign-in-partner#partner=<name-of-partner>&id_token=<ID token>` which signs the user in using `signInWithIdToken()`.

[Joshen] Opting for this UI for this page
image


joshenlim:
Should be done with the UI tweaks now - just need to test once more the E2E flow though nothing functional changed @hf 

although if for some reason we're unable to test on the staging preview, happy to get this through too

joshenlim (approved):
LGTM code wise - Stojan's tested on the staging preview which looks alright except for a 500 thats happening on the /profile endpoint, just a thing to note and investigate separately if needed

PR #37705, opened 2025-08-06
feat: update the main platform SSO documentation

cstockton:
Update the main platform SSO documentation in preparation for the Self-Service SSO release (https://github.com/supabase/supabase/pull/36732).


apps/docs/content/guides/platform/sso/cs-test.mdx
```diff
@@ -0,0 +1,9 @@
+---
```
charislam: Is this meant to be here? I don't think we have a sso_platform_enable partial right now

charislam (approved):

apps/docs/content/guides/platform/sso/cs-test.mdx
cstockton: Whoops didn't mean to commit that, it's been removed.

PR #37707, opened 2025-08-06
Decouple editor panel

SaxonF:
The current inline editor implementation relies on global state, much like the assistant panel, mostly because it was meant to be seen as a layer on top of the dashboard that can be used to run queries that manage entities within the dashboard without leaving the current page (e.g. policies, functions, triggers). This sounds great in theory but you quickly run into state management issues when needing to switch between entity types, supply varying props like templates, and reset values. It can get messy. 

This PR decouples the component from state. State now only manages whether the global editor panel (triggered via top right button) is visible. Pages like policies can make use of the EditorPanel by adding locally and passing in the right props.


joshenlim (approved):

PR #37708, opened 2025-08-06
discourage email comparison

doublethink:
We shouldn't be encouraging developers to use a mutable string for RLS
There is already a user_id to jwt.uid template example so removing this rather than modifying

Fixes SEC-480


joshenlim (approved):
Changes LGTM! Checked staging preview

PR #37721, opened 2025-08-06
fix(studio): small clean up of block classes

kemaldotearth:

YES

My linter was throwing problems because we had two Tailwind block element classes applied to the same elements. Clears that up.

I assume it was a bit of Cursor or Claude leftovers, didn't need both `block` and `flex` on the link elements, one or the other suffices. :)


alaister (approved):
LGTM!

- Tested on preview

PR #37730, opened 2025-08-06
Add CTA to unrestricted badge in table editor

joshenlim:
## Context

With the recent addition of the "Unrestricted" badge in the table editor, we've received a couple of feedbacks to make these callouts a bit more actionable, so that users can quickly address them.

## Changes involved

- Adding a "Learn more" CTA to each of the tooltips which will open the table with the warning popover open
  - There's quite a bit of information about the warning that's too lengthy to be placed in the tooltip, so reckon this might be a better UX instead
- This also shifts the open state for the warning popover to a URL state

https://github.com/user-attachments/assets/18198484-e2f1-4e6b-85a5-c668f7fd1a7d


apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -51,6 +52,12 @@ const GridHeaderActions = ({ table }: GridHeaderActionsProps) => {
   const { data: project } = useSelectedProjectQuery()
   const { data: org } = useSelectedOrganizationQuery()
 
+  const [showWarning, setShowWarning] = useQueryState(
+    'showWarning',
+    parseAsBoolean.withDefault(false)
+  )
+  console.log({ showWarning })
```
alaister: Cheeky one!
```suggestion

```

alaister (approved):
LGTM!

- Tested on preview

PR #37731, opened 2025-08-06
docs(auth): add optional ID token claim instructions for Azure login

mporracindie:

**YES**

---

**Docs update**

---

The [Azure login guide](https://supabase.com/docs/guides/auth/auth-azure) does not mention the need to explicitly add the `email` claim to the ID token in Azure's Token Configuration.

As a result, users—especially those logging in with older Hotmail/Live accounts—may encounter this error when trying to authenticate via Microsoft:
`500: Error getting user email from external provider`

There is currently no mention of this in the guide, leading to confusion and login failures.

---

## What is the new behavior?

This PR adds an **optional but highly recommended** section after the “Obtain a client ID and secret” step. It walks users through the process of adding `email` and `preferred_username` claims to the ID token in Azure's App Registration to ensure Supabase can reliably extract the user’s email from the ID token.

---

## Additional context

This is based on a real issue encountered while implementing Microsoft login with Supabase. Adding these claims immediately resolved the error. The fix was confirmed by decoding the ID token before and after applying the change.

### DIFF
--- a/apps/docs/content/guides/auth/social-login/auth-azure.mdx
+++ b/apps/docs/content/guides/auth/social-login/auth-azure.mdx
@@ -63,6 +63,30 @@ curl -X PATCH "https://api.supabase.com/v1/projects/$PROJECT_REF/config/auth" \
   }'
 ```
 
+## (Optional but Recommended) Add email and username claims to the ID token
+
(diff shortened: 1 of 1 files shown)


Hallidayo:

PR #37738, opened 2025-08-07
[EXPERIMENT] Tailwind v4 Upgrade

Saeris:
First: **this PR is not meant to be merged!** It will stay in draft status until closed.

Looking for ways to speed up local development compile times, I decided to explore what would be needed to make the upgrade from Tailwind v3.4 -> v4.

The short answer is: a lot. It's actually not too difficult to get the upgrade tool `pnpm dlx @tailwindcss/upgrade` to run against `apps/studio`. We are using some non-standard syntax such as `#{!important}` in our global CSS files in addition to some custom utility classes (and even some built-ins which for some reason register as unknown, like `py-2`) that need to be manually fixed or removed in order for this tool to process without throwing.

But after that (most of which you can see in commit history), it'll go and update the class names in most of our components, even in `ui` and `ui-patterns` automatically. This will save a ton of time migrating. The challenge is that we will still need to manually migrate the majority of our custom theme to the new CSS based format, and it's not always clear from the docs what the right way to do that is (such as colors with light/dark mode variants).

Unfortunately this is the most work I've done with Tailwind so far, so I'm more or less stumbling around in the dark trying to figure out what else needs to be done. There are some obvious visual issues with what I've managed to get done here.

However, the app *does* work and it *is* faster overall. I observed in some cases a 3-5x improvement in compile times, especially in route changes in dev mode and when running e2e tests (which this PR includes some WIP changes to our e2e tests for sake of comparison). So I do think that there is a tangible, real benefit to be had by proceeding with this upgrade. The point of this is just to share what I was able to get done in a day and get more knowledgeable eyes on it for evaluation.

Tagging @saltcod @jordienr @joshenlim @MildTomato for review (if you have are interested)


Hallidayo:

Issue #33364, opened 2025-02-05
Unable to edit editable cells by double clicking.


Micah-One:
I know it is not an issue with my application, and I know it is a bug because I once I click other cells a few times, go to other tables a few times, it relieves the bug for a short bit (then goes back to not being editable)

## Describe the bug

Unable to double click and edit cell content on supabase table view. 

## To Reproduce

1. Go to Supabase dashboard table view
2. attempt to edit a simple cell like a bool/int/etc by double clicking it
3. May be able to edit, but may not and appear like it's "Frozen". clicking around on other tables and cells seems to help relieve the bug eventually (temporarily)

## Expected behavior

I am supposed to be able to edit the cell.

## System information

- OS: macOS Sequoia 15.2 (M3 Max)
- Browser: Chrome
- Version of supabase-js: N/A
- Version of Node.js: N/A

## Additional context

I found that other people were having this issue more than 2 years ago, it was claimed to be resolved, but I guess it is back now.

kerowav:
Having the same issue, I have to keep refreshing and switching tables until it eventually decides to work. Really frustrating to work with.

kcbz:
I have the same bug. I've found switching out of table editor and then back works for most cells, but I can't seem to edit float or int cells at all

MicahYea:
To everyone else having this issue: I found the most fool-proof way to bypass the bug is to just close out of the Supabase tab entirely and reopen it. 

I hope it's fixed soon, but this works temporarily.

Hallidayo:
Hi all, I've been trying to recreate this issue but I do not seem to be able to? I'm double clicking in the below boolean field but I'm not getting any freezing?

https://github.com/user-attachments/assets/868cabf2-1620-400e-9872-15e05190cd6b

Has anyone got an example video?

Also which browser/version is being used?

MicahYea:
@Hallidayo Google Chrome for me.

To be honest, it does seem random. about 70% of the time the fields are editable, 30% of the time they get bricked and I have to close and reopen Supabase.

The tables do use realtime, as well as indexing (but not indexing on bricked columns). I have also noticed that it sometimes triggers when a Supabase function updates a row on the table. 

I wish I were more of a help. I would recommend working on Chrome for a while on a project and you will likely see the issue come up.

Hallidayo:
Hi @MicahYea  - Thank you for the response, Is their any chance that when this happens for you could take a video of what you see?

MicahYea:
[...]

PR #37751, opened 2025-08-07
Part 1 of swapping useCheckPermissions with useAsyncCheckProjectPermissions

joshenlim:
## Context

We'll be favouring the use of `useAsyncCheckProjectPermissions` over `useCheckPermissions` as it'll allow us to access the loading state of the permissions. We should then also factor that in to prevent any premature "no permissions" UI rendering 

Opting to do this in smaller PRs since its a bit nuanced 
- We can only swap `useCheckPermissions` with `useAsyncCheckProjectPermissions` for project permissions, organization permissions will still need to use the former for now until we introduce a new `useAsyncCheckOrgPermissions`
- Also need to add loading states where missing for pages that are currently prematurely showing the no perms UI


apps/studio/components/interfaces/Auth/ProtectionAuthSettingsForm/ProtectionAuthSettingsForm.tsx
```diff
@@ -69,18 +69,25 @@ const schema = object({
...
-  const [isUpdatingProtection, setIsUpdatingProtection] = useState(false)
+  const { data: authConfig, error: authConfigError, isError } = useAuthConfigQuery({ projectRef })
+  const { mutate: updateAuthConfig, isLoading: isUpdatingConfig } = useAuthConfigUpdateMutation({
+    onError: (error) => {
+      toast.error(`Failed to update settings: ${error?.message}`)
+    },
+    onSuccess: () => {
+      toast.success('Successfully updated settings')
+    },
+  })
```
Saeris, line 80: I've noticed in the course of writing tests for the modal refactors I did, defining the status hooks here doesn't always work, rather it had to be defined on the mutate function. So just a note here that you may wat to verify these toasts are being called with this change

apps/studio/components/interfaces/JwtSecrets/jwt-secret-keys-table/create-key-dialog.tsx
```diff
@@ -24,8 +23,6 @@ import {
   Textarea,
 } from 'ui'
 
-dayjs.extend(relativeTime)
```
Saeris, line 27: is this unnecessary? just want to know so that if I see it anywhere else I can remove it

apps/studio/hooks/misc/useCheckPermissions.ts
```diff
@@ -133,6 +133,10 @@ export function useGetProjectPermissions(
   }
 }
 
+/**
+ * @deprecated If checking for project permissions, use useAsyncCheckProjectPermissions instead so that we can always
+ * check for loading states to not prematurely show "no perms" UIs. We'll also need a separate async check for org perms too
+ */
```
Saeris, line 141: Add an import hint here to recommend what to replace this with

```suggestion
  *
  * Use `import { useAsyncCheckProjectPermissions } from 'hooks/misc/useCheckPermissions'` instead
 */
```

apps/studio/components/interfaces/JwtSecrets/jwt-secret-keys-table/create-key-dialog.tsx
joshenlim: yup the dayjs extensions are all done on the _app level, so no need to define these in the individual components

joshenlim:
Addressing Terry's comments here (hahaha left in another PR https://github.com/supabase/supabase/pull/37758)

> not directly related, but here's another pesky little guy 
  screenshot-2025-08-07-at-22 01 03

This file wasn't updated as part of this PR, but I'll add that in and fix this

> Weird state on api keys page — not sure if related 
  https://share.cleanshot.com/7779s2DS

Will look into this one

> hummm still seeing the flash on the Auth Providers page

Also not updated as part of this PR, but i'll add that in and fix

saltcod (approved):
Rechecked changes and looks good!

avallete:
Related: https://github.com/supabase/supabase/issues/33364

PR #37773, opened 2025-08-07
fix: queue user identification calls until PostHog is initialized

seanoliver:
## Summary
Fixes a race condition in client-side tracking where we were attempting to `identify` users before PostHog had been fully initialized, resulting in anonymous page views being sent to PostHog.

Resolves GROWTH-462.

## Problem
The PostHog client's `identify()` method was silently failing when called before PostHog finished initializing. This affected all apps (www, docs, studio) but manifested differently:
- **WWW app**: No auth provider, so less affected
- **Docs app**: Auth exists but timing issues prevented proper identification  
- **Studio app**: Full auth but race condition between PostHog init and user identification

## Solution
- Queue the calls to `identify` that happen before PostHog is initialized
- Apply pending identification once PostHog's `loaded` callback fires
- Similar pattern to how we already handle pending groups
- Also ensures groups from pageview properties are properly set

## Testing
To test this in Vercel Preview, log into `studio-staging` using your Supabase email and then copy your auth tokens using the console:

```js
const authData = {
  token: localStorage.getItem('supabase.dashboard.auth.token'),
  user: localStorage.getItem('supabase.dashboard.auth.token-user')
};
copy(JSON.stringify(authData))
```

Then you can go to either `docs` or `www` preview links (below) and manually add your auth tokens:

```js
const authData = /* paste here */;
localStorage.setItem('supabase.dashboard.auth.token', authData.token);
localStorage.setItem('supabase.dashboard.auth.token-user', authData.user);
location.reload();
```

Now the logged in user will be persisted across `docs`, `www`, and `studio` events and it'll be tagged appropriately in posthog.

### DIFF
--- a/packages/common/posthog-client.ts
+++ b/packages/common/posthog-client.ts
@@ -9,6 +9,7 @@ interface PostHogClientConfig {
 class PostHogClient {
   private initialized = false
   private pendingGroups: Record<string, string> = {}
+  private pendingIdentification: { userId: string; properties?: Record<string, any> } | null = null
   private config: PostHogClientConfig
 
   constructor(config: PostHogClientConfig = {}) {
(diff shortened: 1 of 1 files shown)


ivasilov (approved):
Approving to unblock. LGTM.

PR #37784, opened 2025-08-08
Button to add preventDefault onClick if disabled

joshenlim:
## Context

If Button is used with a link element (e.g `<Link>` or `<a>`), users will still be redirected to the link upon clicking the button even if the button is disabled. The changes here should address that by preventing the redirect if the button is disabled

Although do let me know if I might be overlooking something here! 🙏


packages/ui/src/components/Button/Button.tsx
```diff
@@ -6,7 +6,6 @@ import { Loader2 } from 'lucide-react'
 import { cloneElement, forwardRef, isValidElement } from 'react'
 import { SIZE_VARIANTS, SIZE_VARIANTS_DEFAULT } from '../../lib/constants'
 import { cn } from '../../lib/utils/cn'
-import { IconContext } from '../Icon/IconContext'
```
joshenlim, line 9: wasn't being used

alaister (approved):
LGTM!

- Tested on preview

SaxonF (approved):

PR #37785, opened 2025-08-08
chore: disallow larger instance size launch

kevcodez:
There is usually no good reason to immediately launch a project on a very large instance, unless it is fraudulent usage. To prevent easy fraud, we are not allowing to immediately launch >medium instances and just show a hint.

Screenshot 2025-08-08 at 18 15 17


alaister (approved):
LGTM!

- Tested on preview

PR #37787, opened 2025-08-08
Attempt to fix completion

SaxonF:
Noticing some code completion issues on prod. This switches to generateText and a standard fetch for the time being.


joshenlim (approved):
Tested on staging preview with SQL Editor and Edge Functions code editor for inline completions

PR #37788, opened 2025-08-08
AWS Marketplace Onboarding Page

ecktoteckto:
Page the customer gets redirected to from AWS Marketplace after signing a contract there. Long story short, what needs to be done on that page is to link a Supabase organization to the newly created AWS contract/AWS buyer. After linking, the organization will be managed (i.e. plan changes) and billed (i.e. invoicing + charges) by AWS. Before the customer can use that page, they either need to log in to an existing Supabase account or create a new Supabase account.

### Scenario A: Customer has existing organizations
- we distinguish between organizations that can be linked to AWS and organizations that can't be linked to AWS. Reasons why an organization can't be linked are e.g. permissions, outstanding invoices, organization already managed by AWS or Vercel
- in addition to selecting an existing organization, the customer can also create a new organization that will be managed by AWS

existing-orgs

### Scenario B: Customer does not have existing organizations
- the customer needs to create a new organization that will be managed by AWS

no-existing-orgs

### Testing instructions (can be tested locally only)

- **URL:** `http://localhost:8082/aws-marketplace-onboarding?buyer_id=96b6b1ee-1a99-45b9-8ea6-8ce26b2a02c3`
- **Branch to use for backend:** `thomas/billing-aws-marketplace-onboarding` - on that branch a few things are adjusted (e.g. skipping some validations) so that one can actually test the linking without running into errors

### Known issues (will address that later)
- in scenario A the "Link Organization" button is flickering when opening/closing the "Organizations that can't be linked" accordion

### Possible Improvements (not part of initial roll out)
- show reason(s) for why an organization can't be linked. We already get that info from the backend
- maybe show an overview of the AWS contract (start, end, AWS account ID etc.)


apps/studio/components/interfaces/Organization/CloudMarketplace/AwsMarketplaceCreateNewOrg.tsx
```diff
@@ -0,0 +1,94 @@
...
+      )}
+      <ScaffoldSection>
+        <ScaffoldSectionDetail>
+          <p className="mb-6 text-base">
+            You’ve subscribed to the Supabase {onboardingInfo?.plan_name_selected_on_marketplace}{' '}
+            Plan via the AWS Marketplace. As a final step, you need to create a Supabase
+            organization. That organization will be managed and billed through AWS Marketplace.
+            <br />
+            <br />
+            You can read more on billing through AWS in our {''}
```
ignaciodob, line 62: Let's maybe use padding on a different paragraph instead of `<br>`s? I think that will probably be easier to keep the spacing consistent

apps/studio/components/interfaces/Organization/CloudMarketplace/AwsMarketplaceAutoRenewalWarning.tsx
```diff
@@ -0,0 +1,37 @@
+import { Alert_Shadcn_, AlertDescription_Shadcn_, AlertTitle_Shadcn_ } from 'ui'
```
ignaciodob, line 1: I wonder if the Onboarding is the best place to call attention to this. 

What do you think about presenting this globally in Studio, like we do with Outstanding Invoices?

apps/studio/components/interfaces/Organization/CloudMarketplace/AwsMarketplaceLinkExistingOrg.tsx
```diff
@@ -0,0 +1,341 @@
...
+      <ScaffoldSection>
+        <ScaffoldSectionDetail>
+          <>
+            <p className="mb-6 text-base">
+              You’ve subscribed to the Supabase {onboardingInfo?.plan_name_selected_on_marketplace}{' '}
+              Plan via the AWS Marketplace. As a final step, you need to link a Supabase
+              organization to that subscription. Select the organization you want to be managed and
+              billed through AWS.
+              <br />
+              <br />
```
ignaciodob: Same nitpick about not using `<br>`

apps/studio/components/interfaces/Organization/CloudMarketplace/AwsMarketplaceCreateNewOrg.tsx
ecktoteckto: Done 👍

apps/studio/components/interfaces/Organization/CloudMarketplace/AwsMarketplaceLinkExistingOrg.tsx
ecktoteckto: Done 👍

apps/studio/components/interfaces/Organization/CloudMarketplace/AwsMarketplaceAutoRenewalWarning.tsx
ecktoteckto: On the onboarding page, you think that it gets too much attention and distracts from the actual onboarding (also because we already have the email we send) or that it doesn't get enough attention?

Re Studio: Not sure whether a banner at the very top would be too much (so far we're using it for incidents, scheduled maintenance,  overdue invoices). But yeah, maybe somewhere permanent in the Studio. We'd have to retrieve the Auto-Renewal info every time though since we don't get notified about the customer toggling "Auto Renewal".

ignaciodob: >On the onboarding page, you think that it gets too much attention and distracts from the actual onboarding (also because we already have the email we send) or that it doesn't get enough attention?

Distracts from the actual onboarding - yes.

>Re Studio: Not sure whether a banner at the very top would be too much (so far we're using it for incidents, scheduled maintenance, overdue invoices). But yeah, maybe somewhere permanent in the Studio. We'd have to retrieve the Auto-Renewal info every time though since we don't get notified about the customer toggling "Auto Renewal".

We can probably cache this pretty aggressively

kevcodez (approved):

apps/studio/components/interfaces/Organization/CloudMarketplace/AwsMarketplaceAutoRenewalWarning.tsx
ecktoteckto: Acknowledging your feedback @ignaciodob 👍  Gonna merge it now anyway so that we have it ready finally. Whether we show the warning or not is not a blocker in a technical sense. We can simply remove it later.

PR #37799, opened 2025-08-08
feat(studio): add granular data invalidation

taishikato:
# I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

✨ Feature - Adds granular cache invalidation for SQL executed through the AI Assistant to enable automatic UI updates after schema changes

The overall process flow is as follows:
1. Parse the SQL statement executed in the AI Assistant
   - ~~Determine the SQL action type~~
   - ~~Use regular expressions to extract entity type, schema, etc.~~
   - Use `libpg-query` to parse the SQL
2. Based on the information obtained above, use the appropriate invalidation handler and execute the it

🔔 **Update**: We scoped down to `Table`, `Function` and `Cron` entity types + `Create`, `Drop` and `Select` (for cron) actions
reference: https://github.com/supabase/supabase/pull/37799#issuecomment-3208123735

When users execute SQL statements through the AI Assistant, the UI doesn't automatically update to reflect the schema changes. Users must manually refresh the page to see the changes in table lists in the sidebar or other part of studio.

## What is the new behavior?

This PR introduces a granular cache invalidation system that:
- **Parses SQL statements** to detect DDL operations (CREATE, ~~ALTER~~, DROP, ~~ENABLE~~, ~~DISABLE~~, (SELECT for cron job))
- **Identifies affected entities**: tables, functions, procedures, ~~triggers, and policies~~
- **Invalidates specific React Query caches** for affected entities only
- **Updates UI components automatically** without manual refresh

https://github.com/user-attachments/assets/509820bd-bfda-4d60-b23a-393ee7d4a6f9

Key improvements:
- **Granular invalidation**: Only invalidates table-related query keys using `tableKeys.list()`
- **Non-blocking execution**: Uses asynchronous invalidation without blocking the caller
- **Schema-aware**: Targets specific schemas when available
- **Entity type updates**: Automatically updates the entity types list when tables change

### Scope

Currently supports:
- ✅ Tables
- ✅ Functions & Procedures
~~- ✅ Triggers~~
~~- ✅ RLS Policies~~
~~- ✅ RLS enable/disable (same process as table cuz SQL includes the word `table`)~~
~~- ✅ index~~
- ✅ cron job
~~- ✅ view~~
~~- ✅ schema~~
~~- ✅  materialized_view~~

### Tables

**Executed SQL**

- CREATE
```sql
create table public.todos (
  id bigint primary key generated always as identity,
  user_id uuid not null references auth.users(id),
  task text not null,
  is_complete boolean not null default false,
  inserted_at timestamp with time zone not null default now(),
  updated_at timestamp with time zone not null default now()
);
```

- DROP
```sql
drop table if exists public.todos cascade;
```

https://github.com/user-attachments/assets/81ffa3f2-462f-4987-bcbd-3fa92395f59a

https://github.com/user-attachments/assets/c31553ef-8015-4b58-b3fe-d495fa4b1a7e

https://github.com/user-attachments/assets/fb9db5a3-cf91-4c23-8e9f-135d015c6899

### Triggers

**Executed SQL**

- CREATE
```sql
CREATE TRIGGER last_changed_auto_update_timestamp
BEFORE UPDATE ON public.last_changed
FOR EACH ROW
EXECUTE FUNCTION public.update_last_updated_timestamp();
```

- DROP
```sql
DROP TRIGGER IF EXISTS last_changed_auto_update_timestamp ON public.last_changed;
```

https://github.com/user-attachments/assets/c6746608-01e2-4f09-887f-9f4407334125

### Functions

**Executed SQL**

- CREATE
```sql
create or replace function public.add_numbers(a integer, b integer)
returns integer
language sql
as $$
    select a + b;
$$;
```

- DROP
```sql
drop function if exists public.add_numbers(integer, integer);
```

https://github.com/user-attachments/assets/775b1ab3-fab2-4b85-98ae-8b9b2776a791

### Indexes

**Executed SQL**

- CREATE
```sql
CREATE INDEX IF NOT EXISTS todos_user_complete_inserted_idx ON public.todos (user_id, is_complete, inserted_at DESC);
```

- DROP
```sql
DROP INDEX IF EXISTS public.todos_user_complete_inserted_idx;
```

https://github.com/user-attachments/assets/d54d2ab1-dfec-4004-b932-88cc403666ac

### RLS enabling and Policies

**Executed SQL**

- Enabling RLS
```sql
ALTER TABLE public.todos ENABLE ROW LEVEL SECURITY;
```

- CREATE POLICY
```sql
CREATE POLICY "Users can view their own todos" ON public.todos
FOR SELECT TO authenticated
USING ((SELECT auth.uid()) = user_id);
```

- DROP
```sql
DROP POLICY IF EXISTS "Users can view their own todos" ON public.todos;
```

https://github.com/user-attachments/assets/ce1d2c72-ab6f-4a09-9dda-5a304ef2b248

### Cron job
**Executed SQL**
- cron job creation
```sql
SELECT cron.schedule(
    'daily-system-cleanup',
    '0 0 * * *',
    'SELECT public.perform_periodic_cleanup()'
);
```

- cron job deletion
```sql
SELECT cron.unschedule(22);
```

https://github.com/user-attachments/assets/27e613e0-b428-454b-9aa5-a457dc5136a4

### Views

**Executed SQL**

- CREATE
```sql
CREATE VIEW public.example_todos_view WITH (security_invoker=on) AS
SELECT id, user_id, task, is_complete, inserted_at
FROM public.todos
WHERE is_complete = false;
```

- DROP
```sql
DROP VIEW IF EXISTS public.example_todos_view;
```

https://github.com/user-attachments/assets/6ddff938-8542-4168-8464-bc141f071829

### Schema

**Executed SQL**

- CREATE
```sql
create schema if not exists private;
```

- DROP
```sql
drop schema if exists private cascade;
```

https://github.com/user-attachments/assets/f02d8ce7-f26a-456a-b3e8-fd067ee3a82a

### Materialized Views

**Executed SQL**

- CREATE
```sql
create materialized view private.todo_completion_summary as
select
  user_id,
  count(*) filter (where is_complete) as completed_count,
  count(*) filter (where not is_complete) as incomplete_count
from public.todos
group by user_id;
```

- DROP
```sql
drop materialized view if exists private.todo_completion_summary;
```

https://github.com/user-attachments/assets/739ea4a3-292d-4d91-be3b-b92adbf51579


taishikato:
~~This invalidator also works with multiple SQL queries like below. reference:~~
~~https://github.com/supabase/supabase/pull/37799/commits/a355a1cbccc29bdcced22a6fcd7dae33794620e4~~
```sql
create schema if not exists private;

create materialized view private.todo_completion_summary as
select
  user_id,
  count(*) filter (where is_complete) as completed_count,
  count(*) filter (where not is_complete) as incomplete_count
from public.todos
group by user_id;
```

Now `libpg-query` takes care of this.

apps/studio/lib/granular-data-invalidation/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  return events
+}
+
+/**
+ * Process SQL and handle invalidation automatically
+ * This is the main entry point for processing SQL statements
+ */
+export async function invalidateDataGranularly(
```
ivasilov, line 27: Can you try writing 3-4 tests for this function with some bigger queries? You can change the function definition if it makes it easier.

ivasilov (changes requested):
Great work on the PR! I see you're proficient in regex fu :) 

I'm kinda afraid the parsing will be brittle, can you try parsing the query with https://github.com/supabase-community/pg-parser? It supports PG 15,16 and 17 so should be good. I haven't tested but should work in nextjs (ping me if you need help!) 

You don't need to handle all cases since this is proof-of-concept, let's just say 2 actions for 2 entity types + the cron feature.

taishikato:
Hey @ivasilov ty for the comments 🙏 

I tried experimenting with using [@supabase/pg-parser](https://github.com/supabase-community/pg-parser) for a while.  

In short - i couldn’t get past the following error caused by doing `const parser = new PgParser();`:
```
 ⨯ ./node_modules/.pnpm/@supabase+pg-parser@0.1.3/node_modules/@supabase/pg-parser/wasm/17/pg-parser.js:44:35
Module not found: Can't resolve 'module'
  42 |   // When building an ES module `require` is not normally available.
  43 |   // We need to use `createRequire()` to construct the require()` function.
> 44 |   const { createRequire } = await import('module');
     |                                   ^^^^^^^^^^^^^^^^
  45 |   /** @suppress{duplicate} */
  46 |   var require = createRequire(import.meta.url);
  47 |


https://nextjs.org/docs/messages/module-not-found
```

It seems that in this part of `apps/studio/node_modules/@supabase/pg-parser/wasm/[15-17]/pg-parser.js`, Next.js is mixing up the environment with Node.js.
```javascript
var ENVIRONMENT_IS_NODE = typeof process == 'object' && typeof process.versions == 'object' && typeof process.versions.node == 'string' && process.type != 'renderer';
.
.
if (ENVIRONMENT_IS_NODE) {
  // When building an ES module `require` is not normally available.
  // We need to use `createRequire()` to construct the require()` function.
  const { createRequire } = await import('module');
  /** @suppress{duplicate} */
  var require = createRequire(import.meta.url);

}
```

I tried loading the library dynamically / tweaking the webpack settings, but the issue still wasn’t resolved.

FYI: the error occurs with both `pnpm run dev:studio-local` and `pnpm run build:studio`.

apps/studio/components/ui/QueryBlock/QueryBlock.tsx
```diff
@@ -164,10 +164,18 @@ export const QueryBlock = ({
...
       onResults?.(data.result)
       setQueryResult(data.result)
 
+      try {
+        const { parseQuery } = await import('libpg-query')
+        const parsedResult = await parseQuery(sql || '')
+        console.log('Parsed SQL result:', parsedResult)
+      } catch (parseError) {
+        console.error('Error parsing SQL:', parseError)
+      }
```
taishikato: Hey @ivasilov i just want to confirm :D - are you also suggesting that the data invalidation function should be called in the `QueryBlock` component instead of [execute-sql-mutation.ts](https://github.com/taishikato/supabase/blob/feature/invalidate-table-list/apps/studio/data/sql/execute-sql-mutation.ts#L45) as it is now?

ivasilov: Not really, I was just trying to make it work, you can move this code to your function. I would suggest you make it easy for the `invalidateDataGranularly` function to be unit tested without mocking. Ideally it would be a pure function.

taishikato:
Hey @ivasilov I've done these below 😃
- limited entity types to `Table`, `Function` + `Cron`. Actions are only `Create`, `Drop` + `Select` (for cron)  
- switched from regex parsing to using `libpg-query`

taishikato:
I split the functions into smaller parts to narrow the scope of React Query’s influence 🫡

I think writing unit tests for each function should be easier than before with this change.
Now `invalidateDataGranularly` looks below:
```typescript
export async function invalidateDataGranularly(
  queryClient: QueryClient,
  sql: string,
  projectRef: string
): Promise<void> {
  if (!sql || !projectRef) {
    console.warn('invalidateDataGranularly: Invalid input - SQL and projectRef are required')
    return
  }

  try {
    const events = await parseSqlStatements(sql, projectRef)
    const plan = planInvalidationsFromEvents(events)
    await applyInvalidationPlan(queryClient, plan)
  } catch (error) {
    console.error('invalidateDataGranularly: Error processing SQL', error)
  }
}
```

- That said, as long as `invalidateDataGranularly` takes `queryClient` as an argument (and uses it for `applyInvalidationPlan()`, it’s difficult to make `invalidateDataGranularly` itself a pure function.

  For example: if we leave the invalidation to the caller ([execute-sql-mutation.ts](https://github.com/taishikato/supabase/blob/feature/invalidate-table-list/apps/studio/data/sql/execute-sql-mutation.ts) in this case) like below, `invalidateDataGranularly` wouldn’t need to care about `queryClient` anymore.

  ```typescript
  // execute-sql-mutation.ts
  if (granularInvalidation && projectRef) {
    const invalidationActions = await invalidateDataGranularly(queryClient, sql, projectRef)
    const promises = invalidationActions.map((action) =>
      queryClient.invalidateQueries({
        queryKey: action.key,
        exact: action.exact,
        refetchType: action.refetchType,
      })
    )
  
    await Promise.allSettled(promises)
  } else {
  .
  .
  ```

- Since we’re using `libpg-query`, the functions that depend on it (`parseSqlStatements()`) might need to be mocked. With vitest, it tries to fetch the Node binary, and for wasm it seems tricky unless we adjust MSW.

For now, I added [a simple unit test](https://github.com/taishikato/supabase/blob/feature/invalidate-table-list/apps/studio/tests/granular-data-invalidation/plan-invalidations-from-events.test.ts) for `planInvalidationsFromEvents()`.

taishikato:
hey @ivasilov, I made changes to narrow the scope of React Query’s influence 🫡

I think writing unit tests for each function should be easier than before with this change (or at least the structure has become simpler)
Now `invalidateDataGranularly()` looks below:
```typescript
export async function invalidateDataGranularly(
  queryClient: QueryClient,
  sql: string,
  projectRef: string
): Promise<void> {
  if (!sql || !projectRef) {
    console.warn('invalidateDataGranularly: Invalid input - SQL and projectRef are required')
    return
  }

  try {
    const events = await parseSqlStatements(sql, projectRef)
    const plan = planInvalidationsFromEvents(events)
    await applyInvalidationPlan(queryClient, plan)
  } catch (error) {
    console.error('invalidateDataGranularly: Error processing SQL', error)
  }
}
```

- That said, as long as `invalidateDataGranularly()` takes `queryClient` as an argument (and uses it for `applyInvalidationPlan()`, it’s difficult to make `invalidateDataGranularly()` itself a pure function.

  For example: if we leave the invalidation to the caller ([execute-sql-mutation.ts](https://github.com/taishikato/supabase/blob/feature/invalidate-table-list/apps/studio/data/sql/execute-sql-mutation.ts) in this case) like below, `invalidateDataGranularly()` wouldn’t need to care about `queryClient` anymore.

  ```typescript
  // execute-sql-mutation.ts
  if (granularInvalidation && projectRef) {
    const invalidationActions = await invalidateDataGranularly(sql, projectRef)
    const promises = invalidationActions.map((action) =>
      queryClient.invalidateQueries({
        queryKey: action.key,
        exact: action.exact,
        refetchType: action.refetchType,
      })
    )
  
    await Promise.allSettled(promises)
  } else {
  .
  .
  ```

- Since we’re using `libpg-query`, the functions that depend on it (`parseSqlStatements()`) might need to be mocked. With vitest, it tries to fetch the Node binary, and for wasm it seems tricky unless we adjust MSW.

For now I added [a simple unit test](https://github.com/taishikato/supabase/blob/feature/invalidate-table-list/apps/studio/tests/granular-data-invalidation/plan-invalidations-from-events.test.ts) for `planInvalidationsFromEvents()`.

ivasilov:
>  That said, as long as `invalidateDataGranularly()` takes `queryClient` as an argument (and uses it for `applyInvalidationPlan()`, it’s difficult to make `invalidateDataGranularly()` itself a pure function.
>  For example: if we leave the invalidation to the caller ([execute-sql-mutation.ts](https://github.com/taishikato/supabase/blob/feature/invalidate-table-list/apps/studio/data/sql/execute-sql-mutation.ts) in this case) like below, `invalidateDataGranularly()` wouldn’t need to care about `queryClient` anymore.
>  ```ts
>  // execute-sql-mutation.ts
>  if (granularInvalidation && projectRef) {
>    const invalidationActions = await invalidateDataGranularly(sql, projectRef)
>    const promises = invalidationActions.map((action) =>
>      queryClient.invalidateQueries({
>        queryKey: action.key,
>        exact: action.exact,
>        refetchType: action.refetchType,
>      })
>  )
>       
>     await Promise.allSettled(promises)
>   } else {
>   .
>   .
>   ```
Yeah, let's take this approach. It will be easier for testing. Only write the tests for `invalidateDataGranularly(sql, projectRef)`.

>  Since we’re using `libpg-query`, the functions that depend on it (`parseSqlStatements()`) might need to be mocked. With vitest, it tries to fetch the Node binary, and for wasm it seems tricky unless we adjust MSW.

You're right, but it seems to work with `libpg-query` v17.6+. Can you try bumping the version only for `studio` and see if it runs? v17 renames the `parseQuery` function to `parse`, but should work the same. Sorry for the runaround with the pg library.

taishikato:
@ivasilov i tried `libpg-query` v17.6.0 and v17.5.5 but hit the error below. looks like a similar issue as `@supabase/pg-parser` 🤔

implementation I tried in `parse-sql-statements.ts`:
```typescript
const { parse } = await import('libpg-query')
const parsed = await parse(sql)
```

Error:
```
 ⨯ ./node_modules/.pnpm/libpg-query@17.6.0/node_modules/libpg-query/wasm/libpg-query.js:7:599
Module not found: Can't resolve 'fs'
```

taishikato:
Hey @ivasilov, I've made a couple updates 🫡 :
- moved `queryClient` outside of `invalidateDataGranularly()`: https://github.com/supabase/supabase/pull/37799/commits/d3b4756e313dc2c5c89e85a546750652d2762bcb
- added unit tests for `invalidateDataGranularly()` (no mocks): [index.test.ts](https://github.com/taishikato/supabase/blob/feature/invalidate-table-list/apps/studio/tests/granular-data-invalidation/index.test.ts)

joshenlim:
@taishikato just took a quick skim at this one - wondering if what you've done with `granularInvalidation` can just replace `contextualInvalidation` (i think they're both trying to do the same thing?), that way we consolidate this behaviour for both the SQL editor and AI assistant

taishikato:
@joshenlim yea that makes sense!
I initially created `granularInvalidation` as a new setting since I was trying to avoid changing the existing code.
but it might be better to just consolidate everything under `contextualInvalidation`.

Issue #12661, opened 2023-02-24
Kong unable to generate a unique filename


WesleyYue:
I have this intermittent bug on CI (which is on a server with a lot of processors) where Kong will fail to start, and the logs will output something along the lines of `unable to generate a unique filename`. @ColinChartier from webapp.io found the root of the issue, which we believe is:

1. Kong will occasionally execute a script via [pl_utils (a Lua library)](https://github.com/Kong/kong/blob/35c4b7a22d266262e4165fa9869a73c21202c677/kong/tools/utils.lua#L109)
2. [That function will try to make a tmp file](https://stevedonovan.github.io/Penlight/api/source/utils.lua.html)
3. [This fails b/c the docker compose is read only](https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml#L49)
4. [Unable to generate unique filename error is thrown](https://www.lua.org/source/5.3/loslib.c.html#lua_tmpnam:~:text=%22unable%20to%20generate%20a%20unique%20filename%22)

So I think the fix is just setting the kong volume to allow writes in docker compose, but I'm not sure if that will have some other interaction with the rest of the Supabase system.

WesleyYue:
There's also this very old issue from the Kong repo, which I think is related https://github.com/Kong/kong/issues/5134

PR #37810, opened 2025-08-09
fix directory mounting issue in rw

pilinux:

YES

Bug fix

This is related to https://github.com/supabase/supabase/issues/12661. Even though the issue was closed, but the underlying problem was never solved. Hence, still in the master branch, kong cannot run in a docker container.

### Exactly what went wrong?

`./volumes/api/kong.yml:/home/kong/temp.yml:ro,z` mounted just one file in read‑only. `bash -c 'eval "echo \"$$(cat ~/temp.yml)\"" > ~/kong.yml` cannot create any new file in this directory and this forces kong to quit and restart forever.

With this fix, `./volumes/api:/home/kong:z` mounts the directory in r/w mode. Now the entrypoint script can read the template `kong.yml` and write an expanded, interpolated version `kong-local.yml` into the same directory without changing the original `kong.yml` config file.

Changing `KONG_DECLARATIVE_CONFIG` to `/home/kong/kong-local.yml` points kong at the generated file rather than the raw template.

The `:z` label keeps SELinux contexts compatible when needed.

## What is the new behavior?

With this fix, kong container works as it should.

## Summary by CodeRabbit

* **Chores**
  * Updated Kong service volume mounting configuration from single file to directory-based approach.
  * Modified Kong service declarative configuration file path and related entrypoint logic.


aantti:
Tbf, I'm not sure this had been a problem - I hadn't seen it neither across the environments I had access to, nor elsewhere around here. Anyways, the Kong configuration has been re-worked quite a bit in PR #43554.

PR #37828, opened 2025-08-11
docs: add flutter phone factor reference doc and example

grdsdev:

YES/NO

docs update

no reference for `phone` factor MFA enrollment in Flutter docs

## What is the new behavior?

Add example for enrolling phone factor in Flutter

## Additional context

### DIFF
--- a/apps/docs/spec/supabase_dart_v2.yml
+++ b/apps/docs/spec/supabase_dart_v2.yml
@@ -1633,9 +1633,9 @@ functions:
     notes: |
       This section contains methods commonly used for Multi-Factor Authentication (MFA) and are invoked behind the `supabase.auth.mfa` namespace.
 
-      Currently, we only support time-based one-time password (TOTP) as the 2nd factor. We don't support recovery codes but we allow users to enroll more than 1 TOTP factor, with an upper limit of 10.
+      Currently, there is support for time-based one-time password (TOTP) and phone verification code as the 2nd factor. Recovery codes are not supported but users can enroll multiple factors, with an upper limit of 10.
 
-      Having a 2nd TOTP factor for recovery means the user doesn't have to store their recovery codes. It also reduces the attack surface since the recovery factor is usually time-limited and not a single static code.
+      Having a 2nd factor for recovery frees the user of the burden of having to store their recovery codes somewhere. It also reduces the attack surface since multiple recovery codes are usually generated compared to just having 1 backup factor.
 
       Learn more about implementing MFA on your application on our guide [here](https://supabase.com/docs/guides/auth/auth-mfa#overview).
   - id: mfa-enroll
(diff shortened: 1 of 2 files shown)


apps/docs/spec/supabase_dart_v2.yml
```diff
@@ -1633,9 +1633,9 @@ functions:
     notes: |
       This section contains methods commonly used for Multi-Factor Authentication (MFA) and are invoked behind the `supabase.auth.mfa` namespace.
 
-      Currently, we only support time-based one-time password (TOTP) as the 2nd factor. We don't support recovery codes but we allow users to enroll more than 1 TOTP factor, with an upper limit of 10.
+      Currently, there is support for time-based one-time password (TOTP) and phone verification code as the 2nd factor. Recovery codes are not supported but users can enroll multiple factors, with an upper limit of 10.
```
ChrisChinchilla:
```suggestion
      Currently, the Dart SDK supports time-based one-time password (TOTP) and phone verification code as the 2nd factor. Recovery codes are not supported but users can enroll multiple factors, with an upper limit of 10.
```

Assuming that the Dart SDK is what supports those features, if it's Supabase itself, then update to that instead, just make sure what supports what is clear.

"Multiple factors" sounds odd, but I can't think of a better way to say it 🤷 .

Also, if only two factors are supported, why is there a limit of 10? This could be misunderstanding, but I was a little confused by that.

apps/docs/spec/supabase_dart_v2.yml
```diff
@@ -1644,7 +1644,7 @@ functions:
       Starts the enrollment process for a new Multi-Factor Authentication (MFA) factor. This method creates a new `unverified` factor.
       To verify a factor, present the QR code or secret to the user and ask them to add it to their authenticator app.
       The user has to enter the code from their authenticator app to verify it.
-      - Currently, `totp` is the only supported `factorType`. The returned `id` should be used to create a challenge.
+      - Use `totp` or `phone` as the `factorType` and use the returned `id` to create a challenge.
```
ChrisChinchilla:
```suggestion
      - Use `totp` or `phone` as the `factorType` and the returned `id` to create a challenge.
```

apps/docs/spec/supabase_dart_v2.yml
```diff
@@ -1661,6 +1661,10 @@ functions:
         isOptional: true
         type: String
         description: Human readable name assigned to the factor.
+      - name: phone
+        isOptional: true
+        type: String
+        description: Phone number to enroll for Phone factor type.
```
ChrisChinchilla:
```suggestion
        description: Phone number to enroll for phone factor type.
```

apps/docs/spec/supabase_js_v2.yml
```diff
@@ -2128,6 +2128,8 @@ functions:
       Currently, there is support for time-based one-time password (TOTP) and phone verification code as the 2nd factor. Recovery codes are not supported but users can enroll multiple factors, with an upper limit of 10.
 
       Having a 2nd factor for recovery frees the user of the burden of having to store their recovery codes somewhere. It also reduces the attack surface since multiple recovery codes are usually generated compared to just having 1 backup factor.
+
+      Learn more about implementing MFA on your application on our guide [here](https://supabase.com/docs/guides/auth/auth-mfa#overview).
```
ChrisChinchilla:
```suggestion
      Learn more about implementing MFA in your application [in the MFA guide](https://supabase.com/docs/guides/auth/auth-mfa#overview).
```

ChrisChinchilla (changes requested):
Few small things

apps/docs/spec/supabase_dart_v2.yml
grdsdev: @ChrisChinchilla I copy pasted this notes from the JS reference, need changing there too.

It is Supabase itself that supports TOTP and phone based.

> Also, if only two factors are supported, why is there a limit of 10? This could be misunderstanding, but I was a little confused by that.
There are only 2 "kinds" of factors, TOTP and Phone, but you can enroll multiple of each, with a max of 10.

Let me know if I should change notes from JS in this PR, or accept as this, and change all references at once later?

ChrisChinchilla: @grdsdev 

In that case, let's do:

```suggestion
      Currently, Supabase supports time-based one-time password (TOTP) and phone verification code as the 2nd factor. Recovery codes are not supported but users can enroll multiple factors, with an upper limit of 10..
```

And I'll make a separate task to clear up the other references and do them all together.

ChrisChinchilla (approved):
💯 🚀

PR #37837, opened 2025-08-11
fix: Handle a race condition with permissions API and the cell editors in the Table Editor

### LINKED ISSUE #33364: Unable to edit editable cells by double clicking.

ivasilov:
This PR adds code to recreate the grid columns when the `editable` flag changes. This fixes a bug where the permissions would come after the grid has been created and the editable flag will not be propagated when editing table cells.

I've also removed references to `snap.gridColumns` from the Table/JSON/Select editors because the data was already present in the `column` prop. This was a leftover when we used `react-tracked`.

Steps to verify:
1. Add `await new Promise(resolve => setTimeout(() => resolve(), 10000)`
2. Go to Table Editor
3. While the permissions haven't loaded yet, focus a field with text or json editor and try to change its value
4. The editor will be read-only. Once the permissions have loaded, the editor will switch to being editable.

Fixes #33364.

### DIFF
--- a/apps/studio/components/grid/components/editor/BooleanEditor.tsx
+++ b/apps/studio/components/grid/components/editor/BooleanEditor.tsx
@@ -1,5 +1,4 @@
 import type { RenderEditCellProps } from 'react-data-grid'
-import { useTableEditorTableStateSnapshot } from 'state/table-editor-table'
 import { Select } from 'ui'
 
 interface Props<TRow, TSummaryRow = unknown> extends RenderEditCellProps<TRow, TSummaryRow> {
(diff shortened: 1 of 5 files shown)


jordienr (approved):
tested in preview

PR #37853, opened 2025-08-12
fix: resolve 50% session tracking discrepancy on root route

seanoliver:
## Summary
- Fixes race condition causing ~50% discrepancy in session tracking between client and server on root route
- Implements event queuing to capture telemetry events before PostHog initializes
- Adds session ID correlation to enable unified tracking analysis between client and server events

Resolves GROWTH-461

## Problem
We observed a discrepancy in unique session counts on the root route only:
- Client-side tracking was missing ~50% of sessions compared to server-side
- This was only happening on the root route, which uses a lot of dynamic imports compared to other pages
- Hypothesis is that users were bouncing before PostHog client library initialized
- This hypothesis can't be fully validated because we don't currently share a unique session id between client and server

CleanShot 2025-08-11 at 17 59 33

CleanShot 2025-08-11 at 18 25 53@2x

## Solution
1. **Event Queuing**: Queue PV events while PostHog initializes, preventing data loss from early user interactions
2. **Session Correlation**: Extract PostHog session ID from cookies and sync between client/server for unified tracking
3. **Reliable Page Leave**: Use `sendBeacon` API to ensure page-leave events are captured even during navigation
4. **Production Safeguards**: Add `try`/`catch` blocks and memory caps to prevent API errors from breaking our tracking

## Test Plan
- [x] Verify events are queued and sent after PostHog initializes
- [x] Confirm session IDs match between client and server telemetry
- [x] Test page-leave events are captured on navigation
- [x] Vercel preview URLs added to UserCentrics for testing

### DIFF
--- a/packages/common/posthog-client.ts
+++ b/packages/common/posthog-client.ts
@@ -3,7 +3,7 @@ import { PostHogConfig } from 'posthog-js'
 
 // Limit the max number of queued events
 // (e.g. if a user navigates around a lot before accepting consent)
-const MAX_PENDING_EVENTS = 200
+const MAX_PENDING_EVENTS = 20
 
 interface PostHogClientConfig {
   apiKey?: string
@@ -40,6 +40,10 @@ class PostHogClient {
       capture_pageview: false, // We'll manually track pageviews
       capture_pageleave: false, // We'll manually track page leaves
       loaded: (posthog) => {
+        // Apply pending properties that were set before PostHog
+        // initialized due to poor connection or user not accepting
+        // consent right away
+
         // Apply any pending groups
         Object.entries(this.pendingGroups).forEach(([type, id]) => {
           posthog.group(type, id)
@@ -62,7 +66,7 @@ class PostHogClient {
         // Flush any pending events
         this.pendingEvents.forEach(({ event, properties }) => {
           try {
-            posthog.capture(event, properties)
+            posthog.capture(event, properties, { transport: 'sendBeacon' })
           } catch (error) {
             console.error('PostHog capture failed:', error)
           }
@@ -80,6 +84,7 @@ class PostHogClient {
 
     if (!this.initialized) {
       // Queue the event for when PostHog initializes (up to cap)
+      // (e.g. poor connection or user not accepting consent right away)
       if (this.pendingEvents.length >= this.maxPendingEvents) {
         this.pendingEvents.shift() // Remove oldest event
       }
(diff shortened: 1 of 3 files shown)


packages/common/posthog-client.ts
```diff
@@ -42,9 +48,26 @@ class PostHogClient {
 
         // Apply any pending identification
```
pamelachia, line 53: would be good to explain why need to wait for identification

packages/common/posthog-client.ts
```diff
@@ -53,21 +76,49 @@ class PostHogClient {
...
-    if (!hasConsent || !this.initialized) return
+    if (!hasConsent) return
 
-    // Store groups from properties if present (for later group() calls)
-    if (properties.$groups) {
-      Object.entries(properties.$groups).forEach(([type, id]) => {
-        if (id) posthog.group(type, id as string)
-      })
+    if (!this.initialized) {
+      // Queue the event for when PostHog initializes (up to cap)
```
pamelachia, line 86: add comment to explain why we need to do this queueing thing

packages/common/posthog-client.ts
```diff
@@ -1,6 +1,10 @@
 import posthog from 'posthog-js'
 import { PostHogConfig } from 'posthog-js'
 
+// Limit the max number of queued events
+// (e.g. if a user navigates around a lot before accepting consent)
+const MAX_PENDING_EVENTS = 200
```
pamelachia: that's quite a lot of events that are being queued, maybe we can reduce it?

pamelachia: we should probably reduce it to 20

the estimated time for these events to be processed should be quite quick right? I didn't manage to queue it long enough so can't see but we probably shouldn't store such a long 200 element list if it hits max.

pamelachia (approved):
i didn't get any lag when doing identify and event sending so not sure how to test that the queueing thing works but looks good to me otherwise. events do get sent to posthog.

joshenlim (approved):
Code wise generally LGTM - although as per Pam's comment above, not sure how to test this one locally / on staging, approving to unblock although do address Pam's comments before merging! 🙏 🙂

packages/common/posthog-client.ts
seanoliver: I think the only reason it might not be very quick is if someone navigates around on the website a lot before finally clicking "accept consent," but even then you're right that 200 events seems like overkill. Reducing this to 20!

packages/common/posthog-client.ts
seanoliver: Added a comment at the top of this callback! The main two reasons are poor connection and people not accepting consent right away before they start interacting with the page.

packages/common/posthog-client.ts
seanoliver: Done! ✅

PR #37854, opened 2025-08-12
Chore/align bucket and object naming validation with storage api

### LINKED ISSUE #34470: cant create storage bucket with underscrore from UI but can programatically
# Bug report

- [ x] I confirm this is a bug with Supabase, not with my own application. 
- [x ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

In the supabase UI I get an error "The name of the bucket must only container lowercase letters, numbers, dots, and hyphens" if I try to create a bucket with an underscore in the name.
However I can create a bucket with an underscore programatically.
I previously created buckets using the supabase UI with underscore and I need to do this in future because our live site has buckets with underscores in the name and we have a staging site which we frequently reset to be the same as production.

## To Reproduce

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

joshenlim:
Fixes https://github.com/supabase/supabase/issues/34470#issuecomment-2883734179

## Context

There's an inconsistency between the Storage UI on the dashboard and the Storage API itself's naming convention for buckets and folders. This PR aligns the naming validation between both such that the dashboard will be identical to that of the Storage API.

## Changes
- Update bucket naming validation, we'd also try to show which character is causing the error if an invalid character is provided
  - Can test with for e.g `new/bucket/name` (slashes are not allowed)
- Update folder naming validation, we'd also try to show which character is causing the error if an invalid character is provided
  - Can test with for e.g `folder#name` (hashes are not allowed)

### DIFF
--- a/apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
+++ b/apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
@@ -196,10 +196,9 @@ const CreateBucketModal = ({ visible, onClose }: CreateBucketModalProps) => {
               name="name"
               render={({ field }) => (
                 <FormItemLayout
+                  layout="vertical"
                   label="Name of bucket"
                   labelOptional="Buckets cannot be renamed once created."
-                  description="Only lowercase letters, numbers, dots, and hyphens"
-                  layout="vertical"
                 >
                   <FormControl_Shadcn_>
                     <Input_Shadcn_ {...field} placeholder="Enter bucket name" />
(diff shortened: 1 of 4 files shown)


alaister (approved):
LGTM!

- Tested on preview

One small improvement would be to add `underscores` to this list as it seems like they're allowed now:
Screenshot 2025-08-12 at 15 33 36

joshenlim:
> One small improvement would be to add underscores to this list as it seems like they're allowed now:

actually good point - think i'd rather remove this description since its a lot more than underscores that are allowed now iinw 😅

PR #37863, opened 2025-08-12
Chore/support copy row in table editor

joshenlim:
Fixes FE-1641

## Context

We've seen a [request](https://x.com/maxrusakovic/status/1922608175275971009) to support copying rows instead of just exporting rows from the Table Editor, so this PR adds the support for that

## Changes involved
- Adds a new Copy button in the Table Editor header to copy as CSV, SQL or JSON
  image
  - JFYI we're disabling copy if all rows in the table are selected via the "Select all rows in table" button
    - To prevent potential memory issues if the table turns out to be really really large as clipboard contents live in memory
    image

- Also adding a new context menu action to copy the row
image


alaister (approved):
LGTM!

- Tested on preview

PR #37874, opened 2025-08-12
use dynamic max values in database reports

jordienr:
- uses dynamic iops and max connection values

# to test

1. go to db report
2. see max iops and connections
3. go to db settings 
4. update max iops → save
5. go back to report
6. should show new max value


saltcod (approved):
Changed disk size on staging and was reflected in this chart

PR #37879, opened 2025-08-12
chore: add confirmation modal when enabling SMS MFA

ignaciodob:
This PR adds a confirmation modal when a user attempts to enable SMS MFA - better informing the user of the billing implications.

### Testing
- Head over to `http://localhost:8082/project/_/auth/mfa`
- Disable `Phone` if enabled
- Enable `Phone` 
- Click on Save changes
- Assert that the Confirmation modal is shown
- Confirm
- Refresh the page and assert that the changes were saved and that `Phone` is enabled
- Disable `Phone` 
- Click on Save Changes and assert that the confirmation is not shown

### DIFF
--- a/apps/studio/components/interfaces/Auth/MfaAuthSettingsForm/MfaAuthSettingsForm.tsx
+++ b/apps/studio/components/interfaces/Auth/MfaAuthSettingsForm/MfaAuthSettingsForm.tsx
@@ -35,6 +35,7 @@ import {
   WarningIcon,
 } from 'ui'
 import { FormItemLayout } from 'ui-patterns/form/FormItemLayout/FormItemLayout'
+import ConfirmationModal from 'ui-patterns/Dialogs/ConfirmationModal'
 
 function determineMFAStatus(verifyEnabled: boolean, enrollEnabled: boolean) {
   return verifyEnabled ? (enrollEnabled ? 'Enabled' : 'Verify Enabled') : 'Disabled'
@@ -86,6 +87,8 @@ const MfaAuthSettingsForm = () => {
   // Separate loading states for each form
   const [isUpdatingTotpForm, setIsUpdatingTotpForm] = useState(false)
   const [isUpdatingPhoneForm, setIsUpdatingPhoneForm] = useState(false)
+
+  const [isConfirmationModalVisible, setIsConfirmationModalVisible] = useState(false)
 
   const { can: canReadConfig } = useAsyncCheckProjectPermissions(
     PermissionAction.READ,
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Auth/MfaAuthSettingsForm/MfaAuthSettingsForm.tsx
```diff
@@ -426,9 +443,9 @@ const MfaAuthSettingsForm = () => {
                     <WarningIcon />
                     <AlertTitle_Shadcn_>
                       Enabling advanced MFA with phone will result in an additional charge of{' '}
-                      <span translate="no">$75</span>
-                      per month for the first project in the organization and an additional{' '}
-                      <span translate="no">$10</span> per month for additional projects.
+                      <span translate="no">$75</span> per month for the first project in the
+                      organization and an additional <span translate="no">$10</span> per month for
+                      additional projects.
```
ignaciodob: @kevcodez, now that we have the confirmation modal, should we maybe remove this from the form and only keep the warning in the modal?
I can go either way

ecktoteckto: Question was for Kevin, but IMO we should remove it from the form. Makes the form cleaner.

apps/studio/components/interfaces/Auth/MfaAuthSettingsForm/MfaAuthSettingsForm.tsx
```diff
@@ -458,6 +475,30 @@ const MfaAuthSettingsForm = () => {
...
+        title="Confirm SMS MFA"
+        confirmLabel="Confirm and save"
+        onCancel={() => setIsConfirmationModalVisible(false)}
+        onConfirm={() => {
+          setIsConfirmationModalVisible(false)
+          phoneForm.handleSubmit(onSubmitPhoneForm)()
+        }}
+        variant="warning"
+      >
+        <Alert_Shadcn_ variant="warning">
```
ecktoteckto: Maybe let's not use an alert inside the modal but a plain text instead? The modal is "disruptive" enough. The alert inside the modal makes my eyes cry 😵‍💫 😉

apps/studio/components/interfaces/Auth/MfaAuthSettingsForm/MfaAuthSettingsForm.tsx
```diff
@@ -458,6 +475,30 @@ const MfaAuthSettingsForm = () => {
...
+          <WarningIcon />
+          <AlertTitle_Shadcn_>
+            Enabling advanced MFA with phone will result in an additional charge of{' '}
```
ecktoteckto: I know that you simply reused the existing text. Might make sense to use "SMS MFA" instead of "advanced MFA". Furhter down and as the headline of the form we use "SMS MFA".

ecktoteckto (approved):
LGTM. Tested on Preview. Already approved to unblock. Comments are just minor things.

apps/studio/components/interfaces/Auth/MfaAuthSettingsForm/MfaAuthSettingsForm.tsx
ignaciodob: Haha! I agree

PR #37886, opened 2025-08-12
Replace remaining uses of `yup` with `zod` in Studio

Saeris:

YES

We've been using both `yup` and `zod` and it doesn't make sense to have two validation libraries. Talking with @saltcod it appears that `zod` was picked by the team to be the preferred validation library for Studio, so this PR migrates the remaining schemas and associated forms to use the current patterns.

The result here should help to reduce Studio's bundle size by dropping a particularly heavy dependency.

Because most of the affected schemas and forms pertain to Platform specific UIs, I was unable to test them on my local, so we'll have to proceed by testing on staging. Like the other form migrations I've done recently, I would like to add associated UI tests to nail down our desired behavior. But before I add those I'll need to manually test these changes to ensure that the my use of `discriminatedUnion` works as expected for conditional form fields. I'll leave this marked as draft until there is satisfactory test coverage.

**EDIT**: Breaking this up into multiple PRs, keeping this open to track progress

Forms that have been updated and need testing:

Changes branched off into #37913:
- [x] `SignInForm`
- [x] `SignInMFAForm`
- [x] `SignInSSOForm`
- [x] `SignUpForm`

Changes branched off into #38139
- [ ] `DiskSizeConfigurationModal`

Changes branched off into #37909:
- [ ] `AuthProvidersForm`

Changes branched off into `TODO`
- [ ] `CustomDomainsConfigureHostname`
- [ ] `jwt-settings` (extracted to `CustomJWTModal`)
- [ ] `SmtpForm`
- [ ] `SiteURL`
- [ ] `TemplateEditor` in Email Templates

Changes branched off into `TODO`
- [ ] `BasicAuthSettingsForm`
- [ ] `MfaAuthSettingsForm`
- [ ] `ProtectedAuthSettingsForm`


apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersForm.types.ts
Saeris, line 1: moved to `AuthProvidersFormValidation.tsx`

apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersFormValidation.tsx
```diff
@@ -68,149 +118,135 @@ const PROVIDER_EMAIL = {
...
-}
-
-const smsProviderValidation = (config: ProjectAuthConfigData, provider: string) => {
-  return {
-    is: (EXTERNAL_PHONE_ENABLED: boolean, SMS_PROVIDER: string) => {
-      return EXTERNAL_PHONE_ENABLED && SMS_PROVIDER === provider && !config.HOOK_SEND_SMS_ENABLED
-    },
-  }
-}
+} as const satisfies Provider
```
Saeris, line 152: using `as const` here to narrow all of our strings to their literal values

`satisfies Provider` also ensures that we're adhering to the correct types for our form

apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersFormValidation.tsx
```diff
@@ -68,149 +118,135 @@ const PROVIDER_EMAIL = {
...
-      otherwise: (schema) => schema,
+  const twilioSchema = z.object({
+    EXTERNAL_PHONE_ENABLED: z.literal(true),
+    SMS_PROVIDER: z.literal(`twilio`),
+    SMS_TWILIO_ACCOUNT_SID: z.string({ required_error: 'Twilio Account SID is required' }),
+    SMS_TWILIO_AUTH_TOKEN: z.string({ required_error: 'Twilio Auth Token is required' }),
+    SMS_TWILIO_MESSAGE_SERVICE_SID: z.string({
+      required_error: 'Twilio Message Service SID is required',
     }),
+  })
```
Saeris, line 170: note here that we're using `SMS_PROVIDER` as a discriminator in zod's `discriminatedUnion` type

to make this a bit easier to maintain, we're separating out each sms provider into it's own object to declutter the `discriminatedUnion` array itself

apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersFormValidation.tsx
```diff
@@ -446,83 +481,70 @@ const EXTERNAL_PROVIDER_APPLE = {
...
+              console.log(e)
+              return false
+            }
+          },
+          {
+            message: 'Secret key expires in less than 7 days!',
+          }
+        ),
+    }),
+  ]),
```
Saeris, line 547: this was the only provider schema that had this strange dependency whereby a different error message would be shown if the Client ID was not provided

my understanding is that both the Client ID and Secret are required values here, so I opted to simplify this a bit. will update it if there is a compelling reason to

apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersFormValidation.tsx
```diff
@@ -564,25 +586,24 @@ const EXTERNAL_PROVIDER_AZURE = {
...
+      EXTERNAL_AZURE_ENABLED: z.literal(true),
+      EXTERNAL_AZURE_CLIENT_ID: z.string({
+        required_error: 'Application (client) ID is required',
+      }),
+      EXTERNAL_AZURE_SECRET: z.string({ required_error: 'Secret Value is required' }),
+      EXTERNAL_AZURE_URL: z.string().url('Must be a valid URL').optional(),
     }),
-    EXTERNAL_AZURE_URL: string().matches(urlRegex(), 'Must be a valid URL').optional(),
-  }),
+  ]),
```
Saeris, line 601: note that the schema itself doesn't determine what fields get rendered, that is determined by the "properties" key. so that being the case, we *should* be able to use `discriminatedUnion` like this to validate the "enabled/disabled" state of this form. but if it turns out that it causes an error, we should be able to just duplicate the keys and make them optional in the disabled state (where the previously saved values would be rendered but are not required)

apps/studio/components/interfaces/Auth/AuthProvidersForm/FormField.tsx
```diff
@@ -1,288 +1,276 @@
...
-  const [dateAsText, setDateAsText] = useState(
-    formValues[name] ? formatDate(new Date(formValues[name])) : ''
-  )
-
-  useEffect(() => {
-    if (properties.show && properties.show.key && !formValues[properties.show.key]) {
-      setFieldValue(name, '')
-      setDateAsText('')
-    }
-  }, [properties.show && properties.show.key && !formValues[properties.show.key]])
```
Saeris, line 41: this should no longer be necessary since we're switching to field management with `react-hook-form`

apps/studio/components/interfaces/Auth/AuthProvidersForm/FormField.tsx
```diff
@@ -1,288 +1,276 @@
...
-
-  if (properties.show) {
-    if (properties.show.matches) {
-      if (!properties.show.matches.includes(formValues[properties.show.key])) {
-        return null
-      }
-    } else if (!formValues[properties.show.key]) {
-      return null
-    }
-  }
```
Saeris, line 51: opted to lift this out of the field itself and into the `ProviderForm` component to simplify the props

apps/studio/components/interfaces/Auth/AuthProvidersForm/FormField.tsx
```diff
@@ -1,288 +1,276 @@
...
+                      className="text-foreground-lighter"
+                    />
+                  ) : null
                 }
-              }}
-            >
-              <span>Pick</span>
-            </DatePicker>
-          }
-        />
```
Saeris, line 98: this is the only datepicker + text input combo we have like this in the entire studio app. It seemed easier to me for wiring up this to the form to use the datepicker popover button text to display the current value rather than this dummy input

open to suggestions if someone has a better idea of how to do this, but for the moment I just want to get this in a testable state

apps/studio/components/interfaces/Auth/BasicAuthSettingsForm/BasicAuthSettingsForm.tsx
```diff
@@ -84,19 +83,19 @@ const BasicAuthSettingsForm = () => {
     }
   }, [authConfig])
 
-  const onSubmit = (values: any) => {
-    const payload = { ...values }
-    payload.DISABLE_SIGNUP = !values.DISABLE_SIGNUP
-    // The backend uses empty string to represent no required characters in the password
-    if (payload.PASSWORD_REQUIRED_CHARACTERS === NO_REQUIRED_CHARACTERS) {
-      payload.PASSWORD_REQUIRED_CHARACTERS = ''
-    }
```
Saeris, line 93: this key was never in the form's values so the condition would always be false

if it's necessary for the payload I can add back in the empty string

apps/studio/components/interfaces/Auth/SiteUrl/SiteUrl.tsx
```diff
@@ -41,26 +41,27 @@ const SiteUrl = () => {
...
   })
 
   useEffect(() => {
     if (authConfig && !isUpdatingSiteUrl) {
-      siteUrlForm.reset({
+      form.reset({
         SITE_URL: authConfig.SITE_URL || '',
       })
     }
   }, [authConfig, isUpdatingSiteUrl])
```
Saeris, line 57: might be able to remove this `useEffect` by leveraging the `key` property instead

apps/studio/components/interfaces/Auth/AuthProvidersForm/AuthProvidersFormValidation.tsx
Saeris: highlighted this because most of the providers follow this pattern with `discriminatedUnion` for conditional fields

Hallidayo:

PR #37888, opened 2025-08-12
feat: drop file to import data in Table Editor

charislam:
In the Table Editor for an empty table, make the entire grid area a
dropzone so that files can be dropped onto the grid area for importing.
Dropping a file opens the SpreadsheetImport side panel with the dropped
file already loaded.

https://github.com/user-attachments/assets/ce98a13a-5186-4897-a364-6db5b0385545

### DIFF
--- a/apps/studio/components/grid/components/grid/Grid.tsx
+++ b/apps/studio/components/grid/components/grid/Grid.tsx
@@ -1,8 +1,9 @@
-import { forwardRef, memo, Ref, useRef } from 'react'
+import { type DragEvent, forwardRef, memo, Ref, useRef, useState } from 'react'
 import DataGrid, { CalculatedColumn, DataGridHandle } from 'react-data-grid'
 
 import { handleCopyCell } from 'components/grid/SupabaseGrid.utils'
 import { formatForeignKeys } from 'components/interfaces/TableGridEditor/SidePanelEditor/ForeignKeySelector/ForeignKeySelector.utils'
+import { flagInvalidFileImport } from 'components/interfaces/TableGridEditor/SidePanelEditor/SpreadsheetImport/SpreadsheetImport'
 import AlertError from 'components/ui/AlertError'
 import { useForeignKeyConstraintsQuery } from 'data/database/foreign-key-constraints-query'
 import { ENTITY_TYPE } from 'data/entity-types/entity-type-constants'
@@ -53,6 +54,7 @@ export const Grid = memo(
     ) => {
       const tableEditorSnap = useTableEditorStateSnapshot()
       const snap = useTableEditorTableStateSnapshot()
+      const [isDraggedOver, setIsDraggedOver] = useState(false)
 
       const { data: org } = useSelectedOrganizationQuery()
       const { data: project } = useSelectedProjectQuery()
@@ -72,6 +74,7 @@ export const Grid = memo(
 
       const table = snap.table
       const tableEntityType = snap.originalTable?.entity_type
+      const isForeignTable = tableEntityType === ENTITY_TYPE.FOREIGN_TABLE
 
       const { mutate: sendEvent } = useSendEventMutation()
       const { data } = useForeignKeyConstraintsQuery({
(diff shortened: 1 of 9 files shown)


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SpreadsheetImport/SpreadsheetImport.utils.tsx
```diff
@@ -161,3 +170,57 @@ export const acceptedFileExtension = (file: any) => {
...
+export interface ProcessDroppedFileEventDetail {
+  file: File
+}
+
+export interface ProcessDroppedFileEvent extends CustomEvent<ProcessDroppedFileEventDetail> {
+  type: typeof PROCESS_DROPPED_FILE_EVENT
+}
+
+// Helper function to dispatch the custom event
+export function dispatchProcessDroppedFileEvent(file: File): void {
```
charislam: This is done with dispatched events rather than threading the file through the global state objects because passing the file through React state messes with later property accesses. I suspect that the Valtio proxy is messing with File prototype chain checks, but not sure.

joshenlim: mmm i see what you mean here - just checking to see if there's any other workaround for valtio

joshenlim: related discussion: https://github.com/pmndrs/valtio/discussions/739

joshenlim: okay TIL - will need to use `ref` from valtio to handle non plain objects - i'll push an update! 🙏

joshenlim (approved):
LGTM! Tested uploading CSVs via drop, button, and import rows dropdown 🙂 🙏

packages/common/telemetry-constants.ts
```diff
@@ -906,6 +906,26 @@ export interface ImportDataButtonClickedEvent {
...
+ */
+export interface ImportDataFileDroppedEvent {
+  action: 'import_data_file_dropped'
+  properties: {
+    /**
+     * The type of table the data is imported to.
+     * New Table means added when creating new table by clicking from New table sidebar,
+     * Existing Table means added to an existing table by going to the table and clicking from the green Insert button..
+     */
+    tableType: 'New Table' | 'Existing Table'
```
pamelachia: Seems that all the events import_data_file_dropped events sent so far says Existing Table but technically we are dropping from an empty table. So this isn't an accurate comment description and property. Can probably remove properties completely since it isn't relevant and it will always drop on an empty table.

packages/common/telemetry-constants.ts
```diff
@@ -906,6 +906,26 @@ export interface ImportDataButtonClickedEvent {
...
 
+/**
+ * User dropped a file into the import data dropzone on an empty table.
+ *
+ * @group Events
+ * @source studio
+ * @page /dashboard/project/{ref}/editor
+ */
+export interface ImportDataFileDroppedEvent {
+  action: 'import_data_file_dropped'
```
pamelachia: I think dropped can be misinterpreted as deleted or failed. Maybe we can do something like import_data_dropzone_file_added ? We also don't want to introduce too many verbs across our events because it might get confusing or synonym debates overtime. 😅

pamelachia (changes requested):

packages/common/telemetry-constants.ts
charislam: good point, didn't think about that 😄

apps/studio/components/grid/components/grid/Grid.tsx
```diff
@@ -72,8 +74,25 @@ export const Grid = memo(
...
       const { mutate: sendEvent } = useSendEventMutation()
+
+      const { isDraggedOver, onDragOver, onFileDrop } = useCsvFileDrop({
+        enabled: isTableEmpty && !isForeignTable,
+        onFileDropped: (file) => tableEditorSnap.onImportData(valtioRef(file)),
+        onTelemetryEvent: (eventName) => {
+          sendEvent({
+            action: eventName,
+            groups: {
+              project: project?.ref ?? 'Unknown',
```
seanoliver, line 89: It looks like we're already doing this in some places, but moving forward we should probably not send the literal string “unknown.” That turns missing data into a real category and will leak into breakdowns, cohorts, and joins. 

Instead, it'd be cleaner just to omit the property entirely or set it to null. Analysts can then use “has_property/is set” when they want only known values, and missing won’t pollute charts by default. This also avoids future conflicts if we ever introduce a legitimate “Unknown” enum value.

So something like ...(project.ref ? { project: project.ref } : {})

@pamelachia what do you think?

seanoliver (changes requested):

apps/studio/components/grid/components/grid/Grid.tsx
pamelachia: Yeap definitely. But let's handle that as a separate issue for us to handle. I believe frontend is also trying to clean this up.

pamelachia (approved):

seanoliver (approved):
Approving to unblock! Growth eng will take a task to clean up the fallback to "Unknown" thing! Shouldn't block this PR. 😄

PR #37894, opened 2025-08-13
Chore/setup sentry for www

joshenlim:
Related to FE-1807

## Context

Setting up Sentry for www to capture any API issues on the marketing site (e.g submission of Enterprise Demo form)
I think this set up might also be capturing client side errors though, so let me know where we might be able to skim it down - we've hit our Sentry quota again i think so it's quite hard to debug or verify what events are getting sent atm 😓 

Minimally verified that events are getting sent by checking the number of events that are getting dropped from the Sentry usage page 😬 But let me know if I might be missing something in here, or anywhere can be done better - just following the automatic wizard that Sentry provided

Have added the following 2 env vars into the www Vercel app as well:
- `NEXT_PUBLIC_SENTRY_DSN`
- `SENTRY_AUTH_TOKEN`

The next step will then be to set up Alerts for API issues so that we get pinged on Slack accordingly if any endpoints are failing for any reason


pamelachia (approved):

apps/www/app/api-v2/opt-out/[ref]/route.ts
```diff
@@ -78,7 +79,10 @@ export async function POST(req: NextRequest, props: { params: Promise<{ ref: str
       .from('manual_reports')
       .insert([{ project_ref: ref, reason, email }])
 
-    if (supabaseError) throw new Error(`Supabase error: ${supabaseError.message}`)
+    if (supabaseError) {
+      Sentry.captureException(supabaseError)
```
Comment (charislam) [inline apps/www/app/api-v2/opt-out/[ref]/route.ts]: I think this might be double-sending the error, since it's caught below and sent again

apps/www/app/api-v2/ticket-og/route.tsx
```diff
@@ -501,6 +501,7 @@ export async function GET(req: Request, res: Response) {
 
     return await fetch(`${STORAGE_URL}/og/${ticketType}/${username}.png?t=${NEW_TIMESTAMP}`)
   } catch (error: any) {
+    Sentry.captureException(error)
```
charislam, line 504: Since this one is an edge runtime function, it needs to manually flush Sentry events so that they all get sent before the process gets closed: see [related docs PR](https://github.com/supabase/supabase/pull/36377)

```suggestion
    Sentry.captureException(error)
    await Sentry.flush(2000)
```

apps/www/app/global-error.tsx
charislam, line 1: If we don't want to capture client-side errors we can just remove this file

apps/www/pages/_error.tsx
charislam, line 1: Same here, we could remove this to not waste quota on client-side

apps/www/app/layout.tsx
```diff
@@ -0,0 +1,12 @@
+export const metadata = {
+  title: 'Next.js',
+  description: 'Generated by Next.js',
```
charislam: Reminder to change this from the defaults, though do we even need this file? @fsansalvadore

apps/www/instrumentation-client.ts
charislam, line 1: I think we can also remove this and it will stop us sending client-side events at all

apps/www/app/layout.tsx
joshenlim: yeah was thinking if we can just remove this actually - these were all generated as part of the Sentry's set up wizard

apps/www/app/api-v2/opt-out/[ref]/route.ts
Comment (joshenlim) [reply apps/www/app/api-v2/opt-out/[ref]/route.ts]: oh great catch! sorry hahaha

apps/www/app/layout.tsx
joshenlim: i've removed this - reckon we dont need this

turbo.json
```diff
@@ -156,7 +157,8 @@
         "FORCE_ASSET_CDN",
         "ASSET_CDN_S3_ENDPOINT",
         "SITE_NAME",
-        "LUMA_API_KEY"
+        "LUMA_API_KEY",
+        "SENTRY_AUTH_TOKEN"
```
charislam: Nit: I don't think this needs to be listed here because it's never actually mentioned in code

charislam (approved):
Couldn't test properly without quota 😄 so I just ran it locally and waited for the number of rate-limited events to go up

turbo.json
joshenlim: ahh okay gotcha!

PR #37896, opened 2025-08-13
Bump temp API key expiry to an hour just for realtime inspector

joshenlim:
## Context

Realtime inspector now uses the new temporary API keys endpoint when the service role is selected, but the default expiry duration for the temp API keys is at 5 minutes which causes the realtime inspector to error out pretty quickly (it was an hour before)

Opting to bump the expiry for the temp API key just for the realtime inspector to an hour (which is also the max value that the endpoint supports)

## To test
- Double check that the realtime inspector is still working as per expected


charislam (approved):
- Tested using the realtime inspector for 5+ minutes and it keeps working
- But there is a case where it still doesn't work: if you use it for 5 + minutes, then leave the channel and try to join a new one, it fails with an auth error

joshenlim:
> if you use it for 5 + minutes, then leave the channel and try to join a new one, it fails with an auth error

that's odd - the key's expiry is meant to work for an hour now, not just 5 minutes 👀  checking

joshenlim:
i'm gonna push up a change which should refresh the temporary API token when clicking either "Start listening" or "Listen to channel", if the role selected is service role

should improve the auth token expiry issue (ensures to always use a new temp token whenever starting a refresh listen session), although this doesn't address the expiry duration (that'd be on auth end) Still checking on the 5 min duration issue locally

joshenlim:
hmm okay just tested again - seems like i could run the realtime inspector for more than 5 minutes which should be alright, will merge this soon then 🙂

PR #37898, opened 2025-08-13
remove effect for logs redirection

jordienr:
removes redirect effect to sned user to the latest visited logs page, 

since the unified logs effect overwrites it, i think its causing issues


joshenlim (approved):
LGTM! Verified on staging preview 🙂

PR #37823, opened 2025-08-11
ui consistency issues

SaxonF:
Some UI consistency work involving:
- Making use of PageLayout
- Using Scaffold components
- Using Table component in UI
- Some padding issues

PR #37901, opened 2025-08-13
Chore/refactor database publications page

joshenlim:
## Context

General UI tidy up, expanding on what Saxon's done already previously [here](https://github.com/supabase/supabase/pull/37823). This one's just focused on the Database Publications page.

Also addresses a point of confusion that was brought up in an internal friction log:
- At the top level page, we show the number of tables which have publications toggled on for
  image
- However, for `supabase_realtime_messages_publication`, when viewing the publication, i can't seem to find the table that was toggled on
- Realised that we aren't showing the full list of tables, as we're hiding tables from protected schemas
- So this PR also addresses this confusion by showing _all_ tables, but disabling the toggle for tables within protected schemas

## Changes involved
- Refactor Database Publications page to have publications on their own route, rather than controlled by client state
- Replace old table components in the publications page to the new ones from `ui`
- Use PageLayout in Database publications page
- Show _all_ tables within a publication, but show a disabled tooltip for tables which are in protected schemas (we were previously hiding them which was confusing)
  image

### DIFF
--- a/apps/studio/components/interfaces/Database/Publications/PublicationsList.tsx
+++ b/apps/studio/components/interfaces/Database/Publications/PublicationsList.tsx
@@ -166,7 +166,7 @@ export const PublicationsList = () => {
                             {x.tables === null
                               ? 'All tables'
                               : `${x.tables.length} ${
-                                  x.tables.length > 1 || x.tables.length == 0 ? 'tables' : 'table'
+                                  x.tables.length === 1 ? 'table' : 'tables'
                                 }`}
                           </Link>
                         </Button>
--- a/apps/studio/components/interfaces/Database/Publications/PublicationsList.tsx
+++ b/apps/studio/components/interfaces/Database/Publications/PublicationsList.tsx
@@ -1,34 +1,49 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
-import { noop } from 'lodash'
 import { AlertCircle, Search } from 'lucide-react'
+import Link from 'next/link'
 import { useState } from 'react'
 import { toast } from 'sonner'
-import { Button, Input, Toggle } from 'ui'
 
-import Table from 'components/to-be-cleaned/Table'
+import { useParams } from 'common'
+import AlertError from 'components/ui/AlertError'
 import InformationBox from 'components/ui/InformationBox'
 import NoSearchResults from 'components/ui/NoSearchResults'
 import { useDatabasePublicationsQuery } from 'data/database-publications/database-publications-query'
 import { useDatabasePublicationUpdateMutation } from 'data/database-publications/database-publications-update-mutation'
 import { useAsyncCheckProjectPermissions } from 'hooks/misc/useCheckPermissions'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
+import {
+  Button,
+  Card,
+  Switch,
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from 'ui'
+import { Input } from 'ui-patterns/DataInputs/Input'
 import ConfirmationModal from 'ui-patterns/Dialogs/ConfirmationModal'
-import PublicationSkeleton from './PublicationSkeleton'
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/Database/Publications/PublicationsList.tsx
```diff
@@ -97,60 +113,70 @@ export const PublicationsList = ({ onSelectPublication = noop }: PublicationsLis
...
-                                x.tables.length > 1 || x.tables.length == 0 ? 'tables' : 'table'
-                              }`}
+                    <TableCell>
+                      <div className="flex justify-end">
+                        <Button asChild type="default" style={{ paddingTop: 3, paddingBottom: 3 }}>
+                          <Link href={`/project/${ref}/database/publications/${x.id}`}>
+                            {x.tables === null
+                              ? 'All tables'
+                              : `${x.tables.length} ${
+                                  x.tables.length > 1 || x.tables.length == 0 ? 'tables' : 'table'
```
charislam:
```suggestion
                                  x.tables.length === 1 ? 'table' : 'tables'
```

tiniest, tiniest nit, but it's easier to read

apps/studio/pages/project/[ref]/database/publications/[id].tsx
```diff
@@ -41,14 +22,7 @@ const DatabasePublications: NextPageWithLayout = () => {
...
           <FormHeader title="Database Publications" />
-          {selectedPublication === undefined ? (
-            <PublicationsList onSelectPublication={setSelectedPublicationId} />
-          ) : (
-            <PublicationsTables
-              selectedPublication={selectedPublication}
-              onSelectBack={() => setSelectedPublicationId(undefined)}
-            />
-          )}
+          <PublicationsTables />
```
Comment (charislam) [inline apps/studio/pages/project/[ref]/database/publications/[id].tsx]: Minor nit: would it be better to get the id in this component, and pass it as a prop to `PublicationsTables`? Since we know from the filename that this component must have an ID, but it's not clear in `PublicationsTables` that it must be used on a page with an ID.

apps/studio/components/interfaces/Database/Publications/PublicationsTables.tsx
```diff
@@ -95,38 +87,32 @@ export const PublicationsTables = ({
...
+                    <TableHead>Description</TableHead>
+                    {/* 
+                      We've disabled All tables toggle for publications. 
+                      See https://github.com/supabase/supabase/pull/7233. 
+                    */}
+                    <TableHead />
+                  </TableRow>
+                </TableHeader>
+                <TableBody>
+                  {!!selectedPublication &&
```
charislam: This is the current state for invalid publication ID, should we have a more descriptive UI message?

Image

charislam (approved):
- ✅ Checked the publications list and tables list
- Some minor nits, otherwise looks good!

apps/studio/components/interfaces/Database/Publications/PublicationsList.tsx
joshenlim: ooooh yes 💯 agree here

apps/studio/components/interfaces/Database/Publications/PublicationsTables.tsx
joshenlim: oh yes great catch!

apps/studio/pages/project/[ref]/database/publications/[id].tsx
Comment (joshenlim) [reply apps/studio/pages/project/[ref]/database/publications/[id].tsx]: hmm great point here - am thinking for this case specifically because `PublicationsTable` is a specific component thats only used here it should be okay, will avoid the need for prop drilling too)

if say this component ends up being used in more than one place then yup might be better to have `id` passed as a prop

apps/studio/components/interfaces/Database/Publications/PublicationsTables.tsx
joshenlim: will add this empty state
image

joshenlim:
Unit tests are failing due to some issues with coveralls, unrelated to our unit tests i think

PR #37630, opened 2025-08-01
feat(studio): scoped personal access tokens ui

kemaldotearth:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

This PR introduces a more advanced UI for the scoped personal access tokens found in the users account area. This PR introduces.

## What is the current behavior?

The current behaviour is limited in the sense that users cannot set things like token expiry or granular permissions. 

## What is the new behavior?

This PR aims to bring in the kind of control our access tokens previously lacked. In this we address the following:

- Tidy up of the Access Token area. This is more inline with future UI direction.
- An improved Create New Token flow.
- Ability to configure an Expiry date for access tokens.
- Ability to scope your access token's by organizations or projects.
- Ability to set granular access for a variety of permissions.
- Once a token has been created and configured, we also have a panel to show you how everything is configured.

## Additional context

🚧 Awaiting API endpoint updates to be able to get this in there.

PR #37908, opened 2025-08-13
feat(studio): Add expiration date and last used to personal access tokens

kamilogorek:
FE part of https://linear.app/supabase/issue/API-605/add-optional-expiration-date-to-newly-created-pats

Requires https://github.com/supabase/infrastructure/pull/24939 to be deployed first.

Reworked table to use not deprecated `Table` component along the way.

It would be nice if we could use custom calendar dropdown later, but this was beyond me 😢 

image

image


joshenlim (approved):
Tested on staging preview (payload sent when creating PAT includes expires at if an expiry is selected, and doesnt have that property if "never" is selected)

LGTM! Approving to unblock but will mark as draft for now since its blocked on API changes

kemaldotearth:
Jumping on this to migrate some of the scoped PAT UI that we [already designed](https://github.com/supabase/supabase/pull/37630) for expiry stuff in the prototype. Ties in nicely :)

joshenlim (approved):
Tested on staging preview and locally! 🙂

PR #37913, opened 2025-08-13
[Replace `yup`] Refactor Sign In/Up forms to use shadcn and replace `yup` with `zod`

Saeris:
Splitting this off from #37886 so that we can incrementally review the changes.

This migrates all the sign-in/up forms that were still using `yup` over to `zod` schemas, removing our dependency on `yup-password`, and updates their associated forms to follow our convention of `react-hook-form` paired with shadcn based components.

The following forms need to be tested on staging because they are platform specific:

 - [ ] SignInForm
 - [ ] SignInMFAForm
 - [ ] SignInSSOForm
 - [ ] SignUpForm


apps/studio/components/interfaces/SignIn/SignInForm.tsx
```diff
@@ -106,76 +116,78 @@ const SignInForm = () => {
...
+              setCaptchaToken(token)
+            }}
+            onExpire={() => {
+              setCaptchaToken(null)
+            }}
+          />
+        </div>
+
+        <LastSignInWrapper type="email">
+          <Button block form={formId} htmlType="submit" size="large" loading={isSubmitting}>
```
Saeris, line 186: note that `loading` will implicitly cover `disabled`, so there is no need to provide both if they are the same condition

apps/studio/components/interfaces/SignIn/SignUpForm.tsx
```diff
@@ -1,36 +1,60 @@
...
+      // Basic password validation - you can enhance this based on your requirements
+      const hasUppercase = /[A-Z]/.test(password)
+      const hasLowercase = /[a-z]/.test(password)
+      const hasNumber = /[0-9]/.test(password)
+      const hasSpecialChar = /[!@#$%^&*()_+\-=\[\]{};`':"\\|,.<>\/?]/.test(password)
+      const isLongEnough = password.length >= 8
+
+      return hasUppercase && hasLowercase && hasNumber && hasSpecialChar && isLongEnough
+    }, 'Password must contain at least 8 characters, including uppercase, lowercase, number, and special character'),
 })
```
Saeris: note that this was copied from the `ResetPasswordForm` which had previously migrated off of `yup-password`

apps/studio/lib/schemas.test.ts
Saeris, line 1: deleted since we're no longer using `yup-password`

apps/studio/lib/schemas.ts
Saeris, line 1: deleted since we're no longer using `yup-password`

apps/studio/components/interfaces/SignIn/SignInForm.tsx
```diff
@@ -1,34 +1,44 @@
...
+import { Button, Form_Shadcn_, FormControl_Shadcn_, FormField_Shadcn_, Input_Shadcn_ } from 'ui'
 import { LastSignInWrapper } from './LastSignInWrapper'
+import { FormItemLayout } from 'ui-patterns/form/FormItemLayout/FormItemLayout'
 
-const signInSchema = object({
-  email: string().email('Must be a valid email').required('Email is required'),
-  password: string().required('Password is required'),
+const schema = z.object({
+  email: z.string({ required_error: 'Email is required' }).email('Must be a valid email'),
+  password: z.string({ required_error: 'Password is required' }),
```
joshenlim: these validations don't seem to be working - able to submit the form with 0 characters for password

only the email validation seems to be working (RE Must be a valid email)

joshenlim: will need to use `.min()` instead of `required_error`

tho TIL required_error only checks for undefined values 👀 ([ref](https://github.com/colinhacks/zod/issues/3114#issuecomment-1879776472))

Saeris: good to know! I will switch all the instances of `required_error` to `message` instead. I think it's just down to how the resolver for react-hook-form processes zod errors

iirc the original schema here was this simple too, so it makes sense to me that there may be different implicit behaviors between yup and zod

apps/studio/components/interfaces/SignIn/SignUpForm.tsx
```diff
@@ -47,6 +49,7 @@ const formId = 'sign-up-form'
 export const SignUpForm = () => {
   const captchaRef = useRef<HCaptcha>(null)
   const [showConditions, setShowConditions] = useState(false)
+  const [isSubmitting, setIsSubmitting] = useState(false)
```
Saeris: we should be able to get this value from `form.formstate.isSubmitting` instead of creating a new `useSttate` just FYI

joshenlim: oh yes! great catch hahaha will update

joshenlim:
JFYI for transparency: the unit tests in this PR are somehow failing for modal related ones, despite the changes in this PR having unrelated changes. I think its a red herring as other PRs seems to be all okay - so going to merge this in regardless

joshenlim (approved):
Tested on staging preview, form validations for:
- Logging in
- SSO
- Forgot password
- Signing up

PR #37670, opened 2025-08-04
feat: aws marketplace billing in dashboard

ignaciodob:
This PR adds support in Studio for Organizations billed via the AWS Marketplace.

### Changes
- Refactors the computation of the `managed_by` organization property in `organizations-query.ts` to include the AWS Marketplace slug.
- Adds the AWS icon to `PartnerIcon` as well as support for the appropriate tooltip.
- Extends the `PartnerManagedResource` component to support more than one partner.
- Adds the `useAwsRedirectQuery` hook to fetch the AWS redirect URL used in the `PartnerManagedResource` component
- Disables the Cost Control section for AWS orgs.

### Preparation - Create an AWS Org:
- Apply `thomas/billing-aws-marketplace-onboarding` to your backend repo.
- Apply `thomas/billing-1188-aws-marketplace-onboarding-page` to your frontend repo
- Head over to `http://localhost:8082/aws-marketplace-onboarding?buyer_id=96b6b1ee-1a99-45b9-8ea6-8ce26b2a02c3`
- Create an Organization

### Testing
- Head over to `http://localhost:8082/org/_/billing` and pick the AWS org
- Assert that the top banner says `This organization is managed by AWS Marketplace` and that the `Manage` buttons redirects you to AWS.
- Assert that the Spend Cap section is not available
- Assert that the `Past Invoices` are managed by AWS and that the `View invoices` button redirects you to AWS.
- Assert that the `Payment Methods` section is hidden.
- Assert that the `Billing Address & Tax ID` section is hidden

PR #37919, opened 2025-08-13
chore: add BillingManagedBy constant

ignaciodob:
### Changes
- Adds a `BILLING_MANAGED_BY` constant to more easily add support for different billing partners in the future (see https://github.com/supabase/supabase/pull/37670).
- Adds a `BillingManagedBy` type
- Modifies `PartnerManagedResourceProps` to use the new constant

### Testing
- Head over to `http://localhost:8082/org/_/billing` and pick a Vercel-managed organization
- Assert that the `This organization is managed by Vercel.` banner appears at the top of the page
- Assert that the `Invoices` section redirects to Vercel
- Head over to `http://localhost:8082/org/_/billing` and pick a AWS-managed organization
- Assert that the `This organization is managed by AWS Marketplace.` banner appears at the top of the page

### DIFF
--- a/apps/studio/components/ui/PartnerManagedResource.tsx
+++ b/apps/studio/components/ui/PartnerManagedResource.tsx
@@ -3,9 +3,10 @@ import { ExternalLink } from 'lucide-react'
 import { useVercelRedirectQuery } from 'data/integrations/vercel-redirect-query'
 import { Alert_Shadcn_, AlertTitle_Shadcn_, Button } from 'ui'
 import PartnerIcon from './PartnerIcon'
+import { MANAGED_BY, ManagedBy } from 'lib/constants/infrastructure'
 
 interface PartnerManagedResourceProps {
-  partner: 'vercel-marketplace' | 'aws-marketplace'
+  partner: ManagedBy
   resource: string
   cta?: {
     installationId?: string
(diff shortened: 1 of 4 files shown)


apps/studio/lib/constants/infrastructure.ts
```diff
@@ -11,6 +11,14 @@ export const AWS_REGIONS_DEFAULT =
 // TO DO, change default to US region for prod
 export const FLY_REGIONS_DEFAULT = FLY_REGIONS.SOUTHEAST_ASIA
 
+export const BILLING_MANAGED_BY = {
```
ecktoteckto: I'm wondering whether we should either
a) use the same term we use in the backend, i.e. "Billing Partner"
b) use the term "Manged By". Since it's not only about billing being managed by the partner, but also e.g. creating projects (in the case of Vercel)

Just trying to not use too many terms for the same thing to avoid confusion over time.

ignaciodob: Good question! I'm thinking that using "Managed By" makes much more sense - I was missing that key context on creating projects. I'll make that change 🙇

apps/studio/components/ui/PartnerManagedResource.tsx
```diff
@@ -3,9 +3,10 @@ import { ExternalLink } from 'lucide-react'
 import { useVercelRedirectQuery } from 'data/integrations/vercel-redirect-query'
 import { Alert_Shadcn_, AlertTitle_Shadcn_, Button } from 'ui'
 import PartnerIcon from './PartnerIcon'
+import { MANAGED_BY, ManagedBy } from 'lib/constants/infrastructure'
 
 interface PartnerManagedResourceProps {
-  partner: 'vercel-marketplace' | 'aws-marketplace'
+  partner: ManagedBy
```
ecktoteckto, line 9: Maybe `managedBy`

ecktoteckto (approved):

apps/studio/components/ui/PartnerManagedResource.tsx
ignaciodob: Addressed in https://github.com/supabase/supabase/pull/37670, thank you!

PR #37921, opened 2025-08-14
feat: enable/disable dashboard features

alaister:
This PR adds client side (non-API-driven) support for disabling dashboard features. It extends the existing `useIsFeatureEnabled()` hook.

## API Design
- Allow for control from either or both the API side + Frontend side.
- Fully type-safe with proper descriptions to allow for an easy UX. e.g.:
    Screenshot 2025-08-13 at 22 20 35
- Minimal new patterns required.

## How to add support for enabling/disabling a new feature in the dashboard:
1. Add a name and description to `enabled-features.schema.json`

    The naming convention should be `group:feature`. This PR includes the examples: `profile:show_email` and `profile:update`. These names **can** conflict with existing feature names returned by the API. This is so we can have API _or_ Frontend control over a feature.
2. Add the default state (true/false for on/off) in `enabled-features.json`. Generally we should be defaulting to true for hosted studio.
3. Implement `useIsFeatureEnabled()` to conditionally check if a feature is enabled.
    For a single check:
    ```ts
   const profileShowEmailEnabled = useIsFeatureEnabled('profile:show_email')
    ```
    For checking multiple features at a time:
    ```ts
    const { profileShowEmail: profileShowEmailEnabled, profileUpdate: profileUpdateEnabled } =
      useIsFeatureEnabled(['profile:show_email', 'profile:update'])
    ```


charislam (approved):
Tested locally toggling profile:email off and worked as expected!

PR #37924, opened 2025-08-14
Table consistency

SaxonF:
We are at a point where we have a number of different table types across the dashboard. This PR updates a number of project and org tables as a step towards unifying. 

image
image
image
image


apps/studio/components/interfaces/Database/Extensions/Extensions.tsx
joshenlim, line 1: Just to check - this page's padding (or max width?) seems to be inconsistent with for e.g enumerated types, if this is intended? Noticed the "large" prop though

image

image

apps/studio/components/interfaces/Integrations/GraphQL/GraphiQLTab.tsx
```diff
@@ -98,7 +99,26 @@ export const GraphiQLTab = () => {
             </p>
           </div>
 
-          <ExtensionCard extension={pgGraphqlExtension} />
+          <Card>
```
joshenlim: realised we can remove this - we'll never run into this state if the extension is not enabled, will hop on to clean

apps/studio/components/interfaces/Database/Extensions/Extensions.tsx
```diff
@@ -73,49 +72,43 @@ const Extensions = () => {
...
-              ))}
-            </div>
-          </div>
-        </div>
+        <GenericSkeletonLoader />
       ) : (
         <>
-          {extensions.length === 0 && (
+          {extensions.length === 0 ? (
             <NoSearchResults
```
joshenlim: thinking the empty state could also show the headers

image

joshenlim (approved):
Pushed up some minor fixes but generally LGTM! Tested on staging preview for the following

- Database pages
- Integrations -> GraphiQL
- Integrations -> Vault
- Organization members

dnywh (approved):
Looks good! Only minor visual changes from what I can see, all intentional. E.g. headings.

PR #37927, opened 2025-08-14
feat(studio): update chart highlight colour

kemaldotearth:

YES

This is a speculative PR on improving the visibility of chart highlighting function. It switches out our signature green for a more contrast friendly opaque white. While the green looks great (and on brand) on more sparse charts, for dense information it's quite tricky to parse.

See the image below. Whenever we have congested bar charts, the highlight is barely visible. At a glance not visible at all.

Screenshot 2025-08-13 at 15 22 44

## What is the new behavior?

As an experiment, set the value to white. The highlight on condensed bar charts looks a lot more obvious.

Screenshot 2025-08-13 at 15 22 29

## Additional context

This is an open question. If it's something we're happy to change, I think we should make a call based on readability. Open to suggestion here :)


jordienr (approved):

apps/studio/components/ui/Charts/ComposedChart.tsx
dnywh, line 1: Not the same as your hex colours but you could alternatively lean into the variables? E.g.

```jsx
stroke={isDarkMode ? 'hsl(var(--foreground-light))' : 'hsl(var(--brand-default))'}
fill={isDarkMode ? 'hsl(var(--foreground-light))' : 'hsl(var(--brand-default))'}
```

...looks like:

| Light | Dark |
| --- | --- |
| Image | Image |

dnywh (approved):
Nice one! One small suggestion but not blocking.

apps/studio/components/ui/Charts/ComposedChart.tsx
dnywh: Or `foreground-default` for a stronger white.

kemaldotearth: @dnywh good shout with the variables. I used a darker green for light mode too as we had similar issues 😓

PR #37930, opened 2025-08-14
add etl logs

jordienr:


joshenlim (approved):
Verified on staging preview that ETL logs are not available unless the project has access to the replications stuff

PR #37937, opened 2025-08-14
Fix so can change prod branch

SaxonF:
Right now it always sets Github branch to default even if there is an existing connection. A user could update branch but it would appear as if it wasn't saving.


joshenlim (approved):
Verified changes on staging preview! 🙏

PR #37938, opened 2025-08-14
fix: write queries in ai assistant

charislam:
Some read-only errors from running SQL in the AI Assistant aren't being
properly detected, so the action item to review and confirm running the
SQL isn't showing up. Add matching against another variant of read-only
error message that can show up.


joshenlim (approved):
LGTM! 😄

PR #37831, opened 2025-08-11
search snippet list fix

SaxonF:
Searches snippets directly within SQL Editor Nav. This also resolves the issue of snippet items missing a contextual menu when searching.

PR #37945, opened 2025-08-15
Refactor search list in sql editor to have context menu

joshenlim:
## Context

Addresses an internal feedback in which while searching snippets in the SQL Editor, the items do not have context menus like the `SQLEditorNav` in the SQL Editor. This PR thus addresses that by getting `SearchList` to use the same TreeView component as per `SQLEditorNav`, so that the `SQLEditorTreeViewItem` component which has the context menu components can be shared between both UIs

Saxon also had a similar PR [here](https://github.com/supabase/supabase/pull/37831), which opts to deprecate the SearchList component and leverage the `SQLEditorNav`, but I reckon it might be a better UX to have a clear visual difference between the main navigation and the search list
- Search list is a flat list of items to make it easier to find snippets within folders (Similar to how MacOS finder does)
  - It might get confusing if we show a flat list in the same main navigation as if might look like folders are gone

## Changes involved
- Snippet items when searching in the SQL Editor will have the same context menu as per the main navigation in SQL Editor
  - I've deliberate left out the move snippets action in the search list since its a flat list UI, and moving snippets in here might be confusing
  - Also left out all folder actions in SearchList since its not applicable
- Upsert and Delete content will also invalidate the infinite list RQ for content, which is what the `SearchList` component is using
- Include if the snippet belongs to shared or private in the Search list
  image

## To test
- [ ] Verify that all actions within the ContextMenu in `SQLEditorNav` still works
- [ ] Verify that all actions within the ContextMenu in `SearchList` works


apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/DeleteSnippetsModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+}) => {
+  const router = useRouter()
+  const { ref: projectRef, id } = useParams()
+  const tabs = useTabsStateSnapshot()
+  const snapV2 = useSqlEditorV2StateSnapshot()
+
+  const postDeleteCleanup = (ids: string[]) => {
+    const existingSnippetIds = Object.keys(snapV2.snippets).filter((x) => !ids.includes(x))
+
+    if (existingSnippetIds.length === 0) {
```
ivasilov: I'm not sure if you have to do this. If all tabs are removed, the SQL editor will just open a new tab. Opening the first snippet in the state is also weird.

apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
```diff
@@ -41,56 +54,122 @@ export const SearchList = ({ search }: SearchListProps) => {
...
-      )}
-      href={`/project/${ref}/sql/${snippet.id}`}
-    >
-      <SQL_ICON
-        size={16}
-        strokeWidth={1.5}
-        className="w-5 h-5 -ml-0.5 transition-colors fill-foreground-muted group-aria-selected:fill-foreground"
+    <>
+      <div className="flex flex-col flex-grow">
+        {isLoading ? (
```
ivasilov: this `isLoading` should be replaced with the one from `useContentCountQuery`.

apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SQLEditorTreeViewItem.tsx
```diff
@@ -77,7 +82,7 @@ export const SQLEditorTreeViewItem = ({
   onSelectShare,
   onSelectUnshare,
   onSelectDownload,
-  onSelectDuplicate,
+  // onSelectDuplicate,
```
ivasilov: Delete these comments.

ivasilov (approved):

apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
joshenlim: ooooh great catch!

apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/DeleteSnippetsModal.tsx
joshenlim: hmm good point this - will revise

joshenlim: yeah great catch - i think this logic was before we had tabs in place, so no longer applicable

PR #37946, opened 2025-08-15
FE-1813: Audit Logs Improvements

jordienr:
Users are asking to be able to filter audit logs hourly.

- Use LogsDatePicker in Org Audit Logs
- Use LogsDatePicker in Account Audit Logs
- Audit Logs Fix: move filters out of audit logs error state so they can be updated if the error is range exceeded.

- LogsDatePicker: Update time range reset icon so its not confused with a close icon
- LogsDatePicker: Adds hideWarnings prop 
- LogsDatePicker: Fixes "Today" button not scrolling to today

### DIFF
--- a/apps/studio/components/interfaces/Organization/AuditLogs/AuditLogs.tsx
+++ b/apps/studio/components/interfaces/Organization/AuditLogs/AuditLogs.tsx
@@ -131,14 +131,6 @@ const AuditLogs = () => {
     <>
       <ScaffoldContainerLegacy>
         <div className="space-y-4 flex flex-col">
-          {isLoading && (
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Organization/AuditLogs/AuditLogs.tsx
```diff
@@ -147,6 +139,77 @@ const AuditLogs = () => {
             </div>
           )}
 
+          <div className="flex items-center justify-between">
```
ivasilov, line 134: The loading state is above the bar
Image

ivasilov (approved):

PR #37955, opened 2025-08-15
feat: Use v2 endpoint for `getProjects`

lforst:
Based on https://github.com/supabase/infrastructure/pull/24797

### DIFF
--- a/apps/docs/lib/fetch/projects.ts
+++ b/apps/docs/lib/fetch/projects.ts
@@ -8,9 +8,14 @@ const projectKeys = {
 }
 
 export async function getProjects(signal?: AbortSignal) {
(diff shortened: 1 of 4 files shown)


charislam (commented):
Looking at API PR it looks like limit defaults to 100, so we probably want to put in proper pagination handling on the frontend, otherwise it will be confusing to users with > 100 projects if we never fetch the rest and also never tell them that

lforst:
> Looking at API PR it looks like limit defaults to 100, so we probably want to put in proper pagination handling on the frontend, otherwise it will be confusing to users with > 100 projects if we never fetch the rest and also never tell them that

I also just realized that. Found a supid workaround to fix the spec for now: https://github.com/supabase/infrastructure/pull/25000 😂

PR #37956, opened 2025-08-15
Fix studio dynamic route mock setup

Saeris:
The intent of this is to fix for tests that are failing in CI but pass locally.

First I need to test assumptions about the setup code and how it runs in our CI container versus local development environments.

The failing tests appear to all have a root cause of `useParams()` returning `undefined` for dynamic route segment parameters, and the cause of this would be an incorrect configuration of `next-router-mock`. The possible cause of this is that the globs we're using to build this list of paths behaves differently in CI.

### DIFF
--- a/apps/studio/tests/vitestSetup.ts
+++ b/apps/studio/tests/vitestSetup.ts
@@ -18,6 +18,7 @@ import { mswServer } from './lib/msw'
 beforeAll(() => {
   mswServer.listen({ onUnhandledRequest: `error` })
   vi.mock('next/router', () => require('next-router-mock'))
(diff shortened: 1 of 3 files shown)


apps/studio/tests/vitestSetup.ts
```diff
@@ -34,10 +35,6 @@ beforeAll(() => {
       }),
     }
   })
-
-  vi.mock('next/compat/router', () => require('next-router-mock'))
-
-  routerMock.useParser(createDynamicRouteParser(['/projects/[ref]']))
```
Saeris, line 40: want to make sure that this isn't accidentally resetting the list, so I'm removing it as this should be covered in the glob utility anyways

PR #37962, opened 2025-08-16
Replace deprecated `Alert` in `/claim-project` page

Saeris:
This is a simple refactor replacing the deprecated `Alert` component in the claim project page with the equivalent shadcn based version. 

Implementation is based on a similar warning alert in `/project/[ref]/functions/[functionSlug]` route, which has a slightly different visual appearance. See screenshot for comparison (top: original, bottom: new)

image

### DIFF
--- a/apps/studio/pages/claim-project.tsx
+++ b/apps/studio/pages/claim-project.tsx
@@ -1,3 +1,6 @@
+import Head from 'next/head'
+import { useMemo, useState } from 'react'
+
 import { useParams } from 'common'
 import { ProjectClaimBenefits } from 'components/interfaces/Organization/ProjectClaim/benefits'
 import { ProjectClaimChooseOrg } from 'components/interfaces/Organization/ProjectClaim/choose-org'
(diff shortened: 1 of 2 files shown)


apps/studio/pages/claim-project.tsx
```diff
@@ -62,20 +62,22 @@ const ClaimProjectPage: NextPageWithLayout = () => {
...
-            <p>Please retry your claim request from the requesting app</p>
-            {errorProjectClaim != undefined && (
-              <p className="mt-2">Error: {errorProjectClaim?.message}</p>
-            )}
-            {errorRequester != undefined && (
-              <p className="mt-2">Error: {errorRequester?.message}</p>
-            )}
-            <p>Please go back to the requesting app and try again.</p>
-          </Alert>
+          <Alert_Shadcn_ variant="warning">
```
joshenlim: this one's a recent decision with the design team but we've decided to stick to using `Admonition` for all "alert" related UIs sorry 🙏  i'll hop on this PR to make the changes 🙂

joshenlim (approved):
Verified locally minimally!

apps/studio/pages/claim-project.tsx
joshenlim: Although we're planning to consolidate Admonition and Alert_Shadcn so this might change in the near future 😅 🙏

PR #37965, opened 2025-08-16
Replace deprecated components on `/authorize` page

Saeris:
This refactors `/authorize` to replace deprecated components with the recommended shadcn based equivalents. Follows the patterns in `BasicAuthSettingsForm` for the use of `<Card>` components.

I used a `/sandbox` page in development to visually compare basic implementations of the replaced components, however, this should still be inspected for possible errors on staging.

Note that there isn't a "success" alert/admonition, so for sake of consistency with the existing code, this opts to use `<Alet_Shadcn_>` paired with `<CheckIcon />` for the success modal, which will have differences in appearance by necessity. If there is a preferred reference implementation, please let me know!

### DIFF
--- a/apps/studio/pages/authorize.tsx
+++ b/apps/studio/pages/authorize.tsx
@@ -1,5 +1,4 @@
 import dayjs from 'dayjs'
-import { AlertCircle } from 'lucide-react'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { useEffect, useState } from 'react'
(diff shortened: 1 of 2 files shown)


apps/studio/pages/authorize.tsx
```diff
@@ -93,42 +101,51 @@ const APIAuthorizationPage: NextPageWithLayout = () => {
...
-        <div className="px-8 py-6">
-          <Alert withIcon variant="warning" title="Missing authorization ID">
-            Please provide a valid authorization ID in the URL
-          </Alert>
-        </div>
-      </FormPanel>
+      <Card>
+        <CardHeader>Authorization for API access</CardHeader>
+        <CardContent>
+          <Alert_Shadcn_ variant="warning">
```
joshenlim: i know i mentioned in the previous PR that we're using Admonition - but design is gonna look into consolidating alert_shadcn and admonition, so we'll just go ahead with using alert_shadcn for now 🙂

joshenlim (approved):
Verified changes locally!

PR #37968, opened 2025-08-16
add missing onRouterTransitionStart

jordienr:
fixes sentry warning

### DIFF
--- a/apps/studio/instrumentation-client.ts
+++ b/apps/studio/instrumentation-client.ts
@@ -152,3 +152,7 @@ function standardiseRouterUrl(url: string) {
 
   return finalUrl
 }
+
+// This export will instrument router navigations, and is only relevant if you enable tracing.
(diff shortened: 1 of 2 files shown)


apps/studio/instrumentation-client.ts
```diff
@@ -152,3 +152,7 @@ function standardiseRouterUrl(url: string) {
 
   return finalUrl
 }
+
+// This export will instrument router navigations, and is only relevant if you enable tracing.
+// `captureRouterTransitionStart` is available from SDK version 9.12.0 onwards
```
lforst: I feel like the comment about versioning is not necessary.

lforst (approved):

apps/studio/instrumentation-client.ts
jordienr:
```suggestion
```

saltcod (approved):

PR #37982, opened 2025-08-17
Refactor deprecated components in `/project/_/settings/general` route

Saeris:
The goal of this refactor is to replace the majority of deprecated components and patterns within the general project settings page as possible. Notably it skips refactoring of modals in this route to avoid conflicts with other in-flight work awaiting review.

Generally, all of the refactors follow the patterns of use for the `<Card>` components using `BasicAuthSettingsForm` as the example. I decided that it would be best to touch multiple files here to improve consistency in appearance for the entire page. There are some minor visual differences compared to production, such as the Delete Project card, which now has a layout matching other cards on this page. Some headings were also altered for visual consistency.

Besides the updated form for renaming the project, interaction and data fetching logic remains unchanged for the affected components. That said, please check for anything I might have missed!

PR #37996, opened 2025-08-18
Update `/project/_/settings/api` page for consistency with other Project Settings pages

Saeris:
Following on the work in #37982 and #37965, this updates the API Settings page to follow all of the latest patterns for using `<Card>` components for layout. As this was a simpler page, there were no changes to the existing form logic, everything here is just to apply visual consistency with the other pages.

### DIFF
--- a/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
+++ b/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
@@ -37,7 +37,9 @@ import {
   Skeleton,
   Switch,
   WarningIcon,
+  cn,
 } from 'ui'
+import { GenericSkeletonLoader } from 'ui-patterns'
 import { Admonition } from 'ui-patterns/admonition'
 import { FormItemLayout } from 'ui-patterns/form/FormItemLayout/FormItemLayout'
 import {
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -168,275 +166,271 @@ export const PostgrestConfig = () => {
...
+                                        <Button asChild type="default">
+                                          <Link href={`/project/${projectRef}/database/extensions`}>
+                                            Disable the pg_graphql extension
+                                          </Link>
+                                        </Button>
+                                      </>
+                                    )}
+                                  </>
+                                }
+                              />
```
Saeris: for my own reference, do we want to align around *always* using Admonition, or *always* prefer the Alert_shadcn_ primitives? I understand that with Admonition, we don't have as much control over it, such as we're forced into a particular icon per `type` value. However, they also visually differ in colors for `warning` and `destructive` variants, which may read differently depending on whether we're rending inside of a Card or not, so they aren't strictly equivalent at present

would like to be filled in on the consensus here so that I can be more intentional about using one or the other

joshenlim: hmm great question here cc @supabase/design - any thoughts or guidelines as to which we should be following? I've been defaulting to `<Admonition>` if there might be a preference

joshenlim:
pushed up some small fixes, but all looks good nonetheless! will go ahead to merge this PR once the actions pass - and I'll also keep you updated RE the decision between Admonition and Alert_Shadcn 🙂 🙏

joshenlim (approved):

PR #38002, opened 2025-08-18
unified logs: rm org plan check

jordienr:


joshenlim (approved):
Changes LGTM - checked changes locally that unified logs preview is no longer gated by plan

PR #38012, opened 2025-08-18
fix: Revert permissions change in Vercel integrations component.

ivasilov:
This PR reverts this change https://github.com/supabase/supabase/pull/37899/files#diff-1be98329c7b0bc643acf274d192af3447a59c3e4e6dcf62a774c491531d86d89. The component in question `VercelSection` is used both in project and org pages while using `useAsyncCheckProjectPermissions` which requires project ref in the URL (which is not available in org pages). This bug shows up as never ending loading state in the org integrations page.


joshenlim (approved):
Verified changes on staging preview that vercel integration is no longer stuck in a loading state

PR #38019, opened 2025-08-18
feat: disable database features

alaister:
Adds support for enabling/disabling:

- Database replication page
- Database roles page
- Toggle for enabling/disabling RLS when creating a table


packages/common/enabled-features/enabled-features.json
```diff
@@ -23,5 +26,7 @@
 
   "project_homepage:show_all_client_libraries": true,
   "project_homepage:show_instance_size": true,
-  "project_homepage:show_examples": true
+  "project_homepage:show_examples": true,
+
+  "table_editor:creation_enable_rls_toggle": true
```
joshenlim: probably can generalize to enable_rls_toggle - cause this needs to cover the RLS button in the Table Editor header too

apps/studio/components/layouts/DatabaseLayout/DatabaseMenu.utils.tsx
```diff
@@ -64,30 +72,36 @@ export const generateDatabaseMenu = (
...
+                  ]
+                : [
+                    {
+                      name: 'Replication',
+                      key: 'replication',
+                      url: `/project/${ref}/database/replication`,
+                      label: 'Coming Soon',
+                      items: [],
+                    },
+                  ]),
```
joshenlim: unrelated but can be simplied? difference between these 2 seems to just be the labels

joshenlim (approved):
Verified on staging preview that features with flags added in the PR are all still status quo

Verified locally that features with flags added in the PR can be toggled off

PR #38020, opened 2025-08-18
feat: disable users features

alaister:
Adds support for enabling/disabling on auth users page:
- Create user modal
- Provider filter
- Sort by email
- Sort by phone
- Filtering by verified / unverified users


apps/studio/components/interfaces/Auth/Users/AddUserDropdown.tsx
```diff
@@ -58,25 +61,27 @@ const AddUserDropdown = () => {
...
+                  <p>Create new user</p>
+                </DropdownMenuItem>
+              </TooltipTrigger>
+              {!canCreateUsers && (
+                <TooltipContent side="left">
+                  You need additional permissions to create users
+                </TooltipContent>
+              )}
+            </Tooltip>
+          )}
```
joshenlim: this one seems incorrect? we need to gate sending invites instead?

joshenlim (approved):
Verified on staging preview that features with flags added are status quo

Verified locally that features with flags added can be toggled off

PR #38021, opened 2025-08-19
Project list updates

SaxonF:
image

Updates organisation project list page to make use of PageHeader within an addition of a table view via view selector toggle. This brings project list more aligned with other project pages (actions on right, search/filter on left with page header).


joshenlim (approved):

PR #38033, opened 2025-08-19
storage explorer: fix validation in folder creation

jordienr:
- does not allow creating folders with same name but different casing

CleanShot 2025-08-19 at 14 58 34@2x


saltcod (approved):
checked on staging, looks good!

PR #38042, opened 2025-08-19
feat(profile): add self-serve editing of username and primary email

cstockton:
Users can now update `username` and choose a `primary_email` directly in the
Dashboard profile editor.

Changes:
- Add `username` and `primary_email` fields to the profile form.
- Introduce `Select_Shadcn_` dropdown listing the user’s identities (via
  `useProfileIdentitiesQuery`) for selecting a primary email.
- Default `primary_email` selection is derived from the current profile.
- Extend form submission to patch `username` and `primary_email` to the
  profile API.
- Wire zod schema and react-hook-form defaults for new fields.

Rationale:
- Resolves repeated customer issues where changing identity email left the UI stale (old email shown in avatar, invoices, etc.).
- Will allow disabling identity -> profile propagation on sign-in by making changes explicit and user-controlled.
- Reduces duplicate-email login conflicts by requiring deliberate updates.


apps/studio/components/interfaces/Account/Preferences/ProfileInformation.tsx
```diff
@@ -95,6 +125,75 @@ export const ProfileInformation = () => {
...
+                        disabled={profile?.is_sso_user}
+                      >
+                        <SelectTrigger_Shadcn_ className="col-span-8">
+                          <SelectValue_Shadcn_ placeholder="Select primary email" />
+                        </SelectTrigger_Shadcn_>
+                        <SelectContent_Shadcn_ className="col-span-8">
+                          {isIdentitiesSuccess &&
+                            identities.map((identity) => {
+                              const email = identity.identity_data?.email
+                              if (!email) return null
```
joshenlim: could be a filter then? since we're ignoring identities with no emails

joshenlim:
Pushed up some refactors to use the new `Card` component here, and also fix an issue with the primary email select dropdown, in which multiple duplicate emails will be shown in the UI if there's different identities sharing the same email address

Just one bug i noticed which is updating the `username` field doesn't persist, despite the value being sent properly to the PATCH profile request

joshenlim (approved):

JFYI though - for accounts logged in via GH, we use the `username` to derive the URL of the user profile's image, so if the `username` gets updated for GH accounts, then the user profile image will no longer work. Just wondering if we might want to prevent updating username for GH accounts specifically

joshenlim (approved):
All okay now! profile image in user dropdown now depends on identities rather than the user's profile - so users can update their username without affecting the profile image (specifically just for GH accounts)

PR #38060, opened 2025-08-20
chore/disable-more-sections-03

joshenlim:
## Context

Branches out of https://github.com/supabase/supabase/pull/38058 - adds support for enabling/disabling the following UI within account preferences
- Analytics and Marketing section
- Account deletion section

Removes the following flags and reverts the UI changes
- `table_editor:enable_rls_toggle` -> Table Editor should have RLS badge + RLS checkbox in create table side panel
-  `authentication:policies` -> Auth + Realtime policies pages should not longer be flagged

## To test
Locally, test that these can be toggled off
Staging, test that everything is status quo


saltcod (approved):
Checked staging and locally, and enables/disables as expected!

PR #38073, opened 2025-08-20
Chore/custom content hook

joshenlim:
## Context

Adding support for extending the dashboard with defined content. Attempting to do so in a similar fashion as per how we're doing it with the `useIsFeatureEnabled` hook - so hopefully no new patterns here.

The added `useCustomContent` hook is fully type safe with proper descriptions, although I'll probably need some second opinions on how we can do it better (I doubt what I've got set up is optimal atm)

image
image

## How to extend the dashboard with defined content
1. Add a name and description, as well as its expected shape within `custom-content.schema.json`
    - The naming convention should follow that of `useIsFeatureEnabled` (e.g `group:content`) 
    - This PR includes an example: `organization:legal_documents`

2. Add the default state in `custom-content.json`. Generally we should be defaulting to `null` for hosted studio
    - I've added a `custom-content.sample.json` as a reference
    - If the provided state is not `null`, then the dashboard will render the custom content provided 

3. Will also need to add the expected shape in `CustomContent.types.ts`
    - I'm wondering if there's a better way to do this though, but its cause the value for each content is either `null` or something else
    - Type name needs to be the camel case version of the content name (e.g for `organization:legal_documents` -> `organizationLegalDocuments`)

4. Call `useCustomContent` to check if there's any defined custom content. I've stuck to calling multiple contents at a time for simplicity atm (also prefer to just have one way to use the hook)
    - e.g `const { organizationLegalDocuments } = useCustomContent(['organization:legal_documents'])` 

## To test
- [ ] Verify that if custom content defined for `organization:legal_documents` is provided, it'll override the UI for "Legal Documents" under the Organization

### DIFF
--- a/apps/studio/hooks/custom-content/useCustomContent.ts
+++ b/apps/studio/hooks/custom-content/useCustomContent.ts
@@ -1,0 +1,38 @@
+import customContentRaw from './custom-content.json'
+
+// [Joshen] See if we can de-dupe any of the logic here with enabled-features
+// For now just getting something working going first
+
+const customContentStaticObj = customContentRaw as Omit<typeof customContentRaw, '$schema'>
+type CustomContent = keyof typeof customContentStaticObj
+
+type SnakeToCamelCase<S extends string> = S extends `${infer First}_${infer Rest}`
+  ? `${First}${SnakeToCamelCase<Capitalize<Rest>>}`
+  : S
+
+type CustomContentToCamelCase<S extends CustomContent> = S extends `${infer P}:${infer R}`
+  ? `${SnakeToCamelCase<P>}${Capitalize<SnakeToCamelCase<R>>}`
+  : SnakeToCamelCase<S>
+
+function contentToCamelCase(feature: CustomContent) {
+  return feature
+    .replace(/:/g, '_')
+    .split('_')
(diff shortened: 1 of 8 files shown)


apps/studio/hooks/custom-content/useCustomContent.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+const useCustomContent = <T extends CustomContent[]>(
+  contents: T
+): {
+  [key in CustomContentToCamelCase<T[number]>]:
+    | (typeof customContentStaticObj)[CustomContent]
+    | CustomContentTypes[CustomContentToCamelCase<T[number]>]
+} => {
+  // [Joshen] Running into some TS errors without the `as` here - must be overlooking something super simple
```
jordienr, line 34: you'll need to validate the content json with zod or something so typescript can infer the type before returning it.

However I added 2 tests to make sure we don't break this which should keep us covered.

jordienr (approved):
works fine, tested locally, added 2 tests to make sure we don't break the useCustomContent hook.

staaldraad (approved):

apps/studio/hooks/custom-content/useCustomContent.ts
joshenlim: gotcha on this thanks so much!!

PR #38087, opened 2025-08-21
Chore/custom content logs default query

joshenlim:
## Context
Branches out of https://github.com/supabase/supabase/pull/38086, support extending content of the default query when landing on the logs explorer. Refer to custom-content.sample.json for reference on implementation.

## To test
- [ ] Verify that if custom content for logs default query is defined, it'll replace the default query when landing on logs/explorer page


jordienr (approved):
works!

staaldraad (approved):

PR #38090, opened 2025-08-21
Chore/custom content connect UI frameworks

joshenlim:
## Context
Branches out of https://github.com/supabase/supabase/pull/38087, support extending content of the frameworks within the Connect UI. Refer to `custom-content.sample.json` for reference on implementation.

## Changes involved
- Handling of frameworks in the Connect UI will be slightly different:
  - Currently we're handling the content by declaring individual files, then rendering them through `ConnectTabContent` where the file contents will be dynamically imported
  - For custom frameworks, because we're defining the content via JSON, opted to create a separate file to render instead for separation of concerns `ConnectTabContentCustom`
  - The structure of the framework JSON differs slightly between hosted and custom content
    - Custom content has a `files` parameter which is an array of objects that's used to render the UI
    - Opted to extend the existing `ConnectionType` type to handle that and use it in `CustomContent.types.ts`, so that we'd be aware of any potential breaking changes if we ever decide to change the shape there

- `files` parameter in the JSON schema will also support handlebar notation for dynamic content, specifically only for these parameters (these are already what's passed into `ContentFile` when rendering content from files:
   - `{{apiUrl}}`: The API URL of the project
   - `{{anonKey}}`: The anonymous key of the project (if still using legacy API keys)
   - `{{publishableKey}}`: The publishable API key of the project (if using new API keys)
   - Added TS docs within `CustomContent.types.ts`

- Note that current implementation only supports 1 level of dropdown here for simplicity for now. We can reinvestigate if there needs to more dropdown support (like how on hosted we've got dynamic number of dropdowns depending on the framework selected)
  image

## To test
- [ ] Verify that if custom content for connect frameworks is defined, it'll replace the framework tabs in the Connect UI


apps/studio/hooks/custom-content/custom-content.json
```diff
@@ -5,5 +5,40 @@
 
   "project_homepage:example_projects": null,
 
-  "logs:default_query": null
+  "logs:default_query": null,
+
+  "connect:frameworks": {
```
jordienr: need to set this to null

jordienr (approved):
pushed a change to set `connect:frameworks` to null in `custom-content.json`. Looks good!

staaldraad (approved):

PR #38094, opened 2025-08-21
feat(ui): add shadow to tables on inner scroll

kemaldotearth:

YES

This adds a shadow to indicate the tables still have more content to scroll. Previously we weren't showing any sort of indicator the table becomes scrollable on smaller viewports. The shadow appears either side when there's a scroll.

| Before | After |
|--------|--------|
| Screenshot 2025-08-21 at 13 33 37 | Screenshot 2025-08-21 at 13 33 56 | 

No shadows.

## What is the new behavior?

Shadowy coolness.


packages/ui/src/components/shadcn/ui/table.tsx
```diff
@@ -6,12 +6,78 @@ interface TableProps extends React.HTMLAttributes<HTMLTableElement> {
   containerProps?: React.HTMLAttributes<HTMLDivElement>
 }
 
+// Custom hook to detect horizontal scrolling
```
jordienr: no need for the comment here

jordienr (approved):

packages/ui/src/components/ShadowScrollArea/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const ShadowScrollArea = React.forwardRef<HTMLDivElement, ShadowScrollAreaProps>(
+  ({ className, children, ...props }, ref) => {
+    const containerRef = React.useRef<HTMLDivElement>(null)
+    const { hasHorizontalScroll, canScrollLeft, canScrollRight } = useHorizontalScroll(containerRef)
+
+    return (
+      <div className="relative">
+        <div
+          className={cn(
```
dnywh, line 37: Thoughts on different shadow strengths based on `light` or `dark` modes? The current shadow on `light` feels too heavy and on `dark` feels too soft IMO. I assume you were trying to find a value that worked for both.

Here’s an experiment using the `dark:` variant so we can use different values (e.g. `/50` and `/5`) instead of `/40` for both:

```jsx
 <div
          className={cn(
            'absolute inset-0 pointer-events-none z-10',
            'before:absolute before:top-0 before:right-0 before:bottom-0 before:w-6 before:bg-gradient-to-l before:from-black/5 dark:before:from-black/50 before:to-transparent before:opacity-0 before:transition-all before:duration-400 before:easing-[0.24, 0.25, 0.05, 1]',
            'after:absolute after:top-0 after:left-0 after:bottom-0 after:w-6 after:bg-gradient-to-r after:from-black/5 dark:after:from-black/50 after:to-transparent after:opacity-0 after:transition-all after:duration-400 after:easing-[0.24, 0.25, 0.05, 1]',
            hasHorizontalScroll && 'hover:before:opacity-100 hover:after:opacity-100',
            canScrollRight && 'before:opacity-100',
            canScrollLeft && 'after:opacity-100'
          )}
        />
```

| Before | After |
| --- | --- |
| Image | Image |
| Image | Image |

dnywh (approved):
Looks good! One small subjective suggestion.

packages/ui/src/components/ShadowScrollArea/index.tsx
kemaldotearth: Good catch with light mode @dnywh, this should now be taken care of in the updated component 🙇

joshenlim (approved):
Generally LGTM - verified on staging preview checking all tables across the dashboard

JFYI though - i did run into some very subtle issue where the shadows are showing on both ends of the table despite scrolling to the left end (specifically noticed this happening on database/tables, database/indexes and database/publications). Super odd though as we're not able to reproduce this consistently (seems to only be happening on my end)

Alright for now although this might be more obvious on mobile, non-blocking nonetheless, can address separately if needed 🙂

PR #38096, opened 2025-08-21
fix(studio): column sizes for access tokens

kemaldotearth:

YES

This fixes the column widths for the access tokens table so they don't collapse despite small strings for dates.

| Before | After |
|--------|--------|
| Screenshot 2025-08-21 at 14 03 17 | Screenshot 2025-08-21 at 14 03 45 |


ivasilov (changes requested):
Image
The options (three dots) dropdown menu is missing on large screens. It only shows up partially when the screen is smaller (< 1024px, I think)

ivasilov (approved):

PR #38122, opened 2025-08-22
fix wrapper name row ui

jordienr:
- truncate long names
- add tooltip so full name can still be seen

https://github.com/user-attachments/assets/39152cc6-9a52-4c3e-9f88-34a9cfad9394

## BEFORE
CleanShot 2025-08-22 at 11 56 56@2x

## AFTER
CleanShot 2025-08-22 at 11 57 16@2x


raulb (approved):
Nice! (API engineer here, not sure if that's ok with you all 😜)

kevcodez (approved):

PR #38126, opened 2025-08-22
Small style fix

joshenlim:
## Context

Small style fix to `ColumnType` component in `WrapperDynamicColumns` - happening because `WrapperTableEditor` is currently using the old `Form` component which has style overrides for `input[type="text"]`

Before:
image

After:
image


jordienr (approved):
tested in preview lgtm

PR #38127, opened 2025-08-22
Auto select smart region as default in new project page

joshenlim:
## Context

We recently introduced smart regions for project creation (as opposed to choosing a specific country to spin up your project from) although this isn't live yet, only on staging and local.

Changes here are to default the region selection to the recommended smart region as per defined by API, instead of the recommended specific country


alaister (approved):
LGTM!

PR #38129, opened 2025-08-22
Reports consolidation 1: Edge Functions

jordienr:
Goal here is simplify the configuration files for the report pages.

Principles I'm following:
- Consolidate report config into one type (+ remove type casts) 
- No use of dynamic hooks / query factories
- No breaking rules of hooks
- Chart components should be simple and configurable and not do checks for specific metrics
- No data transformations in components, data provider returns data as needed, all transformations live in the same place
- Complex logic (like data transformation) should be easy to extract into functions and tested

## In this PR
- Fixes Edge Functions report
- Update the Edge Functions Report to the new config
- Adds tests
- Simplify data fetching for each metric (theres one function for each metric now)
- Add Report Settings 
CleanShot 2025-08-27 at 16 51 48@2x

## To test
- Create an edge function or two
- Run it a few times (test button / curl) 
- go to edge functions report
- Data in report makes sense

In the next few PRs I will move the rest of the reports and add advanced filtering.


apps/studio/data/reports/v2/edge-functions.config.ts
jordienr, line 1: This file replaces edgefn-charts and edgefn-query

Also uses the new types for report configurations.

I will use these new types on the rest of the charts in next PRs.

apps/studio/pages/project/[ref]/reports/edge-functions.tsx
```diff
@@ -14,38 +14,43 @@ import DefaultLayout from 'components/layouts/DefaultLayout'
...
+export default EdgeFunctionsReportV2
 
 const EdgeFunctionsUsage = () => {
   const { ref } = useParams()
   const { data: functions, isLoading: isLoadingFunctions } = useEdgeFunctionsQuery({
     projectRef: ref,
   })
+
+  const chartSyncId = `edge-functions-${ref}`
+  const _ = useChartHoverState(chartSyncId)
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/reports/edge-functions.tsx]: tiny nit - could just call the hook if we're not using the returned value
```suggestion
useChartHoverState(chartSyncId)
```

apps/studio/pages/project/[ref]/reports/edge-functions.tsx
```diff
@@ -214,17 +225,21 @@ const EdgeFunctionsUsage = () => {
...
-                endDate={selectedDateRange?.period_end?.date}
-                updateDateRange={updateDateRange}
-                functionIds={functionIds}
-              />
-            ))}
+            reportConfig
+              .filter((report) => !report.hide)
+              .map((report, i) => (
+                <ReportChartV2
+                  key={`${report.id}-${i}`}
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/reports/edge-functions.tsx]: tiny nit - do we need `i` here? wondering if `id` is unique enough to be the key

apps/studio/components/ui/Charts/ComposedChart.tsx
```diff
@@ -512,3 +515,5 @@ export default function ComposedChart({
     </div>
   )
 }
+
+export default ComposedChart
```
joshenlim: we're avoiding export defaults actually - so probably can keep the above as `export const ComposedChart = () => ...` ([ref](https://github.com/supabase/supabase/pull/37952))

apps/studio/components/ui/Charts/ChartHeader.tsx
```diff
@@ -202,6 +214,24 @@ const ChartHeader = ({
...
+                icon={<SquareTerminal />}
+                onClick={() =>
+                  router.push(`/project/${ref}/logs/explorer?q=${encodeURIComponent(sql)}`)
+                }
+              />
+            </TooltipTrigger>
+            <TooltipContent side="top" align="center">
+              Open in Log Explorer
+            </TooltipContent>
+          </Tooltip>
```
joshenlim: Could use ButtonTooltip, asChild with Link? 
preference for Link whenever possible instead of router.push

apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
```diff
@@ -0,0 +1,186 @@
...
+            No data available for the selected time range
+          </div>
+        </div>
+      </Panel>
+    )
+  }
+
+  return (
+    <>
+      <Panel
```
joshenlim: Just a heads up too - we're leaning into the `<Card>` component now which (iinw) replaces the Panel component (we should add a deprecated warning in there too)

apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  // LOADING STATE
+  if (isLoadingChart || isLoading) {
+    return (
+      <Panel
+        title={<p className="text-sm">{report.label}</p>}
+        className={cn('h-[260px]', className)}
+      >
+        <div className="flex items-center justify-center h-full">
+          <div className="text-sm text-foreground-light">Loading...</div>
```
joshenlim: tiny tiny nit: using p instead of div for text
likewise for below too error state & empty state

apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
```diff
@@ -0,0 +1,186 @@
...
+    report.availableIn === undefined || (orgPlanId && report.availableIn.includes(orgPlanId))
+
+  const canFetch = orgPlanId !== undefined
+
+  const {
+    data: queryResult,
+    isLoading: isLoadingChart,
+    error,
+  } = useQuery(
+    ['report-v2', report.id, projectRef, startDate, endDate, interval, functionIds],
```
joshenlim: this is just a nit and not sure if it'll be cleaner

for RQ keys we typically start with the project ref first, and wrap configuration related params into a single object
so e.g `['projects', projectRef, 'report-v2', report.id, { startData, endDate, interval, functionIds }]`

starting with the project ref first allows us to invalidate all project related data easily if we need to by invalidating `['projects', projectRef]` (although we don't do that anywhere in the dashboard atm)

apps/studio/components/interfaces/Reports/v2/ReportChartUpsell.tsx
```diff
@@ -0,0 +1,92 @@
...
+        >
+          <Link href={`/org/${orgSlug}/billing?panel=subscriptionPlan&source=reports`}>
+            Upgrade to{' '}
+            <span className="capitalize">
+              {!!report.availableIn?.length ? report.availableIn[0] : 'Pro'}
+            </span>
+          </Link>
+        </Button>
+      </div>
+      <div className="absolute top-0 left-0 w-full h-full z-0"></div>
```
joshenlim: just to check - this one looks unnecessary?

apps/studio/components/interfaces/Reports/v2/ReportChartUpsell.tsx
```diff
@@ -0,0 +1,92 @@
...
+      demo: Math.floor(Math.pow(1.25, index) * 10),
+      max_demo: 1000,
+    }))
+
+  const getDemoChartData = () =>
+    new Array(20).fill(0).map((_, index) => ({
+      period_start: new Date(startDate).getTime() + index * 1000,
+      demo: Math.floor(Math.random() * 10) + 1,
+      max_demo: 1000,
+    }))
```
joshenlim, line 26: should we be removing the demo data?

apps/studio/components/interfaces/Reports/v2/ReportChartUpsell.tsx
```diff
@@ -0,0 +1,92 @@
...
+          ]}
+          label={''}
+          startDate={startDate}
+          endDate={endDate}
+          interval={interval}
+          data={demoData as any}
+          isLoading={false}
+          highlightedValue={0}
+          updateDateRange={() => {}}
+        />
```
joshenlim: also wondering if the params here needs updating like updateDateRange, isLoading, label, etc

apps/studio/components/interfaces/Reports/v2/ReportChartUpsell.tsx
jordienr: Its used in the chart background. Could make it an svg/image but it animates when the user hovers the upgrade button :) 

Image

apps/studio/components/interfaces/Reports/v2/ReportChartUpsell.tsx
jordienr: Not needed, just being used visually for the background. 

I have removed the props for startDate/endDate/interval, they are not necessary really.

apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
jordienr: fixed, also moved the empty, loading and error states to inside the Panel Content to prevent layout shift.

apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
jordienr: 🤝

apps/studio/pages/project/[ref]/reports/edge-functions.tsx
Comment (jordienr) [reply apps/studio/pages/project/[ref]/reports/edge-functions.tsx]: should be yeah some copypasted stuff from the old compo 👍

apps/studio/components/interfaces/Reports/v2/ReportChartUpsell.tsx
joshenlim: ohhhhh sorry! this is all good then hahaha

saltcod (commented):
This looks great on the UI side. Leaving approval for Joshen on the code side.

joshenlim (approved):
LGTM! Verified changes on staging preview by checking the edge functions reports page - let me know if i might have missed anything else! 🙏

PR #38075, opened 2025-08-20
Refactor deprecated components in `/project/_/database/settings` route

Saeris:
Similar to the work in #37982 and #37996, this PR gives the Database Settings page a visual consistency facelift.

Unlike those other PRs, this one takes some design liberties in cleaning up the presentation of the affected components. Rendering logic for all of them remain largely the same except where deprecated hooks were replaced or conditional states were reorganized, which may result in slightly different behavior.

Please verify that these changes look and behave correctly on staging. Associated modals will be refactored in a separate PR.

Here are some screenshots:

### Connection Pooling states

**enable addon**
image

**loading**
image

**error**
image

**success**
image

**disabled**
image

### SSL (w/ error state)

image

### DIsk Management Moved Notification

image

### Disk Management

**disabled**
image

**enabled**
image

### Network Restrictions

**loading**
image

**error**
image

**allow all**
image

**disallow all**
image

**list**
image

### Network bans

**loading**
image

**error**
image

**list**
image

**empty**
image

PR #38137, opened 2025-08-22
Refactor deprecated components in Database Settings -> Network Restrictions modals

Saeris:
Building on #38075 this PR consists of refactors specific to the Network Restrictions section modals for Allow All, Disallow All, Add Restriction and Remove Restriction

This should not be reviewed until #38075 is merged, since this is branched off from that PR. It'll remain in draft status until then.

Because three of these modals are merely confirmations, the logic driving them is virtually unchanged, however they now contain their triggers and check their own permissions in order to be self-contained.

For the Add Restriction modal, it has been updated to use `react-hook-form` and `zod`, so the validation logic has been moved into a zod schema. It likely needs additional scrutiny out of the four refactored modals.


Hallidayo:

PR #38192, opened 2025-08-25
Swap to use paginated projects endpoint

joshenlim:
## Context

This is a pre-requisite to supporting pagination for the projects page - swapping to use the new Version 2 endpoint for GET projects which involves a change in the API response.

This PR is purely just updating all the places in the dashboard that use the `useProjectsQuery()` hook, such that we're accessing the projects via the `projects` parameter from the `data` of the RQ - no change visually.
e.g `const { data: projects = [] } = useProjectsQuery()` -> 
```
const { data } = useProjectsQuery()
const projects = data?.projects ?? []
```

The limit on the API end is 100 results returned on the new GET projects endpoint - which wouldn't affect a majority of users (if not all)

Security review got tagged due to changes in `interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel`

## To test

- [ ] Verify some of the pages that use `useProjectsQuery` RQ hook isn't crashing (e.g projects page, account audit logs, etc, org invite member, etc)
  - If the TS checks are passing I think it's generally a indication that nothing was missed out


jordienr (approved):
lgtm / dont see anything that could lead to a crash

staaldraad (approved):

PR #37860, opened 2025-08-12
Part 2 of swapping useCheckPermissions with useAsyncCheckProjectPermissions

joshenlim:
## Context

Following up from the previous [PR](https://github.com/supabase/supabase/pull/37751), copying the PR description for ease of reference:

We'll be favouring the use of useAsyncCheckProjectPermissions over useCheckPermissions as it'll allow us to access the loading state of the permissions. We should then also factor that in to prevent any premature "no permissions" UI rendering

Opting to do this in smaller PRs since its a bit nuanced

We can only swap useCheckPermissions with useAsyncCheckProjectPermissions for project permissions, organization permissions will still need to use the former for now until we introduce a new useAsyncCheckOrgPermissions
Also need to add loading states where missing for pages that are currently prematurely showing the no perms UI

## Changes involved

- Swap `useCheckPermissions` with `useAsyncCheckProjectPermissions` in more files
  - This should address premature "perms denied" UI issues for branching, and realtime inspector
  - All others either do not have a perms denied UI, or their disabled states are behind button tooltips which are fine as they are
- Refactor a couple more files to directly `export` their components instead of `export default`
- Refactor to use DropdownMenuItemTooltip in TablesList

## Affected UI
Probably just need a quick skim on these pages, mainly just to make sure that there's (1) No incorrect "perms denied" UI, if you're the owner of the org you should have access to everything (2) No pre-mature "perms denied UI" (can test by refreshing on that page)

- Database
  - Tables/Columns
  - Backups
  - Extensions
  - Functions
  - Triggers
  - Publications
  - Roles
- Integrations -> Database Webhooks
- Realtime Inspector
- Branching UI
- Connect UI
- Reports
- Log Drains
- Settings -> JWT keys
- Edge Functions -> Code

PR #38201, opened 2025-08-25
[useAsyncCheckProjectPermissions] Add permission states to SQLEditorMenu

Saeris:
Following on the work started in #37751, #37860, and #37899, this updates the `+` menu on the SQL Editor page with a tooltip and a loading state indicator for the `Create a new snippet` menu item

screenshot of the combined loading state, disabled while loading/disallowed, and tooltip when disallowed:

image

### DIFF
--- a/apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
+++ b/apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
@@ -1,17 +1,18 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { useDebounce } from '@uidotdev/usehooks'
 import { LOCAL_STORAGE_KEYS, useParams } from 'common'
-import { FilePlus, FolderPlus, Plus, X } from 'lucide-react'
+import { Loader2, FilePlus, FolderPlus, Plus, X } from 'lucide-react'
 import { useRouter } from 'next/router'
 import { useEffect, useState } from 'react'
 import { toast } from 'sonner'
 
-import { useCheckPermissions } from 'hooks/misc/useCheckPermissions'
+import { useAsyncCheckProjectPermissions } from 'hooks/misc/useCheckPermissions'
 import { useLocalStorage } from 'hooks/misc/useLocalStorage'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
 import { useProfile } from 'lib/profile'
 import { getAppStateSnapshot } from 'state/app-state'
 import { useSqlEditorV2StateSnapshot } from 'state/sql-editor-v2'
+import { DropdownMenuItemTooltip } from 'components/ui/DropdownMenuItemTooltip'
 import {
   Button,
   DropdownMenu,
@@ -48,10 +49,11 @@ export const SQLEditorMenu = () => {
   const appState = getAppStateSnapshot()
   const debouncedSearch = useDebounce(search, 500)
 
-  const canCreateSQLSnippet = useCheckPermissions(PermissionAction.CREATE, 'user_content', {
-    resource: { type: 'sql', owner_id: profile?.id },
-    subject: { id: profile?.id },
-  })
+  const { can: canCreateSQLSnippet, isLoading: isCheckingPermissions } =
+    useAsyncCheckProjectPermissions(PermissionAction.CREATE, 'user_content', {
+      resource: { type: 'sql', owner_id: profile?.id },
+      subject: { id: profile?.id },
+    })
 
   const createNewFolder = () => {
     if (!ref) return console.error('Project ref is required')
@@ -64,9 +66,6 @@ export const SQLEditorMenu = () => {
     if (!ref) return console.error('Project ref is required')
     if (!project) return console.error('Project is required')
     if (!profile) return console.error('Profile is required')
-    if (!canCreateSQLSnippet) {
-      return toast('Your queries will not be saved as you do not have sufficient permissions')
-    }
     try {
       router.push(`/project/${ref}/sql/new?skip=true`)
       setSearch('')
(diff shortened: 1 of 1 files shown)


apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
```diff
@@ -140,10 +139,25 @@ export const SQLEditorMenu = () => {
...
+                      : 'Your do not have sufficient permissions to create new queries',
+                  },
+                }}
+                onClick={() => handleNewQuery()}
+              >
+                {isCheckingPermissions ? (
+                  <Loader2 className="animate-spin" strokeWidth={1.5} size={16} />
+                ) : (
+                  <FilePlus size={14} />
+                )}
```
Saeris, line 158: @joshenlim lmk if this loading state works for you. I didn't have an example on hand for menu items, but felt like this would be fine to show a spinner icon here

apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
```diff
@@ -48,10 +49,11 @@ export const SQLEditorMenu = () => {
   const appState = getAppStateSnapshot()
   const debouncedSearch = useDebounce(search, 500)
```
Saeris, line 50: unrelated, but could we maybe replace this and the associated `useEffect` with `useDeferredValue` instead for conditionally showing the search results? keeping my eye out for ways we can remove `useEffect` if it isn't necessary

https://react.dev/reference/react/useDeferredValue#examples

joshenlim: ooh interesting - does `useDeferredValue` work the same as a debounce? I dont see it accepting any other parameter apart from the value. I just tried it locally and it doesn't seem to be debouncing - but happy to explore this separately, could pop a demo PR 🙂

apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
```diff
@@ -140,10 +139,25 @@ export const SQLEditorMenu = () => {
               />
             </DropdownMenuTrigger>
             <DropdownMenuContent align="end" side="bottom" className="w-48">
-              <DropdownMenuItem className="gap-x-2" onClick={() => handleNewQuery()}>
-                <FilePlus size={14} />
+              <DropdownMenuItemTooltip
+                className="gap-x-2"
+                disabled={isCheckingPermissions || !canCreateSQLSnippet}
```
joshenlim, line 144: hmm is there a reason for the change in UI behaviour for this one?
we should still let users create a new snippet even if they don't have permissions so that they can still run the query (even if the query is not saved) - hence why the original behaviour is just a toast

apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
```diff
@@ -140,10 +139,25 @@ export const SQLEditorMenu = () => {
...
+                tooltip={{
+                  content: {
+                    text: canCreateSQLSnippet
+                      ? undefined
+                      : 'Your do not have sufficient permissions to create new queries',
+                  },
+                }}
+                onClick={() => handleNewQuery()}
+              >
+                {isCheckingPermissions ? (
```
joshenlim, line 154: so based on the above - i reckon we don't need a loading state here

we can probably update the toast message instead to reflect accordingly
e.g if perms are still loading - "This query will not be saved as permissions are loading and we're unable to check if you have sufficient permissions" (probably something a bit less verbose)

apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
Saeris: previously the behavior was to return an error toast so my assumption was that it blocked creating a new snippet, was that not the case?

apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
```diff
@@ -64,9 +66,6 @@ export const SQLEditorMenu = () => {
     if (!ref) return console.error('Project ref is required')
     if (!project) return console.error('Project is required')
     if (!profile) return console.error('Profile is required')
-    if (!canCreateSQLSnippet) {
-      return toast('Your queries will not be saved as you do not have sufficient permissions')
-    }
```
Saeris, line 69: looks to me like this line is supposed to block navigating to the new sql snippet route

apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
Saeris: No the article I linked points out that it isn't the same as a debounce, rather it's a means to tell React to throw away in-progress renders if the value changes before the render completes

It would make more sense if we were debouncing the search fetch request, but in this case it appears that we're just conditionally rendering the search results list based on whether there is a search value or not. We might still want to debounce the fetch call itself to prevent unnecessary over-fetching, but we might also not introduce a delay for rendering a loading state. It's debatable whether we want instantaneous UI feedback or not.

Agree that we should explore this in a different PR, just came to mind here so I thought I'd mention it!

apps/studio/components/layouts/SQLEditorLayout/SQLEditorMenu.tsx
joshenlim: ahh nope - don't need to block for this specific use case (altho good point RE toast error, reckon could just be a normal toast actually)

the permissions here just checks if the user can save queries - we shouldn't be blocking them from running queries

Saeris: alright I'll change up the implementation here then because when I saw that the navigation was blocked when the user didn't have permission, I figured there had to be something setup incorrectly given the error text for the toast

in other cases where we do want to block user action, does this loading pattern for menu items seem viable?

Hallidayo:

PR #38203, opened 2025-08-26
Shift useFlag hook and configcat library to common package, remove from Studio

joshenlim:
## Context

This is a pre-requisite to setting up ConfigCat for the marketing site - figured it'll be good to shift the ConfigCat logic from the dashboard into the `common` package which `www` and `docs` apps can use easily without duplication of the feature flag set up

There's seemingly a lot of changes but its mainly because of changes to the `package-lock` file

Security got tagged due to changes in `/interfaces/Organization/BillingSettings/`

## Changes involved
- Shift `configcat.ts` and `useFlag` hook to `common` -> `configcat.ts` and `feature-flag.ts` respectively
  - JFYI I've also shifted the test file `configcat.test.ts` over to `common` but we don't have CI running unit tests in that package just yet
- Update all import statements for `useFlag` in `studio` to point to `common`, no other changes to other apps
- Uninstall `configcat` package in `studio`
- Install `configcat` and `vitest` packages in `common`

## To test
- Verify that this new set up works for the dashboard, can easily check with the CreateBranchingModal `allowDataBranching` flag
  - It should be toggled on in the staging preview, where you should see this "Include data" toggle
    image


alaister (approved):
LGTM!

- Tested on preview(s)

staaldraad (approved):

ivasilov (approved):

PR #38204, opened 2025-08-26
Refine the policies page

SaxonF:
image

Update the design of the policies page to be in line with other pages and further emphasize tables with RLS disabled.

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/Policies.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/Policies.tsx
@@ -1,7 +1,6 @@
 import type { PostgresPolicy } from '@supabase/postgres-meta'
 import { isEmpty } from 'lodash'
-import { HelpCircle } from 'lucide-react'
-import { useRouter } from 'next/router'
+import Link from 'next/link'
 import { useState } from 'react'
 import { toast } from 'sonner'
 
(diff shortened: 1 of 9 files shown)


apps/studio/components/interfaces/Auth/Policies/Policies.tsx
```diff
@@ -115,40 +116,21 @@ const Policies = ({
...
+        <CardContent className="flex flex-col items-center justify-center p-8">
+          <h2 className="heading-default">No tables to create policies for</h2>
+
+          <p className="text-sm text-foreground-light text-center mb-4">
+            RLS Policies control per-user access to table rows. Create a table in this schema first
+            before creating a policy.
+          </p>
+          <Button type="default" onClick={() => router.push(`/project/${ref}/editor`)}>
+            Create a table
+          </Button>
```
joshenlim: Should be a Link, avoid router.push where possible

apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/PolicyTableRowHeader.tsx
```diff
@@ -54,28 +54,19 @@ const PolicyTableRowHeader = ({
...
-              <Lock size={14} strokeWidth={2} className="text-brand" />
-            </div>
-          ) : (
-            <div className="flex items-center gap-x-1 text-xs">
-              <Unlock size={14} strokeWidth={2} className="text-warning-600" />
-            </div>
-          )}
-          <h4 className="m-0">{table.name}</h4>
+          <Table strokeWidth={1.5} size={16} className="text-foreground-muted" />
+          <CardTitle className="m-0">{table.name}</CardTitle>
```
joshenlim: we shouldn't be uppercasing here for table names - should follow table name's casing
might be misleading otherwise

joshenlim (approved):
Tested on staging preview and all looks good now!
- Empty state when there's no tables
- Enabling / Disabling RLS (as well as corresponding UI states if theres policies or not)
- Creating, updating, deleting policy via UI
- Updating using assistant

One thing to note - editing policies with assistant doesn't invalidate the policies after the SQL is run from the assistant, so the changes are not reflected until a browser refresh. Assistant SQL blocks needs to trigger come contextual invalidation (similar to the logic that we have in the SQL Editor, but we can do that separately since this is an unrelated issue)

PR #38225, opened 2025-08-26
Refine roles

SaxonF:
image

Moves the of "applied to anonymous" badge into applied to column and groups with other roles


joshenlim (approved):
Verified on staging preview! LGTM 🙂

PR #38236, opened 2025-08-27
Chore/finalize setup configcat www

joshenlim:
## Context

Continuing from the previous PR [here](https://github.com/supabase/supabase/pull/38203), this finalizes setting up ConfigCat feature flags for `www` app. Added comments in the `configcat.ts` file under `common` packages RE setting up ConfigCat feature flags for any other applications in the future

Just needs adding `NEXT_PUBLIC_CONFIGCAT_PROXY_URL` env var to the `www` app on Vercel which has been done!

There's no visual changes involved, but in the future `www` can use ConfigCat feature flags with the `useFlag` hook just like the dashboard

## To test

- [ ] For now you can test this locally by using a flag anywhere on the www site that's been toggled on for staging/local, then console logging out that value  (value should be `true`)
  - Have verified this locally on my end


apps/www/pages/_app.tsx
```diff
@@ -86,7 +87,7 @@ export default function App({ Component, pageProps }: AppProps) {
       />
 
       <AuthProvider>
-        <FeatureFlagProvider API_URL={API_URL} enabled={IS_PLATFORM}>
```
joshenlim, line 89: there's no concept of IS_PLATFORM for www

fsansalvadore (approved):
Tested locally ✅

MildTomato (approved):

PR #38237, opened 2025-08-27
feat(replication): Significantly improve the replication UI behavior

iambriccardo:
This PR improves the replication UI with the following improvements:
- Leverages auto-defaults on the backend side, now the frontend doesn't need anymore to provide defaults as these will be automatically applied by the backend when the field is not sent by the frontend.
- Improved the feeling for enabling and disabling a pipeline, now we optimistically give feedback about starting to enable and disable a pipeline immediately after the button is pressed.
- The deletion dialog user experience has been improved, now when you press delete, you have immediate feedback about the deletion in progress.
- In case a pipeline is running and there are one or more failures in the tables, this is shown in the `View Status` button.
- When creating/editing a destination the confusing enable has been removed. Now the system has `Cancel`, `Apply` and `Apply and Restart/Apply and Start` buttons for a much cleaner UX.
- The system that manages state transitions has been redesigned to be simpler and more reliable.
- Now the system asks you to `Start` or `Stop` a pipeline which is clearer.

### New Replication Status Warning
image

### New Buttons in the Panel
image

### Restarting State
image

### Delete Destination Dialog
image


apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
```diff
@@ -74,6 +76,19 @@ export const DestinationRow = ({
   const pipelineStatus = pipelineStatusData?.status
   const statusName = getStatusName(pipelineStatus)
 
+  // Fetch table-level replication status to surface errors in list view
+  const { data: replicationStatusData } = useReplicationPipelineReplicationStatusQuery(
+    { projectRef, pipelineId: pipeline?.id },
+    {
+      enabled: !!pipeline?.id,
```
joshenlim: shouldn't be needed - cause the RQ hook itself checks for if pipelineId is undefined or not
image

apps/studio/components/interfaces/Database/Replication/DestinationRow.tsx
```diff
@@ -74,6 +76,19 @@ export const DestinationRow = ({
...
+      refetchInterval: STATUS_REFRESH_FREQUENCY_MS,
+    }
+  )
+  const errorCount = (replicationStatusData?.table_statuses || []).filter(
+    (t: any) => t.state?.name === 'error'
```
joshenlim: avoid `any` as much as possible - this one seems unnecessary since table_statuses is typed correctly

I'll push up the fixes btw so no worries - just leaving comments as I review along

apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
```diff
@@ -118,22 +127,25 @@ const DestinationPanel = ({
...
   )
+
   const form = useForm<z.infer<typeof FormSchema>>({
     mode: 'onBlur',
     reValidateMode: 'onBlur',
     resolver: zodResolver(FormSchema),
     defaultValues,
   })
+
   const onSubmit = async (data: z.infer<typeof FormSchema>) => {
```
joshenlim, line 146: There's quite a bit of duplicated logic between this and apply and run - if we can consolidate a bit

apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
```diff
@@ -143,378 +155,484 @@ const DestinationPanel = ({
...
-      onClose()
     } catch (error) {
-      toast.error(`Failed to ${editMode ? 'update' : 'create'} destination`)
+      toast.error(`Failed to ${editMode ? 'apply settings' : 'create'} destination`)
+    } finally {
+      setIsApplying(false)
     }
   }
+
+  const onApplyAndRun = () => {
```
joshenlim: can't we just call onSubmit from form.handleSubmit? with success handlers

joshenlim (approved):
- Creating a new pipeline
- Updating an existing pipeline
- Deleting a pipeline

Also tested deleting SQL snippets due to the change in API types for content delete mutation

PR #38256, opened 2025-08-27
Fix restart DB TS error

joshenlim:
## Context

We received a report that the "Fast database reboot" CTA in projects settings was throwing a toast error - on investigation this is stemming from the client side (API request is fired successfully), and its from a bug that was introduced in this [PR](https://github.com/supabase/supabase/pull/38192)

`setProjectStatus` needed to be updated as it wasn't using the right shape of the projects data (was previously just an array but its now an object like `{ projects: [], pagination: {}}`

Verified the fix working locally

### DIFF
--- a/apps/studio/components/interfaces/Settings/General/Infrastructure/RestartServerButton.tsx
+++ b/apps/studio/components/interfaces/Settings/General/Infrastructure/RestartServerButton.tsx
@@ -80,6 +80,7 @@ const RestartServerButton = () => {
   }
 
   const onRestartSuccess = () => {
+    console.log('onRestartSuccess')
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Settings/General/Infrastructure/RestartServerButton.tsx
```diff
@@ -80,6 +80,7 @@ const RestartServerButton = () => {
   }
 
   const onRestartSuccess = () => {
+    console.log('onRestartSuccess')
```
zamotany:
```suggestion
```

SaxonF (approved):
Confirmed restarting works in preview

PR #38267, opened 2025-08-28
Choose provider and model

SaxonF:
Refactors model logic to allow passing in provider and model name.


apps/studio/components/ui/AIAssistantPanel/Message.tsx
```diff
@@ -108,6 +108,8 @@ const Message = function Message({
 
   const hasTextContent = content && content.trim().length > 0
 
+  console.log(parts)
```
joshenlim: sneaky 😛

joshenlim (approved):
LGTM! Verified changes on staging preview - speed has improved a lot since the last iteration although messages still a bit verbose, but think its good enough to not cause big DX issues. Can iterate and improve from here 🙂 🙏

apps/studio/components/ui/AIAssistantPanel/Message.tsx
```diff
@@ -165,14 +165,15 @@ const Message = function Message({
...
+                            showReasoning={!!part.text}
                             className={cn(
                               'w-full dynamic-tool-item',
                               '[&:not(.dynamic-tool-item+.dynamic-tool-item)]:mt-4 [&.dynamic-tool-item+.dynamic-tool-item]:mt-1 first:!mt-0',
                               '[&:not(:has(+.dynamic-tool-item))]:mb-4'
                             )}
                             isStreaming={part.state === 'streaming'}
                           >
-                            {part.text}
+                            {part.text || part.text}
```
gregnr: I feel like I'm missing something - what does this add?

apps/studio/lib/ai/model.ts
```diff
@@ -1,55 +1,110 @@
...
 
+  const hasAwsCredentials = await checkAwsCredentials()
   const hasAwsBedrockRoleArn = !!process.env.AWS_BEDROCK_ROLE_ARN
   const hasOpenAIKey = !!process.env.OPENAI_API_KEY
 
-  if (hasAwsBedrockRoleArn && hasAwsCredentials) {
-    const bedrockModel = isThrottled || isLimited ? BEDROCK_NORMAL_MODEL : BEDROCK_PRO_MODEL
-    const bedrock = createRoutedBedrock(routingKey)
-    const { model, supportsCachePoint } = await bedrock(bedrockModel)
+  // Auto-pick a provider if not specified defaulting to Bedrock
```
gregnr, line 62: Perhaps we make `provider` mandatory for now until we dive deeper into model selection? Then there is no opportunity for unexpected behaviour.

PR #38270, opened 2025-08-28
Small fix to prevent layout shift in SQL editor when running query

joshenlim:
## Context

Internal feedback pointed out that the SQL Editor layout shifts a bit when running queries (ref before demo video)

This is due to the results footer of the SQL Editor only taking up height when there are results (it's rendered outside of the Resizeable panels), hence leading to the layout shift. Fix to change `max-h-9` to `h-9` so that the results footer takes up the full height irregardless to prevent the layout shift

## Before

https://github.com/user-attachments/assets/7356c679-a657-4289-bb54-7eefac2cb8b7

## After

https://github.com/user-attachments/assets/7b067ad1-e9a1-4cfd-8806-d55db0919d1a


alaister (approved):
LGTM!

- Tested on preview

PR #38272, opened 2025-08-28
feat: clearer upload limits

dnywh:

Reduces possible confusion around [global and bucket-specific file size limits](https://supabase.com/docs/guides/storage/uploads/file-limits).

Problematic UX on the edit/create bucket:
- We show a toast (with confusing copy) instead of inline error

Problematic UX on the global storage settings:
- We don’t expose or notify about bucket-specific limits when touching the global limit

## What is the new behavior?

Improved UX on the edit/create bucket:
- [x] Errors are contextual to their fields. We show an inline error (with clearer copy) instead of the toast
    - [x] Also for MIME type errors
-  [x] We hide all the controls behind a switch so the empty state isn't confusing
- [x] _Restrict file size_ setting: fix both footer text and error footer text to span full width of both fields
- [x] Error should link to global settings (and replace static footer text)
- [x] Replicate changes in _CreateBucketModal_ (or share source to remove duplicity)

Improved design of global storage settings:
- [x] Clearer copywriting
- [x] Fix error state: no red title, proper field error styles, proper width
- [x] Error should link to bucket-specific settings

### Create or edit bucket

The following UI also now applies to MIME type errors.

| Before | After |
| --- | --- |
| edit-before | edit-after |
| Error appears as global toast, off-screen | Inline error |

### Global storage settings

| Before | After |
| --- | --- |
| global-before  | global-after |
| No error. We currently allow these conflicts | Inline error now preventing global sizes smaller than what’s set on a bucket  |

## Docs to update

- [X] Documentation on [limits](https://supabase.com/docs/guides/storage/uploads/file-limits) (stays the same as the UI just makes things “safer”)
- [ ] Support snippets
- [ ] Support [troubleshooting](https://supabase.com/docs/guides/troubleshooting/upload-file-size-restrictions-Y4wQLT)

## In a future PR

- [ ] Change from dialog to sheet
- [ ] Improve analytics buckets vs standard buckets flow
- [ ] Remove `destructive` red title from _StorageSettings_ when errored (just have right-side styled like that)

### DIFF
--- a/apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
+++ b/apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
@@ -1,4 +1,5 @@
 import { zodResolver } from '@hookform/resolvers/zod'
+import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { snakeCase } from 'lodash'
 import { Edit } from 'lucide-react'
 import { useRouter } from 'next/router'
@@ -7,7 +8,6 @@ import { SubmitHandler, useForm } from 'react-hook-form'
 import { toast } from 'sonner'
 import z from 'zod'
 
-import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { useParams } from 'common'
 import { useIcebergWrapperExtension } from 'components/interfaces/Storage/AnalyticBucketDetails/useIcebergWrapper'
 import { StorageSizeUnits } from 'components/interfaces/Storage/StorageSettings/StorageSettings.constants'
@@ -95,17 +95,19 @@ const formId = 'create-storage-bucket-form'
 
 export type CreateBucketForm = z.infer<typeof FormSchema>
 
-const CreateBucketModal = () => {
-  const [visible, setVisible] = useState(false)
+export const CreateBucketModal = () => {
+  const router = useRouter()
   const { ref } = useParams()
   const { data: org } = useSelectedOrganizationQuery()
-  const { mutate: sendEvent } = useSendEventMutation()
-  const router = useRouter()
+
+  const [visible, setVisible] = useState(false)
+
   const { can: canCreateBuckets } = useAsyncCheckProjectPermissions(
     PermissionAction.STORAGE_WRITE,
     '*'
   )
 
+  const { mutate: sendEvent } = useSendEventMutation()
   const { mutateAsync: createBucket, isLoading: isCreating } = useBucketCreateMutation({
     // [Joshen] Silencing the error here as it's being handled in onSubmit
     onError: () => {},
@@ -136,7 +138,6 @@ const CreateBucketModal = () => {
   const isPublicBucket = form.watch('public')
   const isStandardBucket = form.watch('type') === 'STANDARD'
   const hasFileSizeLimit = form.watch('has_file_size_limit')
-  const formattedSizeLimit = form.watch('formatted_size_limit')
   const [hasAllowedMimeTypes, setHasAllowedMimeTypes] = useState(false)
   const icebergWrapperExtensionState = useIcebergWrapperExtension()
   const icebergCatalogEnabled = data?.features?.icebergCatalog?.enabled
@@ -332,7 +333,6 @@ const CreateBucketModal = () => {
 
             {isStandardBucket ? (
               <>
-                {/* Visibility */}
                 <DialogSection className="space-y-3">
                   <FormField_Shadcn_
                     key="public"
@@ -391,6 +391,7 @@ const CreateBucketModal = () => {
                       </FormItemLayout>
                     )}
                   />
+
                   {hasFileSizeLimit && (
                     <div>
                       <FormField_Shadcn_
@@ -451,7 +452,8 @@ const CreateBucketModal = () => {
                           Exceeds global limit of {formattedGlobalUploadLimit}. Increase limit in{' '}
                           <InlineLink
                             className="text-destructive decoration-destructive-500 hover:decoration-destructive"
-                            href={`/project/${ref}/settings/storage`}
+                            href={`/project/${ref}/storage/settings`}
+                            onClick={() => setVisible(false)}
                           >
                             Storage Settings
                           </InlineLink>{' '}
@@ -459,18 +461,19 @@ const CreateBucketModal = () => {
                         </FormMessage_Shadcn_>
                       )}
 
-                      {IS_PLATFORM ? (
+                      {IS_PLATFORM && (
                         <p className="text-sm text-foreground-lighter mt-2">
                           This project has a{' '}
                           <InlineLink
                             className="text-foreground-light hover:text-foreground"
-                            href={`/project/${ref}/settings/storage`}
+                            href={`/project/${ref}/storage/settings`}
+                            onClick={() => setVisible(false)}
                           >
                             global file size limit
                           </InlineLink>{' '}
                           of {formattedGlobalUploadLimit}.
                         </p>
-                      ) : undefined}
+                      )}
                     </div>
                   )}
                 </DialogSection>
@@ -614,5 +617,3 @@ const CreateBucketModal = () => {
     </Dialog>
   )
 }
-
-export default CreateBucketModal
--- a/apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
+++ b/apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
@@ -150,6 +150,7 @@ const CreateBucketModal = () => {
       )
     }
 
+    // [Joshen] Should shift this into superRefine in the form schema
     try {
       const fileSizeLimit =
         values.has_file_size_limit && values.formatted_size_limit !== undefined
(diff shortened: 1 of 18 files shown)


apps/studio/components/interfaces/Storage/StorageSettings/StorageSettings.tsx
```diff
@@ -166,22 +215,58 @@ const StorageSettings = () => {
...
+                          size="large"
+                          disabled={isFreeTier}
+                          checked={field.value}
+                          onCheckedChange={field.onChange}
+                        />
+                      </FormControl_Shadcn_>
+                    </FormItemLayout>
+                  )}
+                />
+              </CardContent>
```
dnywh: Note that _Enable Image Transformation_ now sits above _Global file size limit_. This is purposeful, as it makes the _UpgradeToPro_ callouts more closely tied to what they (mostly) talk about: file size limits.

dnywh:
https://github.com/supabase/supabase/pull/38363 builds upon this for a consolidated edit and create bucket modal component. Separate PR because it’s a large refactor (that went over my head). Works fine in my testing, though.

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
```diff
@@ -340,154 +357,165 @@ const CreateBucketModal = () => {
...
                         )}
                       />
-                    </CollapsibleContent_Shadcn_>
-                  </Collapsible_Shadcn_>
+                      {formattedSizeLimitError?.message === 'exceed_global_limit' && (
+                        <FormMessage_Shadcn_ className="mt-2">
+                          Exceeds global limit of {formattedGlobalUploadLimit}. Increase limit in{' '}
+                          <InlineLink
+                            className="text-destructive decoration-destructive-500 hover:decoration-destructive"
+                            href={`/project/${ref}/settings/storage`}
```
joshenlim: just realised - all these should be pointing to storage/settings instead (settings/storage is an old URL, so redirect to this URL via Link was causing a full re-render)

i'll get these fixed while i review btw no worries 🙂 🙏

joshenlim:
Found a small issue - editing a bucket that doesn't have a file size limit, doesnt init the unit to MB
[edit] Pushed up a fix!

joshenlim:
Small issue found - toggling a bucket to be public or not doesnt show (or hide) the public badge until a refresh

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -106,42 +100,105 @@ export const EditBucketModal = ({ visible, bucket, onClose }: EditBucketModalPro
...
+      )
+
+      if (bucketLimitInBytes > data.fileSizeLimit) {
+        form.setError('formatted_size_limit', { type: 'manual', message: 'exceed_global_limit' })
+        return setIsUpdating(false)
+      }
+    }
+
+    try {
+      const result = await updateBucket({
```
joshenlim: ahh JFYI we should be using the react query mutations, not the update methods directly
so that the related data after updating the bucket gets invalidated and refreshed

although good point on this - we shouldn't export these methods unless needed

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
```diff
@@ -131,38 +128,46 @@ const CreateBucketModal = () => {
...
+          ? convertToBytes(values.formatted_size_limit, selectedUnit as StorageSizeUnits)
+          : undefined
 
-      const allowedMimeTypes =
-        values.allowed_mime_types.length > 0
+      const allowedMimeTypes = hasAllowedMimeTypes
+        ? values.allowed_mime_types.length > 0
           ? values.allowed_mime_types.split(',').map((x) => x.trim())
           : undefined
+        : undefined
```
joshenlim: can be simplified to:

```
const allowedMimeTypes =
        hasAllowedMimeTypes && values.allowed_mime_types.length > 0
          ? values.allowed_mime_types.split(',').map((x) => x.trim())
          : undefined
```

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -195,167 +307,123 @@ export const EditBucketModal = ({ visible, bucket, onClose }: EditBucketModalPro
...
+                                placeholder="0"
+                                {...field}
+                              />
+                            </FormControl_Shadcn_>
+                          </div>
+                          <div className="col-span-4">
+                            <Select_Shadcn_ value={selectedUnit} onValueChange={setSelectedUnit}>
+                              <SelectTrigger_Shadcn_ aria-label="File size limit unit" size="small">
+                                <SelectValue_Shadcn_ asChild>
+                                  <>{selectedUnit}</>
```
joshenlim: unnecessary fragment - can remove asChild property too

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -195,167 +307,123 @@ export const EditBucketModal = ({ visible, bucket, onClose }: EditBucketModalPro
...
+                                </SelectValue_Shadcn_>
+                              </SelectTrigger_Shadcn_>
+                              <SelectContent_Shadcn_>
+                                {Object.values(StorageSizeUnits).map((unit: string) => (
+                                  <SelectItem_Shadcn_ key={unit} value={unit} className="text-xs">
+                                    <div>{unit}</div>
```
joshenlim: unnecessary div here

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
```diff
@@ -195,167 +307,123 @@ export const EditBucketModal = ({ visible, bucket, onClose }: EditBucketModalPro
...
+                      <InlineLink
+                        className="text-foreground-light hover:text-foreground"
+                        href={`/project/${ref}/storage/settings`}
+                        onClick={onClose}
+                      >
+                        global file size limit
+                      </InlineLink>{' '}
+                      of {formattedGlobalUploadLimit}.
+                    </p>
+                  ) : undefined}
```
joshenlim: could do a `&&` if theres no alternative UI to render instead of using a ternary

apps/studio/components/interfaces/Storage/StorageSettings/StorageSettings.tsx
```diff
@@ -149,170 +183,261 @@ const StorageSettings = () => {
...
-              )}
-
-              {!canUpdateStorageSettings && (
-                <CardContent className="pt-0">
-                  <p className="text-sm text-foreground-light">
-                    You need additional permissions to update storage settings
-                  </p>
-                </CardContent>
-              )}
+                  {!canUpdateStorageSettings && (
```
joshenlim, line 407: Just a feedback here: Feels like the "Need additional perms" message to update settings takes up a lot of space
Presuming that we'll be figuring out a pattern for this sort of state - wondering if we should hide the CardFooter?
image

dnywh: Yeah this bugged me too. Thanks for calling out. It predates this PR so I’ve added it to the design audit to fix at a later date/more holistically

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
dnywh: Good catch. Thanks!

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
dnywh: Thanks Joshen. Are [the TanStack docs](https://tanstack.com/query/latest/docs/framework/react/guides/mutations) the best place to learn about this?

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
dnywh: Oh yeah good catch :)

joshenlim (approved):
Verified on staging preview - creating / editing bucket modals show the errors if setting upload limit greater than global limit, and vice versa on the storage settings page if setting global limit lower than any buckets' upload limit

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
dnywh: Duh!

dnywh:
Sibling PR for Storage API: https://github.com/supabase/storage/pull/752

apps/studio/components/interfaces/Storage/EditBucketModal.tsx
joshenlim: hmm the docs there are probably useful to understand how the react queries work
but the way we use react query isnt exactly like how the docs describe it haha

probably easier to reference our existing react query hooks, or our templates [here](https://github.com/supabase/supabase/blob/master/apps/studio/data/__templates/README.md#L16) - although we haven't relooked at the templates README in a long while so just feel free to call out if any parts of it might be confusing or seems wrong 😅

dnywh: Looks great from a glance. Will do, though. Thanks!

apps/studio/components/interfaces/Storage/StorageSettings/StorageSettings.tsx
joshenlim: all good no worries! this is non-blocking feedback btw so feel free to merge either way 🙂 🙏 

i noticed that the S3 Connection panel also follows this pattern atm so i think its fine from a consistency POV

PR #38286, opened 2025-08-28
feat(studio): query insights 

kemaldotearth:

YES

We're looking at introducing a Query Insights tab within our Reports. This is part of broader work aimed at getting `pg_stat_monitor` working with Supabase. 

Currently we have a Query Performance area in Advisors. This will then be deprecated in favour of the the Query Insights tab, that'll live in Reports.

## What is the new behavior?

Introduces a prototype consuming `pg_stat_monitor` query analytics over set periods of time.

## Additional context

This is a prototype and a WIP. We're also developing on self hosted until `pg_stat_monitor` is available on our regular development workflow, so if you're wondering why Reports are visible on self hosted, they're just there for aesthetics and to visualise the section where it will eventually be.


kemaldotearth:
Closing this as we now have a way to consume pg_stat_monitor data locally on regular setup.

PR #38292, opened 2025-08-28
fix: small typo in identites -> identities

cstockton:
Fixes a small typo in the apps/docs/spec files.

I am not exactly sure how this ends up making its way into the documentation. I tried running the make command in the apps/docs/specs directory and it ended with an error before pulling a large amount of changes to various files.

Would appreciate some guidance on how to bring everything up to date.


hf (approved):
I don't remember if these were autogenerated

PR #38297, opened 2025-08-29
Fix policy alignment

SaxonF:
Fixes a container alignment issue on auth/policies page.


joshenlim (approved):
Verified fixes on staging preview!

PR #38298, opened 2025-08-29
Support for projects pagination part 1

joshenlim:
Note: Identified some issues in the API regarding permissions on the `/organization/[slug]/projects` endpoint that API team is currently investigating still, needs to be addressed before this PR can be merged

## Context

This will be part of several PRs to introduce pagination for projects in the dashboard. It'll involve several changes across different parts of the dashboard so will do this incrementally. The previous PR related to this was [here](https://github.com/supabase/supabase/pull/38192) but after planning the work properly I realised the changes there might have been unnecessary as we'd need to use a different RQ for infinite scrolling 😅🙏 

Changes here are solely scoped to the projects home page + projects wildcard page

JFYI there's no virtualization here as (for both card and table view) the project card / row heights are dynamic (not identical) depending on the project's config like if there's vercel / GH integration, or if the project has any exhaustion warnings etc. I think it should be alright for now, but we can adjust if we deem there's a need (which we'll also then need to revisit the UI too)

## Changes involved
- Using `/organization/[slug]/projects` API endpoint to list projects
- Added Infinite scrolling pagination for projects on the projects home page (/org/[slug])
  - For both card view and table view
  - Both views share the same RQ hook, so fetching the next page on one view will also persist the projects on either view (no redundant fetching)
- Searching is shifted to server side via debounce, and also shifted to using URL state (instead of react state)
- Updated project wildcard page view to support infinite scrolling UX
  - No longer showing _all_ organizations and their respective projects in the whole page
  - Organization will need to be selected via a dropdown (default to last visited org, or first org that the user belongs to)
- (Unrelated) Selecting a view on the projects page will now persist in local storage, should return back to that view on a browser refresh
- (Unrelated) `typography.scss` and `main.scss` are both importing `@tailwind ___` files which is resulting in duplicate css styles getting loaded (can verify on prod atm by checking any component with for e.g `text-foreground` class name, you'll see in the Dev tools console that there's duplicate styles getting applied). Am opting to remove the duplicate where able to (utilities seems to be needed in both files unfortuantely), but we'll also need to clean up `main.scss` too - there's probably a lot of unnecessary stuff in there now

## To test
Understandably this is going to be hard to test on staging since we don't have that many projects in the organization, so I'd recommend testing with either of these 2 ways, both of which involve manually setting the page limit to something lower like `12` (I have currently done so in this PR - need to revert before merging)
- Staging: I have an organization that I can add whoever's testing to, there's like 16 projects, enough to test minimally the fetching of the next page
  - Then once tested, we raise it back to `96` before merging
- Local: Manually create at least 16 projects, then reduce the page limit to 12

I've also double checked on mobile as well to ensure that the pagination is working

## What's next
We're going to need to update the UIs for anywhere that's fetching projects (e.g dropdowns, etc)


apps/studio/data/projects/projects-infinite-query.ts
```diff
@@ -0,0 +1,87 @@
...
+import { components } from 'api-types'
+import { get, handleError } from 'data/fetchers'
+import { useProfile } from 'lib/profile'
+import { ResponseError } from 'types'
+import { projectKeys } from './keys'
+
+// [Joshen] Try to keep this value a multiple of 6 (common denominator of 2 and 3) to fit the cards view
+// So that the last row will always be a full row of cards while there's a next page
+// API max rows is 100, I'm just choosing 96 here as the highest value thats a multiple of 6
+const DEFAULT_LIMIT = 12
```
joshenlim: Temporarily set to 12 for testing, need to set to 96 before merging

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
```diff
@@ -1,47 +1,85 @@
...
-  } = useProjectsQuery()
-  const allProjects = data?.projects ?? []
+    isFetching,
+    isFetchingNextPage,
+    hasNextPage,
+    fetchNextPage,
+  } = useOrgProjectsInfiniteQuery(
+    {
+      slug,
+      search: search.length === 0 ? search : debouncedSearch,
```
charislam, line 74:
```suggestion
      search: debouncedSearch,
```

tiny, tiny nit: this shouldn't make a difference, should it? 🤔

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
```diff
@@ -150,13 +166,20 @@ export const ProjectList = ({
               <TableHead>Created</TableHead>
             </TableRow>
           </TableHeader>
+          <TableHeader>
```
charislam: nit: it's not valid HTML to have two `<thead>`s in one table, does it work if we put both `TableRows` in the same `TableHeader`?

apps/studio/pages/project/_/[[...routeSlug]].tsx
Comment (charislam) [inline apps/studio/pages/project/_/[[...routeSlug]].tsx:1]: on this page only, for some reason the next page projects are forever stuck in loading state: 

Image

apps/studio/components/interfaces/HomePageActions.tsx
```diff
@@ -17,45 +19,51 @@ import {
...
+  const [filterStatus, setFilterStatus] = useQueryState(
+    'status',
+    parseAsArrayOf(parseAsString, ',').withDefault([])
+  )
+  const [viewMode, setViewMode] = useLocalStorageQuery(LOCAL_STORAGE_KEYS.PROJECTS_VIEW, 'grid')
+
   return (
     <div className="flex items-center justify-between">
       <div className="flex items-center gap-2">
         <Input
```
charislam, line 58: The search over the network can be a little slow, which sometimes makes it look like nothing is happening. Worth putting a loading state on this?

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
joshenlim: ahh i made it this way so that the UI updates immediately if we're clearing the search, instead of having that 500ms pause in between (makes the UX a bit more snappier, remove unnecessary pauses)

apps/studio/components/interfaces/HomePageActions.tsx
joshenlim: i'll add a loader here as an indicator for the card view 🙂 
image

joshenlim:
Thanks so much for the review @charislam! 🙏 🙏  just pushed up a couple of fixes for the issues you identified 🙂 

Just a heads up tho - there's a known issue on the API where if you filter projects by status, the count isn't getting returned correctly so you might see a loader in either the card or table UI, have poked API about this one 🙏 🙂

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
joshenlim: Yup that works!

apps/studio/pages/project/_/[[...routeSlug]].tsx
Comment (joshenlim) [reply apps/studio/pages/project/_/[[...routeSlug]].tsx:1]: got this fixed!

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
charislam: ahhh I am stupid 😅

charislam (approved):
🙌🏼

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
joshenlim: not at all! hahaha its a new behaviour i'm introducing into the dashboard for where we have debounced search tbh, so i'll incrementally add this to more places as we go along

joshenlim:
Raised page limit back to 96 - reminder to self to check once more before merging 🙏

joshenlim:
oh JFYI there's a small bug on API end that needs addressing (when filtering by status, the count returned doesn't match)

i don't necessarily think its a blocker, but can wait till Monday to merge

FIx is here: https://github.com/supabase/infrastructure/pull/25522

[Edit] I'm gonna reserve this for Monday to merge just in case anything happens 😅

joshenlim:
One more issue found on the API end related to project scoped roles permissions so still blocked - do not merge

joshenlim:
API changes have been addressed and verified working on prod (tested on my end as well)

Will have this merged in a bit today

PR #38299, opened 2025-08-29
chore: minor clarity improvements on new project creation

dnywh:

Some very minor clean up of the new project creation inline-dialog.

- Some warning text (staging only) that is hard to read on light mode and a bit unclear
- Inconsistent use of title case vs sentence case for field labels
- Missing periods

## What is the new behavior?

Fixing all of the above:

| Before | After |
| --- | --- |
| Before | After |

## Additional context

I'm using `text-warning-600` because I couldn't figure out how to set that shade for `text-warning` for light mode. Dark mode doesn't seem to be affected. But still, this should be fixed (in a separate PR) because any time we use `text-warning` on light mode, the text fails colour contrast checks.


ivasilov (approved):

dnywh:
`text-warning` fixes (as mentioned above) are coming in https://github.com/supabase/supabase/pull/38343

PR #38307, opened 2025-08-29
fix(design-system): missing path and deps

kemaldotearth:
Refactored the Drawer component for improved slot-based structure, updated styles, and added 'use client' directive. Added 'date-fns' as a dependency in design-system, updated tsconfig paths for icons, and marked ToggleGroup as a client component.

**NOTE:** Fixes @MildTomato made to get the design system project up and running again. I got some changes I want to make so this unblocks me 🥳


MildTomato:
i copied latest toggle-group and drawer from shadcn
@SaxonF @fsansalvadore might want to double check

apps/design-system/public/llms.txt
```diff
@@ -8,7 +8,7 @@ The Design System used in Supabase Studio.
 - [Changelog](https://supabase-design-system.vercel.app/design-system/docs/changelog)
     - Latest updates and announcements.
 - [Color usage](https://supabase-design-system.vercel.app/design-system/docs/color-usage)
-    - Colors system breakdown with best practise
+    - Colors system breakdown with best practice
```
dnywh, line 11: Mine does this too!

apps/design-system/registry/default/example/drawer-dialog.tsx
```diff
@@ -1,3 +1,5 @@
+'use client'
```
dnywh, line 1: I don’t fully get why we need to use the `client` directive now (and not before) but it seems necessary for the build to work.

apps/design-system/registry/default/example/scroll-area-demo.tsx
```diff
@@ -8,15 +8,12 @@ const tags = Array.from({ length: 50 }).map((_, i, a) => `v1.2.0-beta.${a.length
 export default function ScrollAreaDemo() {
   return (
     <ScrollArea className="h-72 w-48 rounded-md border">
-      <div className="p-4">
+      <div className="p-4 divide-y divide-broder">
```
dnywh:
```suggestion
      <div className="p-4 divide-y divide-border">
```

Minor typo on `divide-border`

dnywh (approved):
Thanks for doing this!

apps/design-system/config/docs.ts
```diff
@@ -389,11 +389,11 @@ export const docsConfig: DocsConfig = {
...
-        {
-          title: 'Toast',
-          href: '/docs/components/toast',
-          items: [],
-        },
+        // {
+        //   title: 'Toast',
+        //   href: '/docs/components/toast',
+        //   items: [],
+        // },
```
joshenlim: just to double check if this is intentional/?

packages/common/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "devDependencies": {
     "@types/lodash": "4.17.5",
+    "@types/node": "catalog:",
```
joshenlim: just to check if this needed?

apps/design-system/registry/default/example/chart-tooltip-demo.tsx
joshenlim, line 1: Unrelated but i realised none of the charts are rendering (also happening on staging)
https://design-system-2325nqkua-supabase.vercel.app/design-system/docs/components/chart

apps/design-system/public/llms.txt
kemaldotearth: This happens when `pnpm build:design-system` is run. It runs a script that generates the file, but I'm not entirely sure why it localises.

Either way, it spells it the CORRECT way.

apps/design-system/registry/default/example/drawer-dialog.tsx
kemaldotearth: Don't think it necessarily needs it, I put it there while debugging, can remove it.

apps/design-system/config/docs.ts
kemaldotearth: Can remove, we currently don't have a Toast page, it was 404'ing - we instead have a "Sonner" page, which can be updated in a separate ticket. Will just remove this for now.

packages/common/package.json
kemaldotearth: I'm not sure why it's showing this as a new line, it was added 5 months ago according to GitLens.

joshenlim: 👀  odd i'm checking on master but i don't see this package in common/package.json
image

maybe we can just remove this then if that's the case

kemaldotearth: Removed.

apps/design-system/registry/default/example/chart-tooltip-demo.tsx
kemaldotearth: @joshenlim this happens on the [current version](https://supabase-design-system.vercel.app/design-system) that's live to the interwebs. I think let's isolate this one outside of the scope of this PR, as I don't want to risk breaking charts across the board.

joshenlim: yup defo! 😄 🙏  happy to tackle this separately 🙂

apps/design-system/registry/default/example/scroll-area-demo.tsx
kemaldotearth: Thanks broder 👊

joshenlim (approved):
LGTM! Minimally checked dashboard staging preview + design system staging preview (in the previous build)

PR #38308, opened 2025-08-29
Fix SQL Editor search not showing snippets in folders

joshenlim:
## Context

We received a report regarding searching in SQL Editor that it doesn't render snippets that are within folders, despite the count showing that there are snippets (Search should be showing a flat list of snippets for easier finding)

e.g:
image

Realised that we need to set (on the client side) `null` for all snippet's `folder_id` parameter so that the UI can render the snippets as a flat list. We don't need the `folder_id` in this specific UI since we dont have any move actions for snippets, but do test a little to check if i might have overlooked anything 🙏 
image

## To test
- [ ] Verify that searching will show results for snippets that are within folders
- [ ] Verify that none of the context menu actions for snippets in the search UI isn't breaking


alaister (approved):
LGTM!

- Tested on preview

PR #38336, opened 2025-09-01
Home New: Top

SaxonF:
image

This is the first step of the new dashboard home which:
- Adds feature flag to show new home if true, otherwise continue to show current
- Adds top section of home
- Sets up sortable sections

The follow up PR's will be adding each additional row/section.


apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureConfiguration/InstanceConfiguration.tsx
```diff
@@ -210,9 +214,9 @@ const InstanceConfigurationUI = () => {
   }, [isSuccessReplicas, isSuccessLoadBalancers, nodes, edges, view])
 
   return (
-    <div className="nowheel border-y">
+    <div className={diagramOnly ? 'h-full' : 'nowheel border-y'}>
```
joshenlim: needs no wheel, otherwise wont be able to scroll while mouse is hovering on the reactflow diagram

joshenlim:
Pushed up some fixes and refactors, some notes:
- Let's avoid `export default` from here, could just do `export` directly on the component
- Had to update `SimpleCodeBlock` as it was having some console errors regarding `key` property getting passed in via spread operator

### Feedback
- It's not very obvious where to find the service statuses now, I only found it cause I was wondering what clicking on "Healthy" does
  image

- The empty state for the branches section is a little off
  image
  - Branch names are based on user input and can be dynamic so this could easily be very long
  - Even if we were to add truncate here i think it might look odd then as it truncates pretty early
    image
  - Wondering if we should keep branches here?

SaxonF:
@joshenlim I made some style adjustments . Let me know if that helps.

joshenlim (approved):
Verified on staging preview! 🙂 🙏

apps/studio/components/interfaces/HomeNew/ActivityStats.tsx
```diff
@@ -0,0 +1,131 @@
...
+          <h4 className="heading-meta text-foreground-light mb-1">
+            {isDefaultProject ? 'Recent branch' : 'Branch Created'}
+          </h4>
+
+          <div className="text-foreground truncate h-[34px] flex items-center capitalize-sentence">
+            {isLoadingBranches ? (
+              <Skeleton className="h-6 w-24" />
+            ) : isDefaultProject ? (
+              <div className="flex items-center gap-2">
+                <GitBranch size={16} strokeWidth={1.5} className="text-foreground-lighter" />
```
dnywh, line 112: Minor low-stakes suggestion: Make the `GitBranch` icon condition conditional so the empty state matches the other rows (no icon). Or visa versa; add icons to the others too. Right now it sticks out a bit.

Image

apps/studio/components/interfaces/HomeNew/ActivityStats.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const list = backupsData?.backups ?? []
+    if (list.length === 0) return undefined
+    return list
+      .slice()
+      .sort((a, b) => new Date(b.inserted_at).valueOf() - new Date(a.inserted_at).valueOf())[0]
+  }, [backupsData])
+
+  return (
+    <div className="@container">
+      <div className="grid grid-cols-2 xl:flex xl:gap-10 gap-4 flex-wrap">
```
dnywh, line 58: Another minor suggestion: these gaps feel a little tight to me, on all breakpoints. I wonder if a bit more breathing room would help. E.g. `xl:gap-[4rem] gap-10 ` feels better on my local.

apps/studio/components/interfaces/HomeNew/ActivityStats.tsx
```diff
@@ -0,0 +1,131 @@
...
+        <div className="block">
+          <p className="heading-meta text-foreground-light mb-1">Status</p>
+          <ServiceStatus />
+        </div>
+
+        <Link className="block" href={`/project/${ref}/database/migrations`}>
```
dnywh, line 64: Might be nice to add a little hover state on the children of these like:

```jsx
<Link className="block hover:[&_div>p]:text-foreground transition-colors" />
```

...so their interactivity is reinforced

dnywh (approved):
Some superficial non-blocking feedback. Looks good!

PR #38338, opened 2025-09-01
chore: upgrade vaul to 1.1.2

dnywh:
Upgrades Vaul to the latest version to fix build errors.


dnywh:
Fixed instead in https://github.com/supabase/supabase/pull/38307

PR #38337, opened 2025-09-01
Home New: Advisors

SaxonF:
image

Adds advisors row to the new home. 

This also includes a new Row component that allows for horizontal scroll whilst showing content outside bounds. 

This also refactors advisors page a little bit by making use of a shared detail component and removing need for selected lint state vs just calculating based on query params. 

Requires: https://github.com/supabase/supabase/pull/38336

PR #38339, opened 2025-09-01
Home New: Project Usage

SaxonF:
image

Adds project usage row to the new home.

This currently uses logs endpoint to calculate which is not a long term solution and will just let us play with what data makes sense without investing time in api. We must transition away before removing new home feature flag.

Requires: 
- https://github.com/supabase/supabase/pull/38336
- https://github.com/supabase/supabase/pull/38337


joshenlim (approved):
Tested on staging preview - LGTM! Although i'm not sure why the ui-library build is failing 😕

PR #38341, opened 2025-09-01
Home New: Report

SaxonF:
image

Adds report row to the new home. This creates a report called "Home" if none exists and embeds on homepage in horizontal layout.

Requires:

https://github.com/supabase/supabase/pull/38336
https://github.com/supabase/supabase/pull/38337


apps/studio/components/interfaces/HomeNew/SnippetDropdown.tsx
```diff
@@ -0,0 +1,107 @@
...
+    root: scrollRootRef.current,
+    threshold: 0,
+    rootMargin: '0px',
+  })
+
+  useEffect(() => {
+    if (entry?.isIntersecting && hasNextPage && !isFetchingNextPage && !isLoading) {
+      fetchNextPage()
+    }
+  }, [entry?.isIntersecting, hasNextPage, isFetchingNextPage, isLoading, fetchNextPage])
```
joshenlim, line 69: oooh seems like a great pattern - will follow this for dropdowns with infinity scrolling

joshenlim:
Two issues found:

- Should I be allowed to add the same snippet more than once? Reckon we could just skip adding a new block if selecting a snippet that's already exist in the section (Filtering from the dropdown will be quite tricky so opting that as the easier workaround)

https://github.com/user-attachments/assets/11859a23-7d09-4a4c-a1ae-f3ee63bc85f0

- Adding a single block, then refreshing, the single block expands to the size of the section - reckon it should stay the same size?

https://github.com/user-attachments/assets/b236ccb2-8960-46d1-ab72-c3ca0fa8009d

SaxonF:
@joshenlim agree on both counts. I've updated Row and added a check for existing blocks.

joshenlim (approved):
Verified changes on staging preview - all looking good! 🙂

PR #38344, opened 2025-09-01
feat: v1 migrations endpoint for self hosted

sweatybridge:

feature

## What is the new behavior?

Adds migrations endpoint for self-hosted.

## Additional context


gregnr (approved):
Tested with local Supabase stack and successfully listed and created migrations via this API 👍

PR #38346, opened 2025-09-01
ref(replication): Remove replication lag

iambriccardo:
This PR removes the replication lag metric from the UI since right now we are not going to release ETL with such a metric and we want to avoid confusing users.


joshenlim (approved):
Code changes LGTM - not tested on staging preview though JFYI

PR #38347, opened 2025-09-01
chore: add ConfigCat custom attributes

alaister:
Adds custom attributes to the ConfigCat user. At this stage just `cloud_provider`


joshenlim (approved):
Code wise LGTM! Verified on staging preview minimally that existing feature flags are still working

PR #38363, opened 2025-09-02
consolidate bucket edit and create modal

dnywh:

Refactoring. Code consolidation and architecture improvement.

- `CreateBucketModal` and `EditBucketModal` contain ~800 lines of duplicated logic
- File size limit validation is scattered across multiple places in component logic
- Form validation logic is inconsistent between create and edit flows
- Maintenance requires updating the same logic in multiple places

## What is the new behavior?

- **Unified `BucketModal` component** consolidates all shared logic between create and edit modes
- **Mode-based architecture** (`mode: 'create' | 'edit'`) handles both flows in a single component
- **Centralized validation** moved from component logic to structured form validation in submit handlers
- **Legacy components** now serve as thin wrappers, maintaining backward compatibility
- **Single source of truth** for all bucket modal functionality

**Code reduction**: ~1,080 lines of duplicated code eliminated across the three components.

## Additional context

This refactoring follows the established pattern used in table editor components and addresses feedback about consolidating create/edit paths. The validation logic was moved to the submit handlers (rather than `superRefine`) due to the need to access project configuration data that isn't available in the form schema context.

All existing imports continue to work without changes, ensuring zero breaking changes while providing a much cleaner, more maintainable architecture going forward.


dnywh:
Too out of date with latest buckets work. Closing.

PR #38367, opened 2025-09-02
Fix button text consistency for database triggers

joshenlim:
Fixes FE-1826

## Context

Tiny consistency fix - when editing a trigger the CTA still says "Create trigger" in the side panel, so this fixes it to show "Save trigger" instead

image

Separately though, there's a general consistency issue with the primary CTA texts across the panels on the database pages (some say "Save", "Confirm", and "Save/Create entity") - we'll need to address this along the way. For now I've also done this update for the functions side panel to say "Save/Create function" instead of just "Confirm"


alaister (approved):
LGTM!

PR #38369, opened 2025-09-02
Add Edge Function Report Filters

jordienr:
CleanShot 2025-09-02 at 10 59 25@2x

- Adds edge function report filters with some basic components
- Plan to move this to the FilterBar component once its production ready
- Adds more tests 
- Filters are now typesafe in the dataProvider 
- Will use same setup for the rest of the reports
- thanks for reading, have a nice day

## To test
- Go to edge functions report and try filtering, check data makes sense

## Demo

https://github.com/user-attachments/assets/f44d56d8-deb8-4881-9be3-1d2dde32e195


apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
```diff
@@ -49,21 +46,16 @@ export const ReportChartV2 = ({
     data: queryResult,
     isLoading: isLoadingChart,
     error,
+    isFetching,
   } = useQuery(
-    ['projects', projectRef, 'report-v2', report.id, { startDate, endDate, interval, functionIds }],
```
joshenlim, line 53: let's keep the projects and projectRef in here? (ref [comment](https://github.com/supabase/supabase/pull/38129#discussion_r2307286979))

apps/studio/components/interfaces/Reports/v2/ReportsNumericFilter.tsx
```diff
@@ -0,0 +1,215 @@
...
+                {getDisplayValue()}
+              </span>
+            )}
+          </span>
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent align="start" className="p-0 w-72" portal={true}>
+        <div className="p-2 border-b border-default flex items-center justify-between">
+          <span className="text-sm text-foreground">{label}</span>
+          <div className="flex gap-2">
```
joshenlim: div here seems unnecessary

apps/studio/components/interfaces/Reports/v2/ReportsNumericFilter.tsx
```diff
@@ -0,0 +1,215 @@
...
+                  : min !== undefined
+                    ? `Min: ${min}`
+                    : `Max: ${max}`}
+              </p>
+            )}
+          </div>
+        </form>
+
+        <div className="flex items-center justify-end gap-2 border-t border-default p-2">
+          <Button size="tiny" type="default" onClick={() => setOpen(false)} htmlType="button">
```
joshenlim, line 196: should have a `disabled={isLoading}`

apps/studio/components/interfaces/Reports/v2/ReportsSelectFilter.tsx
```diff
@@ -0,0 +1,164 @@
...
+                    <span className="text-sm text-foreground-lighter">{option.description}</span>
+                  )}
+                </div>
+              </Label>
+            ))
+          )}
+        </div>
+
+        <div className="flex items-center justify-end gap-2 border-t border-default p-2">
+          <Button size="tiny" type="default" onClick={() => setOpen(false)} htmlType="button">
```
joshenlim, line 130: likewise needs a `disabled={isLoading}` here

joshenlim (approved):
Generally LGTM! Tested on staging preview - just minor feedback and comments 🙂

joshenlim:
## Non-blocking UI feedback

- Could probably render the flags here? Since they match AWS regions
  image
  - Thinking maybe something like this
    image

- Wonder if we might be able to breakdown the data for each edge fn (e.g here - im wondering which functions the counts are from)
  image

apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
jordienr: sorry was meant to be a temporal change and forgot to re add it 🫠

apps/studio/components/interfaces/Reports/v2/ReportsNumericFilter.tsx
jordienr: `<Button>` handles that internally, but I did miss adding passing the loading state to the filter in edge-functions.

https://github.com/supabase/supabase/blob/f927998e4e832345498477dbf53b5f8950fa5fe5/packages/ui/src/components/Button/Button.tsx#L247

joshenlim: oh but this one's the cancel button though - figured this button should be disabled as well while the applying is loading

but actually no worries about this one cause i realised the popover closes after hitting apply, so doesnt really matter in this case

PR #38394, opened 2025-09-03
Gracefully handle feature flag set up

joshenlim:
## Context

This follows up the shifting of feature flag functionality into the common packages and improves the DX for external contributors on the `www` app by gracefully handling the initialisation of feature flags (or anyone who doesn't have access to a proper `.env.local`). 

Two main changes
- in `lib/supabase.ts`, only initialize the supabase client if the necessary env vars are present
  - Otherwise set the supabase client as undefined and gracefully handle accordingly (this one's only used in `Footer.tsx`)
- In `common/configcat.ts`, skip initialization of ConfigCat flags if neither of the required env vars are present
  - Also gracefully handle if falling back to default ConfigCat, but SDK key is not present
- In `common/feature-flags.tsx`, skip fetching PH flags if `API_URL` env var is not present

[Edit] Also made these changes
- `FeatureFlagProvider` supports an object for the `enabled` property for granular provider toggling
  - `{ cc: boolean, ph: boolean }`
  - In this case since `www` doesn't use PH, it's toggled off for now, and local wouldn't throw an error for Failed to fetch if infra is not running
  - `studio` remains `enabled` as just a boolean since we want both if `IS_PLATFORM`
- `FeatureFlagProvider` also no longer expects a `getConfigCatFlags` to be passed as a prop
  - Will just use the default method from `common/config-cat.ts` if not provided
  - `studio` still provides that function as a prop as there's some custom logic

I've verified the changes on both `www` and `studio` that feature flags are still working, and that if I've removed `.env.local` for `www`, the app spins up without any client side errors 🙏 But do let me know if there's anything else that I might be missing!

## To test
- `www`
  - [x] Verify that if you don't have a `.env.local`, `www` still spins up properly with no client side errors
  - [x] Verify that if you do have a `.env.local`, feature flags are then working on `www` (e.g state of startups page)
- `studio`
  - [x] Verify that feature flag are still working for the dashboard
    - Easiest way to verify is the create branch modal, just ensure that the "include data" toggle is present since that's behind a flag
    image
- `docs`
  - Just verify that the site loads correctly (It's not using feature flags atm, but it does render the FeatureFlagProvider)


packages/common/configcat.ts
```diff
@@ -27,17 +27,22 @@ export const fetchHandler: typeof fetch = async (input, init) => {
...
-  if (!process.env.NEXT_PUBLIC_CONFIGCAT_PROXY_URL) {
-    console.error(
-      'Failed to get ConfigCat client: missing env var "NEXT_PUBLIC_CONFIGCAT_PROXY_URL"'
-    )
-  }
-
   if (client) return client
 
+  if (!process.env.NEXT_PUBLIC_CONFIGCAT_SDK_KEY && !process.env.NEXT_PUBLIC_CONFIGCAT_PROXY_URL) {
+    console.log('Skipping ConfigCat set up as env vars are not present')
```
dnywh, line 33: Can confirm I get this logged when on `www` without _env.local_

dnywh:
Predates but related to this work: ConfigCat pollutes the browser console with a bunch of warnings. You can see this on prod, too. Here’s one example:

> ConfigCat - WARN - [3001] Cannot evaluate targeting rules and % options for setting 'advisorRules' (User Object is missing). You should pass a User Object to the evaluation methods like `getValueAsync()` in order to make targeting work properly. Read more: https://configcat.com/docs/advanced/user-object/

That, and all the others, are being triggered by _configcat.ts:72_. Happens on both `www` and `studio`.

joshenlim:
> ConfigCat pollutes the browser console with a bunch of warnings. You can see this on prod, too.

Yup this one's a known issue 😅 and it's caused a couple of red herrings in support tickets - will look into seeing how we can clean those up separately

dnywh (approved):
Nice! Works as expected

PR #38404, opened 2025-09-03
Fix table editor incorrectly showing type change notice for foreign key

joshenlim:
## Context

The ForeignKeySelector of the Table Editor incorrectly shows a type change notice even if the source column type and target column type matches
image

## Changes involved
- Correctly order of early returns in `validateType`
- Avoid unnecessary pushing of `undefined` into typeNotice or typeErrors

## To test
- Select "New table" and ensure that one of your columns is of type `uuid` 
- Add a foreign key to the `auth.users` table on the `id` column
- Because the type matches, you shouldn't see the notice
- Also just check that the notice shows up correctly when it should, by selecting a target column of a different type


charislam (approved):
Tested as described!

PR #38405, opened 2025-09-03
docs: SSR overhaul combined client page PoC

ChrisChinchilla:

YES


apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -39,58 +39,58 @@ pnpm add @supabase/ssr @supabase/supabase-js
 
 ## Set environment variables
 
-In your environment variables file, set your Supabase URL and Supabase Anon Key:
+Create a `.env.local` file in the project root directory. In the file, set the project's Supabase URL and Supabase Publishable Key:
 
 <ProjectConfigVariables variable="url" />
-<ProjectConfigVariables variable="anonKey" />
+<ProjectConfigVariables variable="publishableKey" />
```
charislam:
```suggestion
<ProjectConfigVariables variable="publishable" />
```

based upon your other PR

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -155,297 +156,319 @@ Use the browser client in code that runs on the browser, and the server client i
...
+To access Supabase from a Next.js app, you need 2 types of Supabase clients:
+
+1. **Client Component client** - To access Supabase from Client Components, which run in the browser.
+2. **Server Component client** - To access Supabase from Server Components, Server Actions, and Route Handlers, which run only on the server.
+
+Since Next.js Server Components can't write cookies, you need middleware to refresh expired Auth tokens and store them.
+
+The middleware is responsible for:
+
+1. Refreshing the Auth token by calling `supabase.auth.getUser()`.
```
charislam: This should probably be updated to `supabase.auth.getClaims()` for performance now, see this for an example: https://github.com/vercel/next.js/blob/canary/examples/with-supabase/lib/supabase/middleware.ts

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -155,297 +156,319 @@ Use the browser client in code that runs on the browser, and the server client i
...
 
-<Tabs
-  scrollable
-  size="small"
-  type="underlined"
-  defaultActiveId="client-component"
-  queryGroup="environment"
->
-<TabPanel id="client" label="Client-side">
+<Accordion>
```
charislam, line 175: Something here probably needs a margin:

Image

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -155,297 +156,319 @@ Use the browser client in code that runs on the browser, and the server client i
...
-```ts utils/supabase/middleware.ts
-import { createServerClient } from '@supabase/ssr'
-import { NextResponse, type NextRequest } from 'next/server'
+The code adds a [matcher](https://nextjs.org/docs/app/building-your-application/routing/middleware#matching-paths) so the middleware doesn't run on routes that don't access Supabase.
 
-export async function updateSession(request: NextRequest) {
-  let supabaseResponse = NextResponse.next({
-    request,
-  })
+<Admonition type="danger">
```
charislam, line 225: also probably need more margin here: 

Image

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -155,297 +156,319 @@ Use the browser client in code that runs on the browser, and the server client i
...
-        },
-      },
-    }
-  )
+Be careful when protecting pages. The server gets the user session from the cookies, which can be spoofed by anyone.
 
-  // IMPORTANT: Avoid writing any logic between createServerClient and
-  // supabase.auth.getClaims(). A simple mistake could make it very hard to debug
-  // issues with users being randomly logged out.
+Always use `supabase.auth.getUser()` to protect pages and user data.
```
charislam: this should stay `getClaims` (I guess it's the example that needs to change?)

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -155,297 +156,319 @@ Use the browser client in code that runs on the browser, and the server client i
...
+<Admonition type="note">
+
+Note that `cookies` is called before any calls to Supabase, which opts fetch calls out of Next.js's caching. This is important for authenticated data fetches, to ensure that users get access only to their own data.
+
+See the Next.js docs to learn more about [opting out of data caching](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating#opting-out-of-data-caching).
+
+</Admonition>
+
+<$CodeTabs>
+<$CodeSample
```
charislam: I wonder if we need a simpler example app for this purpose, there's a bunch of extraneous styling and components here that gets in the way of understanding the basic concept

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -155,297 +156,319 @@ Use the browser client in code that runs on the browser, and the server client i
...
-   * JWT before returning the session.
-   */
-  event.locals.safeGetSession = async () => {
-    const {
-      data: { session },
-    } = await event.locals.supabase.auth.getSession()
-    if (!session) {
-      return { session: null, user: null }
-    }
+<$CodeSample
```
charislam, line 274: This should probably also be updated for `getClaims`, I don't know if we have a canonical Svelte example 🤔

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
ChrisChinchilla: Yeah, agreed, that might require some more thought…

ChrisChinchilla:
@charislam I have a thread discussing how `getClaim` might not always be the correct answer, but noted and will update accordingly.

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
ChrisChinchilla: @charislam Coming back to this, I have now also realised that this example is completely unrunnable as nowhere in the docs, old or new, do we tell people to install dependencies etc. I am sure I tested this before, but maybe not 🤷

ChrisChinchilla: It could be that the plan was to say more "here's how to add", but I like to have end-to-end runnable examples. I think for now, i will just rework the code used here, but look to add that fuller functionality later.

ChrisChinchilla: @charislam OK, so after digging into this… `getUser` in the instance of Svelte and Next.js is correct, as at a later stage it shows the email address from the user object.

So, a couple of options…

Simplify the example as above and just authenticate. Consider repurposing the `getUser` flow in a different doc if it doesn't already exist. 

Keep the current usage and expand. The full Next.js example does something similar with user data. So could either add similar to the other code examples in this guide (which was my intention in the long term), as it does show other techniques, but it might be good to mention explicitly in the tutorial the difference between the methods.

In either case, standardise the example across all the frameworks in the doc / examples, possibly adding over time as there's time, rather than spending a lot of time updating Astro, Express, etc right now.

From your other comments, I get the feeling you might say option one, simplifying the example?

charislam: I think simplifying the example is probably best, but don't want it to be a blocker. If the other option is faster, I'm ok with that as a short-term fix too.

apps/www/lib/redirects.js
```diff
@@ -1298,6 +1298,18 @@ module.exports = [
     source: '/docs/guides/auth/auth-helpers/auth-ui-overview',
     destination: '/docs/guides/auth/auth-helpers/auth-ui',
   },
+  {
+    permanent: true,
```
charislam: minor nit: on URLs we might resurrect in the short term I like to make the redirects non-permanent, so we can revive them without worrying about cache-busting redirects

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -156,297 +157,145 @@ Use the browser client in code that runs on the browser, and the server client i
...
-<TabPanel id="server" label="Server-side">
+1. Refreshing the Auth token by calling `supabase.auth.getClaims()`.
+2. Passing the refreshed Auth token to Server Components, so they don't attempt to refresh the same token themselves. This is accomplished with `request.cookies.set`.
+3. Passing the refreshed Auth token to the browser, so it replaces the old token. This is accomplished with `response.cookies.set`.
 
-```ts utils/supabase/server.ts
-import { createServerClient } from '@supabase/ssr'
-import { cookies } from 'next/headers'
+{/* TODO: Admonition or etc */}
+The following code samples are for App Router. For help with Pages Router, see the [Next.js Server-Side Auth guide](/docs/guides/auth/server-side/nextjs?queryGroups=router&router=pages).
```
charislam: This page has been removed, right?

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -156,297 +157,145 @@ Use the browser client in code that runs on the browser, and the server client i
...
+    />
+  </$CodeTabs>
+</div>
 
-  // IMPORTANT: Don't remove getClaims()
-  const { data } = await supabase.auth.getClaims()
+### Hook up middleware
 
-  const user = data?.claims
+Create a `utils/supabase` folder at the root of your project, or inside the `./src` folder if you are using one, with a file for each type of client. Then copy the utility functions for each client type.
```
charislam: This should probably go on top of the `createClient` code block above

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -156,297 +157,145 @@ Use the browser client in code that runs on the browser, and the server client i
...
-  queryGroup="environment"
->
-<TabPanel id="hooks" label="Hooks">
+_Never_ trust `supabase.auth.getSession()` inside server code such as middleware. It isn't guaranteed to revalidate the Auth token.
 
-```ts hooks.server.ts
-import { PUBLIC_SUPABASE_URL, PUBLIC_SUPABASE_PUBLISHABLE_KEY } from '$env/static/public'
-import { createServerClient } from '@supabase/ssr'
-import type { Handle } from '@sveltejs/kit'
+It's safe to trust `getClaims()` because it sends a request to the Supabase Auth server every time to revalidate the Auth token.
```
charislam:
```suggestion
It's safe to trust `getClaims()` because it validates the JWT signature against the project's published public keys every time.
```

apps/docs/content/guides/auth/server-side/creating-a-client.mdx
```diff
@@ -843,114 +692,40 @@ app.post("/hello-world", async function (req, res, next) {
...
 </Tabs>
 
 </TabPanel>
 </Tabs>
 
+## Congratulations
+
+You're done! To recap, you've successfully:
+
+- Called Supabase from a Server Action.
```
charislam: This is showing up for all tabs but the bullet points are pretty specific to Next.js

apps/www/lib/redirects.js
ChrisChinchilla: Good call, I wasn't totally sure on how this worked

ChrisChinchilla:
@charislam For the "Congratulations" text, the non-Next and SvelteKit versions need some more work at some point.

mandarini (commented):
From the SDK perspective, these updates look good to me, and they will set the foundation for further improvements down the line. So it's a ✅ from me.

charislam (approved):

PR #38407, opened 2025-09-03
feat(replication): Add ability to copy pipeline id and filter logs by pipeline id

iambriccardo:
This PR adds the following changes:
- The pipeline ID now can be seen by hovering over the name of the destination (useful for debugging purposes).
- When you click on view logs, now the system filters by `pipeline_id` to show the logs of the specific pipeline.
- The button for starting and stopping the pipeline has been improved and it's now more state aware. When there are state transitions it switches to a disabled state.
- The update config in the destination panel is now working correctly, before it was just referencing the value before the last update.

### Screenshots

image

image

image

### DIFF
--- a/apps/studio/components/interfaces/Database/Replication/PipelineStatus.tsx
+++ b/apps/studio/components/interfaces/Database/Replication/PipelineStatus.tsx
@@ -1,7 +1,8 @@
+import { AlertTriangle, Loader2 } from 'lucide-react'
+
 import { useParams } from 'common'
 import { InlineLink } from 'components/ui/InlineLink'
 import { ReplicationPipelineStatusData } from 'data/replication/pipeline-status-query'
-import { AlertTriangle, Loader2 } from 'lucide-react'
 import { PipelineStatusRequestStatus } from 'state/replication-pipeline-request-status'
 import { ResponseError } from 'types'
 import { cn, Tooltip, TooltipContent, TooltipTrigger, WarningIcon } from 'ui'
@@ -131,6 +132,12 @@ export const PipelineStatus = ({
 
   const statusConfig = getStatusConfig()
 
+  const pipelineLogsUrl = pipelineId
+    ? `/project/${ref}/logs/etl-replication-logs?f=${encodeURIComponent(
+        JSON.stringify({ pipeline_id: pipelineId })
+      )}`
+    : `/project/${ref}/logs/etl-replication-logs`
+
   return (
     <>
       {isLoading && <ShimmeringLoader />}
@@ -157,19 +164,7 @@ export const PipelineStatus = ({
             {['unknown', 'failed'].includes(pipelineStatus?.name ?? '') && (
               <>
                 {' '}
-                Check the{' '}
-                <InlineLink
-                  href={
-                    pipelineId
-                      ? `/project/${ref}/logs/etl-replication-logs?f=${encodeURIComponent(
-                          JSON.stringify({ pipeline_id: pipelineId })
-                        )}`
-                      : `/project/${ref}/logs/etl-replication-logs`
-                  }
-                >
-                  logs
-                </InlineLink>{' '}
-                for more information.
+                Check the <InlineLink href={pipelineLogsUrl}>logs</InlineLink> for more information.
               </>
             )}
           </TooltipContent>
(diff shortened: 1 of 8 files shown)


apps/studio/components/interfaces/Database/Replication/RowMenu.tsx
```diff
@@ -110,17 +131,56 @@ export const RowMenu = ({
...
+          {(() => {
+            // Disable actions during transient request states
+            if (requestStatus !== PipelineStatusRequestStatus.None) {
+              return (
+                <DropdownMenuItem className="space-x-2 opacity-50" disabled>
+                  <Ban size={14} />
+                  <p>Action unavailable</p>
+                </DropdownMenuItem>
+              )
+            }
```
joshenlim: hmm either we hide this, or disable the dropdown menu item with a tooltip

we don't really have a pattern of "action unavailable" around the dashboard

iambriccardo: I am fine with it, just to be consistent.

apps/studio/components/interfaces/Database/Replication/RowMenu.tsx
```diff
@@ -110,17 +131,56 @@ export const RowMenu = ({
...
+              <DropdownMenuItem
+                className={`space-x-2 ${disabled ? 'opacity-50 cursor-not-allowed' : ''}`}
+                disabled={disabled}
+                onClick={onClick}
+              >
+                {icon}
+                <p>{label}</p>
+              </DropdownMenuItem>
+            )
+          })()}
```
joshenlim: ideally i'd like to avoid this way of rendering components from a consistency POV - we can declare the components outside with a `const`

apps/studio/data/replication/update-destination-pipeline-mutation.ts
```diff
@@ -97,11 +97,17 @@ export const useUpdateDestinationPipelineMutation = ({
...
 
         await Promise.all([
+          // Invalidate lists
           queryClient.invalidateQueries(replicationKeys.destinations(projectRef)),
           queryClient.invalidateQueries(replicationKeys.pipelines(projectRef)),
+          // Invalidate item-level caches used by the editor panel
+          queryClient.invalidateQueries(
+            replicationKeys.destinationById(projectRef, destinationId)
+          ),
+          queryClient.invalidateQueries(replicationKeys.pipelineById(projectRef, pipelineId)),
```
joshenlim, line 110: could Promise.all these to run them in parallel

apps/studio/components/interfaces/Database/Replication/ReplicationPipelineStatus.tsx
```diff
@@ -178,16 +178,50 @@ export const ReplicationPipelineStatus = () => {
...
+              href={`/project/${projectRef}/logs/etl-replication-logs${
+                pipelineId
+                  ? `?f=${encodeURIComponent(JSON.stringify({ pipeline_id: pipelineId }))}`
+                  : ''
+              }`}
+            >
+              View logs
+            </Link>
           </Button>
+          {(() => {
```
joshenlim: let's avoid doing this from a consistency POV

apps/studio/components/interfaces/Database/Replication/PipelineStatus.tsx
```diff
@@ -129,6 +132,12 @@ export const PipelineStatus = ({
 
   const statusConfig = getStatusConfig()
 
+  const pipelineLogsUrl = pipelineId
+    ? `/project/${ref}/logs/etl-replication-logs?f=${encodeURIComponent(
+        JSON.stringify({ pipeline_id: pipelineId })
+      )}`
```
joshenlim, line 138: (Non-blocking) JFYI this `f` param support has no UI support yet as this is dependent on the Universal Filter Bar that's currently not live atm ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Settings/Logs/LogsPreviewer.tsx#L241))

I can verify that this is working still though via the network request that gets sent, but there's no way to change the filter thereafter on the logs page with the existing UI on prod atm

joshenlim (approved):
Verified changes on staging! LGTM 🙂 🙏

PR #38411, opened 2025-09-03
update handleError to pass sentry context

jordienr:


apps/studio/data/fetchers.ts
```diff
@@ -135,15 +136,17 @@ export const {
 
 type HandleErrorOptions = {
   alwaysCapture?: boolean
+  sentryContext?: {
```
jordienr: wanted to type this with the same type Sentry uses for the second argument of captureException but they do not export it anywhere 🤷

apps/studio/data/fetchers.ts
```diff
@@ -135,15 +136,17 @@ export const {
...
 
-export const handleError = (
-  error: unknown,
-  options: HandleErrorOptions = { alwaysCapture: false }
-): never => {
+export const handleError = (error: unknown, options: HandleErrorOptions = {}): never => {
   if (error && typeof error === 'object') {
     if (options.alwaysCapture) {
-      Sentry.captureException(error)
+      Sentry.captureException(error, {
```
jordienr: not loving the alwaysCapture argument or having 2 places where we capture exceptions in the handleError fn but want to avoid big changes now

ivasilov (approved):

PR #38416, opened 2025-09-03
storage dnd

fsansalvadore:
Enable moving files and folders around in the Storage UI via drag and drop.

- [x] move files and folders onto different columns
- [x] move files and folders onto folder items
- [x] drag and drop multiple files
- also fixed the `Public` badge from moving when selecting a bucket on the left.

https://github.com/user-attachments/assets/a90bfab3-e6f7-4c0f-b781-fae181368188


apps/studio/state/storage-explorer.tsx
```diff
@@ -1417,6 +1532,406 @@ function createStorageExplorerState({
...
+          .upload(targetFullPath + '/.folder-placeholder', placeholderContent, {
+            upsert: false,
+          })
+
+        if (uploadError) {
+          throw new Error(`Failed to create target folder: ${uploadError.message}`)
+        }
+
+        // Recursively discover all files and folders in the source directory
+        const getAllFilesRecursively = async (
```
joshenlim, line 1689: this seems the same (?) as what `getAllItemsAlongFolder` is trying to do maybe? if so, could we de-dupe?

apps/studio/state/storage-explorer.tsx
```diff
@@ -1417,6 +1532,406 @@ function createStorageExplorerState({
...
+        // Check if selectedBucket and projectRef are available
+        if (!state.selectedBucket || !state.selectedBucket.id) {
+          throw new Error('Selected bucket is not available')
+        }
+
+        if (!state.projectRef) {
+          throw new Error('Project reference is not available')
+        }
+
+        // Create the new folder by uploading a placeholder file
```
joshenlim, line 1675: we don't need to do this actually - folders are an FE concept
so if there's a path say '/path1/path2/file.tsx` then path 1 and path 2 are naturally folders

we only need to add a placeholder file if we want an "empty" folder

joshenlim:
JFYI converted this to draft:
- There's a bug identified with either the `move` or `list` endpoint from storage when moving folders
- The `move` endpoint currently only supports singular file moves, which if we were to move folders that have many files, there's a chance of DDOS-ing our own API since it'll be one request per file

^ Both of these are blocked on storage atm

The `moveFolderWithContents` function in `storage-explorer` can also be streamlined a bit more (some parts in there i think are unnecessary) but I reckon we can refactor when the above 2 are unblocked (otherwise it's quite hard to test the correctness too)

PR #38430, opened 2025-09-04
Fix missing loading prop in ConfirmationModal for disabling extension

joshenlim:
Fixes FE-1830

## Context

Confirmation modal for disabling extension gets stuck in a loading state if there are errors - was just missing a `loading` prop from the disable extension RQ mutation


saltcod (approved):
Tested on staging, looks good!

PR #38426, opened 2025-09-04
fix: estimate time needed to create data branch

sweatybridge:
Show notice on the estimated time to create a data branch.

Screenshot 2025-09-04 at 3 15 01 PM

PR #38454, opened 2025-09-05
only trigger request to /clone in create branch modal when modal is o…

joshenlim:
## Context

We added a check to the /clone endpoint in CreateBranchModal back in [this PR](https://github.com/supabase/supabase/pull/38426), opting to only fire that request when the modal is opened, otherwise the request gets triggered when landing on any project page since the modal is rendered in the header


kevcodez (approved):

joshenlim:

PR #38458, opened 2025-09-05
Swap useCheckPermissions with useAsyncCheckPermissions part 4: Storage

joshenlim:
## Context

Following up from the previous [PR](https://github.com/supabase/supabase/pull/37899), copying the PR description for ease of reference:

We'll be favouring the use of useAsyncCheckProjectPermissions over useCheckPermissions as it'll allow us to access the loading state of the permissions. We should then also factor that in to prevent any premature "no permissions" UI rendering

Opting to do this in smaller PRs since its a bit nuanced

We can only swap useCheckPermissions with useAsyncCheckProjectPermissions for project permissions, organization permissions will still need to use the former for now until we introduce a new useAsyncCheckOrgPermissions
Also need to add loading states where missing for pages that are currently prematurely showing the no perms UI

## Changes involved
- Swap `useCheckPermissions` with `useAsyncCheckProjectPermissions` in more files
  - Note: adjusted loading behaviour for the following
    - `StorageSettings/S3Connection.tsx`
- (Unrelated) Remove more `export default` for Storage related files

## Affected UI
Probably just need a quick skim on these pages, mainly just to make sure that there's (1) No incorrect "perms denied" UI, if you're the owner of the org you should have access to everything (2) No pre-mature "perms denied UI" (can test by refreshing on that page)
- Storage + Storage Settings

### DIFF
--- a/apps/studio/components/interfaces/Storage/StorageExplorer/ColumnContextMenu.tsx
+++ b/apps/studio/components/interfaces/Storage/StorageExplorer/ColumnContextMenu.tsx
@@ -3,7 +3,7 @@ import { Item, Menu, Separator, Submenu } from 'react-contexify'
 import 'react-contexify/dist/ReactContexify.css'
 
 import { PermissionAction } from '@supabase/shared-types/out/constants'
-import { useCheckPermissions } from 'hooks/misc/useCheckPermissions'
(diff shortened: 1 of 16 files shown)


charislam (approved):
Checked preview and LGTM

PR #38461, opened 2025-09-05
Add feature flag to disable /chore requests trigger in CreateBranchModal

joshenlim:
## Context

Following up from the previous [PR here](https://github.com/supabase/supabase/pull/38454) - we've reduced the number of requests going to `/clone` by only triggering the request when the Create branch modal is opened.

This is just adding a kill switch to completely prevent requests to `/clone` in case we need it (meant to be short lived and can be removed once API optimization is done)

Enabling the feature flag will:
- Stop requests to `/clone`
- Remove the physical backups check for "Include data" toggle (FE will let the API do any required validation - if there is any?)
  - Toggle will no longer be disabled

### DIFF
--- a/apps/studio/components/interfaces/BranchManagement/CreateBranchModal.tsx
+++ b/apps/studio/components/interfaces/BranchManagement/CreateBranchModal.tsx
@@ -65,7 +65,7 @@ export const CreateBranchModal = () => {
   const allowDataBranching = useFlag('allowDataBranching')
   // [Joshen] This is meant to be short lived while we're figuring out how to control
   // requests to this endpoint. Kill switch in case we need to stop the requests
-  const disableBackupsCheck = useFlag('DisableBackupsCheckInCreatebranchmodal')
+  const disableBackupsCheck = useFlag('disableBackupsCheckInCreatebranchmodal')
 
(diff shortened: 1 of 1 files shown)


alaister (approved):
LGTM!

apps/studio/components/interfaces/BranchManagement/CreateBranchModal.tsx
```diff
@@ -63,6 +63,9 @@ export const CreateBranchModal = () => {
 
   const gitlessBranching = useIsBranching2Enabled()
   const allowDataBranching = useFlag('allowDataBranching')
+  // [Joshen] This is meant to be short lived while we're figuring out how to control
+  // requests to this endpoint. Kill switch in case we need to stop the requests
+  const disableBackupsCheck = useFlag('disableBackupsCheckInCreatebranchmodal')
```
raulb, line 68: Nitpick: 

```suggestion
  const disableBackupsCheck = useFlag('disableBackupsCheckInCreateBranchModal')
```

joshenlim: ^ this is dependent on config cat itself, so wont be able to change here

PR #38465, opened 2025-09-05
feat: ensure customer puts down address on upgrade

kevcodez:
For customers who have not yet put down a billing address, we want to ensure that they put down an address when they upgrade their plan.

Customers can also no longer unset their billing address.


ecktoteckto (approved):
Looks good on preview

staaldraad (approved):

PR #38470, opened 2025-09-05
feat(studio): simplify query performance filters

kemaldotearth:

YES

This PR is simplifying our Query Performance tabs filter bar. We have a redundant sort dropdown as well as a weird way to find queries in the table. Adding some simplicity to it and removing some noise.


charislam (approved):
LGTM

apps/studio/components/interfaces/QueryPerformance/hooks/useIndexInvalidation.ts
```diff
@@ -34,7 +34,6 @@ export function useIndexInvalidation() {
   const roles = router?.query?.roles ?? []
 
   const queryPerformanceQuery = useQueryPerformanceQuery({
-    searchQuery,
```
charislam: Nit: Can `searchQuery` be removed from `useQueryStates` above? Looks like it's never used.

(Sorry, can't put a suggestion on that line because GitHub UI)

charislam (commented):
Question about the change to local search: I really like the new behavior, but I'm wondering if anyone actually depends on the network search? E.g., I have a very specific query (or set of queries) I want to track performance for, they don't actually show up in the top 20, but in the old behavior, I can query for them from the backend. In the new behavior, I can't fetch them at all.

apps/studio/components/interfaces/QueryPerformance/hooks/useIndexInvalidation.ts
kemaldotearth: Good point re: search, we can revert it back to finding via query. To me it felt a little slow and unreliable to find things inline, but maybe not such an issue? Let me give it a try (and yes, we can remove `searchQuery`!)

kemaldotearth: Try again @charislam - bonus you can now also copy over the url (like before) and share the filtered search too.

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx
```diff
@@ -35,21 +27,15 @@ export const QueryPerformanceFilterBar = ({
...
-  const defaultFilterRoles = router.query.roles ? (router.query.roles as string[]) : []
-  const defaultSortByValue = router.query.sort
-    ? ({ column: router.query.sort, order: router.query.order } as QueryPerformanceSort)
-    : undefined
+  const [{ search: searchQuery }, setSearchParams] = useQueryStates({
+    search: parseAsString.withDefault(''),
+  })
 
-  const [searchInputVal, setSearchInputVal] = useState(defaultSearchQueryValue)
+  const defaultFilterRoles = router.query.roles ? (router.query.roles as string[]) : []
```
joshenlim: nit: we should use `nuqs` to manage URL state here

would look something like 
```const [defaultFilterRoles] = useQueryState('roles', parseAsArrayOf(parseAsString))```

File: apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx [hunk 3]
```diff
@@ -59,73 +45,39 @@ export const QueryPerformanceFilterBar = ({
...
     <div className="px-6 py-2 bg-surface-200 border-t -mt-px flex justify-between items-center">
       <div className="flex items-center gap-x-4">
         <div className="flex items-center gap-x-2">
-          <p className="text-xs prose">Filter by</p>
+          <Input
+            size="tiny"
+            autoComplete="off"
+            icon={<Search size={12} />}
+            value={searchQuery}
+            onChange={(e: any) => onSearchQueryChange(e.target.value)}
```
joshenlim: search needs some debouncing - it's currently firing a network request on each character input which could cause throttle exceptions pretty easily

```
import { useDebounce } from '@uidotdev/usehooks'

const debouncedSearch = useDebounce(search, 500)
```

joshenlim: also avoid usage of `any` as much as possible

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
```diff
@@ -168,7 +168,21 @@ export const QueryPerformanceGrid = ({ queryPerformanceQuery }: QueryPerformance
...
+
+    if (!search || search.trim() === '') {
+      return allData
+    }
+
+    const searchTerm = search.toLowerCase()
+    return allData.filter((row: any) => {
+      // Search in the query text
+      const queryText = row.query?.toLowerCase() || ''
+      return queryText.includes(searchTerm)
```
joshenlim: just to make sure i got the context correct - we're shifting to client side searching because this UI isn't paginated on the server side right? there's no infinite scrolling here? (i think that's the case from skimming the [code](https://github.com/supabase/supabase/blob/feat/query-perf-simplify-filters/apps/studio/components/interfaces/Reports/Reports.constants.ts#L396) in `Reports.constants.ts` but just to double check)

^ If that's the case as well I'm wondering if it might be worth mentioning somewhere in the UI that we're only showing the top 20 rows (to indicate there's no pagination)

joshenlim: There's a small bug here: with this we're currently actually doing _both_ server and client side searching ([ref](https://github.com/supabase/supabase/blob/feat/query-perf-simplify-filters/apps/studio/pages/project/%5Bref%5D/advisors/query-performance.tsx#L46)) as the search query is also being passed into the RQ hook that retrieves the report data

so we'll need to remove the search query from the hook there too

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx
```diff
@@ -59,73 +45,39 @@ export const QueryPerformanceFilterBar = ({
...
 
-  function getSortButtonLabel() {
-    if (sortByValue?.order === 'desc') {
-      return 'Sorted by latency - high to low'
-    } else {
-      return 'Sorted by latency - low to high'
-    }
-  }
-
-  const onSortChange = (order: 'asc' | 'desc') => {
```
joshenlim, line 86: Unrelated, but I actually didn't realise that the columns were sortable by clicking on the header - wondering if we should add a tooltip or some UI indication

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
joshenlim: side thought - im thinking this could get confusing from a UX POV (although i might be over worrying too)

sorting and filtering by roles are done on the server side, but searching is done on the client side (based on the current data that's shown)

is there some context that i might be missing on why we want to shift searching by text to client side?

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx
kemaldotearth: I'll do a follow up updating the UI of the column headers. I think we should use the same pattern as Tables editor cols etc. I used this also in the Query Insights.

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx
kemaldotearth: Done and done.

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx
kemaldotearth: next/router is yesterdays news, seeee ya.

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
kemaldotearth: I think we're back to that with the latest changes. Server side search, last 20 results.

joshenlim (approved):
Verified changes locally

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx
kemaldotearth: See here @joshenlim #38523

PR #38473, opened 2025-09-05
feat(replication): Add UI for updating pipeline versions

iambriccardo:
This PR adds a new UI for updating pipeline images. The new UI works as follows:
- The version of each pipeline is fetched. The backend will return a `new_version` field in case an update is available.
- If an update is available a new button will pop up in the UI and will prompt the user to update a pipeline via a dialog which clearly shows the change in image version.

## Screenshots

image

image


apps/ui-library/public/llms.txt
joshenlim, line 1: seems unnecessary - will revert

apps/studio/data/replication/pipeline-version-query.ts
```diff
@@ -0,0 +1,60 @@
...
+import { replicationKeys } from './keys'
+
+type ReplicationPipelineVersionParams = { projectRef?: string; pipelineId?: number }
+
+export interface ReplicationPipelineVersion {
+  id: number
+  name: string
+}
+
+export interface ReplicationPipelineVersionResponse {
```
joshenlim: should use the API types instead (will push up changes)

apps/studio/data/replication/update-pipeline-version-mutation.ts
```diff
@@ -0,0 +1,73 @@
+import { useMutation, UseMutationOptions, useQueryClient } from '@tanstack/react-query'
+import { toast } from 'sonner'
+
+import type { ResponseError } from 'types'
+import { replicationKeys } from './keys'
+import { handleError, post } from 'data/fetchers'
+
+export type UpdatePipelineVersionParams = {
```
joshenlim: avoid exporting unless being used outside of this file

joshenlim (approved):
Verified changes on staging preview - generally LGTM!

JFYI there's just one last piece of the UI that i've yet to test which is upgrading the pipeline from within the pipeline page, if I could trouble you to help test that before merging 🙂 🙏

PR #38485, opened 2025-09-06
chore: clean up common exports

charislam:
Clean up the exports from the common package.

This was motivated by docs needing to import from `common/enabled-features` in order to bypass the React peer dependency on other parts of common. I didn't like importing from an internal file, so I cleaned up common's exports over the entire monorepo to provide a proper API for this package:

- Default export is `./index.tsx` which contains most things
- Assets directory is exported individually because these are fonts/images, etc.
- `common/constants` is exported because I imagine there are use cases where we also don't want to bring in a bunch of cruft to a script, etc., while just getting constants
- `common/enabled-features` is exported for the same reason
- `common/hooks` just because it's commonly used
- Can no longer directly import from an internal file outside of these export points


joshenlim (approved):
Code changes LGTM from FE - minimally did a brief smoke test on dashboard, www, and docs

PR #38499, opened 2025-09-07
Fix adding bucket policy to schema

SaxonF:
Removes check for bucket to add new policy to storage.objects or storage.buckets schema


joshenlim (approved):

PR #38501, opened 2025-09-08
Fix roles

joshenlim:
## Context

There's a bug introduced [here](https://github.com/supabase/supabase/pull/38428) that's affecting updating of roles for org-scoped roles, whereby the introduced `filter` in `formatMemberRoleToProjectRoleConfiguration` was incorrectly filtering out org-scoped roles, resulting in users seeing this callout when editing roles for a user that only has org-scoped roles
image

## To test
- [ ] Verify that you can change the role of a user with an org-scoped role
- [ ] Verify for a user with project-scoped roles, if one of the project that the user has access to was deleted, you can still update that user's role thereafter


alaister (approved):
LGTM!

PR #38511, opened 2025-09-08
fix: estimate data branch cost before creation

sweatybridge:

docs update

## What is the new behavior?

Present estimated compute and disk cost in create branch modal to avoid bill shock.

Only affects data branch.

## Additional context

Screenshot 2025-09-08 at 5 12 08 PM


apps/studio/components/interfaces/BranchManagement/CreateBranchModal.tsx
```diff
@@ -363,7 +391,7 @@ export const CreateBranchModal = () => {
                       label={
                         <>
                           <Label className="mr-2">Include data</Label>
-                          {!disableBackupsCheck && (isLoadingCloneBackups || noPhysicalBackups) && (
+                          {noPhysicalBackups && (
```
charislam, line 387: I think this is the same as the old behavior, but there's a slightly confusing situation here. Since:

```
const noPhysicalBackups = isErrorCloneBackups || isLoadingCloneBackups
```

You also get the warning badge transiently while loading. This takes several seconds so it's quite confusing if you know you have PITR turned on

https://github.com/user-attachments/assets/17dc67b9-671b-49d4-b464-7bc55cf09805

apps/studio/components/interfaces/BranchManagement/CreateBranchModal.tsx
```diff
@@ -100,10 +109,30 @@ export const CreateBranchModal = () => {
...
+      ? (cloneBackups.target_compute_size as DesiredInstanceSize)
+      : undefined
+  // Ignore failures fetching disk attributes since it only affects cost estimation
+  const { data: disk } = useDiskAttributesQuery(
+    { projectRef },
+    {
+      enabled: showCreateBranchModal && !disableBackupsCheck,
+    }
   )
+  const projectDiskAttributes = disk?.attributes ?? {
```
charislam, line 164: Rather than having defaults of 0 if the disk attributes haven't loaded/have errored, should we make a separate loading/error message? Otherwise seeing that your disk size is 0 GB is a very confusing situation + we could potentially be misleading the user about estimated costs

apps/studio/components/interfaces/BranchManagement/CreateBranchModal.tsx
sweatybridge: Managed to replace the slow query with a simpler disk size estimation. We will eventually follow the project disk size on branches instead of over provisioning.

joshenlim:
JFYI updating copy of the modal as such
- Highlight additional costs in the header + Indicate what the costs are for clarity (disk size + compute)

image

apps/studio/components/interfaces/BranchManagement/CreateBranchModal.tsx
joshenlim: Added loading states!
image

And error state too - so we won’t block the creation of branches, but just indicate that we cant estimate the cost
image

joshenlim:
Adding one last bit here to improve billing transparency
image

joshenlim (approved):
Verified changes on staging preview1

PR #38516, opened 2025-09-08
chore: customizable dashboard auth

alaister:
To test (locally):
- toggle the enabled features in different ways to make sure the different options work
- add a custom sign in provider


joshenlim (approved):
Verified changes locally + verified on staging preview that the affected pages are status quo

PR #38523, opened 2025-09-08
feat(studio): update query performance table ui

kemaldotearth:

YES

This updates the Query Performance table to use some of our newer UI and fix things that aren't currently optimal. Changes introduced are as follows:

- Column header styling brushed up.
- Column headers now sortable as data grids elsewhere are.
- Numbers aligned left for legibility.
- Improved syntax highlighting and how it caps in the cell.
- Cells given minimum widths.
- Columns now ordered by more relevant stats first.
- Fixed Index Advisor popover not being accessible. 
- Fixed total time returning string instead of number.

| Before | After |
|--------|--------|
| Screenshot 2025-09-08 at 21 48 09 | Screenshot 2025-09-08 at 21 47 36 |


apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
```diff
@@ -54,24 +53,66 @@ export const QueryPerformanceGrid = ({ queryPerformanceQuery }: QueryPerformance
...
+                    icon={<ChevronDown size={14} className="text-foreground-muted" />}
+                  />
+                </DropdownMenuTrigger>
+                <DropdownMenuContent align="end" className="w-48">
+                  <DropdownMenuItem
+                    onClick={() => {
+                      setSort({ column: col.id, order: 'asc' })
+                      router.push({
+                        ...router,
+                        query: { ...router.query, sort: col.id, order: 'asc' },
```
joshenlim: reckon we use `nuqs` to handle URL state

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
```diff
@@ -54,24 +53,66 @@ export const QueryPerformanceGrid = ({ queryPerformanceQuery }: QueryPerformance
...
 
   const columns = QUERY_PERFORMANCE_REPORTS[reportType].map((col) => {
+    const nonSortableColumns = ['query']
+
     const result: Column<any> = {
       key: col.id,
       name: col.name,
       resizable: true,
       minWidth: col.minWidth ?? 120,
+      sortable: !nonSortableColumns.includes(col.id),
```
joshenlim, line 57: if we're introducing sorting of column order - reckon we should probably also save the sort order in local storage, otherwise the sort order will keep getting reset on each tab session (or even navigating away then back)

happy to address this in another PR

joshenlim:
Small issue here: Noticed that when opening the dropdown menu here it immediately closes at first, but subsequent clicks are alright

https://github.com/user-attachments/assets/b127b20a-6061-42fe-8550-5e6568d8808e

Also probably good to have a UI indication on the sort of the column? Currently apart from looking at the URL there's no clear indicator on which column its sorted on, nor the sort order
image

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
joshenlim: btw this might not be working - im trying on the staging preview but im not able to sort the column order by dragging them

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
kemaldotearth: Done and done.

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
kemaldotearth: AFAIK the columns being draggable were disabled on the data grids for Query Performance and Advisors. We can handle in a separate ticket I guess. I have a follow up to remove the tabs and merge the columns together here, so let's revisit.

Similarly, lets maybe leave localstorage until we merge all the columns together?

joshenlim: oh oh oh i'm so sorry i understood this change totally wrong - had the impression that the `sortable` property was referring to the "sorting of column orders" like dragging the columns around to sort between them, completely separate functionality from what you're doing in this PR which is "sorting on the column itself" 🤦 🤦 

happy to omit local storage in this PR as well 🙏  (in fact maybe we can just leave it out for now unless we get feedback to persist sorting)

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
```diff
@@ -27,51 +29,77 @@ import {
...
+              <p className="!text-foreground font-medium">{col.name}</p>
+              {col.description && (
+                <p className="text-foreground-lighter font-normal">{col.description}</p>
+              )}
             </div>
-            {sort?.column === col.id && (
-              <>{sort.order === 'desc' ? <ArrowDown size={14} /> : <ArrowUp size={14} />}</>
+
+            {isSortable && (
+              <DropdownMenu>
```
joshenlim, line 72: btw maybe dropdown menu isn't the right component to use here - select might be better? than we have the check indication beside the selected value too
image

kemaldotearth: Not sure Select applies here as we can't change the icon? Took inspiration from our Table Editor. I do agree it needs something, so I've highlighted the sorting that's selected in the dropdown. Same effect more or less. 

Screenshot 2025-09-09 at 18 18 59

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
```diff
@@ -27,51 +29,83 @@ import {
...
+                    onClick={() => {
+                      setSortConfig(col.id, 'asc')
+                    }}
+                    className={cn(
+                      'flex gap-2',
+                      sort?.column === col.id && sort?.order === 'asc' && 'text-foreground'
+                    )}
+                  >
+                    <ArrowUp size={14} />
+                    Sort Ascending
```
dnywh, line 93: Minor nit: make sentence case like other buttons/actions? I.e. `Sort ascending`. Ditto for what’s currently `Sort Descending`

dnywh: Hmm, I see this elsewhere like on our current sorting for Table Editor

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx
```diff
@@ -98,6 +101,12 @@ export const QueryPerformanceFilterBar = ({
             onSaveFilters={onFilterRolesChange}
             className="w-56"
           />
+
+          {sort && (
+            <Button size="tiny" type="link" icon={<X size={12} />} onClick={clearSort}>
+              Clear sort
+            </Button>
+          )}
```
dnywh: I don’t see any visual feedback on _how_ the table is currently sorted. Should that be here?

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
```diff
@@ -27,51 +29,83 @@ import {
...
       resizable: true,
       minWidth: col.minWidth ?? 120,
+      sortable: !nonSortableColumns.includes(col.id),
       headerCellClass: 'first:pl-6 cursor-pointer',
       renderHeaderCell: () => {
+        const isSortable = !nonSortableColumns.includes(col.id)
+
         return (
-          <div
-            className="flex items-center justify-between font-mono font-normal text-xs w-full"
```
dnywh, line 66: Agree that getting rid of `font-mono` looks way better. Should we do this across all full-width tables? E.g. _Security Advisor_ and _Performance Advisor_

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
joshenlim: yeah we probably don't need the icons tbh - but can revisit this again, prioritise on consistency across the dashboard how we'd wanna handle sort dropdowns like these

joshenlim: e.g for storage we have this UI for sorting
image

joshenlim (approved):
Verified changes in staging preview!

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx
```diff
@@ -98,6 +100,20 @@ export const QueryPerformanceFilterBar = ({
...
           />
+
+          {sort && (
+            <div className="text-xs border rounded-md px-2.5 py-1 h-[26px] flex items-center gap-x-2">
+              <p>
+                Sort: {sort.column} <span className="text-foreground-lighter">{sort.order}</span>
+              </p>
+              <Tooltip>
+                <TooltipTrigger onClick={clearSort}>
+                  <X size={14} className="text-foreground-light hover:text-foreground" />
```
fsansalvadore, line 111: @kemaldotearth noticed the text in this element gets stacked if viewport too narrow.

Screenshot 2025-09-10 at 18 06 08

Might want to avoid it going on multiple lines? 
And maybe that whole actions bar could be horizontally scrollable if the width exceeds the available space?

fsansalvadore (approved):
Approving as everything looks good overall, just left one comment @kemaldotearth

apps/studio/components/interfaces/QueryPerformance/QueryPerformanceFilterBar.tsx
kemaldotearth: Good spot, fixing before push!

fsansalvadore: another component where I used the horizontal scroll for responsiveness for reference:

https://github.com/supabase/supabase/blob/master/apps/studio/components/layouts/ProjectLayout/LayoutHeader/LayoutHeader.tsx#L131C1-L132C1

https://github.com/user-attachments/assets/b9c08f92-3b81-472b-b369-fc2cb8d210a9

kemaldotearth: Pushed effectively same fix, thanks! 🙇

PR #38524, opened 2025-09-08
fix: usage page select project ref

alaister:


kemaldotearth (approved):
LGTM 🙇

joshenlim (approved):
Verified on staging preview usage looks good - network requests isn't sending a project ref when all projects is selected

PR #38548, opened 2025-09-09
fix(types): update branching types for api route changes

avallete:

- Fixes the types for `branching` routes for the new `branch_id_or_ref` params
- Regenerated api types.


joshenlim (approved):
sanity checked preview by creating and deleting branches

PR #38550, opened 2025-09-09
Auth Report tests and improvements

jordienr:
- No change in functionality
- moves auth report to new typesafe config ✅
- some clean up ✅ 
- adds data transformation tests ✅ 
- fixes bug where useFilltimeseries wasnt working (receiving wrong timestamp key) ✅ 
- fixes bug where the Tooltip would render incorrect dates ✅ 
- adds Report Settings dropdown ✅ 

## To test
- Create some users / auth events
- Go to report
- Metrics make sense


apps/studio/components/interfaces/Reports/v2/ReportChartUpsell.tsx
```diff
@@ -5,7 +5,16 @@ import { LogChartHandler } from 'components/ui/Charts/LogChartHandler'
 import { ReportConfig } from 'data/reports/v2/reports.types'
 import { Button, Card, cn } from 'ui'
 
-export function ReportChartUpsell({ report, orgSlug }: { report: ReportConfig; orgSlug: string }) {
+export function ReportChartUpsell({
+  report,
+  orgSlug,
+}: {
+  report: {
```
jordienr, line 12: removing unused props so typescript wont ask for them

apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
```diff
@@ -100,10 +107,6 @@ export const ReportChartV2 = ({
       >
         {isLoadingChart ? (
           <Loader2 className="size-5 animate-spin text-foreground-light" />
-        ) : showEmptyState ? (
```
jordienr, line 103: charts have an empty state already, prefer to use that

apps/studio/components/ui/Charts/ComposedChart.tsx
```diff
@@ -416,6 +416,7 @@ export function ComposedChart({
                     attributes?.find((a) => a.attribute === attribute?.name)?.label ||
                     attribute?.name
                   }
+                  maxBarSize={24}
```
jordienr, line 422: makes the bars less thicc when there are little data points

apps/studio/data/reports/auth-charts.ts
jordienr, line 1: removed in favor of the type safe report config

apps/studio/data/reports/auth-report-query.ts
jordienr, line 1: removed in favor of the type safe report config

apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
```diff
@@ -112,7 +115,7 @@ export const ReportChartV2 = ({
           <div className="w-full">
             <ComposedChart
               attributes={dynamicAttributes}
-              data={finalChartData}
+              data={filledChartData}
```
charislam, line 115: is `finalChartData` never used anymore? looks like we could remove it

apps/studio/data/reports/v2/auth.config.ts
```diff
@@ -0,0 +1,478 @@
...
+  order by timestamp desc
+      `
+  },
+}
+
+type AuthReportFilters = {
+  status_code: NumericFilter | null
+}
+
+function filterToWhereClause(filters?: AuthReportFilters): string {
```
charislam, line 158: is this used anywhere?

apps/studio/data/reports/v2/auth.config.ts
```diff
@@ -0,0 +1,478 @@
...
+ * 2. Creates a consistent data structure with `period_start` as the time axis
+ * 3. Initializes all chart attributes to 0, then populates actual values
+ * 4. Sorts timestamps chronologically for proper chart ordering
+ *
+ * @param rawData - Raw analytics data from backend queries containing timestamp and count fields
+ * @param attributes - Chart attribute configuration defining what metrics to display
+ * @returns Formatted data object with consistent time series data and chart attributes
+ */
+export function defaultAuthReportFormatter(
+  rawData: any,
```
charislam:
```suggestion
  rawData: unknown,
```

nit: I think it's better to prefer `unknown` to `any` to force ourselves to defensively code but if it's too much work to change it here then yolo 🤷🏼‍♀️

apps/studio/data/reports/v2/auth-report.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    const result = defaultAuthReportFormatter(data, attributes)
+
+    expect(result).toEqual({
+      data: [{ period_start: '2021-01-01', ActiveUsers: 1 }],
+      chartAttributes: attributes,
+    })
+  })
+
+  it('should format the data correctly with multiple attributes', () => {
```
charislam, line 22: looks like this is testing multiple data points, not multiple attributes, unless I'm misunderstanding

apps/studio/data/reports/v2/auth.config.ts
```diff
@@ -0,0 +1,478 @@
...
+  const data = Array.from(timestamps)
+    .sort()
+    .map((timestamp) => {
+      const point: any = { period_start: timestamp }
+      chartAttributes.forEach((attr) => {
+        point[attr.attribute] = 0
+      })
+      const matchingPoints = result.filter((p: any) => p.timestamp === timestamp)
+      matchingPoints.forEach((p: any) => {
+        point[attributes[0].attribute] = p.count
```
charislam: what's the expected behavior for multiple attributes? I had a bit of trouble testing on the preview, but from here it looks like all the attributes are zeroed, and then only the first one is set to the returned count?

apps/studio/data/reports/v2/auth.config.ts
jordienr: will be used in next PR, decided to keep filters out of this PR since it was quite big already

apps/studio/components/interfaces/Reports/v2/ReportChartV2.tsx
jordienr: correct, cleaned it up

apps/studio/data/reports/v2/auth.config.ts
jordienr: Exactly that. Add all attributes zeroed to all events.

Fixed this test in hopes to explain better what its trying to achieve.
I had some old code mixed up still that made it confusing.

```typescript
it('should format the data correctly with multiple attributes', () => {
    const data = {
      result: [
        { timestamp: String(timestamp), active_users: 1 },
        { timestamp: String(timestamp + 1), sign_in_attempts: 2 },
      ],
    }

    const attributes = [
      { attribute: 'active_users', label: 'Active Users' },
      { attribute: 'sign_in_attempts', label: 'Sign In Attempts' },
    ]

    const result = defaultAuthReportFormatter(data, attributes)

    expect(result).toEqual({
      data: [
        { timestamp: String(timestamp), active_users: 1, sign_in_attempts: 0 },
        { timestamp: String(timestamp + 1), active_users: 0, sign_in_attempts: 2 },
      ],
      chartAttributes: attributes,
    })
  })
```

charislam: ahh, makes sense, test makes it super clear! ❤️

apps/studio/components/ui/Charts/ComposedChart.tsx
```diff
@@ -111,11 +111,12 @@ export function ComposedChart({
   const [focusDataIndex, setFocusDataIndex] = useState<number | null>(null)
   const [hoveredLabel, setHoveredLabel] = useState<string | null>(null)
   const [isActiveHoveredChart, setIsActiveHoveredChart] = useState(false)
+  const [hiddenAttributes, setHiddenAttributes] = useState<Set<string>>(new Set())
   const isDarkMode = resolvedTheme?.includes('dark')
 
   useEffect(() => {
     updateStackedChartColors(isDarkMode ?? false)
-  }, [resolvedTheme])
+  }, [resolvedTheme, isDarkMode])
```
charislam:
```suggestion
  }, [isDarkMode])
```

apps/studio/data/reports/v2/auth.config.ts
```diff
@@ -428,50 +463,53 @@ export const createAuthReportConfig = ({
...
+          attribute: 'request_count',
+          label: 'Request Count',
         },
       ]
 
       const sql = AUTH_REPORT_SQL.SignUpLatency(interval, filters)
       const rawData = await fetchLogs(projectRef, sql, startDate, endDate)
       const transformedData = defaultAuthReportFormatter(rawData, attributes)
 
+      console.log('sign up latency', { sql, rawData, transformedData })
```
charislam:
```suggestion
```

:D

charislam (approved):
LGTM on preview

PR #38572, opened 2025-09-10
fix: table editor menu on mobile

charislam:
Problem:

On mobile, when the schema selector is changed in the table editor menu, the menu immediately closes again before the user has a chance to select a table.

https://github.com/user-attachments/assets/4e5f985f-091b-4b44-b253-42e8453cb30b

Solution:

https://github.com/user-attachments/assets/eb44157a-36be-44ef-9fb9-bcf3df43f3a8


apps/studio/components/ui/SchemaSelector.tsx
```diff
@@ -166,6 +166,7 @@ const SchemaSelector = ({
                         key={schema.id}
                         className="cursor-pointer flex items-center justify-between space-x-2 w-full"
                         onSelect={() => {
+                          console.log('calling onSelectSchema', schema.name)
```
joshenlim: sneaky 😛

joshenlim:
Checking on this one - am wondering if this solution might be a bit of an overkill actually, want to avoid a pattern of  handling mobile and desktop behaviours separately as much as possible

am thinking we could update the `MobileSheetNav` component instead as there's currently a `useEffect` that triggers the closes the `MobileSheetNav` whenever `router.asPath` changes ([ref](https://github.com/supabase/supabase/blob/bc3a47317067f4fd80674051e18f08e8538e9c17/packages/ui-patterns/src/MobileSheetNav/MobileSheetNav.tsx#L19))

thinking we could change that to depend on `router.pathname` instead so the menu doesn't close when query params are changing, only if the whole URL changes. If say we want to close the menu nav when the query params change for a specific scenario, we can manually call `setMobileMenuOpen` from `useAppStateSnapshot` then - wdyt? (Esp if I might be overlooking certain scenarios)

charislam:
@joshenlim yeah, I considered that as well, the only thing I was unsure about is that `router.pathname` in Next.js considers `/project/[ref]/editor/[id]` to be the same path across all variations of `ref` and `id`, and I figured that might possibly be a common scenario. For example, if you change it to `router.pathname`, on the Table Editor menu, switching between tables won't automatically close the menu. (I could always manually close it, but it might be hard to find all the cases.)

What do you think? Happy to do it either way.

charislam:
@joshenlim oh I guess the other thing we could do is manually edit router.asPath to take out the query parameters, which would probably affect fewer scenarios, that might be the best way

joshenlim:
Ahhhh great point!

> manually edit router.asPath to take out the query parameters

this sounds good actually - let's give that a shot 🙂 will help stress test to see if there's any niche cases we might be overlooking

charislam:
@joshenlim made the change and tested navigating around the dashboard in mobile mode to make sure the menu closes when it's expected to, didn't find anything that looked obviously wrong

joshenlim (approved):
LGTM verified on staging preview - the main use case to look out for is just where the URL might change query params when clicking something in the mobile sheet nav, which is afaict only table editor and SQL editor

verified that the original issue is fixed, although for SQL editor JFYI there's a very niche use case whereby if you're already on /sql/new and you click on "New snippet", the menu doesnt close since the URL doesn't change (Same for logs explorer). But I think that's fine for now - our dashboard isn't 100% mobile friendly just yet but we'll make small steps towards there 🙂 🙏

PR #38573, opened 2025-09-10
use ref for selected org to ensure not outdated

SaxonF:
We have an issue where depending on rendering selectedOrg can be undefined by time useChat defines transport. Switched to using ref so its always up to date.


joshenlim (approved):
Code changes LGTM - minimally sanity checked the assistant on the staging preview

PR #38574, opened 2025-09-10
Home New: Getting Started

SaxonF:
image

image

Adds GettingStarted row to the home page. This gives us a framework to start building more interactive steps with deeper integration with docs, dashboard, assistant.


apps/studio/components/interfaces/HomeNew/GettingStarted/GettingStartedSection.tsx
```diff
@@ -0,0 +1,563 @@
...
+      },
+    ],
+    [openConnect, selectedFramework]
+  )
+
+  const codeSteps: GettingStartedStep[] = useMemo(
+    () => [
+      {
+        key: 'install-cli',
+        status: 'incomplete',
```
charislam, line 132: Should there be a way to manually mark these steps (that can't be auto-checked) as complete? I personally would find it mildly infuriating to have them sitting around as incomplete perpetually 😄

apps/studio/components/interfaces/HomeNew/GettingStarted/GettingStartedSection.tsx
```diff
@@ -0,0 +1,563 @@
...
+              <CodeBlock className="w-full text-xs p-3 !bg" language="bash">
+                npm install supabase --save-dev
+              </CodeBlock>
+            ),
+          },
+        ],
+      },
+      {
+        key: 'design-db',
+        status: tablesCount > 0 ? 'complete' : 'incomplete',
```
charislam, line 150: This might be sort of confusing given that it's mixing local and hosted flows. Since this is a local CLI Flow, 'design' should be considered complete once the user has a local migration file written (or pushed to their local DB), but this won't be complete until they have a table in their hosted DB.

apps/studio/components/interfaces/HomeNew/GettingStarted/GettingStartedSection.tsx
```diff
@@ -0,0 +1,563 @@
...
+            onClick: () =>
+              openAiChat(
+                'Generate seed data',
+                'Generate SQL INSERT statements for realistic seed data that I can run via the Supabase CLI.'
+              ),
+          },
+        ],
+      },
+      {
+        key: 'add-rls-policies',
```
charislam, line 197: Again, this step might be confusing because of local/hosted mix.

We've been assuming that the user is working locally all along but 'Create a migration file' suddenly gets them to create a policy in hosted dashboard

charislam (approved):
Nice! Pushed a commit to fix a few minor nits + added some questions about the code-based workflow below, but otherwise it looks good, so approving to unblock!

apps/studio/components/interfaces/HomeNew/GettingStarted/GettingStartedSection.tsx
SaxonF: Code first can be broken down further into local vs remote code first. For local, you run local env etc. For remote, you still use remote database but cli for pushing/pulling. Starting with remote flow first, then can maybe break down into local later?

apps/studio/components/interfaces/HomeNew/GettingStarted/GettingStartedSection.tsx
SaxonF: Agree. Will create a follow-up PR adding this functionality

apps/studio/components/interfaces/HomeNew/GettingStarted/GettingStartedSection.tsx
charislam: Ahh fair point, makes sense!

PR #38576, opened 2025-09-10
Refine prompt

SaxonF:
Since switching to GPT-5 I've noticed it takes some instructions quite literally, especially around planning, checklists and inputs/outputs. This refines the prompt to be a little more balanced and less emphasis on planning.


joshenlim (approved):
Minimally sanity tested the assistant on staging preview 🙏

PR #38578, opened 2025-09-10
chore: consolidate payment method components / ownership

kevcodez:
Security approval is required for PCI compliance. This means we need security approval wherever we embed Stripe for payment methods.

Security approval was previously very coarse and most billing changes required security approval unnecessarily.

This PR aims to move all files relevant for payment method display/changes into a single folder and adjusts the ownership to only require security approval for that.


staaldraad (approved):
🤖 🔏

alaister (approved):
LGTM!

PR #38586, opened 2025-09-10
Chore/refactor dashboard history logic

joshenlim:
## Context
Addresses a bug report which users are running into client errors of `Node with id=xxx doesn't exist in the tree` when landing on the SQL Editor.
Briefly what's happening after investigating:

- User's opening the SQL Editor (`/sql`) which automatically loads the last visited snippet (e.g Snippet ID `a`)
  - Snippet `a` belongs to the user based on `owner_id`
  - Snippet `a` has a `folder_id` as well which indicates it lives in a folder
- However, based on the GET /folders request in the network tab, the contents/folders getting returned are for project ID `1` for example
  - Whereas both the snippet and its folder belong in project ID `2`
  - User has access to both Project IDs `1` and `2`, so there's no security issues here
- Because snippet `a` belongs to the user, dashboard loads the snippet's contents correctly
  - The request to fetch the content of a snippet is solely based on the content ID, doesn't consider the project ref
  - It then tries to open the folder of `a` as a UI indication ([ref](https://github.com/supabase/supabase/blob/97c16123b122f9d3ba6e26ee8bcc32fc7c799c45/apps/studio/pages/project/%5Bref%5D/sql/%5Bid%5D.tsx#L51)) which causes the client error as the folder doesn't exist in the project

## Hypothesis
- For context, the dashboard stores the last visited snippet so that each time you land on `/sql`, we'll just load up the last visited snippet for convenience
- Last visited snippet is saved in local storage under a key that includes the project's ref to uniquely save histories across projects
- Dashboard could've incorrectly saved the last visited snippet under the wrong project ref (more details in next section "Problems identified")
  - Lines up with user running into the error by using visiting /sql

## Problems identified:
- The way we're tracking dashboard history is rather inefficient
- We're loading from local storage, and subsequently then loading into the app valtio store
  - Latter is unnecessary, should just use local storage as source of truth
- Each time we were calling `setDashboardHistory` from `app-state`, we always had to pass in a `ref`
  - Potentially very finicky here, my guess as to what happened to the user was that we saved the last visited snippet from a project into another project's dashboard history

## Changes involved
- These should mitigate saving history into the wrong project
  - Created a `useDashboardHistory` hook, updated UI to use this instead
    - Calls `useParam` to ensure that it's always using the project's `ref` that the user is currently on
  - Removed dashboard history related logic from `app-state.tsx`
- These should help unblock the users atm who are running this erro state on the SQL Editor because of malformed history
  - If the loaded SQL snippet doesn't belong to the project, skip loading into the SQL Editor valtio store + redirect back to /new + reset last visited snippet
  - This will then also prevent the `Node with id=xxx doesn't exist in the tree` if the snippet belonged to a folder as no folders will be opened then

## To test
- [ ] Verify that history for SQL Editor still works
  - Open a snippet, go back to project home page, refresh, then open the SQL Editor via the side nav, last visited snippet should open
- [ ] Verify that history for Table Editor still works
  - Same steps as above, just for Table Editor
- [ ] Verify that history for SQL Editor works across projects
  - In one project, open 2 snippets (remember what the last visited snippet was), then go back to your list of projects + refresh
  - Open another project and go to the SQL editor via the side nav, verify that there's no unexpected outcomes
  - Now in that project, open 2 snippets again (remember what the last visited snippet was here), then go back to your list of projects + refresh
  - Now go back to the first project and open the SQL editor via the side nav, verify that the last visited snippet for this project opens up as expected
  - Go back to the second project and open the SQL editor via the side nav, verify that the last visited snippet for this project opens up as expected
- [ ] Verify that history for Table Editor works across projects
  - Same steps as above, just for Table Editor

### DIFF
--- a/apps/studio/hooks/misc/useDashboardHistory.ts
+++ b/apps/studio/hooks/misc/useDashboardHistory.ts
@@ -1,0 +1,30 @@
+import { LOCAL_STORAGE_KEYS, useParams } from 'common'
+import { useLocalStorageQuery } from './useLocalStorage'
+
+type DashboardHistory = { editor?: string; sql?: string }
+const DEFAULT_HISTORY = { editor: undefined, sql: undefined } as DashboardHistory
+
+export const useDashboardHistory = () => {
+  // [Joshen] History should always refer to the project that the user is currently on
+  const { ref } = useParams()
+
+  const [history, setHistory, { isSuccess }] = useLocalStorageQuery(
+    LOCAL_STORAGE_KEYS.DASHBOARD_HISTORY(ref ?? ''),
+    DEFAULT_HISTORY
+  )
+
+  const setLastVisitedTable = (id?: string) => {
+    setHistory({ ...history, editor: id })
+  }
+
+  const setLastVisitedSnippet = (id?: string) => {
+    setHistory({ ...history, sql: id })
+  }
+
+  return {
+    history,
+    setLastVisitedTable,
+    setLastVisitedSnippet,
+    isHistoryLoaded: isSuccess,
+  }
+}
--- a/apps/studio/hooks/misc/useDashboardHistory.ts
+++ b/apps/studio/hooks/misc/useDashboardHistory.ts
@@ -8,7 +8,7 @@ export const useDashboardHistory = () => {
   // [Joshen] History should always refer to the project that the user is currently on
   const { ref } = useParams()
 
-  const [history, setHistory, { isSuccess }] = useLocalStorageQuery(
+  const [history, setHistory, { isSuccess }] = useLocalStorageQuery<DashboardHistory>(
     LOCAL_STORAGE_KEYS.DASHBOARD_HISTORY(ref ?? ''),
     DEFAULT_HISTORY
   )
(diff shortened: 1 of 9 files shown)


apps/studio/hooks/misc/useDashboardHistory.ts
```diff
@@ -0,0 +1,30 @@
...
+import { useLocalStorageQuery } from './useLocalStorage'
+
+type DashboardHistory = { editor?: string; sql?: string }
+const DEFAULT_HISTORY = { editor: undefined, sql: undefined } as DashboardHistory
+
+export const useDashboardHistory = () => {
+  // [Joshen] History should always refer to the project that the user is currently on
+  const { ref } = useParams()
+
+  const [history, setHistory, { isSuccess }] = useLocalStorageQuery(
```
jordienr:
```suggestion
  const [history, setHistory, { isSuccess }] = useLocalStorageQuery<DashboardHistory>(
```

apps/studio/hooks/misc/useDashboardHistory.ts
```diff
@@ -0,0 +1,30 @@
+import { LOCAL_STORAGE_KEYS, useParams } from 'common'
+import { useLocalStorageQuery } from './useLocalStorage'
+
+type DashboardHistory = { editor?: string; sql?: string }
+const DEFAULT_HISTORY = { editor: undefined, sql: undefined } as DashboardHistory
```
jordienr:
```suggestion
const DEFAULT_HISTORY = { editor: undefined, sql: undefined }
```

jordienr: can remove type cast here for better type safety if you pass the type to the useLocalStorageQuery below

jordienr (approved):
nit

PR #38588, opened 2025-09-10
ref(replication): Add alpha banner and remove batch size

iambriccardo:
This PR removes the batch size configuration from the replication UI and adds a new banner for enabling replication which clearly communicates that it is in alpha.

### Screenshots

image

image


imor (approved):

joshenlim (approved):
Tested on staging preview! 🙏

PR #38592, opened 2025-09-10
Feature/braintrust

SaxonF:
Braintrust for evals and token monitoring


apps/studio/package.json
```diff
@@ -73,7 +73,9 @@
     "@zip.js/zip.js": "^2.7.29",
     "ai": "5.0.45",
     "ai-commands": "workspace:*",
+    "autoevals": "^0.0.131",
```
charislam, line 76: this one can go in devDependencies

apps/studio/lib/ai/evals/generate-task.ts
charislam, line 1: I think we can streamline some of this between the evals and the production code. We want everything to remain exactly the same between the two except for mocking tools. Will work on this...

Hallidayo:

PR #38600, opened 2025-09-10
fix reading time for blog posts

raulb:
> [!NOTE]  
> There are some other changes related to running `pnpm run format` 👀 
> The actual fix for this PR is [this commit](https://github.com/supabase/supabase/pull/38600/commits/d679a7e4866c9d40552ed05a572e447fdd2bdb61).

Noticed that our featured blog post does show the correct reading time:

CleanShot 2025-09-10 at 15 54 36@2x

**However**, when visiting any blog post, reading time is always 0

CleanShot 2025-09-10 at 15 55 42@2x

This pull-requests uses the same way to calculate the time fixing the issue (and simplifying code). Here's how it looks after the fix:

CleanShot 2025-09-10 at 15 56 42@2x

Fixes https://linear.app/supabase/issue/FE-1841/fix-broken-reading-time-display-showing-0-min-on-all-blog-posts


fsansalvadore (approved):
That fixes it, thanks! 👍

PR #38601, opened 2025-09-10
fix(studio): syntax error in banner

kemaldotearth:

YES

A fix for height of layout shifting when coming to Billing page from via the Header banner. @jordienr and myself did a little bit of digging and seems related to overflow/scroll behaviour.

Also fixes a `<Link />` component that was nested inside a `<button />` that was causing some suspense warnings/errors.


kemaldotearth:
This was originally for the viewport height bug, but @SaxonF fixed that in #39754 - however, one other outstanding bug myself @jordienr found was a nested `<a>` tag inside a `<button>` tag within the banner. As it's just a link, no need for the button and potential syntax errors.

ivasilov (approved):

PR #38606, opened 2025-09-10
feat(studio): remove query performance tabs

kemaldotearth:
This removes the tabs from Query Performance with unified columns in the table.

YES

This removes the tabs from the table on Query Performance. It also merges the rows together for a unified view. We will replace the tabs with some metrics above in a separate ticket.

**What changed:**
- Unified columns now that there's no tab logic.
- Removed tabs and the metrics attached to them.
- Kept preset configuration (just incase) and introduced a "unified" preset. 
- New query for merged columns/unified view.

| Before | After |
|--------|--------|
| Screenshot 2025-09-10 at 19 45 16 | Screenshot 2025-09-10 at 19 45 08 | 

cc @jordienr


apps/studio/components/interfaces/QueryPerformance/QueryDetail.tsx
```diff
@@ -25,13 +25,13 @@ const SqlMonacoBlock = dynamic(
 )
 
 export const QueryDetail = ({
-  reportType,
+  // reportType,
```
jordienr: Can be removed

jordienr (approved):
Smol nit

PR #38609, opened 2025-09-10
fix: client side exception in cron jobs details

charislam:
The cron jobs details table will throw a client side exception while cron jobs are in running status. This is because the types that we've defined for cron job details expect both return_message and end_time to be strings, but they will be null until the cron job is finished.

Fixed the types and added some defensive checks to fix the bug.

Also made some minor tweaks to the table. Figured that end time and duration can also be shown for failed jobs, doesn't only have to be succeeded ones.

Before:

CleanShot 2025-09-10 at 18 28 48

After:

CleanShot 2025-09-10 at 18 25 48

## To test

Make a cron job that will run for long enough that you can check the running state. `select pg_sleep_for('5 minutes')` will work.


joshenlim (approved):

PR #38613, opened 2025-09-11
Fix filtering issue

SaxonF:
Error in handlePost: TypeError: Cannot read properties of undefined (reading 'filter')
    at t (.next/server/pages/api/ai/sql/generate-v4.js:8:224)
    at async Module.m (.next/server/pages/api/ai/sql/generate-v4.js:8:3511)
    
   This is one of the remaining errors we get from the generate-v4 endpoint. Hard to pinpoint exactly but could potentially be coming from self-hosted installs.


joshenlim (approved):
Code wise LGTM! Minimally sanity checked assistant on staging preview

PR #38624, opened 2025-09-11
Make project infinite query keys fully unique

joshenlim:
## Context

Internal bug report with Compute and Disk section with running into this client side error when trying to change disk configuration
image

Tracing it down, it looks like this is stemming from [here](https://github.com/supabase/supabase/blob/master/apps/studio/data/projects/projects-query.ts#L91) as `setProjectStatus` is being called when updating addons [here](https://github.com/supabase/supabase/blob/5fa0987a80dd86a5eff2ecb6eba1f19bdd6b9b44/apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx#L228)

## Hypothesis 

- I tried to `console.log` what `old` looks like in `setProjectStatus` and it (inconsistently) returned `old` as `{pages: [...], pageParams: { ... }` which made me think that `projectKeys.list()` might be conflicting between `projects-query` and `projects-infinite-query` (which was recently added [here](https://github.com/supabase/supabase/pull/38298))

## Changes involved
- Am opting to make the query key for `projects-infinite-query` fully unique by renaming the first item in its query key array to `projects-all-infinite` so that there's no chance of conflicting with the key from `projects-query`
- (Unrelated) Fix style issue in Storage Type select here
  image
  image
- (Unrelated) Fix small icon positioning issue in Transfer Project modal here
  image
  image

## To test

Just needs to make sure that this flow works without issues:
- Create project in free organization
- Transfer project to paid organization
- Update compute and disk of project
  - Large, IO2, IOPs at 3500

### DIFF
--- a/apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
+++ b/apps/studio/components/interfaces/DiskManagement/DiskManagementForm.tsx
@@ -190,7 +190,6 @@ export function DiskManagementForm() {
   const isPlanUpgradeRequired = org?.plan.id === 'free'
 
   const { formState } = form
-  const { computeSize: formComputeSize, storageType: formStorageType } = form.watch()
   const usedSize = Math.round(((diskUtil?.metrics.fs_used_bytes ?? 0) / GB) * 100) / 100
   const totalSize = formState.defaultValues?.totalSize || 0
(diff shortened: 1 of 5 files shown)


alaister (approved):
LGTM!

PR #38630, opened 2025-09-11
fix: block data branch creation if PITR is disabled

sweatybridge:

Bug fix

## What is the new behavior?

Discussed with @olirice to enforce PITR check before creating data branches.

## Additional context


charislam (approved):
Looks good on preview

PR #38634, opened 2025-09-11
feat(studio): query performance metric cards 

kemaldotearth:

YES

This PR does a couple of small things. First it brings in some glanceable metrics to the top of the Query Performance view and second it brings in visual progress to the "Time Consumed" column. 

**What this changes:**
- Added a query to retrieve some global metrics from `pg_stat_statements`. 
- Added a component block to render those metrics above.
- Added a progress bar to time consumed column/cell, this is a visual touch in the background (highlighted in screenshot).

| Before | After |
|--------|--------|
| Screenshot 2025-09-12 at 10 29 27 | Screenshot 2025-09-12 at 10 29 39 | 

cc @jordienr


apps/docs/next-env.d.ts
jordienr, line 1: can be removed

jordienr (approved):

apps/docs/next-env.d.ts
kemaldotearth: Ah yes, I'm not sure why this keeps getting generated. Perhaps when I switch over to self hosted and I have to run `pnpm i` for the version of Supabase. Though something seems to have changed there. Either way, deleting!

PR #38635, opened 2025-09-11
Use overflow auto instead of scroll

joshenlim:
## Context

Addressing a scrollbar issue on the projects list - we've introduced infinite scrolling [here](https://github.com/supabase/supabase/pull/38298) for the projects list but got an internal report that the scrollbar seems to be persisting showing up. Realised that it might be cause we're using `overflow-y-scroll`, so just changing that to `overflow-y-auto` and removing an unnecessary `overflow-auto`

image


charislam (approved):
LGTM on preview

PR #38638, opened 2025-09-11
fix: estimate number of users when count is large

charislam:
In the Auth Users table, we always fetch an exact count of users. This can be a problem for projects with many (>50K) users as the count(*) might cause performance issues on the database. We already have logic on the Table Editor to only run automatic count estimates (fetching the exact count only if user requests it), this change ports the same logic over to Auth Users.

https://github.com/user-attachments/assets/2737e5fb-f23e-426f-98c7-ec7ddda87cbf

## To test

For testing, the threshold has been set very low (10 users).

1. Create 11 users on your project.
2. Run `vacuum analyze` to make sure planner statistics are up to date.
3. Check that the Auth Users table has an estimated count at the bottom.
4. Delete a user to get back down to 10.
5. Run `vacuum analyze` again.
6. Check that the Auth Users table has an exact count at the bottom.

### DIFF
--- a/apps/studio/data/auth/users-count-query.ts
+++ b/apps/studio/data/auth/users-count-query.ts
@@ -1,23 +1,9 @@
 import { useQuery, type UseQueryOptions } from '@tanstack/react-query'
 
 import { executeSql, type ExecuteSqlError } from 'data/sql/execute-sql-query'
+import { COUNT_ESTIMATE_SQL, THRESHOLD_COUNT } from 'data/table-rows/table-rows-count-query'
 import { authKeys } from './keys'
 import { type Filter } from './users-infinite-query'
-
-export const USERS_THRESHOLD_COUNT = 10 // Set to 10 for testing, set back to 50_000 before merge
-
-const COUNT_ESTIMATE_SQL = /* SQL */ `
-CREATE OR REPLACE FUNCTION pg_temp.count_estimate(
-    query text
-) RETURNS integer LANGUAGE plpgsql AS $$
-DECLARE
-    plan jsonb;
-BEGIN
-    EXECUTE 'EXPLAIN (FORMAT JSON)' || query INTO plan;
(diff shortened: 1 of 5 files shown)


apps/studio/data/auth/users-count-query.ts
```diff
@@ -1,30 +1,49 @@
-import { useQuery, UseQueryOptions } from '@tanstack/react-query'
+import { useQuery, type UseQueryOptions } from '@tanstack/react-query'
 
-import { executeSql, ExecuteSqlError } from 'data/sql/execute-sql-query'
+import { executeSql, type ExecuteSqlError } from 'data/sql/execute-sql-query'
 import { authKeys } from './keys'
-import { Filter } from './users-infinite-query'
+import { type Filter } from './users-infinite-query'
+
+export const USERS_THRESHOLD_COUNT = 10 // Set to 10 for testing, set back to 50_000 before merge
```
charislam: 🔴 🔴 🔴 🔴 🔴 
**Warning to self to fix this before merge**
🔴 🔴 🔴 🔴 🔴

apps/studio/data/auth/users-count-query.ts
```diff
@@ -1,30 +1,49 @@
...
+    query text
+) RETURNS integer LANGUAGE plpgsql AS $$
+DECLARE
+    plan jsonb;
+BEGIN
+    EXECUTE 'EXPLAIN (FORMAT JSON)' || query INTO plan;
+    RETURN plan->0->'Plan'->'Plan Rows';
+END;
+$$;
+`.trim()
```
joshenlim: probably can import this from table-rows-count-query to reduce duplication

thinking likewise for threshold count - probably something we can make consistent for any UI having to do a count on a table

joshenlim (approved):
Verified changes locally! swapped back count to 50,000 for users page

joshenlim:
Will verify once more on staging preview before merging!

PR #38645, opened 2025-09-11
Added a new case study and blog post.

CoolAssPuppy:

YES

This is a new blog post about edge functions and a new case study of a Firebase switcher: Juniver.

Still need a diagram for the blog post, and the logos for the case study aren't working properly.


apps/www/next-env.d.ts
```diff
@@ -1,6 +1,7 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
 /// <reference types="next/navigation-types/compat/navigation" />
+/// <reference path="./.next/types/routes.d.ts" />
```
dnywh: Was this auto-generated @CoolAssPuppy? Wondering if it’s safe to remove

dnywh: Pretty sure it’s autogenerated, so removed!

dnywh (approved):
Builds fine. I will update the Open Graph and diagram imagery in a separate PR

apps/www/next-env.d.ts
stylessh: yes, it's auto-generating for me as well

PR #38546, opened 2025-09-09
fix(studio): usage project selector truncation

kemaldotearth:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

Fixes the project selector truncation for selected value in the selector.

| Before | After |
|--------|--------|
| Screenshot 2025-09-09 at 09 01 06 | Screenshot 2025-09-09 at 09 01 12 |

PR #38649, opened 2025-09-12
fix: let api handle signing key validation error

sweatybridge:

Bug fix

## What is the new behavior?

Currently unsupported signing key properties like `kid` are filtered out, giving users the wrong impression that their import was complete.

It would be clearer to let api report errors on invalid keys instead.

## Additional context


hf (approved):

joshenlim:
Btw noticed a small style bug here that I think was introduced in this [PR](https://github.com/supabase/supabase/pull/38546) in which a max width was applied to SelectTrigger, pushed up a fix but will need design's +1 as well

Before:
image

After:
image

kemaldotearth (approved):
LGTM, the chevron also stays it's full width, so no need for the width applied to the `<span />` 🙌

joshenlim (approved):
LGTM - minimally verified on staging preview with creating a new standby key

PR #38651, opened 2025-09-12
chore: add show testimonial as an enabled feature

alaister:


joshenlim (approved):
Verified changes locally!

PR #38654, opened 2025-09-12
Support for projects pagination (Part 2)

joshenlim:
Continuing from the previous [PR](https://github.com/supabase/supabase/pull/38298), copying the description there for easier reference.

## Context

This will be part of several PRs to introduce pagination for projects in the dashboard, to support organizations with a large number of projects (allow browser to render nicely). It'll involve several changes across different parts of the dashboard so will do this incrementally.

Generally we'll need to support infinite loading for wherever we call the old GET projects endpoint (that's not paginated) via `useProjectsQuery`. In this one specifically, we're looking into parts of the UI where we support selecting projects via dropdowns, by creating a new component that will support infinite scrolling for project selection.

Decided to replace 3 parts of the dashboard with the new component in this PR to better figure out the requirements based on the different use cases and tweak the necessary props that the new component will support - please let me know any feedback on how we can make it better! 🙏 🙂 

## Changes involved

- Added `OrganizationProjectsSelector` component
- Replaced project selection in the following UIs with the new component
  - `ProjectDropdown.tsx` (In the header)
  - `InviteMemberButton` (In org team settings for Team plan and above when inviting a new member)
  - `SupportFormV2`

## To test

- [ ] Verify that switching projects via the top nav header still works (clicking, keyboard)
- [ ] Verify that selecting projects when inviting a member to your org with project-scoped role still works
- [ ] Verify that selecting projects when submitting a support form still works
  - Esp with things like changing organizations, selecting "No specific project"
- [ ] ^ For these, verify that the infinite loading works (I've set the page limit in `projects-infinite-query` to 12 for now for testing)

## Other notes

Was considering making the new component a bit more generalized like an `InfiniteCommand` component, but reckon it'll be easier if we have the component handle the data loading as well, since infinite queries are a bit more messy than normal fetching queries (otherwise gotta pass a lot of props like `hasNextPage`, `fetchNextPage`, etc, etc)

Not necessarily a deal breaker tbh, although we can revisit this if we reckon it's better to do a more generalized component. (Just that afaict we currently only need this behaviour for projects)

### DIFF
--- a/apps/studio/components/ui/OrganizationProjectSelector.tsx
+++ b/apps/studio/components/ui/OrganizationProjectSelector.tsx
@@ -1,10 +1,12 @@
 import { useDebounce, useIntersectionObserver } from '@uidotdev/usehooks'
 import { OrgProject, useOrgProjectsInfiniteQuery } from 'data/projects/projects-infinite-query'
 import { useSelectedOrganizationQuery } from 'hooks/misc/useSelectedOrganization'
-import { Check, HelpCircle } from 'lucide-react'
+import { Check, ChevronsUpDown, HelpCircle } from 'lucide-react'
 import { ReactNode, useEffect, useMemo, useRef, useState } from 'react'
 
 import {
+  Button,
+  cn,
   Command_Shadcn_,
   CommandGroup_Shadcn_,
   CommandInput_Shadcn_,
@@ -21,11 +23,12 @@ import {
 import ShimmeringLoader from 'ui-patterns/ShimmeringLoader'
 
 interface OrganizationProjectSelectorSelectorProps {
+  slug?: string
   open?: boolean
-  actions?: ReactNode
   selectedRef?: string
   searchPlaceholder?: string
   sameWidthAsTrigger?: boolean
+  checkPosition?: 'right' | 'left'
   setOpen?: (value: boolean) => void
   renderRow?: (project: OrgProject) => ReactNode
   renderTrigger?: ({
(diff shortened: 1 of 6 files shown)


apps/studio/data/projects/projects-infinite-query.ts
```diff
@@ -9,7 +9,7 @@ import { projectKeys } from './keys'
 // [Joshen] Try to keep this value a multiple of 6 (common denominator of 2 and 3) to fit the cards view
 // So that the last row will always be a full row of cards while there's a next page
 // API max rows is 100, I'm just choosing 96 here as the highest value thats a multiple of 6
-const DEFAULT_LIMIT = 96
+const DEFAULT_LIMIT = 12
```
joshenlim: 🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨
**_Swap back to 96 before merging, this is only for testing_**
🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨

apps/studio/components/ui/OrganizationProjectSelector.tsx
```diff
@@ -0,0 +1,241 @@
...
+}
+
+export const OrganizationProjectSelector = ({
+  slug: _slug,
+  open: _open,
+  setOpen: _setOpen,
+  selectedRef,
+  searchPlaceholder = 'Find project...',
+  sameWidthAsTrigger = false,
+  checkPosition = 'right',
```
joshenlim, line 53: this is just to make sure that there's no visual changes (invite member modal + support form had the check icons on the left)

but tbh we can just remove this and make all the check positions on the right if we think its unnecessary
only consideration is the width of the dropdown - invite member modal + support form takes up the width of the trigger which is quite long (although i don't think they need to)

apps/studio/data/projects/projects-infinite-query.ts
jordienr:
```suggestion
const DEFAULT_LIMIT = 96
```

jordienr (changes requested):

jordienr (approved):

PR #38658, opened 2025-09-12
fix(replication): Use correct url for tables

iambriccardo:
This PR fixes the link to point to the right URL for the table editor.


joshenlim (approved):
Approving to unblock!

PR #38662, opened 2025-09-12
feat: alternate search index for nimbus

charislam:
Create an alternate search index for Nimbus that filters out feature-flagged pages (equivalent to setting all feature flags to
false).

Notes:
- Creates two new DB tables, `page_nimbus` and `page_section_nimbus`, which are filtered versions of `page` and `page_section`
- Makes `nimbus` versions of all the DB search functions
- Refactored the embedding upload script. Changes to make it faster (got annoyed by how slow it was when testing...), incorporate retries, and produce better summary logs.
- Upload script, when run with the environment variable `ENABLED_FEATURES_OVERRIDE_DISABLE_ALL`, produces and uploads the alternate search index
- Changed all the search calls in frontend/API to check for `isFeatureEnabled('search:fullIndex')` to determine whether to search the full or alternate index
- Adds Nimbus search index production to GitHub Action that generates embeddings

## To test

It'll be easiest to test this on local.

1. Make sure local Supabase is running: `pnpm supabase start`. If this errors you may need to add dummy `GITHUB_CLIENT_ID` and `GITHUB_SECRET` to your `.env`.
3. Run migrations: `pnpm supabase migration up --local`
4. Make sure your `apps/docs/.env.local` has `NEXT_PUBLIC_SUPABASE_URL`, `NEXT_PUBLIC_SUPABASE_ANON_KEY`, and `SUPABASE_SECRET_KEY` pointing to local Supabase
5. Upload both search indexes: `pnpm embeddings:refresh` and `pnpm embeddings:nimbus`
6. Test out the search function on the docs site
7. In `enabled-features.json`, toggle off the `search:fullIndex` feature flag
8. Test the search function again. Try searching for something that's disabled in the Nimbus build, like the Dart SDK


supabase/migrations/20250910155912_create_nimbus_search_tables.sql
```diff
@@ -0,0 +1,325 @@
+-- Create nimbus tables for feature-flag-filtered search
```
charislam, line 1: I didn't change any logic on these tables and functions, they're just copied over from the existing versions except for appending `_nimbus` to the names.

Probably some of these are redundant or obsolete, but let's fix that later, just getting the alt search index working now

ChrisChinchilla:
@charislam Is there a step missing between 6 and 7? I tried and don't seem to see a difference, even rerunning the embeddings and no difference.

charislam:
@ChrisChinchilla hmmm I don't think so, here is the comparison on my local:

with `search:fullIndex: true`

CleanShot 2025-09-15 at 18 50 16

with `search:fullIndex: false`

CleanShot 2025-09-15 at 18 51 19

(the dart references at results 4 & 5 should be gone)

ChrisChinchilla:
@charislam Either I missunderstood something or a simple "doing it again" fixed the issue. All good now.

ChrisChinchilla (approved):

PR #38672, opened 2025-09-13
Fix(docs): prevent SecurityError from rapid history.replaceState call…

### LINKED ISSUE #33156: Docs: Client error on CLI reference page
# Improve documentation

## Link 

https://supabase.com/docs/reference/cli/supabase-vanity-subdomains

## Describe the problem

when you use the scrollbar of the page to scroll the bottom and top of the page really fast you got a client error. 

## Additional context

This happened on Safari on Mac

Devanshusharma2005:
fix(docs): prevent SecurityError from rapid history.replaceState calls during fast scrolling

YES

Bug fix 
Fixes : #33156

Before : 

https://github.com/user-attachments/assets/865db7e6-9b5b-4864-af40-b1745a84c94a

After : 

https://github.com/user-attachments/assets/7fabb8f1-5d8f-45ce-9fda-6d2f4e857b27

When scrolling really fast on Supabase docs pages (especially CLI reference pages like `/docs/reference/cli/supabase-vanity-subdomains`), users see client-side errors like:

- SecurityError: "Attempt to use history.replaceState() more than 100 times per 10 seconds"
- Application error: "a client-side exception has occurred while loading localhost"

This happens because multiple components (ReferenceSectionWrapper, RefSubLayout, etc.) call `history.replaceState()` without throttling during intersection observer events triggered by rapid scrolling.  
It can even crash the browser’s React app due to hitting built-in browser limits.

## What is the new behavior?

- No more SecurityError during fast scrolling
- No app crashes
- Smooth scrolling experience
- URL still updates as expected (just throttled now)
- Table of Contents highlighting keeps working as before

How it works:
- Introduced a `safeHistoryReplaceState` utility with rate limiting (50 calls per 10 seconds)
- Prevent duplicate URL updates when unnecessary
- Added 500ms debounce to intersection observer events
- Replaced all direct `history.replaceState` calls in docs components with the safe version
- Instead of crashing, errors are now logged as warnings

## Additional context

Files changed:
- `apps/docs/lib/historyUtils.ts`: new rate limiter utility
- `apps/docs/features/docs/Reference.ui.client.tsx`: applied safe wrapper + debounce
- `apps/docs/layouts/ref/RefSubLayout.tsx`: safe wrapper
- `apps/docs/layouts/ref/RefSubLayoutNonFunc.tsx`: safe wrapper  
- `apps/docs/components/AuthSmsProviderConfig/AuthSmsProviderConfig.tsx`: safe wrapper

Testing:
Visit any CLI reference page, scroll up and down quickly using the scrollbar.  
Before this fix, the app would throw errors or crash. Now it works smoothly without any issues.

Browser compatibility:
Works across all modern browsers and respects security limits.


apps/docs/features/docs/Reference.ui.client.tsx
```diff
@@ -35,6 +37,11 @@ export function ReferenceSectionWrapper({
   HTMLAttributes<HTMLElement>) {
   const initialScrollHappened = useContext(ReferenceContentInitiallyScrolledContext)
 
+  const debouncedReplaceState = useMemo(
+    () => debounce(safeHistoryReplaceState, 500), 
+    []
+  )
+
```
Devanshusharma2005: Increased debounce to 500ms for more aggressive throttling

apps/docs/lib/historyUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+  let lastUrl = '' 
+
+  return {
+    canCall: (url: string): boolean => {
+      const now = Date.now()
+      
+      while (calls.length > 0 && calls[0] < now - TIME_WINDOW) {
+        calls.shift()
+      }
+
```
Devanshusharma2005: Removing calls older than TIME_WINDOW

apps/docs/lib/historyUtils.ts
Devanshusharma2005, line 1: * Utility for safely calling history.replaceState with rate limiting
 * to prevent SecurityError: Attempt to use history.replaceState() more than 100 times per 10 seconds

Devanshusharma2005:
@tomaspozo would love your review on this.

Devanshusharma2005:
@charislam would love your review here.

charislam (commented):
Thanks for the PR @Devanshusharma2005! Great work

Just one issue came up when I was testing this: if you do a ton of scrolling, then stop, the URL never updates to match the section where you stopped, because the rate limiter drops all updates that exceed the rate limit. But this doesn't really match typical user behavior, because you want the URL to match where you last stopped, and are typically fine with dropping updates in between.

Can we fix this issue just by debouncing? 10 seconds / 100 updates = 100 ms, so a reasonable debounce could probably prevent this error by itself

Devanshusharma2005:
> Thanks for the PR @Devanshusharma2005! Great work
> 
> Just one issue came up when I was testing this: if you do a ton of scrolling, then stop, the URL never updates to match the section where you stopped, because the rate limiter drops all updates that exceed the rate limit. But this doesn't really match typical user behavior, because you want the URL to match where you last stopped, and are typically fine with dropping updates in between.
> 
> Can we fix this issue just by debouncing? 10 seconds / 100 updates = 100 ms, so a reasonable debounce could probably prevent this error by itself

Fixed!. Would love your feedback again.

apps/docs/lib/historyUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (url === lastUrl) return
+
+  try {
+    window.history.replaceState(null, '', url)
+    lastUrl = url
+  } catch (error) {
+    console.warn('Failed to call history.replaceState:', error)
+  }
+}, 120)
+
```
Devanshusharma2005: why 120 ? 
Browser limit : 100 calls per 10 seconds
Math : 10,000 ms + 100 calls = 100ms minimum
120 ms debounce = ~83 calls per 10 seconds which is a safe margin

apps/docs/lib/historyUtils.ts
```diff
@@ -0,0 +1,21 @@
+import { debounce } from 'lodash-es'
+
+let lastUrl = ''
+
+const debouncedHistoryReplaceState = debounce((url: string) => {
+  if (typeof window === 'undefined') return
+
+  if (url === lastUrl) return
+
```
Devanshusharma2005, line 7: Here tracking the last url to avoid unnecessary duplicate calls and also avoiding updating history when URL hasn't actually changed

charislam (approved):
Thanks @Devanshusharma2005! Made some minor changes and LGTM!

PR #38694, opened 2025-09-15
chore: add sso settings to enabled features

alaister:


joshenlim (approved):
Verified locally that flag toggles org SSO page

Also fixed an org scrolling issue in the org pages + nudged the UnknownInterface rendering in SSO

PR #38695, opened 2025-09-15
chore: move connect modal custom content to enabled features

alaister:
A change in requirements means we can simplify this to just fully disable the tabs


joshenlim:
thinking we could just hide the tabs if its only the connection string (pushed up 🙏)
image

joshenlim (approved):
Verified locally that flag works
Verified on preview that things are status quo

PR #38696, opened 2025-09-15
chore: add compute price to enabled features

alaister:
Screenshot 2025-09-15 at 12 14 51


joshenlim (approved):
Verified locally that flag is working

PR #38702, opened 2025-09-15
Add attachment to sentry permission error

jordienr:


joshenlim (approved):
Approving to unblock! 🙏

PR #38704, opened 2025-09-15
chore: Add labeler action and documentation config

ChrisChinchilla:

YES

Adding labeller action to more clearly identify and group PRs, starting with documentation.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -2,4 +2,4 @@
 # https://github.com/marketplace/actions/labeler
 documentation:
 - changed-files:
-  - any-glob-to-any-file: 'apps/docs'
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,5 @@
+# Add 'documentation' to any change in apps/docs
+# https://github.com/marketplace/actions/labeler
+documentation:
+- changed-files:
+  - any-glob-to-any-file: 'apps/docs'
```
raulb: This is to ensure all subdirectories and nested files also work. I believe `**` would be enough based on https://github.com/marketplace/actions/labeler, but just in case.

```suggestion
  - any-glob-to-any-file: 'apps/docs/**/*'
```

raulb (approved):

PR #38706, opened 2025-09-15
add status code to tags instead of context

jordienr:


apps/studio/data/permissions/permissions-query.ts
```diff
@@ -11,11 +11,18 @@ export type PermissionsResponse = Permission[]
 export async function getPermissions(signal?: AbortSignal) {
   const { data, error } = await get('/platform/profile/permissions', { signal })
   if (error) {
+    const statusCode = (error as any)?.code || 'unknown'
```
charislam:
```suggestion
    const statusCode = (!!error && typeof error === 'object' && (error as any).code) || 'unknown'
```

not sure if it's actually possible for an error to be a non-object in this situation but some defensiveness can't hurt

apps/studio/data/permissions/permissions-query.ts
```diff
@@ -11,11 +11,18 @@ export type PermissionsResponse = Permission[]
...
+    // But we still want to capture errors without a status code or 5XXs
+    // since those may require investigation if they spike
+    const sendError = statusCode >= 500 || statusCode === 'unknown'
     handleError(error, {
-      alwaysCapture: true,
+      alwaysCapture: sendError,
       sentryContext: {
         tags: {
           permissionsQuery: true,
+          statusCode: (error as any)?.code || 'unknown',
```
charislam:
```suggestion
          statusCode,
```

charislam (approved):

PR #38712, opened 2025-09-15
Fix missing app frameworks in connect modal

joshenlim:
## Context

App frameworks contents are currently missing - stemming from this [PR](https://github.com/supabase/supabase/pull/38695)

image

The `ConnectTabContentCustom` was previously introduced in this [PR](https://github.com/supabase/supabase/pull/38090) when we were supporting allowing custom content for app frameworks. But since we no longer need that functionality as per [here](https://github.com/supabase/supabase/pull/38695), it's safe to remove that component, and no longer needs to be dynamically rendered

## To test
- [ ] Verify that all app framework contents are rendering correctly in the Connect modal


apps/studio/components/interfaces/Connect/Connect.tsx
```diff
@@ -413,18 +410,11 @@ export const Connect = () => {
...
-                    framework={FRAMEWORKS.find((x) => x.key === selectedParent)}
-                  />
-                ) : (
-                  <ConnectTabContent
-                    projectKeys={projectKeys}
-                    filePath={filePath}
-                    className="rounded-b-none"
-                  />
-                )}
+                <ConnectTabContent
```
saltcod, line 413: do we also need to delete <ConnectTabContentCustom /> ?

joshenlim: oh yes! good point

joshenlim: pushed up!

saltcod (approved):
tested on staging and looks good

PR #38714, opened 2025-09-15
propagate sentry issue id to form

jordienr:
Makes the Crash page add the sentry issue id to the support form URL.

Then the support form will send this sentry issue id with the form to the backend.

This way we can link Front tickets with Sentry issues, making debugging easier.


apps/studio/components/ui/GlobalErrorBoundaryState.tsx
```diff
@@ -31,6 +30,9 @@ export const GlobalErrorBoundaryState = ({ error, resetErrorBoundary }: Fallback
     ? errorMessage.includes("Failed to execute 'removeChild' on 'Node'")
     : false
 
+  // Get Sentry issue ID from error if available
+  const sentryIssueId = (error as any)?.sentryId || (error as any)?.eventId || ''
```
charislam:
```suggestion
  const sentryIssueId = (!!error && typeof error === 'object' && (error as any).sentryId) ?? ''
```

charislam (approved):
Tested locally throwing an error and the Sentry ID gets sent as promised 😄

PR #38715, opened 2025-09-15
chore: turn on typechecking on prod build

charislam:
It's possible for individual PRs to pass typechecking, while the merge result fails it. To prevent errors from getting into production, we can fail the prod build if typecheck fails. Typecheck is still ignored on preview builds for efficiency, since those are blocked on the GitHub typecheck Action.

### DIFF
--- a/apps/docs/next.config.mjs
+++ b/apps/docs/next.config.mjs
@@ -156,9 +156,10 @@ const nextConfig = {
     ]
   },
   typescript: {
-    // WARNING: production builds can successfully complete even there are type errors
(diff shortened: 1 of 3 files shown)


jordienr (approved):

stylessh (approved):

kemaldotearth (approved):
LGTM. Only comment, does our `design-system` project need this too?

charislam:
> LGTM. Only comment, does our design-system project need this too?

design-system has always had type-checking on all builds by default, we did not include it in our dangerous behavior 😅

PR #38723, opened 2025-09-15
fix(ui/auth-sessions): always show refresh token reuse interval

cstockton:
The reuse interval input was previously hidden when rotation was disabled. This caused confusion, since the interval still applies regardless of rotation setting. Allow zero values for reuse interval as it is supported in auth server and is used by some projects.

* Always render SECURITY_REFRESH_TOKEN_REUSE_INTERVAL field.
* Removed .positive() schema check to allow 0 values
* Improves alignment between dashboard UI and backend behavior


hf (approved):

cemalkilic (commented):
Should we explain what happens when you set to `0`? 👀

charislam (approved):
Successfully tested setting reuse interval to 0

PR #38726, opened 2025-09-15
chore: add auth settings redirect

charislam:


cstockton:
@charislam I asked a couple llm's questions that involved getting the URL for auth settings and they replied with this endpoint. I think it's worth adding the redirect, mind approving?

hf (approved):

charislam:
@cstockton I can't approve my own PR 😅 but I'll get another FE approval

apps/studio/next.config.js
```diff
@@ -387,6 +387,11 @@ const nextConfig = {
         destination: '/organizations',
         permanent: false,
       },
+      {
+        source: '/project/:ref/settings/auth',
+        destination: '/project/:ref/auth',
```
joshenlim, line 392: I'm not at my laptop atm sorry but does landing on this page redirect to an auth related page?

iirc we dont have a page thats just /auth

apps/studio/next.config.js
```diff
@@ -387,6 +387,11 @@ const nextConfig = {
...
+        permanent: false,
```
joshenlim, line 393: guessing we probably want this to be true?

apps/studio/next.config.js
charislam: goes to /auth/users, it's not the best since it's not really a settings page, but better than a 404

apps/studio/next.config.js
charislam: `/project/:ref/settings/auth` is a nice url I could see us reusing one day, and I figured we should make it false in case we ever want to resurrect it, but not terribly tied to this idea 🤷🏼‍♀️

joshenlim: All good yeah! No harm tbh i think hahah

apps/studio/next.config.js
joshenlim: reckon we can just directly redirect to /users then?

joshenlim (approved):
Approving to unblock but reckon we can just redirect settings/auth to auth/users
rather than hopping an additional time settings/auth -> auth -> auth/users

cstockton:
Thanks all, merging this one.

PR #38727, opened 2025-09-15
fix: handle undefined count when fetching table rows

charislam:
count can be undefined in the return from fetching table rows (it uses executeSqlQuery under the hood which has no type safety) so we need to deal with that case


joshenlim (approved):
Verified changes on staging preview - both estimate, and loading count on demand

PR #38733, opened 2025-09-16
Chore/state of startups

MildTomato:
- updated header


apps/www/pages/state-of-startups.tsx
```diff
@@ -206,10 +207,51 @@ function StateOfStartupsPage() {
...
+                  </Link>
+                </li>
+              ))}
+            </ol>
+          </div>
+
+          <SurveySectionBreak />
+        </section>
+
+        {/* Original Header - Commented Out
```
dnywh: Happy for me to remove this in a follow-up PR?

apps/www/components/SurveyResults/StateOfStartupsHeader.tsx
```diff
@@ -0,0 +1,137 @@
...
+export function StateOfStartupsHeader({
+  title,
+  subtitle,
+  chapters,
+  showProgressBars = true,
+}: StateOfStartupsHeaderProps) {
+  return (
+    <header className="mt-32">
+      <div className="max-w-[60rem] mx-auto">
+        <div className="flex flex-col gap-1">
```
dnywh: I think all three DOM elements could just be the `header`. Seems to be the same, visually:

```suggestion
        <header className="mt-32 max-w-[60rem] mx-auto flex flex-col gap-1">
```

dnywh (approved):
Just some minor suggestions. Looks good! Thanks for jumping on.

apps/www/pages/state-of-startups.tsx
MildTomato: oops

apps/www/components/SurveyResults/StateOfStartupsHeader.tsx
MildTomato: yea i think also we should maybe have a `<h1>`
 hidden

PR #38736, opened 2025-09-16
Contextual error for foreign tables in table editor

joshenlim:
## Context

There's potential confusion for wrappers and branching - as branching does not copy table data over to branches.

As such for wrappers which store their keys in the vault schema, foreign tables will not work while on a branch and this might be a source of confusion. Opting to contextualize the error message a bit better as such:

## Before

image

## After

If on main project
image

If on a branch
image

### DIFF
--- a/apps/studio/components/grid/components/grid/Grid.tsx
+++ b/apps/studio/components/grid/components/grid/Grid.tsx
@@ -61,7 +61,6 @@ export const Grid = memo(
       const { data: org } = useSelectedOrganizationQuery()
       const { data: project } = useSelectedProjectQuery()
       const isBranch = project?.parent_project_ref !== undefined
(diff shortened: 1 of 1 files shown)


apps/studio/components/grid/components/grid/Grid.tsx
```diff
@@ -58,6 +60,8 @@ export const Grid = memo(
 
       const { data: org } = useSelectedOrganizationQuery()
       const { data: project } = useSelectedProjectQuery()
+      const isBranch = project?.parent_project_ref !== undefined
+      console.log({ isBranch })
```
charislam:
```suggestion
```

:D

charislam (approved):
Tested on preview and looks good

PR #38738, opened 2025-09-16
fix query performance links

jordienr:
- fixes broken links to query performance


joshenlim (approved):
Minimally verified staging preview for dashboard - CMD K link to query performance, and the redirect in next.config

PR #38740, opened 2025-09-16
docs: update google auth docs

hf:
Updated Google docs based on recent changes to Sign in with Google.


apps/docs/content/guides/auth/social-login/auth-google.mdx
```diff
@@ -2,148 +2,124 @@
...
-1. Go to the [API Credentials page](https://console.cloud.google.com/apis/credentials).
-1. Click `Create credentials` and choose `OAuth Client ID`.
-1. For application type, choose `Web application`.
-1. Under **Authorized JavaScript origins**, add your site URL.
-1. Under **Authorized redirect URLs**, enter the callback URL from the [Supabase dashboard](/dashboard/project/_/auth/providers). Expand the Google Auth Provider section to display it.
+Supabase Auth needs a few scopes granting access to profile data of your end users, which you have to configure in the [**Data Access (Scopes)**](https://console.cloud.google.com/auth/scopes) screen:
 
-   <Admonition type="tip">
+- `openid` (add manually)
+- `.../auth/userinfo.email` (added by default)
```
fadymak, line 29: Do you think we should add the complete scope here since it seems that they all use `https://www.googleapis.com` as the namespace?

fadymak (approved):

apps/docs/content/guides/auth/social-login/auth-google.mdx
hf: I think this is how you register them in the dialog, and that's how they appear:

image

PR #38747, opened 2025-09-16
chore: limit regions for nimbus

alaister:


joshenlim (approved):
Verified locally that flag is working
Verified staging preview that things are status quo

PR #38752, opened 2025-09-16
Use Turbopack for next build

joshenlim:
## Context

Dashboard staging previews have been suffering from OOM issues and relatively long build times ever since the nextjs bump. So we're hoping to use `turbopack` to improve that, which based on observation is able to shave build times from 10 - 15 minutes ([ref](https://vercel.com/supabase/studio-staging/CfqL4FRAF87NCC8SRVyEvF9m7rQa)) currently down to 4 - 7 minutes ([ref](https://vercel.com/supabase/studio-staging/H3Kr4tqL7n8mSDrjzBrJg6k6vJnN))

> [!WARNING]  
> Important to smoke test everything to make sure nothing breaks
> Before merging, ensure that we're **ready to instant rollback** in case anything happens

## Observations

Just a heads up that these are the changes i've observed while testing things locally

### Studio
- There's going to be a lot of `useLayoutEffect` warnings in the server side logs - I don't think this is stemming from any dashboard code though so if I had to guess it might be one of the npm modules. It's just annoying, should be harmless
  image

### Docs
- None that I could find

### WWW
- None that I could find

### Studio
- The main problem was on `generate-v4` endpoint, the staging preview build kept showing the following error RE react-query
  image
  - Realised that `generate-v4` was importing a number of files from our react query hooks (`data/...`)
  - To be fair, we shouldn't mix client side and server side logic, so the main changes were ensuring that `generate-v4` wasn't importing anything that had react query in it
- Affected files are as follows
  - In general, am shifting the function that fetches data into `rq-hook-folder/fetchers.ts` - not sure if this is the best naming convention though so please do let me know if it can be better
  - Like wise for `hooks/misc/aiOptInLevel.ts` - not sure if its best to leave the stuff here (details below in the table)
  - A lot of the file changes are just hence updating import statements

    | File  | Changes |
    | `execute-sql-query.ts`  | Shifted `executeSql` into a separate file under `execute-sql.ts`. Also updated all direct imports of `executeSql()` to import from `execute-sql.ts` instead  |
    | `database-functions-query.ts`  | `fallback-tool.ts` was using `getDatabaseFunctions`, so shifted that out into 1`database-functions/fetchers.ts`  |
    | `database-policies-query.ts`  | `fallback-tool.ts` and `schema-tools.ts` were using `getDatabasePolicies`, so shifted that out into `database-policies/fetchers.ts`  |
    | `entity-definitions-query.ts` | `fallback-tool.ts` was using `getEntityDefinitionsSql`, so shifted that out into `database/fetchers.ts`  |
    | `getOrgAiDetails` | This function innately calls `getOrganizations` and `getProjects` so had to shift those out as well into `organization/fetchers.ts` and `projects/fetchers.ts` |
    | `getAiOptInLevel` | This function was living in `useOrgOptedIntoAi` which imported several react hooks, so shifted out into `hooks/misc/aiOptInLevel.ts` |

### Docs
- None that I could find

### WWW
- None that I could find

## To test

- [ ] Full smoke test of the dashboard
- [ ] Full smoke test of the docs site
- [ ] Full smoke test of the www site


joshenlim:
Just picking this up from Kevin, currently known issues

## Studio
- Bug: CSP issue for font assets
  image
- Warning: There's a blast of console warnings when running studio locally for `useLayoutEffect`?
  image


## Docs
- Bug: 404 link for "Local Development" here
  - This makes me a bit nervous, only caught this through randomly skimming and checking pages
  - On prod it links to `/docs/guides/local-development/cli/getting-started`
  - On preview it links to `/docs/guides/cli/getting-started`
  image


## www
Looks alright - just skimmed the previews across multiple pages, all looking good

kevcodez:
> CSP issue for font assets

Should be unrelated to this PR - this is because we don't have the specific vercel preview URL in Cloudflare configured

joshenlim:
@kevcodez hmm i was wondering about that but the preview builds for other PRs all load the fonts properly though

one observation (not sure if related)
on this preview the fonts are getting loaded as such with these paths
image


but on staging, the fonts are getting loaded with these paths (as tho they're obfuscated)
image

kevcodez:
> all load the fonts properly though

Other previews do not use the asset cdn - I have explicitly activated the asset CDN for that branch to ensure that there are no issues with the asset upload process

You can disable it by removing the FORCE_ASSET_CDN env var for that branch - might be good at this point to ensure nothing is broken just because of CF CORS settings

PR #38753, opened 2025-09-16
Add self-remediation for orphan prefixes and refactor delete objects to be a singular request

joshenlim:
## Context

There seems to be an issue that's happening in the Dashboard's Storage Explorer, whereby if you try to delete a folder through the UI, there's a chance that the folder that you deleted persists in the UI despite it being empty. And trying to delete that folder again subsequently just shows success toasts but the folder continues to persist

image
image

This happens rather inconsistently and can be reproduced occasionally with just a flat folder (e.g I reproduced this with a folder that contains 70 files, no nested folders)

## Hypothesis

Current assumption is that there might be some race condition happening on the Storage BE side of things (unconfirmed - but given it's inconsistent nature) as the dashboard fires a series of network requests when deleting folders - specifically we're firing a DELETE object request per 2 files. So if there's 70 files, that'll mean 35 requests. I'm actually not sure why we're doing that to be honest 🤦‍♂️

However, this seems to be a known problem from the Storage team ("Orphan prefix") which they've suggested a simple fix would just be to clean that row from the `storage.prefixes` table.

## Changes involved

- Added self-remediate step in `deleteFolder` where if the number of files to be deleted is 0, we'll trigger a function to delete from the `storage.prefixes` table
- Opt to trigger just one singular request when deleting files since DELETE objects accepts an array of strings for `paths`
  - This hopefully aims to mitigate the orphan prefix issue which I haven't been able to reproduce since, but I cannot confirm if it resolves it entirely
  - NOTE: The UI consequence of this change is that there's no longer a progress indicator since it just waits on when the request finishes (that might have been a factor of consideration for the old implementation, but reckon this is still better than almost DDOSing the project's API)

## To test
- [ ] Try to reproduce the Orphan prefix issue by uploading a folder with some files, then deleting the folder
- [ ] If you have the Orphan prefix issue, try deleting the empty folder - it should clear up nicely
- [ ] Verify that you can also still delete multiple files (not just the whole folder)

### DIFF
--- a/apps/studio/data/storage/bucket-prefix-delete-mutation.ts
+++ b/apps/studio/data/storage/bucket-prefix-delete-mutation.ts
@@ -1,0 +1,35 @@
+import { executeSql } from 'data/sql/execute-sql-query'
+
+/**
+ * [Joshen] JFYI this is solely being used in storage-explorer.tsx and hence doesn't have a useMutation hook
+ * It's solely supposed to aid users to self-remediate a known issue with orphan prefixes so the condition to
+ * clean the prefix from the storage.prefixes is intentionally strict here (e.g only level 1, since we've only
+ * seen this bug happening at the top level folder)
+ */
+
+type DeleteBucketPrefixParams = {
+  projectRef?: string
+  connectionString?: string
+  bucketId?: string
+  prefix?: string
+}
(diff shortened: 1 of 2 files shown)


jordienr (approved):
cannot repro anymore so i think its fixed!

PR #38761, opened 2025-09-16
fix: failing type errors in build

charislam:
Typecheck within a Vercel build is stricter than typecheck without building due to Next generated types, therefore we had a bunch of accumulated type errors that needed fixing:

**Docs**
- `route.ts` files can only export the API handlers, not utility functions

**Studio**
- pages/api can only contain API handlers, not utility files
- apiWrapper was not correctly inferring types

**www**
- params in App Router pages is now a promise and needs to be awaited

Tested by running `NEXT_PUBLIC_VERCEL_ENV=production pnpm build` locally


apps/docs/app/api/revalidate/route.utils.ts
```diff
@@ -0,0 +1,103 @@
+import { createClient } from '@supabase/supabase-js'
```
charislam, line 1: Cut and pasted from `revalidate/route.ts`

fsansalvadore (approved):
Tested the local "prod" build 👍

joshenlim (approved):
Minimally did a skim of staging preview for dashboard

PR #38770, opened 2025-09-17
feat: improved issue select

dnywh:

Minor improvements to our support form:

- Removed default value for the _What areas are you having problems with?_ dropdown 
- Replaced it with a placeholder and error states (because it’s a required field)
- Improved language for clarity

We assume the user wants to file a ticket specifically about _Issues with APls / client libraries_:

CleanShot 2025-09-17 at 11 47 27@2x

## What is the new behavior?

We don't assume (and we make language a bit cleaner):

CleanShot 2025-09-17 at 12 13 22@2x

## Additional context

- [Linear](https://linear.app/supabase/issue/DEPR-64/remove-preset-value-for-problem-area)


CarelFdeWaal (approved):
Looks like request to API won't change so I am happy

charislam (approved):

PR #38773, opened 2025-09-17
feat: select ‘25 banner

dnywh:

A dismissible banner on `www` for [Supabase Select](https://select.supabase.com/).

No banner.

## What is the new behavior?

The banner, using existing logic set up for Launch Weeks. That includes a key and value in local storage so if the user dismisses the banner, that choice will be remembered for that browser.

## Additional context

This removes, replaces, or adds to some ‘placeholder’ banner props used for Launch Week. That’s so it can be used for non Launch Week things like this. See @fsansalvadore’s https://github.com/supabase/supabase/pull/37718/files for prior LW-focussed approach (which was hidden from view).


SaxonF (approved):
Approved to unblock and assumption we will keep iterating

MildTomato (approved):
approve. although not seen visually

PR #38775, opened 2025-09-17
FE-1856: fix shared api report filters

jordienr:
## to test
- use shared api report filter
- they filter


joshenlim (approved):
Checked reports on staging preview, but can only verify database reports as that's the only one that i've got data for 🙏

PR #38777, opened 2025-09-17
FE-1860: Shared API Report Fixes

jordienr:
- fixes open in log explorer button
- removes old button, opt for top right button instead
- fixes shared api report not refetching on filterby change
- adds new "sql" prop for components to grab the sql ran to pass on to log explorer link

## To test
- create some users or grab a project with usage
- go to reports / auth
- click open in log explorer in one of the metrics
- SQL makes sense / runs


charislam (approved):
Tested as described

PR #38779, opened 2025-09-17
feat: support verify_jwt per function in self-hosted docker

w0rmh0l3:

YES

feature

The self-hosted docker does not support per function jwt verification.

## What is the new behavior?

This allows for configuring per function verify_jwt to true/false in volumes/functions/config.toml.

## Additional context

This PR does not take away the current VERIFY_JWT, instead uses that as fallback(default) for when a function does not exist in config.toml.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -317,7 +317,8 @@ services:
     image: supabase/edge-runtime:v1.69.6
     restart: unless-stopped
     volumes:
-      - ./volumes/functions:/home/deno/functions:Z
+      - ./volumes/functions:${FUNCTIONS_VOLUME_PATH}:Z
+      - ./volumes/functions/config.toml:${FUNCTIONS_VOLUME_PATH}/config.toml:Z
     depends_on:
       analytics:
         condition: service_healthy
@@ -329,11 +330,12 @@ services:
       SUPABASE_DB_URL: postgresql://postgres:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
       # TODO: Allow configuring VERIFY_JWT per function. This PR might help: https://github.com/supabase/cli/pull/786
       VERIFY_JWT: "${FUNCTIONS_VERIFY_JWT}"
+      FUNCTIONS_VOLUME_PATH: ${FUNCTIONS_VOLUME_PATH}
     command:
       [
         "start",
         "--main-service",
-        "/home/deno/functions/main"
+        "${FUNCTIONS_VOLUME_PATH}/main"
       ]
 
   analytics:
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -109,7 +109,10 @@ OPENAI_API_KEY=
 # Functions - Configuration for Functions
 ############
 # NOTE: VERIFY_JWT applies to all functions. Per-function VERIFY_JWT is not supported yet.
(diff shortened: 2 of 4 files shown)


w0rmh0l3:
@laktek as suggested by @sweatybridge [here](https://github.com/supabase/cli/pull/786#issuecomment-3300748395), tagging you for the review.

w0rmh0l3:
@aantti - do you reckon this feature is still desired/in roadmap? While I proposed this to be merged, we have been using this in our prod env via `docker-compose.override.yml`. And it would help us to know if this is undesired feature, so that we are not introducing breaking changes in our env in long term.

aantti:
I'll leave it to @kallebysantos to review this one if I may. There's currently a couple of projects related to both edge functions, and CLI (that already supports this option) - but I have to see where what approach settles in the end. Ideally, it would be great to have platform, CLI, and self-hosted aligned with regard to this kind of behavior.

kallebysantos (approved):
LGTM! Just recommend to also add it to [documentation](https://supabase.com/docs/guides/self-hosting/self-hosted-functions)

Btw, using the `config.toml` also gives possibility for custom `entrypoint, import_map` or any other relevant option.

docker/docker-compose.yml
```diff
@@ -329,11 +330,12 @@ services:
       SUPABASE_DB_URL: postgresql://postgres:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
       # TODO: Allow configuring VERIFY_JWT per function. This PR might help: https://github.com/supabase/cli/pull/786
```
thunze, line 331:
```suggestion
```

IIUC, this PR addresses this TODO, doesn't it?

aantti: Frankly, that PR for CLI isn't really helpful. [Self-hosted Supabase](https://supabase.com/docs/guides/self-hosting) configuration is not the same stack of containers as the one started via [Local development & CLI](https://supabase.com/docs/guides/local-development). This particular PR seem to be adding a new code to the main function code so that a custom `config.toml` can get parsed - but `config.toml` doesn't exist in self-hosted Supabase configuration, and adding it just for edge runtime is probably not the most elegant way of solving it.

Aside from that, I think there's currently some WIP regarding how edge functions support the new API keys, and I'd like it to settle a bit. I'm not sure how it would ultimately look like with the preferred way of using the new `sb_` opaque key for API auth verification in functions.

I'll maybe let @kallebysantos to add more.

kallebysantos: Hey mates, 

> that PR for CLI isn't really helpful.

I don't see any problem with that, 

>  This particular PR seem to be adding a new code to the main function code so that a custom config.toml can get parsed...

I think its the only way to support it on self-hosted, `main/index.ts` should be responsible to locate the function as well use any extra metadata for customisation. Keeping it as `config.toml` just enforces a similar pattern of CLI, and I don't see any problem with that. 
Who is self-hosting/maintaining is free to update `main/index.ts` file to grab this metadata from any other place, like scoped `config.toml/json` per function or external feature-flag/metadata provider service.

--

I did already approved this PR in the past, and my opinion remais the same.
But is for @aantti to find what should be the best place to default add this extra props.

> Aside from that, I think there's currently some WIP regarding how edge functions support the new API keys,

I don't think that API keys will block this `verify_jwt` per function feature. On my point the keys will should be supplied via Docker env.

PR #38781, opened 2025-09-17
Improve logic to calculate time range when log chart is clicked

jordienr:
- removes export default
- fixes all imports of logspreviewer
- improves calculation of time range for logs bar chart click
- moves calculation to external function to be used elsewhere
- adds tests to this function


charislam (approved):
Tested clicking the charts in preview, LGTM

PR #38783, opened 2025-09-17
feat(studio): add query perf symlink to reports

kemaldotearth:

YES

Introduces Query Performance as a tab in the Reports submenu. Until we can migrate it over or replace with Query Insights, no harm in improving discoverability of it.

Like so:
Screenshot 2025-09-17 at 12 23 39


apps/studio/components/layouts/ReportsLayout/ReportsMenu.tsx
```diff
@@ -146,6 +146,12 @@ const ReportsMenu = () => {
               },
             ]
           : []),
+        {
+          name: 'Query Performance',
+          key: 'query-performance',
+          url: `/project/${ref}/advisors/query-performance${preservedQueryParams}`,
```
jordienr: think we can remove the preserverdQueryParams here since the query performance page doesn't use them for filtering

jordienr (approved):

PR #38786, opened 2025-09-17
feat(replication): Show warning when publication is not found

iambriccardo:
This PR shows a warning when the publication name of this pipeline is not there anymore.

### Screenshots

image


joshenlim (approved):
Verified changes on preview!

PR #38789, opened 2025-09-17
Address various console warnings

joshenlim:
## Context

Addresses various console warnings that show up in the developer tools when landing on the dashboard

- Many many configcat warnings (shows up on prod too)
    image
  - Addressed by passing in a dummy user when calling `getAllValuesAsync` from configcat

- `motion is deprecated` (shows up on local dev)
    image
  - As per the message, just swapping `motion()` to `motion.create()` in a couple of files

- `Failed to load chart sync settings from localStorage` (shows up on local dev and build logs)
    image
  - This one's just because `useChartHoverState` is trying to access local storage before the browser is loaded, just added a check for `window`

- `Warning: Missing Description for DialogContent` (shows up on local dev)
    image
  - Just shows up when opening the create branch modal, just needed to add `aria-describedBy={undefined}` in `DialogContent` (and we should do this whenever we're not rendering a `<DialogDescription>`)

## To test

I'm mainly worried about the feature flags stuff, so just need to test if the flags are still working
A quick way to do so is to open the create branch modal to verify if the "include data" toggle is visible (that UI is behind a feature flag that's set to `true`)

### DIFF
--- a/apps/studio/components/interfaces/APIKeys/APIKeyRow.tsx
+++ b/apps/studio/components/interfaces/APIKeys/APIKeyRow.tsx
@@ -19,7 +19,7 @@ export const APIKeyRow = ({
   apiKey: Extract<APIKeysData[number], { type: 'secret' | 'publishable' }>
   lastSeen?: { timestamp: string }
 }) => {
-  const MotionTableRow = motion(TableRow)
+  const MotionTableRow = motion.create(TableRow)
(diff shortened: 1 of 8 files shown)


charislam (approved):
Checked 'include data' flag and clicked around a bunch on the API keys page + tried the assistant in project layout

PR #38797, opened 2025-09-17
feat: local mcp server

gregnr:
Adds a local HTTP-based MCP server to `studio` at `/api/mcp`. This endpoint is only available on local/self-hosted instances.

To keep this PR small, we only implement 3 operations _(more in future PRs)_:
- `executeSql`
- `listMigrations`
- `applyMigration`

## Quick background
Our [MCP server](https://github.com/supabase-community/supabase-mcp) exists as a standalone NPM package at `@supabase/mcp-server-supabase`. It exposes APIs that allow you to implement separate backends, called `platform`. This approach allows us to re-use the majority of MCP logic between platforms (i.e. between management API and local/self-hosted), but still allow each platform to implement their own specific differences (like how `executeSql` is implemented).

## Refactors
The above operations already live at the following Next.js API endpoints:
- `/platform/pg-meta/[ref]/query` (`executeSql`)
- `/v1/projects/[ref]/database/migrations` (`listMigrations` & `applyMigration`)

So to share the logic between these endpoints and MCP, we move the common logic to `./lib/api/self-hosted`:
- `./lib/api/self-hosted/query.ts`
- `./lib/api/self-hosted/migrations.ts`

## Testing
1. Checkout this branch `feat/local-mcp` and run local studio stack:
    ```shell
    pnpm dev:studio-local
    ```
    > If you receive errors about `GITHUB_CLIENT_ID` not being available, just comment out the `[auth.external.github]` section in `./supabase/config.toml` (we should make this not required)

2. Add to your preferred MCP client using the URL `http://localhost:8082/api/mcp`.

    - If using Cursor you can use this button:

        [Install MCP Server](https://cursor.com/en/install-mcp?name=local-supabase&config=eyJ1cmwiOiJodHRwOi8vbG9jYWxob3N0OjgwODIvYXBpL21jcCJ9)

    - If using Claude Code you can use this command:

        ```shell
        claude mcp add --transport http local-supabase http://localhost:8082/api/mcp
        ```

3. Test the tools:
    - "What tables do I have?"
    - "What migrations have I run?"
    - "Create a todos table"

Resolves AI-150


mattrossman:
Works for me in Cursor 👍

apps/studio/lib/api/local/query.ts
joshenlim, line 1: Wondering if this file might be doing the same thing as this one?

https://github.com/supabase/supabase/blob/master/apps%2Fstudio%2Flib%2Fself-hosted.ts

gregnr: Ah yup looks like it. I'll consolidate. Ok to organize under `lib/api/local/*`? Can also rename to `lib/self-hosted/*` if that's a better term

apps/studio/pages/api/mcp/index.test.ts
charislam, line 1: This needs to be moved out of the `api` folder. Only actual API routes can go in this folder, otherwise typecheck will fail 😢

charislam (commented):
Tested with Claude Code and it worked fine for me. Just two type errors:

- One is the test file which needs to be moved
- The other is this: `studio:typecheck: lib/ai/supabase-mcp.ts:1:10 - error TS2305: Module '"@supabase/mcp-server-supabase"' has no exported member 'createSupabaseApiPlatform'.` which is all kinds of bizarre given that it _works_

apps/studio/pages/api/mcp/index.test.ts
gregnr: Thank you! [Moved](https://github.com/supabase/supabase/pull/38797/commits/6671bbf8bf6e1ff26e1037338f3c47d52b961d26) to `./tests/pages/api/mcp/index.test.ts`. Let me know if there is a better place

gregnr:
@charislam I forgot that we moved `createSupabaseApiPlatform` to its own package export. Fixed [here](https://github.com/supabase/supabase/pull/38797/commits/cc823d0dedfc29d8bd3e29035a28ca75363aa13b).

_(still very strange that it worked)_

charislam (approved):
Tested again and looks good!

apps/studio/lib/api/local/query.ts
gregnr: Consolidated both execute sql functions and renamed directory to `lib/api/self-hosted/*`

apps/studio/data/sql/execute-sql-query.ts
```diff
@@ -44,10 +44,10 @@ export async function executeSql<T = any>(
   >,
   signal?: AbortSignal,
   headersInit?: HeadersInit,
-  fetcherOverride?: (
-    sql: string,
+  fetcherOverride?: (options: {
+    query: string
     headers?: HeadersInit
-  ) => Promise<{ data: T } | { error: ResponseError }>
+  }) => Promise<{ data: T } | { error: ResponseError }>
```
gregnr, line 50: The new `lib/api/self-hosted/*` functions follow this object notation, so just makes it easier to pass in. Happy to revert though.

madhavpujara-at:
Umm, the tests are failing can anyone check why are they failing
image
this is the error image
 the script is running out of memory for tests
 I am waiting for this feature

PR #38802, opened 2025-09-18
Refine top section home new

SaxonF:
image

Adjusts home top section styling


joshenlim (approved):
Verified changes on staging!

JFYI i swapped the amber colors to gray instead to match the dropdown
Thinking it might be better to reserve amber color too for other callouts - but feel free to swap back though! 🙏

PR #38803, opened 2025-09-18
feat: select ‘25 banner styled

dnywh:

Adds some visual styles to the Supabase Select banner introduced yesterday in https://github.com/supabase/supabase/pull/38773

https://github.com/supabase/supabase/pull/38773 brought in a plain banner; white text on black background irrespective of user theme preference:

Database  Open source SQL Database

## What is the new behavior?

We still force dark mode but I’ve made the banner more closely aligned with the [Supabase Select](https://select.supabase.com/) branding:

https://github.com/user-attachments/assets/1d2bce1b-720d-4e3a-aa0f-24271edd30ec

## Additional context

Since this banner is quite different to the template used in Launch Weeks, I’ve reverted the JSON and prop changes I introduced in https://github.com/supabase/supabase/pull/38773. That means future Launch Weeks can just roll as per usual. So ignore my comments about this yesterday @fsansalvadore! All is back to normal. Except the _Announcement_ component which forces a dark `<X size={16} />`.


fsansalvadore (approved):
nice!

PR #38804, opened 2025-09-18
update onboarding

SaxonF:
image

This is the start of a series of branches that update the style and behaviour of the Assistant.

This PR updates onboarding state to include performance and security lints to give lints and Assistant abilities more exposure.

PR #38805, opened 2025-09-18
Chore/ai flow model

SaxonF:
This is the second pull request in a series of improvements to the Assistant.

This requires https://github.com/supabase/supabase/pull/38804 to be merged first.

This PR:
- Updates ai and ai-sdk/open packages which contain a fix for sending reasoningEffort to OpenAI. This means we can switch back to responses with minimal reasoning effort. The effect is quicker responses.
- Fixes an issue with incompatible tool states


charislam (approved):

apps/studio/pages/api/ai/sql/generate-v4.ts
```diff
@@ -117,6 +97,28 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse) {
     }
   }
 
+  // Only returns last 7 messages
+  // Filters out tools with invalid states
+  // Filters out tool outputs based on opt-in level using renderingToolOutputParser
+  const messages = (rawMessages || []).slice(-7).map((msg: any) => {
```
charislam, line 103: I don't like that any but since it's the old behavior let's just ignore that for now 😄 I'll make another PR to fix typing for this

PR #38806, opened 2025-09-18
Assistant action orientated approach

SaxonF:
image

This is the last PR in a series to update the styling and behaviour of the Assistant. 

Requires https://github.com/supabase/supabase/pull/38805

The main change behind this PR is that we removing the display query and edge function tools in favour of client-side execute_sql and deploy_edge_function tools. This , along with prompt changes, gives the assistant greater capability of running tools and reading results (as long as permissions are right) whilst giving users visibility over what they are running.

At the same time we're making some style refinements to messages and tools.

How
- Introduces two new client-side tools
- Filters out these tools via MCP so that client-side take precedence
- Refactors client-side to simplify how queries are run. This means results are only stored in tool results now when confirmed or skipped. Running directly from QueryBlock does not store results which is a compromise.
- Does the same for edge functions
- in generate-v4 we filter out tool results from these tools if opt in permission levels are not sufficient

To test:
- `pnpm i`
- Ask Assistant to read data in one of your tables and then try skipping the query and running the query
- Ask Assistant to create a new table, then skip, then try run the query and notice confirmation
- Ask Assistant to write an example query and notice no confirm to run (just displays via markdown)
- Ask Assistant to write an edge function and then confirm deploy


apps/studio/components/interfaces/Reports/ReportBlock/ReportBlock.tsx
```diff
@@ -57,89 +59,107 @@ export const ReportBlock = ({
         if (failureCount >= 2) return false
         return true
       },
+      onSuccess: (contentData) => {
+        if (!isSnippet) return
```
charislam, line 64: Do we always want the snippet to be run as soon as it's fetched? It creates a sort of odd situation in custom reports, where as soon as you open the report, writes are run (and then fail), and you get the generic error message rather than seeing what the saved snippet is

Image

apps/studio/components/ui/AIAssistantPanel/elements/Tool.tsx
```diff
@@ -0,0 +1,55 @@
+/* no react imports to avoid type resolution issues here */
```
charislam: What problems do you run into with type resolution? I added a type import from react and it passed typecheck, but is there is a separate issue?

SaxonF: we can remove

apps/studio/components/interfaces/Reports/ReportBlock/ReportBlock.tsx
SaxonF: Lets work on this in a separate PR because it also affects new home page.

joshenlim: i think running it is fine since FE doesn't validate client side if the snippet is read only or not (we'd have to let the database tell us)

however fwiw this was the original behaviour for queries that have write behaviour - wondering if we can keep it that way? reckon write queries wouldn't make sense to render within ReportBlock for custom reports or the new home page too hence why thinking of keeping this original UI

I'll see how we might be able to keep this UI with minimal changes

image

apps/studio/components/interfaces/Reports/ReportBlock/ReportBlockContainer.tsx
```diff
@@ -39,35 +41,28 @@ export const ReportBlockContainer = ({
...
-              className="!text-xs font-medium text-foreground-light flex-1 truncate"
+            <div
+              className={cn(
+                'flex items-center gap-2 flex-1 transition-opacity',
+                showDragHandle && 'group-hover:opacity-25'
+              )}
             >
-              {label}
-            </h3>
+              <h3 className="heading-meta truncate">{label}</h3>
```
joshenlim, line 64: i'd suggest we don't uppercase the label here since its based on a user input (snippet title) as it could be confusing to the user

this is a bad example but for e.g if the user has 2 snippets named
- "Users table"
- "users table"

there was a similar issue for this in the auth policies page, where we were uppercasing the table names

image

apps/studio/components/ui/AIAssistantPanel/EdgeFunctionRenderer.tsx
```diff
@@ -0,0 +1,159 @@
...
+        showReplaceWarning={showReplaceWarning}
+        onCancelReplace={() => setShowReplaceWarning(false)}
+        onConfirmReplace={() => void performDeploy()}
+        onDeploy={handleDeploy}
+        hideDeployButton={showConfirmFooter}
+      />
+      {showConfirmFooter && (
+        <div className="mx-4">
+          <ConfirmFooter
+            message="Assistant wants to deploy this EdgeFunction"
```
joshenlim: nit spacing "Edge Function"

apps/studio/components/ui/AIAssistantPanel/EdgeFunctionRenderer.tsx
```diff
@@ -0,0 +1,159 @@
...
+            cancelLabel="Skip"
+            confirmLabel={isDeploying ? 'Deploying…' : 'Deploy'}
```
joshenlim: Just to verify - is this something that we want to use from here on? Not strong preference, just considering consistency (but also the former is not that intuitive to type)
`…` vs `...`

apps/studio/components/ui/AIAssistantPanel/Message.Parts.tsx
```diff
@@ -0,0 +1,327 @@
...
+    return (
+      <div className="my-4 rounded-lg border bg-surface-75 heading-meta h-9 px-3 text-foreground-light flex items-center gap-2">
+        <Loader2 className="w-4 h-4 animate-spin" />
+        Writing Edge Function...
+      </div>
+    )
+  }
+
+  if (state === 'output-error') {
+    return <div className="text-xs text-danger">Failed to deploy Edge Function.</div>
```
joshenlim: nit - could use `p` given that it's purely text

apps/studio/components/ui/AIAssistantPanel/Message.hooks.ts
joshenlim, line 1: nit but i reckon this can live within `lib/profile.tsx` - `UserDropdown.tsx` will benefit from this too

apps/studio/components/ui/AIAssistantPanel/Message.hooks.ts
```diff
@@ -0,0 +1,18 @@
...
+  avatarUrl: string | undefined
+} {
+  const { profile } = useProfile()
+  const username = profile?.username
+
+  const { data: identitiesData } = useProfileIdentitiesQuery()
+  const githubProfileData = identitiesData?.identities?.find(
+    (x) => x.provider === 'github'
+  )?.identity_data
+  const avatarUrl = githubProfileData?.avatar_url
```
joshenlim: Hmm should use the `getGitHubProfileImgUrl` i reckon as there's some minor optimizations there when fetching the avatar image

apps/studio/components/ui/AIAssistantPanel/Message.utils.ts
joshenlim, line 1: can't leave a comment on L18 cause its unchanged but `findResultForManualId` doesn't seem to be used anymore? will clean it up

apps/studio/components/ui/SqlWarningAdmonition.tsx
```diff
@@ -7,32 +7,47 @@ export interface SqlWarningAdmonitionProps {
...
       type="warning"
       className={`mb-0 rounded-none border-0 shrink-0 bg-background-100 ${className}`}
     >
       <p className="text-xs !mb-1">
-        {warningType === 'hasWriteOperation'
-          ? 'This query contains write operations.'
-          : 'This query involves running a function.'}{' '}
-        Are you sure you want to execute it?
+        {message ?? (
+          <>
```
joshenlim, line 37: could do a string interpolation instead of fragment here

joshenlim: the message conditional could also be wrapping the `p` element? otherwise it'll be an empty `p` if there's no message

joshenlim (approved):
LGTM! Verified changes on staging preview

PR #38809, opened 2025-09-18
Address odd bug between data grid contextmenu and overlay

joshenlim:
## Context

We received an internal bug report that in the Auth Users UI, if you were to open the delete user confirmation modal through right clicking on the user - once the modal is closed (no matter cancelling or confirming the delete), nothing else in the dashboard is clickable.

## Investigation
- This is very very oddly specifically happening when Data Grid + Context Menu + Dialog (or any component that uses Shadcn Overlay, which includes Sheet) components are being used together (Using a Context Menu from Shadcn in a Data Grid to trigger a Dialog)
- A `pointer-events: none` style is getting applied to the `body` element of the page when the modal gets closed, but that style isn't getting cleaned up - hence why the dashboard becomes unresponsive after
- I can reproduce this not just in the Auth Users UI, but also in the Cron Jobs Table UI which has a similar set up of this 3 things
- I cannot reproduce this in the SQL Editor UI (Side nav snippet list) as it doesn't use Data Grid (Only context menu and dialog)

## Changes involved
- Am opting to simplify and scoping down the fix by removing the `pointer-events` style on the body if it exists when closing the confirmation modal for just the Auth Users DeleteUserModal + CreateCronJobSheet + DeleteCronJob, since its specifically only happening in these 3 areas
- Table Editor is not affected because it's not using the ContextMenu from Shadcn

## To test
- [ ] Verify in Auth Users UI, right click user -> Delete user, close modal, can still click around in the dashboard
- [ ] Verify in Cron UI, right click job -> Edit / Dlete job, close sheet / modal, can still click around in the dashboard

### DIFF
--- a/apps/studio/components/interfaces/Auth/Users/DeleteUserModal.tsx
+++ b/apps/studio/components/interfaces/Auth/Users/DeleteUserModal.tsx
@@ -3,7 +3,6 @@ import { toast } from 'sonner'
 
 import { useUserDeleteMutation } from 'data/auth/user-delete-mutation'
 import { User } from 'data/auth/users-infinite-query'
-import { timeout } from 'lib/helpers'
 import ConfirmationModal from 'ui-patterns/Dialogs/ConfirmationModal'
 
 interface DeleteUserModalProps {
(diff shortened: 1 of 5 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #38812, opened 2025-09-18
fix: add overflow to resizable panels

dnywh:

Fixes an issue where ResizablePanel contents weren’t scrollable, meaning critical information was hidden and inaccessible on (at least) the following pages:
- Security Advisor
- Performance Advisor

This PR also allows the LinterPageFooter to be dismissed, just like its sibling in QueryPerformance.

- Users can’t scroll down to critical information, as [reported](https://supabase.slack.com/archives/C0429V78ACX/p1758029655881809?thread_ts=1758027312.503529&cid=C0429V78ACX) by @inian
- The footer doesn’t provide very useful information, and isn’t dismissable
  - That exacerbates the above issue

## What is the new behavior?

- The ResizablePanel is scrollable when it has overflowing contents
- The LinterPageFooter is dismissable

https://github.com/user-attachments/assets/8a7d70a0-d319-434e-a5c1-0c353d8b4515


apps/studio/components/interfaces/Linter/LinterDataGrid.tsx
```diff
@@ -166,7 +166,12 @@ const LinterDataGrid = ({
       {selectedLint !== null && (
         <>
           <ResizableHandle withHandle />
-          <ResizablePanel defaultSize={30} maxSize={45} minSize={30} className="bg-studio border-t">
+          <ResizablePanel
+            defaultSize={30}
+            maxSize={45}
+            minSize={30}
+            className="bg-studio border-t flex flex-col h-full"
```
dnywh, line 173: Needs to be a flex container for the below change to take effect

apps/studio/components/interfaces/Linter/LinterDataGrid.tsx
```diff
@@ -176,7 +181,7 @@ const LinterDataGrid = ({
               </div>
               <Button type="text" icon={<X />} onClick={handleSidepanelClose} />
             </div>
-            <div className="p-6">
+            <div className="p-6 flex-grow min-h-0 overflow-y-auto">
```
dnywh, line 184: This fixes the issue where content wasn’t scrollable. It’s analogous to QueryPerformanceGrid

apps/studio/components/interfaces/Linter/LinterPageFooter.tsx
```diff
@@ -14,8 +17,25 @@ const LinterPageFooter = ({
   refetch,
   hideDbInspectCTA,
 }: LinterPageFooterProps) => {
+  const [showBottomSection, setShowBottomSection] = useLocalStorageQuery(
+    LOCAL_STORAGE_KEYS.LINTER_SHOW_FOOTER,
```
dnywh, line 21: This local storage key is shared for both Performance and Security Advisor. I figured that’s fine since they are so similar in nature; if a user dismisses one, they’re likely to have the other dismissed too.

apps/studio/components/interfaces/Linter/LinterPageFooter.tsx
```diff
@@ -14,8 +17,25 @@ const LinterPageFooter = ({
...
-    <div className="px-6 py-6 flex gap-x-4 border-t ">
+    <div className="px-6 py-6 flex gap-x-4 border-t relative">
+      <Button
+        className="absolute top-1.5 right-3 px-1.5"
+        type="text"
+        size="tiny"
+        onClick={() => setShowBottomSection(false)}
+      >
+        <X size="14" />
+      </Button>
```
dnywh, line 38: This adds the close button. Similar to what QueryPerformance does.

apps/studio/components/interfaces/Linter/LinterPageFooter.tsx
dnywh: See [below](https://github.com/supabase/supabase/pull/38812/files#r2357976056) for the close button where this state logic is used

packages/common/constants/local-storage.ts
```diff
@@ -39,6 +39,7 @@ export const LOCAL_STORAGE_KEYS = {
   CLS_DIFF_WARNING: 'cls-diff-warning-dismissed',
   CLS_SELECT_STAR_WARNING: 'cls-select-star-warning-dismissed',
   QUERY_PERF_SHOW_BOTTOM_SECTION: 'supabase-query-perf-show-bottom-section',
+  LINTER_SHOW_FOOTER: 'supabase-linter-show-footer',
```
dnywh, line 42: As mentioned above: this is shared across two pages, since they are so similar in nature.

kemaldotearth (approved):
LGTM

PR #38823, opened 2025-09-18
FE-1858 FE-1853: fix pooler max connections

jordienr:
## BEFORE
- its showing 0 max connections

CleanShot 2025-09-18 at 16 47 36@2x

## AFTER
- shows correct max connections

CleanShot 2025-09-18 at 16 46 52@2x


charislam (commented):
which charts is this meant to show on? I don't get tooltips for most of the charts on the database reports page, am I looking in the wrong place? 😅 

https://github.com/user-attachments/assets/b70e17a6-18d6-4113-a4e7-7cec7254323f

jordienr:
> which charts is this meant to show on? I don't get tooltips for most of the charts on the database reports page, am I looking in the wrong place? 😅
> 
>  CleanShot.2025-09-18.at.16.30.04.mp4

You'll need to check with an org in Team or Enterprise for this one!

charislam (approved):
Looks good on preview!

PR #38838, opened 2025-09-19
update coming soon copy replication

SaxonF:
image

Updates copy of the coming soon state for database/replication to reduce confusion between the already available read replicas and upcoming ETL focus / external destinations.


dnywh (approved):
Lookin’ good!

PR #38840, opened 2025-09-19
feat: blog post postgrest 13

steve-chavez:
Highlights the main changes in https://github.com/PostgREST/postgrest/releases/tag/v13.0.0 

preview: https://zone-www-dot-com-git-blog-pgrst-v13-supabase.vercel.app/blog/postgrest-13-release


dshukertjr (approved):

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+## Automatic tsvector convertion
+
+Previously you could only use the full text search operator on `tsvector` columns, now you can do it on `text` columns too:
+
+```js
+const { data, error } = await supabase.from('titles').textSearch('primary_name', `'god' & 'father'`)
+```
+
+This works because `text` columns will be automatically converted with `to_tsvector`.
```
kiwicopple: this is very cool steve

i imagine it will be quite slow since there is not pre-build index? should we still recommend they create a generated column/index if they are using this frequently?

kiwicopple: (and if not necessary, it would be cool to explain why)

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
```diff
@@ -0,0 +1,141 @@
...
+## Content-Length header
+
+For observability, you can now verify the response body size in bytes in the `Content-Length` header.
+
+```bash
+
+HTTP/1.1 200 OK
+Content-Length: 104
+Content-Location: /items
+```
```
kiwicopple, line 162:
```suggestion
```bash
HTTP/1.1 200 OK
Content-Length: 104
Content-Location: /items
```
```

File: apps/www/_blog/2025-09-18-postgrest-13-release.mdx [hunk 3]
```diff
@@ -0,0 +1,141 @@
...
+## Proxy-Status header
+
+The PostgREST error code is now present in the `Proxy-Status` header.
+
+```bash
+HTTP/1.1 406 Not Acceptable
+Proxy-Status: PostgREST; error=PGRST116
+```
+
+You can check this header and the Content-Length one the Supabase Logs Explorer.
```
kiwicopple:
```suggestion
You can check the `Proxy-Status` and `Content-Length` headers in the Supabase Logs Explorer.
```

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
```diff
@@ -0,0 +1,141 @@
...
+  .from('people')
+  .update({ primary_name: 'Marlon Brando Jr.' })
+  .eq('primary_name', 'Marlon Brando')
+  .limit(1)
+  .order('nconst')
+```
+
+But this never gave an error, it only did the equivalent of a `UPDATE .. LIMIT` SQL statement. It also required adding a PK in the `order` method, which was a confusing interface.
+
+To do this now, you should use the new Max Affected instead.
```
kiwicopple:
```suggestion
To do this now, you should use the new `Max Affected` instead.
```

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
```diff
@@ -0,0 +1,141 @@
...
+```bash
+HTTP/1.1 406 Not Acceptable
+Proxy-Status: PostgREST; error=PGRST116
+```
+
+You can check this header and the Content-Length one the Supabase Logs Explorer.
+
+## Breaking Changes
+
+### Limited Update/Delete
```
kiwicopple: was anything actually removed/breaking here? Or is it more of a suggestion to move to `max affected`? If so, it could be worth mentioning that so that it's very clear to users what action they should take

I would recommend putting the equivelent code using `maxAffected` so that they can compare. It could also be good to add tabs which have the `cURL` request?

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
```diff
@@ -0,0 +1,141 @@
...
+  .eq('primary_name', 'Marlon Brando')
+  .limit(1)
+  .order('nconst')
+```
+
+But this never gave an error, it only did the equivalent of a `UPDATE .. LIMIT` SQL statement. It also required adding a PK in the `order` method, which was a confusing interface.
+
+To do this now, you should use the new Max Affected instead.
+
+### JWT kid validation
```
kiwicopple:
```suggestion
### JWT `kid` validation
```

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+PostgREST now validates the JWT `kid` claim. If your JWT contains a `kid`, it will try to match the `kid` of one of the keys in the configured JSON Web Key Set. Check the [official docs](https://docs.postgrest.org/en/v13/references/auth.html#jwk-kid-validation) for more details.
```
kiwicopple:
```suggestion
PostgREST now validates the JWT `kid` claim. If your JWT contains a Key ID (`kid`), it will try to match this with one of the `kid`'s in the configured JSON Web Key Set. Check the [official docs](https://docs.postgrest.org/en/v13/references/auth.html#jwk-kid-validation) for more details.
```

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+If you use Supabase Auth or the CLI to create JSON Web Keys, you shouldn’t worry about this change. For users that integrate with other Auth systems, check that both your JWT and JWKS follow the above rules.
+
+### Schema validation in PostgREST search path
+
+The schemas inside `db-schemas` and `db-extra-search-path` are now validated. This means you cannot put a nonexistent schema there, if you do PostgREST will faill with an error message.
```
kiwicopple:
```suggestion
The schemas inside `db-schemas` and `db-extra-search-path` are now validated. This means you cannot put a nonexistent schema there, if you do PostgREST will fail with an error message.
```

questions:

- what happens if people upgrade and the schema doesnt exist? will the upgrade fail?
- what happens if someone drops a schema after it's been added to PostgREST?

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
steve-chavez: Thanks for catching that :pray: . Added a note mentioning an index.

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
steve-chavez: > what happens if people upgrade and the schema doesnt exist? will the upgrade fail?

It will fail, the user would have to remove the schema first.

> what happens if someone drops a schema after it's been added to PostgREST?

It would fail too, the user would need to ensure the schema drop is done with transactional DDL while keeping PostgREST in sync, like so:

```sql
begin;
drop schema old_schema;
alter role authenticator set pgrst.db_schemas = 'public, pg_graphql, others'; -- make sure old_schema is not present here
commit;
```

I've added this on the blog post for now. But I do see a problem with this when users drop schemas with the GUI, so I'll see if PostgREST can return to being more lenient or we could also remove the schema automatically with an event trigger.

apps/www/_blog/2025-09-18-postgrest-13-release.mdx
steve-chavez: > was anything actually removed/breaking here? Or is it more of a suggestion to move to max affected?

Right there isn't something actually breaking here (no error given), so I've removed this section. I think it's enough to have `maxAffected` as a feature, and that's already included above.

> It could also be good to add tabs which have the cURL request?

This is now done for all snippets. Also included swift snippets.

apps/www/_blog/2025-09-30-postgrest-13-release.mdx
```diff
@@ -0,0 +1,204 @@
...
+const { data, error } = await supabase.from('titles').select(`
+  title:primary_title,
+  ...people(actors:primary_name),
+  ...genres(genres:name)
+`)
+```
+
+```bash name=Bash
+curl 'https://<your-domain>/titles'
+  -d "select=title:primary_title,...people(actors:primary_name),...genres(genres:name)"
```
laurenceisla: Adding `-d` turns it into a `POST` request, it needs `--get` to make it work as a query-string.

```suggestion
curl --get 'https://<your-domain>/titles'
  -d "select=title:primary_title,...people(actors:primary_name),...genres(genres:name)"
```

apps/www/_blog/2025-09-30-postgrest-13-release.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+<$CodeTabs>
+
+```js name=JavaScript
+const { data, error } = await supabase.from('titles').textSearch('primary_name', `'god' & 'father'`)
+```
+
+```bash name=Bash
+curl 'https://<your-domain>/titles'
+  -d "primary_name=fts.'god'%26'father"
```
laurenceisla: Ditto

```suggestion
curl --get 'https://<your-domain>/titles'
  -d "primary_name=fts.'god'%26'father"
```

Issue #38839, opened 2025-09-19
Hiding Authorization headers on DB webhooks is undocumented


adboio:
# Improve documentation

## Link 

- Primary docs: https://supabase.com/docs/guides/database/webhooks#creating-a-webhook
- Also (maybe more importantly) inline docs for the webhook edit UI

## Describe the problem

Supabase UI hides `Authorization` headers from DB webhooks. This behavior is undocumented.

## Describe the improvement

- Add documentation inline for the webhook create/edit UI that `Authorization`-named headers may be hidden.

## Additional context

Code ref: https://github.com/supabase/supabase/blob/031c5644e7d6d3068206e7ce4ab6905f1634af20/apps/studio/components/interfaces/Database/Hooks/FormContents.tsx#L104

adboio:
I think docs should be added here: https://github.com/supabase/supabase/blob/031c5644e7d6d3068206e7ce4ab6905f1634af20/apps/studio/components/interfaces/Database/Hooks/HTTPRequestFields.tsx#L82

adboio:
Looks like this may be a deeper issue - not only are the auth headers hidden, they are deleted, unbeknownst to the client. Opened a new bug issue here: https://github.com/supabase/supabase/issues/38848

Closing in favor of the bug issue...

PR #38844, opened 2025-09-19
docs: update webhook authorization handling in Studio UI

Devanshusharma2005:

YES

Docs update (both inline UI docs and primary documentation)

When users create/edit database webhooks in Studio and add Authorization headers, those headers disappear from the UI after saving. This confuses users who think their headers are lost, but they're actually just hidden by the UI for security reasons. The behavior isn't documented anywhere, leading to confusion and bug reports.
Related to issue #38839 where contributors reported user confusion about "missing" Authorization headers.

## What is the new behavior?
Added inline documentation in the Studio webhook UI explaining that Authorization headers are automatically managed
Added a clear note in the Database Webhooks documentation guide explaining when/why Authorization headers are hidden
Both docs explain the three scenarios: auto-added for Edge Functions with JWT verification, removed otherwise

Screenshot 2025-09-19 at 10 32 10 AM

## Additional context
The Authorization header filtering happens in FormContents.tsx where headers are automatically added/removed based on Edge Function JWT settings. This is actually helpful behavior but was completely undocumented, causing unnecessary confusion.

### DIFF
--- a/apps/docs/content/guides/database/webhooks.mdx
+++ b/apps/docs/content/guides/database/webhooks.mdx
@@ -57,7 +57,7 @@ When creating or editing webhooks through the Supabase Studio UI, headers named 
 - For Edge Functions with JWT verification enabled: Authorization headers are automatically added with the appropriate service role token
 - For regular HTTP requests or Edge Functions without JWT verification: Authorization headers are automatically removed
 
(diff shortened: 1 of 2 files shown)


Devanshusharma2005:
@charislam again up for review. Lmk if there is any improvements needed.  ^^

Devanshusharma2005:
@joshenlim would love your review too.

joshenlim:
hey there @Devanshusharma2005! thanks so much for this one - just checking real quick, this one seems more like a bug to be honest and should be addressed in the dashboard rather than providing docs for this

I'll take a look at the bug instead but will close this PR for now 🙂 🙏

Devanshusharma2005:
Thank you @joshenlim will also take a look at dashboard.

joshenlim:
Got a fix here! 🙂 🙏 
https://github.com/supabase/supabase/pull/38894

PR #38846, opened 2025-09-19
Chore/attempt to address ts build issue

joshenlim:
## Context

We've recently enabled type checking as part of our production builds on Vercel [in this PR](https://github.com/supabase/supabase/pull/38715) in order to catch TS errors from merges. However we've since observed build errors of "Type instantiation is excessively deep and possibly infinite." happening intermittently which unless we're attentive to our production builds in Vercel, it'll easily get overlooked and fixes wouldn't have made it through to production until the next successful build.

## Hypothesis

This one's quite hard to track as no matter how many times I try to run `npx tsc --noEmit` I just can't seem to reproduce that error? There's no specific line or file that shows up from the build logs too which makes it hard to pinpoint. But a number of articles mentioned `zod` as a potential candidate contributing to this problem (more specifically having multiple versions of `zod` installed).

We use `zod` in several different packages that the dashboard depends on (e.g ui-patterns, packages) - they technically all have version `^3.25.76` defined in their respective `package.json` files, but am opting to standardize the version through pnpm's catalog.

Separately, am going to set `noErrorTruncation` as `true` for dashboard's `tsconfig.json` so that if this happens again we'll be able to get more details from the error

There's also articles talking about `z.infer` being the possible culprit, or even react query's nested query keys too but reckon we only investigate those as we get more details, otherwise kinda flying blind

## Changes involved

- Standardize version of `zod` package to `^3.25.76` via pnpm catalog in `pnpm-workspace.yaml`
- Set `noErrorTruncation` to `true` in dashboard's `tsconfig.json` -> `compilerOptions`

### DIFF
--- a/apps/design-system/package.json
+++ b/apps/design-system/package.json
@@ -45,7 +45,7 @@
     "ui": "workspace:*",
     "ui-patterns": "workspace:*",
     "unist-util-visit": "^5.0.0",
-    "zod": "^3.25.76"
+    "zod": "catalog:"
   },
   "devDependencies": {
     "@shikijs/compat": "^1.1.7",
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -120,7 +120,7 @@
     "uuid": "^9.0.1",
     "valtio": "catalog:",
     "yaml": "^2.4.5",
-    "zod": "^3.25.76"
+    "zod": "catalog:"
   },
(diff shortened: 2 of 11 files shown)


alaister (approved):
LGTM!

- Tested on preview

dnywh (approved):
Working as expected locally

PR #38851, opened 2025-09-19
Improve member invite UI

joshenlim:
Fixes FE-1795
Fixes FE-1840

## Context

Changes here addresses some confusion with 2 states in the organization member invite flow (obscure errors from the users' POV)
- Seeing an invite when not logged in yet
  image
- Seeing an invite that's already been accepted
  image

## Changes involved

Generally did a skim of the whole flow and made the following changes:
- Added better loading states to reduce layout shift
- Only load invite token query if user is logged in (based on `profile`)
- Tweaked UI when user is not logged in yet (Don't show the unnecessary error)
  not-signed-in
- Tweaked UI when invite has already been accepted / declined
  - Invites get deleted in the BE when they get accepted or declined so FE doesn't have context on which organization the invite was for. Opting for a clearer copy + CTA to head back to dashboard
  Screenshot 2025-09-19 at 14 24 36

## To test
- [ ] Smoke test inviting an organization member, and accepting that invite

### DIFF
--- a/apps/studio/components/interfaces/OrganizationInvite/OrganizationInvite.tsx
+++ b/apps/studio/components/interfaces/OrganizationInvite/OrganizationInvite.tsx
@@ -8,13 +8,14 @@ import { useOrganizationAcceptInvitationMutation } from 'data/organization-membe
 import { useOrganizationInvitationTokenQuery } from 'data/organization-members/organization-invitation-token-query'
 import { useProfile } from 'lib/profile'
 import { ResponseError } from 'types'
-import { Button, Loading, cn } from 'ui'
+import { Button, cn } from 'ui'
+import { Admonition, GenericSkeletonLoader } from 'ui-patterns'
 import { OrganizationInviteError } from './OrganizationInviteError'
 
 export const OrganizationInvite = () => {
   const router = useRouter()
   const { profile } = useProfile()
-  const { slug, token, name } = useParams()
+  const { slug, token } = useParams()
 
   const { data, error, isSuccess, isError, isLoading } = useOrganizationInvitationTokenQuery(
     { slug, token },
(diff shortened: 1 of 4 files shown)


apps/studio/components/interfaces/OrganizationInvite/OrganizationInvite.tsx
```diff
@@ -8,26 +8,29 @@ import { useOrganizationAcceptInvitationMutation } from 'data/organization-membe
...
       retry: false,
       refetchOnWindowFocus: false,
+      enabled: !!profile,
     }
   )
   const hasError =
     isError || (isSuccess && (data.token_does_not_exist || data.expired_token || !data.email_match))
-  const isNotSignedInError = error?.message.includes('No authorization token was found')
+  const inviteHasBeenAccepted =
+    error?.code === 401 && error?.message.includes('Failed to retrieve organization')
```
matlin: Probably more of an API fix but `Error('Failed to retrieve organization') = invitedAccepted` is confusing. Maybe the variable could be better named `inviteIsNoLongerValid`?

apps/studio/components/interfaces/OrganizationInvite/OrganizationInvite.tsx
```diff
@@ -47,57 +50,88 @@ export const OrganizationInvite = () => {
...
             </p>
-            <div className="flex justify-center gap-3">
-              <Button asChild type="default">
-                <Link href={loginRedirectLink}>Sign in</Link>
-              </Button>
-              <Button asChild type="default">
-                <Link href={loginRedirectLink}>Create an account</Link>
-              </Button>
-            </div>
+            {isSuccess && slug && (
```
matlin: Just because this is a in a big ternary, would be helpful to have a more specific variable name here. Like `succesfullyFetchedInvitation` or something like that

apps/studio/pages/join.tsx
```diff
@@ -1,39 +1,48 @@
...
-          className="block h-[24px] cursor-pointer rounded"
-        />
-      </Link>
-      <div
-        className={cn(
-          'mx-auto overflow-hidden rounded-md border',
-          'border-muted bg-alternative text-center shadow',
-          'md:w-[400px]'
+        {mounted && (
+          <img src={imgUrl} alt="Supabase" className="block h-[24px] cursor-pointer rounded" />
```
matlin, line 28: Is this to prevent flickering between light and dark mode?

matlin (approved):
Looks a lot better! Left some optional nits but LGTM

apps/studio/components/interfaces/OrganizationInvite/OrganizationInvite.tsx
joshenlim: oooooh yes good point! since this is the case for either accepted or deleted

apps/studio/pages/join.tsx
joshenlim: im somehow running into some hydration + rendering errors in this one 😕 whereby without mounted, despite the imgUrl being updated once the page has rendered, the rendered HTML somehow doesn't update in the DOM

PR #38852, opened 2025-09-19
Fix missing parenthesis in policy editor

joshenlim:
## Context
One of our users pointed out a missing closing parenthesis in the policy editor UI (purely UI, not functionality) if there's both a using and with check block (for commands "UPDATE" and "ALL") (Notice line 8 in the screenshot)
image

## Changes involved
Adds the missing parenthesis

image

## To test
- [ ] Swap through the commands in the policy editor - just make sure that the UI code boilerplate generally looks right


alaister (approved):
LGTM!

- Tested on preview

PR #38854, opened 2025-09-19
Update global error boundary to add context for insertBefore error

joshenlim:
Fixes FE-1870

## Context

We've received quite a number of tickets regarding client crashes due to `Failed to execute 'insertBefore' on 'Node': ...` and `Failed to execute 'removeChild' on 'Node'` - these are both known to be caused by Chrome's translate feature in the browser as mentioned [here](https://github.com/facebook/react/issues/24865) and [here](https://github.com/facebook/react/issues/17256) respectively.

While we give contextual information regarding the 'removeChild on Node' error, realised we were missing the same for the 'insertBefore on Node' error so this PR adds that in

Note that this is not a full resolution as we can't fix this entirely in the dashboard really - it's dependent on how Chrome plays with the client code when translating. The theoretical fix is that we need to wrap every single text in the dashboard with a `<span>` but that's really too much of a change on our end in my opinion to manage this bug that's caused by Chrome's translate

Also opting to reduce prominence of the support button in the case of these 2 specific errors
image


alaister (approved):
LGTM!

PR #38858, opened 2025-09-19
chart footer tooltip improvs

jordienr:
- adds a bit of delay to the tooltip group
- fixes tooltip trigger only wrapping text, not whole clickable area
- makes the color on disabled attributes gray.

---

- tooltip only triggers if you hover the text

- before (idk why the previews are not working)

https://github.com/user-attachments/assets/9836943d-21fa-4712-aca3-8dd96109c239

- after

https://github.com/user-attachments/assets/dd4ef25a-f4fe-470e-8143-524f0e811bc4

- disabled attribute still shows colored dot
CleanShot 2025-09-19 at 11 53 19@2x
- after
CleanShot 2025-09-19 at 12 00 08@2x

- adds a bit more offset to the tooltip
- be4
CleanShot 2025-09-19 at 11 54 40@2x
- after
CleanShot 2025-09-19 at 12 01 39@2x

### DIFF
--- a/apps/studio/components/ui/Charts/ComposedChart.utils.tsx
+++ b/apps/studio/components/ui/Charts/ComposedChart.utils.tsx
@@ -2,7 +2,7 @@
 
 import dayjs from 'dayjs'
 import { useState } from 'react'
-import { cn, Tooltip, TooltipContent, TooltipTrigger } from 'ui'
+import { cn, Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from 'ui'
 import { CHART_COLORS, DateTimeFormats } from './Charts.constants'
 import { numberFormatter } from './Charts.utils'
 import { formatBytes } from 'lib/helpers'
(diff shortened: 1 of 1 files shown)


joshenlim (approved):
LGTM! Tested on staging preview - just a quick feedback RE the delays though, wondering why we're adding delay for these tooltips? (if it might be a pattern we should be following for future/other UIs)

Also very tiny nit noticed but the delay no longer takes place once you hover across 2 tooltip groups

https://github.com/user-attachments/assets/2a4f89df-a8c5-422f-b3d5-6af4481543ff

jordienr:
> LGTM! Tested on staging preview - just a quick feedback RE the delays though, wondering why we're adding delay for these tooltips? (if it might be a pattern we should be following for future/other UIs)
> 
> Also very tiny nit noticed but the delay no longer takes place once you hover across 2 tooltip groups
> 
>  Screen.Recording.2025-09-23.at.11.17.55.mov

I found it annoying that the tooltip would appear as soon as you hovered so added the delay that felt natural, with the <TooltipProvider> around the group of tooltips it follows the rule of not having to wait if tooltips are part of the same group (see this: https://x.com/emilkowalski_/status/1962500739336462340)

PR #38866, opened 2025-09-19
coveralls: fail-on-error false

jordienr:
allow ci to continue if coveralls step fails, tests have to pass but coveralls step can fail


charislam (approved):
hallelujah 😅

PR #38869, opened 2025-09-19
refactor(docs): enhance the docs sidebar UI 

Devanshusharma2005:

YES/NO

fixes : #38861

Before : 
Screenshot 2025-09-19 at 9 07 54 PM

After : 

https://github.com/user-attachments/assets/80f502e8-80e5-4a52-bf84-28a3025c77ea

## What is the new behavior?

## Additional context


apps/docs/components/Navigation/NavigationMenu/RecursiveNavigation.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const pathname = usePathname()
+
+  const itemPath = path ? `${path}.${item.title}` : item.title
+  const isActive = item.url === pathname
+  const hasChildren = item.children && item.children.length > 0
+
+  if (item.enabled === false) {
+    return null
+  }
+
```
Devanshusharma2005: used memo so that the Only re-renders when item, depth, or path actually change. 
This should prevent the cascade re-renders.

apps/docs/components/Navigation/NavigationMenu/RecursiveNavigation.tsx
```diff
@@ -0,0 +1,127 @@
...
+const RecursiveNavigation: React.FC<RecursiveNavigationProps> = ({ items, className }) => {
+  return (
+    <nav className={cn('w-full space-y-1', className)}>
+      {items.map((item, index) => (
+        <RecursiveNavItem key={item.url || `${item.title}-${index}`} item={item} />
+      ))}
+    </nav>
+  )
+}
+
```
Devanshusharma2005: This is actually the main component that holds the entire navigation structure.

apps/docs/components/Navigation/NavigationMenu/RecursiveNavigation.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { cn } from 'ui'
+
+interface MenuItem {
+  title: string
+  icon?: string
+  url?: string
+  type?: 'link' | 'category'
+  enabled?: boolean
+  children?: MenuItem[]
+}
```
Devanshusharma2005: this is a recursive menu system using a clean MenuItem TypeScript interface. This means each menu item can have children, and those children can have their own and so on. No hardcoding. No limits.

Devanshusharma2005:
@charislam would love your review again . hehe ^^

Devanshusharma2005:
@kiwicopple would love your review too. hehe ^^

Devanshusharma2005:
Gentle reminder on this one 🙏

stylessh:
this is nice, thanks for the contribution!

a small issue: in larger sections the transition feels way to fast, and this causes the elements to overlap each other when hiding the section. 

https://github.com/user-attachments/assets/a574ffc6-f120-4d19-8889-bcd805e9cb61

also: when i join a section inside a dropdown and refresh, that collapsible should be open so I'd know where I'm located.

apps/docs/components/Navigation/NavigationMenu/RecursiveNavigation.tsx
charislam: our old types are already recursive, could we make this work with something closer to the old types? this will prevent having to deal with a bunch of edge cases during migration, e.g., not all icons have light mode, etc.

apps/docs/components/Navigation/NavigationMenu/NavigationMenuGuideListItems.tsx
```diff
@@ -31,32 +32,56 @@ const ContentAccordionLink = React.memo(function ContentAccordionLink(props: any
...
-        href={props.url}
-        className={props.className}
-        target={isExternal ? '_blank' : undefined}
-        rel={isExternal ? 'noopener noreferrer' : undefined}
-      >
-        {props.children}
-      </Link>
-    )
-  }
+  const LinkContainer = React.forwardRef<HTMLButtonElement | HTMLAnchorElement, any>(
```
charislam: let's move this out into an independent component, it'll be redefined (and be unmounted and remounted) on every render otherwise

charislam (commented):
This has been a problem for a while now, thanks for addressing it! ❤️ Looks really good, just some small feedback on top of @stylessh's

Devanshusharma2005:
> this is nice, thanks for the contribution!
> 
> a small issue: in larger sections the transition feels way to fast, and this causes the elements to overlap each other when hiding the section.
> 
>  Screen.Recording.2025-09-25.at.10.42.37.mov 
> 
> also: when i join a section inside a dropdown and refresh, that collapsible should be open so I'd know where I'm located.

Hey Alan , could you please help me with the logic of how can I make the collapsible state persist even after a refresh ? I am not able to find that logic.

Devanshusharma2005:
@charislam @stylessh I over-engineered the stuff too much. Made a good debugging and cleanup now.
State is persisting even after a refresh. Types are now recursive also as charis said. 

https://github.com/user-attachments/assets/e9733e79-0bdb-4472-8655-70a379c12a55

charislam:
Thanks @Devanshusharma2005! The more localized updates are much easier to review ❤️ 

Just one small outstanding problem:

In some cases, the positioning of the button shifts as the accordion opens and closes

https://github.com/user-attachments/assets/fa072377-2f28-4a98-9f76-e7cfc5566b62

Devanshusharma2005:
> Thanks @Devanshusharma2005! The more localized updates are much easier to review ❤️
> 
> Just one small outstanding problem:
> 
> In some cases, the positioning of the button shifts as the accordion opens and closes
> 
>  CleanShot.2025-09-26.at.16.11.17.mp4

@charislam This is just following the default behavior we have everywhere else. It's working fine for other nested items, so I figured keeping it consistent made sense. If I lock the chevron in a fixed position, it might actually be more annoying for users since they'd expect it to move with the content. Plus, as you can see, the border line is also stretching and opening according to the collapse which was the original behaviour, so everything moves together as intended. But hey, I'll be happy to be wrong here! What do you think?

charislam:
@Devanshusharma2005 yep, I saw it was the old behavior that the width of the entire thing changes 😅 I think it was fine before because it was mostly the line shrinking/expanding which is relatively subtle and doesn't disturb the user experience, but the dropdown indicator shifting is just really odd

it's probably a simple CSS thing to make the inner element take up the full space, just need to find out where to apply it 😄

Devanshusharma2005:
@charislam figured it out. It was a very tiny fix. Idk why it took me so much time to debug it. Maybe I need to re-visit my css skills haha ^^

charislam (approved):
Nice, this is great, thank you!

Devanshusharma2005:
@charislam thank you ^_^

PR #38881, opened 2025-09-19
FE-1864 FE-1868 reports kaizen 2

jordienr:
- crosshair in charts
- make chart highlight actions configurable per chart
- add chart actions to auth report

## to test
- go to auth or database report
- select a part of the chart
- go to logs
- shows the logs filtered by the range you selected in the chart

### DIFF
--- a/apps/studio/components/ui/Charts/ChartHighlightActions.tsx
+++ b/apps/studio/components/ui/Charts/ChartHighlightActions.tsx
@@ -1,6 +1,7 @@
-import React, { useEffect, useMemo, useState } from 'react'
 import dayjs from 'dayjs'
 import { ArrowRight, SearchIcon } from 'lucide-react'
+import { ReactNode, useEffect, useMemo, useState } from 'react'
+
 import {
   cn,
   DropdownMenu,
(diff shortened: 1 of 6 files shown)


apps/studio/components/ui/Charts/ChartHighlightActions.tsx
```diff
@@ -60,31 +91,38 @@ const ChartHighlightActions = ({
...
+          }
+          let rightNode: React.ReactNode = null
+          if (typeof action.rightSlot === 'function') {
+            rightNode = ctx ? action.rightSlot(ctx) : null
+          } else {
+            rightNode = action.rightSlot ?? null
+          }
+          return (
+            <DropdownMenuItem
+              key={action.id}
```
joshenlim: missing a `disabled={disabled}` property on the DropdownMenuItem here?
should apply the styles for disabled state and won't need the bg-transparent below i think

image

apps/studio/components/ui/Charts/ComposedChartHandler.tsx
```diff
@@ -231,6 +232,22 @@ const ComposedChartHandler = ({
           : (firstData.data[firstData.data.length - 1] as any)?.[firstAttr.attribute]
   }, [highlightedValue, attributes, attributeQueries])
 
+  const highlightActions: ChartHighlightAction[] = useMemo(() => {
+    return [
+      {
+        id: 'open-logs',
+        label: 'Open in Postgres Logs',
+        icon: <LogsIcon size={12} />,
```
joshenlim: Very minor but reckon we can use the same icon as per the one in the side nav

image
image

joshenlim: pushed up changes for this 🙏 🙂

apps/studio/components/ui/Charts/ChartHighlightActions.tsx
joshenlim: pushed up changes for this 🙏 🙂

joshenlim (approved):
LGTM! Tested on staging preview 🙂

jordienr:
tests passed but coveralls broke again so merging

PR #38891, opened 2025-09-20
Fix new table button not working if on unknown table in table editor

joshenlim:
## Context

Addresses an issue where the "new table" button in the table editor wasn't working if we're on a unknown table (e.g /editor/123, where 123 is a random non existent ID)

Realised this was due to an early return statement in `TableGridEditor.tsx`, just had to shift the Unknown table UI state into the main return statement and also shift `SidePanelEditor` outside of the `TableEditorTableStateContextProvidert

^ This is fine since none of the side panels are using the `table-editor-table` state (we only need to make sure that the TableEditor side panel doesn't use actually since all the other panels cannot be accessed while on the Unknown table UI state.)

## To test
- [ ] Verify that you can create a new table while on an unknown table (e.g `/editor/123`)

### DIFF
--- a/apps/studio/data/table-editor/table-editor-query.ts
+++ b/apps/studio/data/table-editor/table-editor-query.ts
@@ -33,7 +33,7 @@ export async function getTableEditor(
     signal
   )
 
-  return (result[0]?.entity ?? undefined) as Entity | undefined
+  return (result[0]?.entity ?? null) as Entity | undefined
 }
 
 export type TableEditorData = Awaited<ReturnType<typeof getTableEditor>>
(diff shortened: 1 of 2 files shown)


apps/studio/data/table-editor/table-editor-query.ts
```diff
@@ -33,7 +33,7 @@ export async function getTableEditor(
     signal
   )
 
-  return (result[0]?.entity ?? undefined) as Entity | undefined
+  return (result[0]?.entity ?? null) as Entity | undefined
```
joshenlim, line 36: this is just to address a react query error - we shouldn't be saving `undefined` data into the react query state
image

jordienr (approved):
tested in preview
Image

PR #38893, opened 2025-09-20
a new hope

jordienr:
- runs e2e tests against CLI studio build in vercel
- fixes flaky tests


apps/studio/state/tabs.tsx
```diff
@@ -347,7 +347,7 @@ function createTabsState(projectRef: string) {
               router.push(`/project/${router.query.ref}/editor`)
               break
             default:
-              router.push(`/project/${router.query.ref}/${editor}`)
+              router.push(`/project/${router.query.ref}/${editor === 'table' ? 'editor' : 'sql'}`)
```
jordienr, line 350: found this bug while testing. 

we were redirecting to /ref/table instead of /ref/editor

joshenlim (approved):
Generally LGTM! Thanks so much for this ❤️ 🙏 🙂

PR #38894, opened 2025-09-20
Fix database webhooks http headers for edge functions not loading auth headers on first open

### LINKED ISSUE #38848: `Authorization` headers are deleted from DB webhooks on update
# Bug report

## Describe the bug

`Authorization`-type HTTP headers are deleted from HTTP database webhooks on webhook update, with no notification to the client.

These headers can be added successfully, however:
- They are ["hidden" from the client](https://github.com/supabase/supabase/blob/031c5644e7d6d3068206e7ce4ab6905f1634af20/apps/studio/components/interfaces/Database/Hooks/FormContents.tsx#L105) when re-opening the webhook edit panel
- They are deleted when any changes are made to the webhook (on "Update webhook"), whether headers were updated or not (the filtered list of headers is saved, effectively deleting the auth headers which were hidden)

## To Reproduce

1. Create a new table `foo` via UI with defaults
2. Create a DB webhook on `foo` for any operation, with a custom HTTP header `Authorization: Bearer 123`. I pointed this to a simple server which only logged headers/body and returned.
3. Save the webhook
4. Trigger the webhook, view request headers - both `Content-type` and `Authorization` come through. So far so good.
5. Open the webhook edit panel. Notice the `Authorization` header is "hidden" from the client ([code ref](https://github.com/supabase/supabase/blob/031c5644e7d6d3068206e7ce4ab6905f1634af20/apps/studio/components/interfaces/Database/Hooks/FormContents.tsx#L105))
6. **Do not change anything in the edit panel**, but click "Update webhook"
7. Trigger the webhook again. **The Authorization header is gone.**

## Expected behavior

I'd like some help defining the requirements here - happy to create a PR, but not sure what expected behavior _should_ be.

In my mind:
- Authorization headers can be added to HTTP webhooks
- Authorization headers are never [hidden](https://github.com/supabase/supabase/blob/031c5644e7d6d3068206e7ce4ab6905f1634af20/apps/studio/components/interfaces/Database/Hooks/FormContents.tsx#L105) from clients
- Authorization headers are not deleted when unrelated changes are made to an HTTP webhook

Let me know if you agree - but I'm sure the "hiding" was done for a reason, so I figured this should be discussed first.

### Fix options
I think there are two options:
1. Do not hide auth headers from clients - or, if this was for security purposes, show "Hidden headers..." and still allow clients to view them. Then, obviously, also don't delete them on save.
2. Disallow auth headers entirely, with clear communication when attempted. I personally don't like this option (I rely on the auth headers in my projects), but may be preferred, depending on why this behavior exists to begin with

## Screenshots

Video demo: https://www.loom.com/share/c3129858c963410d9f0f404adf3b3b29?sid=a68635a8-f7e4-42e3-b5d1-7bf78268fcb8

## System information

- OS: macOS
- Browser: tested on Chrome/Safari
- Version of Node.js: v22.19.0

## Additional context

I previously opened an issue to document this "hiding" behavior, but realized it was deeper than that (headers are actually deleted, not only hidden) - https://github.com/supabase/supabase/issues/38839

joshenlim:
Fixes https://github.com/supabase/supabase/issues/38848, related to https://github.com/supabase/supabase/pull/38844

## Context

Realised that for database webhooks, when editing a webhook that uses an edge function the authorization header doesn't load when you open the panel for the first time which seemingly looks like the authorization header was being hidden (but its not the case)

To reproduce
- Create a webhook that uses an edge function (Ensure that there's an auth header)
- Refresh the page then edit the webhook, notice that the authorization header is missing
- Close the panel then edit the webhook again, notice that the authorization header is now present

https://github.com/user-attachments/assets/d9b795f8-7d58-4831-8828-2184912e6d77

This was happening as the list of edge functions was not loaded yet in initial loading of HTTP headers in the client state within the useEffect. Just needed to check for `isSuccess` from the `useEdgeFunctionsQuery`

### DIFF
--- a/apps/studio/components/interfaces/Database/Hooks/EditHookPanel.tsx
+++ b/apps/studio/components/interfaces/Database/Hooks/EditHookPanel.tsx
@@ -44,6 +44,7 @@ export const EditHookPanel = ({ visible, selectedHook, onClose }: EditHookPanelP
   // hence why this external state as a temporary workaround
   const [events, setEvents] = useState<string[]>(selectedHook?.events ?? [])
   const [eventsError, setEventsError] = useState<string>()
+  const [isSubmitting, setIsSubmitting] = useState(false)
 
   // For HTTP request
   const [httpHeaders, setHttpHeaders] = useState<HTTPArgument[]>(() => {
@@ -65,6 +66,7 @@ export const EditHookPanel = ({ visible, selectedHook, onClose }: EditHookPanelP
       value,
     }))
   })
+
   const [httpParameters, setHttpParameters] = useState<HTTPArgument[]>(() => {
     if (typeof selectedHook === `undefined`) {
       return [{ id: uuidv4(), name: '', value: '' }]
@@ -90,7 +92,7 @@ export const EditHookPanel = ({ visible, selectedHook, onClose }: EditHookPanelP
     projectRef: project?.ref,
     connectionString: project?.connectionString,
   })
-  const [isSubmitting, setIsSubmitting] = useState(false)
+
   const { mutate: createDatabaseTrigger } = useDatabaseTriggerCreateMutation({
     onSuccess: (res) => {
       toast.success(`Successfully created new webhook "${res.name}"`)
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Database/Hooks/EditHookPanel.tsx
joshenlim, line 1: this was just shifting stuff around sorry 🙏

jordienr (approved):

PR #38897, opened 2025-09-21
Update overview.mdx docs: Article change from 'an' to 'a'

tusharkashyap63:

YES

Docs update

The _database overview_ docs use the wrong article _an_ before SQL
> In 1994, Postgres95 was built on top of POSTGRES code, adding an SQL language interpreter as a replacement for PostQUEL.

## What is the new behavior?

> In 1994, Postgres95 was built on top of POSTGRES code, adding a SQL language interpreter as a replacement for PostQUEL.


ChrisChinchilla:
@tusharkashyap63 So… SQL is an odd case, as it depends if you say "ES Queue Ell" or "See Queue Ell"… and generally the Postgres community says "ES Queue Ell", so the current is correct… But thanks for the PR!

PR #38903, opened 2025-09-22
Refine home styles

SaxonF:
- Inverts background images in getting started depending on theme
- Each row in home is now consistent in height

Note that to do this I also updated NoDataPlaceholder and LogsBarChart to adopt height of container via 100% height vs static height value.


dnywh (approved):
Looks good! Charts working as expected too.

FYI the supporting image for the _Create your first table_ step is not showing on staging but works fine locally. If I remember right, it was an issue in a prior preview link too. So unrelated and likely not a problem.

dnywh (approved):
Ran again locally with most recent changes. Looks good

PR #38906, opened 2025-09-22
In UI library docs, selecting a line shouldn't select $

Dineshs91:

YES

Docs update to UI Library

Currently when a user selects the command from the docs, it selects the `$` as well.

CleanShot 2025-09-22 at 10 29 01@2x

## What is the new behavior?

This will ensure that only the command is selected. `$` is excluded from the selection

CleanShot 2025-09-22 at 10 30 35@2x

## Additional context

Here's a demo

https://github.com/user-attachments/assets/eaf7914f-164b-4703-b3fa-78fe5e816b61

### DIFF
--- a/apps/ui-library/components/command.tsx
+++ b/apps/ui-library/components/command.tsx
@@ -67,7 +67,7 @@ export function Command({ name, highlight }: CommandCopyProps) {
             <TabsContent_Shadcn_ key={manager} value={manager} className="m-0">
               <div className="flex items-center">
                 <div className="flex-1 font-mono text-sm text-foreground relative z-10">
-                  <span className="mr-2 text-[#888]">$</span>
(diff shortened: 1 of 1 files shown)


apps/ui-library/components/command.tsx
```diff
@@ -67,7 +67,7 @@ export function Command({ name, highlight }: CommandCopyProps) {
             <TabsContent_Shadcn_ key={manager} value={manager} className="m-0">
               <div className="flex items-center">
                 <div className="flex-1 font-mono text-sm text-foreground relative z-10">
-                  <span className="mr-2 text-[#888]">$</span>
+                  <span className="mr-2 text-[#888] select-none">$</span>
```
saltcod, line 70: clever. 
came here expecting to see a js solution!

saltcod (approved):
clever, clever, clever

PR #38908, opened 2025-09-22
Bump zod to v4

joshenlim:
## Context

In efforts to address "Type instantiation is excessively deep and possibly infinite." - many articles suggest Zod V3 might be the culprit, and that upgrading to V4 may be the solution, so giving that a try in this PR. The previous PR [here](https://github.com/supabase/supabase/pull/38846) standardises the Zod version across all apps and packages in the monorepo to use the same version of Zod which mights this transition slightly easier.

Am mainly using this zod-v3-to-v4 codemod [here](https://github.com/nicoespeon/zod-v3-to-v4) to help with the transition, with some manual eye-balling to make sure there's no weird changes

# Changes involved

- Shifted `zod` and `@hookforms/resolver` to `pnpm-workspace` catalog
  - Bump `zod` to `^4.0.0`
  - Bump `@hookforms/resolver` to `^5.1.0`
- Updated usage of `zod` based on v4 migration using codemod
- Added `tailwindcss` as a dev dependency to `docs`
  - Tbh i'm really not sure why this is needed but without this, `docs` and `www` build were complaining of `Error: Cannot find module 'tailwindcss' errors ([ref](https://vercel.com/supabase/docs/2MXiYUJcTqy67FWSoRGMnuFmXKFR)) while others like `ui-library`, `design-system` were complaining of `Module not found: Can't resolve 'tailwindcss/plugin'` ([ref](https://vercel.com/supabase/ui-library/CHBRyRJygTcaekMKxxoDs9e6TkXB))

## To test
- [ ] General smoke test of the dashboard
  - Specifically also look out for these 2 (had to manually fix outside of the codemod)
    - Moving queries in SQL Editor (to existing folder and new folder)
    - Updating AI opt in levels in org settings
  - And this one too (had to bump `hookforms/resolvers` because of Zod errors were throwing client side exceptions which our unit tests caught)
    - Storage -> Delete bucket modal, verify that error handling (e.g not typing the correct bucket name in the input) is handled as expected, shouldn't be a client side crash
- [ ] General smoke test of the docs
- [ ] General smoke test of the www

### DIFF
--- a/apps/design-system/package.json
+++ b/apps/design-system/package.json
@@ -16,7 +16,7 @@
     "typecheck": "contentlayer2 build && tsc --noEmit -p tsconfig.json"
   },
   "dependencies": {
-    "@hookform/resolvers": "^3.1.1",
+    "@hookform/resolvers": "catalog:",
     "contentlayer2": "0.4.6",
     "date-fns": "^2.30.0",
     "eslint-config-supabase": "workspace:*",
--- a/apps/design-system/registry/default/example/calendar-form.tsx
+++ b/apps/design-system/registry/default/example/calendar-form.tsx
@@ -25,8 +25,8 @@ import {
 
 const FormSchema = z.object({
   dob: z.date({
-    required_error: 'A date of birth is required.',
-  }),
+      error: (issue) => issue.input === undefined ? 'A date of birth is required.' : undefined
+}),
 })
 
 export default function CalendarForm() {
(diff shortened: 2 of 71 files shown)


kevcodez:
Studio (tsc diagnostics)

Before
Instantiations:            13656016
Memory used:               2059744K
Total time:                  57.83s

After
Instantiations:            13400288
Memory used:               1980829K

Seems like the culprit for the many instantiations may not be zod (still good to upgrade)

staaldraad (approved):

PR #38910, opened 2025-09-22
Getting started step status

SaxonF:
image

This hooks up the status of each getting started step. Not completely accurate just yet and stats may change but lays the foundation . You can check `useGettingStartedProgress` to see how we are checking step.  `Connect your application` step just checks requests which is not ideal and instead should be watching for external requests.

Main concern is just around performance since steps are pretty wide ranging so will need to make sure refetching is optimal. 

To test, try create a new project and start configuring things.


apps/studio/components/interfaces/HomeNew/CustomReportSection.tsx
```diff
@@ -278,7 +278,7 @@ export function CustomReportSection() {
   return (
     <div className="space-y-6">
       <div className="flex items-center justify-between">
-        <h3 className="heading-section">At a glance</h3>
+        <h3 className="heading-section">Report</h3>
```
joshenlim: nit - reckon plural might be better?

apps/studio/components/interfaces/HomeNew/CustomReportSection.tsx
```diff
@@ -298,14 +298,18 @@ export function CustomReportSection() {
         {(() => {
           if (layout.length === 0) {
             return (
-              <div className="flex min-h-[270px] items-center justify-center rounded border-2 border-dashed p-16 border-default">
+              <div className="h-64 flex flex-col items-center justify-center rounded border-2 border-dashed p-16 border-default">
+                <h4>Build a custom report</h4>
```
joshenlim: wondering if stating that these are from SQL snippets might be better? Since the add block dropdown will show a list of SQL snippets

"Build a custom report with SQL"

joshenlim: Feedback - reckon we could add a CTA at the bottom here to "Create your first block with SQL" which directs them to the SQL editor

image

Also feedback - at the top of my head wondering if it might make sense to introduce a category of SQL snippet templates for building custom reports (understandably the needs of all projects will be different, but perhaps just the rough building blocks might be useful - e.g templates with SQL aggregation of data)

apps/studio/components/interfaces/HomeNew/CustomReportSection.tsx
```diff
@@ -298,14 +298,18 @@ export function CustomReportSection() {
...
+                <p className="text-sm text-foreground-light mb-4">
+                  Keep track of your most important metrics{' '}
```
joshenlim:
```suggestion
                  Keep track of your most important metrics
```

apps/studio/components/interfaces/HomeNew/CustomReportSection.tsx
```diff
@@ -298,14 +298,18 @@ export function CustomReportSection() {
         {(() => {
```
joshenlim: unrelated but just noticed - this dynamic function seems unnecessary, can be simplified with `{condition && component}` i think

apps/studio/components/interfaces/HomeNew/CustomReportSection.tsx
```diff
@@ -295,66 +295,65 @@ export function CustomReportSection() {
...
-              onDragStart={handleDragStart}
-              onDragEnd={handleDragEnd}
+        {layout.length === 0 ? (
+          <div className="h-64 flex flex-col items-center justify-center rounded border-2 border-dashed p-16 border-default">
+            <h4>Build a custom report</h4>
+            <p className="text-sm text-foreground-light mb-4">
+              Keep track of your most important metrics
+            </p>
+            {canUpdateReport || canCreateReport ? (
+              <SnippetDropdown
```
joshenlim, line 305: unrelated as well but i just realised SnippetDropdown's internal searching is only client side searching but we need to support server side searching due to pagination (will look into it)

apps/studio/components/interfaces/HomeNew/CustomReportSection.tsx
joshenlim: [Edit] Added the CTA at the bottom
image

apps/studio/components/interfaces/HomeNew/GettingStarted/useGettingStartedProgress.ts
```diff
@@ -0,0 +1,147 @@
...
+  const statuses = useMemo<GettingStartedStatuses>(() => {
+    const hasTables = (tablesData?.length ?? 0) > 0
+    const hasCliSetup = (migrationsData?.length ?? 0) > 0
+    const hasSampleData = (tablesData ?? []).some(
+      (table) => Number(table?.live_rows_estimate ?? 0) > 0
+    )
+    const hasRlsPolicies = (policiesData?.length ?? 0) > 0
+    const allowSignupsEnabled = authConfig ? !authConfig.DISABLE_SIGNUP : false
+    const emailProviderEnabled = !!authConfig?.EXTERNAL_EMAIL_ENABLED
+    const hasConfiguredAuth = allowSignupsEnabled && emailProviderEnabled
```
joshenlim, line 97: This check will return true for all new projects btw - if it might be an issue (looks weird from a UX POV)

image

apps/studio/components/interfaces/HomeNew/GettingStarted/useGettingStartedProgress.ts
```diff
@@ -0,0 +1,147 @@
...
+    const hasFirstUser = (usersCountData?.count ?? 0) > 0
```
joshenlim: we've recently introduced an estimated count for users [here](https://github.com/supabase/supabase/pull/38638) as doing a select * from auth.users on a project with millions of users can cause the database to degrade

the estimate count isn't guaranteed to be accurate hence why this may end up being checked for a brand new project as per the screenshot [here](https://github.com/supabase/supabase/pull/38910/files#r2405207047)

i'll update this check to consider the estimation

apps/studio/components/interfaces/HomeNew/GettingStarted/useGettingStartedProgress.ts
joshenlim, line 1: just a heads up -  we can address this separately but do we have plans to hide the GettingStarted section completely for older projects when we launch the new Home V2?

concern here being that we're firing a lot of queries on the database from this particular UI which may cause performance issues on projects which DB are considerably big

apps/studio/components/interfaces/HomeNew/GettingStarted/GettingStarted.utils.tsx
```diff
@@ -0,0 +1,401 @@
...
+      ],
+    },
+    {
+      key: 'upload-file',
+      status: hasStorageObjects ? 'complete' : 'incomplete',
+      title: 'Upload a file',
+      icon: <Upload strokeWidth={1} className="text-foreground-muted" size={16} />,
+      description:
+        'Integrate file storage by creating a bucket and uploading a file, starting from the buckets dashboard linked below.',
+      actions: [{ label: 'Buckets', href: `/project/${ref}/storage/buckets`, variant: 'default' }],
```
joshenlim: Nit but just wondering - given this is part of the code workflow - should the CTA here be to upload a file through the UI? Or using a client library?

apps/studio/components/interfaces/HomeNew/GettingStarted/GettingStarted.utils.tsx
```diff
@@ -0,0 +1,401 @@
...
+      icon: <Code strokeWidth={1} className="text-foreground-muted" size={16} />,
+      description:
+        'Add server-side logic by creating and deploying your first Edge Function—a lightweight TypeScript or JavaScript function that runs close to your users—then revisit the list to monitor and iterate on it.',
+      actions: [
+        {
+          label: 'Create a function',
+          href: `/project/${ref}/functions/new`,
+          variant: 'default',
+        },
+        { label: 'View functions', href: `/project/${ref}/functions`, variant: 'default' },
```
joshenlim, line 209: likewise here - given that this is part of the code workflow, should the CTA here be to use the CLI? or the dashboard

joshenlim (approved):
Approving to unblock but JFYI with a new project the "Allow user signups" step is already checked which might be odd for someone coming into Supabase for the first time

image

PR #38914, opened 2025-09-22
fix redirect loop

jordienr:
fixes infinite redirect loop in sql editor


alaister (approved):
LGTM!

- Tested on preview and self-hosted preview

PR #38931, opened 2025-09-22
Improve stability and explicitness of table editor tab state syncing between URL params and local storage

matlin:

YES

This fixes some buggy/inconsistent behavior with how caching of the params to the Table Editor tabs

In some cases (like when clicking on a referenced table link from one tab to another) filters would correctly update the state and URL query params but would not update the Local Storage cache. This would cause weird behaviors when clicking different links in the UI which sometimes would load correctly from storage while others would "revert" to an older and incorrect state. 

Also trying to update the state by manually change or deleting the URL params for `filter` and `sort` works in all cases except if you delete all params in which case it implicilty (and unexpectedly IMO) loads from cache. 

## What is the new behavior?

Now all you have to do is update the URL params and both the state and cache/storage will update automatically--you no longer need to manually update the cache. Loading from cache/local-storage now requires an explicit URL param: `loadFromCache=true`. The extra URL param is somewhat annoying but I think it's worth it for the caching stability.


charislam (approved):
Pushed a commit with one fix: the tab state was getting reset when navigating to another page and then back to the Table Editor using the side navigation bar

Otherwise LGTM!

PR #38933, opened 2025-09-22
feat: add basic table template infrastructure (1 of 3)

seanoliver:
## Summary
Implements foundation for [table quickstart experiment](https://www.notion.so/supabase/Table-Editor-Quickstart-2735004b775f80f680eaeb3c6e581633?source=copy_link) with pre-defined templates.

Users can select from categorized templates to quickly create common table structures.

Resolves GROWTH-504

## Changes
- Add pre-defined table templates for common use cases (social network, e-commerce, blog, to-do list, analytics)
- Create simple template selector UI with category browsing
- Add shared UI components for template selection
- Integrate template selection into TableEditor
- Set up type definitions and utils
- Add localStorage handling for dismiss state
- Add toast notifications for template application

## How to Test
1. Navigate to the **Table Editor** page
2. Click **"Create a table"** button
3. The template selector will appear at the top of the form (if project is less than 7 days old)

### Setting the Feature Flag
To see the template selector, you need to set the feature flag:
1. Go to PostHog (make sure you're in the **staging** project)
2. Find the `tableQuickstart` feature flag
3. Click **Edit**
4. Under **"Optional override"**, select **"templates"**
5. Save the changes

**Note:** In this PR, the quickstart component will ONLY appear when the feature flag is set to `"templates"`. Any other value (including `"control"`, `"ai"`, or undefined) will hide the component.

## Related PRs
- **Part 1 of 3** in the table quickstart feature series
- Part 2: #38934 (View system + AI generation)
- Part 3: #38932 (Telemetry tracking)

## Screenshots
Step 1: Quickstart chooser 
CleanShot 2025-10-10 at 10 50 24@2x

Step 2: Table selector and autofill
CleanShot 2025-10-10 at 10 52 30@2x

PR #38932, opened 2025-09-22
feat: add telemetry tracking for table quickstart experiment (3 of 3)

seanoliver:
## Summary
Adds comprehensive telemetry tracking for the [table quickstart experiment](https://www.notion.so/supabase/Table-Editor-Quickstart-2735004b775f80f680eaeb3c6e581633?source=copy_link).

Tracks user interactions across both templates and AI variants to measure engagement and conversion.

Resolves GROWTH-504

### Event Tracking
- `tableeditor_quickstart_viewed` - Exposure event on mount
- `tableeditor_quickstart_category_selected` - Template category chosen
- `tableeditor_quickstart_template_selected` - Template applied
- `tableeditor_quickstart_ai_selected` - AI option chosen
- `tableeditor_quickstart_quick_idea_selected` - Quick idea clicked
- `tableeditor_quickstart_ai_generation_started` - AI generation initiated
- `tableeditor_quickstart_ai_generation_succeeded` - AI tables generated
- `tableeditor_quickstart_ai_generation_failed` - AI generation error
- `tableeditor_quickstart_abandoned` - User left without completing

### Metadata Tracking
- Template name and source (template vs AI)
- Time spent in quickstart flow
- User interaction stage when abandoned
- AI prompt length and tables generated
- Error messages for debugging

### Implementation
- Event hooks in TableTemplateSelector component
- Abandonment detection on unmount
- Metadata passed through to table creation
- Type-safe event definitions in telemetry-constants

## How to Test
1. Set feature flag to `"templates"` or `"ai"`
2. Open browser DevTools → Network tab
3. Create a new table and interact with quickstart
4. Verify telemetry events are sent to PostHog
5. Check event properties contain expected metadata

### Events to Verify
- Viewed event fires on mount
- Category/AI selection events fire on clicks
- Template selection includes template name
- Abandonment event fires on dismiss/back navigation
- AI generation events include prompt length

## Related PRs
- Part 1: #38933 (Basic templates foundation)
- Part 2: #38934 (AI generation + view system)
- **Part 3 of 3** (this PR - Telemetry tracking)

## Dependencies
- Requires #38934 to be merged first (adds view system and AI generation)

PR #38934, opened 2025-09-22
feat: add AI-powered table generation variant (2 of 3)

seanoliver:
## What's this?
Adds AI-powered table generation as the second variant for the [table quickstart experiment](https://www.notion.so/supabase/Table-Editor-Quickstart-2735004b775f80f680eaeb3c6e581633?source=copy_link). This helps new users get started faster by letting them describe their app and having AI generate a complete database schema.

Resolves GROWTH-505

### Architecture: Moved to Empty State (After Design Review)
After discussing with design, we moved the quickstart widget **out of the table editor sidebar** and into the **table editor empty state** (`NewTab.tsx`). This provides a better UX by:
- Making the quickstart more discoverable when users first land on the Table Editor
- Avoiding sidebar clutter and keeping the create table form clean
- Better aligning with the user's mental model (start here → then customize in the form)
- Allowing the widget to be larger and more prominent

The widget appears on the empty state page for projects less than 7 days old, positioned between the "Create a table" action card and recent items.

### AI Table Generation
- Prompt-based schema generation using `gpt-5-mini`
- Generates exactly 3 related tables per prompt
- Streaming responses with AI SDK's `streamObject()` for real-time feedback
- Quick idea buttons for common use cases ("Recipe sharing app", "Event ticketing system", etc.)

### API Implementation (`/api/ai/table-quickstart/generate-schemas`)
- Uses AI SDK's `streamObject()` with Zod schema validation for structured generation
- Streams partial objects as they're generated (tables appear incrementally)
- 50KB buffer limit to prevent memory issues
- Client-side uses `Object.assign()` to merge partial objects efficiently
- Proper abort handling when user navigates away

### State Management
- Extended `onAddTable()` to accept optional `templateData` parameter
- Templates/AI tables flow through the same state mechanism
- When user selects a generated table, it opens the table editor form pre-filled with all fields

## Why these choices?

**Why empty state instead of sidebar?**
Design feedback showed that putting this in the sidebar felt cramped and less discoverable. The empty state is:
- The first thing users see when they land on Table Editor with no tables
- Spacious enough for the widget to breathe
- Naturally positioned as the "starting point" before creating custom tables
- Less intrusive than injecting into the create form

**Why exclude foreign keys?**
The AI can't create meaningful foreign key relationships to tables that don't exist yet. We can add this later if useful, but for now it's cleaner to skip it and let users add relationships manually and use the generated tables as a personalized starting point.

**Why 7-day project gate?**
Only show this to new projects where users are still figuring things out. More established projects probably don't need this kind of hand-holding. Also this is purely an activation experiment for now. If we wanted to build this out into a more fully fledged feature, we'd need to prioritize additional capabilities (foreign keys, multi-table creation, etc.).

**Why 50KB buffer limit?**
Protects against pathological AI responses that could cause memory issues. 50KB is more than enough for 3 table schemas with columns and descriptions. If we hit the limit, we can still use the partial response.

### AI Variant (New in this PR)

**Basic Steps**
1. Navigate to the Table Editor page
2. If you have existing tables, delete them all to see the empty state
3. The AI table generator will appear below the "Create a table" button (if project is less than 7 days old)

**Setting the Feature Flag**

To see the AI generator, you need to set the feature flag:
1. Go to PostHog (make sure you're in the staging project)
2. Find the `tableQuickstart` feature flag
3. Click Edit
4. Under "Optional override", select `"ai"`
5. Save the changes

**What to expect**: 
- Input field with quick idea buttons below the "Create a table" card
- Type a prompt (e.g., "fitness tracking app") or click a quick idea
- Hit Generate, tables stream in near-instantly
- Click a table to open the form pre-filled with all fields

### Templates Variant (For validation - already reviewed in #38933)

**Basic Steps**
1. Navigate to the Table Editor empty state
2. The template selector will appear (if project is less than 7 days old)

**Setting the Feature Flag**

To see the template selector:
1. Go to PostHog (staging project)
2. Find the `tableQuickstart` feature flag
3. Set to `"templates"`

**What to expect**: Category tabs at the top, browse templates by category, click to apply.

### Control

To hide the component entirely (baseline for experiment):
1. Set the `tableQuickstart` flag to `"control"` (or disable the flag)

**Note:** The component only shows for projects fewer than 7 days old.

## Notes
This branch contains **both** the templates variant from #38933 and the new AI variant. The templates PR was merged to master first, so this branch has both implementations for easier testing of the experiment variants side-by-side.

Related: Part 3 (#38932) adds telemetry tracking for the experiment.

## Screenshots
`ai` variation on empty state with quick idea suggestions:
CleanShot 2025-10-20 at 20 14 20@2x

`ai` variation results view showing 3 generated tables:
CleanShot 2025-10-20 at 20 15 51@2x

`templates` variation shown in the table editor empty state page:
CleanShot 2025-10-20 at 20 12 54@2x


pamelachia:
Thoughts as I was testing: 
- if there was an option to tick and add sample data together with the table
- if i can create multiple tables at the same time
- can it read the schema of some earlier created tables to create new tables based on that

I guess that is something our AI assistant can do so maybe after the first table creation we redirect them to the assistant?

apps/studio/pages/api/ai/table-quickstart/generate-schemas.ts
```diff
@@ -0,0 +1,221 @@
...
+})
+
+const ResponseSchema = z.object({
+  tables: z.array(TableSchema).describe('Array of related database tables for the application'),
+  summary: z.string().optional().describe('Brief summary of the generated schema'),
+})
+
+type SchemaStreamResult = ReturnType<typeof streamObject<typeof ResponseSchema>>
+
+const handler = async (req: NextApiRequest, res: NextApiResponse) => {
```
charislam: For consistency with other API routes, we like to wrap the handler like this:

```
async function handler(req: NextApiRequest, res: NextApiResponse) {
  const { method } = req

  switch (method) {
    case 'POST':
      return handlePost(req, res)
    default:
      res.setHeader('Allow', ['POST'])
      res.status(405).json({
        data: null,
        error: { message: `Method ${method} Not Allowed` },
      })
  }
}
```

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
```diff
@@ -0,0 +1,416 @@
...
+    if (Array.isArray(value)) {
+      const existingArray = Array.isArray(target[key]) ? (target[key] as Array<any>) : []
+      const merged = value.map((item, index) => {
+        if (item === null || item === undefined) {
+          return existingArray[index]
+        }
+
+        if (typeof item === 'object' && !Array.isArray(item)) {
+          const existing = existingArray[index]
+          return mergeDeep((existing as Record<string, any>) ?? {}, item as Record<string, any>)
```
charislam: Is `existing` guaranteed to be a plain object here, or is it potentially an array or primitive?

Would it make sense to use [lodash's merge](https://lodash.com/docs/4.17.15#merge) or do we need slightly different functionality? If we need our own, we should write comprehensive tests for it because it doesn't have much type safety and is probably prone to breaking unexpectedly on edge cases

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
```diff
@@ -0,0 +1,416 @@
...
+
+    try {
+      const headers = await constructHeaders()
+      headers.set('Content-Type', 'application/json')
+      headers.set('Accept', 'text/event-stream')
+
+      const response = await fetch(`${BASE_PATH}/api/ai/table-quickstart/generate-schemas`, {
+        method: 'POST',
+        headers,
+        credentials: 'include',
```
charislam: Is this necessary? The request shouldn't be cross-origin.

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
charislam, line 1: In general there's a lot of low-level stuff in here that we can avoid having to debug/maintain if we use the [AI SDK](https://ai-sdk.dev/docs/introduction), which we're already using for all our other AI functionality

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/TableTemplateSelector.tsx
charislam, line 1: There's a lot of conditional logic and many branches to keep in mind now because the two variants are both mixed in here. It would be better to have two different components for the variants, and a main switch component that switches between the two. If any design scaffolding needs to be shared, it can be abstracted out.

charislam (changes requested):
It occurs to me now that the "quickstart dismissed" value in local storage should be keyed to project ref, so a user will still see quickstarts for new projects even if they've dismissed it in the past

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/TableTemplateSelector.tsx
seanoliver: Agreed — good call!

apps/studio/pages/api/ai/table-quickstart/generate-schemas.ts
seanoliver: Ah okay thanks for pointing this out! I'd noticed that pattern but thought it might be overkill for a single-method endpoints. I'll update this to align!

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
seanoliver: Great point. I'll refactor to use lodash's `merge`.

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
seanoliver: Ah thanks for catching this. It's a force of habit but not necessary in this case.

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
seanoliver: Yeah, good call. The original intent was to move most of the parsing client-side to speed up AI responses, but you’re right — it’s going to be tricky to maintain. I’ve reverted it to use the AI SDK instead.

I also optimized the prompt a bit to cut down latency, so responses are now around 6–8 seconds. That should be fine for now since this is still experimental. If it proves successful and we decide to roll it out more broadly, we can look into adding a smaller model to handle requests even faster.

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/QuickstartAIWidget.tsx
```diff
@@ -0,0 +1,190 @@
...
+            }}
+            disabled={isGenerating || disabled}
+            aria-label="Table description for AI generation"
+            className="pr-24"
+          />
+          <Button
+            type="default"
+            size="tiny"
+            disabled={!aiPrompt.trim() || isGenerating || disabled}
+            onClick={() => handleGenerateTables()}
```
charislam, line 119:
```suggestion
            onClick={handleGenerateTables}
```

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/QuickstartAIWidget.tsx
```diff
@@ -0,0 +1,190 @@
...
+            icon={<AiIconAnimation loading={isGenerating} />}
+            className="absolute right-1 top-1/2 -translate-y-1/2 !space-x-1"
+          >
+            {isGenerating ? 'Generating...' : 'Generate'}
+          </Button>
+        </div>
+
+        {apiError && (
+          <div className="text-sm text-red-600" role="alert">
```
charislam, line 132: We can leave the `role="alert"` as we have no better mechanism to handle this right now, but dynamically adding `role="alert"` typically doesn't work for screen readers -- it's more robust to have a static alert element that's always on the page, and to populate it with new messages as needed

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/QuickstartAIWidget.tsx
```diff
@@ -0,0 +1,190 @@
...
+            {apiError}
+          </div>
+        )}
+
+        {tables.length === 0 && (
+          <div>
+            <div className="text-xs text-foreground-light mb-2">Quick ideas:</div>
```
charislam: a header here would have better semantics

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
```diff
@@ -0,0 +1,294 @@
...
+  const generateTables = useCallback(async (prompt: string): Promise<TableSuggestion[]> => {
+    if (abortControllerRef.current) {
+      abortControllerRef.current.abort()
+    }
+
+    if (prompt.length > LIMITS.MAX_PROMPT_LENGTH) {
+      const message = `Your description is too long. Try shortening it to under ${LIMITS.MAX_PROMPT_LENGTH} characters.`
+      if (isMountedRef.current) {
+        setError(message)
+      }
```
charislam:
```suggestion
        setError(message)
```

Any particular reason for needing to check `isMountedRef` here? This is before any asynchronous activity so I don't think there is an opportunity for unmounting between the click event and this running, but I could be missing something

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    const abortController = new AbortController()
+    abortControllerRef.current = abortController
+
+    if (isMountedRef.current) {
+      setIsGenerating(true)
+      setError(null)
+      setTables([])
+    }
```
charislam:
```suggestion
      setIsGenerating(true)
      setError(null)
      setTables([])
```

same here

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    try {
+      const headers = await constructHeaders()
+      headers.set('Content-Type', 'application/json')
+
+      const response = await fetchHandler(`${BASE_PATH}/api/ai/table-quickstart/generate-schemas`, {
```
charislam, line 185: Consider using the `useObject` API if possible here: https://ai-sdk.dev/cookbook/next/stream-object#client

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
```diff
@@ -0,0 +1,294 @@
...
+    if (abortControllerRef.current) {
+      abortControllerRef.current.abort()
+      abortControllerRef.current = null
+    }
+
+    if (isMountedRef.current) {
+      setIsGenerating(false)
+      setError(null)
+      setTables([])
+    }
```
charislam:
```suggestion
      setIsGenerating(false)
      setError(null)
      setTables([])
```

same here about the `isMountedRef`

charislam:
Oh, another thing I noted when testing:
- If the AI generates multiple tables, and you create one
- You then lose the others, because it immediately opens the newly created table in a new tab
- And even if you close that tab, the old AI suggestions are lost so you have to regenerate from scratch

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/useAITableGeneration.ts
seanoliver: @charislam and I touched base on slack and agreed it’s best to hold off on adopting `useObject` until it’s out of experimental. I’ve added a note in the code to capture that decision so future contributors know this was intentional. Once the API stabilizes, we can revisit and simplify this logic (it should replace roughly 150 lines of custom code).

seanoliver:
> Oh, another thing I noted when testing:
> 
> * If the AI generates multiple tables, and you create one
> * You then lose the others, because it immediately opens the newly created table in a new tab
> * And even if you close that tab, the old AI suggestions are lost so you have to regenerate from scratch


Great point! I've updated it to persist the last generation in localStorage so users can create multiple tables from the same results and even come back later to find them still there. They only get cleared when you click "Generate new" or do a new search.

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/QuickstartAIWidget.tsx
seanoliver: ah! that's good to know! i'll make note of this for future

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableQuickstart/QuickstartAIWidget.tsx
seanoliver: ah yeah — i used the inline arrow function here since `handleGenerateTables` takes an optional `tableData` param. if we pass it directly as a function reference, React sends the click event as that first arg and things break.

charislam: ahh missed that, thank you!

charislam (approved):

dnywh (approved):
Smoke tested both `ai` and templates` variants locally. I love this.

Will leave minor (non-blocking) feedback elsewhere.

PR #38939, opened 2025-09-23
feat: Add typechecking as a required check for the merge queue

ivasilov:
This PR adds the `typecheck` Github Action as a required check when merging PRs via the merge queue.

This check is added to avoid issues where the master branch and the PR branch have incompatible changes and may result in a bad build. This used to be checked by the Vercel build, but the Vercel build took too long so we removed typechecking as a step.

### DIFF
--- a/.github/workflows/typecheck.yml
+++ b/.github/workflows/typecheck.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - 'master'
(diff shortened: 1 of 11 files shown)


joshenlim (approved):
LGTM! Minimally did a basic smoke test on staging preview

fsansalvadore (approved):

PR #38941, opened 2025-09-23
Fix Postgres password issue and add password changing guidelines

dhirajjat:
… POSTGRES_PASSWORD

This PR addresses the issue #30524 regarding changing the Postgres password in .env while using Docker Compose.

Problem:
When the Postgres password is changed in .env, running docker compose up -d again causes errors. The documentation did not mention any steps to safely change the password.

Improvement / Fix:

Added clear instructions in the documentation on how to change the Postgres password safely.

Steps included:

Run docker compose down -v

Remove existing database data if needed: rm -rf volumes/db/data/

Run docker compose up -d again

This ensures changing the password does not result in errors.

Link to Documentation Updated:
[Supabase Docker Self-Hosting Guide](https://supabase.com/docs/guides/self-hosting/docker)

### DIFF
--- a/apps/docs/content/guides/self-hosting/docker.mdx
+++ b/apps/docs/content/guides/self-hosting/docker.mdx
@@ -98,9 +98,57 @@ docker compose up -d
 
 <Admonition type="tip">
 
-If you are using rootless docker, edit `.env` and set `DOCKER_SOCKET_LOCATION` to your docker socket location. For example: `/run/user/1000/docker.sock`. Otherwise, you will see an error like `container supabase-vector exited (0)`.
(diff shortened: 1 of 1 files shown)


singh-inder:
This method is only useful if you don't care about data in your db as this will delete everything.

dhirajjat:
> This method is only useful if you don't care about data in your db as this will delete everything.

Thank you for the feedback! I’ve updated the PR to include a safe method to change the Postgres password without deleting existing data.
The previous destructive method is still documented optionally for users who want a fresh database, but now the main instructions preserve data and include an optional backup step for extra safety.

aantti:
Some notes added [here](https://supabase.com/docs/guides/self-hosting/docker#changing-database-password). I'm going to close this PR.

PR #38952, opened 2025-09-23
FE-1857 update default api filter

jordienr:
- updates default api filter in reports from "matches" to "is"


charislam (approved):
checked on local that default value is 'is' and filtering works

PR #38957, opened 2025-09-23
Project dropdown fix staying on current page for selected project in onSelect

joshenlim:
Fixes DEPR-76

## Context

Realised that `onSelect` (using keyboard shortcuts) for `ProjectDropdown` wasn't using the route query to redirect the user when switching projects. Should stay on the same page when switching projects. Clicking works though, was just missing that behaviour for keyboard shortcut

## To test

- [x] From a project (any page e.g settings/infrastructure/general), swap to another project via the top nav dropdown. You should stay on that same page just different project


charislam (approved):
tested on preview, looks good

dnywh (approved):
Also tested on staging preview, also looks good!

PR #38962, opened 2025-09-23
feature: prioritize query performance in cmdk menu

charislam:
When the user's search term includes `query`, Query Performance Reports should jump to the top of the list.


jordienr (approved):
tested in preview all good!

PR #38965, opened 2025-09-23
fix(local-mcp): database operations

gregnr:
I accidentally broke MCP database operations during a refactor. This fixes them by unwrapping the return value instead of returning the wrapper itself. Also:
- Fixes `listMigrations()` to return empty array if the migrations table doesn't exist
- Adds `assertSelfHosted()` to prevent accidental usage of these functions outside of a self-hosted environment
- Improves types, removes casts

### DIFF
--- a/apps/studio/lib/api/self-hosted/query.ts
+++ b/apps/studio/lib/api/self-hosted/query.ts
@@ -1,23 +1,47 @@
-import { fetchPost } from 'data/fetchers'
 import { PG_META_URL } from 'lib/constants/index'
-import { ResponseError } from 'types'
 import { constructHeaders } from '../apiHelpers'
+import { DatabaseError, databaseErrorSchema, WrappedResult } from './types'
+import { assertSelfHosted } from './util'
 
 export type QueryOptions = {
   query: string
   headers?: HeadersInit
 }
 
-export async function executeQuery<T = unknown>({ query, headers }: QueryOptions) {
-  const response = await fetchPost<T[]>(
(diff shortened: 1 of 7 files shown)


apps/studio/lib/api/self-hosted/query.ts
```diff
@@ -1,23 +1,47 @@
...
-  } else {
-    return { data: response }
+  const response = await fetch(`${PG_META_URL}/query`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      ...constructHeaders(headers ?? {}),
+    },
+    body: JSON.stringify({ query }),
+  })
```
gregnr, line 29: Switching this to a vanilla `fetch()` in order to capture pg-meta specific database errors. I'm 99% sure we don't need the special `fetchPost()` logic in the context of server-side self-hosted code, but let me know if this is wrong.

apps/studio/lib/api/self-hosted/types.ts
```diff
@@ -0,0 +1,20 @@
+import z from 'zod/v4'
+
+export type WrappedSuccessResult<T> = { data: T; error: undefined }
+export type WrappedErrorResult = { data: undefined; error: Error }
+export type WrappedResult<R> = WrappedSuccessResult<R> | WrappedErrorResult
```
gregnr, line 5: We were already using this pattern previously on self-hosted functions, this just adds better type safety by making it explicit

charislam (approved):
tested locally with Claude Code

PR #38969, opened 2025-09-24
Reinstate ignoreBuildErrors as false for studio prod

joshenlim:
Reinstating `ignoreBuildErrors` to `false` for dashboard production

Now that we're running typechecking outside of the build, reckon we can flip this back to `false` to catch any errors arising from the typecheck


alaister (approved):
LGTM!

PR #38973, opened 2025-09-24
feat(etl): Add replication lag UI

iambriccardo:
This PR adds a new UI for exposing replication lag metrics to users which helps them to understand how well a pipeline is catching up with their inserts. In addition, it adds better copy for replication states and also shows when the replication live update is disconnected.

### Screenshots

image

image


joshenlim (approved):
Generally looks alright - Saxon's providing a mockup to improve UI consistency but reckon we can tackle that separately 🙂

PR #36867, opened 2025-07-03
feat: MCP URL builder

charislam:
- Creates an MCP URL builder component that allows a user to choose options via a UI, then automatically builds the URL and config needed for various MCP clients
- Adds this component to studio Config modal and docs MCP page

~~**Do not merge until closer to launch**
(Integration doesn't actually work right now because still waiting on OAuth flows for MCP server, but the UI should be fully functional)~~

image

## Merge instructions

Review https://github.com/supabase/supabase/pull/38981 which is a follow-up on this PR.

After merging this, merge that PR.

PR #38981, opened 2025-09-24
feat: local/self-hosted support in MCP URL builder

mattrossman:
Adds a platform selector to the MCP URL builder in docs so you can install MCP for:
- Hosted Supabase, (default, enables project selection)
- Local / self-hosted

The MCP URL construction logic for `studio` is similar to [this API_URL logic](https://github.com/supabase/supabase/blob/57fb359434ec72a4452006867175d2994985b8d2/apps/studio/lib/constants/index.ts#L7-L18) to handle different environments.

The `docs` assumes a local MCP URL of `http://127.0.0.1:54321/mcp` as prescribed in https://github.com/supabase/cli/pull/4186

Resolves AI-152

## Merge Instructions

Review https://github.com/supabase/supabase/pull/36867 first for the initial MCP URL builder implementation. 

Merge this PR after that. This PR adds self-hosted/local info to the initial implementation and makes more parts of the URL construction logic configurable.


mattrossman:
Screenshots of how the URL looks in the various apps/states/environments:

<details>

<summary>
Docs preview (hosted selected)
</summary>

CleanShot 2025-09-25 at 09 47 56@2x

</details>

<details>

<summary>
Docs preview (local selected)
</summary>

CleanShot 2025-09-25 at 09 48 03@2x

</details>


<details>

<summary>
Studio staging preview
</summary>

CleanShot 2025-09-25 at 09 48 22@2x

</details>


<details>

<summary>
Studio self-hosted preview
</summary>

CleanShot 2025-09-25 at 09 51 54@2x

</details>

<details>
<summary>
Studio self-hosted localhost
</summary>
CleanShot 2025-09-25 at 09 54 48@2x
</details>

<details>
<summary>
Studio localhost
</summary>
CleanShot 2025-09-25 at 09 58 21@2x
</details>

I think the only environment where MCP URL is incorrect is the preview deployment of "studio-self-hosted", but that appears to be the case for other connect URLs too.

packages/ui-patterns/src/McpUrlBuilder/constants.tsx
```diff
@@ -51,6 +51,10 @@ export const FEATURE_GROUPS: McpFeatureGroup[] = [
   },
 ]
 
+export const FEATURE_GROUPS_NON_PLATFORM = FEATURE_GROUPS_PLATFORM.filter((group) =>
+  ['docs', 'database'].includes(group.id)
```
mattrossman: Maybe could wire this up to the same list of constants as the query schema in the server implementation:

https://github.com/supabase/supabase/blob/90976b798848ccc5bcb9c9b67a302f51609882ad/apps/studio/pages/api/mcp/index.ts#L10

gregnr: This might be over engineered, but it would be neat to auto detect feature groups based on tools available, regardless of platform (i.e. create a temp MCP client, list tools, deduce groups). Would probably require extra metadata sent back to map tools to groups, so not sure if it's worth the effort.

gregnr:
A few things:
- `project_ref` isn't supported on local/self-hosted today (since there's only the single project), so we should omit this query param there
- For docs, I'm not positive that self hosted (Docker compose) setups will use the same URL `http://127.0.0.1:54321` as local CLI. Maybe let's check our self hosted docs to see if there's a common default for this? Otherwise maybe we just call this platform "CLI" for now to keep things simple?

mattrossman:
- `project_ref` is now omitted for local/self-hosted
-  "CLI / Self-hosted" option is renamed to "CLI" for simplicity

Based on ["Self-Hosting with Docker"](https://supabase.com/docs/guides/self-hosting/docker), [docker-compose.yml](https://github.com/supabase/supabase/blob/e2960f38ed39907366b8b8dc393e5ae48158f70a/docker/docker-compose.yml), and [kong.yml](https://github.com/supabase/supabase/blob/e2960f38ed39907366b8b8dc393e5ae48158f70a/docker/volumes/api/kong.yml), I believe the current default for accessing MCP with self-hosted is `http://localhost:8000/api/mcp`, though this may change somwhat after ["[Self hosted] Create kong entry for MCP"](https://linear.app/supabase/issue/DEVWF-704/self-hosted-create-kong-entry-for-mcp)

apps/docs/features/ui/McpConfigPanel.tsx
```diff
@@ -108,26 +108,111 @@ function ProjectSelector({
...
+  className?: string
+  selectedPlatform: 'hosted' | 'local'
+  onPlatformSelect?: (platform: 'hosted' | 'local') => void
+}) {
+  const [open, setOpen] = useState(false)
+
+  const platforms = [
+    { value: 'hosted', label: 'Hosted' },
+    { value: 'local', label: 'CLI' },
+  ]
```
gregnr: If we mark this `as const`, we won't need the ` as 'hosted' | 'local'` casting below.

```suggestion
  const platforms = [
    { value: 'hosted', label: 'Hosted' },
    { value: 'local', label: 'CLI' },
  ] as const
```

apps/docs/features/ui/McpConfigPanel.tsx
```diff
@@ -108,26 +108,111 @@ function ProjectSelector({
...
+          </CommandList_Shadcn_>
+        </Command_Shadcn_>
+      </PopoverContent_Shadcn_>
+    </Popover_Shadcn_>
+  )
+}
+
 export function McpConfigPanel() {
   const [selectedProject, setSelectedProject] = useState<{ ref: string; name: string } | null>(null)
+  const [selectedPlatform, setSelectedPlatform] = useState<'hosted' | 'local'>('hosted')
```
gregnr, line 189: This is used in a lot of places, worth abstracting to a type?

```ts
type PlatformType = 'hosted' | 'local'
```

packages/ui-patterns/src/McpUrlBuilder/McpConfigPanel.tsx
```diff
@@ -18,26 +18,37 @@ export interface McpConfigPanelProps {
   onClientSelect?: (client: McpClient) => void
   theme?: 'light' | 'dark'
   className?: string
+  isPlatform: boolean // For docs this is controlled by state, for studio by environment variable
+  apiUrl: string | null
```
gregnr: Consider making this optional instead of `null`?

```suggestion
  apiUrl?: string
```

packages/ui-patterns/src/McpUrlBuilder/constants.tsx
gregnr: ^ follow up to my last comment - not a priority, let's keep it simple for now, duplication ok.

packages/ui-patterns/src/McpUrlBuilder/constants.tsx
```diff
@@ -197,4 +201,5 @@ export const MCP_CLIENTS: McpClient[] = [
   },
 ]
 
-export const MCP_URL = process.env.NEXT_PUBLIC_MCP_URL ?? 'http://localhost:8080/mcp'
+export const DEFAULT_MCP_URL_PLATFORM = 'http://localhost:8080/mcp'
+export const DEFAULT_MCP_URL_NON_PLATFORM = 'http://localhost:54321/mcp'
```
gregnr, line 180: Did we confirm that these can't be derived from `apiUrl` or an env var?

mattrossman: Normally it is derived from `apiUrl` and the `NEXT_PUBLIC_MCP_URL` env var:
https://github.com/supabase/supabase/blob/400f845895e74e50c1558c98f281190d0ad48067/packages/ui-patterns/src/McpUrlBuilder/utils/getMcpUrl.ts#L57-L68

These defaults are fallbacks for cases such as:
- Local dev when `NEXT_PUBLIC_MCP_URL` isn't set
- The `docs` app when you select CLI platform

packages/ui-patterns/src/McpUrlBuilder/McpConfigPanel.tsx
mattrossman: Updated https://github.com/supabase/supabase/pull/38981/commits/f252efe7fdcd7dabec467b5dfd30e63e43387d17

apps/docs/features/ui/McpConfigPanel.tsx
mattrossman: Updated in https://github.com/supabase/supabase/pull/38981/commits/a4b502ce269c82aeee6f561c5a3541ce409c1266 to a const assertion, move its definition outside the component, use `satisfies` to enforce shape, and infer a sharable `PlatformType` from its definition.

apps/docs/features/ui/McpConfigPanel.tsx
mattrossman: Updated as part of https://github.com/supabase/supabase/pull/38981#discussion_r2383160185

charislam (approved):
Tested on all preview sites & locally

packages/ui-patterns/src/McpUrlBuilder/constants.tsx
```diff
@@ -147,6 +151,31 @@ export const MCP_CLIENTS: McpClient[] = [
...
+      }
+    },
+    alternateInstructions: () => (
+      <p className="text-xs text-foreground-light">
+        Alternatively, configure through the UI by going to{' '}
+        <span className="font-medium">Settings</span> &gt;{' '}
+        <span className="font-medium">Integrations</span> and entering the URL above.
+      </p>
+    ),
+  },
```
gregnr: This might have snuck back in here

gregnr: Ah ignore this - I see this uses the older JSON approach vs Connectors

mattrossman: You're right that was intended to be removed. Last I checked the JSON approach didn't really work with Claude Desktop either as remote MCP support is removed there in favor of connectors and the `remote-mcp` shim gave me auth trouble. https://github.com/supabase/supabase/pull/36867#issuecomment-3305414573

mattrossman: Re-removed in https://github.com/supabase/supabase/pull/38981/commits/a11f98a76cafe8d08531ed00aff70bd786acfcdc

mattrossman:
Conflicts resolved, diff looks back to normal ✅ (will rebase to keep stacked PRs updated in the future to avoid messy PR)
Only other change is the read-only toggle is now hidden on self-hosted as that feature won't be ready in time ([Slack](https://supabase.slack.com/archives/C051L8U2EJF/p1759331398206479?thread_ts=1759266317.250659&cid=C051L8U2EJF))

PR #38988, opened 2025-09-24
feat: studio hide tabs button

Hallidayo:

YES

Supabase Studio - Table Editor/SQL Editor

This is a feature request from the discussions - #38958

## What is the new behavior?

A button has been added which will show and hide the tabs on both the Table and SQL editor:

https://github.com/user-attachments/assets/ac369ef3-6da9-4bbf-806a-eff9c2d6c187

## Summary by CodeRabbit

* **New Features**
  * Added the ability to toggle editor tab visibility using a new button in the product menu bar.
  * Icon button displays an eye or eye-off icon to indicate the current tab state.
  * Tabs can be hidden or shown with smooth animations for improved workspace flexibility.


charislam:
Thanks for the PR @Hallidayo! I think we may need to think more about this feature and whether it's worth building, because it's going to be complicated: we'll need to have two different forms of state handling between the "no tabs" and "tabs" preference. Otherwise, if we just hide the tabs display but continue to have tabs under the hood, we could get into weird states like tables being marked as open without being visually open 🤔

PR #38998, opened 2025-09-25
Add callouts on suggestions for certain support issues

joshenlim:
## Context

On the topic of enabling users to self-debug issues better, opting to add these callouts in the support form when either of these issue categories are selected. Also aims to improve visibility into these sections of the dashboard as well

image

image

image


alaister (approved):
LGTM!

- Tested on preview

PR #39004, opened 2025-09-25
Chore/contextual errors for table editor

joshenlim:
## Context

On the topic of enabling users to self debug issues, opting to provide contextual errors in the table editor especially when applying filters.

A common issue users run into is when providing a filter value that doesn't match the column data type, understandably with the current error UI it looks like it's something wrong with the dashboard. Opting to show actionable feedback to users when running into errors - we can incrementally expand this (We've already got one for foreign tables).

## Before

image

## Changes involved

- Add error UI for "invalid input syntax" errors
  - Added a "remove filters" CTA as well to allow users to quickly reset back to a working state
  image
- Add error UI for "could not identify ordering operator" errors
  - Added a "remove sorts" CTA as well to allow users to quickly reset back to a working state.
  image
- Also opting to set `retry` to false when fetching table rows
  - I reckon we don't need to retry the network request if the request is failing, and just immediately show the error to users. Otherwise atm it looks like the dashboard is really slow as it retries 3 times before finally showing the error
- Opt to not hide JSON columns from sort dropdown
  - We currently are, which can be confusing - opting to show a tooltip instead
    image

### DIFF
--- a/apps/studio/data/table-rows/table-rows-query.ts
+++ b/apps/studio/data/table-rows/table-rows-query.ts
@@ -236,9 +236,7 @@ export async function getTableRows(
 
     return { rows }
   } catch (error) {
-    throw new Error(
-      `Error fetching table rows: ${error instanceof Error ? error.message : 'Unknown error'}`
-    )
+    throw new Error(error instanceof Error ? error.message : 'Unknown error')
   }
 }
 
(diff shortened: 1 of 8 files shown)


apps/studio/data/table-rows/table-rows-query.ts
```diff
@@ -236,9 +236,7 @@ export async function getTableRows(
 
     return { rows }
   } catch (error) {
-    throw new Error(
-      `Error fetching table rows: ${error instanceof Error ? error.message : 'Unknown error'}`
-    )
+    throw new Error(error instanceof Error ? error.message : 'Unknown error')
```
joshenlim, line 239: The prefix here should be handled on the component that's consuming the error

alaister (approved):
LGTM!

- Tested on preview

PR #39011, opened 2025-09-25
fix(studio): human readable formatting for total consumed times

kemaldotearth:

YES

Fixes how our Time Consumed seconds look when the numbers are quite large. Previously they were not human readable, now they should be. Please @jordienr or someone double check I'm converting the right value 🙏 

| Before | After |
|--------|--------|
| Screenshot 2025-09-25 at 13 11 52 | Screenshot 2025-09-25 at 13 12 44 |


apps/studio/components/interfaces/QueryPerformance/QueryPerformance.utils.ts
```diff
@@ -0,0 +1,23 @@
+import dayjs from 'dayjs'
+
+export const formatDuration = (milliseconds: number) => {
```
jordienr, line 6: good candidate for a few tests

jordienr (approved):

apps/studio/components/interfaces/QueryPerformance/QueryPerformance.utils.ts
kemaldotearth: Any example of how we test right now, happy to do as part of this @jordienr

jordienr (approved):

PR #39014, opened 2025-09-25
Fix BucketRow long name hides more button

joshenlim:
## Context

As per PR title

## Before
image

## After
image


jordienr (approved):
:shipit:

PR #39018, opened 2025-09-25
feat: `pnpm i` will only run postbuild scripts for approved dependencies

ivasilov:


charislam (approved):
briefly smoke tested all apps

PR #39022, opened 2025-09-25
feat: self-hosted MCP development tools

mattrossman:
Follow-up from https://github.com/supabase/supabase/pull/38797

Implements `DevelopmentOperations` for the self-hosted remote MCP endpoint.

Refactors common logic to be shared with MCP from `/api/platform/projects/[ref]/settings` into
- `lib/api/self-hosted/settings.ts`
- `lib/api/self-hosted/typescript.ts`

To test run `pnpm dev:studio-local` and configure `http://localhost:8082/api/mcp` as a remote MCP server in e.g. Cursor

### Notes:
- I tweaked the error case from `/typecript` to align with [recent changes](https://github.com/supabase/supabase/pull/38965) to `/query`
  - Compare [here](https://github.com/supabase/supabase/blob/fdd99f1528f14952b6da35bbb0ec995bee0326f0/apps/studio/pages/api/v1/projects/%5Bref%5D/types/typescript.ts#L28-L37) and [here](https://github.com/supabase/supabase/blob/fdd99f1528f14952b6da35bbb0ec995bee0326f0/apps/studio/pages/api/platform/pg-meta/%5Bref%5D/query/index.ts#L27-L36)
- I didn't see existing test cases for these endpoints so I assume no tests need adding/updating

Resolves AI-151


apps/studio/lib/api/self-hosted/generate-types.ts
gregnr, line 1: nit: perhaps this could be called `generate.ts` or `generate-types.ts`?

apps/studio/pages/api/v1/projects/[ref]/types/typescript.ts
```diff
@@ -21,35 +21,18 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+  const { data, error } = await generateTypescriptTypes({ headers })
 
-  if (response.error) {
-    const { code, message } = response.error
-    return res.status(code).json({ message })
+  if (error) {
+    if (error instanceof PgMetaDatabaseError) {
+      const { statusCode, message, formattedError } = error
+      return res.status(statusCode).json({ message, formattedError })
+    }
```
Comment (gregnr) [inline apps/studio/pages/api/v1/projects/[ref]/types/typescript.ts]: I don't think `error` will ever be an instance of `PgMetaDatabaseError` since this is an error we manually create in `lib/api/self-hosted/query.ts`. Since you are using the `fetchGet` wrapper, you can probably check for `instanceof ResponseError` instead, then pull out the `message` and `code` as we did previously.

apps/studio/pages/api/v1/projects/[ref]/types/typescript.ts
```diff
@@ -21,35 +21,18 @@ async function handler(req: NextApiRequest, res: NextApiResponse) {
...
+    const { message } = error
```
Comment (charislam) [inline apps/studio/pages/api/v1/projects/[ref]/types/typescript.ts]: Am I missing something or is it impossible for this to be returned as a `PgMetaDatabaseError`? The only functions that check an error to possibly wrap it as a `PgMetaDatabaseError` are `executeQuery` and the `handlePost` from `/platform/pg-meta/[ref]/query`, but this is using `fetchGet`

charislam (approved):
Tested both MCP operations from Claude Code. Approving since only nit is an unused error handling branch

We should probably run the self-hosted e2e tests on this but I have no idea how to approve them, let me ask 😅

apps/studio/pages/api/v1/projects/[ref]/types/typescript.ts
Comment (mattrossman) [reply apps/studio/pages/api/v1/projects/[ref]/types/typescript.ts]: Good catch. Rereading [fetchGet](https://github.com/supabase/supabase/blob/5b38e124a42f5c1cce133b4a49ba694d1e3333fb/apps/studio/data/fetchers.ts#L248C1-L271C2), it looks like this helper performs error wrapping already, I didn't notice cause without a type argument the response collapses to `any`. I don't understand how the old error handling in `/typescript` worked since it looks for an error instance on `response.error` but IIUC the entire `response` would itself be a `ResponseError`.

I've updated this logic in https://github.com/supabase/supabase/pull/39022/commits/71c2218598df3b5c5e777e6b8af1400613f09083 to return the response from `fetchGet` (with type argument filled) so it preserves the result / error types and updated the error checks accordingly.

apps/studio/pages/api/v1/projects/[ref]/types/typescript.ts
Comment (mattrossman) [reply apps/studio/pages/api/v1/projects/[ref]/types/typescript.ts]: You're right! I just updated this after a similar comment from Greg https://github.com/supabase/supabase/pull/39022#discussion_r2380333851

apps/studio/lib/api/self-hosted/generate-types.ts
mattrossman: Yeah it's a little confusing with `types.ts` right next to it. Updated to `generate-types.ts` in https://github.com/supabase/supabase/pull/39022/commits/064ead2dac37a8585edc78874b6b04a3837824cb

PR #39026, opened 2025-09-26
Remove useProjectsQuery from support form

joshenlim:
## Context

This removes all usage of `useProjectsQuery` from the support form to ensure that organizations with large numbers of projects do not get stuck on the Support page (since that query will fetch _all_ projects belonging to the user)

## Changes involved

- `SupportFormV2` to use `getProjectDetails` call to validate the existence of the project
  - This only runs once as a initial prefill of the form based on the URL params 
- Support success state to use `useProjectDetails` instead to get project name from the provided `ref` via prop
- Within `new` at the bottom "Having trouble submitting the form" - updated "include your project ID" tooltip
  - This was previously also listing all projects under the user, but am opting to change the UX here to avoid fetching all projects
  - Added the project ID as a field description under "Which project is affected"
  - Clicking on "include your project ID" will scroll to that field and (hopefully clearly) inform the user where to copy the ID from
  - They can subsequently use the project dropdown selector to find the project and its ID (since that component handles infinite loading + searching of projects)
  - Am thinking this flow should work as the user would've tried to submit a ticket before going to the troubleshooting section if anything failed (and would have already selected the project that they wanted to submit a ticket for). So immediately directing them to where to get that project's ref should be alright (instead of listing _all_ the projects and their ref in a tooltip)
  - Lemme know wdyt of this UX though, happy to revise

    https://github.com/user-attachments/assets/28b04b57-c91d-48e8-8286-09d294e39253

## To test

- [ ] Verify that URL params for `projectRef` still works (project ref in URL param should be prefilled in the form)
- [ ] (UX) Verify that the UX flow for troubleshooting section doesn't feel awkward

### DIFF
--- a/apps/studio/components/interfaces/Support/Success.tsx
+++ b/apps/studio/components/interfaces/Support/Success.tsx
@@ -2,6 +2,7 @@ import { Check, ExternalLink, Mail, Search } from 'lucide-react'
 import Link from 'next/link'
 import { useState } from 'react'
 
+import { useProjectDetailQuery } from 'data/projects/project-detail-query'
 import { useProfile } from 'lib/profile'
 import { Button, Input, Separator } from 'ui'
 import { CATEGORY_OPTIONS } from './Support.constants'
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Support/Success.tsx
```diff
@@ -2,25 +2,24 @@ import { Check, ExternalLink, Mail, Search } from 'lucide-react'
...
-  projects = [],
-}: SuccessProps) => {
+export const Success = ({ sentCategory = '', selectedProject = 'no-project' }: SuccessProps) => {
   const { profile } = useProfile()
   const respondToEmail = profile?.primary_email ?? 'your email'
 
-  const project = projects.find((p) => p.ref === selectedProject)
+  const { data: project } = useProjectDetailQuery(
+    { ref: selectedProject },
+    { enabled: selectedProject !== 'no-project' }
```
alaister, line 21: I don't love the magic string `'no-project'` if I'm honest. I feel like we'll forget to check for it somewhere in future and accidentally try and use it as a real project ref. wdyt?

alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/interfaces/Support/Success.tsx
joshenlim: good point on this - it should just be a falsey value if there's no project selected

just quick a did check - i think its possible to change `no-project` to just a `null` value actually

i'll add a comment on this refactor to address separately 🙂

PR #39027, opened 2025-09-26
Fix FormItemLayout mobile responsivity when layout is flex row reverse

joshenlim:
Using organization settings as the demo here

## Before
image

## After
image


kemaldotearth (approved):
LGTM @joshenlim 🔥 I would say the inputs could be full width on smaller screens, but no deal break :)

joshenlim:
@kemaldotearth oh yup defo! noticed that the second input field is actually longer than its supposed to be - pushed up a fix 🙂 🙏 

image

PR #39028, opened 2025-09-26
Fix default order by columsn to exclude json columns

joshenlim:
## Context

Our current `getDefaultOrderByColumns` picks the first column in the table to sort on if the table has no primary key

However, JSON based columns cannot be sorted on and hence if the table's first column is of JSON type, the user will run into the following error with no way of remediating:

image

## Changes involved

- Omit json based columns in `getDefaultOrderByColumns`

## To test

- [ ] Create a table with no PK, and just one JSON type column
- [ ] Verify that you can open that table in the table editor


charislam (approved):
tested as described

PR #39032, opened 2025-09-26
feat(replication): Add dialog to clarify replication requirements and limitations

iambriccardo:
This PR adds a dialog which shows the user the current limitations of replication and some things to note. The dialog is meant to be shown only when creating a pipeline.

### Screenshots

image

### DIFF
--- a/apps/studio/components/interfaces/Database/Replication/ReplicationDisclaimerDialog.tsx
+++ b/apps/studio/components/interfaces/Database/Replication/ReplicationDisclaimerDialog.tsx
@@ -1,0 +1,102 @@
+import {
+  Accordion_Shadcn_,
+  AccordionContent_Shadcn_,
+  AccordionItem_Shadcn_,
+  AccordionTrigger_Shadcn_,
+  Button,
+  Dialog,
+  DialogContent,
+  DialogFooter,
+  DialogHeader,
+  DialogSection,
+  DialogSectionSeparator,
+  DialogTitle,
+} from 'ui'
+import { Admonition } from 'ui-patterns'
+
+interface ReplicationDisclaimerDialogProps {
+  open: boolean
+  isLoading: boolean
+  onOpenChange: (value: boolean) => void
+  onConfirm: () => void
+}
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Database/Replication/ReplicationDisclaimerDialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+            Creating this replication pipeline will immediately start syncing data from your
+            publication into the destination. Make sure you understand requirements and limitations
+            of the system before proceeding.
+          </p>
+
+          <Admonition type="warning" className="space-y-1 px-4 py-3">
+            <p className="font-medium text-foreground">Primary keys are required for all tables</p>
+            <p className="text-foreground-light">
+              Every table included in the publication must expose a primary key for replication to
+              work correctly.
```
joshenlim: might be better to be clearer here RE how the behaviour will be if this condition isn't met

also wondering if its possible to catch this requirement on the client side and we block starting a replication to mitigate guess work if there's issues. ideal scenario i suppose is that we're able to highlight upfront which tables in the selected publication are missing primary keys

iambriccardo: We investigated checking this upfront but the problem is that a user can change these settings in their database after the fact, so we would do a check which doesn't guarantee down the line that it is upheld.

joshenlim: so this disclaimer dialog then just focuses on the limitations than the requirements

iambriccardo: Yes, exactly.

joshenlim (approved):
Checked changes on staging preview - although will be good for you to also take a look once more before merging @iambriccardo 🙂 🙏

iambriccardo:
Looks good!

PR #39041, opened 2025-09-26
feat: read-only mode for self-hosted MCP

mattrossman:
- `readOnly` option added to self-hosted `executeQuery` (default false)
- `crypto-js` and `@types/crypto-js` deps added matching [infra encryption behavior](https://github.com/supabase/infrastructure/blob/64162cead1a418ebd3e4b2438c6e5aec15ef9f9a/worker/src/lib/helpers.ts#L29-L31)
- `read_only` parameter supported in MCP endpoint
- `POSTGRES_PASSWORD`, `POSTGRES_HOST`, `POSTGRES_DB`, `POSTGRES_PORT`, `PG_META_CRYPTO_KEY` added to `turbo.json` so values configured from env files produce the correct connection string
- Updated `generateLocalEnv.js` to generate defaults for these for `dev:studio-local`

Many of the `POSTGRES_*` vars have defaults in the self-hosting example env file:
https://github.com/supabase/supabase/blob/3bda4dbac7d3b86c492a38a437e0efdc70f8192c/docker/.env.example#L20-L22

`POSTGRES_PASSWORD` is required to be set by self-hosters:
https://supabase.com/docs/guides/self-hosting/docker#update-secrets

Otherwise for CLI it defaults to "postgres": 
https://github.com/supabase/cli/blob/05a2f39026049ab060874e05736c65a2d6c535aa/pkg/config/config.go#L337

Similar to Management API, this read-only implementation sends an encrypted connection string to pg-meta with a read only postgres user. ~~The key is hardcoded to `"SAMPLE_KEY"` based on [this pg-meta fallback](https://github.com/supabase/postgres-meta/blob/dcb8e9ba9a70f05aab4865144743233ed285db59/src/server/constants.ts#L9), see [slack thread](https://supabase.slack.com/archives/C07E5GFAHTM/p1758215024704829) for further context.~~ (edit: this is now configurable with `PG_META_CRYPTO_KEY`)

Test by running `pnpm dev:studio-local` with remote MCP url `http://localhost:8082/api/mcp?read_only=true`, it should fail to perform writes via `execute_sql`.

Resolves [AI-153](https://linear.app/supabase/issue/AI-153/support-read-only-mode)


apps/studio/lib/api/self-hosted/constants.ts
```diff
@@ -0,0 +1,9 @@
+// Constants specific to self-hosted environments
+
+export const ENCRYPTION_KEY = 'SAMPLE_KEY' // See postgres-meta `CRYPTO_KEY` fallback
```
charislam: I'm not super familiar with the hosted platform, should this be synced with `postgres-meta`'s implementation of `getSecret('CRYPTO_KEY')`?

charislam (commented):
Tested on local. Looks good pending a question about the `ENCRYPTION_KEY`, will follow up in Slack!

apps/studio/lib/api/self-hosted/constants.ts
mattrossman: Based on [this message](https://supabase.slack.com/archives/C07E5GFAHTM/p1759279805285839?thread_ts=1758215024.704829&cid=C07E5GFAHTM), yeah we're better off making it configurable.

I've made a change in https://github.com/supabase/supabase/pull/39041/commits/7d307983d0533a8164345e0fbad0792aaf41a2d0 to have it controlled in both services via a `PG_META_CRYPTO_KEY` env var.

I also drafted a CLI change to populate the variable in both services, though that might not be necessary as I figure CLI would use "SAMPLE_KEY" either way. https://github.com/supabase/cli/pull/4236

mattrossman: Got one informal approval on the change [here](https://supabase.slack.com/archives/C07E5GFAHTM/p1759355234903599?thread_ts=1758215024.704829&cid=C07E5GFAHTM)

sweatybridge (approved):

supa-mdx-lint/Rule003Spelling.toml
```diff
@@ -246,6 +246,7 @@ allow_list = [
     "PostQUEL",
     "PostgREST",
     "Postgres",
+    "postgres-meta",
```
mattrossman, line 249: This is the spelling used already in the docs

https://github.com/supabase/supabase/blob/2b5caf9977ee87e90f76a2674ae48c720c0c8b8c/apps/docs/content/guides/self-hosting/docker.mdx?plain=1#L324

charislam (approved):

PR #39067, opened 2025-09-29
Fix query builder wrap table schema name in double quotes

joshenlim:
## Context

Edge case - but the Table Editor will error out when fetching rows for tables with names that are reserved keywords (e.g `system_user`) Fix is to wrap the table names in double quotes in the query from the query builder.

## To test

- [ ] Create a table with a name like "system_user" in the table editor
- [ ] Table should not error out when loading rows after opening it


charislam (approved):
tested in preview

PR #39072, opened 2025-09-29
Context gathering prominence

SaxonF:
Puts a little more emphasis on context gathering tools


charislam (approved):

charislam (approved):
Approved to unblock though it seems to be maybe a little overeager about this:

Image

PR #39092, opened 2025-09-30
Fix auth pages error state consistency and missing loading states

joshenlim:
## Context

Just addressing something that I noticed while developing locally - error states on the auth pages are not using the AlertError components + missing loading states

## Before

image

## After

image

## To test

- [ ] Just verify that the auth pages load as expected and the loading states do not feel awkward


charislam (approved):
Tested loading states on preview

PR #39093, opened 2025-09-30
chore: feature flag for new Storage UI

dnywh:

Resolves DEPR-89. Adds a feature flag for incoming Storage UI (and a bare bones first version of this for testing).

There is only one version of Storage UI in Studio. 

## What is the new behavior?

There is now another “v2” version, behind a feature flag, being incrementally built-upon to support two new bucket types described [here](https://linear.app/supabase/project/40139f777444).

To make this feature flag more tangible, I’ve also included the following v2 work within this PR:
- Storage sidebar now contains a list of bucket _types_ rather than a list of buckets

Note that the main content area of Storage, under this feature flag, only shows placeholder UI. In other words: **users cannot access or create buckets** in the GUI. This is so we can make this large change to Studio incrementally across small PRs. It’s why the feature flag is limited to just myself and whoever tests this work.

| Before | After |
| --- | --- |
| CleanShot 2025-09-30 at 12 42 57@2x | CleanShot 2025-09-30 at 12 40 49@2x |

_Duplicated description in above screenshot is purposeful and will be addressed in DEPR-90, started in [dnywh/feat/storage-buckets-inline](https://github.com/supabase/supabase/pull/39154)_

## Additional context

I’d love to know if this is the best approach to gating features to (a handful of) folks. Keep in mind that the [this work](https://linear.app/supabase/project/40139f777444) will extend to foreign tables, i.e. outside of /storage.

There are some known imperfections:
- Users not on the feature flag who go to `/project/:ref/storage/media` or another new `bucketType` will see that above new state momentarily, before the `false` is returned. I felt this was a better approach because most folks won’t experience these new routes and it avoids layout shift/flash in the inverse case.

## To test

- [ ] Then test that the newer, in-progress, Storage UI is shown **only** when that feature flag is enabled via either of the following:
  - [ ] Add your local/staging account email address to the `storageAnalyticsVector` flag in ConfigCat
  - [ ] Just hardcode any instance of `isStorageV2` to `true` instead of referencing `useFlag('storageAnalyticsVector')`


apps/studio/components/layouts/StorageLayout/StorageLayout.tsx
```diff
@@ -10,12 +13,14 @@ export interface StorageLayoutProps {
 }
 
 const StorageLayout = ({ title, children }: StorageLayoutProps) => {
+  const isStorageV2 = useFlag('storageAnalyticsVector')
+
   return (
     <ProjectLayout
-      stickySidebarBottom
+      stickySidebarBottom={isStorageV2 ? false : true}
```
dnywh, line 19: This entire prop will be removed in DEPR-96

apps/studio/components/interfaces/Storage/Storage.constants.ts
```diff
@@ -1,3 +1,6 @@
+import { DOCS_URL } from 'lib/constants'
+
+// Original storage constants
```
dnywh, line 3: Soon-to-be outdated storage constants like `STORAGE_SORT_BY_ORDER` will be removed in DEPR-95

apps/studio/components/interfaces/Storage/StorageMenuV2.tsx
```diff
@@ -0,0 +1,32 @@
+import Link from 'next/link'
+import { useParams } from 'common'
+import { Menu } from 'ui'
+import { BUCKET_TYPES, BUCKET_TYPE_KEYS, DEFAULT_BUCKET_TYPE } from './Storage.constants'
+
+const StorageMenuV2 = () => {
```
dnywh: Note this `ComponentNameV2` pattern will be used extensively as part of this work so the existing Storage UI can continue to function.

apps/studio/components/layouts/StorageLayout/StorageLayout.tsx
```diff
@@ -10,12 +13,14 @@ export interface StorageLayoutProps {
 }
 
 const StorageLayout = ({ title, children }: StorageLayoutProps) => {
+  const isStorageV2 = useFlag('storageAnalyticsVector')
```
dnywh: See ConfigCat for corresponding feature flag on both environments. Feel free to add your testing account email address here.

apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  useEffect(() => {
+    if (!config) {
+      router.replace(`/project/${ref}/storage`)
+    }
+  }, [config, ref, router])
+
+  return (
+    <div>
+      {/* [Danny] Purposefully duplicated directly below StorageLayout‘s config.description for now. Will be placed in a conditional empty state in next PR. TODO: consider reusing FormHeader for non-empty state.*/}
+      <p className="text-foreground-light mb-4">{config.description}</p>
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx]: As mentioned above: this `div` forms the placeholder for both the empty state and 1+ bucket state header. It will be replaced in the next PR as part of DEPR-90

joshenlim:
Left a comment in linear but just also mentioning here for ease of reference

Since the changes here are looking to replace an existing UI, will be good for us to follow this [playbook](https://github.com/supabase/playbooks/blob/main/playbooks/frontend/incremental-rollout-strategy.md) 🙂 

So the main things that I'd change in this PR is to 
- Have the UI behind a feature preview (instead of a feature flag)
- Have the feature preview behind the feature flag instead

apps/studio/components/interfaces/Storage/Storage.constants.ts
```diff
@@ -56,3 +59,31 @@ export const CONTEXT_MENU_KEYS = {
   STORAGE_ITEM: 'STORAGE_ITEM',
   STORAGE_FOLDER: 'STORAGE_FOLDER',
 }
+
+// New bucket types configuration
+export interface BucketTypeConfig {
```
joshenlim, line 64: i'd opt to shift this to Storage.types.ts for consistency

apps/studio/components/interfaces/Storage/Storage.constants.ts
```diff
@@ -56,3 +59,31 @@ export const CONTEXT_MENU_KEYS = {
...
+    displayName: 'Analytics',
+    description: 'Purpose-built storage for analytical workloads.',
+    docsUrl: `${DOCS_URL}/guides/storage/analytics/introduction`,
+  },
+  vectors: {
+    displayName: 'Vectors',
+    description: 'Purpose-built storage for vector data.',
+    docsUrl: `${DOCS_URL}/guides/storage/vectors`,
+  },
+} as const satisfies Record<string, BucketTypeConfig>
```
joshenlim: do we need this actually? BUCKET_TYPES would already be inferred correctly even without this declaration
image

apps/studio/components/interfaces/Storage/StorageMenuV2.tsx
joshenlim: yup sounds good! easier to clean up at the end too RE identifying the old files

apps/studio/components/interfaces/Storage/StorageMenuV2.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <Menu type="pills" className="mt-6 flex flex-grow flex-col">
+      <div className="mx-3">
+        <Menu.Group title={<span className="uppercase font-mono">Bucket Types</span>} />
+
+        {BUCKET_TYPE_KEYS.map((bucketTypeKey, idx: number) => {
+          const isSelected = selectedBucketType === bucketTypeKey
+          const config = BUCKET_TYPES[bucketTypeKey]
+
+          return (
+            <Link key={`${idx}_${bucketTypeKey}`} href={`/project/${ref}/storage/${bucketTypeKey}`}>
```
joshenlim: don't need to use idx here - just the bucketTypeKey is good enough

apps/studio/components/interfaces/Storage/StorageMenuV2.tsx
joshenlim: nit: we're following a code style whereby we export the component directly, rather than declaring a export default at the bottom

so this can be `export const StorageMenuV2 = () => { ... }`

apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <DocsButton href={config.docsUrl} />
+    </div>
+  )
+}
+
+BucketTypePage.getLayout = (page) => {
+  // We need to get the bucketType from the router since it's not available in page.props
+  const BucketTypeLayout = () => {
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx]: could live in layouts/StorageLayout

joshenlim:
Pushed up some tiny refactors which should resolve all the comments i left 🙏

joshenlim (approved):
Verified on staging preview that feature preview and feature flag works!

apps/studio/components/interfaces/App/FeaturePreview/FeaturePreview.constants.tsx
```diff
@@ -1,9 +1,16 @@
...
+    key: LOCAL_STORAGE_KEYS.UI_PREVIEW_NEW_STORAGE_UI,
+    name: 'New Storage interface',
+    discussionsUrl: undefined,
+    isNew: true,
+    isPlatformOnly: false,
+  },
   {
     key: LOCAL_STORAGE_KEYS.UI_PREVIEW_UNIFIED_LOGS,
-    name: 'New Logs Interface',
+    name: 'New Logs interface',
```
dnywh, line 13: Nit. Just making language consistent with elsewhere

apps/studio/components/interfaces/App/FeaturePreview/NewStorageUIPreview.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const NewStorageUIPreview = () => {
+  return (
+    <div className="text-sm text-foreground-light">
+      <p className=" mb-4">
+        Experience our enhanced Storage interface with support for analytics and vector bucket
+        types.
+      </p>
+      <Image
+        alt="new-storage-preview"
+        src={`${BASE_PATH}/img/previews/new-storage-preview.png`}
```
dnywh, line 23: Figma mockup so not 1:1 with reality but will replace over time.

apps/studio/components/interfaces/App/FeaturePreview/UnifiedLogsPreview.tsx
```diff
@@ -26,8 +26,8 @@ export const UnifiedLogsPreview = () => {
         <p className="text-sm">Enabling this preview will:</p>
         <ul className="list-disc pl-6 text-sm text-foreground-light space-y-1">
           <li>
-            Replace the current logs interface on the{' '}
-            <InlineLink href={`/project/${ref}/logs`}>logs page</InlineLink> with a unified view
+            Replace the current Logs interface on the{' '}
+            <InlineLink href={`/project/${ref}/logs`}>Logs page</InlineLink> with a unified view
```
dnywh, line 30: Just some nit language/copywriting consistency

apps/studio/components/interfaces/Storage/Storage.constants.ts
```diff
@@ -56,3 +59,26 @@ export const CONTEXT_MENU_KEYS = {
   STORAGE_ITEM: 'STORAGE_ITEM',
   STORAGE_FOLDER: 'STORAGE_FOLDER',
 }
+
+// New bucket types configuration
+
+export const BUCKET_TYPES = {
+  files: {
+    displayName: 'Files',
```
dnywh, line 67: Renamed `Media` to `Files` for clarity

apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx
Comment (dnywh) [reply apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx]: Thanks. Will make a note for the next PR as I’m making related changes [there](https://github.com/supabase/supabase/compare/dnywh/feat/storage-buckets-inline?diff=split&w)

joshenlim (approved):
Double approve ✅ ✅ 😄

PR #39095, opened 2025-09-30
Give password protection visibility within auth protection

SaxonF:
image

We've had feedback that finding "Prevent use of leaked passwords" has been challenging so this just gives it more visibility within the protection page. The setting just links off to email provider as the two are tied together.


joshenlim (approved):
Verified on staging preview!

PR #39102, opened 2025-09-30
Add callout for leaked password that its on the pro plan and above

joshenlim:
## Context

Realised that "Leaked password protection" is a feature that's only available on the pro plan and above as highlighted in our docs [here](https://supabase.com/docs/guides/auth/password-security#password-strength-and-leaked-password-protection), but its not an information that's obvious from the dashboard which resulted in some users tripping up here

## Changes involved

- Adds a callout to the field's description if it's a paid feature + organization is on the free plan, link goes to the organization billing page with the subscription panel open
  image
  - This wouldn't show up for an organization that's on Pro plan and above
  image

- Also added a tooltip inline to the field label if there's a corresponding link to the docs site
  image

### DIFF
--- a/apps/studio/components/interfaces/Auth/AuthProvidersForm/FormField.tsx
+++ b/apps/studio/components/interfaces/Auth/AuthProvidersForm/FormField.tsx
@@ -7,6 +7,7 @@ import { Markdown } from 'components/interfaces/Markdown'
 import { DatePicker } from 'components/ui/DatePicker'
 import { BASE_PATH } from 'lib/constants'
 import { Button, Input, InputNumber, Listbox, Toggle } from 'ui'
+import { InfoTooltip } from 'ui-patterns/info-tooltip'
 import type { Enum } from './AuthProvidersForm.types'
 
 interface FormFieldProps {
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
joshenlim, line 1: changes here are unrelated but nit stuff i came across while working on this

apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
```diff
@@ -613,7 +610,7 @@ const NewOrgForm = ({ onPaymentMethodReset, setupIntent, onPlanSelected }: NewOr
                   </div>
                   <div className="text-foreground-light text-xs">
                     {orgProjects.length <= 2 ? (
-                      <span>{orgProjects.join('and ')}</span>
+                      <span>{orgProjects.join(' and ')}</span>
```
joshenlim, line 613: missing space:
image

apps/studio/components/interfaces/Organization/NewOrg/NewOrgForm.tsx
```diff
@@ -571,9 +572,9 @@ const NewOrgForm = ({ onPaymentMethodReset, setupIntent, onPlanSelected }: NewOr
 
       <ConfirmationModal
         size="large"
-        loading={false}
+        loading={newOrgLoading}
```
joshenlim, line 575: was missing loading state

fadymak (approved):

apps/studio/components/interfaces/Auth/AuthProvidersForm/FormField.tsx
```diff
@@ -223,17 +224,18 @@ const FormField = ({
           id={name}
           name={name}
           disabled={disabled}
```
fadymak, line 226: nit: should we add the `cursor-not-allowed` class when the field is disabled?

charislam (approved):
Checked on preview that callout only shows up on free project

apps/studio/components/interfaces/Auth/AuthProvidersForm/FormField.tsx
joshenlim: that should be alright - the toggle is not clickable while it's disabled and we currently don't have a pattern of using `cursor-not-allowed` for disabled buttons / interfaces

PR #39113, opened 2025-09-30
feat(studio): add auth overview feature flag

kemaldotearth:

YES

This brings our new Auth Overview page behind a feature flag.


charislam (approved):
Approved with one suggestion:

- Even with the feature flag toggled off, it's possible for someone to navigate directly to the page and see it in its 🚧 state. Perhaps you can add a check on the page itself that redirects away when the feature is disabled?

kemaldotearth:
> Approved with one suggestion:
> 
> * Even with the feature flag toggled off, it's possible for someone to navigate directly to the page and see it in its 🚧 state. Perhaps you can add a check on the page itself that redirects away when the feature is disabled?

Thanks @charislam that should be there now! Feel free to test and let me know if anything else can improve there.

PR #39116, opened 2025-09-30
Flatten table editor grid context menu copy action

joshenlim:
## Context

Opting to flatten the context menu of the table editor grid into a single level menu since copy cell value is a rather common action and this 2nd level UI slows down the UX

## Before

image

## After

image

## To test

- [ ] Verify that copy cell and copy row are still working in the table editor


ivasilov (approved):

saltcod (approved):
Works great! thanks!

PR #39137, opened 2025-10-01
chore: disable billing support form options

alaister:
Fixes FE-1899


joshenlim (approved):
Verified changes locally, tested on staging preview that things are status quo + tested sending a ticket

PR #39138, opened 2025-10-01
fix(enabled-features): auth providers settings link

alaister:
When auth providers wasn't being shown, the link would go to a page that doesn't exist.


joshenlim (approved):
Verified changes locally, tested on preview that things are status quo

PR #39139, opened 2025-10-01
fix: edge function url in list

alaister:
Updated to match the URL logic in `apps/studio/components/interfaces/Functions/EdgeFunctionDetails/EdgeFunctionDetails.tsx`


joshenlim (approved):

PR #39141, opened 2025-10-01
chore: disable org security tab

alaister:


joshenlim (approved):
Verified changes locally, tested on preview that things are status quo

PR #39142, opened 2025-10-01
chore: hide edge function examples

alaister:


joshenlim (approved):
Verified changes locally, tested on preview that things are status quo in edge function details page -> invocation examples

PR #39143, opened 2025-10-01
chore: disable client libraries in support form

alaister:


joshenlim (approved):
Verified changes locally, tested on preview that things are status quo by submitting a support ticket with api and client libraries category selected

PR #39145, opened 2025-10-01
Chore/ai realtime

SaxonF:
image

We are advising people to move towards broadcasting messages from your database vs postgres changes when it comes to creating realtime experiences connected to your database. This PR is a stepping stone towards that direction:
- It replaces the enable realtime toggle within the table editor with a link to Triggers which is a key primitive when creating realtime experiences using broadcast. This is behind a feature flag `triggersInsteadOfRealtime`
- It adds a more informative empty state for triggers highlighting realtime as a use case
- It adds a more informative empty state for realtime pointing users to triggers as a first step

**To test:**
- enable the feature flag
- notice the triggers button next to policies in the table editor header
- go to triggers and notice empty state
- go to realtime and notice empty state

**Measuring**
We want to make sure realtime adoption is not hurt too badly with this change (I suspect a decrease will happen) so will be monitoring realtime adoption rates as well gradually roll out feature flag to more people.


apps/studio/components/interfaces/Realtime/Inspector/AnimatedCursors.tsx
charislam, line 1: So pretty! 😻

charislam (approved):
Nice!

PR #39149, opened 2025-10-01
re-add chart config

SaxonF:
image

re-adds chartconfig to rendering tool executeSql so that results can be rendered in a chart vs a table.

To test, ask Assistant to query data in a way that would benefit from being rendered in a chart.


charislam (approved):
tested in preview

PR #39150, opened 2025-10-01
Add flag to disable wrappers UI

joshenlim:
## Context

Adds a flag in enabled features to disable everything related to wrappers in the UI. Also considers empty state if flag is off and landing on a wrappers URL

## To test

- [ ] Verify that flag works locally (And let me know if I overlooked any sections)
- [ ] Verify that things are status quo on staging preview


charislam (approved):
tested as described

PR #39152, opened 2025-10-01
Add flag for metadata IP logs explorer template

joshenlim:
## Context

Adds a flag in enabled features to toggle visibility of the "Metadata IP" template in logs explorer

## To test

- [ ] Verify behaviour locally when flag is off
- [ ] Verify on staging preview that things are status quo


charislam (approved):
Tested as described! Also pushed a commit to hide from templates page as well (though in practice maybe the templates page will be disabled if the metadata IP template is disabled?)

PR #39153, opened 2025-10-01
Add flag for restarting project

joshenlim:
## Context

Adds a flag in enabled features to show the option to restart project. If off, the option will be to restart the database instead (which runs the fast database reboot)

Flag on:
image

Flag off:
image

## To test

- [ ] Verify changes locally when flag is off


charislam (approved):
tested as described

PR #39154, opened 2025-10-01
feat(studio): inline storage buckets

dnywh:

Feature: inline storage (file) buckets for (internal) users on the `UI_PREVIEW_NEW_STORAGE_UI` feature preview. Resolves DEPR-90.

https://github.com/supabase/supabase/pull/39093 introduced the new sidebar pattern, showing bucket _types_ on the sidebar. But this removed the ability to enter buckets or see much about them.

## What is the new behavior?

This follow-up adds:
- Full-container empty states for each bucket type
- Tabbed `navigationItems` for _file_ bucket type
- Lists all file buckets within a `Table` component
  - Search file buckets

| Files (Empty) | Files (1+ Buckets) | Files → Settings |
| --- | --- | --- |
| Storage  Supabase | Storage  Supabase | Storage  Supabase |

## Additional context

Still to come:
- View each (file) bucket _contents_
- Add analytics bucket contents and creation
- Add vector bucket contents and creation

## To test

- [ ] Then test that the newer Storage UI mentioned above in _new behavior_ is shown and functional **only** when that feature flag is enabled via either of the following:
  - [ ] Add your local/staging account email address to the `storageAnalyticsVector` flag in ConfigCat
  - [ ] Just hardcode any instance of `isStorageV2` to `true` instead of referencing `useFlag('storageAnalyticsVector')`


apps/studio/components/layouts/StorageLayout/BucketLayout.tsx
```diff
@@ -8,22 +9,66 @@ import { PageLayout } from '../PageLayout/PageLayout'
...
-          subtitle={config?.description || 'Manage your storage buckets and files.'}
-          secondaryActions={secondaryActions}
-        >
-          <ScaffoldContainer>{children}</ScaffoldContainer>
-        </PageLayout>
+        {isEmpty ? (
+          // For empty state, render directly without PageLayout/ScaffoldContainer
+          children
+        ) : (
+          // For normal state, use PageLayout with ScaffoldContainer
```
dnywh: This is required because the empty state design requires no `PageLayout` or `ScaffoldContainer` to be present. In other words: if 0 buckets exist, a different layout should be shown.

Happy to implement this another way. I spent a bit of time on this, and the `isEmpty` check here is the best I could come up with!

apps/studio/components/layouts/StorageLayout/BucketLayout.tsx
```diff
@@ -8,22 +9,66 @@ import { PageLayout } from '../PageLayout/PageLayout'
...
+    {
+      label: 'Policies',
+      href: `/project/${ref}/storage/files/policies`,
+    },
+  ]
+
+  // Use provided navigationItems or default based on bucket type
+  // Show files navigation items if we're on files bucket type OR on a files sub-route
+  const finalNavigationItems =
+    navigationItems || (bucketTypeKey === 'files' || isOnFilesSubRoute ? filesNavigationItems : [])
```
dnywh: This is so we can programmatically pass a value to `PageLayout`’s `navigationItems` prop. The _files_ storage type requires this tabbed pattern (for /policies and /settings that are only for that type of bucket. The the other two bucket types (_vector_ and _analytics_) are much simpler and don’t require `navigationItems` (yet).

Happy to implement this a different way if you have suggestions. The `isOnFilesSubRoute` irks me a bit but was the only way I could have it stick around when on /policies and /settings.

apps/studio/components/interfaces/Storage/CreateBucketModalSpecialized.tsx
dnywh, line 1: A reusable dialog for creating vector and analytics bucket types. These are very different to file bucket types yet very similar to each other. They both:

- Require a bucket name (with differing validation e.g. length, underscores or dashes)
- Give notice of a wrapper that will be installed on their behalf (if not already installed)

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
```diff
@@ -95,12 +95,26 @@ const formId = 'create-storage-bucket-form'
 
 export type CreateBucketForm = z.infer<typeof FormSchema>
 
-export const CreateBucketModal = () => {
+interface CreateBucketModalProps {
+  hideAnalyticsOption?: boolean
```
dnywh: Backwards compatibility. Since this dialog in Storage V2 is just for file buckets, we don’t need the below `"type"` part of the form.

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
```diff
@@ -225,26 +239,29 @@ export const CreateBucketModal = () => {
...
+                text: !canCreateBuckets
+                  ? 'You need additional permissions to create buckets'
+                  : undefined,
+              },
+            }}
+          >
+            New bucket
+          </ButtonTooltip>
+        </DialogTrigger>
+      )}
```
dnywh: This component will now be called from a few places, in a different style to this. Maintaining for backwards compatibility behind a `externalVisible` prop.

apps/studio/pages/project/[ref]/storage/files/policies.tsx
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/files/policies.tsx:1]: We need to make a ‘copy’ of the /policies and /settings pages so that:
- They appear at the correct path for Storage V2 users, e.g.: /storage/files/policies
- They remain at the existing path (with a different header) for normal users, e.g. storage/policies

apps/studio/components/interfaces/Storage/AnalyticsBuckets.tsx
```diff
@@ -0,0 +1,6 @@
+import { EmptyBucketState } from './EmptyBucketState'
+
+export const AnalyticsBuckets = () => {
+  // Placeholder component - will be implemented in a later PR
+  return <EmptyBucketState bucketType="analytics" onCreateBucket={() => {}} />
```
dnywh: Like in `VectorBuckets` this file will just render the empty state for now. Bucket creation split out into future PR to keep this one lightweight.

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
dnywh: Storage V2 will have a different dialog for analytics (and vector) buckets.

apps/studio/components/interfaces/Storage/EmptyBucketState.tsx
dnywh, line 1: Reused component across all three bucket types when <1 bucket exists.

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
dnywh, line 1: This interface controls everything to do with the standard bucket type for general file storage. This PR (re)introduces the following:
- True empty state
- List of buckets
- Bucket creation dialog

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -0,0 +1,182 @@
...
+    switch (activeTab) {
+      case 'settings':
+        return (
+          <div className="space-y-12">
+            <StorageSettings />
+            <S3Connection />
+          </div>
+        )
+      case 'policies':
+        return <StoragePolicies />
```
dnywh: This file will show the various settings and policies components if the user has selected the corresponding tab (`navigationItems`)

apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx
```diff
@@ -15,6 +21,40 @@ const BucketTypePage: NextPageWithLayout = () => {
...
+      default:
+        const defaultFilesBuckets = buckets.filter((bucket: Bucket) => bucket.type === 'STANDARD')
+        return {
+          component: <FilesBuckets />,
+          isEmpty: defaultFilesBuckets.length === 0 && !bucketsLoading,
+        }
+    }
+  }
+
+  const bucketTypeResult = getBucketTypeResult()
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx]: Required to determine empty state or not. Happy to be corrected!

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
joshenlim: i reckon we actually don't need a prop here - could just listen to the feature preview
since only the new UI will need this to be true

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  return (
+    <>
+      {renderTabContent()}
+      <CreateBucketModal
+        hideAnalyticsOption={true}
+        visible={showCreateBucketModal}
+        onOpenChange={setShowCreateBucketModal}
+      />
```
joshenlim: hmm we'll be rendering 2 of this components if the tab is `bucket` - does the create bucket modal need to be present across all tabs?

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -0,0 +1,182 @@
...
+          </div>
+        )
+      case 'policies':
+        return <StoragePolicies />
+      case 'buckets':
+      default:
+        if (filesBuckets.length === 0) {
+          return (
+            <>
+              <EmptyBucketState bucketType="files" onCreateBucket={handleCreateBucket} />
```
joshenlim: feedback (not sure if already planned down the road)

Just wondering cause the settings and policy pages won't be available without a bucket created, which will be hiding a couple of things that aren't bucket related like Settings -> S3 Connection, and Policies -> Policies on the storage schema directly

apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/storage/[bucketType]/index.tsx:1]: hmm im actually wondering if we need this route actually

given that we're introducing specific pages like files/policies, and files/settings
all the other bucket types could have their own specific route as well - so no need for dynamic param (preference for things to be more declarative when possible)

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
```diff
@@ -95,12 +95,26 @@ const formId = 'create-storage-bucket-form'
 
 export type CreateBucketForm = z.infer<typeof FormSchema>
 
-export const CreateBucketModal = () => {
+interface CreateBucketModalProps {
+  hideAnalyticsOption?: boolean
+  visible?: boolean
+  onOpenChange?: (open: boolean) => void
```
joshenlim: hmm im on the fence with introducing an external state like this - since there will (?) always be a button trigger to open this modal? e.g CreateBucketModal could be rendered in the EmptyState instead

apps/studio/components/interfaces/Storage/EmptyBucketState.tsx
```diff
@@ -0,0 +1,48 @@
...
+          <p className="text-foreground-light text-sm">
+            {config.description}{' '}
+            <Link
+              href={config.docsUrl}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="text-foreground-lighter underline hover:text-foreground transition-colors"
+            >
+              Learn more
+            </Link>{' '}
```
joshenlim: thoughts on a dedicated docs button instead of an inline URL?

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -0,0 +1,182 @@
...
+  if (isLoading) {
+    return (
+      <div className="flex items-center justify-center h-32">
+        <div className="animate-spin rounded-full h-8 w-8 border-b-2 border-foreground"></div>
+      </div>
+    )
+  }
+
+  // Render content based on active tab
+  const renderTabContent = () => {
```
joshenlim: this one's just personal preference but i'd prefer these conditionals to be within the final return statement of the component instead - keep all the JSX in one place rather than within a function

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -0,0 +1,182 @@
...
+    if (path.includes('/storage/files/settings')) return 'settings'
+    if (path.includes('/storage/files/policies')) return 'policies'
+    return 'buckets'
+  }
+
+  const activeTab = getActiveTab()
+
+  if (isLoading) {
+    return (
+      <div className="flex items-center justify-center h-32">
```
joshenlim, line 57: wondering if we should apply a background color here? the background color change feels quite jarring from the loading state to the empty state

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const getActiveTab = () => {
+    const path = router.asPath
+    if (path.includes('/storage/files/settings')) return 'settings'
+    if (path.includes('/storage/files/policies')) return 'policies'
+    return 'buckets'
+  }
+
+  const activeTab = getActiveTab()
+
+  if (isLoading) {
```
joshenlim: the early return statement here will unnecessarily block settings and policies i think (those pages don't need the buckets to load first for e.g)

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -0,0 +1,182 @@
...
+    return (
+      <div className="flex items-center justify-center h-32">
+        <div className="animate-spin rounded-full h-8 w-8 border-b-2 border-foreground"></div>
```
joshenlim: will this eventually be a loader that we'll want to use everywhere in the dashboard? (for consistency)

apps/studio/components/layouts/StorageLayout/BucketLayout.tsx
joshenlim: hmm okay gotcha on this - i'll make some changes to simplify things here that's related to the comment [here](https://github.com/supabase/supabase/pull/39154/files#r2396934688) since layout is specific to the files page

apps/studio/components/interfaces/Storage/EmptyBucketState.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export const EmptyBucketState = ({ bucketType, onCreateBucket }: EmptyBucketStateProps) => {
+  const config = BUCKET_TYPES[bucketType]
+
+  return (
+    <div className="flex flex-col justify-center h-full w-full bg-muted">
+      {/* Centered container with padding */}
+      {/* TODO: pt-[8rem] or similar and include illustration */}
+      <div className="w-full max-w-sm mx-auto bg-surface-100 rounded-lg px-8 pt-10 pb-8 flex flex-col gap-6 border">
```
joshenlim, line 19: wondering if we should remove the background color here as the transition is quite jarring

loading -> empty
empty -> buckets exist state

apps/studio/components/interfaces/Storage/EmptyBucketState.tsx
```diff
@@ -0,0 +1,48 @@
...
+interface EmptyBucketStateProps {
+  bucketType: keyof typeof BUCKET_TYPES
+  onCreateBucket: () => void
+}
+
+export const EmptyBucketState = ({ bucketType, onCreateBucket }: EmptyBucketStateProps) => {
+  const config = BUCKET_TYPES[bucketType]
+
+  return (
+    <div className="flex flex-col justify-center h-full w-full bg-muted">
```
joshenlim: wondering if we should remove the background color (bg-muted) here as the transition is quite jarring

loading -> empty
empty -> buckets exist state

apps/studio/pages/project/[ref]/storage/files/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+  useEffect(() => {
+    if (!isStorageV2) router.replace(`/project/${ref}/storage`)
+  }, [isStorageV2, ref, router])
+
+  if (isLoadingBuckets) {
+    return (
+      <div className="flex items-center justify-center h-full">
+        <div className="animate-spin rounded-full h-8 w-8 border-b-2 border-foreground" />
+      </div>
+    )
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/storage/files/index.tsx]: there's still an awkward loading state transition if you refresh on this page
is this a loading indicator that we'd want to standardize across the dashboard?

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
dnywh: Nope, just a placeholder one that I plan to swap out later for Loading component or similar. Good eyes!

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
dnywh: Noted, thanks for letting me know

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
dnywh: Discussed on Slack with @burmecia:

> ...those S3 connection config are providing S3-compatible access, so they’re relevant to all bucket types including analytic and vector bucket type ...It doesn’t matter how many buckets are created, all the created buckets are using the same S3 connection info. Even with no bucket created, those info can still be there. 

So I need to extract S3 Connection as a separate sidebar item outside of bucket types. Sorry for this change!

dnywh: Following up about policies on same thread in Slack.

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
dnywh: Hmm. Should just be the following instances:

- One `CreateBucketModal` trigger on empty state
- One `CreateBucketModal` trigger on bucket tab (1+ buckets)

The `CreateBucketModal` trigger is not available on the other tabs

apps/studio/pages/project/[ref]/storage/files/index.tsx
Comment (dnywh) [reply apps/studio/pages/project/[ref]/storage/files/index.tsx]: Good call! Let’s discuss on Slack

packages/ui/src/components/Input/Input.tsx
```diff
@@ -38,7 +38,7 @@ export interface Props
 }
 
 /**
- * @deprecated Use `import { Input_shadcn_ } from "ui"` instead or ./ui-patterns/data-inputs/input
+ * @deprecated Use `import { Input_Shadcn_ } from 'ui'` instead or `import { Input } from 'ui-patterns/DataInputs/Input'`
```
dnywh, line 41: Updated wrong import information on comment.

apps/studio/pages/project/[ref]/storage/s3.tsx
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/s3.tsx:1]: S3 settings is now global ([context](https://supabase.slack.com/archives/C0161K73J1J/p1759385963488829?thread_ts=1759303258.525489&cid=C0161K73J1J)) and therefore its own page in Storage.

apps/studio/pages/project/[ref]/storage/files/settings.tsx
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/files/settings.tsx:1]: Settings that just pertain to the Files storage type.

apps/studio/components/layouts/StorageLayout/StorageSpecializedLayout.tsx
dnywh, line 1: I made this shared layout instead of:
- StorageAnalyticsLayout
- StorageVectorsLayout

They are basically the same at this point.

apps/studio/components/layouts/StorageLayout/StorageFilesLayout.tsx
dnywh, line 1: I suppose we could create a generic layout like StorageTypeLayout that conditionally shows the  `navigationItems` for files. But, as discussed elsewhere, we are going for clarity over code minimalism.

apps/studio/components/interfaces/Storage/VectorsBuckets.tsx
dnywh, line 1: Will be fleshed out with vector storage contents in an upcoming PR.

apps/studio/components/interfaces/Storage/StorageMenuV2.tsx
```diff
@@ -1,29 +1,47 @@
...
 import { Menu } from 'ui'
 import { BUCKET_TYPES, BUCKET_TYPE_KEYS, DEFAULT_BUCKET_TYPE } from './Storage.constants'
 
 export const StorageMenuV2 = () => {
-  const { ref, bucketType } = useParams()
-  const selectedBucketType = bucketType || DEFAULT_BUCKET_TYPE
+  const router = useRouter()
+  const { ref } = useParams()
+
+  const page = router.pathname.split('/')[4] as undefined | 'files' | 'analytics' | 'vectors' | 's3'
```
dnywh: Now includes root S3 item. See other comment about it now being a shared setting across storage types ([context](https://supabase.slack.com/archives/C0161K73J1J/p1759385963488829?thread_ts=1759303258.525489&cid=C0161K73J1J)).

apps/studio/components/interfaces/Storage/Storage.constants.ts
```diff
@@ -65,20 +65,25 @@ export const CONTEXT_MENU_KEYS = {
...
+    bucketName: 'an analytics bucket',
     description: 'Purpose-built storage for analytical workloads.',
+    valueProp: 'Store large datasets for analytics and reporting.',
     docsUrl: `${DOCS_URL}/guides/storage/analytics/introduction`,
   },
   vectors: {
     displayName: 'Vectors',
+    bucketName: 'a vector bucket',
     description: 'Purpose-built storage for vector data.',
+    valueProp: 'Store, index, and query your vector embeddings at scale.',
```
dnywh, line 84: The empty state has slightly different copywriting (a value proposition) to complement the description. Both are shown on the same page.

apps/studio/components/interfaces/Storage/Storage.constants.ts
```diff
@@ -65,20 +65,25 @@ export const CONTEXT_MENU_KEYS = {
...
   analytics: {
     displayName: 'Analytics',
+    bucketName: 'an analytics bucket',
     description: 'Purpose-built storage for analytical workloads.',
+    valueProp: 'Store large datasets for analytics and reporting.',
     docsUrl: `${DOCS_URL}/guides/storage/analytics/introduction`,
   },
   vectors: {
     displayName: 'Vectors',
+    bucketName: 'a vector bucket',
```
dnywh: English is weird with articles(“an” vs “a“) based on subsequent word sound. Ditto for whether a word should be plural.

Clearer (I think) to write this explicitly as a constant for each than to have weird button copywriting like “Create files bucket”

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
dnywh: The table of buckets can now be searched, just like [database tables](https://supabase.com/dashboard/project/_/database/tables).

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -0,0 +1,172 @@
...
+        ? true
+        : bucket.name.toLowerCase().includes(filterString.toLowerCase())
+    )
+  return (
+    <>
+      {!isLoading &&
+      buckets.filter((bucket) => !('type' in bucket) || bucket.type === 'STANDARD').length === 0 ? (
+        <EmptyBucketState bucketType="files" />
+      ) : (
+        <div className="pt-12 flex flex-col gap-y-4">
```
dnywh: `pt-12` is visually to large of a gap but is necessary here because any smaller would be misaligned with other tabs.

Something to polish in post.

dnywh (commented):
Several updates centred around:
- Simpler empty states
- S3 settings in root sidebar

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
joshenlim: reckon we can use ScaffoldSection for consistency? (the padding is handled there)
e.g Auth -> Sign In/Providers has a similar layout like this (BasicAuthSettingsForm.tsx)

apps/studio/components/interfaces/Storage/Storage.constants.ts
joshenlim: yup that's alright!

apps/studio/components/interfaces/Storage/StorageMenuV2.tsx
```diff
@@ -1,29 +1,47 @@
...
+        </div>
 
-          return (
-            <Link key={bucketTypeKey} href={`/project/${ref}/storage/${bucketTypeKey}`}>
-              <Menu.Item rounded active={isSelected}>
-                <p className="truncate">{config.displayName}</p>
+        <div className="h-px w-full bg-border" />
+        <div className="mx-3">
+          <Menu.Group title={<span className="uppercase font-mono">Configuration</span>} />
+          {IS_PLATFORM && (
```
joshenlim: given that there's only one link here - IS_PLATFORM can probably wrap this whole section (the border + configuration group)

apps/studio/components/layouts/StorageLayout/StorageFilesLayout.tsx
joshenlim: tbh actually given that we're no longer doing a full page empty state - i'm happy to combine all the layouts into one, makes more sense now (opted to separate all previously as buckets page had a conditional rendering pattern than analytics + vectors didn't need

will combine this and StorageSpecializedLayout 🙂

joshenlim: reckon s3 page could use the same layout too

apps/studio/pages/project/[ref]/storage/files/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+import type { NextPageWithLayout } from 'types'
+
+const StorageFilesPage: NextPageWithLayout = () => {
+  const { ref } = useParams()
+  const router = useRouter()
+  const isStorageV2 = useIsNewStorageUIEnabled()
+
+  useEffect(() => {
+    if (!isStorageV2) router.replace(`/project/${ref}/storage`)
+  }, [isStorageV2, ref, router])
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/storage/files/index.tsx]: we can then shift all these re-routing logic to the single layout then 🙂

joshenlim (approved):
Verified on staging preview all looks good! Both existing UI and new UI 🙂 🙏

apps/studio/components/layouts/StorageLayout/StorageFilesLayout.tsx
dnywh: Sounds good!

apps/studio/components/interfaces/Storage/StorageMenuV2.tsx
dnywh: Ah yep!

apps/studio/pages/project/[ref]/storage/files/index.tsx
Comment (dnywh) [reply apps/studio/pages/project/[ref]/storage/files/index.tsx]: Love it. `StorageBucketsLayout` coming through with the goods

SaxonF (approved):
Tested old/new on staging

apps/studio/components/layouts/StorageLayout/StorageBucketsLayout.tsx
```diff
@@ -31,10 +39,16 @@ export const StorageBucketsLayout = ({ children }: PropsWithChildren) => {
...
+    if (!isStorageV2) router.replace(`/project/${ref}/storage/buckets`)
+  }, [isStorageV2, ref, router])
+
   return (
     <PageLayout
-      title={config?.displayName || 'Storage'}
-      subtitle={config?.description || 'Manage your storage buckets and files.'}
+      title={title || (config?.displayName ?? 'Storage')}
+      subtitle={
+        hideSubtitle ? config?.description || 'Manage your storage buckets and files.' : null
```
dnywh, line 50: FYI @joshenlim I’ve inverted this ternary in [the upcoming PR](https://github.com/supabase/supabase/pull/39342/files#diff-b6ba82aef3e975e7844954dc1181988785d1517ad63aeba4295c77025920e6beL49) because this hides the subtitle text. Let me know if that was intentional (I can see merit for leaving it off!)

PR #39156, opened 2025-10-01
Fix upgrading compute for provider

joshenlim:
## Context

Upgrading of compute wasn't working for when project is on a specific provider - this fixes that and also hides disk configuration related fields which aren't relevant for that provider

Also, within database settings -> disk management for a specific provider "Increase disk size" button will be hidden + "View detailed summary" will be behind the `reports:all` flag

## To test

Locally: Will need to manually update your local database infra config for that project to that provider (ping me if need more information)
  - [ ] Verify that you can change your compute size for project on that specific provider
  - [ ] Verify that there's no disk configuration related stuff for project on that specific provider
  - [ ] Verify that there's no "increase disk size" button in database settings -> disk management on that specific provider
  - [ ] Verify that "View detailed summary" button is hidden if `reports:all` flag is off

Staging Preview:
- [ ] Verify that things are status quo on staging preview


alaister (approved):
LGTM!

- Tested as instructed

PR #39160, opened 2025-10-01
FE-1862 Report Chart: Always show tooltip values

jordienr:
More info in the Linear issue → https://linear.app/supabase/issue/FE-1862/always-show-the-0-field

## BEFORE

CleanShot 2025-10-01 at 10 54 50

## AFTER

CleanShot 2025-10-01 at 10 54 10


kemaldotearth (approved):
I love it when things get simpler. LGTM 👍

PR #39161, opened 2025-10-01
Add flag for network restrictions section

joshenlim:
## Context

Adds a flag in enabled features to toggle the visibility of the Network Restrictions section in Database settings

## To test

- [ ] Verify locally that section is hidden when flag is off locally


charislam (approved):
tested as described

PR #39166, opened 2025-10-01
Chore/direct upgrade prompts to support based on flag

joshenlim:
## Context

Updates all CTAs for upgrading plan or toggling spend caps to direct to the support form instead of the organization billing page if `billing:all` flag is off.

Added a new ticket category "Plan Upgrades" for easier triaging which only shows up if `billing:all` is off
I've tested locally sending a support ticket with this category, API passes but I can't verify atm if our Support system received a ticket - need to testing on staging
image

Support form content should be prefilled so that users can conveniently hit submit ticket immediately
image
image

For plan upgrades, there's some areas of the dashboard where we don't specify what plan to upgrade to (e.g `BillingMetric.tsx` -> Organization/Usage page, Monthly SSO metric for example)
- In this case, expect the placeholder `<Specify which plan to upgrade to: Pro | Team | Enterprise>` to be there
- I've also added form validation to check for that placeholder
  image

## To test
All these needs to be tested on a Free plan organization
- [ ] Verify locally upgrade plan / toggle spend cap CTAs direct to support form when `billing:all` flag is off
  - Example for plan upgrade: Organization Audit Logs
  - Example for toggling spend cap: Realtime settings
- [ ] Verify that there's no changes on staging preview
  - We also need to test sending a "Plan Upgrade" category ticket on the preview and verify that it get's sent to our Support system correctly (I've temporarily made that option visible in this PR)
  - The only visible change actually is just the upgrade call out in Authentication -> Sessions, the current placement is a bit awkward so shifted it to be similar to how we do the call out in Storage Settings
    - Before:
      image
    - After:
      image


apps/studio/components/interfaces/Support/SupportFormV2.tsx
```diff
@@ -161,7 +180,13 @@ export const SupportFormV2 = ({
         option.value === SupportCategories.SALES_ENQUIRY
       ) {
         return isBillingEnabled
+      } else if (option.value === 'Plan_upgrade') {
+        // [Joshen] THIS IS JUST FOR TESTING OF SENDING A SUPPORT TICKET ON STAGING PREVIEW
+        // REVERT THIS VALUE ONCE TESTING IS COMPLETED BEFORE MERGING
+        // return !isBillingEnabled
+        return true
```
joshenlim: 🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨
[Joshen] Reminder to revert this value back after testing before merging
🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨

staaldraad (approved):

alaister (approved):
LGTM!

- Tested as instructed

PR #39171, opened 2025-10-01
Add flag for logs explorer showing metadata

joshenlim:
## Context

Adds a flag in enabled features to toggle the visibility of the `metadata` property from log results in the Logs Explorer

If the flag is off, there shouldn't be a `metadata` column shown in the results nor in the Log Detail side panel when selecting a particular log

## To test

- [x] Verify the above behaviour locally
- [x] Verify that staging preview has no changes


matlin (approved):
LGTM

matlin (approved):
raw_log_data.metadata update working well 👌

Just adjusted the conditional to skip property checks if logsMetadata is enabled

PR #39175, opened 2025-10-01
fix(studio): use session storage as default cache for tab state

matlin:

YES

Bug fix

Because local storage is used to store editor tab state, working across multiple _browser_ tabs can lead to confusing behavior of loading tab state from the previously used tab when clicking around the UI.

## What is the new behavior?

Now tab state is saved to both session storage (scoped to the individual browser tab) as well as local storage and then when reading defaults to session storage. For a new tab that doesn't have anything stored in it's session it will read from local storage which will reflect the most recent tab state.

## Additional context

**To test**: open table editor in multiple browser tabs of the same browser and ensure that when clicking between tabs or other table links that the state is consistent to the individual tab and doesn't affect other tabs.


charislam (approved):
tested as described

PR #39179, opened 2025-10-01
feat(pgmeta): filter by composite values

charislam:
Current behavior:

Only single columns can be used as filters: `where version = 2`

New behavior:

Composite values can be used as filters: `where (id, version) = (1, 2)`

Motivation:

In order to correct an existing bug when fetching table rows, we need to use cursor pagination on the primary key, but in order to handle composite primary keys, we need to be able to filter by composite values.

### DIFF
--- a/packages/pg-meta/src/query/Query.utils.ts
+++ b/packages/pg-meta/src/query/Query.utils.ts
@@ -166,6 +166,22 @@ function applyFilters(query: string, filters: Filter[]) {
   if (filters.length === 0) return query
   query += ` where ${filters
     .map((filter) => {
+      // Handle composite values
+      if (Array.isArray(filter.column)) {
+        switch (filter.operator) {
+          case 'is':
+          case '~~':
+          case '~~*':
+          case '!~~':
+          case '!~~*':
+            throw new Error(`Cannot use ${filter.operator} operator in a tuple filter`)
+          case 'in':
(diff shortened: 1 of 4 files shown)


packages/pg-meta/src/query/Query.utils.ts
```diff
@@ -166,6 +166,22 @@ function applyFilters(query: string, filters: Filter[]) {
   if (filters.length === 0) return query
   query += ` where ${filters
     .map((filter) => {
+      // Handle composite values
+      if (Array.isArray(filter.column)) {
+        switch (filter.operator) {
+          case 'is':
```
matlin: Is there a source for the operators that aren't supported with tuples? If so might be worth documenting with a comment there or switch more to a whitelist support where we accept the filters that we know to work and reject the rest.

Alternatively, could it make sense to just let Postgres reject bad filters?

charislam: good point, let's make it a whitelist

joshenlim (approved):
code changes LGTM! although not tested on preview though - not sure if the changes here affect anything in the dashboard?

charislam:
> not sure if the changes here affect anything in the dashboard?

shouldn't change anything since we're not using this new functionality yet, I just did a quick smoke test of the tables-related parts

PR #39181, opened 2025-10-01
fix(table editor): inconsistent serialization and serializing stale values for filter state

charislam:
We have some reports of filters getting corrupted as users switch or reload tabs.

Some fixes that might help the issue:
- Making sure we write the latest values to storage (we were writing stale values before)
- Using the same serialization/deserialization logic anywhere that needs to sync with URL state

There's probably more we could do to refactor out some common utils, just want to get a basic fix in this PR to see if it helps with a support ticket.

To test:
1. Use an `in` filter to test against a set of matching values on a table
2. Close and reload studio
3. Go back to the same table, the deserialization should fail and cause an error in prod; this should be fixed in this PR


alaister (approved):
LGTM!

- Tested on preview

charislam:
Closed because fixed independently in another PR

PR #39182, opened 2025-10-01
chore: add nimbus auto sync

alaister:


.github/workflows/trigger-nimbus-sync.yml
```diff
@@ -0,0 +1,18 @@
...
+  trigger-nimbus-sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger sync in supabase-nimbus
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.NIMBUS_SYNC_TOKEN }}
+          repository: supabase/supabase-nimbus
+          event-type: sync_from_upstream
+          client-payload: '{"ref": "${{ github.ref }}", "sha": "${{ github.sha }}"}'
```
charislam: Will this be used for something downstream eventually? (I don't see it being read currently in supabase-nimbus.)

I'm kind of dubious about that `github.ref` depending on its final use, since it's untrusted input, but I guess as it stands only people with write access can trigger this job 🤔

charislam (approved):

.github/workflows/trigger-nimbus-sync.yml
alaister: Yep good point! I'll remove

PR #39187, opened 2025-10-02
fix(studio): remove password managers from inputs

dnywh:

Bug fix

1Password shows an annoying widget and tooltip in the _Bucket name_ field when it has nothing to do with what a user might have in a 1Password vault.

## What is the new behavior?

This 1Password widget no longer shows.

| Before | After |
| --- | --- |

## Additional context

As explained [here](https://flaviocopes.com/how-to-disable-1password-in-an-input-field/), 1Password adds this based on the `name="name"` value. I added the appropriate `ignore` attributes for 1Password and similar password managers. I also did the same for edge function secrets.


saltcod (approved):
tested on staging, both work!

PR #39188, opened 2025-10-02
docs(design-system): NavMenu component

dnywh:

Design system docs

We don’t mention the NavMenu component which is used extensively through Studio as a third level of hierarchy.

## What is the new behavior?

- Adds NavMenu component documentation
- Adds general Navigation documentation

| Navigation | NavMenu |
| --- | --- |
| Navigation | NavMenu |

## Additional context

The general Navigation documentation is housed under a new ‘UI Patterns’ section, much like [GitHub Primer](https://primer.style/product/ui-patterns/navigation/). I find their directory structure much clearer and something to work towards, but also understand we currently use `ui-patterns` to mean a certain type of component in code.

I’m proposing (in later PRs and over time) we update our design system to follow a similar structure:

1. Getting Started
  - Figma
  - Tailwind
  - Etc
2. Primitives
  - Color
  - Size
  - Typography
  - Etc
3. UI Patterns
  - Loading states
  - Navigation
  - Saving
  - Etc
4. Components
  - Admonition
  - NavMenu
  - TreeView
  - Etc

Like this:

NavMenu

That means—again, over time—breaking away from a section that maps 1:1 with Shadcn’s docs and the more abstract UI Patterns components. We can and should still stick to the registry.


SaxonF (approved):
Looks good. NavMenu primarily (only?) used as sub navigation within PageLayout which might be worth mentioning.

dnywh:
> Looks good. NavMenu primarily (only?) used as sub navigation within PageLayout which might be worth mentioning.

Good shout @SaxonF! Added.

PR #38904, opened 2025-09-22
feat: better Postgres 17 upgrade instructions for deprecated extensions

dnywh:
## What kind of change does this PR introduce?

Improves our flow for [upgrading a project’s Postgres version](https://supabase.com/dashboard/project/_/settings/infrastructure) in the case that extension(s) need to be first disabled. 

We get a tonne of support tickets about this flow as it currently stands. Both because it’s confusing and because most users are affected by this deprecated extensions case.

## What is the current behavior?

We show the extensions to be disabled in a bulleted list. Each item in the list is a hyperlink, although the styling doesn’t make this clear.

CleanShot 2025-09-22 at 12 21 58@2x

## What is the new behavior?

We show the extensions to be disabled in a clearer, actionable, list with a _Manage [extension]_ button. The link to the docs now goes to a more pertinent section, with clearer copywriting around `pgjwt` and its history in Supabase projects (enabled in all projects up until recently, so most likely safe to remove).

CleanShot 2025-09-22 at 12 23 22@2x

## Additional context

You can mock this `hasUnsupportedExtensions` state by temporarily hardcoding `unsupported_extensions` in [useProjectUpgradeEligibilityQuery](https://github.com/supabase/supabase/blob/c440226633d4f36ff9f44bc9f248148f25380368/apps/studio/data/config/project-upgrade-eligibility-query.ts#L52), like so:

```tsx
export const useProjectUpgradeEligibilityQuery = <TData = ProjectUpgradeEligibilityData>(
  { projectRef }: ProjectUpgradeEligibilityVariables,
  {
    enabled = true,
    ...options
  }: UseQueryOptions<ProjectUpgradeEligibilityData, ProjectUpgradeEligibilityError, TData> = {}
) => {
  const { data: project } = useProjectByRefQuery(projectRef)
  return useQuery<ProjectUpgradeEligibilityData, ProjectUpgradeEligibilityError, TData>(
    configKeys.upgradeEligibility(projectRef),
    ({ signal }) => getProjectUpgradeEligibility({ projectRef }, signal),
    {
      enabled:
        enabled &&
        project !== undefined &&
        project.status === PROJECT_STATUS.ACTIVE_HEALTHY &&
        typeof projectRef !== 'undefined' &&
        IS_PLATFORM,
      // TEMPORARY MOCK FOR TESTING - REMOVE AFTER TESTING
      select: (data) => {
        if (data) {
          return {
            ...data,
            eligible: false,
            unsupported_extensions: [
              'pgjwt',
[...]

PR #39189, opened 2025-10-02
chore(studio): improve Postgres upgrade experience

dnywh:

Makes the Postgres upgrade experience clearer. 
- Fixes DEPR-80
- Fixes DEPR-82

- The `ObjectsToBeDroppedWarning` just links to the docs. It doesn’t make it clear that the user needs to remove the affected function(s).
- The _Service Versions_ section doesn’t make it clear that this is where the user goes to upgrade Postgres
  - This is exacerbated by the fact that the API call to determine upgrade eligibility can take ~10 seconds

## What is the new behavior?

Specifically for the `ObjectsToBeDroppedWarning`:

| Before | After |
| --- | --- |
| Infrastructure  Supabase | Infrastructure  Supabase |

To address upgrade clarity: the description for the _Service Versions_ section has been amended.

## Additional context

This is a follow up from https://github.com/supabase/supabase/pull/38904.

I would prefer that each line in `ObjectsToBeDroppedWarning` links out directly to its corresponding function. For example:

```tsx
<Button size="tiny" type="default" asChild>
  <Link href={`/project/${ref}/database/functions?search=${obj}`}>Manage</Link>
</Button>
```

But the API only returns `objects_to_be_dropped` as an array of strings like:
- function public.test_noop_plv8 is using obsolete language plv8

That requires backend changes to provide the name of the function and name of the extension, not just the string.

### DIFF
--- a/apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
+++ b/apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
@@ -33,7 +33,7 @@ export const ObjectsToBeDroppedWarning = ({
           <p className="mb-1">
             The following objects are not supported and must be removed before upgrading.{' '}
             <InlineLink
-              className="text-foreground-lighter hover:text-foreground transition-colorsrs"
+              className="text-foreground-lighter hover:text-foreground transition-colors"
               href={`${DOCS_URL}/guides/platform/upgrading#extensions`}
               target="_blank"
               rel="noopener noreferrer"
--- a/apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
+++ b/apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
@@ -24,36 +24,37 @@ export const ObjectsToBeDroppedWarning = ({
 }: {
   objectsToBeDropped: string[]
 }) => {
+  const { ref } = useParams()
   return (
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
```diff
@@ -24,36 +24,37 @@ export const ObjectsToBeDroppedWarning = ({
...
+    <Alert_Shadcn_ title="A newer version of Postgres is available for your project">
+      <AlertTitle_Shadcn_>A newer version of Postgres is available</AlertTitle_Shadcn_>
       <AlertDescription_Shadcn_ className="flex flex-col gap-3">
-        <div>
-          <p className="mb-1">The following objects have to be removed before upgrading:</p>
+        <>
+          <p className="mb-1">
+            The following objects are not supported and must be removed before upgrading.{' '}
+            <InlineLink
+              className="text-foreground-lighter hover:text-foreground transition-colorsrs"
```
charislam:
```suggestion
              className="text-foreground-lighter hover:text-foreground transition-colors"
```

charislam (approved):

apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
dnywh: Good catch, thanks!

PR #39224, opened 2025-10-03
Rate assistant responses

SaxonF:
image

This adds ThumbsUp and ThumbsDown message actions which are tracked alongside a categorisation of the conversation and an optional reason if downvoting.

Starting with just a category enum due to privacy concerns of summarising the entire conversation into a more freeform string.


charislam (approved):
Nice! Tested the flow and it works.

Just one minor nit: the thumbs up/down buttons show even as the Markdown is streaming in, can we make them appear only after the streaming is complete?

packages/common/telemetry-constants.ts
```diff
@@ -1189,6 +1189,35 @@ export interface AiAssistantInSupportFormClickedEvent {
...
 }
 
+/**
+ * User rated an AI assistant message with thumbs up or thumbs down.
+ *
+ * @group Events
+ * @source studio
+ */
+export interface AssistantMessageRatedEvent {
+  action: 'assistant_message_rated'
```
pamelachia: rename action to assistant_message_rating_submitted to keep it in line with [naming conventions](https://www.notion.so/supabase/Event-tracking-in-Posthog-1435004b775f8027a624dcd9325bd966?source=copy_link#1435004b775f80b0bcbfe5fbcbb0070a)

pamelachia (changes requested):

pamelachia (approved):

PR #39230, opened 2025-10-03
Add callout for migration

joshenlim:
Note: Changes here will only be visible for projects with > 10M objects in their storage (so most of the project's wouldn't be seeing this) This will be based on whether listV2 is false and their upstream target is main from the storage config

## Context

Changes are to make storage upgrade self-serve for certain projects so that their able to control when to do the upgrade since it involves running a migration on the database schema.

UI is controlled by the feature flag `storageMIgrationCallout` which toggles the visibility of the related UI

## If upgrade is available

- Tooltip will show for Storage settings in the side nav
  image
- Callout will show in storage settings
  image
- Clicking the CTA will show this confirmation dialog
  image
- In which upgrading will then flip the callout to this state (This callout will disappear once the migration is completed)
  image
- Callout will change to a warning if it's not completed and it's one month to the deadline (currently set as 1st Jan 2026 for e.g)
  image
  - Using TimestampInfo for convenience here as well
    image

## Next steps (Can be done separately - non blocking since its behind feature flag)

- [ ] Dashboard Storage Explorer needs to use new List V2 endpoint once project has access to that
  - Currently blocked on API endpoint availability

### DIFF
--- a/apps/studio/components/interfaces/DiskManagement/fields/DiskSizeField.tsx
+++ b/apps/studio/components/interfaces/DiskManagement/fields/DiskSizeField.tsx
@@ -88,7 +88,7 @@ export function DiskSizeField({
   const mainDiskUsed = Math.round(((diskUtil?.metrics.fs_used_bytes ?? 0) / GB) * 100) / 100
 
   return (
-    <div className="grid @xl:grid-cols-12 gap-5">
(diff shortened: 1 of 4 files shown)


charislam (approved):
Didn't know how to test E2E, but manually overrode the network responses to check the UI

PR #39277, opened 2025-10-06
chore: project daily stats caching / deprecated props

kevcodez:
Remove deprecated props that have no effect on behaviour and fix the caching (based on date instead of timestamp)

### DIFF
--- a/apps/studio/components/interfaces/Reports/ReportBlock/ChartBlock.tsx
+++ b/apps/studio/components/interfaces/Reports/ReportBlock/ChartBlock.tsx
@@ -23,7 +23,6 @@ import { Dashboards } from 'types'
 import { WarningIcon } from 'ui'
 import { METRIC_THRESHOLDS } from './ReportBlock.constants'
 import { ReportBlockContainer } from './ReportBlockContainer'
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/Reports/ReportBlock/ChartBlock.tsx
```diff
@@ -23,6 +23,7 @@ import { Dashboards } from 'types'
 import { WarningIcon } from 'ui'
 import { METRIC_THRESHOLDS } from './ReportBlock.constants'
 import { ReportBlockContainer } from './ReportBlockContainer'
+import { startOfDay } from 'date-fns'
```
joshenlim: unused import?

joshenlim (approved):
Verified that daily-stats.controller doesn't have `interval` param + all good on removing `databaseIdentifier`

Also verified that daily-stats.controller only cares about YYYY-MM-DD (ignores time) so the formatting in ChartBlock makes sense

Minimally checked custom reports page on staging preview that all seems okay (Based on the changes only that page should be affected) - ComposedChartHandler being used in built in reports are all getting data from infra-monitoring

PR #39278, opened 2025-10-06
chore: cache time usage / do not preload invoice breakdown

kevcodez:
- Increase cache time of usage to 60m
- Ensure org is fully loaded before rendering invoice breakdown (this currently leads to starting to load this endpoint and then cancelling it when directly landing on the org billing page)


kanishkdudeja (approved):

joshenlim (approved):
Code changes LGTM! Verified BillingBreakdown changes on staging preview

PR #39281, opened 2025-10-06
Switch Assistant to Streamdown

SaxonF:
Switches Assistant from react-markdown to https://github.com/vercel/streamdown for improved performance and component streaming. I believe this resolves re-rendering issues we were facing previously and means we no longer need to hold off rendering SQL queries until streaming is complete. 

To test, ask Assistant to:
- Write a query without executing it
- Execute a query
- Write an edge function
- Write js code


joshenlim:
Very tiny nit but there seems to be 2 list decorations
image

joshenlim:
Also tiny nit: Noticed that we no longer block the output with "Writing SQL..."
But we still have that behaviour for edge functions - "Writing Edge Function..." -> wondering if we wanna make this consistent? Not a biggie though the Edge function code output spits out pretty fast tbh

joshenlim (approved):
LGTM! Verified on staging preview with both SQL, Edge Function, and JS code output in Assistant

PR #39285, opened 2025-10-06
fix auth logs bar chart

jordienr:
auth logs bar chart is not rendering in prod because the status is a string


cemalkilic (approved):

joshenlim (approved):
Changes LGTM - but not able to test on staging preview (got no auth logs in my projects 😅 )

PR #39298, opened 2025-10-06
fix(onboarding)add-me-to-supabase

smeubank:
Part of onboarding, and my general growing excitement about my new life and glory at supabase

🥁 🎭 🎉


smeubank:
@charislam attaching you for PR review cause I saw you reviewed a few of these such changes already 😅. I don't get from the onboarding if that's the normal MO

charislam (approved):
Welcome! 🎉

PR #39280, opened 2025-10-06
feat: custom bucket icons

dnywh:
## What kind of change does this PR introduce?

- Adds `Bucket` and `BucketAdd` custom icons
- Fixes a bunch of errors and documentation around custom icons

## What is the current behavior?

- We use the `Archive` [Lucide icon](http://lucide.dev/icons/archive) to represent buckets
- Our custom icon pipeline is broken and documentation fragmented
- Our bucket empty state is lacking iconography

## What is the new behavior?

- Replaces `Archive` [Lucide icon](http://lucide.dev/icons/archive) with `Bucket` icon to represent buckets
- Fixes custom icon creation pipeline
- Uses these two new icons in Studio
- Brings correct product icons to `ProjectUsage` and `ProjectUsageSection`

| Empty State | Policies |
| --- | --- |
| Storage  Supabase | Storage  Supabase |

## Additional context

- `BucketAdd` naming convention used instead of `BucketNew` or `BucketCreate` to distinguish it from the actual “new/create bucket” actions
- I wonder if we should rename the `Bucket` icon for similar reasons ([naming collision example](https://github.com/supabase/supabase/pull/39280/commits/e06d6e506d8a73ad0df0af24e1e6cbcdade93392))

This necessarily touches multiple parts of the repo:
- `/design-system` to store the new icons, documentation, and build process
- `/studio` to use the icons in production

## To test

Please follow [the feature preview instructions on this PR](https://github.com/supabase/supabase/pull/39154) in order to test the `BucketEmptyState` implementation of `BucketNew`.

- [ ] Smoke-test `interfaces/Home` and `interfaces/HomeNew` for fully functional icons
- [ ] Smoke-test `BucketEmptyState` (see above) for `BucketAdd` icon
- [ ] Smoke-test Storage Policies page (same with or without feature preview)

PR #39308, opened 2025-10-07
feat(studio): full file bucket functionality in storage

dnywh:

- Feature parity for file buckets
- Resolves DEPR-109

- Users with `isNewStorageUIEnabled` set to `true` could only _view_ a list of their file buckets

## What is the new behavior?

- Users with `isNewStorageUIEnabled` set to `true` now have full feature parity with everyone else
  - View bucket contents
  - Edit bucket
- We use the new table layout to show useful information about each bucket
  - Policies
  - Allowed MIME types
  - File size limit
  
| File Buckets | File Explorer |
| --- | --- |
| Storage  Supabase | Buckets  Supabase | 

## Additional context

Note that the new icons from https://github.com/supabase/supabase/pull/39280 are yet to be merged to `master` as of writing this. They shouldn’t clash.

## To test

The `storageAnalyticsVector` feature flag is set to `true` for all users on local and staging. You’ll just need to enable to corresponding feature preview.

- [ ] Then test that the newer Storage UI mentioned above in _new behavior_ is shown and functional


apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
dnywh, line 1: I’ve edited FileExplorerHeader so that the breadcrumbs don’t show until the user is 1+ directory deep. Otherwise it looks really odd with all the names and arrows pointing various directions:

Image

dnywh: This is done just for the `isNewStorageUIEnabled` folks so existing UI is unaffected.

apps/studio/components/interfaces/Storage/StorageExplorer/StorageExplorer.tsx
```diff
@@ -53,9 +53,6 @@ export const StorageExplorer = ({ bucket }: StorageExplorerProps) => {
   // Things like showing results from a search filter is "temporary", hence we use react state to manage
   const [itemSearchString, setItemSearchString] = useState('')
 
-  // Requires a fixed height to ensure that explorer is constrained to the viewport
-  const fileExplorerHeight = window.innerHeight - 122
-
```
dnywh, line 58: This code was four years old so I presume flex box improvements have allowed for these changes. We no longer need to hardcode this.

Affects current UI too.

dnywh: I had to make this change because the height is different for the new implementation of StorageExplorer. We show a header above this.

apps/studio/hooks/storage/useStoragePolicyCounts.ts
dnywh, line 1: I made this hook because:
- the API response does not include the count of policies
- the new design calls upon this information (per bucket) in two different areas
  - the FilesBuckets table column _Policies_
  - the [bucketid].tsx header button

apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx:1]: This is a fork of the [existing [bucketId].tsx file](https://github.com/supabase/supabase/blob/master/apps/studio/pages/project/%5Bref%5D/storage/buckets/%5BbucketId%5D.tsx). I made a fork because:
- This one has a new URL path (now nested inside of `/files/`)
- The logic is now quite different (no check for `bucket.type === 'ANALYTICS'`)

That said, I’m happy to consolidate via `isNewStorageUIEnabled`

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
dnywh: Here’s what it looks like now:
CleanShot 2025-10-07 at 15 36 23@2x

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -28,13 +34,27 @@ import { EmptyBucketModal } from './EmptyBucketModal'
...
   const [filterString, setFilterString] = useState('')
 
-  const { data: buckets = [], isLoading } = useBucketsQuery({ projectRef: ref })
+  const { data: buckets = [], isLoading: isLoadingBuckets } = useBucketsQuery({ projectRef: ref })
+  const { getPolicyCount, isLoading: isLoadingPolicies } = useStoragePolicyCounts(buckets)
+
+  const { data } = useProjectStorageConfigQuery({ projectRef: ref }, { enabled: IS_PLATFORM })
+
+  // Helper function to format bytes consistently
+  const formatBytes = (bytes: number) => {
```
joshenlim: we have a helper function under lib/helpers which should be able to replace this i think if we only need the string representation

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
```diff
@@ -85,28 +112,74 @@ export const FilesBuckets = () => {
                     </TableRow>
                   )}
                   {filesBuckets.map((bucket) => (
-                    <TableRow key={bucket.id}>
+                    <TableRow
+                      key={bucket.id}
+                      onClick={() => {
```
joshenlim: understandably we cant make table rows an `a` element, but what i like to do for this if we're making the row a link is to also consider if the meta key is pressed down in the onClick event so that we can still mimic the native browser event of opening links in a new tab if the meta key is pressed

i'll add this in no worries! 🙂 🙏

apps/studio/components/interfaces/Storage/StorageExplorer/StorageExplorer.tsx
```diff
@@ -178,7 +175,7 @@ export const StorageExplorer = ({ bucket }: StorageExplorerProps) => {
       ) : (
         <FileExplorerHeaderSelection />
       )}
-      <div className="flex h-full" style={{ height: fileExplorerHeight }}>
+      <div className="flex flex-1 min-h-0">
```
joshenlim, line 178: :chefs-kiss:

apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx
Comment (joshenlim) [reply apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx:1]: yup this is alright - keeps changes exclusive to the existing and feature preview UIs so we don't have to worry about existing UI getting affected from changes here

apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx
```diff
@@ -0,0 +1,110 @@
...
+  // [Joshen] Checking against projectRef from storage explorer to check if the store has initialized
+  if (!project || !projectRef) return null
+
+  if (isError) {
+    return <StorageBucketsError error={error as any} />
+  }
+
+  if (!isSuccess) {
+    return <div />
+  }
```
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx]: nit could return null and consolidate with line 31

joshenlim:
## Feedback

These are just some thoughts so not necessarily need to address (and can also address separately from this PR)

- RE breadcrumbs given that we're no longer showing the bucket name within the storage explorer itself, the "Navigate" CTA becomes a bit awkward then as it only shows up on hover. 
  image
  - Am wondering:
    - (1) We could probably keep this a permanently visible CTA or 
    - (2) Could we improve this UI? (or even remove really) Tbh this was there just so that users can navigate to a folder if they know a path but im not sure if its that useful at its current state (we don't have analytics to track user patterns for this too)

- Also RE breadcrumbs given that we've got space in the headers now, wdyt of making the breadcrumbs same behaviour as per list view where we show all the folders that are currently opened?
  image

- From a UX POV (no BE support atm) reckon it'll be nice to include the bucket size here actually, but we'll need to chat with Storage if that data can be returned on the /buckets endpoint.
  image
  - We could subsequently then sort of "link" this page to the org usage page (e.g how much of the org plan's limit for storage is this project currently using) - provide some sort of visibility / transparency from the usage/billing POV

joshenlim (approved):
Verified changes on staging preview! both with and without the feature preview

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
dnywh: Good catch, thanks! I hate being on the receiving end of that as a user (when it’s not set up).

PR #39310, opened 2025-10-07
Chore/projects pagination part 01

joshenlim:
## Context

Ongoing efforts to use the paginated projects endpoint, this PR deprecates the use of the old `useProjectsQuery` in the following places:

- `RouteValidationWrapper.tsx`
- Account Audit Logs
- Organization Audit Logs
- Organization Card (In the organizations home page - data is just to show the project counts)
- New project page

## Other changes:

- Renamed the existing `projects-infinite-query` to `org-projects-infinite-query`, so that we can implement another `projects-infinite-query` which retrieves all projects from a user (Used in account audit logs)
- Added some padding to organizations home page to reduce layout shift between organizations home page and organization home page (the one that shows the list of projects for a selected org)
- For organization audit logs, i've set `enabled` on GET organizations, roles, projects to be waiting on `isLoading` and `error` from the Audit logs request (reckon we don't need to fetch those data if the org doesn't have access to audit logs to begin with)

## To test

- [ ] Verify that RouteValidationWrapper logic is correct, and that you (1) can access projects you're in (2) cannot access projects you're not in
- [ ] Account + Organization audit logs can filter logs by selected project
  - Just FYI there's a known bug that I've left a comment for whereby the selected projects disappear from the label when doing async search, will address that separately as it involved several changes that's not related to this topic
- [ ] Verify that you can still create a new project
- [ ] Verify that you can see the project counts on each organization card in the organizations home page


apps/studio/components/interfaces/Account/AuditLogs.tsx
```diff
@@ -88,6 +105,13 @@ const AuditLogs = () => {
               valueKey="ref"
               activeOptions={filters.projects}
               onSaveFilters={(values) => setFilters({ ...filters, projects: values })}
+              search={search}
```
ivasilov, line 108: 2 issues:
- there's no loading indicator
- when filtering, the project name is lost somehow from the log rows (EDIT: saw it in description, it's known)

https://github.com/user-attachments/assets/98be0648-89bc-4104-8446-03f14f55a454

ivasilov (approved):

PR #39318, opened 2025-10-07
Fix org usage hidden UI

joshenlim:
## Context

Tiny fix in org usage page

## Before

image

## After

image


kemaldotearth (approved):
Not sure why `overflow-hidden` broke this, but :shipit: !

PR #39324, opened 2025-10-07
feat: add iceberg (analytics bucket) destination in UI

imor:
This PR adds the iceberg (analytics bucket) destination type to the UI.

image

image

Accompanying API PR: https://github.com/supabase/infrastructure/pull/26282


apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
```diff
@@ -283,16 +426,72 @@ export const DestinationPanel = ({
...
+    if (editMode) return
+
+    if (
+      selectedType === 'Analytics Bucket' &&
+      warehouseName &&
+      namespace &&
+      projectRef &&
+      (!s3Keys || s3Keys.bucketName !== warehouseName)
+    ) {
+      const createKeys = async () => {
```
joshenlim: any reason why we're doing this in a useEffect? I'd rather keys be created in the onSubmit handler instead as this creates a new set of keys each time the type is changed

joshenlim: i'm also running into an error here - but there's no network request getting triggered so there's something on the client side triggering this error
image

apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
```diff
@@ -283,16 +426,72 @@ export const DestinationPanel = ({
...
+        try {
+          const { snakeCase } = await import('lodash')
```
joshenlim: any reason why we're importing this way? could just import at the top of the page, i think this might be what's causing the client side error actually

apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
```diff
@@ -283,16 +426,72 @@ export const DestinationPanel = ({
...
+            }
+            setS3Keys(newS3Keys)
+
+            // Update form fields directly without resetting the entire form
+            form.setValue('s3AccessKeyId', newS3Keys.accessKey)
+            form.setValue('s3SecretAccessKey', newS3Keys.secretKey)
+          }
+        } catch (error) {
+          console.error('Failed to create S3 access keys:', error)
+          toast.error('Failed to create S3 access keys')
```
joshenlim: generally we should always float the error message up to the UI so users can understand why it failed

joshenlim (approved):
Verified on staging preview but I'd recommend checking through / tested once more on the staging preview before merging as I made quite a number of changes 🙏

apps/studio/components/interfaces/Database/Replication/NewPublicationPanel.tsx
```diff
@@ -171,4 +171,4 @@ const NewPublicationPanel = ({ visible, sourceId, onClose }: NewPublicationPanel
   )
 }
 
-export default NewPublicationPanel
+export { NewPublicationPanel }
```
joshenlim: could just export the component directly up top

apps/studio/components/interfaces/Database/Replication/DestinationPanel/DestinationPanelFields.tsx
```diff
@@ -0,0 +1,500 @@
...
+      enabled: type === 'Analytics Bucket' && !!catalogUri && !!warehouseName && !!serviceApiKey,
+    }
+  )
+
+  const { mutateAsync: createNamespace, isLoading: isCreatingNamespace } =
+    useIcebergNamespaceCreateMutation({
+      onSuccess: () => {
+        refetchNamespaces()
+      },
+    })
```
joshenlim: do we need this? we aren't creating name spaces here right?

joshenlim: removed this, as well as isCreatingNamespace usage

joshenlim (approved):
Double approve!

PR #39330, opened 2025-10-07
fix(ai): increase request body size limit

gregnr:
Currently API request bodies cap at 1MB. For our AI assistant, folks with a large amount of chat context sometimes reach this limit, so this bumps it to 5MB (then we will continue to monitor).


jordienr (approved):

saltcod (approved):
Tested on staging, looks good

PR #39338, opened 2025-10-07
feat: entitlements

ignaciodob:
https://linear.app/supabase/issue/BILLING-1678/centralized-entitlements-frontend-queries-and-methods

This PR introduces a new entitlements system that allows organizations to view their feature entitlements based on their subscription plan and any custom overrides.

# Proposed changes

### `useCheckEntitlements.ts`
This hook provides a way to check if a user has access to specific features based on their organization's entitlements. It abstracts away the complexity of fetching and processing entitlement data.
- Fetches organization entitlements using React Query
- Finds specific entitlement by `featureKey` and returns access status
- Returns both boolean access status and raw entitlement value

### `entitlements-query.ts`
This file contains the low-level data fetching logic for organization entitlements using React Query. Composed of:
- `getEntitlements()`: Async function that fetches entitlements from the API
- `useEntitlementsQuery()`: React Query hook that wraps the API call with caching and state management

### DIFF
--- a/packages/api-types/types/platform.d.ts
+++ b/packages/api-types/types/platform.d.ts
@@ -1251,6 +1251,46 @@ export interface paths {
     patch?: never
     trace?: never
   }
+  '/platform/organizations/{slug}/entitlements': {
+    parameters: {
+      query?: never
+      header?: never
+      path?: never
+      cookie?: never
+    }
+    /**
+     * Get entitlements for an organization
+     * @description Returns the entitlements available to the organization based on their plan and any overrides.
+     */
+    get: operations['OrganizationEntitlementsController_getEntitlements']
+    put?: never
+    post?: never
+    delete?: never
+    options?: never
+    head?: never
+    patch?: never
+    trace?: never
+  }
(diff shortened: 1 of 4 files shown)


packages/api-types/types/platform.d.ts
```diff
@@ -4738,7 +4778,16 @@ export interface components {
       description?: string
       name: string
       /** @enum {string} */
-      type: 'postgres' | 'bigquery' | 'webhook' | 'datadog' | 'elastic' | 'loki'
+      type:
```
kevcodez: you'll have to remove these changes as they cause the TS failures and it's waiting on analytics team to implement, have flagged with them already

packages/api-types/types/platform.d.ts
```diff
@@ -6896,7 +6950,16 @@ export interface components {
       name: string
       token: string
       /** @enum {string} */
-      type: 'postgres' | 'bigquery' | 'webhook' | 'datadog' | 'elastic' | 'loki'
+      type:
```
kevcodez: same here

kevcodez (approved):

PR #39342, opened 2025-10-08
feat(studio): analytics bucket functionality in storage

dnywh:

- Feature parity for analytics buckets
- Resolves DEPR-110

Users with `isNewStorageUIEnabled` set to `true` can only see the empty state for analytics storage.

## What is the new behavior?

Users with `isNewStorageUIEnabled` set to `true` can now:
- [x] View existing buckets instead of the empty state, if applicable
- [x] Open the bucket creation dialog
- [ ] Create analytics buckets (if they have the feature enabled)
  - [ ] Alert correctly shows whether extension and/or integration need to be installed on their behalf
    - [ ] That extension and/or integration are indeed installed on the user’s behalf
  - [x] Are blocked from creating analytics buckets if the feature is not enabled for them
- [ ] View bucket contents
- [ ] Create tables within buckets
  - [ ] ~~Be able to create new schema from the _Target schema_ dropdown~~

## Additional context

Vector storage is not yet built out, hence the code ‘stubs’ in the bucket creation dialog shared between `ANALYTICS` and `VECTOR`.

Known issues with this draft PR:
- [ ] The `wrappersExtension?.installed_version` returns `null`, meaning we can’t conditionally show when just that wrapper needs to be installed
- [x] I don’t have the analytics feature enabled for my project `pdprmxybiwdwlmntsbxh` so can’t progress past the creation dialog, which means:
  - [ ] Actually installing the Wrappers extension and/or Iceberg Wrapper integration on the user’s behalf probably doesn’t work yet

## To test

The `storageAnalyticsVector` feature flag is set to `true` for all users on local and staging. You’ll just need to enable to corresponding feature preview.


dnywh:
Cancelling as https://github.com/supabase/supabase/pull/39545 has subsumed this work into existing route:
`apps/studio/components/interfaces/Storage/AnalyticsBuckets.tsx`

PR #39343, opened 2025-10-08
Fix project filter on homepage

joshenlim:
## Context

The project status filter on the homepage had a bit of an odd behaviour which starts of as all checked, but if you select one of the status, it checks that status while unchecking the other

Fixing that odd behaviour such that
- filter statuses are unchecked by default (i.e. all projects will be returned)
- checking a status will then return projects based on the status filter (so if both active healthy and paused are selected, only projects of those 2 status will be returned)

## To test
- [ ] Ensure that the status works for returning the expected projects
- [ ] Just verify that the UX doesn't feel weird too


alaister (approved):
LGTM!

- Tested on preview

PR #39344, opened 2025-10-08
Handle context exceeded errors

SaxonF:
On the occasion when context windows has been exceeded, provide more explicit instructions and a way to create new chat.


apps/studio/components/ui/AIAssistantPanel/AIAssistant.tsx
```diff
@@ -305,6 +305,11 @@ export const AIAssistant = ({ className }: AIAssistantProps) => {
     [chatMessages, project?.ref, selectedOrganization?.slug, rateMessage, sendEvent]
   )
 
+  const isContextExceededError =
+    error &&
+    (JSON.stringify(error).includes('context_length_exceeded') ||
+      JSON.stringify(error).includes('exceeds the context window'))
```
joshenlim: we could use `error.message` instead of JSON.stringifying the error object here

apps/studio/components/ui/AIAssistantPanel/AIAssistant.tsx
```diff
@@ -458,29 +463,50 @@ export const AIAssistant = ({ className }: AIAssistantProps) => {
...
+                      {isContextExceededError ? (
+                        <p>
+                          This conversation has become too long for the Assistant to process. Please
+                          start a new chat to continue.
+                        </p>
+                      ) : (
+                        <p>
+                          Sorry, I'm having trouble responding right now. If the error persists
+                          while retrying, you may try clearing the conversation's messages and try
+                          again.
```
joshenlim: come to think of it - we've removed the ability to clear messages in a conversation right?
there's only starting a new chat?

SaxonF: you can delete messages instead of clearing but yes good point let's reframe this copy too

joshenlim (approved):
Minimally tested the assistant on staging preview

PR #39348, opened 2025-10-08
Chore/projects pagination part 04

joshenlim:
## Context

- `OrgProjectSwitcher`
- `NotificationsFilter`
- `NotificationsPopover`
- `ProjectDropdown` 
  - Replace `useProjectsQuery` with `useProjectDetailQuery` for getting parent project detail if on a branch
- `LoadingState` 
  - Replace `useProjectsQuery` with `useProjectDetailQuery`
- `org-ai-details` 
  - Replace `getProjects` with `getProjectDetail`
- `org/[slug]/index.tsx` 
  - Remove `useAutoPrefetchProjects`
- `NewOrgForm` 
  - Replace `useProjectsQuery` with `useProjectsInfiniteQuery`

Probably just need to test Notifications Filter from the header nav, the rest are pretty subtle and have no changes in terms of functionality

But this will prevent calls to /projects when landing on the home page

## Things to note

- RE `OrgProjectSwitcher` (within Cmd K) -> swapping to use the paginated endpoint means that we won't be listing every single project in Cmd K. 
  - Ideally we're able to support some async searching within Cmd K but reckon we can do that separately
  - cc @charislam on this one in case you have some ideas at the top of your head - not a priority though so no worries
- RE `NotificationsFilter` -> Opting to _not_ support infinite loading here as the Filter Popover UI here is quite complex with a mix of sync and async data. 
  - Opting to show this text call out instead if the number of projects the user has access to exceeds the default limit (I'd suppose only a subset of customers will be running into this). 
     image
  - We'll still support async search here nonetheless so not a blocker for finding projects
    image
- RE `NewOrgForm` because we're using a paginated endpoint, there's a chance that not all projects will be factored in here (page limit is 100 results).
  - The projects data is mainly used to compute `hasFreeOrgWithProjects` within `onSubmit` to show the confirmation modal whether the user wants to upgrade existing projects instead
  - I _don't  think_ (?) this is critical functionality so am okay for now, but ideally perhaps this data can be computed on the API and returned in /profile or something instead (since this data is on the account level) - cc @kevcodez on this one


staaldraad (approved):

apps/studio/lib/ai/org-ai-details.ts
```diff
@@ -11,28 +11,20 @@ export const getOrgAIDetails = async ({
...
-      },
-    }),
+  const headers = {
+    'Content-Type': 'application/json',
+    ...(authorization && { Authorization: authorization }),
+  }
+
+  const [organizations, selectedProject] = await Promise.all([
+    getOrganizations({ headers }),
+    getProjectDetail({ ref: projectRef }, undefined, headers),
```
ivasilov, line 21: Will this fetch in case the project ref is preview branch ref?

ivasilov (approved):

apps/studio/lib/ai/org-ai-details.ts
joshenlim: yup `getProjectDetail` works for both primary branches and preview branches

PR #39349, opened 2025-10-08
feat: optimized users page

hf:
Provides an optimized view in the Users page that is light on the database while providing the essential search function for users.

## To test
- [ ] Verify that the UX between "optimized" and "freeform" search modes do not feel awkward
- [ ] Verify that both search modes are working as expected


saltcod:
this is slightly more concise: 
This may impact your database—especially if your project has many users. Use with caution.

than this:
This may adversely impact your database, in particular if your project has a large number of users - use with caution.

saltcod (approved):
Tested all and looks good on staging.

saltcod:
Small nit: I'm left wondering what "a lot of users" is. 100? 1000? 10,000?

saltcod:
Maybe one more tiny nit: 
"be careful" instead of "not recommended"? 



screenshot-2025-10-13-at-20 36 57

joshenlim:
> this is slightly more concise:

gotcha sounds great!

> I'm left wondering what "a lot of users" is.

It's a great question and i was on the fence about putting a user-facing number to this since its quite arbitrary 😅 
but decided that we'll land with > 1M as the benchmark
image


> "be careful" instead of "not recommended"?

oh yup i updated the copy to "Searching across all columns is not recommended"

PR #39350, opened 2025-10-08
Chore/projects pagination part 05

joshenlim:
Branches out of https://github.com/supabase/supabase/pull/39348

## Context

- `AIAssistantOption`
  - Remove use of `useProjectsQuery` entirely
  - Default to wildcard route for AI Assistant call out if no projects selected (I think this makes more sense to rather than just directing to the first project that the user has access to)
- `PlanUpdateSidePanel`
  - Replace `useProjectsQuery` with `useOrgProjectsInfiniteQuery`
  - cc @kevcodez JFYI for this one - projects data is being used in the `hasMembersExceedingFreeTierLimit`, to check if the current organization has any running projects, but because we're using a paginated endpoint there's technically a chance that some projects might be omitted here
  - Although I think for this specific case it shouldn't really matter since we're just checking for the existence of a running project (chances of an org having > 100 projects all paused i suppose will be incredibly rare 😅)
- `NoProjectsOnPaidPlan`
  - Replace `useProjectsQuery` with `useOrgProjectsInfiniteQuery`
- `IntegrationPanels`
  - Replace `useProjectsQuery` with `useProjectDetailQuery`
- `IntegrationConnection`
  - Replace `useProjectsQuery` with `useProjectDetailQuery`


apps/studio/components/interfaces/Support/AIAssistantOption.tsx
```diff
@@ -42,16 +39,8 @@ export const AIAssistantOption = ({
...
-      return projectRef
-    }
-    // If no specific project selected, use first project from the org
-    const orgProjects = projects?.filter((p) => p.organization_slug === organizationSlug)
-    return orgProjects?.[0]?.ref || '_'
-  }
-
-  const aiLink = `/project/${getProjectRef()}?aiAssistantPanelOpen=true`
+  // If no specific project selected, use the wildcard route
+  const aiLink = `/project/${projectRef !== 'no-project' ? projectRef : '_'}?aiAssistantPanelOpen=true`
```
ivasilov: This is a change in behaviour, it used to preselect the first project in the org but now it always sends you to select a project. Not sure if expected.

ivasilov (approved):

apps/studio/components/interfaces/Support/AIAssistantOption.tsx
joshenlim: yup! i was thinking this might make more sense actually since we're only sending users to the wildcard route if "no specific project" is selected

was thinking the current behaviour feels a little weird:
e.g if i selected "no specific project" in the support and I click on the CTA to use the AI assistant, then i get sent to (from the user's POV at least, despite it being the first project in the org) a random project to use the assistant

although what do you reckon though? happy to revert this nonetheless if you reckon the original behaviour is better 🙏

doublethink (approved):

apps/studio/components/interfaces/Support/AIAssistantOption.tsx
joshenlim: i'll keep this change but added one small change

AI Link when linking to the project wildcard route, should minimally pre-select the organization based on the selected organization in the support form, so that users can choose which project from within that organization to use the AI assistant in

joshenlim: i'll keep this change but added one small change

AI Link when linking to the project wildcard route, should minimally pre-select the organization based on the selected organization in the support form, so that users can choose which project from within that organization to use the AI assistant in

PR #39351, opened 2025-10-08
chore/projects pagination part 06

joshenlim:
Branches out of https://github.com/supabase/supabase/pull/39350

## Context

- `ProjectLinker`
  - This component is being used by the below two files, am shifting the logic of fetching supabase projects into this component (rather than being passed in as a prop)
  - Use `OrganizationProjectSelector` component instead to handle the project selection / loading
- `SidePanelVercelProjectLinker`
- `choose-project`
  - JFYI when this component was passing in the supabase projects as prop into `ProjectLinker`, its specifically passing in projects of a set of statuses that seems to just omit "Paused" or "Pausing"
  - While the new paginated endpoint has a `statuses` filter, we're only supporting 'ACTIVE_HEALTHY' and 'INACTIVE' as valid status values so we can't support all those status that the file was initially filtering against
  - So with this change on this UI, all projects of all statuses will be shown in the dropdown, but I'm gonna show a "Paused" or "Pausing" badge to somewhat keep that original behaviour from a UX POV
  image


staaldraad (approved):

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
```diff
@@ -54,6 +54,8 @@ export const PlanUpdateSidePanel = () => {
   const { data: selectedOrganization } = useSelectedOrganizationQuery()
   const { mutate: sendEvent } = useSendEventMutation()
 
+  console.log({ selectedOrganization })
```
ivasilov: Forgotten console.log.

ivasilov (approved):

PR #39354, opened 2025-10-08
change connection string via dropdown

fsansalvadore:
This PR updates the _Connection String_ tab layout in the `Connect` dialog to reduce cognitive load when needing to find the necessary connection string.

- A new `Method` dropdown is added with "Direct Connection", "Transaction Pooler" and "Session Pooler" options
- Updated layout to show descriptive content on the left column and actionable content on the right
- Positioned the actual connection string to be more prominent and obvious
- Forced the connection string to break on a new line instead of being hidden by overflow
- Also added sync with url state to enable deep-linking

Additional change:
- fix three step layout that appears on some of the _Connection String_ `Type` options that becomes inconsistent with new layout proposal.

## Before

Screenshot 2025-10-08 at 18 40 18

"steps" layout that also needs to be aligned:
Screenshot 2025-10-13 at 14 49 22

## After

Screenshot 2025-10-13 at 15 10 32

Screenshot 2025-10-08 at 18 38 39

Fixed "Steps layout":
Screenshot 2025-10-13 at 14 51 41


apps/studio/components/interfaces/Connect/DatabaseSettings.utils.ts
```diff
@@ -38,7 +38,7 @@ export const getConnectionStrings = ({
 } => {
   const isMd5 = poolingInfo?.connectionString.includes('options=reference')
   const { projectRef } = metadata
-  const password = '[YOUR-PASSWORD]'
+  const password = '[YOUR_PASSWORD]'
```
fsansalvadore, line 41: this is a silly change but

1. word-breaks as one word
2. slightly improves UX as it can be selected via double click

dnywh (approved):
Smoke tested on staging and flicked through changes here. Looks good!

FYI: dialog looks a bit funny on smaller breakpoints but this is unrelated to your changes:
- Square edges on dialog
- Centered title and subtitle

joshenlim (approved):

PR #39357, opened 2025-10-08
feat: use new org daily stat endpoint

kevcodez:
This adjusts the frontend to use the new org daily stats endpoint.

Staging pipeline is not hooked up, so there is no data to display.


packages/api-types/types/api.d.ts
```diff
@@ -4280,7 +4322,7 @@ export interface operations {
         organization_slug?: string
         redirect_uri: string
         /** @description Resource indicator for MCP (Model Context Protocol) clients */
-        resource?: 'https://api.supabase.green/mcp' | 'https://mcp.supabase.green/mcp'
+        resource?: 'http://localhost:8080/mcp' | 'http://localhost:8080/mcp'
```
joshenlim: @kevcodez this one's not FE specific but wondering if this is intentional on the API end?

kevcodez: Not intentional, I just regenerated specs, I can remove this one

apps/studio/components/interfaces/Organization/Usage/Usage.tsx
```diff
@@ -112,6 +113,16 @@ export const Usage = () => {
...
   }, [projectRef, isSuccessProjectDetail])
 
+  const { data: orgDailyStats, isLoading: isLoadingOrgDailyStats } = useOrgDailyStatsQuery(
+    {
+      orgSlug: slug,
+      projectRef,
+      startDate,
+      endDate,
+    },
+    { enabled: slug !== undefined }
```
joshenlim: this one's unnecessary since its already handled in the query itself (i'll push up the change)

joshenlim: might be missing an error handler here (will add it in)

packages/api-types/types/api.d.ts
joshenlim: that's all good no worries - guessing this needs to be addressed on the API end then

joshenlim (approved):
Minimally verified on staging preview call to /usage/daily succeeds + UI renders correctly! 🙂 🙏

ecktoteckto (approved):
Tested locally with data from our Supabase org on lilve.

PR #39361, opened 2025-10-08
fix crash in r2np

jordienr:
I think the API will not return the ref if the "og" project was removed.


joshenlim (approved):
Changes LGTM!

PR #39363, opened 2025-10-08
Add perms check for storage migration callout

joshenlim:
## Context

As per PR title - just adding permissions checking for the storage migration callout (which is equivalent to the perms for updating storage config). Will show a tooltip with the button disabled if the user doesn't have sufficient perms

image


alaister (approved):
LGTM!

PR #39364, opened 2025-10-08
Mitigate SQL Editor "Unable to find snippet" issue

joshenlim:
## Context

We've been receiving multiple reports of users running into "Unable to find snippet ID..." errors in the SQL Editor after typing for a bit.
image

## Investigation
- This _may_ very well be due to the replication lag problems that dashboard has been facing, but I noticed there's some inefficiencies on the dashboard that we can improve on
- Noticed that lodash's `debounce` in `MonacoEditor` wasn't working as expected ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/SQLEditor/MonacoEditor.tsx#L165))
  - If you add a `console.log` in `setSql` in the `sql-editor-v2` valtio store, it's getting triggered even if I'm continuously typing ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/state/sql-editor-v2.ts#L118))
  - I'm not too sure why this is happening to be very honest, and I haven't checked in other parts of the dashboard that we're using lodash's `debounce`
- The fact that we're trying to save the snippet once the user starts typing is also probably what's causing the issue ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/SQLEditor/MonacoEditor.tsx#L186))

## Changes involved
- Opting to use `useDebounce` instead with a local state to track the input from the MonacoEditor
  - We then have a `useEffect` depending on the debounced value to fire `setSql`
- In `handleEditorChange`, opting to _not_ upsert the snippet to the DB immediately when creating a new snippet
  - Let debounce handle the snippet saving which I think makes more sense
- With these changes at the very least, users won't run into the "Unable to find snippet" error _while_ typing, but after the debounce

## To test
Understandably this doesn't happen all the time, especially locally but probably just good to verify running queries on a "New" tab that everything works as expected. Do this multiple times just to stress test things

## What else we might be able to do
- Another idea that i'm thinking on exploring in this PR as well, is the error handling itself
- e.g if fetching content from a "new" snippet and it errors out, don't block the UI - show a toast or silently try again in the background
- but I'll need to figure out how to differentiate that from an actual non-existing snippet

### DIFF
--- a/apps/studio/components/interfaces/SQLEditor/MonacoEditor.tsx
+++ b/apps/studio/components/interfaces/SQLEditor/MonacoEditor.tsx
@@ -1,8 +1,8 @@
 import Editor, { Monaco, OnMount } from '@monaco-editor/react'
-import { debounce } from 'lodash'
 import { useRouter } from 'next/router'
-import { MutableRefObject, useEffect, useRef } from 'react'
+import { MutableRefObject, useEffect, useRef, useState } from 'react'
 
+import { useDebounce } from '@uidotdev/usehooks'
 import { LOCAL_STORAGE_KEYS, useParams } from 'common'
 import { useLocalStorageQuery } from 'hooks/misc/useLocalStorage'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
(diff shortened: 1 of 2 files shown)


charislam (approved):
tested as described

PR #39379, opened 2025-10-08
fix(studio): dynamic size for query perf time consumed col

kemaldotearth:

YES

This adds a dynamic resizing for when there's overflow data in the time consumed column.

| Before | After |
|--------|--------|
| Screenshot 2025-10-08 at 22 07 35 | Screenshot 2025-10-08 at 22 21 54 |


apps/studio/components/interfaces/QueryPerformance/QueryPerformanceGrid.tsx
```diff
@@ -38,6 +38,30 @@ interface QueryPerformanceGridProps {
   queryPerformanceQuery: DbQueryHook<any>
 }
 
+const calculateTimeConsumedWidth = (data: any[]) => {
```
jordienr: is it possible to use the type here? If not we should validate the rows before trying to access the properties, so that it doesn't crash if something unexpected is returned from the API.

jordienr (changes requested):

jordienr (approved):

PR #39384, opened 2025-10-09
Add hide/show anon key button to edge function details page

joshenlim:
## Context

Adds a "Hide/show anon key" button in Edge Function details page that'll toggle the visibility of the anon key in the code block. Defaults to hidden at the beginning which will be helpful if the user is screen sharing 

image

Separately, also removed `react-copy-to-clipboard` package dependency - swapped out to use our `copyToClipboard` util method instead

## For Design team:
JFYI i've also updated the `CodeBlock` and `SimpleCodeBlock` components to support an optional `handleCopy` prop, which will be called instead of the internal copy method within the component if provided. 

This is mainly to support copying content with obfuscated values when displayed in the UI, but you'd want the actual values to be in the copied content (such as in this case, the anon key is obfuscated in the UI but you'd want the key's value to not be obfuscated when copying the code content)

Although separately, wondering if we should / could de-dupe the `CodeBlock` and `SimpleCodeBlock` components as well - we'd possibly be able to remove some package dependency as well since both these components I think serve the same purpose but are using different libraries for syntax highlighting

## To test

- [ ] Verify that you can hide / show anon key in the edge functions details page
- [ ] Verify that when you copy the curl value in the "Invoke function" while key is hidden, the copied value will contain the actual key value


dnywh (approved):
Can confirm:

> - [x] Verify that you can hide / show anon key in the edge functions details page
> - [x] Verify that when you copy the curl value in the "Invoke function" while key is hidden, the copied value will contain the actual key value

Looks good! Will leave broader design feedback elsewhere.

PR #39386, opened 2025-10-09
Chore/consolidate copy icons

joshenlim:
## Context

This just consolidate all icons for "Copy" actions to use the Copy icon instead of the Clipboard icon

Also addresses a separate issue in the Product Menu -> The background color of the `Menu.Item` in light mode when `active` has barely any differentiation (if any?) against the background color of the side panel (with `bg-surface-200`). Opting to use `bg-surface-400` in light mode when active but i reckon we might need to revisit the light mode colors nonetheless

## Before
image

## After
image

### DIFF
--- a/packages/ui/src/lib/theme/defaultTheme.ts
+++ b/packages/ui/src/lib/theme/defaultTheme.ts
@@ -1189,7 +1189,7 @@ export default {
             group-hover:border-foreground-muted`,
           active: `
             font-semibold
-            bg-sidebar-accent dark:bg-surface-200
+            bg-sidebar-accent
             text-foreground-lighter
             z-10 rounded-md
           `,
--- a/packages/ui/src/lib/theme/defaultTheme.ts
+++ b/packages/ui/src/lib/theme/defaultTheme.ts
@@ -1182,20 +1182,16 @@ export default {
           rounded: `rounded-md`,
         },
         pills: {
-          base: `
-            px-3 py-1
-          `,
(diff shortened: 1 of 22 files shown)


packages/ui/src/lib/theme/defaultTheme.ts
```diff
@@ -1191,7 +1191,7 @@ export default {
             group-hover:border-foreground-muted`,
           active: `
             font-semibold
-            bg-surface-200
+            bg-surface-400 dark:bg-surface-200
```
dnywh: Thoughts on using the same `active:bg-sidebar-accent` as the root sidebar?

This seems noticeably darker than what’s currently in prod ...which I just noticed is also not using the same background color as the root sidebar!

- `active:bg-sidebar-accent` seems to map to `rgb(237, 237, 237);`
- Selected item on ‘inner’ sidebar maps to:
  - `rgb(243, 243, 243);` on prod
  - `#e5e5e5` on this PR

 | Before | After |
| --- | --- |
| Image | Image |

Anyway, the bigger issue for me here—as a daily light mode user—is that this feels too dark. I didn’t have a problem with the currently-in-prod grey, and it seems that using the root sidebar tint (which is darker) might fix both things?

joshenlim: ohhhh that sounds better! yup +1 😄

dnywh (approved):
Looks good!

PR #39392, opened 2025-10-09
[FE-1926] chore: show sign in error

alaister:
Currently errors while signing in result in an ephemeral toast. For example:
Screenshot 2025-10-09 at 17 27 42

This PR adds an inline error message along with actions that the user can take:
Screenshot 2025-10-09 at 17 27 28

To test:

- [ ] Ensure sign in works as normal (happy path)
- [ ] Test error message actions. You can fake an error while logged out by visiting a path like: `/sign-in-mfa?error=server_error&error_code=unexpected_failure&error_description=Error+getting+user+email+from+external+provider&method=snapchat&returnTo=%2Forganizations#error=server_error&error_code=unexpected_failure&error_description=Error+getting+user+email+from+external+provider`

### DIFF
--- a/apps/studio/components/interfaces/SignIn/SignInMfaForm.tsx
+++ b/apps/studio/components/interfaces/SignIn/SignInMfaForm.tsx
@@ -8,6 +8,7 @@ import { useEffect, useState } from 'react'
 import { SubmitHandler, useForm } from 'react-hook-form'
 import z from 'zod'
 
+import { useAuthError } from 'common'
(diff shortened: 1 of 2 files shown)


encima:
With MFA disabled for this, should the title of the page be different?

ivasilov (approved):

alaister:
@encima there isn't a super easy way to determine if MFA is disabled specifically, so we can customise this as part of the follow up tasks here as well.

PR #39398, opened 2025-10-09
fix: During project claim, disable the orgs for choosing who can't add another projects

ivasilov:
This PR fixes a flow where the user might choose an org for project claiming which can't create/add another projects because of limits for free orgs.

Fixes FE-1937.


apps/studio/components/interfaces/Organization/ProjectClaim/confirm.tsx
```diff
@@ -41,13 +41,22 @@ export const ProjectClaimConfirm = ({
...
 
   const { mutateAsync: claimProject, isLoading: isClaiming } = useOrganizationProjectClaimMutation()
 
   const onClaimProject = async () => {
     try {
+      // try {
       await approveRequest({ id: auth_id!, slug: selectedOrganization.slug })
+      // } catch (error: any) {
+      //   // if the org is already approved with this auth id, we can ignore the error. Otherwise, we throw it again
+      //   if (!error.message.includes('already approved')) {
```
charislam: curious why we're not going with the commented out approach? seems safer than blanket ignoring all errors

ivasilov: Oh, you're right, that code shouldn't be commented, sorry!

phamhieu (approved):

w3b6x9 (changes requested):
@ivasilov isn't there an issue with this flow if the user is blocked from claiming project yet the oauth app is approved in the background? Bit of an unexpected state where user was unable to claim the project yet AI Builder is able to get their OAuth app approved.

w3b6x9 (approved):

apps/studio/components/interfaces/Organization/ProjectClaim/confirm.tsx
```diff
@@ -41,13 +41,14 @@ export const ProjectClaimConfirm = ({
   const queryClient = useQueryClient()
 
   const { mutateAsync: approveRequest, isLoading: isApproving } =
-    useApiAuthorizationApproveMutation()
+    useApiAuthorizationApproveMutation({ onError: () => {} })
```
ivasilov, line 44: To avoid double toasts, there's another in a `try/catch` block which catches errors from 2 API calls.

apps/studio/components/ui/org-selector.tsx
```diff
@@ -66,29 +142,7 @@ export function OrganizationSelector({ onSelect, maxOrgsToShow = 5 }: ProjectCla
               <div className="text-center text-foreground-light py-6">No organizations found.</div>
             )}
             {filteredOrgs.map((org) => (
-              <Card
```
ivasilov, line 69: Moved to a separate component to fetch org info whether the org can add/create another project.

charislam (approved):
Tested in local with mocked requests and works as described!

PR #39403, opened 2025-10-09
Auth Report Fixes

jordienr:
- moves usage queries to 1 query

Having different queries for each makes it difficult to always have the same number of data points and timestamps in the final data, specially after transformations, by moving it to 1 query we make sure the 4 related charts have the same data, and comparing the different metrics more robust.

- fixes refresh

### DIFF
--- a/apps/studio/data/reports/v2/auth.config.ts
+++ b/apps/studio/data/reports/v2/auth.config.ts
@@ -145,14 +145,6 @@ function buildAuthLatencyAllSQL(
   endDate: string
 ) {
   const granularity = analyticsIntervalToGranularity(interval)
-  const groupByProvider = Boolean(filters?.provider && filters.provider.length > 0)
-
-  const providerClause =
-    filters?.provider && filters.provider.length > 0
-      ? `AND COALESCE(JSON_VALUE(event_message, '$.provider'), 'unknown') IN (${filters.provider
-          .map((p) => `'${p.replace(/'/g, "\\'")}'`)
-          .join(', ')})`
-      : ''
 
   return `
     -- Auth latency (sign-in + sign-up), single scan, FE does zero-fill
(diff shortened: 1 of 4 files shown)


apps/studio/data/reports/v2/auth.config.ts
```diff
@@ -33,177 +33,205 @@ const METRIC_KEYS = [
...
+    base AS (
+      SELECT
+        TIMESTAMP_TRUNC(timestamp, ${granularity}) AS ts,
+        COALESCE(JSON_VALUE(event_message, '$.auth_event.action'), JSON_VALUE(event_message, '$.action')) AS action,
+        COALESCE(JSON_VALUE(event_message, '$.auth_event.actor_id'), JSON_VALUE(event_message, '$.user_id')) AS actor_id,
+        COALESCE(JSON_VALUE(event_message, '$.provider'), 'unknown') AS provider,
+        JSON_VALUE(event_message, '$.login_method') AS login_method,
+        CASE
+          WHEN JSON_VALUE(event_message, '$.metering') IS NOT NULL THEN LOWER(JSON_VALUE(event_message, '$.metering'))
+          WHEN JSON_QUERY(event_message, '$.metering') = 'true' THEN 'true'
```
charislam, line 79: Not super familiar with BigQuery, but is it impossible for the second branch to be hit?

charislam (commented):
Dunno if I'm thinking about this the wrong way, but the numbers aren't what I expected. I added a single new user on a project I haven't touched in days, and the logs show what I expect with a single call to `/admin/users`:

Image

But in reports, I get 2 active users:

Image

apps/studio/data/reports/v2/auth.config.ts
jordienr: yup sorry not done with the PR, shouldve made a draft

PR #39409, opened 2025-10-09
chore: add telemetry for users search

fadymak:
Adds telemetry for searches conducted on `/dashboard/project/{ref}/auth/users` to better understand usage patterns and tune queries and UI to better fit the use cases.


apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -182,6 +186,18 @@ export const UsersV2 = () => {
   // [Joshen] Only relevant for when selecting one user only
   const selectedUserFromCheckbox = users.find((u) => u.id === [...selectedUsers][0])
 
+  const telemetryProps = {
+    sort_column: sortColumn,
+    sort_order: sortOrder,
+    providers: selectedProviders,
+    user_type: filter,
```
seanoliver: It might be clearer to match the `"all"` behavior here to `undefined`, just like on L177 and L156 above. That way, it’s explicit that `"all"` means "no filter" rather than a literal `"all"` user type and it keeps the telemetry data consistent with the query payload.

You could use the same ternary logic:
```js
user_type: filter === 'all' ? undefined : filter
```

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -331,8 +347,18 @@ export const UsersV2 = () => {
...
-                      setFilterKeywords(search.trim().toLocaleLowerCase())
+                      const query = search.trim()
+                      setSearch(query)
+                      setFilterKeywords(query.toLocaleLowerCase())
+                      sendEvent({
+                        action: 'auth_users_searched',
+                        properties: {
+                          trigger: 'search_input',
+                          ...telemetryProps,
+                          keywords: query,
```
seanoliver: This could cause data consistency issues since we’re sending the non-lowercased `query` here, but all other calls use the lowercased value (set on L352 and passed through on L194).

For example, the first search event might send `JohnDoe@example.com`, but subsequent events (filter, refresh, etc.) would send `johndoe@example.com`.

seanoliver (approved):
Overall this LGTM! Just added a few thoughts on data cleanliness to make analysis easier down the line.

packages/common/telemetry-constants.ts
```diff
@@ -1901,6 +1901,49 @@ export interface ApiDocsCodeCopyButtonClickedEvent {
...
 
+/**
+ * User performed a search via the Auth Users page.
+ *
+ * @group Events
+ * @source studio
+ * @page /dashboard/project/{ref}/auth/users
+ */
+export interface AuthUsersSearchedEvent {
+  action: 'auth_users_searched'
```
seanoliver:
```suggestion
  action: 'auth_users_search_submitted'
```

packages/common/telemetry-constants.ts
```diff
@@ -1901,6 +1901,49 @@ export interface ApiDocsCodeCopyButtonClickedEvent {
...
 }
 
+/**
+ * User performed a search via the Auth Users page.
+ *
+ * @group Events
+ * @source studio
+ * @page /dashboard/project/{ref}/auth/users
+ */
+export interface AuthUsersSearchedEvent {
```
seanoliver: One quick note — I was chatting with @pamelachia and we were thinking it might be clearer to rename this to something like `AuthUsersSearchSubmittedEvent`. “Submitted” is the verb we already use in our event naming, and it better reflects the user action. This would also keep our event terminology consistent going forward.

```suggestion
export interface AuthUsersSearchSubmittedEvent {
```

packages/common/telemetry-constants.ts
```diff
@@ -2014,3 +2057,4 @@ export type TelemetryEvent =
   | TableCreatedEvent
   | TableDataAddedEvent
   | TableRLSEnabledEvent
+  | AuthUsersSearchedEvent
```
seanoliver:
```suggestion
  | AuthUsersSearchSubmittedEvent
```

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -448,7 +499,23 @@ export const UsersV2 = () => {
...
                   </DropdownMenuTrigger>
                   <DropdownMenuContent className="w-44" align="start">
-                    <DropdownMenuRadioGroup value={sortByValue} onValueChange={setSortByValue}>
+                    <DropdownMenuRadioGroup
+                      value={sortByValue}
+                      onValueChange={(value) => {
+                        const [sortColumn, sortOrder] = value.split(':')
+                        setSortByValue(value)
+                        sendEvent({
+                          action: 'auth_users_searched',
```
seanoliver:
```suggestion
                          action: 'auth_users_search_submitted',
```

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -389,7 +429,18 @@ export const UsersV2 = () => {
                     labelClass="text-xs"
                     maxHeightClass="h-[190px]"
                     className="w-52"
-                    onSaveFilters={setSelectedProviders}
+                    onSaveFilters={(providers) => {
+                      setSelectedProviders(providers)
+                      sendEvent({
+                        action: 'auth_users_searched',
```
seanoliver:
```suggestion
                        action: 'auth_users_search_submitted',
```

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -349,7 +375,21 @@ export const UsersV2 = () => {
                 />
 
                 {showUserTypeFilter && (
-                  <Select_Shadcn_ value={filter} onValueChange={(val) => setFilter(val as Filter)}>
+                  <Select_Shadcn_
+                    value={filter}
+                    onValueChange={(val) => {
+                      setFilter(val as Filter)
+                      sendEvent({
+                        action: 'auth_users_searched',
```
seanoliver:
```suggestion
                        action: 'auth_users_search_submitted',
```

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -331,8 +347,18 @@ export const UsersV2 = () => {
                   onChange={(e) => setSearch(e.target.value)}
                   onKeyDown={(e) => {
                     if (e.code === 'Enter' || e.code === 'NumpadEnter') {
-                      setSearch(search.trim())
-                      setFilterKeywords(search.trim().toLocaleLowerCase())
+                      const query = search.trim()
+                      setSearch(query)
+                      setFilterKeywords(query.toLocaleLowerCase())
+                      sendEvent({
+                        action: 'auth_users_searched',
```
seanoliver:
```suggestion
                        action: 'auth_users_search_submitted',
```

seanoliver (changes requested):
One last thing — we’ve been trying to keep our event naming verbs consistent across the codebase, and “submitted” is the pattern we use for user-initiated actions. I missed this on my first pass — sorry about that.

I went ahead and updated the references to `AuthUsersSearchSubmittedEvent` to align with the convention. Let me know if this causes any issues on your side.

cc @pamelachia for visibility

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
fadymak: Good call! I'll update the code accordingly 👌

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
fadymak: Good catch! Fixed ✅

fadymak:
Thanks for the feedback @seanoliver — really helpful! 💚 I believe I've applied all the suggestions, let me know if I missed anything!

One small note: I've added an additional field `filter_column` which was part of the recent refactor/changes to this page.

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -292,7 +309,15 @@ export const UsersV2 = () => {
                   setSearch={setSearch}
                   setFilterKeywords={(s) => {
                     setFilterKeywords(s)
-                    setSelectedUser(undefined)
```
ivasilov, line 295: I'm guessing this was accidentally removed? 

It causes a bug where if you open a user info and then search, if you have zero results, it shows empty side panel.

ivasilov (approved):

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
fadymak: Good catch! This was indeed unintentional during the rebase 🙇

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -293,6 +310,15 @@ export const UsersV2 = () => {
                   setFilterKeywords={(s) => {
                     setFilterKeywords(s)
                     setSelectedUser(undefined)
+                    sendEvent({
+                      action: 'auth_users_search_submitted',
+                      properties: {
+                        trigger: 'search_input',
+                        ...telemetryProps,
+                        keywords: search.trim().toLocaleLowerCase(),
```
seanoliver: I might recommend just sing `s` here since `search` could be stale due to async state updates and you're already normalizing the value you pass into `setFilterKeywrods` upstream in the `UsersSearch` component:

```suggestion
                        keywords: s,
```

seanoliver (approved):
LGTM! Just one minor suggestion!

PR #39410, opened 2025-10-09
refactor,tests(support form)

charislam:
Refactor support form to make it easier to maintain:
- Split up large components into smaller components and hooks
- Lift state up so we don't have to do complex child/parent
state-syncing via useEffect
- Use nuqs parsing for consistent serialization/deserialization of
support form prefilled fields

Add comprehensive tests for support form

## Testing

Test by submitting a support form (maybe try different variations of fields and attachments).
Support forms submitted on preview sites go to the same Front instance, so make sure to clearly mark the ticket as testing and submit from a free account (to not give our support team a heart attack, guilty of this the first time 😬)


apps/studio/lib/helpers.ts
```diff
@@ -348,3 +348,5 @@ export const cleanPointerEventsNoneOnBody = (timeoutMs: number = 300) => {
     }, timeoutMs)
   }
 }
+
+export function neverGuard(_: never): any {}
```
joshenlim, line 352: just curious what this does? can see that it's being used in SupportForm.state.ts

apps/studio/components/interfaces/Support/SupportFormV2.tsx
```diff
@@ -1,297 +1,91 @@
...
       library:
-        values.category === 'Problem' && selectedLibrary !== undefined ? selectedLibrary.key : '',
-      message: formatMessage(values.message, attachments, error),
+        values.category === SupportCategories.PROBLEM && selectedLibrary !== undefined
+          ? selectedLibrary.key
+          : '',
+      message: formatMessage(values.message, attachments, initialError),
       verified: true,
       tags: ['dashboard-support-form'],
       siteUrl: '',
```
joshenlim, line 95: i couldn't leave a comment on `affectedServices` but this is just a tiny nit

given that affected services field is hidden if the support category selected is login issues or plan upgrade
we could default that to an empty string here in those scenarios
image

similar to support_access_categories

apps/studio/components/interfaces/Support/AttachmentUpload.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { toast } from 'sonner'
+// End of third-party imports
+
+import { uuidv4 } from 'lib/helpers'
+import { cn } from 'ui'
+import { createSupportStorageClient } from './support-storage-client'
+
+const MAX_ATTACHMENTS = 5
+
+export const uploadAttachments = async (ref: string, files: File[]) => {
```
joshenlim: can remove `export` since its not being used else where

apps/studio/components/interfaces/Support/SubmitButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+      >
+        Send support request
+      </Button>
+      <div className="flex flex-col items-end gap-1">
+        <div className="space-x-1 text-xs">
+          <span className="text-foreground-light">We will contact you at</span>
+          <span className="text-foreground font-medium">{userEmail}</span>
+        </div>
+        <span className="text-foreground-light text-xs">
+          Please ensure emails from supabase.io are allowed
```
joshenlim: i'm actually wondering if it's still `io` or `com` now haha

joshenlim: yup okay can confirm its .com now

apps/studio/components/interfaces/Support/DocsSuggestions.tsx
joshenlim, line 1: nit but this file could just be DocsSuggestions.tsx?

apps/studio/components/interfaces/Support/SupportFormPage.tsx
```diff
@@ -0,0 +1,249 @@
...
+    Sentry.captureMessage(`Failed to submit Support Form: ${curr.message}`)
+    dispatch({ type: 'RETURN_TO_EDITING' })
+  })
+
+  return (
+    <SupportFormWrapper>
+      <SupportFormHeader />
+      {!!projectRef && !!orgSlug && (
+        <AIAssistantOption projectRef={projectRef} organizationSlug={orgSlug} />
+      )}
```
joshenlim: any reason why we're conditionally rendering this? changing the organization causes this component to re-render

joshenlim:
Incredibly tiny nit sorry 😅 some padding differences

## Before
image

## After
image

joshenlim:
Note for no project selected is missing

Ahh its cause we consolidated both plan expectation and no project selected into one component - will split em out

## Before
image

## After
image

joshenlim:
Will change the text here to an admonition

## Before
image

## After
image

apps/studio/components/interfaces/Support/ProjectAndPlanInfo.tsx
```diff
@@ -0,0 +1,243 @@
...
+          )}
+
+          <div className="flex flex-col sm:flex-row items-start sm:items-center gap-y-2 sm:gap-x-2">
+            <Button asChild>
+              <Link
+                href={`/org/${orgSlug}/billing?panel=subscriptionPlan&source=planSupportExpectationInfoBox`}
+              >
+                Upgrade project
+              </Link>
+            </Button>
```
joshenlim: needs to be behind the billing:all enabled features flag

joshenlim: actually including the enquire about enterprise too

joshenlim:
Tiny padding issue here:
image

apps/studio/components/interfaces/Support/SupportFormV2.tsx
```diff
@@ -1,990 +1,171 @@
...
-    state.status === 'partialResults' ||
-    (state.status === 'loading' && state.staleResults.length > 0)
-
-  const onFilesUpload = async (event: ChangeEvent<HTMLInputElement>) => {
-    event.persist()
-    const items = event.target.files || (event as any).dataTransfer.items
-    const itemsCopied = Array.prototype.map.call(items, (item) => item) as File[]
-    const itemsToBeUploaded = itemsCopied.slice(0, MAX_ATTACHMENTS - uploadedFiles.length)
+  const onSubmit: SubmitHandler<SupportFormValues> = async (values) => {
+    const attachments = await attachmentUpload.createAttachments(projectRef)
```
joshenlim, line 75: submit button doesnt have a loading state while files are being uploaded

joshenlim: ahhh i think the dispatch event just needs to be shifted upwards

joshenlim (commented):
Tested on staging preview and mostly LGTM except one odd bug that I'm running into in the staging preview but not locally

For preflling the form with URL params, somehow for `category`, `dashboard_bug` and `abuse` do not automatically prefill for me but all other values for `category` seems to be working 🤔

apps/studio/lib/helpers.ts
charislam: It's a way to get type safety on switch statements so the switch statement is guaranteed to be exhaustive. If anything falls through to the default case, then it's passed through as the arg to neverGuard, which means it has to match type `never`, but nothing matches `never`, so it'll error

CleanShot 2025-10-10 at 13 58 44

charislam:
> dashboard_bug and abuse do not automatically prefill for me

fixed! was a case sensitivity thing

also pushed up a few minor clean-ups and fixes

joshenlim (approved):
Verified changes on staging preview + verified testing submitting ticket on preview

PR #39414, opened 2025-10-09
fix(pgmeta): return pks in order

charislam:
to match pg-meta pr here: https://github.com/supabase/postgres-meta/pull/996


joshenlim (approved):
LGTM - i dont think dashboard is using this atm actually so all good

PR #39424, opened 2025-10-10
Remove has configured auth step

SaxonF:
By default authentication (sign up/in) is enabled on new projects. In our new getting started section we were showing a step for enabling which could lead to confusion. Reduce steps by removing auth check and lean more into the sign up user step.


joshenlim (approved):
Code changes LGTM!

PR #39425, opened 2025-10-10
chore: branding customisation

alaister:
Adds support for larger logos.

Also swaps Next Image to regular `` tags since for svgs there's no benefit.


packages/common/enabled-features/enabled-features.json
```diff
@@ -24,8 +24,11 @@
   "authentication:show_sort_by_email": true,
   "authentication:show_sort_by_phone": true,
   "authentication:show_user_type_filter": true,
+
   "billing:all": true,
 
+  "branding:large_logo": false,
```
alaister, line 30: Default of `false` is correct here.

ivasilov (approved):

PR #39431, opened 2025-10-10
feat(studio): query performance metrics chart

kemaldotearth:

YES

Brings in a chart tracking specific metrics for our Query Performance view.


apps/studio/components/interfaces/QueryPerformance/QueryPerformance.utils.ts
```diff
@@ -24,3 +24,14 @@ export const formatDuration = (milliseconds: number) => {
 
   return parts.length > 0 ? parts.join(' ') : '0s'
 }
+
+export const transformLogsToJSON = (log: string) => {
+  try {
+    let jsonString = log.replace('[pg_stat_monitor] ', '')
+    jsonString = jsonString.replace(/""/g, '","')
```
jordienr, line 31: are we sure this wont break anything? 🧐

apps/studio/components/interfaces/QueryPerformance/QueryPerformance.tsx
```diff
@@ -1,158 +1,54 @@
...
-  )
-
-  const handleRefresh = () => {
-    queryPerformanceQuery.runQuery()
-    queryHitRate.runQuery()
-    queryMetrics.runQuery()
-  }
-
-  const { data: databases } = useReadReplicasQuery({ projectRef: ref })
+  const [isMounted, setIsMounted] = useState(false)
```
jordienr: I don't understand why the mounted state is necessary

kemaldotearth: I had this initially due to it defaulting to `pg_stat_statements` loading, can shore that up/remove.

apps/studio/components/interfaces/QueryPerformance/QueryPerformance.utils.ts
kemaldotearth: Ideally this won't be longterm, but if there's a better approach to this, happy to do it!

jordienr (approved):

PR #39468, opened 2025-10-13
chore: update self-hosted image versions

kiwicopple:


docker/docker-compose.yml
```diff
@@ -343,7 +343,7 @@ services:
 
   analytics:
     container_name: supabase-analytics
-    image: supabase/logflare:1.22.4
```
aantti, line 346: There was an earlier commit here hidden in a different PR (not a chore/update versions one) that changed it to 1.22.6; this can be edited and resolved while merging, though.

docker/docker-compose.yml
```diff
@@ -392,7 +392,7 @@ services:
   # Comment out everything below this point if you are using an external Postgres database
   db:
     container_name: supabase-db
-    image: supabase/postgres:15.8.1.085
+    image: supabase/postgres:17.6.1.017
```
aantti: This looks like a breaking change to me.

aantti (dismissed):
There's a minor merge conflict, and a breaking change of Pg 15 → Pg 17?

docker/docker-compose.yml
sweatybridge: Can you share more context? AFAIK we need to bump the logflare version to 1.23.0+ to fix issues with cron logs.

docker/docker-compose.yml
sweatybridge: Yes, let's revert this until we've figured out the upgrade path for pg17?

docker/docker-compose.yml
aantti: Yes, that'd be fine.

docker/docker-compose.yml
aantti: Agreed.

sweatybridge (approved):

PR #39471, opened 2025-10-13
perf: pass org slug / project ref to resource warnings endpoint

kevcodez:
Pass in project ref / org slug to ensure we filter down and not query across all orgs unnecessarily

Prod deploy is still outstanding for the query params to be accepted by the backend, but it is safe to merge the change either way given it's just additive


apps/studio/components/interfaces/Database/Replication/DestinationPanel.tsx
```diff
@@ -123,20 +123,24 @@ export const DestinationPanel = ({
     pipelineId: existingDestination?.pipelineId,
   })
 
-  const defaultValues = useMemo(
-    () => ({
+  const defaultValues = useMemo(() => {
+    const bigQueryConfig =
```
kevcodez, line 127: Fixed TS errors due to new spec

apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureActivity.tsx
```diff
@@ -59,7 +59,7 @@ export const InfrastructureActivity = () => {
   })
   const isFreePlan = organization?.plan?.id === 'free'
 
-  const { data: resourceWarnings } = useResourceWarningsQuery()
+  const { data: resourceWarnings } = useResourceWarningsQuery({ ref: projectRef })
   const projectResourceWarnings = resourceWarnings?.find((x) => x.project === projectRef)
```
joshenlim, line 64: guessing this line can be cleaned up once the BE changes are on prod?

kevcodez: yep, wanted to keep as is to ensure backwards comp

apps/studio/data/usage/keys.ts
```diff
@@ -2,5 +2,6 @@ export const usageKeys = {
   usage: (projectRef: string | undefined) => ['projects', projectRef, 'usage'] as const,
   orgUsage: (orgSlug: string | undefined, projectRef?: string, start?: string, end?: string) =>
     ['organizations', orgSlug, 'usage', projectRef, start, end] as const,
-  resourceWarnings: () => ['projects', 'resource-warnings'] as const,
+  resourceWarnings: (slug?: string, projectRef?: string) =>
+    ['projects', 'resource-warnings', slug ?? '-', projectRef ?? '-'] as const,
```
joshenlim: this is just a nit but reckon the key here could be an object for clarity - since both params are optional

```['projects', 'resource-warnings', { slug, ref }]```

File: apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx [hunk 4]
```diff
@@ -86,7 +86,7 @@ export const ProjectList = ({ organization: organization_, rewriteHref }: Projec
     isError: isErrorPermissions,
     error: permissionsError,
   } = usePermissionsQuery()
-  const { data: resourceWarnings } = useResourceWarningsQuery()
+  const { data: resourceWarnings } = useResourceWarningsQuery({ slug })
```
joshenlim, line 89: possible refactor RE optimization (okay to do separately)

but given that this is the only file that's calling resource warnings with `slug`
this logic could probably be shifted into `ProjectTableRow` and `ProjectCard` with each component then calling `useResourceWarningsQuery` with `ref` instead

So other parts of the UI that are calling `useResourceWarningsQuery` with `ref` can just access the cached data from react query

joshenlim: nope cancel this - as you mentioned would cause a barrage of requests for orgs with many projects

joshenlim (approved):
verified changes on staging preview!

PR #39472, opened 2025-10-13
AI model selection

SaxonF:
image

This adds the ability to switch between gpt-5 and gpt-5-mini within the Assistant. There are a couple of reasons for this:
- Transparency around which models power responses
- Switching to mini for speed purposes
- Opens up ability to offer more models in future

To test:
- Test model selection on pro plan
- Test model selection free plan and ensure you can't switch to gpt-5


apps/studio/state/ai-assistant-state.tsx
```diff
@@ -233,6 +239,7 @@ export const createAiAssistantState = (): AiAssistantState => {
 
   const state: AiAssistantState = proxy({
     ...initialState, // Spread initial values directly
+    model: initialState.model,
```
charislam:
```suggestion
```

Shouldn't this already be present from the `initialState` spread?

apps/studio/pages/api/ai/sql/generate-v4.ts
```diff
@@ -79,7 +80,14 @@ async function handlePost(req: NextApiRequest, res: NextApiResponse) {
...
   }
 
-  const { messages: rawMessages, projectRef, connectionString, orgSlug, chatName } = data
+  const {
+    messages: rawMessages,
+    projectRef,
+    connectionString,
+    orgSlug,
+    chatName,
+    model: requestedModel,
```
charislam, line 89: I don't know if we care about this, but even though the model selection is gated on plan, we have no checks to enforce it

apps/studio/components/ui/AIAssistantPanel/ModelSelector.tsx
```diff
@@ -0,0 +1,103 @@
...
+export const ModelSelector = ({ selectedModel, onSelectModel }: ModelSelectorProps) => {
+  const router = useRouter()
+  const { data: organization } = useSelectedOrganizationQuery()
+
+  const [open, setOpen] = useState(false)
+
+  const canAccessProModels = organization?.plan?.id !== 'free'
+  const slug = organization?.slug ?? '_'
+
+  const upgradeHref = `/org/${slug ?? '_'}/billing?panel=costControl&source=ai-assistant-model`
```
charislam:
```suggestion
  const upgradeHref = `/org/${slug ?? '_'}/billing?source=ai-assistant-model`
```

This opens the cost control side panel, whereas the upgrade subscription plan button is on the main page

apps/studio/pages/api/ai/sql/generate-v4.ts
SaxonF: isLimited will/should revert to mini if true

apps/studio/components/ui/AIAssistantPanel/ModelSelector.tsx
SaxonF:
```suggestion
  const upgradeHref = `/org/${slug ?? '_'}/billing?panel=subscriptionPlan&source=ai-assistant-model`
```

Ah the panel should be subscriptionPlan

charislam (approved):

PR #39475, opened 2025-10-13
Refactor methods from projects query

joshenlim:
## Context

Given that we're looking into deprecating `projects-query` (which is the unpaginated projects endpoint), changes here just aims to remove as much usage of methods from `projects-query` so that we can just delete that file once we've fully moved off from that endpoint

## Changes involved
- Shifted `setProjectStatus` and `setProjectPostgrestStatus` into `project-detail-query`
  - Opt to convert all these `setXXX` functions into hooks so that client components do not need to import `useQueryClient` just to pass it back into those functions
  - `setProjectStatus` will update the query data from  `org-projects-infinite-query`, `projects-infinite-query`, `project-details-query`, and `projects-query` -> The last one we can clean up once `projects-query` is fully deprecated
- Shift `invalidateProjectsQuery` into `org-projects-infinite-query`
  - Will invalidate keys from `org-projects-infinite-query`, `projects-infinite-query`, and `projects-query` -> The last one we can clean up once `projects-query` is fully deprecated
- Deprecate `org-projects` query in favor of `org-projects-infinite-query`
  - `DiskUsage.tsx` was the only file using the former query, but these 2 are pinging the same endpoint, so just swapped to use the latter query in `DiskUsage.tsx`


apps/studio/components/interfaces/Organization/Usage/UsageSection/DiskUsage.tsx
```diff
@@ -72,7 +63,8 @@ const DiskUsage = ({
           })
           .filter((it) => it.ref === projectRef || !projectRef)
       : []
-  }, [diskUsage, projectRef])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isSuccess, projectRef])
```
alaister: Should `projects` also be in this array?
```suggestion
  }, [isSuccess, projects, projectRef])
```

apps/studio/components/interfaces/Organization/Usage/UsageSection/DiskUsage.tsx
```diff
@@ -31,32 +31,23 @@ export interface DiskUsageProps {
...
-      orgSlug: slug,
-    },
-    {
-      enabled: currentBillingCycleSelected,
-    }
+  const { data, isError, isLoading, isSuccess, error } = useOrgProjectsInfiniteQuery(
+    { slug },
+    { enabled: currentBillingCycleSelected }
   )
+  const projects = useMemo(() => data?.pages.flatMap((page) => page.projects), [data?.pages]) || []
```
alaister: nit:
```suggestion
  const projects = useMemo(() => data?.pages.flatMap((page) => page.projects) || [], [data?.pages])
```

alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/interfaces/Organization/Usage/UsageSection/DiskUsage.tsx
joshenlim: oh yup good point!

apps/studio/components/interfaces/Organization/Usage/UsageSection/DiskUsage.tsx
joshenlim: oh great catch haha

PR #39483, opened 2025-10-13
Auth Overview fixes

jordienr:
- removes arrows from % ✅ 
- links to report with 24h filter ✅ 
- cleans old code ✅ 
- adds tables with common errors. ✅ 
- 
CleanShot 2025-10-13 at 17 14 20@2x


kemaldotearth (approved):
LGTM! Finally a page we can experiment with :)

PR #39486, opened 2025-10-13
chore: Migrate eslint for all apps to use flat config

ivasilov:
This PR migrates all eslint rules to use the new flag config which is mandatory in eslint v9. Nextjs has also deprecated their `next lint` command in v16 so that's another reason to migrate.

I've also turned on warnings for `reportUnusedDisableDirectives` which will flag all eslint overrides that aren't working.


charislam (approved):
tested linting locally on all apps

fsansalvadore (approved):
Looks good 👍

PR #39497, opened 2025-10-13
fix: omit quotes in values when `.env` contents are pasted into input

johnstonmatt:
Fixes FE-1941

YES

Bug Fix

The current behavior when pasting the contents of a `.env` file into the Functions Secrets Dashboard GUI is to split each line into a key/value pair. The value component of the pasted content is sometimes wrapped in `"` or `'` and in these cases they are preserved as a part of the value. This is not in-line with how most systems handle `.env` files, where quotes at the beginning and end are typically stripped from the values.

This would be fine except for the fact that the strings are also hidden from the user so they may not ever notice the error.

## What is the new behavior?

The new behavior is to strip the `"` and `'` characters if they exist in each value string after it has been trimmed of whitespace.

## Additional context


joshenlim (approved):
LGTM! Verified on staging preview 🙂🙏

PR #39500, opened 2025-10-14
feat(studio): list existing analytics buckets

dnywh:

- Feature
- Resolves DEPR-128

Users with `isNewStorageUIEnabled` set to `true` can only see the empty state for analytics storage.

## What is the new behavior?

Users with `isNewStorageUIEnabled` set to `true` can now:
- [x] View existing analytics buckets instead of the empty state, if applicable
- [x] Open the bucket creation dialog (with a hardcoded integration status)
- [x] Open the bucket deletion dialog

## Additional context

See https://github.com/supabase/supabase/pull/39342 for a further-developed version of analytics buckets, which this pull request is working towards. That branch includes:
- Hooks for actually determining wrapper and extension status
- The bucket page itself

## To test

Please test the following:

Please ensure the following specific features are working:

- [ ] The list of buckets immediately updates to include a newly-created bucket (just like file buckets)
  - [ ] Going from 0 buckets to 1 bucket
  - [ ] Going from n buckets to n+1 buckets

The `storageAnalyticsVector` feature flag is already set to `true` for all users on local and staging. You’ll just need to enable to corresponding feature preview.

The trickier thing to test is viewing a list of _existing_ analytics buckets. There are a few ways you can achieve this:

- On local: [hardcode](https://github.com/supabase/infrastructure/blob/rs/storage-temp/api/apps/mgmt-api/src/routes/platform/storage/ref/buckets/buckets.controller.ts#L38) some buckets
- On staging: switch the feature flag off, make some analytics buckets the old way, and then switch the flag back on


apps/studio/components/interfaces/Storage/__tests__/CreateBucketModal.test.tsx
```diff
@@ -54,7 +54,7 @@ describe(`CreateBucketModal`, () => {
       expect(screen.getByRole(`dialog`)).toBeInTheDocument()
     })
 
-    const nameInput = screen.getByLabelText(`Name of bucket`)
+    const nameInput = screen.getByLabelText(`Bucket name`)
```
dnywh, line 57: _Name of bucket_ feels a bit clumsy as a field title. Suggested rephrasing.

apps/studio/components/interfaces/Storage/__tests__/EditBucketModal.test.tsx
```diff
@@ -73,7 +73,7 @@ describe(`EditBucketModal`, () => {
     await userEvent.click(openButton)
     await screen.findByRole(`dialog`)
 
-    const nameInput = screen.getByLabelText(`Name of bucket`)
+    const nameInput = screen.getByLabelText(`Bucket name`)
```
dnywh, line 76: Match above renaming.

apps/studio/components/interfaces/Storage/Storage.constants.ts
dnywh, line 1: I’ve split `label` into `singularName` and `article` because we use them in various ways in language. For example:

> Create a file bucket

> New file bucket

apps/studio/components/layouts/StorageLayout/StorageBucketsLayout.tsx
```diff
@@ -47,7 +47,7 @@ export const StorageBucketsLayout = ({
     <PageLayout
       title={title || (config?.displayName ?? 'Storage')}
       subtitle={
-        hideSubtitle ? config?.description || 'Manage your storage buckets and files.' : null
+        !hideSubtitle ? config?.description || 'Manage your storage buckets and files.' : null
```
dnywh, line 41: This check was the wrong way around

apps/studio/components/interfaces/Storage/AnalyticsBuckets.tsx
```diff
@@ -1,6 +1,185 @@
...
+                            className="text-sm text-foreground-light"
+                          />
+                        </p>
+                      </TableCell>
+
+                      <TableCell>
+                        <div className="flex justify-end gap-2">
+                          <Button asChild type="default">
+                            <Link
+                              href={`/project/${ref}/storage/analytics/buckets/${encodeURIComponent(bucket.id)}`}
```
dnywh, line 126: Note that this route does not exist yet, so will 404. See https://github.com/supabase/supabase/pull/39342 for how it might function

apps/studio/components/interfaces/Storage/AnalyticsBuckets.tsx
```diff
@@ -1,6 +1,185 @@
...
+          )}
+        </ScaffoldSection>
+      )}
+
+      {selectedBucket && (
+        <>
+          <DeleteBucketModal
+            visible={modal === `delete`}
+            bucket={selectedBucket}
+            onClose={() => setModal(null)}
```
dnywh: No bucket deletion functionality added yet (although it is welcome!)

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
```diff
@@ -264,7 +267,7 @@ export const CreateBucketModal = ({
 
       <DialogContent>
         <DialogHeader>
-          <DialogTitle>Create storage bucket</DialogTitle>
+          <DialogTitle>Create {isStorageV2 ? config.singularName : 'storage'} bucket</DialogTitle>
```
dnywh: This dialog is currently used in prod in these two ways:
- To create a file bucket
- To create a storage bucket

But in `isStorageV2`, we only use it for file buckets. Perhaps this check is overkill?

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
```diff
@@ -279,8 +282,8 @@ export const CreateBucketModal = ({
                 render={({ field }) => (
                   <FormItemLayout
                     name="name"
-                    label="Name of bucket"
-                    labelOptional="Buckets cannot be renamed once created."
+                    label="Bucket name"
+                    labelOptional="Cannot be changed after creation"
```
dnywh, line 286: Just some gentle copywriting improvements, as mentioned in the corresponding tests files above.

apps/studio/components/interfaces/Storage/CreateSpecializedBucketModal.tsx
dnywh, line 1: I‘ve decided that analytics and vector bucket types will get their own (probably shared) “create bucket” modal  for these reasons:
- The create _file_ bucket is has quite a few file-specific options
  - Public or private
  - Restrict file size
  - Restrict MIME types
- Analytics and vector buckets have very similar setups
  - Bucket name
  - A check for relevant extensions and integrations

apps/studio/components/interfaces/Storage/CreateSpecializedBucketModal.tsx
```diff
@@ -0,0 +1,265 @@
...
+                      target="_blank"
+                      rel="noreferrer"
+                      className="text-foreground-lighter hover:text-foreground transition-colors"
+                    >
+                      Learn more
+                    </InlineLink>
+                  </p>
+                </Admonition>
+              </DialogSection>
+            )}
```
dnywh: Hardcoded for now. See https://github.com/supabase/supabase/pull/39342 for a fancy hook that checks for the installation status of either and renders text accordingly.

apps/studio/components/interfaces/Storage/EmptyBucketState.tsx
```diff
@@ -10,19 +11,35 @@ export const EmptyBucketState = ({ bucketType }: EmptyBucketStateProps) => {
...
+            bucketType === 'vectors'
+              ? {
+                  content: {
+                    side: 'bottom',
+                    text: 'Vector buckets are not supported yet',
+                  },
+                }
+              : undefined
+          }
+        />
```
dnywh, line 42: Assuming we use `CreateSpecializedBucketModal` for both analytics and vectors, we can just simplify the ternary and set set `bucketType={bucketType}`.

Perhaps I‘m jumping the gun here and we should just stick to a `&&` to render each.

apps/studio/components/interfaces/Storage/EmptyBucketState.tsx
```diff
@@ -10,19 +11,35 @@ export const EmptyBucketState = ({ bucketType }: EmptyBucketStateProps) => {
...
-          <h3>Create {config.label}</h3>
-          <p className="text-foreground-light text-sm">{config.valueProp}</p>
-        </div>
+    <aside className="mt-12 border border-dashed w-full bg-surface-100 rounded-lg px-4 py-10 flex flex-col gap-y-3 items-center text-center gap-1 text-balance">
+      <BucketAdd size={24} strokeWidth={1.5} className="text-foreground-muted" />
+      <div className="flex flex-col items-center text-center">
+        <h3>
+          Create {config.article} {config.singularName} bucket
+        </h3>
+        <p className="text-foreground-light text-sm">{config.valueProp}</p>
```
dnywh, line 20: Just some spacing and markup cleanup. Visually very similar.

apps/studio/components/interfaces/Storage/CreateBucketModal.tsx
joshenlim: this is alright i think

apps/studio/components/interfaces/Storage/AnalyticsBuckets.tsx
joshenlim: fwiw im guessing this already works actually? i dont think there's a different delete endpoint for analytics buckets specifically (i can't verify myself since i'm not able to properly create new analytics buckets on staging, but it does look like this should already work)

apps/studio/components/interfaces/Storage/CreateSpecializedBucketModal.tsx
```diff
@@ -0,0 +1,265 @@
...
+            {bucketType === 'analytics' && (
+              <DialogSection className="space-y-3">
+                <Admonition type="default">
+                  <p>
+                    Supabase will install the Wrappers extension and Iceberg Wrapper integration on
+                    your behalf.{' '}
+                    <InlineLink
+                      href={`${DOCS_URL}/guides/database/extensions/wrappers/iceberg`}
+                      target="_blank"
+                      rel="noreferrer"
```
joshenlim: these are already handled within InlineLink so no need to pass them

apps/studio/components/interfaces/Storage/CreateSpecializedBucketModal.tsx
```diff
@@ -0,0 +1,265 @@
...
+                      className="text-foreground-lighter hover:text-foreground transition-colors"
```
joshenlim: just to check if we'd want this styling to be used for InlineLink in general?

ideally we keep the styling consistent whereever we use the InlineLink component i reckon

apps/studio/components/interfaces/Storage/CreateSpecializedBucketModal.tsx
joshenlim, line 1: there's a number of unused variables here which i think as you mentioned in the slack thread that we need to figure out how to properly create the analytics buckets here yeah?

joshenlim (approved):
Code wise LGTM - although JFYI i'm not able to test this due to some blockers on the BE side of things. Approving to unblock FE progress nonetheless, we can do an E2E test at the end 🙂

apps/studio/components/interfaces/Storage/AnalyticsBuckets.tsx
dnywh: I have some on prod I can test 🫠

apps/studio/components/interfaces/Storage/CreateSpecializedBucketModal.tsx
dnywh: Oh duh!

apps/studio/components/interfaces/Storage/CreateSpecializedBucketModal.tsx
dnywh: Agreed. We should try and avoid overrides because that leads to design debt.

Right now the `InlineLink` component [hardcodes](https://github.com/supabase/supabase/blob/dnywh/feat/analytics-bucket-basics/apps/studio/components/ui/InlineLink.tsx#L13-L14) `text-foreground`. This is why I added the `text-foreground-lighter` (since that’s what the surrounding paragraph is, at least in future PRs). Otherwise it’s weirdly prominent.

Anyway, what do you think of me changing that (in `InlineLink`) to `text-inherit`?

PR #39506, opened 2025-10-14
Drag query to home

SaxonF:
image

Adds ability to drag queries from Assistant into the reports section of new home.


apps/studio/components/ui/AIAssistantPanel/DisplayBlockRenderer.tsx
```diff
@@ -85,7 +85,9 @@ export const DisplayBlockRenderer = ({
   const [rows, setRows] = useState<any[] | undefined>(
     Array.isArray(initialResults) ? initialResults : undefined
   )
-  const isDraggableToReports = canCreateSQLSnippet && router.pathname.endsWith('/reports/[id]')
+  const isReportsPage = router.pathname.endsWith('/reports/[id]')
+  const isHomePage = router.pathname === '/project/[ref]'
```
charislam, line 89: Do we also need to do a check that they have the new home page version? (Is that still feature-flagged?)

charislam (approved):
Tested on preview

PR #39507, opened 2025-10-14
chore(studio): consistent table rows across pages

dnywh:
## What kind of change does this PR introduce?

- UI cleanup
- Fixes DEPR-119

## What is the current behavior?

- The [S3 Configuration](https://supabase.com/dashboard/project/_/storage/s3) page is using a deprecated table components
  - It’s visually out of date with other table instances on Studio
  - This inconsistency will be exacerbated by incoming Storage sibling pages, which also deal with S3 configuration (or at least present data in a similar way
- The [API Keys](https://supabase.com/dashboard/project/_/settings/api-keys/new) page is similarly out of date with table components and incoming similar Storage UI

## What is the new behavior?

- The S3 Configuration table uses the latest table components
- The API Keys page is updated to match both the latest table components and how we’re presenting secrets in Storage

| Before | After |
| --- | --- |
| S3 Configuration  Supabase | S3 Configuration  Supabase | 
| Settings  Supabase | Settings  Supabase | 

## To test

Please test the following at various breakpoints:

- [ ] Viewing [secret keys](https://supabase.com/dashboard/project/_/settings/api-keys/new)
- [ ] Viewing S3 Configuration [access keys](https://supabase.com/dashboard/project/_/storage/s3)
  - [ ] Viewing 1+ items and their values shown in the table

PR #39509, opened 2025-10-14
chore(studio): consistent copy/reveal input behaviour

dnywh:

- UI improvement
- Design documentation

- We mix-and-match how we handle read-only fields and values

## What is the new behavior?

- The `Input` component now has dedicated `read-only:*` styles
- `Input` instances that are truely “read only” have `disabled` removed so their text is more easily selectable
- S3 access keys use this new pattern instead of their bespoke one

| Before | After |
| --- | --- |
| S3 Configuration  Supabase | S3 Configuration  Supabase | 

## Additional context

You might be wondering: isn’t weird to have an input inside a table row?

I think that’s more a stylistic concern than a semantic one, hence the first-pass at styling the `Input` component.

This PR is part of a larger effort to tighten how we present read-only values and tables. It comes after https://github.com/supabase/supabase/pull/39507 and before https://github.com/supabase/supabase/pull/39511.

That last one especially will have this effort make sense; we are using input-esque UI to show values, so we might as well consolidate.

## To test

- [ ] View S3 Configuration [access keys](https://supabase.com/dashboard/project/_/storage/s3) with 1+ key
  - [ ] Copy the value with one click
  - [ ] Select the text easily


apps/studio/components/interfaces/Settings/API/ServiceList.tsx
```diff
@@ -95,7 +95,7 @@ export const ServiceList = () => {
                         : 'RESTful endpoint for querying and managing your database'
                   }
                 >
-                  <Input copy readOnly disabled className="input-mono" value={endpoint} />
```
dnywh, line 98: I assume the `disabled` prop was passed here (and elsewhere) to achieve the _look_ of a disabled input.

I’ve instead added `read-only` specific styles to the Input component’s definition. This is important for ergonomics. I want to be able to select text of a read-only input. Disabled inputs make that (appear) harder.

apps/studio/components/interfaces/Storage/StorageSettings/StorageCredItem.tsx
```diff
@@ -50,12 +49,7 @@ export const StorageCredItem = ({
         <span className="text-foreground">{description}</span>
       </td>
       <td>
-        <div className="flex items-center justify-between">
-          <span className="text-ellipsis font-mono cursor-default">{access_key}</span>
-          <span className="w-24 text-right opacity-0 group-hover:opacity-100 transition-opacity">
-            <CopyButton text={access_key} type="default" />
-          </span>
-        </div>
+        <Input readOnly copy value={access_key} className="font-mono" />
```
dnywh, line 54: Replacing a bespoke input (or value, really) with what we’re moving towards everywhere for anything that has a copy-able (and optional reveal) button.

apps/studio/components/ui/ProjectSettings/DisplayApiSettings.tsx
```diff
@@ -10,7 +10,8 @@ import Panel from 'components/ui/Panel'
 import { useJwtSecretUpdatingStatusQuery } from 'data/config/jwt-secret-updating-status-query'
 import { useProjectSettingsV2Query } from 'data/config/project-settings-v2-query'
 import { useAsyncCheckPermissions } from 'hooks/misc/useCheckPermissions'
-import { Input } from 'ui'
+import { Input } from 'ui-patterns/DataInputs/Input'
+import { FormLayout } from 'ui-patterns/form/Layout/FormLayout'
```
dnywh, line 14: Using the newer version of Input here on the (legacy) API keys view.

This is important for achieving the same behaviour has elsewhere.

packages/ui-patterns/src/DataInputs/Input.tsx
```diff
@@ -71,6 +71,7 @@ const Input = forwardRef<
           {...props}
           onCopy={onCopy}
           value={reveal && hidden ? HIDDEN_PLACEHOLDER : props.value}
+          disabled={reveal && hidden ? true : props.disabled}
```
dnywh, line 74: This prop change means that _before_ reveal the inputs contents are disabled, and _after_ reveal they are not.

packages/ui/src/components/shadcn/ui/input.tsx
```diff
@@ -9,8 +9,8 @@ export interface InputProps
 
 export const InputVariants = cva(
   cn(
-    'flex h-10 w-full rounded-md border border-control bg-foreground/[.026] px-3 py-2 text-sm file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-foreground-muted',
-    'focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-background-control focus-visible:ring-offset-2 focus-visible:ring-offset-foreground-muted disabled:cursor-not-allowed disabled:opacity-50',
+    'flex h-10 w-full rounded-md border border-control read-only:border-secondary bg-foreground/[.026] px-3 py-2 text-sm file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-foreground-muted read-only:text-foreground-lighter/75',
+    'focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-background-control focus-visible:ring-offset-2 focus-visible:ring-offset-foreground-muted disabled:cursor-not-allowed disabled:text-foreground-muted',
```
dnywh: This is the heart of it: `read-only`-specific styles that make us not rely on `disabled` as a crutch.

apps/studio/components/interfaces/Settings/API/ServiceList.tsx
```diff
@@ -95,7 +95,7 @@ export const ServiceList = () => {
...
+                  <Input copy readOnly className="font-mono" value={endpoint} />
```
dnywh, line 98: `input-mono` seems to just be custom class wrapper on `font-mono`.

apps/design-system/__registry__/index.tsx
dnywh, line 1: A few examples in the design system

apps/design-system/config/docs.ts
```diff
@@ -117,6 +117,11 @@ export const docsConfig: DocsConfig = {
           href: '/docs/fragments/inner-side-menu',
           items: [],
         },
+        {
+          title: 'Data Input',
+          href: '/docs/fragments/data-input',
+          items: [],
+        },
```
dnywh, line 124: I’ve documented this component under _Fragments_ since:

1. That’s what it is.
2. We already have (confusingly) an Input component under the Components category.

We should obviously consolidate, merge, or clarify the difference. But one step at a time.

dnywh: This confusion also why I’m referring to this one as _Data_ Input. At least for now.

apps/design-system/registry/default/example/data-input-with-copy-secret.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+  return (
+    <Input
+      containerClassName="w-full max-w-sm"
+      readOnly
+      copy
+      value={maskedValue}
+      onCopy={() => {
+        navigator.clipboard.writeText(actualValue)
+      }}
```
dnywh, line 15: This callback allows for custom values to be copied, not what the actual value is. That’s really useful here, where we want to copy the `actualValue` even when `maskedValue` is what’s shown.

API keys etc make use of this.

apps/design-system/README.md
dnywh, line 1: Just some general README improvements for future design system contributors (i.e. future notes to self).

apps/studio/components/interfaces/Storage/StorageSettings/S3Connection.tsx
```diff
@@ -167,11 +167,19 @@ export const S3Connection = () => {
...
-                        <Input className="input-mono" copy disabled value={project?.region} />
+                        <Input
+                          readOnly
+                          copy
+                          value={project?.region}
+                          data-1p-ignore
+                          data-lpignore="true"
+                          data-form-type="other"
+                          data-bwignore
+                        />
```
dnywh, line 187: This was getting the pesky password manager widgets on it.

joshenlim (approved):
Approving to unblock, just a couple of non-blocking feedback:

- Wondering if we should just use `text-foreground` for the read only input here to make it look "less disabled"
image

- We probably need to investigate our light mode color pallette imo, the contrast is really different from dark mode, e.g here i can barely notice the background color of the input field compared to the card
image

- Not just that but the background color of cards as well perhaps, within a modal i then realised that the inputs actually have borders on them (Nothing sensitive in the screenshot below, its from a local project anyways)
image

dnywh:
> Approving to unblock, just a couple of non-blocking feedback:

Thanks Joshen! For anyone following along, I’ve made changes in https://github.com/supabase/supabase/pull/39509/commits/fec4fdffa5ffccaed3719b6ce62ad0afafec10c2. You can see screenshots [here](https://supabase.slack.com/archives/C0161K73J1J/p1760574523083489?thread_ts=1760497439.543969&cid=C0161K73J1J).

SaxonF (approved):

PR #39525, opened 2025-10-14
cleanup: consolidate database report

jordienr:
Currently we have 2 report configurations, it was used initially to separate free from paid report, but since then we updated the config to allow metric-based limits (the allowedIn: ['free', 'pro'...] attribute. 

This removes the old config and leaves just 1 report config for Database.


apps/studio/data/reports/database-charts.ts
```diff
@@ -407,13 +203,45 @@ export const getReportAttributesV2: (
...
+      ],
+    },
     {
       id: 'client-connections',
       label: 'Database Connections',
       syncId: 'database-reports',
       valuePrecision: 0,
-      availableIn: ['team', 'enterprise'],
-      hide: false,
+      availableIn: ['free', 'pro', 'team', 'enterprise'],
```
ivasilov: is `free` suppose to be in this array? Other similar attributes don't have it.

ivasilov (approved):

PR #39527, opened 2025-10-14
feat: Redirect the user after claiming a project

ivasilov:
This PR adds a redirect after the user claims a project. The redirect will the match the `redirect_uri` parameter from when the source `/v1/oauth/authorize/project-claim` API call. The redirect URL still needs to math the redirect URI in the oAuth app.

### DIFF
--- a/apps/studio/components/interfaces/Organization/ProjectClaim/confirm.tsx
+++ b/apps/studio/components/interfaces/Organization/ProjectClaim/confirm.tsx
@@ -47,16 +47,22 @@ export const ProjectClaimConfirm = ({
 
   const onClaimProject = async () => {
     try {
-      await approveRequest({ id: auth_id!, slug: selectedOrganization.slug })
+      const response = await approveRequest({ id: auth_id!, slug: selectedOrganization.slug })
       await claimProject({
         slug: selectedOrganization.slug,
         token: claimToken!,
       })
 
       toast.success('Project claimed successfully')
-      // invalidate the org projects to force them to be refetched
-      queryClient.invalidateQueries(projectKeys.list())
-      router.push(`/org/${selectedOrganization.slug}`)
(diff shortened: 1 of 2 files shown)


apps/studio/pages/claim-project.tsx
```diff
@@ -21,7 +21,7 @@ const ClaimProjectPageLayout = ({ children }: PropsWithChildren) => {
   return (
     <>
       <Head>
-        <title>Claim project | {appTitle || 'Supabase'}</title>
+        <title>{`Claim project ${appTitle ?? 'Supabase'}`}</title>
```
charislam:
```suggestion
        <title>{`Claim project | ${appTitle ?? 'Supabase'}`}</title>
```

doesn't make much sense as one sentence I think

charislam (approved):
approving to unblock since code changes look fine, but haven't tested since I'm not sure how to trigger project claims flow

apps/studio/components/interfaces/Organization/ProjectClaim/confirm.tsx
```diff
@@ -47,16 +47,22 @@ export const ProjectClaimConfirm = ({
...
-      router.push(`/org/${selectedOrganization.slug}`)
+      try {
+        // check if the redirect url is valid. If not, redirect the user to the org dashboard
+        const url = new URL(response.url)
+        window.location.href = url.toString()
+      } catch {
+        // invalidate the org projects to force them to be refetched
+        queryClient.invalidateQueries(projectKeys.list())
+        router.push(`/org/${selectedOrganization.slug}`)
+      }
```
phamhieu, line 65: I can’t think of any reason the redirect URL would be invalid, but it’s not a blocker.

phamhieu (approved):

PR #39543, opened 2025-10-15
fix(support form): create api route to upload attachments

charislam:


apps/studio/pages/api/generate-attachment-url.ts
```diff
@@ -0,0 +1,86 @@
...
+
+const GenerateAttachmentUrlSchema = z.object({
+  filenames: z.array(z.string()),
+})
+
+async function handlePost(req: NextApiRequest, res: NextApiResponse) {
+  const { claims, error: userClaimsError } = await getUserClaims(req.headers.authorization!)
+  if (userClaimsError || !claims) {
+    return res.status(401).json({ error: { message: 'Unauthorized' } })
+  }
```
joshenlim, line 18: do we need this check here? `apiWrapper` calls `apiAuthenticate` internally which also does a similar check ([ref](https://github.com/supabase/supabase/blob/master/apps/studio/lib/api/apiAuthenticate.ts#L21))

apps/studio/pages/api/generate-attachment-url.ts
```diff
@@ -0,0 +1,86 @@
...
+import { createClient } from '@supabase/supabase-js'
+import { components } from 'api-types'
+import apiWrapper from 'lib/api/apiWrapper'
+import { getUserClaims } from 'lib/gotrue'
+import z from 'zod'
+
+export const maxDuration = 120
+
+type ProjectsResponse = components['schemas']['ListProjectsPaginatedResponse']
+type Project = ProjectsResponse['projects'][number]
```
joshenlim: doesnt seem to be used atm

apps/studio/components/interfaces/Support/AttachmentUpload.tsx
```diff
@@ -12,12 +15,20 @@ const MAX_ATTACHMENTS = 5
 
 const uploadAttachments = async (ref: string, files: File[]) => {
```
joshenlim: can remove ref param - no longer using

apps/studio/components/interfaces/Support/AttachmentUpload.tsx
```diff
@@ -12,12 +15,20 @@ const MAX_ATTACHMENTS = 5
...
   const supportSupabaseClient = createSupportStorageClient()
+  const userId = await gotrueClient.getSession().then((res) => res?.data.session?.user.id)
+  if (!userId) {
+    console.error(
+      '[Support Form > uploadAttachments] Could not upload attachments because no user ID found'
+    )
+    toast.error('Could not upload attachments')
+    return []
+  }
```
joshenlim: im wondering if we should just let the API handle the prefix so client side doesn't need to check for userId

joshenlim: oh sorry ignore this - understand now, we're uploading files from the client, then calling the API to get the signed URLs, as opposed to generating the signed URLs from the client side

apps/studio/components/interfaces/Support/AttachmentUpload.tsx
```diff
@@ -32,13 +43,25 @@ const uploadAttachments = async (ref: string, files: File[]) => {
...
   if (keys.length === 0) return []
 
-  const { data, error } = await supportSupabaseClient.storage
-    .from('support-attachments')
-    .createSignedUrls(keys, 10 * 365 * 24 * 60 * 60)
-  if (error) {
-    console.error('Failed to retrieve URLs for attachments', error)
+  try {
+    const headers = await constructHeaders()
+    const response = await fetch(`${BASE_PATH}/api/generate-attachment-url`, {
```
joshenlim: will shift this into `data`

joshenlim (approved):
Verified on staging preview

PR #39545, opened 2025-10-15
studio(chore): improve analytics bucket page

dnywh:

- UI update
- Resolves DEPR-129

- Analytics buckets (under the new Storage UI feature preview) have no page route
- The existing analytics bucket page is out of date with latest design work

## What is the new behavior?

- Analytics buckets (under the new Storage UI feature preview) now each have their own dynamic page route
- They reuse the existing analytics bucket interface by @ivandotv 
  - This interface has been given some design love (whilst remaining backwards compatible)

| Before | After: Existing | After: Flag |
| --- | --- | --- |
| Buckets  Supabase | Buckets  Supabase| Buckets  Supabase |
| Buckets  Supabase| Buckets  Supabase | Buckets  Supabase |

## Additional context

I mentioned elsewhere that this PR would duplicate the efforts of `/components/interfaces/Storage/AnalyticBucketDetails/index.tsx`. I've since opted to work _within_ that interface to reduce overhead and redundancy. Backwards compatibility isn’t as difficult here as I thought. 

That said, the approach to namespaces and tables does differ in the new storage feature flag.

**Current approach:** schema and namespace can be different
**New approach:** we force them to be the same

Hence why I have one page section for each, depending on the `isStorageV2` boolean.

## Questions

1. This pull request introduces `storage/analytics/buckets/[bucketId]` as the page route. Is that `/buckets/` path necessary? It is for `files/buckets` because there are other routes that could collide (e.g. `files/policies` and a bucket called `policies`). But not here. What’s better, brevity or consistency?

## To test

Please test the following:

Please ensure the following specific features are working:

- [ ] The correct warning admonition shows based upon the installation state (mocked) of the Wrappers extension and/or Iceberg Wrapper integration
- [ ] Going back to to storage/analytics correctly shows the table of buckets (not an empty state)
- [ ] Deleting a bucket from both `storage/analytics` and `storage/analytics/buckets/[bucketId]` works as expected

Known non-working features:

- Tapping _New table_ doesn’t do anything
- Tables for `isStorageV2` is hardcoded in an empty state.

Both of these require some more infrastructural work that I’ll need to pair with someone on, or handoff entirely.

The `storageAnalyticsVector` feature flag is already set to `true` for all users on local and staging. You’ll just need to enable to corresponding feature preview.

The trickier thing to test is viewing a list of _existing_ analytics buckets. There are a few ways you can achieve this:

- On local: [hardcode](https://github.com/supabase/infrastructure/blob/rs/storage-temp/api/apps/mgmt-api/src/routes/platform/storage/ref/buckets/buckets.controller.ts#L38) some buckets
- On staging: switch the feature flag off, make some analytics buckets the old way, and then switch the flag back on


apps/studio/components/interfaces/Storage/AnalyticBucketDetails/constants.ts
dnywh, line 1: Match capitalisation convention from rest of Studio: only Title Case proper nouns. These ones aren’t, from [what I can see](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html).

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/CopyEnvButton.tsx
```diff
@@ -34,29 +34,14 @@ export const CopyEnvButton = ({
...
-            </span>
-          ),
-        },
-      }}
-    >
-      Copy env values
-    </ButtonTooltip>
+    <Button type="default" loading={isLoading} icon={<Copy />} onClick={onCopy}>
+      Copy all
+    </Button>
```
dnywh, line 45: I don’t think we need a tooltip here, nor a verbose title. They can click and find out! The toast can confirm it.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/DecryptedReadOnlyInput.tsx
dnywh, line 1: Mostly just using the latest componentry. Stylistic cleanup.

The maybe-too-stark input style is fixed in https://github.com/supabase/supabase/pull/39509

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/DecryptedReadOnlyInput.tsx
```diff
@@ -41,52 +42,57 @@ export const DecryptedReadOnlyInput = ({
     : value
 
   return (
-    <Input
-      readOnly
-      // If the value is secure, allow copying to clipboard if the value is revealed. Otherwise, always allow copying
-      copy={!secureEntry || (!isDecryptedValueLoading && showHidden)}
-      disabled
```
dnywh, line 48: We don’t need `disabled` if `readOnly` is the intent. That’s something I’ve cleaned up stylistically in https://github.com/supabase/supabase/pull/39509

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
dnywh, line 1: This file is the guts of the already-in-prod analytics bucket page. As the description of this pull request mentions, I decided to work within in.

The changes in this file can be boiled down to:
- Stylistic improvements
- Support for isStorageV2

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
```diff
@@ -131,6 +137,8 @@ export const AnalyticBucketDetails = ({ bucket }: { bucket: Bucket }) => {
 
   const wrappersExtension = extensionsData?.find((ext) => ext.name === 'wrappers')
 
+  const config = BUCKET_TYPES['analytics']
```
dnywh, line 140: Just to nab the right docsUrl etc. Feels a bit redundant though!

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
```diff
@@ -140,131 +148,227 @@ export const AnalyticBucketDetails = ({ bucket }: { bucket: Bucket }) => {
...
+        breadcrumbs={
+          isStorageV2
+            ? [
+                {
+                  label: 'Analytics',
+                  href: `/project/${project?.ref}/storage/analytics`,
+                },
+              ]
+            : []
+        }
```
dnywh, line 163: Doesn’t make sense to show breadcrumbs on `!isStorageV2` because there is no parent-child relationship. The bucket is on the sidebar.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
```diff
@@ -140,131 +148,227 @@ export const AnalyticBucketDetails = ({ bucket }: { bucket: Bucket }) => {
...
+                            <p className="text-sm text-foreground">No tables yet</p>
+                            <p className="text-sm text-foreground-lighter">
+                              Create an analytics table to get started
+                            </p>
+                          </TableCell>
+                        </TableRow>
+                      </TableBody>
+                    </Table>
+                  </Card>
+                </ScaffoldSection>
```
dnywh, line 226: `isStorageV2` is quite different to the current approach. We are showing _tables_ that are connected to namespaces. Not the other way around.

We need to support both, and this is hard to test. Hence why there is this duplicative table and why it’s hardcoded as empty.

I’d love some support getting the real table data in :)

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
```diff
@@ -283,39 +387,35 @@ const ExtensionNotInstalled = ({
     (wrappersExtension?.default_version ?? '') < (wrapperMeta?.minimumExtensionVersion ?? '')
 
   return (
```
dnywh, line 387: These changes are mainly copywriting (and componentry) simplifications of the warnings so what the user needs to do is clearer.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/SimpleConfigurationDetails.tsx
dnywh, line 1: I find it a bit odd that this file duplicates the same section inside `/interfaces/Storage/AnalyticBucketDetails/index.tsx`, but one step at a time!

apps/studio/components/ui/InlineLink.tsx
dnywh, line 1: Sneaking this one in. What do you think? See discussion https://github.com/supabase/supabase/pull/39500#discussion_r2430843995

dnywh: Happy to leave it for another day!

apps/studio/pages/project/[ref]/storage/analytics/buckets/[bucketId].tsx
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/analytics/buckets/[bucketId].tsx:1]: Turning this page just into a thin wrapper since `apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx` now handles the bucket contents.

apps/studio/components/ui/InlineLink.tsx
joshenlim: oh probably will be good to add a hover:text-foreground i think, to match the decoration-foreground on hover

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/SimpleConfigurationDetails.tsx
joshenlim: oh you mean like a nested ScaffoldSection yeah? we could probably bring them out to be siblings instead of nesting actually - will look into that

joshenlim:
> But not here. What’s better, brevity or consistency?

yeah i reckon unless analytics is gonna have other pages like `analytics/settings` perhaps - can remove `buckets` otherwise. Do we foresee if there's gonna be any pages apart from buckets under analytics?

happy to keep it as is for now

joshenlim:
Running into an issue with deleting analytics bucket - investigating
image

joshenlim (approved):
Verified changes on staging! Deleting analytics buckets is currently not working but its a known issue that will be addressed separately 🙂 🙏

apps/studio/components/ui/InlineLink.tsx
dnywh: Nice!

PR #39548, opened 2025-10-15
fix: support nimbus project urls for testing edge functions

alaister:


joshenlim (approved):
minimally verified that staging preview is status quo

PR #39549, opened 2025-10-15
perf: remove customer endpoint call to determine missing address

kevcodez:
Backend now exposes flag


ecktoteckto (approved):
Tested on Preview

PR #39554, opened 2025-10-15
Chore/simplified support form

joshenlim:
## Context

Opting to show a simplified support form for certain organizations where we'd hide the following fields
- category
- severity
- affected services

image

## To test
- [ ] Opt in staging/local org slugs into config cat with the `simplifiedSupportForm` flag to test
- [ ] Also test for when `simplifiedSupportForm` is empty to verify that this change going to prod shouldn't affect anything


alaister (approved):
LGTM!

- Tested on staging

PR #39559, opened 2025-10-15
fix(feedback): generate image urls on backend

charislam:
Testing: the feedback send endpoint on staging seems to be broken (I tested on master and it also gives an authorization error), but you can check the network requests to make sure the URL is generated and included in the message.


joshenlim (approved):
Verified changes on staging preview for both feedback and support

PR #39567, opened 2025-10-15
Table Editor to hide count for foreign tables

joshenlim:
## Context

Opting to disable + hide count for foreign tables as technically the table will be an unknown size.

## Changes involved

Because the Pagination in the Table Editor depends on the count, am opting to not show `Page x of y` for foreign tables, but instead just `Page x`:
image

So users can just keep paginating to the next page until the row count of the current page is lower than the `rowsPerPage` value. If say the last page's row count matches that of `rowsPerPage`, then we also handle that state with this:
image

## To test

- [ ] Verify that pagination behaviour is status quo for non foreign tables in Table Editor
- [ ] Verify that this new pagination behaviour works as expected for foreign tables in Table Editor
- [ ] Verify that there's no query getting fired to retrieve counts for Foreign Table in network inspector

### DIFF
--- a/apps/studio/components/grid/components/footer/pagination/Pagination.tsx
+++ b/apps/studio/components/grid/components/footer/pagination/Pagination.tsx
@@ -96,7 +96,7 @@ export const Pagination = () => {
 
   // [Joshen] This is only applicable for foreign tables, as we use the number of rows on the page to determine
   // if we've reached the last page (and hence disable the next button)
-  const { data: rowsData } = useTableRowsQuery(
+  const { data: rowsData, isLoading: isLoadingRows } = useTableRowsQuery(
     {
       projectRef: project?.ref,
       connectionString: project?.connectionString,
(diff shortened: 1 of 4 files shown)


apps/studio/components/grid/components/footer/pagination/Pagination.tsx
```diff
@@ -141,11 +186,59 @@ const Pagination = () => {
     if (isError && snap.enforceExactCount && error?.code === 408) {
       snap.setEnforceExactCount(false)
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isError, snap.enforceExactCount, error?.code])
 
+  if (isForeignTableSelected) {
+    return (
+      <div className="flex items-center gap-x-2">
```
joshenlim, line 194: the components here are very similar to the return statement after this, but i opted to keep this separate as nesting a lot of `isForeignTableSelected` checks will make it really hard to read the logic 😅 

e.g the loader should only show if `(!isForeignTableSelected && isLoading) || isForeignTableSelected` -> we'd need to spray a lot of these elsewhere too

saltcod (approved):

saltcod:
Tested with stripe fdw on staging. Count query doesn't fire.

PR #39572, opened 2025-10-15
feat(support form): add commit sha

charislam:
Knowing the commit SHA will help us debug issues by matching them to Git timeline. Added the commit SHA in two places:

1. Automatically as part of the message on support requests
2. Logged within the console, so we can ask the user to double-check when checking if issues persist/are fixed


joshenlim (approved):
Verified SHA shows up in ticket + inspector

PR #39582, opened 2025-10-16
Hide GettingStarted if mature

SaxonF:
The GettingStarted section is primarily an onboarding experience that is relevant to new projects. It includes quite a few requests to gather the necessary information to check whether each task is complete or not. Ideally we have a programmatic, non time based way to determine whether the section is still relevant, but for the time being this PR applies a limit of 10 days so that mature projects will not see it when we launch the new home.


joshenlim (approved):
Verified changes on prod with a project that's > 10 days old, and a project thats < 10 days old

PR #39599, opened 2025-10-16
fix(users search): add explanation for 'all columns' search

charislam:


joshenlim (approved):

PR #39601, opened 2025-10-16
fix(policy editor): reduce number of table role access queries

charislam:
On the Policies page, every single PolicyTableRow fires off its own table roles access query (because each query is keyed to an individual table). For users with many tables (400+), this causes the API to rate-limit the dashboard.

Changed so that the table roles access query now fetches information for the entire schema in one go, then organizes it by table.

## Testing

1. Use the following SQL query to generate 400 tables on your project:
```sql
BEGIN;

DO $$
DECLARE
  i int;
  tbl text;
  policy_name text;
BEGIN
  FOR i IN 1..400 LOOP
    tbl := format('generated_table_%s', lpad(i::text, 3, '0'));
    policy_name := format('allow_all_%s', lpad(i::text, 3, '0'));

    EXECUTE format(
      'CREATE TABLE IF NOT EXISTS %I.%I (
         id bigserial PRIMARY KEY,
         data jsonb,
         created_at timestamptz DEFAULT now()
       )',
      'public', tbl
    );

    EXECUTE format('ALTER TABLE %I.%I ENABLE ROW LEVEL SECURITY', 'public', tbl);


    EXECUTE format(
      'CREATE POLICY %I ON %I.%I FOR ALL USING (true) WITH CHECK (true)',
      policy_name, 'public', tbl
    );

    EXECUTE format('GRANT SELECT, INSERT, UPDATE, DELETE ON %I.%I TO PUBLIC', 'public', tbl);
  END LOOP;
END;
$$ LANGUAGE plpgsql;

COMMIT;
```

2. In master, try to create a new RLS policy from the dashboard, and notice that the role selector does not load. Also check that the network tab has tons of `query?key=TableRolesAccess-public` queries
3. Check that it's fixed on this preview


apps/studio/data/tables/table-roles-access-query.ts
```diff
@@ -29,40 +27,52 @@ export type TableRolesAccessVariables = TableRolesAccessArgs & {
 }
 
 export async function getTableRolesAccess(
-  { schema, table, projectRef, connectionString }: TableRolesAccessVariables,
+  { schema, projectRef, connectionString }: TableRolesAccessVariables,
```
joshenlim: nit but i'd just rename this to schema roles access then for clarity

joshenlim: or tables-role-access

apps/studio/data/tables/table-roles-access-query.ts
```diff
@@ -4,19 +4,17 @@ import { tableKeys } from './keys'
...
   schema: string
-  table: string
 }
 
 /**
  * [Joshen] Specifically just checking for anon and authenticated roles since this is
  * just to verify if the table is exposed via the Supabase API
  */
-export const getTableRolesAccessSql = ({ schema, table }: TableRolesAccessArgs) => {
+export const getTableRolesAccessSql = ({ schema }: TableRolesAccessArgs) => {
```
joshenlim: i'll shift this over to packages/pg-meta

awaseem (approved):
LGTM

joshenlim (approved):

PR #39611, opened 2025-10-17
[FE-1938] feat: show region capacity issues

alaister:
Screenshot 2025-10-17 at 16 27 13

### DIFF
--- a/apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
@@ -1,6 +1,7 @@
 import { ControllerRenderProps, UseFormReturn } from 'react-hook-form'
 
 import { useFlag, useParams } from 'common'
+import AlertError from 'components/ui/AlertError'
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
```diff
@@ -73,6 +83,10 @@ export const RegionSelector = ({
     process.env.NEXT_PUBLIC_ENVIRONMENT === 'local' ||
     process.env.NEXT_PUBLIC_ENVIRONMENT === 'staging'
 
+  if (isErrorAvailableRegions) {
```
ivasilov, line 86: The selector can still work without the available regions, not sure if this is needed.

ivasilov (approved):

apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
alaister: In the very rare case of no capacity being available, this endpoint will error:
image (4)

PR #39629, opened 2025-10-17
fix: add time zone to commit time and add to support message

charislam:
The time zone isn't getting logged on the commit times, which makes it confusing what the actual time is. Also added it to the support message alongside the version number, to help quickly rule out outdated version issues.


joshenlim (approved):
Verified changes on staging preview by checking browser logs

PR #39631, opened 2025-10-17
feat(region selector): get region status by desired size

charislam:
The available-regions endpoint now takes a desired_instance_size query parameter to return more accurate status data.

Merge after https://github.com/supabase/infrastructure/pull/26729

### DIFF
--- a/apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
@@ -1,9 +1,10 @@
-import { ControllerRenderProps, UseFormReturn } from 'react-hook-form'
+import { ControllerRenderProps } from 'react-hook-form'
 
 import { useFlag, useParams } from 'common'
 import AlertError from 'components/ui/AlertError'
 import { useDefaultRegionQuery } from 'data/misc/get-default-region-query'
 import { useOrganizationAvailableRegionsQuery } from 'data/organizations/organization-available-regions-query'
+import type { DesiredInstanceSize } from 'data/projects/new-project.constants'
 import { BASE_PATH, PROVIDERS } from 'lib/constants'
 import type { CloudProvider } from 'shared-data'
 import {
@@ -27,7 +28,7 @@ import { getAvailableRegions } from './ProjectCreation.utils'
 interface RegionSelectorProps {
   cloudProvider: CloudProvider
   field: ControllerRenderProps<any, 'dbRegion'>
-  form: UseFormReturn<any>
+  instanceSize?: DesiredInstanceSize
   layout?: 'vertical' | 'horizontal'
 }
 
@@ -38,6 +39,7 @@ interface RegionSelectorProps {
 export const RegionSelector = ({
   cloudProvider,
   field,
+  instanceSize,
   layout = 'horizontal',
 }: RegionSelectorProps) => {
   const { slug } = useParams()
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
```diff
@@ -54,7 +56,10 @@ export const RegionSelector = ({
     isLoading: isLoadingAvailableRegions,
     isError: isErrorAvailableRegions,
     error: errorAvailableRegions,
-  } = useOrganizationAvailableRegionsQuery({ slug, cloudProvider }, { enabled: smartRegionEnabled })
+  } = useOrganizationAvailableRegionsQuery(
+    { slug, cloudProvider, desiredInstanceSize: instanceSize },
+    { enabled: smartRegionEnabled, staleTime: 1000 * 60 * 5 } // 5 minutes
```
seanoliver: Super nit: 5 mins seems a little long for this kind of data. The risk is low but I'm just thinking that in the event of an outage or a capacity issue users might be more likely to see a stale "available" status than if the timeout was closer to 1-2 mins.

apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
```diff
@@ -54,7 +56,10 @@ export const RegionSelector = ({
     isLoading: isLoadingAvailableRegions,
     isError: isErrorAvailableRegions,
     error: errorAvailableRegions,
-  } = useOrganizationAvailableRegionsQuery({ slug, cloudProvider }, { enabled: smartRegionEnabled })
+  } = useOrganizationAvailableRegionsQuery(
```
seanoliver, line 59: This query includes the new `instanceSize` param, but it looks like the same query in the parent component (`/pages/new/[slug].tsx:L244-257`) does not. Do we want to include it in both queries? If not, we'd have two separate React Query cache entries and potentially return different values for `availableRegionsData`.

```
  // `/pages/new/[slug].tsx:L244-L257`
  const { data: availableRegionsData, error: availableRegionsError } =
    useOrganizationAvailableRegionsQuery(
      {
        slug: slug,
        cloudProvider: PROVIDERS[defaultProvider].id,
      },
      {
        enabled: smartRegionEnabled,
        refetchOnMount: false,
        refetchOnWindowFocus: false,
        refetchInterval: false,
        refetchOnReconnect: false,
      }
    )
```

seanoliver (commented):
LGTM from a growth-eng POV. One thing to double-check: do we also want to pass `instanceSize` to the `useOrganizationAvailableRegionsQuery` call in `/pages/new/[slug].tsx`?

`RegionSelector.tsx` (its child) includes it, but the query in its parent does not. If we keep them different, React Query will treat them as two separate cache entries and we could get different values for `availableRegionsData`.

apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
joshenlim: think this makes sense!

joshenlim (approved):
Verified on staging preview by minimally creating a project|

Smol non-blocking UI issue here: Reckon we can omit the region code in the SelectValue
image

seanoliver (approved):

charislam:
> Smol non-blocking UI issue here: Reckon we can omit the region code in the SelectValue

going to save this one for a separate PR because it involves changing the state and I wanna minimize the number of things I touch on this page at one time (always cautious about this high-value workflows 😬)

PR #39649, opened 2025-10-19
Chore/persist auth users search configuration

joshenlim:
## Context

Related to the recent changes to the Auth Users UI - opting to persist search configuration to improve the UX here.

- Filter type, user filter type, and filter keywords will be URL states
  - So that you'll be able to share the view, or refresh with going back to the same view
- Only filter type will be stored in local storage
  - Will be initialized when landing on `/auth/users` and only if there's no filter type in the URL params
  - `id` value will be ignored since that's the default

## To test
- [ ] Verify that the URL states are working
- [ ] Verify that filter type will be initialized based on previous selection when landing on `/auth/users`


saltcod (approved):
Tested on staging and state persists as expected!

PR #39661, opened 2025-10-20
chore(studio): add feature preview for auth security template UI

dnywh:

- Chore: feature preview
- Resolves DEPR-135
- Resolves DEPR-136

- No feature preview for the incoming security notifications work
- _Authentication_ sidebar has two section: _Manage_ and _Configuration_

## What is the new behavior?

- Added feature preview for the incoming security notifications work
- When that feature preview is on, the _Authentication_ sidebar pulls _Emails_ out from _Configuration_ into a new _Notifications_ section

| Before | Preview | After |
| --- | --- | --- |
| Authentication  Supabase | Authentication  Supabase | Authentication  Supabase|

## Additional context

- Feature flag: `securityNotifications`
- More items coming to the _Notifications_ sidebar section soon: _Push_ and _SMS_, for example


apps/studio/components/interfaces/App/FeaturePreview/FeaturePreview.constants.tsx
```diff
@@ -1,6 +1,13 @@
 import { LOCAL_STORAGE_KEYS } from 'common'
 
 export const FEATURE_PREVIEWS = [
+  {
+    key: LOCAL_STORAGE_KEYS.UI_PREVIEW_SECURITY_NOTIFICATIONS,
+    name: 'Security notification templates',
+    discussionsUrl: undefined,
+    isNew: true,
+    isPlatformOnly: false,
```
dnywh: Should `isPlatformOnly` be `true`? My understanding of `platform` is that it means the “Supabase-hosted version” of Studio. Does that mean these email templates only apply when that’s the case?

CC @issuedat

apps/studio/components/interfaces/App/FeaturePreview/FeaturePreviewContext.tsx
```diff
@@ -138,11 +144,19 @@ export const useFeaturePreviewModal = () => {
...
       }
     },
-    [gitlessBranchingEnabled, advisorRulesEnabled, isUnifiedLogsPreviewAvailable]
+    [
+      gitlessBranchingEnabled,
+      advisorRulesEnabled,
+      isUnifiedLogsPreviewAvailable,
+      isNewStorageUIAvailable,
+      isSecurityNotificationsAvailable,
+    ]
```
dnywh, line 159: These dependencies were missing my earlier `isNewStorageUIAvailable`. So I added that and this new `isSecurityNotificationsAvailable` too.

apps/studio/components/layouts/AuthLayout/AuthLayout.utils.ts
```diff
@@ -21,6 +22,7 @@ export const generateAuthMenu = (
     authenticationAttackProtection,
     authenticationAdvanced,
     authenticationShowOverview,
+    authenticationShowSecurityNotifications,
```
dnywh, line 25: Following the same naming convention and declaration as set out by the other feature flags and previews: `authenticationFeatureDoesThis`.

apps/studio/components/layouts/AuthLayout/AuthLayout.utils.ts
```diff
@@ -71,7 +93,7 @@ export const generateAuthMenu = (
                     },
                   ]
                 : []),
-              ...(authenticationEmails
+              ...(authenticationEmails && !authenticationShowSecurityNotifications
```
dnywh, line 96: When the feature preview is off, leave the _Emails_ section here. But when it’s on, remove it (because it’s shown above).

apps/studio/components/layouts/AuthLayout/AuthLayout.utils.ts
```diff
@@ -33,6 +35,26 @@ export const generateAuthMenu = (
...
+                      pages: ['templates', 'smtp'],
+                      url: `/project/${ref}/auth/templates`,
+                      items: [],
+                    },
+                  ]
+                : []),
+            ],
+          },
+        ]
+      : []),
```
dnywh: Should this be wrapped in a `IS_PLATFORM` check like some items in _Configuration_?

apps/studio/components/layouts/AuthLayout/AuthLayout.utils.ts
```diff
@@ -33,6 +35,26 @@ export const generateAuthMenu = (
...
+          {
+            title: 'Notifications',
+            items: [
+              ...(authenticationEmails
+                ? [
+                    {
+                      name: 'Email',
+                      key: 'email',
+                      pages: ['templates', 'smtp'],
+                      url: `/project/${ref}/auth/templates`,
```
dnywh, line 49: Singular name (_Email_ not _Emails_) because you’re configuring the notification type “email”. This will make more sense when other (singular) options are added: “Push”, ”SMS”.

Goes to existing `url` for now.

apps/studio/components/interfaces/App/FeaturePreview/FeaturePreview.constants.tsx
joshenlim: ^ yup! for self-hosted / local, there's no dashboard support for emails so this is correct

apps/studio/components/layouts/AuthLayout/AuthLayout.utils.ts
joshenlim: yup! lets do that - cause i reckon even after this UI becomes the default, there's no dashboard support for this for self-hosted anyways (i'll add that in)

joshenlim: we should also add the enabled features emails check too btw

joshenlim (approved):
Verified changes on staging preview - also checked configcat that prod flag is off for public

PR #39665, opened 2025-10-20
fix(docs): Clarify auth.admin.generateLink does not send emails or OTPs

thewheat:

YES

- In relation to `auth.admin.generateLink`  it mentions "to be sent via a custom email provider" which can be confusing as this method does not send any emails / OTPs, it just generates them
- Removed that mention and clarified it does not send out emails / OTPs
- I just searched for the phrased and replaced them all, so I'm not 100% sure what each file does. If there are any issues just let me know

### DIFF
--- a/apps/docs/features/docs/__snapshots__/Reference.typeSpec.test.ts.snap
+++ b/apps/docs/features/docs/__snapshots__/Reference.typeSpec.test.ts.snap
@@ -42109,7 +42109,7 @@ exports[`TS type spec parsing > matches snapshot 1`] = `
           }
         },
         "comment": {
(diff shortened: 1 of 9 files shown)


grdsdev (approved):
Approved, but saw you changed dart and kotlin references, let's make sure to also check for Swift and Python.

thewheat:
Seems like 
- Python does not have explanation text with the confusing sentence  https://supabase.com/docs/reference/python/auth-admin-generatelink
- Swift doesn't seem to any generateLink method only inviteUserByEmail  https://supabase.com/docs/reference/swift/auth-admin-inviteuserbyemail
image

So will merge this and we can discuss what possible other changes may need to be done as well

PR #39670, opened 2025-10-20
fix: allow merge request with stale PR to be closed

sweatybridge:

Bug fix

## What is the new behavior?

A merge request may be requested on a branch that has a stale PR. Allows such merge requests to be closed by conditioning on the `review_requested_at` field instead.

## Additional context

### DIFF
--- a/apps/studio/pages/project/[ref]/branches/merge-requests.tsx
+++ b/apps/studio/pages/project/[ref]/branches/merge-requests.tsx
@@ -258,6 +258,9 @@ const MergeRequestsPage: NextPageWithLayout = () => {
                               rowLink={rowLink}
                               external={isPR}
                               rowActions={
+                                // We always want to show the action button to close a merge request
+                                // when user has requested review from dashboard. It doesn't matter
+                                // whether the branch is linked to a GitHub PR.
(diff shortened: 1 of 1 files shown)


apps/studio/pages/project/[ref]/branches/merge-requests.tsx
```diff
@@ -258,7 +258,7 @@ const MergeRequestsPage: NextPageWithLayout = () => {
                               rowLink={rowLink}
                               external={isPR}
                               rowActions={
-                                !isPR && (
+                                branch.review_requested_at && (
```
Comment (ivasilov) [inline apps/studio/pages/project/[ref]/branches/merge-requests.tsx:264]: Shouldn't this be `!isPR || branch.review_requested_at`?

ivasilov (approved):

apps/studio/pages/project/[ref]/branches/merge-requests.tsx
Comment (sweatybridge) [reply apps/studio/pages/project/[ref]/branches/merge-requests.tsx:264]: ~Actually we can just remove this extra condition because we want to show the action buttons on every merge request. The `mergeRequestBranches` variable is already filtered by `review_requested_at`.~

Thanks for the suggestion, will add a comment.

PR #39680, opened 2025-10-20
add coming up state to new project home DESIGN-300

jordienr:
- New project homepage will now show coming up state 

Before: We show the "building" state at the layout level

Now: 
- We show "Coming up" as another project state
- We disable other sections (lower opacity and prevent clicks)

CleanShot 2025-10-20 at 16 35 08@2x


SaxonF (commented):
Looking good! Couple small requests:
- Once project is out of initial coming up status and individual services are coming up, we can show the normal 6 dot service status indicators rather than loading spinner
- When coming up, advisor section remains in loading state, we can just show empty state

```
useEffect(() => {
    if (ref) state.setSelectedDatabaseId(ref)
  }, [ref])
  }, [ref, state])
```
I'm unsure if this would have any negative consequences but it does mean that `state.setSelectedDatabaseId(ref)` is called on at least every page load/route change instead of just once.

jordienr:
> Looking good! Couple small requests:
> 
> * Once project is out of initial coming up status and individual services are coming up, we can show the normal 6 dot service status indicators rather than loading spinner
> * When coming up, advisor section remains in loading state, we can just show empty state
> 
> ```
> useEffect(() => {
>     if (ref) state.setSelectedDatabaseId(ref)
>   }, [ref])
>   }, [ref, state])
> ```
> 
> I'm unsure if this would have any negative consequences but it does mean that `state.setSelectedDatabaseId(ref)` is called on at least every page load/route change instead of just once.

fixed!

apps/studio/components/layouts/ProjectLayout/ProjectLayout.tsx
```diff
@@ -383,7 +395,16 @@ const ContentWrapper = ({ isLoading, isBlocking = true, children }: ContentWrapp
...
+  if (shouldRedirectToHomeForBuilding) {
+    return <Loading />
+  }
+
+  if (
+    requiresDbConnection &&
+    isProjectBuilding &&
+    !isBranchesPage &&
+    !(isHomeNewFlag && isHomePage) // [jordi] We won't be showing the building state with the new home page
+  ) {
```
awaseem: nit: I like how you assigned the boolean evaluation to a variable for `shouldRedirectToHomeForBuilding`. Can we do something similar here, its a bit cleaner to read the variable intention than doing mental boolean gymnastics 😆

apps/studio/components/interfaces/HomeNew/AdvisorSection.tsx
```diff
@@ -199,15 +210,21 @@ export const AdvisorSection = () => {
...
+    <Card className="bg-transparent">
+      <CardContent className="flex flex-col items-center justify-center gap-2 p-16">
+        <Shield size={20} strokeWidth={1.5} className="text-foreground-muted" />
+        <p className="text-sm text-foreground-light text-center">
+          No security or performance errors found
+        </p>
+      </CardContent>
+    </Card>
+  )
+}
```
awaseem, line 230: nit: possible to keep a single component per file? or move this to a emptyStates file like we do for the old home?

awaseem (commented):
Minor nits, hope I wasn't super annoying 😬

apps/studio/components/interfaces/HomeNew/AdvisorSection.tsx
jordienr: Think i'll keep this one internal since its only used here, can move it out when needed elsewhere.

apps/studio/components/layouts/ProjectLayout/ProjectLayout.tsx
jordienr: 100% good catch

awaseem (approved):
LGTM

Issue #39656, opened 2025-10-19
Enabling "Expose Queues via PostgREST" fails on newest Postgres Version due to pgmq.read parameter change


NicolaiKobras:
# Bug report

- [ x ] I confirm this is a bug with Supabase, not with my own application. 
- [ x ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Enabling "Expose Queues via PostgREST" fails on newest Postgres Version because it expects a function with 3 parameters, but the function takes 4  (text, integer, integer, json)

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Enable Queues Integration
2. try to enable "Expose Queues via PostgREST"
4. See error:
`Failed to toggle queue exposure via PostgREST: function pgmq.read(text, integer, integer) does not exist`

## Expected behavior

It should create the correct public schema

Hallidayo:
Hi @NicolaiKobras - Thanks for opening this one. I've tried the above on a test project I have and I'm able to enable without error.

Could you run the following: `SELECT * FROM pg_extension` and see if `pgmq` exists?

NicolaiKobras:
Hi @Hallidayo,
this is the output from `SELECT * FROM pg_extension `:

<img width="939" height="39" alt="Image" src="https://github.com/user-attachments/assets/447a7955-5139-4d87-9824-4cf0136a27f8" />

Also this is my Versions:

<img width="1195" height="346" alt="Image" src="https://github.com/user-attachments/assets/bcff2074-272d-41dd-bd97-af06ae8626e7" />

Hallidayo:
Ahh my test project is on an old version of Postgres, I'm just going to upgrade now and see if I can recreate. On my current version I was running `pgmq` on version 1.4.4 so I think thats why I was not having the issue.

EDIT: I'm on the most recent version and I am getting the error now. I'll flag this up.

lassestolley:
I get the same error: 
`Failed to toggle queue exposure via PostgREST: function pgmq.read(text, integer, integer) does not exist.`

PostgREST version: 13.0.5
Postgres version: 17.6.1.021
pgmg version: 1.5.1

Is there a known way to fix this in the short term?

NicolaiKobras:
I guess we could just insert the sql command that this executes and fix the function mappings
[...]

PR #39695, opened 2025-10-20
fix #39693 "Expose Queues via PostgREST" not working because of new parameter condition in pgmq.read 

NicolaiKobras:

YES

Bug fix

https://github.com/supabase/supabase/issues/39656#issuecomment-3423777486

## What is the new behavior?

the public wrapper function for pgmq.read now includes the optional parameter `condition`
The default value will be {} as to not break any current implementations of this on supabase instances.

This will enable the UI to "Expose Queues via PostgREST" to work again while not breaking any existing instances.


samrose:
Hi thank you for this PR. We're looking into the best way to address this, and will respond ASAP

NicolaiKobras:
This was implemented better and more feature complete in #40023 :)

PR #39697, opened 2025-10-21
Fixes diff editor widget

SaxonF:
This error only occurs locally as far as I know but when using the inline assistant for sql editor and ai editor when you accept or reject an answer it throws an error:

```
TextModel got disposed before DiffEditorWidget model got reset

Error: TextModel got disposed before DiffEditorWidget model got reset
```

To test:
- Locally, use SQL Editor , prompt assistant , accept answer and see error
- Switch to branch and do the same 
- Test same on staging


joshenlim (approved):
Verified changes locally + sanity checked staging preview

joshenlim:
Closing in favor of https://github.com/supabase/supabase/pull/39703

PR #39610, opened 2025-10-17
Swap analytics buckets endpoint

joshenlim:
## Context

Depends on infrastructure PR here: https://github.com/supabase/infrastructure/pull/26697

Swaps out endpoints for listing, creating and deleting analytics buckets with the new buckets that are specifically for analytics buckets. All actions have been tested locally and working, so just pending on the BE PR above to be merged so we can test on staging preview

[Update 291025] Verified locally that all functionalities except deleting analytics bucket is working with the existing UI
However, deleting analytics bucket is currently broken on prod actually since its using the old UI

## Note

This will serve as the main branch atm for Analytics Buckets changes while we're waiting on the BE PR to get through, so related changes will be PR-ed + code reviewed into this branch. And when the time comes to merge this one into master, we'll just need to do an E2E test on staging preview

## Pull requests that have gone into this branch:
All the changes from the following PRs should've been reviewed and tested before going into this branch. So when doing a final check on this PR before merging, we can just do a smoke test and skip code review
- https://github.com/supabase/supabase/pull/39701
- https://github.com/supabase/supabase/pull/39902
- https://github.com/supabase/supabase/pull/39906

## To test
- [ ] Can create analytics bucket
- [ ] Can view analytics bucket
- RE deleting analytics bucket, it's currently also error-ing on prod so feel free to ignore this for now

PR #39701, opened 2025-10-21
chore(studio): analytics buckets contents

dnywh:

- Chore
- UI improvements to analytics buckets

- Analytics buckets have a rudimentary set up and outdated table/namespace UI

## What is the new behavior?

- Analytics buckets reflect the latest UI with focus on Tables, grouped by namespace/schema (instead of the other way around)
  - Current behavior is mantained for users with `isStorageV2` set to `false`
- Dummy _Connect tables_ dialog
  - `disabled` to avoid confusion until we implement this feature end-to-end

## Additional context

- See [this update](https://supabase.slack.com/archives/C0543H0QL1Z/p1761197693642859?thread_ts=1761197667.230239&cid=C0543H0QL1Z) for a comprehensive (visual) walkthrough
- See @joshenlim’s base branch PR https://github.com/supabase/supabase/pull/39610 that this will merge into
  - See the infra PR https://github.com/supabase/infrastructure/pull/26697 that both of these branches rely upon
 
## Known issues

- Going to a non-existent bucket route will not show the empty state like it should. It instead redirects back to the org page. I believe there is something wrong with the `projectRef` lookup from useStorageExplorerStateSnapshot. But this predates this PR (and its base branch). ([source](https://github.com/supabase/supabase/blob/f83bba4e2439b161ee608a4564681e31a57d2849/apps/studio/pages/project/%5Bref%5D/storage/analytics/buckets/%5BbucketId%5D.tsx#L22-L23))


apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
```diff
@@ -189,116 +179,57 @@ export const AnalyticBucketDetails = ({ bucket }: { bucket: AnalyticsBucket }) =
 
           {state === 'added' && wrapperInstance && (
             <>
-              {isStorageV2 ? (
```
dnywh, line 192: I heard that no non-Supabase folks have access to analytics buckets anyway, so removed this check. Confirming [here](https://supabase.slack.com/archives/C08R9PG2PMK/p1761197987607049?thread_ts=1761106881.052869&cid=C08R9PG2PMK).

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
dnywh, line 1: A better name for this component may now be “NamespaceCluster” or ”TableGroup” because it is a grouping of tables by namespace/target schema.

Anyway, a change for another day.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
```diff
@@ -23,9 +49,122 @@ type NamespaceRowProps = {
...
+
+  const handleQueryTable = () => {
+    // TODO: Implement query table functionality
+    console.log('Query table:', tableName)
+  }
+
+  const handleDeleteTable = () => {
+    // TODO: Implement delete table functionality
+    console.log('Delete table:', tableName)
+  }
```
dnywh: Not yet implemented. Would love help with this! Also unsure what happens here. Are both tables deleted? Just the target?

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
```diff
@@ -23,9 +49,122 @@ type NamespaceRowProps = {
...
+            <Button asChild type="default" size="tiny">
+              <Link href={`/project/${project?.ref}/editor/${tableName}`}>
+                <p>View table</p>
+              </Link>
+            </Button>
+            <DropdownMenu>
+              <DropdownMenuTrigger asChild>
+                <Button type="default" className="px-1" icon={<MoreVertical />} />
+              </DropdownMenuTrigger>
+
```
dnywh: Non of these dropdown actions are functional yet. Just getting the UI ready.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
```diff
@@ -23,9 +49,122 @@ type NamespaceRowProps = {
...
+        </TableCell>
+      )}
+    </TableRow>
+  )
+}
+
 export const NamespaceRow = ({
   bucketName,
   namespace,
+  sourceType = 'direct',
```
dnywh: Whether this is coming from an Iceberg client (`replication` or from Supabase table `direct`). Has UI implications inside of this component

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
```diff
@@ -65,66 +206,154 @@ export const NamespaceRow = ({
...
-    if (isLoadingNamespaceTables) return 'Loading tables...'
-    if (missingTables.length > 0) return `Found ${missingTables.length} new tables`
-    if (tables.length === 0) return 'No tables found'
+    if (isImportingForeignSchema) return 'Looking for new tables...'
+    if (isLoadingNamespaceTables) return 'Loading new tables...'
+    if (missingTables.length > 0)
+      return `${missingTables.length} new table${missingTables.length > 1 ? 's' : ''} found`
+    if (tables.length === 0) return 'No new tables found'
     return 'All tables are up to date'
   }, [isImportingForeignSchema, isLoadingNamespaceTables, missingTables.length, tables.length])
```
dnywh, line 74: Unused but keeping around for a moment in case we reintroduce.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
```diff
@@ -65,66 +206,154 @@ export const NamespaceRow = ({
...
+  }, [schema, tables.length])
+
+  // Determine what schema name to display
+  const displaySchema = useMemo(() => {
+    // If we have a target schema, use it
+    if (schema) return schema
+
+    // Otherwise, show the incoming schema (namespace name)
+    return `fdw_analytics_${namespace.replaceAll('-', '_')}`
+  }, [schema, namespace])
```
dnywh: Mock logic here to catch clashes and handle prefixes.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
```diff
@@ -65,66 +206,154 @@ export const NamespaceRow = ({
...
+                    {tables.length === 0 && <Info size={12} />}
+                  </span>
+                </TooltipTrigger>
+                <TooltipContent>
+                  <p>Target schema{tables.length === 0 && ' that will be created'}</p>
+                </TooltipContent>
+              </Tooltip>
+            </TooltipProvider>
+          )}
+        </CardTitle>
```
dnywh: This title has different colour and tooltip treatment depending on the following factors:
- Does the target schema already exist?
- Do tables already exist?
- Is the source a Postgres table or Iceberg client?

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
```diff
@@ -65,66 +206,154 @@ export const NamespaceRow = ({
...
+
+        <div className="flex flex-row gap-x-2">
+          {/* {tables.length > 0 && <p className="text-sm text-foreground-muted">{scanTooltip}</p>} */}
+          {/* {scanTooltip && <p className="text-sm text-foreground-muted">{scanTooltip}</p>} */}
```
dnywh: Keeping this around because we may reuse. Feel free to delete though

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
```diff
@@ -65,66 +206,154 @@ export const NamespaceRow = ({
...
+                index={index}
+              />
+            ))
+          )}
+        </TableBody>
+      </Table>
       <ImportForeignSchemaDialog
         bucketName={bucketName}
         namespace={namespace}
+        circumstance="clash"
```
dnywh: Now that we likely are only showing the `ImportForeignSchemaDialog` when there is a clash, I’ve hardcoded this new prop. I figured we could remove the prop once confirmed (and remove the 'generic' case.

apps/studio/components/interfaces/Storage/AnalyticsBuckets.tsx
```diff
@@ -85,7 +85,7 @@ export const AnalyticsBuckets = () => {
                 </TableHeader>
                 <TableBody>
                   {analyticsBuckets.length === 0 && filterString.length > 0 && (
-                    <TableRow>
+                    <TableRow className="[&>td]:hover:bg-inherit">
```
dnywh, line 88: Don’t show highlight on hover if there is no content. Honestly this should go in the component itself, IMO.

apps/studio/components/interfaces/Storage/AnalyticsBuckets.tsx
```diff
@@ -95,15 +95,7 @@ export const AnalyticsBuckets = () => {
                     </TableRow>
                   )}
                   {analyticsBuckets.map((bucket) => (
-                    <TableRow
-                      key={bucket.id}
-                      className="cursor-pointer"
-                      onClick={(event) => {
-                        const url = `/project/${ref}/storage/analytics/buckets/${bucket.id}`
-                        if (event.metaKey) window.open(url, '_blank')
-                        else router.push(url)
```
dnywh, line 104: Removed the ability to click the whole row. Was annoying and redundant given the _View contents_ button.

apps/studio/components/interfaces/Storage/ConnectTablesDialog.tsx
dnywh, line 1: Basic new dialog to handle the table connection process. In other words:
1. Supabase user wants to stream changes from a Postgres table into an analytics bucket.
2. They tap ‘Connect table’.
3. This dialog appears.
4. They select which table(s) they want to connect.

The current state of this dialog only:
- Lists existing tables

Nothing else. It should definitely:
- Begin the replication process upon _Connect_ tapped

It should arguably:
- Leave out schemas that are read-only
- Leave out tables that are already connected to another bucket

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
dnywh, line 1: Same “don’t show hover treatment if there are no rows” as above.

apps/studio/components/interfaces/Storage/ImportForeignSchemaDialog.tsx
dnywh, line 1: As mentioned above, I’ve just added a prop to handle the clash case, which might be the only case in the near future.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
```diff
@@ -47,11 +46,13 @@ import {
 } from 'ui'
 import { Admonition } from 'ui-patterns/admonition'
 import { GenericSkeletonLoader } from 'ui-patterns/ShimmeringLoader'
+import { ConnectTablesDialog } from '../ConnectTablesDialog'
 import { DeleteBucketModal } from '../DeleteBucketModal'
 import { DESCRIPTIONS, LABELS, OPTION_ORDER } from './constants'
 import { CopyEnvButton } from './CopyEnvButton'
 import { DecryptedReadOnlyInput } from './DecryptedReadOnlyInput'
 import { NamespaceRow } from './NamespaceRow'
+import { NamespaceWithTables } from './NamespaceWithTables'
```
dnywh, line 56: Separate component so existing `NamespaceRow` users aren’t affected.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
```diff
@@ -191,39 +192,47 @@ export const AnalyticBucketDetails = ({ bucket }: { bucket: AnalyticsBucket }) =
...
+                    <aside className="border border-dashed w-full bg-surface-100 rounded-lg px-4 py-10 flex flex-col gap-y-3 items-center text-center gap-1 text-balance">
+                      <SquarePlus size={24} strokeWidth={1.5} className="text-foreground-muted" />
+                      <div className="flex flex-col items-center text-center">
+                        <h3>Connect database tables</h3>
+                        <p className="text-foreground-light text-sm">
+                          Stream data from tables for archival, backups, or analytical queries.
+                        </p>
+                      </div>
+                      <ConnectTablesDialog />
+                    </aside>
```
dnywh, line 218: Proper empty state. I wonder if this should be a shared component now that I’ve used it a few times?

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceWithTables.tsx
dnywh, line 1: As mentioned elsewhere, this is now a forked version of `NamespaceRow` to support a big deviation from how we currently present tables (and their namespaces).

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceWithTables.tsx
```diff
@@ -0,0 +1,345 @@
...
+                  <p>Postgres schema{tables.length === 0 && ' that will be created'}</p>
+                </TooltipContent>
+              </Tooltip>
+            </TooltipProvider>
+          )}
+        </CardTitle>
+
+        <div className="flex flex-row gap-x-2">
+          {/* {tables.length > 0 && <p className="text-sm text-foreground-muted">{scanTooltip}</p>} */}
+          {/* {scanTooltip && <p className="text-sm text-foreground-muted">{scanTooltip}</p>} */}
```
dnywh, line 277: Would be good to keep around until we resolve language and UI details.

apps/studio/components/interfaces/Storage/ConnectTablesDialog.tsx
dnywh: After speaking with @ivandotv, @imor, and @fenos: we’ve decided to disable this dialog entry until we’ve built out the working end-to-end ETL flow. The `disabled` `ButtonTooltip` lets us do both.

The reason for this is to balance:
- Getting this in the hands of testers ASAP
- Whilst allowing this in-Studio flow to continue being developed

apps/studio/pages/project/[ref]/storage/analytics/buckets/[bucketId].tsx
```diff
@@ -20,13 +23,22 @@ const AnalyticsBucketPage: NextPageWithLayout = () => {
...
+            <Admonition
+              className="max-w-md"
+              type="default"
+              title="Unable to find bucket"
+              description={`${bucketId ? `The template "${bucketId}"` : 'This template'} doesn’t seem to exist.`}
+            >
+              <Button asChild type="default" className="mt-2">
+                <Link href={`/project/${projectRef}/storage/analytics`}>Head back</Link>
+              </Button>
+            </Admonition>
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/analytics/buckets/[bucketId].tsx:42]: Better empty states as discussed elsewhere. Documenting as a pattern for future reference.

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceWithTables.tsx
```diff
@@ -0,0 +1,345 @@
...
+const TableRowComponent = ({
+  tableName,
+  isConnected,
+  schema,
+  isLoading,
+  index,
+}: {
+  tableName: string
+  isConnected: boolean
+  schema?: string
```
joshenlim: unused prop

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceWithTables.tsx
```diff
@@ -0,0 +1,345 @@
...
+
+    // Otherwise, show the incoming schema (namespace name)
+    return `fdw_analytics_${namespace.replaceAll('-', '_')}`
+  }, [schema, namespace])
+
+  return (
+    <Card>
+      <CardHeader className="flex flex-row justify-between items-center px-4 py-5 space-y-0">
+        <CardTitle className="text-sm font-normal font-sans normal-case leading-none flex flex-row items-center gap-x-1">
+          <TooltipProvider>
```
joshenlim, line 237: don't need to declare TooltipProvider here as there's already one at the _app level

only need to declare if using custom properties like delay duration

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
joshenlim: i might be overlooking something but where else are we using this empty state?

apps/studio/components/interfaces/Storage/AnalyticsBuckets.tsx
joshenlim: agree, although we can look into that separately as we'll need to double check the other UIs that are using this component

apps/studio/components/interfaces/Storage/ConnectTablesDialog.tsx
```diff
@@ -0,0 +1,186 @@
...
+        message: !!match
+          ? `Bucket name cannot contain the "${match}" character`
+          : 'Bucket name contains an invalid special character',
+      })
+    }
+  })
+
+const formId = 'connect-tables-form'
+
+export type ConnectTablesForm = z.infer<typeof FormSchema>
```
joshenlim: don't need to export if not being imported elsewhere

apps/studio/components/interfaces/Storage/ConnectTablesDialog.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+interface ConnectTablesDialogProps {}
```
joshenlim: can remove this if no props

apps/studio/components/interfaces/Storage/ImportForeignSchemaDialog.tsx
```diff
@@ -23,6 +23,7 @@ export interface ImportForeignSchemaDialogProps {
   bucketName: string
   namespace: string
   wrapperMeta: WrapperMeta
+  circumstance: 'fresh' | 'clash'
```
joshenlim: should this be an optional parameter? its causing a TS issue in `NamespaceRow.tsx`
[edit] ahhh yes it can be - saw that the default value for it is clash

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceWithTables.tsx
```diff
@@ -0,0 +1,345 @@
...
+                index={index}
+              />
+            ))
+          )}
+        </TableBody>
+      </Table>
+      <ImportForeignSchemaDialog
+        bucketName={bucketName}
+        namespace={namespace}
+        circumstance="clash"
```
joshenlim: just to double check if "clash" is meant to be hardcoded here

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceRow.tsx
joshenlim: we can likely look into this when we're implementing "Connect tables" next - given that it's disabled for now

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceWithTables.tsx
dnywh: Ah gotcha. Thanks!

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/NamespaceWithTables.tsx
dnywh: `ImportForeignSchemaDialog` is now used in two different ways:

- **Existing analytics buckets UI:** users must manually link a schema to the namespace, no matter what. `circumstance="fresh"`.
- **New anlaytics buckets in `isStorageV2`**: we do this automatically and _only_ show this dialog in the case there is a clash with what would have been the schema. It should be very rare. That’s `circumstance="clash"`.

It’s hardcoded in _NamespaceWithTables_ because that component is only available in `isStorageV2`.

But, now that I write this all out, I suppose we should fork `ImportForeignSchemaDialog` into a different dialog given they are different purposes?

PR #39703, opened 2025-10-21
Silence text model error in diff editor

joshenlim:
## Context

Alternative to this PR [here](https://github.com/supabase/supabase/pull/39697) RE "Error: TextModel got disposed before DiffEditorWidget model got reset" error happening only locally.

Saw this comment in this [thread](https://github.com/suren-atoyan/monaco-react/issues/647#issuecomment-2897027817) that suggests this fix, which seems to be working

## To reproduce
- Locally, use SQL Editor , prompt assistant , accept answer and see error


alaister (approved):
LGTM!

PR #39388, opened 2025-10-09
fix: update queries for local edge function and cron logs

sweatybridge:
## What kind of change does this PR introduce?

Bug fix

## What is the new behavior?

Functions

Screenshot 2025-10-09 at 5 28 49 PM

Cron

Screenshot 2025-10-09 at 5 28 38 PM

## Additional context

Discussed with @Ziinc and agree that it's not ideal to have local function logs differ from hosted. But fixing the translation query in local logflare is more complex, so we can merge this as a stop gap.

closes https://github.com/supabase/supabase/issues/38121
closes https://github.com/supabase/supabase/issues/36378
closes https://github.com/supabase/supabase/issues/36378
closes https://github.com/supabase/supabase/issues/33834
closes https://github.com/supabase/supabase/issues/33549

PR #39704, opened 2025-10-21
fix: cron logs query to handle type cast

sweatybridge:

Bug fix

## What is the new behavior?

Follow up from https://github.com/supabase/supabase/pull/39388#issuecomment-3419806412

## Additional context

To test locally

```
docker build -f apps/studio/Dockerfile . -t supabase/studio:dev
echo 'dev' > supabase/.temp/studio-version
npx supabase@latest start
```

Then add a new cron job in studio
```
create extension pg_cron;

select
  cron.schedule(
    'call-hello-world-every-5-minutes',
    '*/5 * * * *',
    'select hello_world()'
  );
```


joshenlim (commented):
Approving to unblock!

joshenlim (approved):

PR #39716, opened 2025-10-21
docs: vector and analytics storage

ChrisChinchilla:

- Resolves DOCS-469
- Affects DEPR-194

- Storage docs currently:
  - Focus mainly on file storage
  - Have outdated, sparse, content on analytics buckets

## What is the new behavior?

- Expanding Storage docs to:
  - Give love to analytics and vector buckets
  - Give them equal standing to file buckets


devkyputra03-dot:
Ya

dnywh:
Just did some basic scaffolding. 

I wonder if the analytics and vector accordion sections should become core sidebar sections, given that everything else within _Storage_ is about **file** storage.

In other words: it feels strange that “File buckets” isn’t one of these accordion sections. But then again, if it were, it would house _all the other content on the page_. So breaking analytics and vector buckets into proper sections seems like the cleanest solution.

Thoughts?

CleanShot 2025-11-06 at 14 18 22@2x

charislam (approved):

PR #39720, opened 2025-10-21
Swap project query to fix team role picker

saltcod:
swap project query


awaseem (approved):
LGTM

joshenlim (approved):
Verified changes both locally and on staging preview!

PR #39728, opened 2025-10-21
docs(readme): fix dev instructions for internal/external

charislam:


awaseem (approved):
WOOO to updated docs!

PR #39730, opened 2025-10-21
cleanup(region selector): remove code from selected region

charislam:
Selector looks a bit busy and you probably know the code if you've already selected it, so removing it from the dropdown as suggested in a previous PR.

### DIFF
--- a/apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
@@ -89,8 +89,7 @@ export const RegionSelector = ({
     process.env.NEXT_PUBLIC_ENVIRONMENT === 'staging'
 
   const allSelectableRegions = [...smartRegions, ...regionOptions]
-  const selectedRegion =
-allSelectableRegions.find((region) => {
+  const selectedRegion = allSelectableRegions.find((region) => {
     return !!region.name && region.name === field.value
   })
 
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
```diff
@@ -88,6 +88,14 @@ export const RegionSelector = ({
     process.env.NEXT_PUBLIC_ENVIRONMENT === 'local' ||
     process.env.NEXT_PUBLIC_ENVIRONMENT === 'staging'
 
+  const allSelectableRegions = [...smartRegions, ...regionOptions]
+  const selectedRegion =
+    field.value !== undefined
+      ? allSelectableRegions.find((region) => {
+          return !!region.name && region.name === field.value
+        })
+      : undefined
```
awaseem: Aren't these type checks a bit overkill? This can be simplifed to:


```suggestion
  const selectedRegion =
allSelectableRegions.find((region) => {
    return !!region.name && region.name === field.value
  })
```

The value is going to get checked to see if its undefined in the renderer right? Tried this on my local and no TS errors

awaseem (commented):
Minor changes

apps/studio/components/interfaces/ProjectCreation/RegionSelector.tsx
charislam: Yep, you're right! Thanks for the catch

awaseem (approved):
LGTM 🚀

PR #39733, opened 2025-10-21
ci(dependabot): don't get prompted for major version updates for github actions

charislam:
Major version updates require more preparation and testing, so we only want to update these on our schedule / if actually deprecated, not be prompted as soon as a new major version comes out. We mostly want to keep patch/minor updates so we can be aware of any vulnerability fixes.


awaseem (approved):
LGTM

PR #39742, opened 2025-10-21
Fix: Filtering values on a self hosted instance within edge_logs throwing errors

awaseem:

YES

Bug fix, root cause:
I think this is a typecast issue, if you pass in a single where clause Postgres engine can’t evaluate the type inference where it can with multiple statements.

https://linear.app/supabase/issue/FE-1532/fix-filters-in-self-hosted

Selecting a single value for the edge_logs within the "product" dropdown causes an error for the analytics service. This doesn't cause the error anymore. This does create some strange behavior that I can't understand. investigation below 👇 

## Additional context

There is a strange unknown issue that I can't solve, When I query multiple filters I get results. Example query:

```
-- local dev edge_logs query
select id, edge_logs.timestamp, event_message, request.method, request.path, request.search, response.status_code
from edge_logs
cross join unnest(metadata) as m
  cross join unnest(m.request) as request
  cross join unnest(m.response) as response
where (request.path::text like '/auth/%' or request.path::text like '/storage/%')
order by timestamp desc
limit 100;
```

I get the following results:
Screenshot 2025-10-21 at 12 28 28 PM

but when I query a single value, I get nothing. Example query:

```
-- local dev edge_logs query
select id, edge_logs.timestamp, event_message, request.method, request.path, request.search, response.status_code
from edge_logs
cross join unnest(metadata) as m
  cross join unnest(m.request) as request
  cross join unnest(m.response) as response
where (request.path::text like '/auth/%')
order by timestamp desc
limit 100;
```

And on top of that we have zero errors from the analytics service locally

### Weirder edge case

or some reason this query does return results, its just the same where clause duplicated with an OR

```
-- local dev edge_logs query
select id, edge_logs.timestamp, event_message, request.method, request.path, request.search, response.status_code
from edge_logs
cross join unnest(metadata) as m
  cross join unnest(m.request) as request
  cross join unnest(m.response) as response
where (request.path::text like '/auth/%' or request.path::text like '/auth/%')
order by timestamp desc
limit 100;
```


awaseem:
Moving to draft and waiting for PSQL devs to comment

PR #39745, opened 2025-10-21
Add type-safe event tracking utility

seanoliver:
## Summary

Adds a clean, type-safe wrapper for telemetry event tracking that automatically injects project and organization context.

Resolves GROWTH-517

### Before
```tsx

sendEvent({
  action: 'table_created',
  properties: { method: 'sql_editor' },
  groups: {
    project: project?.ref ?? 'Unknown',
    organization: org?.slug ?? 'Unknown'
  }
})
```

### After
```tsx
const track = useTrack()
track('table_created', { method: 'sql_editor' })
```

## Changes

- Export `TelemetryGroups` type from telemetry-constants
- Add `useTrack()` hook with full TypeScript event validation
- Refactor project creation events to use new API (demonstrates usage)

## Benefits

- **Less boilerplate**: One import instead of three hooks
- **Type-safe**: TypeScript validates action names and properties
- **Automatic context**: Project/org injected from React Query hooks
- **Cleaner code**: ~10 lines reduced to ~2 lines per event

## Test Plan

- [x] Type check passes
- [x] Refactored 4 project creation events as proof of concept
- [x] Events fire correctly with proper context

### DIFF
--- a/apps/studio/lib/telemetry/track.ts
+++ b/apps/studio/lib/telemetry/track.ts
@@ -1,0 +1,61 @@
+import { sendTelemetryEvent } from 'common'
+import { TelemetryEvent, TelemetryGroups } from 'common/telemetry-constants'
+import { useSelectedOrganizationQuery } from 'hooks/misc/useSelectedOrganization'
+import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
+import { API_URL } from 'lib/constants'
+import { useRouter } from 'next/router'
+import { useCallback } from 'react'
+
+type EventMap = {
+  [E in TelemetryEvent as E['action']]: E
+}
+
+type PropertiesForAction<A extends keyof EventMap> = EventMap[A] extends { properties: infer P }
+  ? P
+  : never
+
+type HasProperties<A extends keyof EventMap> = EventMap[A] extends { properties: any }
+  ? true
+  : false
+
+/**
+ * Hook for type-safe telemetry event tracking with automatic project/org context injection.
+ *
+ * @example
+ * const track = useTrack()
+ * track('table_created', { method: 'sql_editor', schema_name: 'public' })
+ * track('help_button_clicked')
+ */
+export const useTrack = () => {
+  const { data: project } = useSelectedProjectQuery()
+  const { data: org } = useSelectedOrganizationQuery()
+  const router = useRouter()
+
+  const track = useCallback(
+    <A extends keyof EventMap>(
+      action: A,
+      ...args: HasProperties<A> extends true
+        ? [properties: PropertiesForAction<A>, groupOverrides?: Partial<TelemetryGroups>]
+        : [properties?: undefined, groupOverrides?: Partial<TelemetryGroups>]
+    ) => {
+      const [properties, groupOverrides] = args
+
+      const groups = {
+        ...(project?.ref && { project: project.ref }),
+        ...(org?.slug && { organization: org.slug }),
+        ...groupOverrides,
+      }
+
+      const event = {
+        action,
+        ...(properties && { properties }),
+        ...(groups && { groups }),
+      } as EventMap[A]
+
+      sendTelemetryEvent(API_URL, event, router.pathname)
+    },
+    [project?.ref, org?.slug, router.pathname]
+  )
+
+  return track
+}
--- a/apps/studio/pages/new/[slug].tsx
+++ b/apps/studio/pages/new/[slug].tsx
@@ -43,7 +43,7 @@ import {
   ProjectCreateVariables,
   useProjectCreateMutation,
 } from 'data/projects/project-create-mutation'
-import { useSendEventMutation } from 'data/telemetry/send-event-mutation'
+import { useTrack } from 'lib/telemetry/track'
 import { useCustomContent } from 'hooks/custom-content/useCustomContent'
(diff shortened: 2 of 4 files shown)


apps/studio/pages/new/[slug].tsx
```diff
@@ -246,16 +246,16 @@ const Wizard: NextPageWithLayout = () => {
...
+        {
           instanceSize: form.getValues('instanceSize'),
         },
-        groups: {
+        {
           project: res.ref,
           organization: res.organization_slug,
-        },
-      })
+        }
```
Comment (seanoliver) [inline apps/studio/pages/new/[slug].tsx:258]: Note: We still need to pass `project` here because the project hasn't technically been created yet.

apps/studio/lib/telemetry/track.ts
```diff
@@ -0,0 +1,59 @@
...
+      const groups = {
+        ...(project?.ref && { project: project.ref }),
+        ...(org?.slug && { organization: org.slug }),
+        ...groupOverrides,
+      }
+
+      const event = {
+        action,
+        ...(properties && { properties }),
+        ...(Object.keys(groups).length > 0 && { groups }),
```
awaseem: Pretty sure this keys check is not needed you can achieve the same using the following: 
```suggestion
        ...(groups && { groups }),
```

apps/studio/lib/telemetry/track.ts
```diff
@@ -0,0 +1,59 @@
...
+      ...args: HasProperties<A> extends true
+        ? [properties: PropertiesForAction<A>, groupOverrides?: Partial<TelemetryGroups>]
+        : [properties?: undefined, groupOverrides?: Partial<TelemetryGroups>]
+    ) => {
+      const [properties, groupOverrides] = args
+
+      const groups = {
+        ...(project?.ref && { project: project.ref }),
+        ...(org?.slug && { organization: org.slug }),
+        ...groupOverrides,
```
awaseem, line 46: In a lot of places, the send event actually default the values to `Unknown`. To keep the parity we should do the same in this wrapper right? Example files:
- supabase/apps/studio/components/layouts/ProjectLayout/LayoutHeader/MergeRequestButton.tsx
- /supabase/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx

Its in a bunch of other places, we do go that route the object spread becomes simple:

 
```suggestion
        { project: project.ref ?? 'Unknown',   organization: org.slug ?? 'Unknown', ...groupOverrides }
```

awaseem (commented):
Minor suggestions, overall I like this change but would want someone like @joshenlim, @charislam to review for a second opinion.

apps/studio/lib/telemetry/track.ts
seanoliver: Good catch! Though we're actually moving away from hardcoding `"Unknown"` and just letting these fall back to `undefined` when data isn’t available.

Having "Unknown" show up in event payloads tends to cause analytics headaches since it gets treated as a real value instead of missing data. Leaving it `undefined` makes filtering and grouping a lot cleaner downstream.

Context here: https://linear.app/supabase/issue/GROWTH-459/remove-unknown-fallbacks-from-telemetry-events

apps/studio/lib/telemetry/track.ts
seanoliver: Agreed. There will be some cases where we pass `groups: {}` as part of the payload (e.g. signup events), but our analytics tools will just ignore that. This version’s a lot cleaner from a code perspective — thanks @awaseem!

apps/studio/lib/telemetry/track.ts
```diff
@@ -0,0 +1,59 @@
...
+  const { data: project } = useSelectedProjectQuery()
+  const { data: org } = useSelectedOrganizationQuery()
+  const { mutate: sendEvent } = useSendEventMutation()
+
+  const track = useCallback(
+    <A extends keyof EventMap>(
+      action: A,
+      ...args: HasProperties<A> extends true
+        ? [properties: PropertiesForAction<A>, groupOverrides?: Partial<TelemetryGroups>]
+        : [properties?: undefined, groupOverrides?: Partial<TelemetryGroups>]
```
charislam, line 39: This is absolutely a nit, but the signature isn't as ergonomic as it could be, because you have to pass `undefined` in properties to get to `groupOverrides`: can we omit it entirely?

seanoliver: Yeah good call. I actually thought about that too and looked through the codebase to see where we’d actually need group overrides. The only real cases are entity-creation events (like project or org creation) where the new entity isn’t yet in React Query cache, so we have to manually pass the fresh values:

``` ts
onSuccess: (res) => {
  track(
    'project_creation_simple_version_submitted',
    { instanceSize: form.getValues('instanceSize') },
    { project: res.ref, organization: res.organization_slug }
  )
}
```

But those always include `properties` anyway, so in practice we never hit the awkward “undefined middle param” case.

For everything else, `project` and `organization` are already available in cache, so calls like `track('form_submitted')` will pass valid `groups` data inherently without having to explicitly pass `undefined` for an empty `properties` object.

So your point makes sense, but my inclination is to keep the current signature for now since changing it would make other calls more verbose without solving an actual problem. If we ever hit a case that needs it, we can always revisit. WDYT? 🙂

pamelachia (approved):
just need to rmb to test org level and user level events

awaseem (approved):
LGTM on my end

PR #39748, opened 2025-10-21
Add realtime button A/B test with three variants (GROWTH-524)

seanoliver:
## What changed

We're running an A/B test on the Enable Realtime button in Table Editor to see if we can improve onboarding for new users. The experiment has three variants and only runs for projects created in the last 7 days.

**Variants:**
- `control` - Shows the normal Enable Realtime button (baseline)
- `hide-button` - Hides the button entirely
- `triggers` - Replaces button with "Manage triggers" link

Implementation uses a PostHog flag `realtimeButtonVariant` that returns one of those three values. Old projects (>7 days) always see the normal button regardless of flag value.

## Why limit to new projects

We only want to test behavior changes on users who haven't already learned the current flow. Changing UI on existing users would contaminate the data and annoy people. Projects older than 7 days are excluded from the experiment entirely - they always see the button, no tracking.

## What got added

**Experiment logic:**
- Check project age on load, only expose flag if ≤7 days old
- Track exposure once per session when viewing a table with all data loaded
- Hide or replace button based on variant (applies to grid header + side panel)

**New telemetry events:**
- `realtime_experiment_exposed` - Fires once when user sees table editor (includes variant, project age, current realtime state)
- `table_realtime_enabled` - Fires on successful enable (replaces old tracking)
- `table_realtime_disabled` - Fires on successful disable (new)

Refactored `realtime_toggle_table_clicked` and `table_rls_enabled` to use `useTrack()` hook instead of `useSendEventMutation()`.

**Constants and types:**
- `RealtimeButtonVariant` enum with three variants
- `NEW_PROJECT_THRESHOLD_DAYS = 7` for consistent project age checks
- TypeScript interfaces for all three new events

## Testing

Verified:
- Control variant shows button normally
- Hide variant removes button from both locations
- Triggers variant shows "Manage triggers" link
- Old projects always see button regardless of flag
- Exposure tracking fires once per session, only for new projects
- Enable/disable events fire on successful mutation
- All variants work in grid header and side panel

[GROWTH-524](https://linear.app/supabase/issue/GROWTH-524)

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
@@ -4,7 +4,7 @@ import Link from 'next/link'
 import { useState } from 'react'
 import { toast } from 'sonner'
 
-import { useFlag, useParams } from 'common'
+import { useParams } from 'common'
 import { RefreshButton } from 'components/grid/components/header/RefreshButton'
 import { getEntityLintDetails } from 'components/interfaces/TableGridEditor/TableEntity.utils'
 import { APIDocsButton } from 'components/ui/APIDocsButton'
@@ -22,12 +22,12 @@ import {
   isView as isTableLikeView,
 } from 'data/table-editor/table-editor-types'
 import { useTableUpdateMutation } from 'data/tables/table-update-mutation'
-import { useSendEventMutation } from 'data/telemetry/send-event-mutation'
 import { useAsyncCheckPermissions } from 'hooks/misc/useCheckPermissions'
 import { useIsFeatureEnabled } from 'hooks/misc/useIsFeatureEnabled'
-import { useSelectedOrganizationQuery } from 'hooks/misc/useSelectedOrganization'
+import { useRealtimeExperiment, RealtimeButtonVariant } from 'hooks/misc/useRealtimeExperiment'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
 import { useIsProtectedSchema } from 'hooks/useProtectedSchemas'
+import { useTrack } from 'lib/telemetry/track'
 import { DOCS_URL } from 'lib/constants'
 import { parseAsBoolean, useQueryState } from 'nuqs'
 import { useTableEditorTableStateSnapshot } from 'state/table-editor-table'
@@ -54,7 +54,7 @@ export interface GridHeaderActionsProps {
 export const GridHeaderActions = ({ table, isRefetching }: GridHeaderActionsProps) => {
   const { ref } = useParams()
   const { data: project } = useSelectedProjectQuery()
-  const { data: org } = useSelectedOrganizationQuery()
+  const track = useTrack()
 
   const [showWarning, setShowWarning] = useQueryState(
     'showWarning',
@@ -69,7 +69,6 @@ export const GridHeaderActions = ({ table, isRefetching }: GridHeaderActionsProp
   const isView = isTableLikeView(table)
   const isMaterializedView = isTableLikeMaterializedView(table)
 
-  const triggersInsteadOfRealtime = useFlag<boolean>('triggersInsteadOfRealtime')
   const { realtimeAll: realtimeEnabled } = useIsFeatureEnabled(['realtime:all'])
   const { isSchemaLocked } = useIsProtectedSchema({ schema: table.schema })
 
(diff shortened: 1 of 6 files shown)


pamelachia:
is there a reason why enabled and disabled events are included? Why don't we just have the click event to show if it was to enable or to disable?

packages/common/telemetry-constants.ts
```diff
@@ -422,6 +422,58 @@ export interface RealtimeToggleTableClickedEvent {
...
+    /**
+     * Schema name
+     */
+    schema_name: string
+    /**
+     * Table name
+     */
+    table_name: string
+  }
+  groups: Partial<TelemetryGroups>
```
pamelachia: wouldn't this be at a project level and we don't need to do Partial?

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
```diff
@@ -162,10 +176,31 @@ export const GridHeaderActions = ({ table, isRefetching }: GridHeaderActionsProp
...
+      isTable &&
+      hideRealtimeButton !== undefined
+    ) {
+      hasTrackedExposure.current = true
+
+      const daysSinceCreation = Math.floor(
+        (Date.now() - new Date(project.inserted_at).getTime()) / (1000 * 60 * 60 * 24)
+      )
+
+      track('realtime_experiment_exposed', {
```
pamelachia: on usePHFlag() we track Feature flag called and the flag and value so we would know when it was exposed. Do we really need the other properties in this event here?

packages/common/telemetry-constants.ts
seanoliver: Correct! Good catch!

apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
seanoliver: Good question! I compared results for the `tableQuickstart` experiment using both approaches (custom exposure event vs. built-in global exposure event) and found a significant discrepancy.

The built-in "Feature flag called" event from `usePHFlag()`:
  - Tracks ~46k exposures for `tableQuickstart`
  - Fires as soon as the flag value is evaluated, regardless of UI context
  - Can fire before project, org, or other required data loads
  - Counts users who may have navigated away before actually seeing the experiment

The custom exposure event:
  - Tracks ~12k exposures for `tableQuickstart`
  - Only fires when user is in the relevant context (viewing a table with all data loaded)
  - More accurately reflects users who were actually exposed to the treatment

The 4x difference suggests the built-in tracking over-counts because it fires before we know the user is actually at the specific point in the UI to see the experiment.

I'm torn here because I don't want to proliferate custom exposure events for every experiment. But looking at how `usePHFlag`'s tracking is implemented (lines 28-34 in `useFlag.ts`), it seems to be more designed for general feature flag monitoring, not A/B testing. It doesn't have the context-specific guards needed to ensure we only count a true exposure.

For experiments where the initial exposure happens deeply nested in the UI (like this one), I think we probably need a custom event. 

What do you think? If you agree I can document this pattern for future experiments? Or even better would be to make the built-in tracking more context aware?  Like maybe it accepts a contextual ref or something to ensure the user is actually seeing the relevant UI for the experiment before it fires?

image

image

awaseem:
@seanoliver is this still relevant?

seanoliver:
> @seanoliver is this still relevant?

@awaseem It is. I've just been incorporating some feedback from Saxon and preparing for review. I will post it in FE Slack shortly with a bit more context for the team!

awaseem:
@seanoliver if that is the case, can we move this draft for now. I'm just setting up a slack bot that lets the team know of PRs that are over 24 hours and need review. This PR is showing up, but if its not ready lets move to draft please 🙏

seanoliver:
@awaseem done! And i'll be sure to use draft mode moving forward until things are ready. Note that this should be ready to come out of draft mode imminently so don't be surprised if the slack bot soon picks it up! 🙏

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.tsx
```diff
@@ -396,23 +400,16 @@ export const TableEditor = ({
           </Admonition>
         )}
 
-        {realtimeEnabled && (
+        {activeRealtimeVariant !== RealtimeButtonVariant.HIDE_BUTTON && realtimeEnabled && (
```
awaseem, line 403: @seanoliver Curious for self hosted do we want to experiment with that? My guess is no. Should we just return the control value for self hosted envs. This would also help with our e2e tests

awaseem (changes requested):
Just a concern about self hosted. Let me know if you have any questions or concerns

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.tsx
seanoliver: @awaseem agreed and good catch! Just pushed this change: 43050a63421bf397c4b6161014b1af326f1c4e9d to return `null` (control experience) and skip tracking for self-hosted instances.

awaseem (approved):
LGTM

awaseem:
@seanoliver I'm fixing the tests, feel free to merge as they are related to your changes

pamelachia (approved):

PR #39754, opened 2025-10-22
Fix gap issue

SaxonF:
image

When visiting a page that is linked to an id e.g. `/billing#address` there is a small gap that appears at the bottom of the viewport. Strangely this appears to be due to our sidebar and specifically a div used to set width on desktop and its use of svh.


joshenlim (approved):
Verified on preview for both desktop and mobile

dnywh (approved):
Smoke tested on staging on desktop. Also tested on iOS 26 because I was worried about the `svh` ramifications with Safari’s wonky changing address bar. Seems okay.

PR #39755, opened 2025-10-22
Improve messaging in policy table rows

joshenlim:
## Context

Noticed this while reviewing another PR on the Policy page - the messaging here can be more granular in particular whether the schema is exposed through the API or not

## Before
Example here is for a custom schema, in particular the 2nd one which has RLS enabled but no policies
Messaging here is technically wrong since while true that no data will be selectable via API, its not because there's no policies, but because the schema isn't exposed through the API (via postgrest)

image

## After
Opting to be more specific that the data isn't selectable because schema isn't exposed via the API, and add a CTA to the API settings if users want to configure that

image


alaister (approved):
LGTM!

- Tested on preview

PR #39758, opened 2025-10-22
Chore/table editor stay on schema after creating table

joshenlim:
## Context

Just some minor improvements to the Table Editor that I came across while working on it

## Changes involved

- Creating a new table in a schema outside of the `public` schema should keep the selected schema as that schema (currently goes back to `public`)
  - Selected schema affects the list of tables shown in the Table Editor menu
- Small improvement to tab closing behaviour when closing a tab that's in focus
  - Will open the tab that's adjacent to it instead of defaulting to the first tab

## To test
- [ ] Verify that creating a table in a schema apart from `public` will stay on that schema
- [ ] Verify that tab closing behaviour for the tab that's in focus doesn't feel awkward


alaister (approved):
LGTM!

- Tested on preview

PR #39774, opened 2025-10-22
swap error rate for success rate and add tests

jordienr:
- swaps error rate to success rate
- adds tests 
- moves some operations to a separate file

## BEFORE
CleanShot 2025-10-22 at 15 26 03@2x

## AFTER

CleanShot 2025-10-22 at 15 25 23@2x


awaseem (approved):
Clean!

PR #39775, opened 2025-10-22
command events

fsansalvadore:
Add CommandMenu telemetry events in studio, docs and www:
- `command_menu_opened`
- `command_menu_input_typed`
- `command_menu_command_selected`


packages/ui-patterns/src/CommandMenu/api/CommandMenu.tsx
```diff
@@ -143,6 +161,47 @@ function CommandMenuTrigger({ children }: PropsWithChildren) {
   return childWithClickHandler
 }
 
+function CommandMenuTriggerInput({
```
fsansalvadore, line 163: I abstracted the Command input in order to have it consistent throughout apps.
The idea is to also add it in Studio in an upcoming pr.

packages/common/telemetry-constants.ts
```diff
@@ -1946,7 +1946,75 @@ export interface AuthUsersSearchSubmittedEvent {
...
+}
+
+/**
+ * User typed a search term in the command menu input.
+ *
+ * @group Events
+ * @source ui-patterns
+ * @page any
+ */
+export interface CommandMenuInputTypedEvent {
```
fsansalvadore: New `command_menu_input_typed` event triggered when typing in the command input (debounced every 500ms)

packages/common/telemetry-constants.ts
```diff
@@ -1946,7 +1946,75 @@ export interface AuthUsersSearchSubmittedEvent {
...
+}
+
+/**
+ * User opened the command menu.
+ *
+ * @group Events
+ * @source ui-patterns
+ * @page any
+ */
+export interface CommandMenuOpenedEvent {
```
fsansalvadore, line 1959: New `command_menu_opened` event distinguished by keyboard_shortcut or search_input trigger and by app.

packages/common/telemetry-constants.ts
```diff
@@ -1946,7 +1946,75 @@ export interface AuthUsersSearchSubmittedEvent {
...
+}
+
+/**
+ * User selected a command from the command menu.
+ *
+ * @group Events
+ * @source ui-patterns
+ * @page any
+ */
+export interface CommandMenuCommandSelectedEvent {
```
fsansalvadore, line 2007: New `command_menu_command_selected` event to see what users actually select.
It could help understand how users actually use the menu.

kemaldotearth (approved):
Great text size improvement :shipit:

apps/studio/hooks/misc/useStudioCommandMenuTelemetry.ts
```diff
@@ -0,0 +1,37 @@
...
+        },
+      }
+
+      sendEvent(eventWithGroups)
+    },
+    [projectRef, organization?.slug, sendEvent]
+  )
+
+  return { onTelemetry }
+}
```
seanoliver, line 40: Nice! FYI we're building a generic telemetry enrichment utility (#39745) that could eventually replace these app-specific hooks for consistency.

packages/ui-patterns/src/CommandMenu/api/CommandInput.tsx
```diff
@@ -50,8 +52,68 @@ const CommandInput = forwardRef<
...
+      }
+
+      if (debounceTimeoutRef.current) {
+        clearTimeout(debounceTimeoutRef.current)
+      }
+
+      debounceTimeoutRef.current = setTimeout(() => {
+        logTelemetryEvent(value)
+        previousValueRef.current = value
+      }, 500)
```
seanoliver: This is a nit but ideally the timeout value here is a named constant at the top of the file.

packages/ui-patterns/src/CommandMenu/api/CommandInput.tsx
```diff
@@ -50,8 +52,68 @@ const CommandInput = forwardRef<
...
   useEffect(() => {
     setInputValue(query)
+    previousValueRef.current = query
   }, [query])
 
+  // Get telemetry context
+  const telemetryContext = useCommandMenuTelemetryContext()
+
+  // Debounced telemetry tracking
+  const debounceTimeoutRef = useRef<NodeJS.Timeout>()
```
seanoliver: Is there a reason we're not using [`useDebounce`](https://github.com/supabase/supabase/blob/ac192a50244888800c6016d08e18c9c21cd8244d/packages/common/hooks/useDebounce.tsx) here instead? It would simplify this component and take care of any potential cleanup issues.

packages/ui-patterns/src/CommandMenu/api/CommandInput.tsx
charislam:
```suggestion
      }, 2500)
```

We can make this much longer, since it's not providing user feedback, and there's not much value in sending all the half-typed queries

charislam (approved):

packages/ui-patterns/src/CommandMenu/api/CommandInput.tsx
seanoliver: +1

packages/ui-patterns/src/CommandMenu/api/CommandProvider.tsx
```diff
@@ -37,13 +54,14 @@ const CommandShortcut = ({ openKey }: { openKey: string }) => {
       if (evt.key === openKey && usesPrimaryModifier && !otherModifiersActive) {
         evt.preventDefault()
         toggleOpen()
+        sendTelemetry('keyboard_shortcut')
```
seanoliver: It looks like `sendTelemetry` is always sending a `command_menu_opened` event even though the keyboard shortcut here is a toggle. So if the menu is already open and someone uses the keyboard shortcut to close it, we'll still send an `_opened` event.

You might want to consider checking the current state of the command menu and only sending when the menu is going from closed -> open.

packages/common/telemetry-constants.ts
```diff
@@ -81,7 +81,7 @@ export interface ConnectionStringCopiedEvent {
      */
     connectionMethod: 'direct' | 'transaction_pooler' | 'session_pooler'
   }
-  groups: TelemetryGroups
+  groups: Partial<TelemetryGroups>
```
seanoliver: Hmm... this feels a little overly broad. We're effectively changing `groups: TelemetryGroups` to `groups: Partial<TelemetryGroups>` across all ~70 existing telemetry events, which weakens type safety unnecessarily.

The command menu events need `Partial<TelemetryGroups>` because they fire from docs/www (which don't have project/org context). However, most existing events are Studio-only and should still require full context.

I'd recommend we only apply `Partial<TelemetryGroups>` to the three new command menu events and leave the existing ones as-is to maintain type safety.

Is there a specific reason we want to do this across the board?

packages/common/telemetry-constants.ts
seanoliver: Recommend `CommandMenuSearchSubmittedEvent` instead of `CommandMenuInputTypedEvent` to match existing patterns like `AuthUsersSearchSubmittedEvent`. Avoids introducing new terminology ("typed", "input") to the telemetry system.

seanoliver (changes requested):
Nice work! Solid telemetry implementation across all apps.

To main pieces of feedback are:

1. Event naming - see inline comments about maintaining consistency with existing patterns
2. Type scope - `Partial<TelemetryGroups>` change affects all events; recommend limiting to just the new command menu events

Bonus: Consider using `useDebounce` from `common` to simplify the debouncing logic!

packages/ui-patterns/src/CommandMenu/api/CommandInput.tsx
fsansalvadore: The only down side of this is that if the user selects a command before the timer ends, the "typed" event isn't triggered because the debounce is cleaned up on unmount.

apps/studio/hooks/misc/useStudioCommandMenuTelemetry.ts
fsansalvadore: Good to know!

packages/common/telemetry-constants.ts
fsansalvadore: Nice, thanks for the correction 👍

packages/common/telemetry-constants.ts
fsansalvadore: Removing all those as it was helping type used in ui-patterns (which wouldn't directly have prog and org). But agreed that it was too broad in the context of this pr.

packages/ui-patterns/src/CommandMenu/api/CommandProvider.tsx
fsansalvadore: Great catch, @seanoliver!

This is fixed now

fsansalvadore:
@seanoliver thanks for the review, I addressed and resolved all your comments.
Ready for review again ✌️

seanoliver (approved):
This looks great!

PR #39778, opened 2025-10-22
fix(billing): do not show auto-renewal banner on aws mp onboarding page for private offers

ecktoteckto:
On our AWS Marketplace onboarding page, we show a banner in case Auto-Renewal is disabled for the AWS contract. With this PR we don't show the banner in case the contract was created through a Private Offer. The Auto-Renewal feature does not exist for Private Offers.

Screenshot 2025-10-22 at 16 34 28


packages/api-types/types/platform.d.ts
```diff
@@ -6941,7 +6749,16 @@ export interface components {
       name: string
       token: string
       /** @enum {string} */
-      type: 'postgres' | 'bigquery' | 'webhook' | 'datadog' | 'elastic' | 'loki'
+      type:
```
kevcodez: You might have to remove these changes to not have TS errors

packages/api-types/types/platform.d.ts
```diff
@@ -4778,7 +4669,16 @@ export interface components {
       description?: string
       name: string
       /** @enum {string} */
-      type: 'postgres' | 'bigquery' | 'webhook' | 'datadog' | 'elastic' | 'loki'
+      type:
```
kevcodez: You might have to remove these changes to not have TS errors

kevcodez (approved):

PR #39779, opened 2025-10-22
Chore/improve error handling ux for insert before and remove child errors

joshenlim:
## Context

We frequently receive support tickets for the following 2 client exception errors
- `Failed to execute 'removeChild' on 'Node'`
- `Failed to execute 'insertBefore' on 'Node'`

These 2 particular errors are commonly known issues in React and Chrome ([here](https://github.com/facebook/react/issues/17256) and [here](https://github.com/facebook/react/issues/24865)), and our responses to such tickets are usually just to disable the Chrome translation feature.

Reckon we can reduce the friction in this flow but providing a more specific error handler for these 2 specific errors by providing clearer information about them upfront to reduce cognitive load, and enable users to self-debug better

## Changes involved

- Updated `GlobalErrorBoundaryState` specifically for handling `insertBefore` or `removeChild` client side errors
- Added `notranslate` meta tag to `_app`

## Before

image

## After

image

Clicking "Still stuck" will then still direct to the support form, but the error will be kept separate from the message body. Note that this applies to all other errors caught by `GlobalErrorBoundaryState` as well - reckon this makes sense as getting tickets with only just the browser errors ends up with more back and forth.

image

## To test

This is likely easiest to test locally with React Developer Tools, but we'll only be able to simulate triggering the ErrorBoundary, not the exact `insertBefore` error, can follow the guide [here](https://dev.to/ninariccimarie/how-to-trigger-react-error-boundary-with-react-developer-tools-1ccg) on how to do so


apps/studio/components/ui/ErrorBoundary/InsertBeforeRemoveChildErrorHandler.tsx
```diff
@@ -0,0 +1,87 @@
...
+        className="text-center text-xs text-foreground-lighter hover:text-foreground-light transition"
+        queryParams={{
+          category: SupportCategories.DASHBOARD_BUG,
+          subject: `Client error: Failed to execute '${isRemoveChildError ? 'removeChild' : 'insertBefore'}' on 'Node'`,
+          sid: sentryIssueId,
+          error: urlMessage,
+        }}
+      >
+        Still stuck?
+      </SupportLink>
```
awaseem, line 86: I thought we agreed to remove the ability to not create a support ticket on this? Might be mistake on it

apps/studio/components/ui/ErrorBoundary/GlobalErrorBoundaryState.tsx
```diff
@@ -0,0 +1,152 @@
...
+                <Button type="outline" onClick={() => router.reload()}>
+                  Reload dashboard
+                </Button>
+              ) : (
+                <Button type="outline" onClick={() => resetErrorBoundary()}>
+                  Return to dashboard
+                </Button>
+              )}
+            </div>
+          </>
```
awaseem: What do you think about creating this as its own component? It'll make the conditional in here a lot cleaner. We can name it something like `ClientSideExceptionError`

awaseem: We can also move methods like `handleClearStorage` into said component, reducing a bit of code here and making it a tad bit cleaner

apps/studio/components/ui/ErrorBoundary/GlobalErrorBoundaryState.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { useIsFeatureEnabled } from 'hooks/misc/useIsFeatureEnabled'
+import { Button, cn } from 'ui'
+import { Admonition } from 'ui-patterns'
+import CopyButton from '../CopyButton'
+import { InlineLinkClassName } from '../InlineLink'
+import { InsertBeforeRemoveChildErrorHandler } from './InsertBeforeRemoveChildErrorHandler'
+
+export type FallbackProps = {
+  error: unknown
+  resetErrorBoundary: (...args: any[]) => void
```
awaseem: nit: maybe unknown is better than any since we don't even reference the args? 


```suggestion
  resetErrorBoundary: (...args: unknown[]) => void
```

awaseem (approved):
Go to go, would like the suggested changes and thoughts

apps/studio/components/ui/ErrorBoundary/InsertBeforeRemoveChildErrorHandler.tsx
joshenlim: ahh yup - we'd still let the users open a support ticket, but just make sure that they enter something in the message field minimally (previously because we appended the error message in the message field, users could then just send the support ticket immediately without any other input which then likely leads to some back and forth for an error that's known and has resolutions on the user's end)

awaseem: Ahh got it!

apps/studio/components/ui/ErrorBoundary/GlobalErrorBoundaryState.tsx
joshenlim: gotchaaa pushed up the refactor 🙏 🙂

PR #39782, opened 2025-10-22
Add threshold value to persist sort by in auth users

joshenlim:
## Context

Another small improvement to improve the UX of the Auth Users page for most users

## Changes

- Opt to persist the "Sort by" value in the Auth Users page in local storage, ONLY IF the number of users is within an arbitrary threshold of 10,000. This value will then be loaded whenever the user lands on `/auth/users` (e.g from clicking on the dashboard)
  - This check is intentionally strict so if for some reason the count doesn't successfully load, we won't load the sort by value from local storage
- Also opt to skip the warning dialog when selecting "unified search" if number of users is within that same threshold

## To test
This is probably easier to test locally unless you have > 10,000 users. Can set `SORT_BY_VALUE_COUNT_THRESHOLD` to a low value of 10 for example
- [ ] Verify that sort by value gets persisted + loaded if user count is within threshold
- [ ] Verify that warning dialog for selecting "unified search" doesn't show up if user count is within threshold
- [ ] Verify that sort by value does not persist / loaded if user count is outside of the threshold
- [ ] Verify that warning dialog for selecting "unified search" still shows up if user count is outside threshold

### DIFF
--- a/apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
+++ b/apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
@@ -55,7 +55,7 @@ import { formatUserColumns, formatUsersData } from './Users.utils'
 import { UsersFooter } from './UsersFooter'
 import { UsersSearch } from './UsersSearch'
 
-const SORT_BY_VALUE_COUNT_THRESHOLD = 10
+const SORT_BY_VALUE_COUNT_THRESHOLD = 10_000
 
 export const UsersV2 = () => {
   const queryClient = useQueryClient()
--- a/apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
+++ b/apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
@@ -55,7 +55,6 @@ import { formatUserColumns, formatUsersData } from './Users.utils'
 import { UsersFooter } from './UsersFooter'
 import { UsersSearch } from './UsersSearch'
 
-// [Joshen] Arbitary threshold value for what's considered to be a lot of users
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -54,6 +55,9 @@ import { formatUserColumns, formatUsersData } from './Users.utils'
 import { UsersFooter } from './UsersFooter'
 import { UsersSearch } from './UsersSearch'
 
+// [Joshen] Arbitary threshold value for what's considered to be a lot of users
+const SORT_BY_VALUE_COUNT_THRESHOLD = 10_000
```
awaseem: nit: Don't think the comment is needed, variable defines it fine

```suggestion
const SORT_BY_VALUE_COUNT_THRESHOLD = 10_000
```

awaseem (approved):
Minor comment nit, looks good otherwise 🚀

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -134,6 +146,23 @@ export const UsersV2 = () => {
   const [isDeletingUsers, setIsDeletingUsers] = useState(false)
   const [showFreeformWarning, setShowFreeformWarning] = useState(false)
 
+  const { data: countData, isSuccess: isCountLoaded } = useUsersCountQuery(
```
alaister: Might be worth renaming this to indicate it's the **total** users count?

alaister (approved):
LGTM!

PR #39786, opened 2025-10-22
feat(studio): give action shortcut for persistent branches deletion

avallete:

Little UI improvement, there is multiples reports from user being confused when downgrading / attempting to delete their branches.

We don't allow deletion to persistent branches anymore. This aim to reduce the confusion with a clear call to action for user that want to cleanup their branches while preserving the intention of making it hard to delete a persistent branch by mistake.

If you click the `Delete` on a persistent branch you have an error response. Saying you can't delete persistent branch.

## What is the new behavior?

We now show a warning modal, explaining that the branch must be switched to preview first before deletion. With an easy way to perform the action:
Screenshot 2025-10-22 at 19 10 34


apps/studio/components/interfaces/BranchManagement/Overview.tsx
```diff
@@ -401,6 +428,7 @@ const MainBranchActions = ({ branch, repo }: { branch: Branch; repo: string }) =
           {repo ? (
             <Link passHref href={`/project/${projectRef}/settings/integrations`}>
               <DropdownMenuItem asChild className="gap-x-2">
+                {/** biome-ignore lint/a11y/useValidAnchor: href is passed above in the Link component */}
                 <a>Change production branch</a>
```
sweatybridge, line 426: does it mean the `<a>` tag is redundant here?

apps/studio/components/interfaces/BranchManagement/Overview.tsx
```diff
@@ -308,12 +313,20 @@ const PreviewBranchActions = ({
...
+              }
             }}
             onClick={(e) => {
-              e.stopPropagation()
-              onSelectDeleteBranch()
+              if (isPersistentBranch) {
+                setShowPersistentBranchDeleteConfirmationModal(true)
+              } else {
+                e.stopPropagation()
+                onSelectDeleteBranch()
```
sweatybridge: Shall we move this check inside `onSelectDeleteBranch` to reduce duplication?

sweatybridge (approved):
just some nitpicks

apps/studio/components/interfaces/BranchManagement/Overview.tsx
avallete: No it just means we allow the `<a>` tag can be without any `href`, the tag is still necessary for the styling of the dropdown to work properly.

apps/studio/components/interfaces/BranchManagement/Overview.tsx
avallete: Can't put it in there since it comes from a props. But I can make a wrapper around this to dedup 👍

apps/studio/components/interfaces/BranchManagement/Overview.tsx
ivasilov: `biome-ignore`? Does that work with `eslint`?

ivasilov (approved):

apps/studio/components/interfaces/BranchManagement/Overview.tsx
avallete: My bad, I thought we were using `biome` linter everywhere (this is a biome specific rule).

Reverted this change since eslint doesn't complain.

PR #39787, opened 2025-10-22
fix: add modal to realtime settings to warn users of disconnect

filipecabaco:

YES

Add modal to realtime settings to warn users of disconnect

https://github.com/user-attachments/assets/0a8ee12c-c846-478c-b64e-b1553f052064

### DIFF
--- a/apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
+++ b/apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
@@ -1,6 +1,7 @@
 import { zodResolver } from '@hookform/resolvers/zod'
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import Link from 'next/link'
+import { useState } from 'react'
 import { SubmitHandler, useForm } from 'react-hook-form'
 import { toast } from 'sonner'
 import * as z from 'zod'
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
```diff
@@ -636,7 +639,8 @@ export const RealtimeSettings = () => {
                   )}
                   <Button
                     type="primary"
-                    htmlType="submit"
+                    htmlType="button"
+                    onClick={() => setIsConfirmNextModalOpen(true)}
                     form={formId}
                     disabled={!canUpdateConfig || isUpdatingConfig || !form.formState.isDirty}
                     loading={isUpdatingConfig}
```
awaseem: This removes enter to submit on text form fields right? @ivasilov might know better. 

If thats okay then this should be fine and good to go

awaseem (approved):
Minor callout, not sure if its true though 😅

joshenlim (commented):
Opting to render the `ConfirmationModal` outside of the form, and letting the "Save changes" button remain as `htmlType=submit` so that it can trigger the RHF validations

Otherwise if the form values are failing validations (e.g connection pool value is 0), the form validation only kicks in within the confirmation modal. When this happens, field input gets highlighted in the background, but the modal is still opened and shows a loading state (as though its still trying to save)( the form's `onSubmit` will not be triggered in this case if any validation fails)

The many line changes is just the shift in indentation because of the fragment, but the flow will now be
- Click "Save changes" -> triggers RHF validation
  - If fail validation, show error in form
  - If pass, only then show confirmation modal

joshenlim (approved):
Verified changes on staging preview updating realtime settings!

apps/studio/components/interfaces/Realtime/RealtimeSettings.tsx
filipecabaco: @joshenlim might have fixed that also 😃 thank you for the callout 🙏

PR #39790, opened 2025-10-22
fix,perf(policies search)

charislam:
## Overview

Fixes two problems with the policy search:
1. Search field query keeps getting corrupted
2. Search is extremely laggy when user has hundreds of tables

### Buggy search problem

Try searching in staging and again in the preview site. In staging, when you type fast, only bits and pieces of the searched query are preserved. This should be fixed in preview.

### Laggy search problem

Add 700 tables to your project using the following SQL:

```
DO
$$
DECLARE
  target_schema  text := 'public';
  prefix         text := 'rnd';
  to_create      int  := 700;
  created        int  := 0;
  tbl_name       text;
BEGIN
  -- Ensure schema exists (no-op for public)
  EXECUTE format('CREATE SCHEMA IF NOT EXISTS %I;', target_schema);

  WHILE created < to_create LOOP
    -- Random-ish, reasonably short, valid identifier
    tbl_name := prefix || '_' || substr(
      md5(random()::text || clock_timestamp()::text || created::text),
      1, 10
    );

    BEGIN
      EXECUTE format(
        'CREATE TABLE %I.%I (
           id   BIGSERIAL PRIMARY KEY,
           name TEXT NOT NULL
         );',
        target_schema, tbl_name
      );
      created := created + 1;
    EXCEPTION
      WHEN duplicate_table THEN
        -- Rare collision: try again without incrementing 'created'
        CONTINUE;
    END;
  END LOOP;
END;
$$;
```

Take a performance profile while searching in both staging and preview. Notice the lag on staging and the number of extremely long tasks. (Lag is especially obvious when you backspace to clear your search.) Compare preview -- should be faster and much fewer, shorter long tasks.

## Breakdown by commit

[fix(policies search): replace useUrlState with nuqs](https://github.com/supabase/supabase/pull/39790/commits/8b329ee854a11b354825e678a38d7316b0130fa8) 
There is a bug when typing quickly into the search field on the policies
page: because useUrlState naively uses router.replace, which is
asynchronous, the search term can get corrupted, e.g., searching for
"dummy" might end up with a final value of "dum". Switched to nuqs for
better handling.

[perf(policies page): memoize and optimize for projects with many tables](https://github.com/supabase/supabase/pull/39790/commits/7c69ff15b2acf315b620cec399af93917ececfb0)
A project with hundreds of tables will have extremely slow search
performance for the policies page, because of heavy rerendering. Made
some optimizations:
- Memoized to reduce rerendering
- Hid filtered-out tables instead of unmounting, so we don't have to
remount when the search is cleared

[refactor(policies page): pass policies info through context](https://github.com/supabase/supabase/pull/39790/commits/e5e5c15b2b85fc580cad63890f49518cdb713593)
Avoid having a mess of props by creating a PoliciesDataContext to share
policy information.

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/Policies.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/Policies.tsx
@@ -1,7 +1,7 @@
 import type { PostgresPolicy } from '@supabase/postgres-meta'
 import { isEmpty } from 'lodash'
 import Link from 'next/link'
-import { useCallback, useMemo, useState } from 'react'
+import { useCallback, useState } from 'react'
 import { toast } from 'sonner'
 
 import { useParams } from 'common'
@@ -11,15 +11,11 @@ import {
 } from 'components/interfaces/Auth/Policies/PolicyTableRow'
 import { ProtectedSchemaWarning } from 'components/interfaces/Database/ProtectedSchemaWarning'
 import { NoSearchResults } from 'components/ui/NoSearchResults'
-import { useProjectPostgrestConfigQuery } from 'data/config/project-postgrest-config-query'
 import { useDatabasePolicyDeleteMutation } from 'data/database-policies/database-policy-delete-mutation'
 import { useTableUpdateMutation } from 'data/tables/table-update-mutation'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
-import type { ResponseError } from 'types'
 import { Button, Card, CardContent } from 'ui'
 import ConfirmModal from 'ui-patterns/Dialogs/ConfirmDialog'
-
-const EMPTY_POLICIES: PostgresPolicy[] = []
 
 interface PoliciesProps {
   search?: string
@@ -27,10 +23,6 @@ interface PoliciesProps {
   tables: PolicyTableRowProps['table'][]
   hasTables: boolean
   isLocked: boolean
-  policies: PostgresPolicy[]
-  isLoadingPolicies: boolean
-  isPoliciesError: boolean
-  policiesError?: ResponseError | Error
   visibleTableIds: Set<number>
   onSelectCreatePolicy: (table: string) => void
   onSelectEditPolicy: (policy: PostgresPolicy) => void
@@ -43,10 +35,6 @@ export const Policies = ({
   tables,
   hasTables,
   isLocked,
-  policies,
-  isLoadingPolicies,
-  isPoliciesError,
-  policiesError,
   visibleTableIds,
   onSelectCreatePolicy,
   onSelectEditPolicy: onSelectEditPolicyAI,
@@ -54,7 +42,6 @@ export const Policies = ({
 }: PoliciesProps) => {
   const { ref } = useParams()
   const { data: project } = useSelectedProjectQuery()
-  const { data: postgrestConfig } = useProjectPostgrestConfigQuery({ projectRef: project?.ref })
 
   const [selectedTableToToggleRLS, setSelectedTableToToggleRLS] = useState<{
     id: number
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/Auth/Policies/Policies.tsx
```diff
@@ -139,18 +149,26 @@ export const Policies = ({
...
+                    isLocked={schema === 'realtime' ? true : isLocked}
+                    onSelectToggleRLS={onSelectToggleRLS}
+                    onSelectCreatePolicy={handleCreatePolicy}
+                    onSelectEditPolicy={onSelectEditPolicy}
+                    onSelectDeletePolicy={onSelectDeletePolicy}
+                  />
+                </section>
+              )
+            })}
+            {Boolean(search) && visibleTableIds.size === 0 && (
```
awaseem: nit: should we use !! to keep it standard? 
```suggestion
            {!!search && visibleTableIds.size === 0 && (
```

apps/studio/components/interfaces/Auth/Policies/PoliciesDataContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+type TableKey = `${string}.${string}`
+
+interface PoliciesDataContextValue {
+  getPoliciesForTable: (schema: string, table: string) => PostgresPolicy[]
+  isPoliciesLoading: boolean
+  isPoliciesError: boolean
+  policiesError?: ResponseError | Error
+  exposedSchemas: Set<string>
+}
```
awaseem: nit: Some places we use types and other places we use interfaces? Thoughts on just standardizing this to types? 


```suggestion
type PoliciesDataContextValue = {
  getPoliciesForTable: (schema: string, table: string) => PostgresPolicy[]
  isPoliciesLoading: boolean
  isPoliciesError: boolean
  policiesError?: ResponseError | Error
  exposedSchemas: Set<string>
}
```

apps/studio/components/interfaces/Auth/Policies/PoliciesDataContext.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return context
+}
+
+interface PoliciesDataProviderProps {
+  policies: PostgresPolicy[]
+  isPoliciesLoading: boolean
+  isPoliciesError: boolean
+  policiesError?: ResponseError | Error
+  exposedSchemas: string[]
+}
```
awaseem: Same as above


```suggestion
type PoliciesDataProviderProps = {
  policies: PostgresPolicy[]
  isPoliciesLoading: boolean
  isPoliciesError: boolean
  policiesError?: ResponseError | Error
  exposedSchemas: string[]
}
```

apps/studio/components/interfaces/Realtime/Policies.tsx
```diff
@@ -26,7 +30,30 @@ export const RealtimePolicies = () => {
...
+    projectRef: project?.ref,
+    connectionString: project?.connectionString,
+  })
+  const exposedSchemas = useMemo(() => {
+    if (!postgrestConfig?.db_schema) return []
+    return postgrestConfig.db_schema
+      .split(',')
+      .map((schema) => schema.trim())
+      .filter((schema) => schema.length > 0)
+  }, [postgrestConfig?.db_schema])
```
awaseem, line 58: Thoughts on this?


```suggestion
const exposedSchemas = useMemo(() => {
  const dbSchema = postgrestConfig?.db_schema
  if (!dbSchema) return []

  return dbSchema
    .split(',')
    .map((schema) => schema.trim())
    .filter((schema) => schema.length > 0)
}, [postgrestConfig?.db_schema])

```

apps/studio/pages/project/[ref]/auth/policies.tsx
```diff
@@ -141,32 +201,25 @@ const AuthPoliciesPage: NextPageWithLayout = () => {
...
+              schema={schema}
+              tables={tablesWithVisibility}
+              hasTables={(tables ?? []).length > 0}
+              isLocked={isSchemaLocked}
+              visibleTableIds={visibleTableIds}
+              onSelectCreatePolicy={handleSelectCreatePolicy}
+              onSelectEditPolicy={handleSelectEditPolicy}
+              onResetSearch={handleResetSearch}
+            />
+          </PoliciesDataProvider>
```
Comment (awaseem) [inline apps/studio/pages/project/[ref]/auth/policies.tsx:222]: This feels a lot cleaner!

awaseem (approved):
Minor changes, let me know your thoughts and not a big deal

PR #39597, opened 2025-10-16
feat: Vector buckets - initial work

ivasilov:
> [!NOTE]
> Introduces Vector buckets (list/detail pages, create/delete bucket and table, data queries/mutations) and refactors Analytics buckets (page layout, deletion, connect tables), with feature flags and API type updates.
> 
> - **Storage – Vector Buckets (new)**:
>   - Pages: `pages/project/[ref]/storage/vectors` and `vectors/buckets/[bucketId]` with list/detail views.
>   - UI: `CreateVectorBucketDialog`, `CreateVectorTableSheet`, `DeleteVectorBucketModal`, `DeleteVectorTableModal`, `VectorBucketDetails`.
>   - Data layer: queries/mutations for vector buckets and indexes (`vector-buckets[-query]`, `vector-bucket[-query]`, `vector-buckets-indexes[-query]`, create/delete index & bucket), new cache keys.
>   - Navigation: `StorageMenuV2` shows Vectors (ALPHA) via feature flag `useIsVectorBucketsEnabled`.
>   - Empty state: supports vectors via `CreateVectorBucketDialog`.
> - **Storage – Analytics Buckets (refine)**:
>   - Convert details page to `PageLayout`; fetch selected bucket internally; improved error handling.
>   - Split deletion into `DeleteAnalyticsBucketModal`; `ConnectTablesDialog` now uses `bucketId` and validates inputs.
>   - Config/connection details components accept optional bucket name; wrapper instance hook gating.
> - **Config/Flags**:
>   - Add `useIsVectorBucketsEnabled` (feature flag) and surface ALPHA badges.
> - **API/types**:
>   - Update `api-types` (storage config fields, logs docs, restore status, vectors endpoints payloads, etc.).
> - **Misc/UI**:
>   - Add Studio UI rules (React Query + formId pattern); minor text class fix; simplify standard bucket deletion.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2d40bc703a86cd9a1d8f3f3c0c361769328f0105. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #39794, opened 2025-10-23
chore(studio): vector buckets polish B

dnywh:

Chore: UI more improvements to the vector buckets and their details

- Vector bucket page load  `GenericSkeletonLoader` is full-width
- The bucket deletion action happens instantly after tapping _Delete bucket_
  - On both the ‘outside’ (/storage/vectors) and ‘inside’ (storage/vectors/buckets/test)

## What is the new behavior?

- `GenericSkeletonLoader` is constrained to the loading of indexes, everything else should load instantly (just like analytics buckets)
- The bucket deletion action calls the `DeleteBucketModal`
  - On both the ‘outside’ (/storage/vectors) and ‘inside’ (storage/vectors/buckets/test)

### Known issues to fix

- [ ] I’ve introduced a bit of a regression with the `GenericSkeletonLoader` fix. Although it is now properly constrained to just the tables (indexes), the entire `VectorBucketDetails` is (presumedly) waiting for `isSuccess` from the parent `[bucketId]`. That means a momentary blank screen when going from ‘outside’ (/storage/vectors) to ‘inside’ (storage/vectors/buckets/test). I’m not sure why this is the case given analytics buckets work in the same way (without any flash). See https://github.com/supabase/supabase/pull/39610.
- [ ] I had to edit the `DeleteBucketModal` component to handle the custom delete _vector_ bucket mutation. This goes over my head, but I’m wondering if we can avoid this customisation since analytics buckets deletion works without any customisation.

## Additional context

Based branch: https://github.com/supabase/supabase/pull/39597.

## To test

Use `MOCK_VECTOR_INDEXES` for `allIndexes` in `VectorBucketDetails` ([here](https://github.com/supabase/supabase/blob/4394ca48f0ee3c2c0f4717b52ff08f639d65c8e3/apps/studio/components/interfaces/Storage/VectorBuckets/VectorBucketDetails.tsx#L67)) if you need mock table data.

### DIFF
--- a/apps/studio/components/interfaces/Storage/DeleteBucketModal.tsx
+++ b/apps/studio/components/interfaces/Storage/DeleteBucketModal.tsx
@@ -33,11 +33,17 @@ export interface DeleteBucketModalProps {
   visible: boolean
   bucket: Bucket
   onClose: () => void
+  onDelete?: (bucket: Bucket) => void
 }
 
 const formId = `delete-storage-bucket-form`
 
-export const DeleteBucketModal = ({ visible, bucket, onClose }: DeleteBucketModalProps) => {
+export const DeleteBucketModal = ({
+  visible,
+  bucket,
+  onClose,
+  onDelete, // Temporary prop for vector buckets specific deletion mutation
+}: DeleteBucketModalProps) => {
   const router = useRouter()
   const { ref: projectRef } = useParams()
   const { data: project } = useSelectedProjectQuery()
@@ -103,7 +109,13 @@ export const DeleteBucketModal = ({ visible, bucket, onClose }: DeleteBucketModa
   const onSubmit: SubmitHandler<z.infer<typeof schema>> = async () => {
     if (!projectRef) return console.error('Project ref is required')
     if (!bucket) return console.error('No bucket is selected')
-    deleteBucket({ projectRef, id: bucket.id, type: bucket.type })
+
+    // Temporary change for vector buckets specific deletion mutation
+    if (onDelete) {
+      onDelete(bucket)
+    } else {
+      deleteBucket({ projectRef, id: bucket.id, type: bucket.type })
+    }
   }
 
   return (
--- a/apps/studio/components/interfaces/Storage/VectorBuckets/VectorBucketDetails.tsx
+++ b/apps/studio/components/interfaces/Storage/VectorBuckets/VectorBucketDetails.tsx
@@ -34,6 +34,7 @@ import {
 } from 'ui'
 import { Input } from 'ui-patterns/DataInputs/Input'
 import { GenericSkeletonLoader } from 'ui-patterns/ShimmeringLoader'
+import { DeleteBucketModal } from '../DeleteBucketModal'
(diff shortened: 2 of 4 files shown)


apps/studio/components/interfaces/Storage/VectorBuckets/VectorBucketDetails.tsx
```diff
@@ -41,11 +43,16 @@ interface VectorBucketDetailsProps {
...
 export const VectorBucketDetails = ({ bucket }: VectorBucketDetailsProps) => {
+  const [modal, setModal] = useState<'delete' | null>(null)
   const { ref: projectRef } = useParams()
   const router = useRouter()
 
   // Use the correct query for bucket contents
-  const { data, isLoading, error } = useVectorBucketsIndexesQuery({
+  const {
+    data,
+    isLoading: isLoadingIndexes,
```
dnywh, line 53: Being specific about what’s loading so we can load the page and just put the loading state on the indexes (tables). Similar to https://github.com/supabase/supabase/pull/39610

apps/studio/components/interfaces/Storage/VectorBuckets/VectorsBuckets.tsx
```diff
@@ -150,6 +155,24 @@ export const VectorsBuckets = () => {
...
+      <DeleteBucketModal
+        visible={modal === 'delete'}
+        bucket={{
+          id: selectedBucket?.vectorBucketName || '',
+          name: selectedBucket?.vectorBucketName || '',
+          created_at: selectedBucket?.creationTime || '',
+          updated_at: selectedBucket?.creationTime || '',
+          owner: '',
+          public: false,
+          type: 'STANDARD' as const,
```
dnywh, line 168: I had to add these values to keep `DeleteBucketModal` happy. Not sure why, as analytics buckets work fine as-is. See https://github.com/supabase/supabase/pull/39610

apps/studio/components/interfaces/Storage/DeleteBucketModal.tsx
```diff
@@ -33,11 +33,17 @@ export interface DeleteBucketModalProps {
   visible: boolean
   bucket: Bucket
   onClose: () => void
+  onDelete?: (bucket: Bucket) => void
```
dnywh, line 36: Temporary prop for bespoke vector buckets deletion mutation

apps/studio/pages/project/[ref]/storage/vectors/buckets/[bucketId].tsx
```diff
@@ -25,7 +23,6 @@ const VectorsBucketPage: NextPageWithLayout = () => {
     <div className="storage-container flex flex-grow p-4">
       {isError && <StorageBucketsError error={error as any} />}
 
-      {isLoading && <GenericSkeletonLoader className="w-full" />}
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/vectors/buckets/[bucketId].tsx:28]: As mentioned elsewhere: I'm following a pattern from https://github.com/supabase/supabase/pull/39610 where we put the loading at the section level rather than blocking the page. That's the idea, anyway!

PR #39795, opened 2025-10-23
Sidebar manager

SaxonF:
This is the first step in introducing a generic sidebar space that Assistant, Inline Editor and Advisor will use. For this PR we are focusing on introducing sidebar-manager-state to manage sidebars, SidebarProvider to register these sidebars inc hotkey, and LayoutSidebar to render them. To start we migrate Assistant over to use this new sidebar manager which means removing all open/close logic out of the assistant state and any components that call it. 

To test:
-  Validate that Assistant can be toggled using top header button
- Validate Assistant can be toggled via cmd + i hotkey
- Validate all places we make use of Assistant across dashboard still open the Assistant


apps/studio/components/layouts/ProjectLayout/LayoutSidebar/index.tsx
```diff
@@ -0,0 +1,34 @@
+import { Fragment } from 'react'
+import { useSidebarManagerSnapshot } from 'state/sidebar-manager-state'
+import { ResizableHandle, ResizablePanel, cn } from 'ui'
+
+export const LayoutSidebar = () => {
+  const { activeSidebar } = useSidebarManagerSnapshot()
+
+  if (!activeSidebar) return null
+
```
charislam:
```suggestion
```

`activeSidebar?.component` will check for both

apps/studio/components/layouts/ProjectLayout/LayoutSidebar/index.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+describe('LayoutSidebar', () => {
+  beforeEach(() => {
+    routerMock.setCurrentUrl('/projects/default')
+  })
+
+  afterEach(() => {
+    resetSidebarManagerState()
+    localStorage.clear()
+    vi.clearAllMocks()
+    routerMock.setCurrentUrl('/projects/default')
```
charislam: Does this need to be done in both `beforeEach` and `afterEach`?

apps/studio/components/layouts/ProjectLayout/LayoutSidebar/index.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+}))
+
+const resetSidebarManagerState = () => {
+  Object.keys(sidebarManagerState.sidebars).forEach((id) => {
+    sidebarManagerState.unregisterSidebar(id)
+  })
+  sidebarManagerState.closeAll()
+}
+
+describe('LayoutSidebar', () => {
```
charislam, line 22: nice 😄

apps/studio/components/layouts/ProjectLayout/LayoutSidebar/LayoutSidebarProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+  useRegisterSidebar(SIDEBAR_KEYS.AI_ASSISTANT, () => <AIAssistant />, {}, 'i')
+
+  const router = useRouter()
+  const { openSidebar } = useSidebarManagerSnapshot()
+
+  // Handle sidebar URL parameter
+  useEffect(() => {
+    if (!router.isReady) return
+
+    const sidebarParam = router.query.sidebar
```
charislam, line 20: Ideally we would have enum validation on this, but can do without for now

apps/studio/components/layouts/ProjectLayout/ProjectLayout.tsx
```diff
@@ -227,25 +216,9 @@ export const ProjectLayout = forwardRef<HTMLDivElement, PropsWithChildren<Projec
...
-                        'w-screen h-[100dvh]',
-                        'md:absolute md:h-auto md:w-3/4',
-                        'xl:relative xl:border-l-0'
-                      )}
-                    >
-                      <AIAssistant className="w-full h-[100dvh] md:h-full max-h-[100dvh]" />
-                    </ResizablePanel>
-                  </>
-                )}
+                <LayoutSidebarProvider>
```
charislam, line 218: so much cleaner 🤩

apps/studio/components/ui/AIAssistantPanel/AIAssistant.tsx
```diff
@@ -457,11 +460,13 @@ export const AIAssistant = ({ className }: AIAssistantProps) => {
         },
       ]}
     >
-      <div className={cn('flex flex-col h-full', className)}>
+      <div
+        className={cn('flex flex-col h-full w-full h-[100dvh] md:h-full max-h-[100dvh]', className)}
```
charislam: I think the `h-[100dvh]` is clobbering the `h-full` here?

apps/studio/state/sidebar-manager-state.tsx
```diff
@@ -0,0 +1,189 @@
...
+  onClose?: () => void
+}
+
+type ManagedSidebar = SidebarHandlers & {
+  id: string
+  component: () => ReactNode | undefined
+}
+
+type SidebarManagerData = {
+  sidebars: Partial<Record<string, ManagedSidebar>>
```
charislam, line 17: Ideally we should have an enum for this instead of string, but not critical

apps/studio/state/sidebar-manager-state.tsx
```diff
@@ -0,0 +1,189 @@
...
+const createSidebarManagerState = () => {
+  const state: SidebarManagerState = proxy({
+    ...INITIAL_SIDEBAR_MANAGER_DATA,
+
+    registerSidebar(
+      id: string,
+      component: () => ReactNode | undefined,
+      handlers: SidebarHandlers = {}
+    ) {
+      const existing = state.sidebars[id]
```
charislam:
```suggestion
      const existing = !!state.sidebars[id]
```

nit: only used to check existence, so this will make intention clearer

apps/studio/state/sidebar-manager-state.tsx
```diff
@@ -0,0 +1,189 @@
...
+      state.sidebars[id] = {
+        id,
+        component,
+        ...handlers,
+      }
+
+      const panel = state.sidebars[id]
+      if (!panel) return
```
charislam: by this point, it should be impossible for `panel` to be falsy, iinw

apps/studio/state/sidebar-manager-state.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+      if (state.activeSidebar?.id === id) {
+        panel.onOpen?.()
+      } else if (!existing) {
+        panel.onClose?.()
```
charislam: out of curiosity, what case is this catching? under what circumstances is it possible for this to be open if it didn't exist before?

apps/studio/state/sidebar-manager-state.tsx
```diff
@@ -0,0 +1,189 @@
...
+      panel.onClose?.()
+      state.activeSidebar = undefined
+      return
+    },
+
+    isSidebarOpen(id: string) {
+      return state.activeSidebar?.id === id
+    },
+
+    closeAll() {
```
charislam: This name is a little misleading IMO. `closeAll` suggests it's possible for more than one to be open, but IIUC only one can be open, and this closes it. Maybe `closeCurrent`?

charislam (commented):
Really nice! Liking how clean this will be. Pushed a tiny commit to add a missing `openSidebar` on the new homepage.

apps/studio/state/sidebar-manager-state.tsx
SaxonF: @charislam curious how you would approach this. I was hoping to keep sidebar-manager-state unaware of which sidebars are making use of it but would be nice to enforce. Make use of sidebar keys from provider?

charislam: Good point, maybe it needs to be generic and then you provide the keys when you call it? Like `createSidebarManager<SidebarKeys>()`. Then the actual creation call, and the snapshots, could be moved up to the provider level, which knows what the keys are.

Not a big deal either way, just a possibility

charislam (approved):

PR #39797, opened 2025-10-23
Fix region empty after selecting cloud provider

joshenlim:
## Context

Switching cloud providers in the new project page causes the region field to be empty

Realised this was happening as `useOrganizationAvailableRegionsQuery` was a static value depdning on `defaultProvider` for its `cloudProvider` parameter.

Separately the new project page has gotten rather big and complex so I'll look into separately to refactor that page into smaller components

## To test

- [ ] Verify that changing cloud providers should update the db region as per what's returned in the `useOrganizationAvailableRegionsQuery` (i don't think there's any difference from the API atm)
- [ ] Verify that you can create a new project


alaister (approved):
LGTM!

PR #39819, opened 2025-10-23
fix(support form): bug with log snapshot

charislam:
The log snapshot is properly cut off at support form navigation in the preview, but not when it is actually uploaded. That happens because we try to take ownership of the current log snapshot twice, so the second time it falls back to the non-snapshotted logs. Changed to take ownership of the snapshot once when the support form is mounted and then pass it to child components/handlers.


awaseem (approved):
tested in staging, looks good!

PR #39826, opened 2025-10-23
perf(storage menu): improve bucket list for users with many thousand buckets

charislam:
The bucket list breaks for users with several thousand buckets.

We've brought up the issue with the Storage team, and they're working on pagination. In the meantime, this PR virtualizes and memoizes to improve performance, making the UI at least usable.


awaseem (approved):
LGTM, was able to test the buckets view but not with a larger list 😅

PR #39829, opened 2025-10-23
Fix #39827

### LINKED ISSUE #39827: Studio storage UI failing authentication with resumable endpoint
# Bug report

## Describe the bug

Using the storage GUI in Supabase Studio for file uploads is currently broken.

After an investigation this is what I've found:

Studio doesn't send the `Bearer ${token}` that the storage server expects, it sends `Basic ${credentials}` instead which causes an `Access Denied` error. Kong protects the Studio with Basic Auth, but it's very strange that the Studio itself is using Basic instead of Bearer in the `Authorization` header to communicate with Storage. I believe this to be the mechanism of the bug, and in a quick test locally, setting `Authorization: Bearer ${temp_token}` fixed the bug:

```ts
onBeforeRequest: async (req) => {
  try {
    const data = await getTemporaryAPIKey({ projectRef: state.projectRef })
    req.setHeader('apikey', data.api_key)
    req.setHeader('Authorization', "Bearer " + data.api_key) // experimental (WORKS!)
  } catch (error) {
    throw error
  }
},
```

But also, curiously, in the hosted platform it doesn't send any `Authorization` header at all. Just `Apikey`. As far as I tell from a brief look at the storage server `jwt` plugin, it expects `Authorization` and `Apikey` shouldn't work:

```ts
// storage/src/http/plugins/jwt.ts

fastify.addHook('preHandler', async (request) => {
      request.jwt = (request.headers.authorization || '').replace(BEARER, '')
```

yet it does work in the hosted platform with just `Apikey: ${temp_key}`

So there's 2 mysteries here:
- Why does the self-hosted Studio set `Authorization: Basic ${temp_key}` when the hosted platform doesn't?
- Will the storage server look at `headers.apikey` in the absence of `headers.authorization`?

If the answer to the second question is yes, then we need to look at simply preventing the setting of `Authorization` on the TUS request.

If the answer is no, then we can use the fix I've just pasted above.

I think we really ought to understand why an `Authorization` header of type `Basic` is somehow making it into the TUS request. Totally perplexing.

## To Reproduce

1. Start up a self-hosted Supabase instance, either via the CLI or the method I was personally using when I found the bug: `git clone https://github.com/BenIsenstein/pgonrails.git && cd pgonrails && ./setup.sh && docker compose up`
2. Attempt to upload a file to storage using the Studio GUI

## Expected behavior

Uploads via the GUI are consistently successful when within the size limits, MIME types etc that have been configured.

## Screenshots

None.

## System information

- OS: MacOS Ventura
- Browser: Chrome
- Version of Node.js: latest self-hosted (22 slim)

## Additional context

None

BenIsenstein:

YES

Bug fix

Studio storage UI failing authentication with resumable endpoint

https://github.com/supabase/supabase/issues/39827

## What is the new behavior?

Before initiating a resumable upload, the TUS upload `onBeforeRequest` handler adds a correct `Authorization` header with the temporary token. Overriding the previous behaviour of a `Basic` token that used the dashboard credentials, with the proper `Bearer` token.

## Additional context

It is still unknown how or why an `Authorization` header with the value of `Basic ${base64(USER:PASSWORD)}` was being set on the TUS resumable upload request. Nowhere in the code is it explicitly performed. We might want to look into default behaviour of the `tus-js-client` looking into browser cookies, still a complete mystery. In any case, this PR fixes the Auth header issue for the storage UI.

### DIFF
--- a/apps/studio/state/storage-explorer.tsx
+++ b/apps/studio/state/storage-explorer.tsx
@@ -1198,6 +1198,9 @@ function createStorageExplorerState({
                 try {
                   const data = await getTemporaryAPIKey({ projectRef: state.projectRef })
                   req.setHeader('apikey', data.api_key)
+                  if (!IS_PLATFORM) {
+                    req.setHeader('Authorization', `Bearer ${data.api_key}`)
+                  }
                 } catch (error) {
                   throw error
                 }


awaseem:
@BenIsenstein I tested this in the studio and it seems to work, once all the CI passes we can merge this!

awaseem (approved):
LGTM

awaseem:
Coveralls.io going through heavy load right now 🙈

BenIsenstein:
A fellow Calgarian! Nice to meet you Ali :)

awaseem:
Likewise, thanks for the contribution to Supabase 💚

PR #39834, opened 2025-10-24
Editor sidebar manager

SaxonF:
This is a follow up to our refactoring of sidebars PR https://github.com/supabase/supabase/pull/39795

This focuses on moving the EditorPanel over to the new sidebar manage state. 

In the process this removes some of the opinionated parts of the editor and how it connects with the rest of the dashboard. It is simply an SQL editor that other parts of the dash can make use of. We currently have a feature preview "Directly edit database entities" which opens this inline editor instead of the standard UI Sheet when creating/editing triggers, functions, policies. The intention is for this to be more of an advanced setting that experienced users may prefer to use. 

With the advancements of LLM it may also become a more approachable experience via inline assistant / code generation.


apps/studio/components/ui/EditorPanel/EditorPanel.tsx
```diff
@@ -28,89 +29,51 @@ import {
...
-  onRunSuccess?: (value: any[]) => void
-  onRunError?: (value: any) => void
-  functionName?: string
-  templates?: Template[]
-  initialPrompt?: string
-  onChange?: (value: string) => void
-  isInlineEditorHotkeyEnabled?: boolean
-}
+  const label = 'SQL Editor'
+  const isInlineEditorHotkeyEnabled = true
```
charislam: We'll have to hook this up properly to the user preference so it'll keep the hot key disabled if they don't want it

apps/studio/components/ui/EditorPanel/EditorPanel.tsx
```diff
@@ -171,286 +131,237 @@ export const EditorPanel = ({
...
+            onCancel={() => setShowWarning(undefined)}
+            onConfirm={() => {
+              setShowWarning(undefined)
+              onExecuteSql(true)
+            }}
+          />
+        )}
+
+        {results !== undefined && results.length > 0 && (
+          <div className={`max-h-72 shrink-0 flex flex-col ${showResults && 'h-full'}`}>
```
charislam:
```suggestion
          <div className={cn(`max-h-72 shrink-0 flex flex-col`, showResults && 'h-full')}>
```

charislam (changes requested):

charislam (approved):

PR #39835, opened 2025-10-24
Improve database functions list

SaxonF:
image

- Adds return type and security filters
- truncates cells
- If function has trigger return type, links to triggers list with function name as search. Extended triggers list search to include function_name.


alaister (approved):
LGTM!

- Tested on preview

PR #39837, opened 2025-10-24
chore: respect catalog version for postgrest-js

mandarini:

YES

`@supabase/postgrest-js` should respect catalog version.

By keeping version to `*`, it would resolve older version from `pnpm-lock.yaml`, so the lock file would have two different versions for `postgrest-js`.

## What is the new behavior?

`@supabase/postgrest-js` version is set in `catalog`, and the `ui-library` version respects that. It follows the convention already in place for version management.


mandarini:
Ref: https://github.com/supabase/postgrest-js/commit/6705c5c5f3b600e3e8cc74bc063c86ba90cb456c

avallete (approved):

Issue #39791, opened 2025-10-22
This specific section of the documentation contains a typo.


DominicNikolai:
# Improve documentation

[## Link ](https://supabase.com/docs/reference/javascript/range)

A typo was found in this part of the documentation

## Image
<img width="1893" height="851" alt="Image" src="https://github.com/user-attachments/assets/e7e4f97a-b3dc-41c1-8b34-0606a008bd6e" />

BenHavis:
Just to clarify — the typo seems to refer to the mismatch between the two code examples in this section.
The JavaScript snippet uses .from('countries'), but the SQL example below creates a characters table.
Updating one of them for consistency should resolve the confusion.

marsou001:
Good catch! I'm on it.

marsou001:
I just create [a PR](https://github.com/supabase/supabase/pull/39840) for it! @ChrisChinchilla

Hallidayo:
Closing as PR merged. Thanks @marsou001 !

PR #39840, opened 2025-10-24
fix(docs): change 'countries' to 'characters' to match JS example to data source

marsou001:
…data source

YES

Docs update.

The JavaScript example and the Data source don't reference the same table. The JavaScript example should fetch rows from the 'characters' table, just like the Data source, because the response shows data from 'characters', not from 'countries'.

## What is the new behavior?

The JavaScript example and the Data source match now.

## Additional context

This PR is solving [this issue](https://github.com/supabase/supabase/issues/39791)

Here is a screenshot illustrating the issue:

mimtch

### DIFF
--- a/apps/docs/spec/supabase_js_v2.yml
+++ b/apps/docs/spec/supabase_js_v2.yml
@@ -6571,7 +6571,7 @@ functions:
         code: |
           ```ts
           const { data, error } = await supabase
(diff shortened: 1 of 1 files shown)


ChrisChinchilla (approved):
Great, thanks! @marsou001

PR #39848, opened 2025-10-24
chore: compute Instances entitlements

ignaciodob:
This PR changes the way that we check if the user is allowed to change the compute instance of a project from checking if the user is on the Free Plan to an entitlement check.

### Testing
- Head over to `/project/_/settings/compute-and-disk`
- Assert that an Org on the Free Plan cannot change the Compute Size.
- Assert that an Org on the Free Plan sees the "Compute and Disk configuration is not available on the Free Plan" notice
- Assert that an Org on the Pro Plan and above can change the Compute Size.

### Optional testing with backend feature catalog changes
- Run your local backend local environment and local studio
- Modify [this line](https://github.com/supabase/infrastructure/blob/develop/api/packages/core/src/lib/entitlements/catalog.ts#L40) to `enabled: true` to enable updating compute sizes on the Free Plan.
- Assert that you're able to select a compute size update on the frontend


kevcodez (commented):
Before starting to actually use entitlements in the frontend, let's ensure two things

- On a plan change or any change that would invalidate entitlements, we need to invalidate the react-query cache
- Have a higher stale time to avoid refetching (let's do 30m-60m)

ignaciodob:
> Before starting to actually use entitlements in the frontend, let's ensure two things
> 
> * On a plan change or any change that would invalidate entitlements, we need to invalidate the react-query cache
> * Have a higher stale time to avoid refetching (let's do 30m-60m)

Addressed in https://github.com/supabase/supabase/pull/39905

kanishkdudeja (approved):
Tested and it works as expected.

kevcodez (approved):

PR #39850, opened 2025-10-24
Fix: infinite re-render on table when computing foreign keys

### LINKED ISSUE #39833: Bug Add foreign keys
i cant select column 

Image

awaseem:

YES

Bug fix for https://github.com/supabase/supabase/issues/39833

When loading the table editor, this is causing an infinite re-render because of the following:

- This .filter() call creates a new array reference on every single render
- Even if the actual data hasn't changed, foreignKeys is a brand new array each time
- React's useEffect uses shallow comparison on dependencies
- It sees foreignKeys as "different" on every render (new reference) and re-runs the effect

## What is the new behavior?

no more unused re-renders 🙈

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.tsx
@@ -1,6 +1,6 @@
 import type { PostgresTable } from '@supabase/postgres-meta'
 import { isEmpty, isUndefined, noop } from 'lodash'
-import { useEffect, useState } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 import { toast } from 'sonner'
 
 import { DocsButton } from 'components/ui/DocsButton'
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.tsx
```diff
@@ -141,14 +141,18 @@ export const TableEditor = ({
     (constraint) => constraint.type === CONSTRAINT_TYPE.PRIMARY_KEY_CONSTRAINT
   )
 
-  const { data: foreignKeyMeta, isSuccess: isSuccessForeignKeyMeta } =
+  const { data: foreignKeyMeta = [], isSuccess: isSuccessForeignKeyMeta } =
```
charislam, line 144: Nit: Correct me if I'm wrong, but won't `[]` be a new instance every render, so the fallback empty array is never memoized in practice? It doesn't currently cause infinite rerender because of the way we have useEffects set up, but it could be a footgun since at first glance it _looks_ memoized

charislam (approved):
Verified fix works in preview

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/TableEditor.tsx
awaseem: You are right here, let me revert this small change back and post a new PR

PR #39857, opened 2025-10-25
fix: confirmation modal on paid org creation

kevcodez:
The submit handler is triggered both on initially clicking the "Create organization" button and on the confirmation button, but it always early returns if the customer has other free plan orgs with projects, so it doesn't let the customer submit the form.

How to reproduce:
- Create a free plan org with an active project
- Create a new paid plan organization
- Confirmation modal appears
- Clicking the button just closes the modal and does not create the org as there is an early return in the onSubmit handler

Moved away from ConfirmationModal to an inline info instead

Screenshot 2025-10-25 at 13 00 16


joshenlim (approved):
Verified on staging preview! Tested creating free, pro, team org

PR #39869, opened 2025-10-25
chore: user v2 content count query for search

kevcodez:
Instead of using two different endpoint versions, we can rely on v2 for all cases.

### DIFF
--- a/apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
+++ b/apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
@@ -52,8 +52,6 @@ export const SearchList = ({ search }: SearchListProps) => {
   )
   const totalNumber = count ? count.private + count.shared : 0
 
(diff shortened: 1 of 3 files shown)


apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
```diff
@@ -45,13 +45,14 @@ export const SearchList = ({ search }: SearchListProps) => {
...
-      cumulative: true,
       type: 'sql',
       name: search,
     },
     { keepPreviousData: true }
   )
-  const totalNumber = (count as unknown as { count: number })?.count ?? 0
+  const totalNumber = count ? count.private + count.shared : 0
+
+  console.log({ totalNumber })
```
awaseem: Not needed, we can remove this
```suggestion

```

awaseem (approved):
LGTM after console log is removed!

joshenlim (approved):
verified changes on staging preview RE count number tallies when searching in SQL Editor

PR #39875, opened 2025-10-25
fix: revert 39818

### LINKED ISSUE #39872: Can't edit or delete rows
# Bug report

Project: Colorado Springs Brewery App
Studio Build / Commit: 977b867bef6819077c7c306f3d4c07ac0557d9ab (deployed 2025-10-25 06:17-06:00 America/Denver)
Browser(s) Tested: [Chrome X.Y / Firefox X.Y / Safari X.Y]
OS: [Windows 10/11, macOS etc]
Environment: Hosted (Supabase managed) — not self-hosted.
Affected Tables: public.device_tokens, public.beer_releases (view), public.events_table (view)
Summary:

Inline edits (clicking a cell and editing) succeed.

Right-click → Edit row and Delete row actions in the Table Editor fail silently or show no effect.

In the browser console, the following error is logged:

Uncaught TypeError: Cannot destructure property 'rowIdx' of 'r' as it is undefined

No API request appears in the Network tab for the right-click actions (indicates the UI crashes before sending the request).

Schema review: device_tokens is a table, has PK id, the key columns device_token and platform are NOT NULL without defaults. RLS policy for DELETE was missing — we added a DELETE policy + a BEFORE UPDATE trigger to handle NULLs, but the UI bug still blocks the context-menu paths.

The issue reproduces on multiple machines/browsers.

Steps to reproduce:

Open Supabase Studio → “Table Editor”.

Select table public.device_tokens.

Right-click on any row → choose Edit row → attempt a change → no effect, console error appears.

Alternatively: right-click → Delete row → no effect, console error appears.

Inline edit a cell → change value → save → works.

From Network tab: no API call is made during the context‐menu edit/delete.

Schema, policies and PKs have been verified correct; other edit paths (SQL editor, REST API) succeed.

Expected behaviour:

Right-click → Edit row should open the row editor dialog, allow changes, submit a PATCH and succeed (given proper permissions).

Right-click → Delete row should issue DELETE request and remove the row (given permissions).

No front-end crash should occur; rowIdx property should be defined.

Actual behaviour:

UI crashes or silently fails with rowIdx undefined.

No API request is made for the failed action.

Only inline edit path works.

Impact:

Limits usability of Supabase Studio’s Table Editor for row-level context menu actions.

Requires using SQL editor or REST API manually for edits/deletes.

Degrades developer experience and slows workflows.

Additional details / troubleshooting done:

Verified schema: device_tokens table has primary key id, columns device_token and platform are NOT NULL without defaults.

Verified RLS/policies: DELETE policy was missing — added CREATE POLICY device_tokens_delete_authenticated … FOR DELETE TO authenticated USING (true).

Created BEFORE UPDATE trigger to coalesce nulls for NOT NULL columns.

Checked pg_class.relrowsecurity = false for the table; relforcerowsecurity = false.

Browser DevTools console shows the JS error listed above.

Network tab shows no request for the failed context-menu actions.

Works via SQL editor (UPDATE/DELETE) and via REST API.

Suggested fix / reasoning:

The error indicates a front-end bug: the variable r (likely the row object) is undefined or missing rowIdx in the context menu code path.

Possibly caused by the table-row object not being passed correctly to the context-menu handler.

The fix likely requires the Studio team to guard against r being undefined (and revert to inline edit mode) or correct how the row index is passed to the menu component.

Request:
Please investigate this issue in Supabase Studio. Given the commit date (2025-10-25) and repeatable repro across browsers/cloud projects, it may be a regression.
Could you:

Confirm a bug ticket for this exists (or open one if none).

Provide an ETA or note when fix will be merged.

Provide a workaround or flag (temporary) so right-click Edit/Delete works while fix is pending.

Hallidayo:

YES

Supabase Studio > Table Editor > Right Click Row

The right click options on the table editor do not work and console error.

## What is the new behavior?

Reverted back and right click options work

## Additional context

Closes #39872

### DIFF
--- a/apps/studio/components/grid/components/menu/RowContextMenu.tsx
+++ b/apps/studio/components/grid/components/menu/RowContextMenu.tsx
@@ -75,20 +75,17 @@ export const RowContextMenu = ({ rows }: RowContextMenuProps) => {
         <Copy size={12} />
         <span className="ml-2 text-xs">Copy row</span>
       </Item>
-      {snap.editable && (
-        <>
-          <DialogSectionSeparator className="my-1.5" />
-          <Item onClick={onEditRowClick} data="edit">
-            <Edit size={12} />
-            <span className="ml-2 text-xs">Edit row</span>
-          </Item>
-          <DialogSectionSeparator className="my-1.5" />
-          <Item onClick={onDeleteRow} data="delete">
-            <Trash size={12} />
-            <span className="ml-2 text-xs">Delete row</span>
-          </Item>
-        </>
-      )}
+
+      <DialogSectionSeparator className="my-1.5" />
+      <Item onClick={onEditRowClick} data="edit">
+        <Edit size={12} />
+        <span className="ml-2 text-xs">Edit row</span>
+      </Item>
+      <DialogSectionSeparator className="my-1.5" />
+      <Item onClick={onDeleteRow} data="delete">
+        <Trash size={12} />
+        <span className="ml-2 text-xs">Delete row</span>
+      </Item>
     </Menu>
   )
 }


awaseem (approved):
LGTM, tested and works!

PR #39879, opened 2025-10-25
chore: storage settings entitlements

ignaciodob:
This PR changes how we check the `Global file size limit` in Storage Settings to use the Entitlements API.

### Changes
- Adds new methods to the `useCheckEntitlements` hook to better expose the underlying entitlement configuration without the client having to be aware of the configuration structure or details.
- Adds the hook in `StorageSettings` to replace the current plan checks with the appropriate entitlement.

### Testing
- Head over to `project/_/storage/settings`
- With an Org in the Free Plan, assert that you are not able to edit the `Global Size Limit`
- With an Org on the Pro Plan or above, assert that the max `Global Size Limit` is 500GB. Anything over it should show a client-side validation error, `Maximum limit is up to 500 GB.`

### DIFF
--- a/apps/studio/hooks/misc/useCheckEntitlements.ts
+++ b/apps/studio/hooks/misc/useCheckEntitlements.ts
@@ -3,6 +3,20 @@ import { useMemo } from 'react'
 import { useSelectedOrganizationQuery } from './useSelectedOrganization'
 import type { EntitlementConfig } from 'data/entitlements/entitlements-query'
 
+function isNumericConfig(
+  config: EntitlementConfig
+): config is { enabled: boolean; unlimited: boolean; value: number } {
+  return 'unlimited' in config && 'value' in config
+}
+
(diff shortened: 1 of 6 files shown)


apps/studio/hooks/misc/useCheckEntitlements.ts
```diff
@@ -29,27 +50,48 @@ export function useCheckEntitlements(
...
   const isSuccess = shouldGetSelectedOrg
     ? isSuccessSelectedOrg && isSuccessEntitlements
     : isSuccessEntitlements
 
-  return { hasAccess, entitlementConfig, isLoading, isSuccess }
+  return {
+    hasAccess,
+    isLoading,
+    isSuccess,
+    getEntitlementNumericValue: () =>
```
kevcodez: Instead of inlining them, let's extract them to a regular function?

kevcodez (approved):

PR #39891, opened 2025-10-27
fix: use Next Link for proper prefixing

kevcodez:
<a> does not automatically use the `/dashboard` prefix on prod/staging, leading to wrong links for org restriction banner


alaister (approved):
LGTM!

PR #39892, opened 2025-10-27
Moves inline editor from feature preview to setting

SaxonF:
image

Required: https://github.com/supabase/supabase/pull/39834

With the introduction of a shared sidebar space, we are at a point where it makes sense to replace the inline editor feature preview with a user based setting, starting with local storage. The intention of this is to offer a more native/code first experience for users who prefer to work that way (vs UI). LLM's are also becomingly increasingly capable which makes our editor that has inline assistant built a more approachable option.


apps/studio/hooks/misc/useIsInlineEditorEnabled.ts
joshenlim, line 1: i reckon this can live within InlineEditorSettings? keep these 2 related things beside each other

apps/studio/components/interfaces/Account/Preferences/InlineEditorSettings.tsx
```diff
@@ -0,0 +1,86 @@
...
+  inlineEditorEnabled: z.boolean(),
+})
+
+export const useIsInlineEditorEnabled = () => {
+  const [inlineEditorEnabled] = useLocalStorageQuery(
+    LOCAL_STORAGE_KEYS.UI_PREVIEW_INLINE_EDITOR,
+    true
+  )
+
+  return inlineEditorEnabled ?? true
```
joshenlim: this should be false? otherwise users who hasn't toggled this feature preview before will get defaulted to the inline editor

apps/studio/components/interfaces/Account/Preferences/InlineEditorSettings.tsx
```diff
@@ -0,0 +1,86 @@
...
+    true
+  )
+  const { data: org } = useSelectedOrganizationQuery()
+
+  const { mutate: sendEvent } = useSendEventMutation()
+
+  const form = useForm<z.infer<typeof InlineEditorSchema>>({
+    resolver: zodResolver(InlineEditorSchema),
+    values: {
+      inlineEditorEnabled: inlineEditorEnabled ?? true,
```
joshenlim: likewise should be false here - should be a power user opt in imo

joshenlim (approved):
Verified on staging previews setting inline editor, and verified that it works on policies, functions and trigger pages. Also double checked AI assistant within inline editor

packages/common/telemetry-constants.ts
```diff
@@ -2181,6 +2181,24 @@ export interface TableQuickstartAssistantOpenedEvent {
...
 
+/**
+ * User toggled the inline editor setting in account preferences.
+ *
+ * @group Events
+ * @source studio
+ * @page /dashboard/account/preferences
+ */
+export interface InlineEditorSettingToggledEvent {
+  action: 'inline_editor_setting_toggled'
```
pamelachia: Could we use clicked as the verb of choice instead of toggled? We aim to re-use verbs if they're similar in functionality.

pamelachia (approved):

packages/common/telemetry-constants.ts
pamelachia: approving to unblock but just change this before merging

PR #39893, opened 2025-10-27
chore(studio): put focus on Discord in help menu

dnywh:

Clarifies Discord as our community resource in Studio. Resolves DEPR-150.

`HelpPopover` component currently:

- Displays both Discord and GitHub Discussions with equal prominence and value proposition
- Has a bunch of language that is arguably redundant

## What is the new behavior?

`HelpPopover` component now:

- Only displays Discord as the community resource of choice
- More clearly pushes community support in general
- Removes redundant language
  - Prioritisation based on plan (only relevant to support form, and repeated there)
  - Support only available for platform use (just hide those options when `IS_PLATFORM` is false)

### Hosted

When `IS_PLATFORM` is true.

User is inside a project:

| Before | After |
| --- | --- |
| Prime Walkers  The Long Walk  Supabase | Prime Walkers  The Long Walk  Supabase |

User is at org level:

| Before | After |
| --- | --- |
| Supabase | Supabase |

### Not hosted

When `IS_PLATFORM` is false.

User is inside a project:

| Before | After |
| --- | --- |
| Prime Walkers  The Long Walk  Supabase | Prime Walkers  The Long Walk  Supabase |

User is at org level:

| Before | After |
| --- | --- |
| Supabase | Supabase |

## Additional context

Some purposeful design decisions:

- **Lightweight:** this is something meaningful we can get out the door now. I do see larger issues in regards to the the overlap with _Feedback_ dropdown, but we think about any larger redesigns more holistically. One step at a time.
- **Ordering**: we discussed putting the Discord half on top of `HelpPopover`. This doesn’t feel right in practice, as arguably the Assistant and docs (current top-level items) are better resources. The text flow and prominence now should achieve a similar result; putting _emphasis_ on Discord.

Please feel free to push back on these things in particular:

1. Should we be presenting this to Team and Enterprise users as-is? Is it a bit on-the-nose to be saying “Start with our docs or community” to them? What else would we say in its place??
2. Any strong feelings about the “Our Discord community” copy? It has room for improvement, in my opinion.


SaxonF (approved):

saltcod (approved):
Looks good on staging!

joshenlim (approved):
Triple stamp! ✅  Verified on staging preview 🙂

PR #39894, opened 2025-10-27
feat: allow skipping retries on specific pathnames

alaister:
**Changes:**
- Tracks request pathname on `ResponseError`.
- Introduces a set of pathnames that retries should be skipped on (requested by API team).

**To test (locally):**

- [ ] Add a random `throw new Error('whatever')` into one of the listed API routes (I did `/run-lints`) on your local infra setup, then make sure the FE doesn't retry when requesting that route in the network console.
- [ ] Also adding another random throw to a good API route to test that retries are continuing to work.

### DIFF
--- a/apps/studio/data/query-client.ts
+++ b/apps/studio/data/query-client.ts
@@ -1,6 +1,8 @@
 import { QueryClient, onlineManager } from '@tanstack/react-query'
-import { IS_PLATFORM } from 'lib/constants'
+import { match } from 'path-to-regexp'
 import { useState } from 'react'
+
+import { IS_PLATFORM } from 'lib/constants'
 import { ResponseError } from 'types'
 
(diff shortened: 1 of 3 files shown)


apps/studio/data/query-client.ts
```diff
@@ -9,6 +11,14 @@ if (!IS_PLATFORM) {
   onlineManager.setOnline(true)
 }
 
+const SKIP_RETRY_PATHNAME_MATCHERS = [
+  '/system/projects/:ref/run-lints',
```
phamhieu: dashboard does not use this system route

phamhieu (approved):

kamilogorek (approved):

charislam (approved):
Tested as described

PR #39899, opened 2025-10-27
feat(auth): add account changes notifications template schemas

fadymak:

- Feature: new email templates
- Resolves DEPR-138
- Resolves AUTH-824

- Only the existing authentication email templates are available

## What is the new behavior?

- Account changes notifications email templates (`security`) are also available (to people within the feature preview)
- Email template schemas have been expanded to note the `emailTemplateType`: `authentication` or `security`

| Before | After |
| --- | --- |

## To test

- [ ] For users not under the feature preview, do the existing `authentication` templates render and work as expected?
- [ ] When the feature preview is on, can you see the new `security` templates?
  - [ ] Can you enable and disable them from the root page?
- [ ] [We deleted](https://github.com/supabase/supabase/pull/39899#discussion_r2472227920) the `misc` properties because they seemed to be unused. Are they truely safe to delete?

### DIFF
--- a/apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
+++ b/apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
@@ -32,6 +32,9 @@ const CONFIRMATION: FormSchema = {
   validationSchema: object().shape({
     MAILER_SUBJECTS_CONFIRMATION: string().required('Subject heading is required.'),
   }),
+  misc: {
+    emailTemplateType: 'authentication',
+  },
 }
 
 const INVITE: FormSchema = {
@@ -63,6 +66,9 @@ const INVITE: FormSchema = {
   validationSchema: object().shape({
     MAILER_SUBJECTS_INVITE: string().required('Subject heading is required.'),
   }),
+  misc: {
+    emailTemplateType: 'authentication',
+  },
 }
 
 const MAGIC_LINK: FormSchema = {
@@ -94,6 +100,9 @@ const MAGIC_LINK: FormSchema = {
   validationSchema: object().shape({
     MAILER_SUBJECTS_MAGIC_LINK: string().required('Subject heading is required.'),
   }),
+  misc: {
+    emailTemplateType: 'authentication',
+  },
 }
 
 const EMAIL_CHANGE: FormSchema = {
@@ -126,6 +135,9 @@ const EMAIL_CHANGE: FormSchema = {
   validationSchema: object().shape({
     MAILER_SUBJECTS_EMAIL_CHANGE: string().required('Subject heading is required.'),
   }),
+  misc: {
+    emailTemplateType: 'authentication',
+  },
 }
 
 const RECOVERY: FormSchema = {
@@ -157,6 +169,9 @@ const RECOVERY: FormSchema = {
   validationSchema: object().shape({
     MAILER_SUBJECTS_RECOVERY: string().required('Subject heading is required.'),
   }),
+  misc: {
+    emailTemplateType: 'authentication',
+  },
 }
 const REAUTHENTICATION: FormSchema = {
   $schema: JSON_SCHEMA_VERSION,
@@ -185,6 +200,9 @@ const REAUTHENTICATION: FormSchema = {
   validationSchema: object().shape({
     MAILER_SUBJECTS_REAUTHENTICATION: string().required('Subject heading is required.'),
   }),
+  misc: {
+    emailTemplateType: 'authentication',
+  },
 }
 
 // Notifications
@@ -214,6 +232,9 @@ const PASSWORD_CHANGED_NOTIFICATION: FormSchema = {
       'Subject heading is required.'
     ),
   }),
+  misc: {
+    emailTemplateType: 'security',
+  },
 }
 
 const EMAIL_CHANGED_NOTIFICATION: FormSchema = {
@@ -241,6 +262,9 @@ const EMAIL_CHANGED_NOTIFICATION: FormSchema = {
   validationSchema: object().shape({
     MAILER_SUBJECTS_EMAIL_CHANGED_NOTIFICATION: string().required('Subject heading is required.'),
   }),
+  misc: {
+    emailTemplateType: 'security',
+  },
 }
 
 const PHONE_CHANGED_NOTIFICATION: FormSchema = {
@@ -269,6 +293,9 @@ const PHONE_CHANGED_NOTIFICATION: FormSchema = {
   validationSchema: object().shape({
     MAILER_SUBJECTS_PHONE_CHANGED_NOTIFICATION: string().required('Subject heading is required.'),
   }),
+  misc: {
+    emailTemplateType: 'security',
+  },
 }
 
 const IDENTITY_LINKED_NOTIFICATION: FormSchema = {
@@ -296,6 +323,9 @@ const IDENTITY_LINKED_NOTIFICATION: FormSchema = {
   validationSchema: object().shape({
     MAILER_SUBJECTS_IDENTITY_LINKED_NOTIFICATION: string().required('Subject heading is required.'),
   }),
+  misc: {
+    emailTemplateType: 'security',
+  },
 }
 
 const IDENTITY_UNLINKED_NOTIFICATION: FormSchema = {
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
```diff
@@ -141,13 +124,8 @@ const EMAIL_CHANGE: FormSchema = {
...
   },
   validationSchema: object().shape({
-    MAILER_SUBJECTS_EMAIL_CHANGE: string().required('"Subject heading is required.'),
+    MAILER_SUBJECTS_EMAIL_CHANGE: string().required('Subject heading is required.'),
   }),
-  misc: {
-    iconKey: 'email-icon2',
-    helper: `To complete setup, add this authorisation callback URL to your app's configuration in the Apple Developer Console.
-[Learn more](${DOCS_URL}/guides/auth/social-login/auth-apple#configure-your-services-id)`,
-  },
```
dnywh, line 150: @issuedat just checking that removing these `misc` properties of each item was purposeful?

Just double-checking as I have no idea what this is or what the implications are of removing it!

apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
```diff
@@ -226,4 +392,12 @@ export const TEMPLATES_SCHEMAS = [
...
   RECOVERY,
   REAUTHENTICATION,
+  // Notifications
+  PASSWORD_CHANGED_NOTIFICATION,
+  EMAIL_CHANGED_NOTIFICATION,
+  PHONE_CHANGED_NOTIFICATION,
+  IDENTITY_LINKED_NOTIFICATION,
+  IDENTITY_UNLINKED_NOTIFICATION,
+  MFA_FACTOR_ENROLLED_NOTIFICATION,
+  MFA_FACTOR_UNENROLLED_NOTIFICATION,
```
dnywh, line 441: Thoughts on removing the `_NOTIFICATION` part of each of these @issuedat?

dnywh: I had another question for you which might affect your answer:

I want to be able to reference _only_ the notification type emails, and _only_ the (existing authentication type emails.

E.g. to be able to split them into their own tabs and/or sections.

So I think I need to:

- A) add a property to each item in here that says whether it is `"authentication"` or `"notification"` ("or is "security" clearer?)
- B) just add that specification in either this `export const TEMPLATES_SCHEMAS` or an additional one, like `export const TEXMPLATES_SCHEMAS_NOTIFICATIONS`

What do you think is best?

apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
fadymak: I didn't see it being used anywhere on the pages and the text seemed irrelevant to the email templates. A second pair of eyes to confirm would be good though!

apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
fadymak: > Thoughts on removing the _NOTIFICATION part of each of these @issuedat?

That should be fine! I mostly added them for consistency with how we differentiate between authentication templates vs notification templates. However, the `id` field will need to have the `_NOTIFICATION` suffix since it's used as the slug in the APIs as the template identifier.

> I want to be able to reference only the notification type emails, and only the (existing authentication type emails.
> 
> E.g. to be able to split them into their own tabs and/or sections.

I think a separate export (`export const TEXMPLATES_SCHEMAS_NOTIFICATIONS`) might be the least friction if we plan to separate them into different sections.

apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
dnywh: Cool. Good idea. I’ll check too, and then note this in the PR body for team/dashboard to doublecheck when reviewing.

apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
dnywh: Cool. I like your change with `misc.emailTemplateType` for now, and will ask the reviewer for their input too.

Zooming out: would should reconcile our language? We use “notifications” and “security” to mean the same thing, as far as I can tell. Are there any implications for the API?

apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
dnywh: For the reviewer: it seems this `misc.helper` is used for the emails sent by other providers, namely [Sign In with Apple](https://supabase.com/docs/guides/auth/social-login/auth-apple#configure-your-services-id). Is that right?

[Code reference here](https://github.com/supabase/supabase/blob/iat/notifications-schemas/apps/studio/components/interfaces/Auth/AuthProvidersForm/ProviderForm.tsx#L245)

apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
dnywh: For the reviewer:

Instead of a misc.emailTemplateType, should we instead/additionally do something like the below? That would simplify our `map` in the `EmailTemplates` component.

We _do_ need to know the `emailTemplateType` in the _TemplateEditor_ though, as we will show things dynamically there based on the type of template.

```tsx
export const TEMPLATES_SCHEMAS_AUTHENTICATION = [
  CONFIRMATION,
  INVITE,
  MAGIC_LINK,
  EMAIL_CHANGE,
  RECOVERY,
  REAUTHENTICATION,
]

export const TEMPLATES_SCHEMAS_SECURITY = [
  PASSWORD_CHANGED_NOTIFICATION,
  EMAIL_CHANGED_NOTIFICATION,
  PHONE_CHANGED_NOTIFICATION,
  IDENTITY_LINKED_NOTIFICATION,
  IDENTITY_UNLINKED_NOTIFICATION,
  MFA_FACTOR_ENROLLED_NOTIFICATION,
  MFA_FACTOR_UNENROLLED_NOTIFICATION,
]

export const TEMPLATES_SCHEMAS = [
  ...TEMPLATES_SCHEMAS_AUTHENTICATION,
  ...TEMPLATES_SCHEMAS_SECURITY,
]
```

apps/studio/components/interfaces/Auth/EmailTemplates/EmailTemplates.tsx
```diff
@@ -52,51 +123,145 @@ export const EmailTemplates = () => {
...
+                                href={`/project/${projectRef}/auth/templates/${templateSlug}`}
+                                className="py-6 pr-6"
+                              >
+                                <ChevronRight
+                                  size={16}
+                                  className="text-foreground-muted hover:text-foreground transition-colors"
+                                />
+                              </Link>
+                            </div>
+                          </CardContent>
```
dnywh, line 219: This `CardContent` is structured a little differently to the one [above](https://github.com/supabase/supabase/pull/39899/files#diff-52c294357eea7e2aae352c20bec994baa039d188cc7b8beb6d8ee8b293b9bbc4R141) for the `authentication` emails mapping. That’s because we need to support:

1. The row being tappable, taking the user to the template page
2. The switch also being tappable, changing the state of the template

In my tinkering, `e.stopPropagation()` on the switch wasn’t enough to stop the parent row taking the user to the template page. This approach basically makes three segments instead. The side-benefit is that the switch has a bit of breathing room; if the user misclicks by a few pixels, they aren’t hijacked to another page.

apps/studio/components/interfaces/Auth/EmailTemplates/EmailTemplates.tsx
```diff
@@ -29,12 +58,54 @@ export const EmailTemplates = () => {
...
+    PermissionAction.UPDATE,
+    'custom_config_gotrue'
+  )
 
   const builtInSMTP =
     isSuccess &&
     authConfig &&
     (!authConfig.SMTP_HOST || !authConfig.SMTP_USER || !authConfig.SMTP_PASS)
 
+  const getNotificationEnabledKeys = () =>
```
joshenlim: reckon this can just be a const?

joshenlim (approved):
Verified on staging preview both with flag off and on - LGTM! Tested updating templates, and toggling notifications

PR #39902, opened 2025-10-27
Clean up iceberg wrapper when deleting analytics bucket

joshenlim:
Branches out of this PR: https://github.com/supabase/supabase/pull/39610

## Context

Cleans up the iceberg FDW when deleting analytics bucket, otherwise will be unable to create another analytics bucket with the same name

Also cleans up S3 access key as well

## To test
Changes can only be tested locally against this PR branch: `joshen/add-skeleton-of-analytics-buckets-to-storage` ([PR](https://github.com/supabase/infrastructure/pull/26697))
- [ ] Ensure that new storage UI feature preview is toggled
- [ ] Create an analytics bucket
- [ ] Delete the analytics bucket
- [ ] Verify that S3 key associated with analytics bucket is also removed under Configuration -> S3
- [ ] Try to create another analytics bucket with the same name

### DIFF
--- a/apps/studio/components/interfaces/Storage/DeleteBucketModal.tsx
+++ b/apps/studio/components/interfaces/Storage/DeleteBucketModal.tsx
@@ -1,5 +1,5 @@
 import { zodResolver } from '@hookform/resolvers/zod'
-import { get as _get, find } from 'lodash'
+import { get as _get, find, snakeCase } from 'lodash'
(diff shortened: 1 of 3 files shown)


charislam (commented):
In testing, I can create, delete, and recreate the bucket, but it doesn't delete the access key under S3 Configuration

PR #27436, opened 2024-06-21
blog: postgis vector tiles

thorwebdev:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES/NO

## What kind of change does this PR introduce?

Bug fix, feature, docs update, ...

## What is the current behavior?

Please link any relevant issues here.

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context

Add any other context or screenshots.

PR #39904, opened 2025-10-27
docs: Add schema qualified object references throughout the docs

dshukertjr:

YES

Update the docs to use schema-qualified object references. For
1. Mitigating [search path poisoning attack](https://www.postgresql.org/support/security/CVE-2020-14349/)
2. Making the Supabase CLI able to push migrations that are copied from the docs.


apps/docs/content/guides/database/extensions/postgis.mdx
```diff
@@ -35,11 +35,8 @@ You can get started with PostGIS by enabling the PostGIS extension in your Supab
 <TabPanel id="sql" label="SQL">
 
 ```sql
--- Create a dedicated separate schema
-create schema if not exists "gis";
-
 -- Example: enable the "postgis" extension
-create extension postgis with schema "gis";
```
dshukertjr, line 42: By default, from the dashboard Postgis gets installed on `extensions` schema, so I'm updating this doc to use `extensions` schema as well. 

It seems like the original reason why this `gis` schema was introduced was becauase of [this line](https://github.com/supabase/supabase/pull/27436/files#diff-5217b4ca6938ea74ac61a9a410667e0be3f61cd6772aee8d8889ab6c40dc47b2R74) from a blog post, which I feel like the schema of the extension shouldn't have much effect on the performance of the database, but let me know if I'm wrong and we want to keep the `gis` schema.

sweatybridge: I agree there's no performance benefits to creating postgis extension in its own schema. But if we want to recommend the `extensions` schema, all function references to `gis` schema will also need to be updated. For eg.

https://github.com/supabase/supabase/pull/39904#diff-08d7c28bda4c1d1495f9073f44664ef9e7987485e17abdee2f7c7d2fb9a65ce5R216

```sql
select id, name, extensions.st_y(location::extensions.geometry) as lat, extensions.st_x(location::extensions.geometry) as long, extensions.st_distance(location, extensions.st_point(long, lat)::extensions.geography) as dist_meters
```

sweatybridge (commented):
vector changes look good.

apps/docs/content/guides/database/extensions/postgis.mdx
sweatybridge: So I think it's simpler to keep this blogpost unchanged because it's fine to use a separate `gis` schema to keep the extensions namespace clean.

dshukertjr: Awesome! I didn't touch the blog post, but fixed the guide.

dshukertjr:
Done updating the `gis` references that I missed 👍

sweatybridge (approved):

PR #39905, opened 2025-10-27
chore: invalidate entitlements

ignaciodob:
This PR adds invalidation to the entitlements query when the Organization's plan is updated and changes the staleTime to 30 minutes.


apps/studio/data/entitlements/entitlements-query.ts
kevcodez: Should use `organizationKeys.entitlements` here now

kevcodez (approved):

PR #39906, opened 2025-10-27
Hook up connecting tables for analytics buckets

joshenlim:
Branches out of https://github.com/supabase/supabase/pull/39610

## Pre-requisite

Heads up that this cannot be tested locally (unless you have ETL set up locally), nor on staging unless your staging project is flagged to be able to use replication - which is quite a bit of a blocker to have this tested. I haven't been able to test the changes here E2E as a result.

As such, am okay for just a code review for now as we'd still need to do an E2E test either way before merging the main branch

## Context

Hooks up connecting tables for Analytics Bucket, which is essentially trying to do the same as what's `DestinationPanel.tsx` doing. The changes here _only_ cover the flow of connecting tables from an empty state (e.g new Analytics Bucket) so there's probably more cases to consider thereafter too (can follow up with subsequent PRs)

In general from the user's POV they're only just selecting which tables they want to send data from
image

But behind the scenes thereafter, we're
- Creating a new publication based on the selected tables (Default name to `{bucketId}_publication`)
- Creating a new namespace (Default name to `{bucketId}_namespace`)
- Creating a new destination pipeline (Default name to `{bucketId}_destination`)
- Starting the newly created pipeline

However, what's missing thereafter still are
- Managing namespaces
  - I suppose we also need to support deleting namespaces? Where can I check what's the endpoint for that?
- Removing / Adding more tables
  - Question: Is this just a matter of updating the publication?
- Error handling due to multiple async processes

### DIFF
--- a/apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
@@ -1,16 +1,14 @@
-import { snakeCase, uniq } from 'lodash'
+import { uniq } from 'lodash'
 import { SquarePlus } from 'lucide-react'
 import Link from 'next/link'
 import { useMemo, useState } from 'react'
 
 import { useIsNewStorageUIEnabled } from 'components/interfaces/App/FeaturePreview/FeaturePreviewContext'
 import { INTEGRATIONS } from 'components/interfaces/Integrations/Landing/Integrations.constants'
-import { WRAPPER_HANDLERS } from 'components/interfaces/Integrations/Wrappers/Wrappers.constants'
 import { WrapperMeta } from 'components/interfaces/Integrations/Wrappers/Wrappers.types'
 import {
   convertKVStringArrayToJson,
   formatWrapperTables,
-  wrapperMetaComparator,
 } from 'components/interfaces/Integrations/Wrappers/Wrappers.utils'
 import { BUCKET_TYPES } from 'components/interfaces/Storage/Storage.constants'
 import { PageLayout } from 'components/layouts/PageLayout/PageLayout'
@@ -27,7 +25,6 @@ import {
   DatabaseExtension,
   useDatabaseExtensionsQuery,
 } from 'data/database-extensions/database-extensions-query'
-import { useFDWsQuery } from 'data/fdw/fdws-query'
 import { AnalyticsBucket } from 'data/storage/analytics-buckets-query'
 import { useIcebergNamespacesQuery } from 'data/storage/iceberg-namespaces-query'
 import { useIcebergWrapperCreateMutation } from 'data/storage/iceberg-wrapper-create-mutation'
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
```diff
@@ -55,56 +52,35 @@ import { DecryptedReadOnlyInput } from './DecryptedReadOnlyInput'
 import { NamespaceRow } from './NamespaceRow'
 import { NamespaceWithTables } from './NamespaceWithTables'
 import { SimpleConfigurationDetails } from './SimpleConfigurationDetails'
+import { useAnalyticsBucketWrapperInstance } from './useAnalyticsBucketWrapperInstance'
 import { useIcebergWrapperExtension } from './useIcebergWrapper'
 
 export const AnalyticBucketDetails = ({ bucket }: { bucket: AnalyticsBucket }) => {
+  const config = BUCKET_TYPES['analytics']
```
awaseem: nit: why not the dot operator?


```suggestion
  const config = BUCKET_TYPES.analytics
```

apps/studio/components/interfaces/Storage/ConnectTablesDialog.tsx
```diff
@@ -25,66 +34,151 @@ import {
...
+      })
+
+      const decryptedValues = await getDecryptedValues({
+        projectRef,
+        connectionString: project?.connectionString,
+        ids: Object.entries(wrapperValues)
+          .filter(([k, v]) => {
+            if (['vault_aws_access_key_id', 'vault_aws_secret_access_key'].includes(k)) return v
+          })
+          .map((x) => x[1]),
```
awaseem: I have a heard time reading this, can we:
- Assigns this to its own variable and reference it here. 
- Maybe simplify it to the following? maybe thats worse haha, let me know your thoughts

```suggestion
        ids: Object.entries(wrapperValues)
          .filter(
            ([key, value]) =>
              ['vault_aws_access_key_id', 'vault_aws_secret_access_key'].includes(key) && value
          )
          .map(([, value]) => value),
```

awaseem (approved):
Minor nits, everything else looks great!

apps/studio/components/interfaces/Storage/AnalyticBucketDetails/index.tsx
joshenlim: oh yup good point hahah

apps/studio/components/interfaces/Storage/ConnectTablesDialog.tsx
joshenlim: good point though RE readability - i'll update it as such, hopefully a bit more clearer

```
const keysToDecrypt = Object.entries(wrapperValues)
    .filter(([name]) =>
      ['vault_aws_access_key_id', 'vault_aws_secret_access_key'].includes(name)
    )
    .map(([_, keyId]) => keyId)
const decryptedValues = await getDecryptedValues({
    projectRef,
    ids: keysToDecrypt,
})
```

PR #39907, opened 2025-10-27
fix(settings): add Zod validation to limit project name length to 64 chars

kartik-212004:

YES

Fix the issue with updating the project name in settings. Currently, there’s no proper Zod validation for the name length. Names longer than 64 characters are accepted, causing the UI to break. Added validation to enforce a maximum length of 64 characters.

https://github.com/user-attachments/assets/239cd8cf-ce5a-401d-9507-835b721f4dfd

## What is the new behavior?

couldn't add the video as there is no option to modify the project name during development

### DIFF
--- a/apps/studio/components/interfaces/Settings/General/General.tsx
+++ b/apps/studio/components/interfaces/Settings/General/General.tsx
@@ -50,7 +50,9 @@ const General = () => {
     ref: z.string().optional(),
   })
 
-  const validateForm = (values: any) => {
+  type SettingsFormValues = z.infer<typeof SettingsFormSchema>
+
+  const validateForm = (values: SettingsFormValues) => {
     const result = SettingsFormSchema.safeParse(values)
     if (result.success) return {}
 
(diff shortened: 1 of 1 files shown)


kartik-212004:
@charislam  would appreciate your feedback when you get a chance.

charislam:
Thanks for the PR @kartik-212004!

My view on this is that if it's allowed by the API, it should be allowed on the frontend. Otherwise we might "fix" it, but it'd still be broken for anyone who created their project on the API. What do you think about changing the project name display so it's truncated instead?

kartik-212004:
> Thanks for the PR @kartik-212004!
> 
> My view on this is that if it's allowed by the API, it should be allowed on the frontend. Otherwise we might "fix" it, but it'd still be broken for anyone who created their project on the API. What do you think about changing the project name display so it's truncated instead?

yeah, so the api accepts the name string without any limit.
but when creating a project, there’s a 64-char limit, while updating doesn’t have any restriction.

also, there’s no check for duplicate project names, so multiple projects can have the same name right now when modifying the project name.

maybe I should revert the Settings validation, add a small TruncatedText component if you want, and just truncate the display on the front end.

Hallidayo:

PR #39916, opened 2025-10-28
fix(studio): troubleshooting links

dnywh:

Bug fix

Links to the Troubleshooting page were using an old URL, which in turn was redirected to a very specific “[Diagnose HTTP API issues](https://supabase.com/docs/guides/troubleshooting/http-api-issues)” page.

## What is the new behavior?

We link to the root Troubleshooting page with the [`?products=platform`](https://supabase.com/docs/guides/troubleshooting?products=platform) search param.


apps/studio/components/interfaces/Support/SubjectAndSuggestionsInfo.tsx
dnywh, line 1: Just removing unnecessary `<span>` elements

joshenlim (approved):
Verified URL changes!

PR #39919, opened 2025-10-28
fix: usage page project filtering via URL param

kevcodez:
When opening the usage page with a projectRef param (filter by project), it lead to some odd behaviour where the project is not preselected and then after picking a project, it would end up showing the initially filtered project from the URL.

PR removes the odd all-projects logic and uses nuqs for URL state mgmt

### DIFF
--- a/apps/studio/components/interfaces/Organization/Usage/Usage.tsx
+++ b/apps/studio/components/interfaces/Organization/Usage/Usage.tsx
@@ -1,7 +1,7 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import dayjs from 'dayjs'
 import Link from 'next/link'
-import { useEffect, useMemo, useState } from 'react'
+import { useMemo, useState } from 'react'
 
 import { useParams } from 'common'
 import {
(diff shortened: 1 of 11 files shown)


apps/studio/components/interfaces/Organization/Usage/Usage.tsx
```diff
@@ -67,6 +52,13 @@ export const Usage = () => {
     isSuccess: isSuccessSubscription,
   } = useOrgSubscriptionQuery({ orgSlug: slug })
 
+  const { data: selectedProject } = useProjectDetailQuery(
+    {
+      ref: selectedProjectRef!,
+    },
+    { enabled: selectedProjectRef != null }
+  )
```
alaister: nit:
```suggestion
  const { data: selectedProject } = useProjectDetailQuery(
    {
      ref: selectedProjectRef ?? undefined,
    }
  )
```
(the query has a built in enabled check)

alaister (approved):
LGTM!

- Tested on preview

One small nit:

PR #39922, opened 2025-10-28
Simplify paused state

SaxonF:
**New:**
image

**Old:**
image

Simplifies paused state in an effort to make it a little less scary via less alerts  and condensed copy.


joshenlim:
Adjusting some of the states (e.g loading, error, and pause disabled)

Just to show all states here:

Loading (for pause status):
image

Error state:
image

Paused + Able to resume:
image

image


Resume confirmation:
image

Paused + Past 90 days threshold:
image

joshenlim (approved):
Verified changes on staging preview but reckon good to do one more double checking on copy before merging just to make sure we didn't leave any info out from the old UI

saltcod (commented):
looks great!

PR #39923, opened 2025-10-28
studio(fix):don’t presume organization on API authorize step

dnywh:

Fixes DEPR-131

Our API authorisation step (used for example when connecting to an MCP client like Cursor) currently presumes the organisation based on whatever is first alphabetically. This is the fallback behaviour when no `organization_slug` is provided.

## What is the new behavior?

When no `organization_slug` is provided, we make no assumptions. We make the user actively select the organization they wish to grant access to. Additionally:

- A placeholder “Select an organization” string is supplied
- We don’t disable the buttons. Instead, if the user tries to proceed without selecting an authorization, we tell them what to fix

## Additional context

| Before | After |
| --- | --- |
| Authorize API access  Supabase | Authorize API access  Supabase |
| Presumes first organisation alphabetically | Forces me to choose an organisation |

## To test

You can use the Connect dialog (`?showConnect=true&connectTab=mcp`) to one-click add Supabase to Cursor. From there, tap _Install_ and then _Connect_. That should bring you to the authorization step in question. Please test:

- [x] Trying to proceed without supplying an organis£ation
- [x] Selecting an organisation and then proceeding

It would also be great if you could test another situation where `organization_slug` is provided:
- [x] Does it render correctly?
- [x] Does it function as per usual?

### DIFF
--- a/apps/studio/pages/authorize.tsx
+++ b/apps/studio/pages/authorize.tsx
@@ -108,24 +108,17 @@ const APIAuthorizationPage: NextPageWithLayout = () => {
     )
   })
 
-  const onDeclineRequest = async () => {
+  const onDeclineRequest = form.handleSubmit((values) => {
     if (!auth_id) {
       return toast.error('Unable to decline request: auth_id is missing ')
     }
 
-    // Guard: ensure organizations are loaded
-    if (!organizations || organizations.length === 0) {
-      return toast.error('Unable to decline request: Organizations not loaded')
-    }
-
-    const slugToUse = selectedOrgSlug || organization_slug || organizations[0]?.slug
-    if (!slugToUse) {
-      return toast.error('Unable to decline request: No organization context available')
-    }
-
     setIsDeclining(true)
-    declineRequest({ id: auth_id, slug: slugToUse }, { onError: () => setIsDeclining(false) })
-  }
+    declineRequest(
+      { id: auth_id, slug: values.selectedOrgSlug },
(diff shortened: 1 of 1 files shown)


charislam (approved):

apps/studio/pages/authorize.tsx
```diff
@@ -108,17 +108,24 @@ const APIAuthorizationPage: NextPageWithLayout = () => {
...
     if (!auth_id) {
       return toast.error('Unable to decline request: auth_id is missing ')
     }
 
+    // Guard: ensure organizations are loaded
+    if (!organizations || organizations.length === 0) {
+      return toast.error('Unable to decline request: Organizations not loaded')
+    }
+
+    const slugToUse = selectedOrgSlug || organization_slug || organizations[0]?.slug
```
charislam: I'm thinking it would be better to stick with `values.selectedOrgSlug`. With all the fallbacks here, it might be easy to get into a bizarre state where the user has no idea what they're declining

dnywh: Good shout. I’ve just simplified/reverted the change so it does what you suggested.

Note: for similar reasons, I left the _Decline_ button enabled in the case that no organisation is selected. It shows the form validation error in the case a user clicks it without having selected an organisation:

> Please select an organization

This is instead of disabling the button until a organisation is selected. That would be technically neater but is cryptic. I.e. the user could reasonably have no idea why they can’t decline.

PR #39924, opened 2025-10-28
chore: remove sidebar labels

dnywh:

- UI improvement
- Design documentation
- Fixes DEPR-127

We have a bunch of pills denoting “Alpha” or ”New” or ”Coming Soon” on the sidebar. Many of these are outdated or lack follow-up information.

## What is the new behavior?

We get rid of outdated pills from the sidebar.

## Additional context

Our sidebars are already quite hectic and these little pills don’t provide much value on their own. Ideally we don’t use them unless:

- Users _must_ know that the feature may behave unexpectedly
- And in that case, we should have an Admonition inline with a CTA towards a place to leave feedback or get help


joshenlim (approved):

craigcannon (approved):

PR #39925, opened 2025-10-28
fix: wait for auth to load before fetching feature flags

alaister:
Currently on initial render, we're sending two requests to `/platform/telemetry/feature-flags` because we're not waiting for the user to load:
Screenshot 2025-10-28 at 16 18 20

The first request is unauthenticated, followed by another request with the user email attached right after it.

This PR simply waits for the authentication state to finish loading before sending the request.

**To test:**

- [ ] Verify only one request is being sent
- [ ] Ensure feature flags still function as normal


ivasilov (approved):

joshenlim (approved):
Double checked on preview that there's only 1 x request for feature-flags and configv6
and data from those look correct too

PR #39927, opened 2025-10-28
chore: move password check to frontend

alaister:
**Changes:**
- Switches the password check to import the library directly instead of calling an API route
- Removes debouncing on the password checks (no longer needed since we're not making requests)
- Removes the self-hosted API route
- Fixes one tiny issue with no space being shown before the Generate a Password text

**To test:**

- [x] Ensure that the password strength bar still works in all places
- [x] Verify the chunk for the password check starts loading when you start typing into the input


apps/studio/lib/password-strength.ts
```diff
@@ -13,29 +15,20 @@ export default async function passwordStrength(value: string) {
...
-
-        // warning message for anything below 4 strength :string
-        if (resultScore < DEFAULT_MINIMUM_PASSWORD_STRENGTH) {
-          warning = `${
-            result?.feedback?.warning ? result?.feedback?.warning + '.' : ''
-          } You need a stronger password.`
-        }
-      } else {
-        toast.error(`Failed to check password strength: ${(error as ResponseError).message}`)
+      const result = zxcvbn(value)
```
alaister, line 16: Here's the main change

apps/studio/lib/password-strength.ts
```diff
@@ -13,29 +13,20 @@ export default async function passwordStrength(value: string) {
...
-        const resultScore = result?.score ?? 0
+      const result = zxcvbn(value)
+      const resultScore = result?.score ?? 0
 
-        const score = (PASSWORD_STRENGTH as any)[resultScore]
-        const suggestions = result.feedback?.suggestions
-          ? result.feedback.suggestions.join(' ')
-          : ''
+      const score = (PASSWORD_STRENGTH as any)[resultScore]
+      const suggestions = result.feedback?.suggestions ? result.feedback.suggestions.join(' ') : ''
```
charislam:
```suggestion
      const suggestions = result.feedback?.suggestions?.join(' ') ?? ''
```

tiny nit

charislam (approved):
Tested as described (except for the Vercel integration one, just did a code check for that)

charislam:
Looks like wording on the password strength tests needs to be updated though

PR #39942, opened 2025-10-28
chore: update self-hosted image versions

kiwicopple:


docker/docker-compose.yml
```diff
@@ -343,7 +343,7 @@ services:
 
   analytics:
     container_name: supabase-analytics
-    image: supabase/logflare:1.22.6
+    image: supabase/logflare:1.23.2
```
aantti: Waiting for feedback on whether the previous issues with the logs were fixed.

aantti (approved):

docker/docker-compose.yml
aantti: Keeping logflare at 1.22.6 for now.

PR #39986, opened 2025-10-29
Fix table_quickstart_opened event to track control group

seanoliver:
## Problem

The `table_quickstart_opened` event wasn't tracking users in the control group, preventing baseline measurement for the A/B test.

**Root cause:** `activeQuickstartVariant` filters out `CONTROL` variants, so the tracking logic never fired for control group users.

## Solution

- Add `shouldTrackExposure` boolean to separate tracking from rendering logic
- Track `tableQuickstartVariant` directly (includes control) instead of `activeQuickstartVariant`

## Test Plan

- [x] Verify control group users send `table_quickstart_opened` event with `variant: 'control'`
- [x] Verify treatment users still send events with correct variant names
- [x] Verify no events sent for projects > 7 days old
- [x] Verify no events sent in SQL editor mode
- [x] Verify event only fires once per session (ref check)

### DIFF
--- a/apps/studio/components/layouts/Tabs/NewTab.tsx
+++ b/apps/studio/components/layouts/Tabs/NewTab.tsx
@@ -81,7 +81,7 @@ export function NewTab() {
 
   /**
    * Returns:
-   * - `QuickstartVariant`: user variation (`ai`, `templates`, `assistant`)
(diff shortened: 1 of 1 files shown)


awaseem (approved):
Tested locally to make sure there wasn't any strange re-renders. LGTM

PR #39991, opened 2025-10-29
Updates to Edge Functions dashboard code editor

laktek:

Main highlights:
* Removed the client-side eszip parsing in favor of mgmt-api endpoint (parsing of eszip happens on server-side
* Support for deno.json files
* Support for adding static files (eg: CSV, HTML, etc)
* Option to drag and drop files to the editor (handy when adding binary files like wasm)

(internal only) screencast: https://supabase.slack.com/archives/C02KMRX22NR/p1758850132342759

The backend changes are now deployed to production so we can merge this.


joshenlim (approved):
- Creating a new edge function
- Editing an existing edge function
- Drop file to add + save edge function (file persists after refreshing)
- Testing edge function in dashboard

Also pushed up a small UI nudge to edit the drag over state to be more consistent with our dashboard colors

PR #39992, opened 2025-10-29
refactor: use @tanstack/react-virtual for virtualization 

charislam:
Migrates from `react-window` to `@tanstack/react-virtual` for virtualization.

Motivation:
- `@tanstack/react-virtual` is entirely headless, which gives us more flexibility. Because `react-window` adds extra DOM elements, it's more difficult to control styling and scroll container placement when virtualizing elements that are very strict about their DOM structure, like tables.

Affected areas to test:
- [x] Storage buckets menu
- [x] Storage file explorer
- [x] Table Editor menu
- [x] Notifications popover (easiest to test this by just overriding the notifications response)

Resolves FE-1988


awaseem (approved):
LGTM

PR #39995, opened 2025-10-30
studio(feat): nested help and feedback

dnywh:

- UI feature

- Help and Feedback are two distinct entry points into a very busy top nav

## What is the new behavior?

- Help and Feedback are nested in the UserDropdown

## To do

Still to come:

- [ ] Restore `IS_PLATFORM` checks in the right places (e.g. not showing feedback, support to platform users)


dnywh:
Cancelling as we decided to keep _Feedback_ as its own dedicated navigation item

PR #40001, opened 2025-10-30
feat(etl): Improve destination panel UI

iambriccardo:
This PR improves the ETL UI with the following improvements:
- Each destination is gated behind a feature flag, enabling us to control which destinations are available to our users.
- The copy of the entire ETL UI has been improved and aligned.
- A new UI for changing destinations has been added which also makes it clear that changing a destination after pipeline creation is not possible.
- Better handling of required fields has been added, with a clear UI showing which fields are mandatory and properly disabling buttons when not all required fields are filled in.
- Now there is a `Restart` and `Stop` button for all active states of a pipeline, to give our users more control.
- A subtitle has been added under the replication title to clearly specify what replication is.
- A better division of sections of the panel has been implemented, to clearly define the hierarchy of information.

### Screenshots

UI when no destinations are available
image

Subtitle
image

New destination selection
image

Updating a destination
image

New stop and restart buttons
image

Improved advanced settings
image


joshenlim (approved):
Verified UI changes on staging preview!

PR #40009, opened 2025-10-30
Add a grace period heading

saltcod:
Adds a new heading to billing faq to address grace periods


pamelachia (approved):
clarified the auto removal part

PR #40021, opened 2025-10-30
fix(storage ui v2): virtualize files buckets table

charislam:
Resolves FE-1989


apps/studio/components/ui/VirtualizedTable.tsx
```diff
@@ -0,0 +1,238 @@
...
+  )
+}
+
+export const VirtualizedTableHeader = TableHeader
+
+export const VirtualizedTableHead = forwardRef<
+  HTMLTableCellElement,
+  React.ComponentProps<typeof TableHead>
+>(({ className, ...props }, ref) => {
+  return <TableHead ref={ref} className={cn('sticky top-0 z-10 bg', className)} {...props} />
```
joshenlim: probably use `bg-200` here to be consistent with `TableHead`?

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
joshenlim, line 1: another observation - because of the virtualization, there's a tendency for the header widths to shift based on the width of the bucket name column, probably set a fixed width?

https://github.com/user-attachments/assets/cbd29e08-648a-4d49-8947-db178ca34e00

joshenlim: also very tiny nit haha - but the border bottom on the header goes away once you scroll

image

apps/studio/components/ui/VirtualizedTable.tsx
charislam: ahh good point, I just roughly eyeballed the colours

apps/studio/components/interfaces/Storage/FilesBuckets.tsx
dnywh: Agree. Perhaps we should also show a slight shadow underneath `TableHead` to reinforce underlying contents. But I can do that in post.

dnywh (approved):
Pulled locally and tested with 100 fake buckets. Looks great. Some non-blocking feedback about `TableHead` styling. Thanks for this!

joshenlim (approved):
Verified changes on staging preview for both virtualized and unvirtualized UI

PR #40025, opened 2025-10-31
Fix: Bring back roles filter in Query performance

awaseem:

YES

Bug fix: bring back roles filter for query performance

You can't filter by roles within query performance as you use to be able to 

## What is the new behavior?

You can now filter by roles

## Additional context

Test:
- Go to query performance
- Click roles, you can filter by the ones there
- You can also append ?role=postgres and have the filter work


kemaldotearth (approved):
Nice catch! Thanks for this. I made a small adjustment re: hiding it on the pg_stat_monitor version of the Query Performance view. Roles filter doesn't quite work there mainly due to how we consume the data.

PR #40026, opened 2025-10-31
Unify Inbox and Advisor

SaxonF:
image

This is the last piece of work surrounding a redesign of our advisor experience which moves advisor up into the top bar. This PR merges the existing Inbox/Notifications with the new Advisor sidebar. It treats advisor issues and notifications similar in concept but still able to be separated via filters.

A few notes:
- Advisor issues not available at org level just yet, however we want a consistent experience moving between project<>org, so instead of hiding items we just direct user to select a project. In future, let's work on making advisors accessible outside of a project / across projects.
- Notifications are limited to 10 for this first release. Query supports pagination but for now let's keep it simple with the assumption we extend once we decide how notifications/advisors work better together from the back-end
- Sidebar has been lifted up into DefaultLayout so that they are accessible at the org level
- To allow this, when registering sidebars you can now provide an "enabled" argument which in the case of Assistant and Editor is set to false when at org level

To test:
- Toggle advisor sidebar and notice messages tab
- View messages in project level and at org level
- Toggle other sidebars and move between org <> project to make sure it works as expected


apps/studio/components/ui/AdvisorPanel/AdvisorDetail.tsx
```diff
@@ -0,0 +1,40 @@
...
+interface AdvisorDetailProps {
+  item: AdvisorItem
+  projectRef: string
+  onUpdateNotificationStatus?: (id: string, status: 'archived' | 'seen') => void
+}
+
+export const AdvisorDetail = ({
+  item,
+  projectRef,
+  onUpdateNotificationStatus = () => {},
```
alaister: Small nit: use `noop()` (you'll need to add an import) instead of `() => {}` in React component default arguments. If you ever pass `onUpdateNotificationStatus` to a `useEffect()` dependency array with the default value, it'll cause the `useEffect()` to rerun on every render. A fun footgun of React 😅

```suggestion
  onUpdateNotificationStatus = noop,
```

alaister (approved):
LGTM!

- Tested on preview

One small nit + some of the copilot nits look worthwhile

alaister (changes requested):
Actually I just checked self-hosted and it looks like this is broken due to the `/api/platform/notifications` endpoint being missing. It should just be a case of disabling it as I'm guessing notifications aren't relevant for self-hosted.

SaxonF:
@alaister this should be resolved now. Checking for IS_PLATFORM as part if query enable and not showing messages tab. We previously had the condition in layout header but since the merge we have to add to panel.

alaister (approved):
LGTM!

PR #40029, opened 2025-10-31
fix responsive issues

SaxonF:
Makes a number of small changes to improve responsiveness throughout the dashboard. A user is more likely to encounter these issues due to more emphasis being put on right sidebar. Uses container queries for more control.


joshenlim (approved):
Did a quick skim of the updated pages - generally LGTM

PR #40043, opened 2025-10-31
Update lints for self-hosted

joshenlim:
Fixes FE-2059

Note: Hold for now, still trying to figure out if there's actually a need to remove that WHERE clause below for it to work

## Context

Our lints for self-hosted hasn't been updated with what's on [here](https://github.com/supabase/splinter/blob/main/splinter.sql) in a while so it was omitted a couple of new lints like `rls_disabled_in_public`.

## Important note
- Self hosted lints query doesnt set the `pgrst.db_schemas` setting value before running the lints SQL
  - Which means that any of the lints which are checking against `current_setting('pgrst.db_schemas', 't')` will not return any results.
- This PR appends to add that setting value before running the lint query with `set local pgrst.db_schemas` which should work
- However, self hosted has no context of which schemas are being exposed atm
 - On the local API route `config/postgrest`, the `db_schemas` are being hard-coded which is incorrect
 - This value is being set properly at the docker compose level [here](https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml#L183) (Set via an env var [here](https://github.com/supabase/supabase/blob/474a78721e510301d15ca9dbd41f05ce10fa29e5/docker/.env.example#L55))
- I'm temporary patching this by hardcoding the exposed schemas when getting lints but we need to figure out a way separately for the self-hosted dashboard to retrieve the exposed schemas value from docker-compose as the source of truth

## To test

This needs to be tested on self-hosted local (`pnpm run dev:studio-local`)

- [ ] Verify that for a table in the public schema that has RLS disabled, the unrestricted badge shows up in the table editor

### DIFF
--- a/apps/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
+++ b/apps/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
@@ -89,6 +89,8 @@ const EntityListItem = ({
     projectRef: project?.ref,
   })
 
+  console.log(lints)
+
   const tableHasLints: boolean = getEntityLintDetails(
     entity.name,
     'rls_disabled_in_public',
--- a/apps/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
+++ b/apps/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
@@ -89,8 +89,6 @@ const EntityListItem = ({
     projectRef: project?.ref,
   })
 
-  console.log(lints)
-
   const tableHasLints: boolean = getEntityLintDetails(
     entity.name,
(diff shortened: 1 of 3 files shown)


apps/studio/components/layouts/TableEditorLayout/EntityListItem.tsx
```diff
@@ -89,6 +89,8 @@ const EntityListItem = ({
     projectRef: project?.ref,
   })
 
+  console.log(lints)
+
```
awaseem:
```suggestion
```

awaseem (approved):
Minor change

PR #40044, opened 2025-10-31
feat: make log table scrollable FE-2060

jordienr:
Makes log table scrollable making it easier for users to compare log results in Log Explorer.

CleanShot 2025-10-31 at 16 14 09


kemaldotearth (approved):
Nice!

PR #40057, opened 2025-11-01
fix: resolve circular dependency in Docker Compose setup

AkprasadoP:

YES

Bug fix

Self-hosting setup fails with a circular dependency error in Docker Compose, causing the Postgres (`db`) container to restart indefinitely.

Root cause: `db → vector → analytics → db` circular dependency.

## What is the new behavior?

- Removed `vector` dependency from `db` service  
- Increased retry duration for vector log sinks from 10s → 30s  
- Added retry backoff configuration  
- Docker Compose now validates without circular dependency errors  
- All services start in the correct order without infinite restarts

## Additional context

Fixes the blocking circular dependency issue in Supabase self-hosting setup.  
Tested locally with successful startup of all containers.


aantti:
Thanks, will address this in PR #42857

There's no circular dependency though... Seem to be the following:

- vector (no deps)
- db (depends on vector)
- analytics (depends on db)
- kong (depends on analytics)

PR #40068, opened 2025-11-02
Add link to function and table editor

SaxonF:
Adds a link from trigger list to the corresponding function as well as a link to table via table editor.

To test:
- Create a trigger and function
- Go to triggers list and notice function column now links to function
- Click link and notice takes to function page with search applied

Note that in future we could open the function sidebar or definition on link click but holding off on that for now due to a few moving sidebar parts.

### DIFF
--- a/apps/studio/components/interfaces/Database/Triggers/TriggersList/TriggerList.tsx
+++ b/apps/studio/components/interfaces/Database/Triggers/TriggersList/TriggerList.tsx
@@ -1,10 +1,12 @@
+import { PostgresTrigger } from '@supabase/postgres-meta'
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { includes, sortBy } from 'lodash'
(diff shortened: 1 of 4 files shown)


apps/studio/components/interfaces/Database/Triggers/TriggersList/TriggerList.tsx
```diff
@@ -111,15 +113,33 @@ const TriggerList = ({
...
+            )}
           </TableCell>
 
           <TableCell className="space-x-2">
-            <p title={x.function_name} className="truncate">
-              {x.function_name}
-            </p>
+            {x.function_name ? (
+              <Link
+                href={`/project/${projectRef}/database/functions?search=${x.function_name}`}
```
joshenlim: nit here but we'll also need to consider the function schema the function selected might not be in the public schema (i'll hop on!)

joshenlim (approved):

PR #40069, opened 2025-11-03
feat(studio): email template configuration

dnywh:

- Feature
- Resolves DEPR-177

- Security emails can only be turned on/off from the root email templates page
- ~~Email templates that are dependent on a setting to fire don’t communicate this to the user~~
- Spam validation was hard to read and interact with

## What is the new behavior?

- Security emails can now be turned on/off from within their respective template page
- ~~Email templates that are dependent on a setting to fire now communicate this to the user, if applicable~~
- Miscellaneous copywriting and design improvements to templates and SMTP presentation
- Spam validation is now a first-class citizen

| Before | After |
| --- | --- |

## Known issues

The following issues pre-date this work but would be nice to fix:

- [ ] Switching between _Source_ and _Preview_ (and/or back again) makes the window jump to the top

These were introduced in this PR:

- [ ] ~~The related setting is **assumed off** when it’s actual state should be read~~
- [ ] Triggering the spam filter, then replacing the offending text with ‘clean’ text does not re-enable the _Save changes_ button

## To test

- [ ] Does the `[templateId]` page correctly represent the state of the security email?
  - [ ] Does changing (and saving) the switch in the `[templateId]` page truly change the state of the template?
- [ ] Does adding problematic “spammy” text show the correct warning(s) table?
  - [ ] Does removing the text re-enable the primary button?
  - [ ] Does using built-in SMTP _block_ the template from being saved?
  - [ ] Does using custom SMTP just _warn_ the user whilst saving the template?
- [ ] Do both the root templates and each `[templateId]` page work as expected when the feature preview is disabled?

### Spam template

Here is a reliable email template body for triggering the spam filter:

```txt
Congratulations!

You've been selected to receive a special reward worth up to $1000. All you need to do is confirm your information within 24 hours.

- Click here to claim your prize: http:// example-rewards.click/claim

Don't miss this limited-time opportunity! Your unique reward ID: #WIN-983245

Best regards, Rewards Center Team
```

### Custom SMTP

You can just add fake details to test custom SMTP.


apps/studio/components/interfaces/Auth/EmailTemplates/SpamValidation.tsx
dnywh, line 1: Instead of permanently showing an entire row with:

> Email content is unlikely to be marked as spam

...for most users, let’s just make this component focus on rendering the contents of resulting actual spam issues, and show that conditionally.

apps/studio/components/interfaces/Auth/EmailTemplates/TemplateEditor.tsx
```diff
@@ -284,76 +288,80 @@ export const TemplateEditor = ({ template }: TemplateEditorProps) => {
...
+                  width={60}
+                  options={['preview', 'source']}
+                  activeOption={activeView}
+                  onClickOption={(option: 'source' | 'preview') => setActiveView(option)}
+                  borderOverride="border-muted"
+                />
+              </div>
+              {activeView === 'source' ? (
+                <>
+                  <div className="overflow-hidden rounded-md border dark:border-control overflow-hidden [&_.monaco-editor]:outline-0 [&_.monaco-editor-background]:!bg-surface-200/30 [&_.monaco-editor_.margin]:!bg-surface-200/30 dark:[&_.monaco-editor-background]:!bg-surface-300 dark:[&_.monaco-editor_.margin]:!bg-surface-300">
```
dnywh, line 298: Style the code editor so it’s similar to other fields and textfields. Insetting slightly also helps with readability in the card.

This is quite different to, say, Edge Functions, where a full-bleed code editor is warranted.

apps/studio/components/interfaces/Auth/EmailTemplates/TemplateEditor.tsx
```diff
@@ -284,76 +288,80 @@ export const TemplateEditor = ({ template }: TemplateEditorProps) => {
...
+                </>
+              )}
             </CardContent>
-            <CardFooter className="justify-end space-x-2">
+
+            {spamRules.length > 0 && (
+              <CardContent>
+                <SpamValidation validationResult={validationResult} />
+              </CardContent>
+            )}
```
dnywh: As mentioned above: only show the SpamValidation component when there are spam issues.

apps/studio/components/interfaces/Auth/SmtpForm/SmtpForm.tsx
dnywh, line 1: All copywriting and component improvements.

apps/studio/components/interfaces/Auth/AuthTemplatesValidation.tsx
dnywh, line 1: The phrase “Subject heading” reads a bit funny/redundant to me. Just “Subject” is enough, I think.

apps/studio/pages/project/[ref]/auth/templates/[templateId].tsx
```diff
@@ -1,23 +1,44 @@
+import { zodResolver } from '@hookform/resolvers/zod'
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { useEffect } from 'react'
+import { useForm } from 'react-hook-form'
+import { toast } from 'sonner'
+import { z } from 'zod'
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/auth/templates/[templateId].tsx:8]: Bringing in form validation because we are now handling the template on/off within the component (for security emails).

apps/studio/pages/project/[ref]/auth/templates/[templateId].tsx
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/auth/templates/[templateId].tsx:1]: this one's just a feedback but it looks a little weird imo that we have a mix of horizontal (name) and vertical (message) layout for the form input here
image

it could be my own bias too but it feels more natural with the original layout
image

joshenlim:
Not related to this PR i think but there's an odd layout bug i noticed with this UI when you scroll down
image

joshenlim (commented):
Code changes LGTM - but there's a number of issues that i reckon we should resolve first before getting this through 🙂 🙏 
- Triggering the spam filter, then replacing the offending text with ‘clean’ text does not re-enable the Save changes button
- Layout height scrolling issue on emails page
- (Low prio) Switching between Source and Preview (and/or back again) makes the window jump to the top

apps/studio/pages/project/[ref]/auth/templates/[templateId].tsx
Comment (dnywh) [reply apps/studio/pages/project/[ref]/auth/templates/[templateId].tsx:1]: I was wondering the same in the back of my mind. Thanks for flagging!

apps/studio/components/interfaces/Auth/EmailTemplates/TemplateEditor.tsx
```diff
@@ -1,13 +1,14 @@
...
 import { editor } from 'monaco-editor'
 import { useEffect, useMemo, useRef, useState } from 'react'
 import { useForm } from 'react-hook-form'
 import ReactMarkdown from 'react-markdown'
 import { toast } from 'sonner'
 
 import { useParams } from 'common'
+import { useIsSecurityNotificationsEnabled } from 'components/interfaces/App/FeaturePreview/FeaturePreviewContext'
 import CodeEditor from 'components/ui/CodeEditor/CodeEditor'
+import TwoOptionToggle from 'components/ui/TwoOptionToggle'
```
dnywh, line 10: More appropriate component for switching between different views of the same thing. Much like Data and Definition on Table Editor, or Segmented Control on iOS.

apps/studio/components/interfaces/Auth/EmailTemplates/TemplateEditor.tsx
```diff
@@ -75,9 +74,13 @@ export const TemplateEditor = ({ template }: TemplateEditorProps) => {
   const [bodyValue, setBodyValue] = useState((authConfig && authConfig[messageSlug]) ?? '')
   const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false)
   const [isSavingTemplate, setIsSavingTemplate] = useState(false)
+  const [activeView, setActiveView] = useState<'source' | 'preview'>('source')
+  const [lastAttemptedSaveValues, setLastAttemptedSaveValues] = useState<{
+    formValues: { [key: string]: string }
+    bodyValue: string
+  } | null>(null)
```
dnywh: This approach to enabled/disabling the _Save changes_ button is necessary, as far as I can see. Otherwise when spam is detected, the save is blocked but `hasChanges` remains true because the values weren’t ever saved.

The currently-in-prod approach is to set `preventSaveFromSpamCheck` in the disabled condition. I believe that depends on the current `validationResult`, which can change, making the button state unreliable.

So this is an attempt at a simpler fix: track a flag indicating a spam-blocked save attempt, and using that in the disabled condition.

apps/studio/components/interfaces/Auth/EmailTemplates/EmailTemplates.tsx
```diff
@@ -191,7 +188,7 @@ export const EmailTemplates = () => {
                               )}
                             </Link>
 
-                            <div className="flex items-center gap-4 group h-full pl-2">
+                            <div className="flex items-center gap-4 h-full pl-2 relative">
```
dnywh, line 191: Fixes the overflow issue. Documented in `switch.mdx` above.

apps/design-system/content/docs/components/switch.mdx
dnywh, line 1: The overflow issue was hard to pin down, and turned out to be an issue with `Switch` requiring an ancestor to be positioned `relative`. I thought it’d be good to document this on the component’s design system page.

apps/studio/components/interfaces/Auth/EmailTemplates/TemplateEditor.tsx
joshenlim: hmm one sec - i don't think we need to save this value

will look into simplifying this, the save being blocked was probably the issue (although i can't recall what it was before)

joshenlim: can verify that removing this state will still allow users to hit save now after the validation fails

joshenlim (approved):

PR #40070, opened 2025-11-03
switch to default

SaxonF:
Fixes an issue where chat form overflow appears under button due to no fill. 

New:

image

Old:

image


joshenlim (approved):

PR #40089, opened 2025-11-03
test: improve local setup with docker for e2e tests in logflare

nelsonmestevao:
When doing integration tests with logflare, analytics will not be a necessary service and I am making some overrides in some fields to make it easy.

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -92,8 +92,8 @@ ENABLE_PHONE_AUTOCONFIRM=true
 # Studio - Configuration for the Dashboard
 ############
 
-STUDIO_DEFAULT_ORGANIZATION=Default Organization
-STUDIO_DEFAULT_PROJECT=Default Project
+STUDIO_DEFAULT_ORGANIZATION="Default Organization"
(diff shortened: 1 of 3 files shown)


aantti:
Hey @nelsonmestevao, I'm having difficulties interpreting this PR :) This looks like your internal project that accidentally became a PR? In case I'm wrong, could you possibly provide extended description and rationale for the changes?

nelsonmestevao:
Hi @aantti. Sorry, it was my fault to not provide more details upfront.

I am working on integration tests between logflare and supabase. The approach I am trying is the self-host flow from supabase to test against each PR in logflare. By creating a `docker-compose.e2e.yml` I am looking into adding the necessary configurations for that. cc/ @Ziinc

Ziinc:
@nelsonmestevao could we keep this override file in the logflare repo and just copy the file over in the ci ?

nelsonmestevao:
> @nelsonmestevao could we keep this override file in the logflare repo and just copy the file over in the ci ? 

Yes. I was just thinking if the setup in the Supabase repo changes, this should also be edited and might be hard to keep track.

I will move it there for now 🙏

PR #40095, opened 2025-11-03
feat(studio): log drains entry point from logs

kemaldotearth:

YES

This adds a small entry point from our logs view to log drains for awareness more than anything.


apps/studio/components/layouts/LogsLayout/LogsSidebarMenuV2.tsx
```diff
@@ -366,9 +367,27 @@ export function LogsSidebarMenuV2() {
...
+              React.cloneElement(type.icon, { height: 20, width: 20 })
+            )}
+          </div>
+        }
+        actions={
+          <Button asChild type="default">
+            <Link href={`/project/${ref}/settings/log-drains`}>Go to Log Drains</Link>
+          </Button>
+        }
+      />
```
dnywh, line 392: I believe this component is for the old logs, right? If so, this feedback can likely be ignored since I understand we’re moving away from it:

- This panel looks a bit funny next to the centered text in the aside above it
- Maybe a `Separator` above this would help make it not look like a child of _Queries_

dnywh: P.S. is this `React.cloneElement` the reason for the peculiar `import React, { useState } from 'react'` import?

Seems like a funny bit of syntax to map through icons, but I’ll leave that to frontender(s) to assess. It’s more likely that I need to brush up.

apps/studio/components/ui/Logs/LogsExplorerHeader.tsx
dnywh, line 79: Not part of this PR but `text-brand-link` would make the green text more accessible.

Or we could just use `InlineLink`, as it probably doesn’t need to be green text.

dnywh, line 97: Again, not part of this PR but whilst I’m looking at it:

I wonder if a dropdown would be a better pattern here, given there are so many tabs

apps/studio/components/ui/DownloadResultsButton.tsx
dnywh, line 1: FYI this button has no gap on old logs:

Image

apps/studio/components/ui/DataTable/FilterSideBar.tsx
```diff
@@ -60,6 +63,23 @@ export function FilterSideBar({ dateRangeDisabled }: FilterSideBarProps) {
           />
         )}
         <DataTableFilterControls dateRangeDisabled={dateRangeDisabled} />
+        <FeaturePreviewSidebarPanel
```
dnywh, line 66: Similar feedback as above: maybe a separator between the actual filters and this panel. But that sort of breaks the visuals here.

apps/studio/components/ui/Logs/LogsExplorerHeader.tsx
```diff
@@ -62,7 +62,7 @@ const LogsExplorerHeader = ({ subtitle }: LogsExplorerHeaderProps) => {
               onClick={() => setShowReference(true)}
               icon={<BookOpen strokeWidth={1.5} />}
             >
-              Field Reference
+              <span>Field Reference</span>
```
dnywh, line 65: Why the `span`? Seems to render fine without it

dnywh: I like this ‘naked’ button style FYI. I should use it more

packages/ui/src/components/SidePanel/SidePanel.tsx
```diff
@@ -106,11 +106,7 @@ const SidePanel = ({
 
   return (
     <Dialog.Root open={open} onOpenChange={handleOpenChange} defaultOpen={defaultOpen}>
-      {triggerElement && (
-        <Dialog.Trigger asChild className={__styles.trigger}>
-          {triggerElement}
-        </Dialog.Trigger>
-      )}
+      {triggerElement && <Dialog.Trigger asChild>{triggerElement}</Dialog.Trigger>}
```
dnywh, line 109: Not sure what the change here is for but seems to work fine in testing

dnywh (approved):
Tested on local. Works as expected, and looking good!

Few superficial comments, some unrelated to this PR (sorry).

apps/studio/components/ui/DownloadResultsButton.tsx
kemaldotearth: Fixed this irrespective of PR. Should also horizontally scroll now on smaller viewports like our other toolbar areas.

packages/ui/src/components/SidePanel/SidePanel.tsx
kemaldotearth: We had some old sass styling applied to the element that was breaking the child button's styling. Seems like this was the only place we use it, so remove the class.

apps/studio/components/layouts/LogsLayout/LogsSidebarMenuV2.tsx
kemaldotearth: `React.cloneElement` was used to be able to pass sizing props to an icon coming from a constants object. I saw one other use of it elsewhere and gave it a bash!

PR #40107, opened 2025-11-03
Fix: Updated to use optimized search columns when getting user counts

awaseem:

YES

Optimizes the users count query to remove like statements clauses

Unneed performance hit to the database

## What is the new behavior?

Column specific logic per filter

## Additional context

N/A


apps/studio/components/interfaces/Auth/Users/UsersFooter.tsx
```diff
@@ -42,6 +43,13 @@ export const UsersFooter = ({
       filter: filter === 'all' ? undefined : filter,
       providers: selectedProviders,
       forceExactCount,
+      // Force exact count for optimized search since result sets are small
+      // forceExactCount: forceExactCount || specificFilterColumn !== 'freeform',
```
awaseem: I wasn't super sure if we always want a specific count when the filter is applied. This is kinda the big reason why all the estimates are so strange. Right now the count query users an explain, inside the `count_estimate` function so its counting the stats row? We can remove this but wanted to ask if we should use the exact value here now that we have proper count filtering

packages/pg-meta/src/sql/studio/get-users-count.ts
```diff
@@ -7,46 +9,72 @@ export const getUsersCountSQL = ({
...
+    } else if (column === 'phone') {
+      const range = stringRange(formattedKeywords)
+      conditions.push(`phone >= '${range[0]}'${range[1] ? ` and phone < '${range[1]}'` : ''}`)
+    } else if (column === 'id') {
+      const isMatchingUUIDValue = prefixToUUID(formattedKeywords, false) === formattedKeywords
       conditions.push(
-        `(raw_app_meta_data->>'providers')::jsonb ?| array[${providers.map((p) => `'${p}'`).join(', ')}]`
+        isMatchingUUIDValue
+          ? `id = '${formattedKeywords}'`
+          : `id >= '${prefixToUUID(formattedKeywords, false)}' and id < '${prefixToUUID(formattedKeywords, true)}'`
```
awaseem: A lot of this logic just copied over from the paginated query WITHOUT the startAt, assuming the count does not need to be paginated thats why I didn't copy that logic over. Let me know if I'm totally wrong here

awaseem (commented):
Let me know if this makes any sense @joshenlim

packages/pg-meta/test/sql/studio/get-users-common.test.ts
```diff
@@ -0,0 +1,208 @@
+import { expect, test, describe } from 'vitest'
```
awaseem, line 1: Doesn't fit the style of integration tests only so we can move them, but I added some unit tests for these functions

apps/studio/data/auth/keys.ts
```diff
@@ -39,6 +41,7 @@ export const authKeys = {
       filter: string | undefined
       providers: string[] | undefined
       forceExactCount?: boolean
+      column?: OptimizedSearchColumns
```
joshenlim, line 44: note to self: just wondering whether we should be including this in the key for `usersInfinite` (wondering why is there a different here between `usersCount` and `usersInfinite`)

or on the flip side - why do we need this in the key for `usersCount`

joshenlim: ahh okay we need `column` in the key because we have 2 places retrieving count

UsersV2 keeps track of the full count
UsersFooter syncs the count with whats in view

apps/studio/components/interfaces/Auth/Users/UsersFooter.tsx
joshenlim: hmm from a performance POV i think we should still default to false for this unless unified search
so that we default to checking rel_tuples initially rather than running `count(*)`

> estimates are so strange

this is alright imo - it's expected that the estimates are not always accurate and there's an escape hatch to fetch the actual row count which we show a warning before we run the query

altho please do let me know if i might be overlooking anything! 🙏

packages/pg-meta/src/sql/studio/get-users-count.ts
joshenlim: yup count doesnt need to be paginated - should be considering the total count, not count per page

apps/studio/components/interfaces/Auth/Users/UsersFooter.tsx
awaseem: Make sense, I'll remove the comment

packages/pg-meta/test/sql/studio/get-users-count.test.ts
```diff
@@ -0,0 +1,357 @@
+import { expect, test, describe } from 'vitest'
```
soedirgo: These unit tests are brittle as it's testing the implementation instead of the behavior - any future refactors would very likely trigger a failure in these tests, even if they don't introduce any bugs. Can we adapt it to something more like [this](https://github.com/supabase/supabase/blob/c474d3dbbd1e761c97644624c0243b9f7a95c397/packages/pg-meta/test/schemas.test.ts#L44)?

One smoke test for the trivial case and one regression test for the `stringMatch` bug should be enough

Sorry for the extra work 🙏

soedirgo: `get-users-count.test.ts` are fine though

apps/studio/components/interfaces/Auth/Users/UsersFooter.tsx
```diff
@@ -42,6 +43,10 @@ export const UsersFooter = ({
       filter: filter === 'all' ? undefined : filter,
       providers: selectedProviders,
       forceExactCount,
+      // Use optimized search when filtering by specific column
+      ...(specificFilterColumn !== 'freeform'
+        ? { column: specificFilterColumn as OptimizedSearchColumns }
+        : { column: undefined }),
```
avallete, line 50: **question**

Why doesn't this `specificFilterColumn` use the `SpecificFilterColumn` type ?

packages/pg-meta/src/sql/studio/get-users-count.ts
```diff
@@ -7,46 +9,72 @@ export const getUsersCountSQL = ({
...
       )
     }
+  } else {
+    // Unified search mode - apply all filters
+    if (hasValidKeywords) {
+      // [Joshen] Escape single quotes properly
+      const formattedKeywords = keywords.replaceAll("'", "''")
+      conditions.push(
+        `id::text ilike '%${formattedKeywords}%' or email ilike '%${formattedKeywords}%' or phone ilike '%${formattedKeywords}%'`
+      )
```
avallete: **question**

Would it be a good case for using `literal(keywords)` here  to escape things ? Like we do for https://github.com/supabase/supabase/blob/a00fdc1c71b37a63c4d6026e4eae456011746b73/packages/pg-meta/src/query/Query.utils.ts#L121



```suggestion
      const escapedSearchKeywords = literal('%' + formattedKeywords + '%')
      conditions.push(
        `id::text ilike ${escapedSearchKeywords} or email ilike ${escapedSearchKeywords} or phone ilike ${escapedSearchKeywords}`
      )
```

packages/pg-meta/src/sql/studio/get-users-count.ts
```diff
@@ -7,46 +9,72 @@ export const getUsersCountSQL = ({
...
+    } else if (filter === 'unverified') {
+      conditions.push(`email_confirmed_at IS NULL AND phone_confirmed_at IS NULL`)
+    }
+
+    if (providers && providers.length > 0) {
+      // [Joshen] This is arguarbly not fully optimized, but at the same time not commonly used
+      // JFYI in case we do eventually run into performance issues here when filtering for SAML provider
+      if (providers.includes('saml 2.0')) {
+        conditions.push(
+          `(select jsonb_agg(case when value ~ '^sso' then 'sso' else value end) from jsonb_array_elements_text((raw_app_meta_data ->> 'providers')::jsonb)) ?| array[${providers.map((p) => (p === 'saml 2.0' ? `'sso'` : `'${p}'`)).join(', ')}]`.trim()
```
avallete: **nitpick**

`literal` escape should also work for arrays ensuring each element in it is properly escaped.

```suggestion
          `(select jsonb_agg(case when value ~ '^sso' then 'sso' else value end) from jsonb_array_elements_text((raw_app_meta_data ->> 'providers')::jsonb)) ?| array[${literal(providers.map((p) => (p === 'saml 2.0' ? 'sso' : p)))}]`.trim()
```

packages/pg-meta/test/sql/studio/get-users-count.test.ts
avallete: Same comment here, would rather have only two tests that ensure that the queries run on a database and retrieve the data expected rather than having tests over what the query should contain / how each sub-component works internally.

`withDatabase` tests wrapper will give you a supabase DB (not initialized though), but should be able to implement something 90% similar to how go-true works by adding the init gotrue migration: https://github.com/supabase-community/gotrue/blob/master/migrations/00_init_auth_schema.up.sql

Something like:

```ts
import { expect, test, beforeAll, afterAll } from 'vitest'
import { createTestDatabase, cleanupRoot } from './db/utils'
import { randomUUID } from 'crypto'

beforeAll(async () => {
  // Any global setup if needed
})

afterAll(async () => {
  await cleanupRoot()
})

const withTestDatabase = (
  name: string,
  fn: (db: Awaited<ReturnType<typeof createTestDatabase>>) => Promise<void>
) => {
  test(name, async () => {
    const db = await createTestDatabase()
    db.executeQuery(`
      -- auth.users definition

CREATE TABLE IF NOT EXISTS auth.users (
	instance_id uuid NULL,
	id uuid NOT NULL UNIQUE,
	aud varchar(255) NULL,
	"role" varchar(255) NULL,
	email varchar(255) NULL UNIQUE,
	encrypted_password varchar(255) NULL,
	confirmed_at timestamptz NULL,
	invited_at timestamptz NULL,
	confirmation_token varchar(255) NULL,
	confirmation_sent_at timestamptz NULL,
	recovery_token varchar(255) NULL,
	recovery_sent_at timestamptz NULL,
	email_change_token varchar(255) NULL,
	email_change varchar(255) NULL,
	email_change_sent_at timestamptz NULL,
	last_sign_in_at timestamptz NULL,
	raw_app_meta_data jsonb NULL,
	raw_user_meta_data jsonb NULL,
	is_super_admin bool NULL,
	created_at timestamptz NULL,
	updated_at timestamptz NULL,
	CONSTRAINT users_pkey PRIMARY KEY (id)
);
CREATE INDEX IF NOT EXISTS users_instance_id_email_idx ON auth.users USING btree (instance_id, email);
CREATE INDEX IF NOT EXISTS users_instance_id_idx ON auth.users USING btree (instance_id);
comment on table auth.users is 'Auth: Stores user login data within a secure schema.';

-- auth.refresh_tokens definition

CREATE TABLE IF NOT EXISTS auth.refresh_tokens (
	instance_id uuid NULL,
	id bigserial NOT NULL,
	"token" varchar(255) NULL,
	user_id varchar(255) NULL,
	revoked bool NULL,
	created_at timestamptz NULL,
	updated_at timestamptz NULL,
	CONSTRAINT refresh_tokens_pkey PRIMARY KEY (id)
);
CREATE INDEX IF NOT EXISTS refresh_tokens_instance_id_idx ON auth.refresh_tokens USING btree (instance_id);
CREATE INDEX IF NOT EXISTS refresh_tokens_instance_id_user_id_idx ON auth.refresh_tokens USING btree (instance_id, user_id);
CREATE INDEX IF NOT EXISTS refresh_tokens_token_idx ON auth.refresh_tokens USING btree (token);
comment on table auth.refresh_tokens is 'Auth: Store of tokens used to refresh JWT tokens once they expire.';

-- auth.instances definition

CREATE TABLE IF NOT EXISTS auth.instances (
	id uuid NOT NULL,
	uuid uuid NULL,
	raw_base_config text NULL,
	created_at timestamptz NULL,
	updated_at timestamptz NULL,
	CONSTRAINT instances_pkey PRIMARY KEY (id)
);
comment on table auth.instances is 'Auth: Manages users across multiple sites.';

-- auth.audit_log_entries definition

CREATE TABLE IF NOT EXISTS auth.audit_log_entries (
	instance_id uuid NULL,
	id uuid NOT NULL,
	payload json NULL,
	created_at timestamptz NULL,
	CONSTRAINT audit_log_entries_pkey PRIMARY KEY (id)
);
CREATE INDEX IF NOT EXISTS audit_logs_instance_id_idx ON auth.audit_log_entries USING btree (instance_id);
comment on table auth.audit_log_entries is 'Auth: Audit trail for user actions.';

-- auth.schema_migrations definition

CREATE TABLE IF NOT EXISTS auth.schema_migrations (
	"version" varchar(255) NOT NULL,
	CONSTRAINT schema_migrations_pkey PRIMARY KEY ("version")
);
comment on table auth.schema_migrations is 'Auth: Manages updates to the auth system.';
		
-- Gets the User ID from the request cookie
create or replace function auth.uid() returns uuid as $$
  select nullif(current_setting('request.jwt.claim.sub', true), '')::uuid;
$$ language sql stable;

-- Gets the User ID from the request cookie
create or replace function auth.role() returns text as $$
  select nullif(current_setting('request.jwt.claim.role', true), '')::text;
$$ language sql stable;
    `)
    try {
      await fn(db)
    } finally {
      await db.cleanup()
    }
  })
}

// Example of refactored tests using the wrapper
withTestDatabase('list with system schemas', async ({ executeQuery }) => {
    // Generate 5 unique user IDs and emails
  const users = Array.from({ length: 5 }, (_, i) => ({
    id: randomUUID(),
    email: `user${i + 1}@example.com`,
    aud: 'authenticated',
    role: 'authenticated',
  }))

  // Insert all 5 users in a single query
  const values = users
    .map((user) => `('${user.id}', '${user.email}', '${user.aud}', '${user.role}', NOW())`)
    .join(', ')
  await executeQuery(`
    INSERT INTO auth.users (id, email, aud, "role", created_at)
    VALUES ${values}
  `)
  const userCountSql = getUsersCountSQL({})
  const userCountResult = await executeQuery(userCountSql)
  expect(userCountResult).toBe(5)
})
```

awaseem: That makes sense, I'll trim the migration schema down and convert this to int. tests

apps/studio/components/interfaces/Auth/Users/UsersFooter.tsx
awaseem: Miss on my part, moving it to constants in the component lib and referencing here (in constants because freeform has no context in pg-meta types)

packages/pg-meta/src/sql/studio/get-users-count.ts
awaseem: This might be a better question for @hf because my SQL is bad :(

avallete (approved):
LGTM 🚀

soedirgo (approved):

alaister (approved):
LGTM!

- Tested on preview and self hosted preview

PR #40108, opened 2025-11-03
fix(support link): auto-fill organization when on org page

charislam:
If user clicks the Support Link from an org page (not a project page), pass along the organization slug in the query params to auto-fill the Support Form.

Resolves FE-1611


awaseem (approved):
tested in preview!, LGTM

PR #40110, opened 2025-11-03
fix(integrations page): overflow on integrations card loading states

charislam:
Resolves FE-1371


awaseem (approved):
Tested in Preview, LGTM

PR #40117, opened 2025-11-04
Fix branch merge request link

joshenlim:
## Context

Bug report that the links do not tally in the branches merge request page if the merge request corresponds to one in GH
image

Realised that this was due to `BranchRow` component using `BASE_PATH` to open an external URL which isn't required

## Changes involved

- Update `BranchRow` to use `Link` component instead to open links rather than manually calling `window.open` or `router.push`
- (Unrelated) Remove odd padding in `BranchManagementSection`
  image
- (Unrelated) Fix premature empty state for merge requests before branches are loaded

## To test

- [ ] Verify that merge requests from GH in the dashboard open the correct links when clicking on the branch name


alaister (approved):
LGTM!

- Tested on preview

PR #40125, opened 2025-11-04
feat: add min calls filter to Query Performance FE-2071

jordienr:
https://linear.app/supabase/issue/FE-2071/add-min-calls-filter-to-query-performance

## To Test
- go to query performance
- try to filter by min calls (e.g. 1000) 
- table should update and show correct results
- sorts and query filter should still work as expected


jeremyisatrecharm:
Unfortunately I am unqualified to give any helpful feedback/review here, but wanted to say: wow! That was fast. Thanks so much.

kemaldotearth (approved):
🚢

PR #40137, opened 2025-11-04
fix(mssql foreign tables): disallow sort on columns filtered for equality

charislam:
There is an edge case interaction between the Postgres query parser and MSSQL foreign tables, where the query parser may drop sort clauses that are redundant with applied filters. This leads to invalid MSSQL syntax, because the resulting query has a `limit` but no `sort`, and the user sees a confusing error message.

This PR detects this edge case on MSSQL foreign tables. There are three cases:
1. The user filters by a column, but there are still other columns available for sorting. The default search for a sorting column will leave out the filtered column.
2. The user filters by a column/several columns, and there are no more columns that can be used for sorting. We stop the query and show an admonition.
3. The user filters by a column, then tries to sort by the same column. We stop the query and show an admonition.

## Testing

- [ ] Make sure non-MSSQL tables behave as usual

On MSSQL tables:
- [ ] Find the default sorting column. Filter by that column. Make sure that column is not the sorting column in the final query
- [ ] Filter by all sortable columns. Make sure admonition appears and query does not run.
- [ ] Filter by a column and sort by the same column. Make sure admonition appears and query does not run.

Resolves FE-1990

### DIFF
--- a/apps/studio/components/grid/components/footer/pagination/Pagination.tsx
+++ b/apps/studio/components/grid/components/footer/pagination/Pagination.tsx
@@ -47,7 +47,11 @@ const RowCountSelector = ({
   )
 }
 
-export const Pagination = () => {
+type PaginationProps = {
+  disableForeignRowsQuery?: boolean
+}
+
+export const Pagination = ({ disableForeignRowsQuery = false }: PaginationProps) => {
   const { id: _id } = useParams()
   const id = _id ? Number(_id) : undefined
 
@@ -108,7 +112,7 @@ export const Pagination = () => {
       roleImpersonationState: roleImpersonationState as RoleImpersonationState,
     },
     {
-      enabled: isForeignTableSelected,
+      enabled: isForeignTableSelected && !disableForeignRowsQuery,
     }
   )
   const isLastPage = (rowsData?.rows ?? []).length < tableEditorSnap.rowsPerPage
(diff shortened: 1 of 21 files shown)


apps/studio/components/grid/components/footer/pagination/Pagination.tsx
```diff
@@ -47,7 +47,11 @@ const RowCountSelector = ({
   )
 }
 
-export const Pagination = () => {
+type PaginationProps = {
+  disableForeignRowsQuery?: boolean
```
soedirgo: nit: can we use positive states for variables? (i.e. `enableForeignRowsQuery`)

packages/pg-meta/src/sql/foreign-tables.ts
```diff
@@ -3,10 +3,17 @@ select
...
+  fs.srvname as foreign_server_name,
+  fdw.fdwname as foreign_data_wrapper_name,
+  handler.proname as foreign_data_wrapper_handler
 from
   pg_class c
   join pg_namespace n on n.oid = c.relnamespace
+  left join pg_foreign_table ft on ft.ftrelid = c.oid
+  left join pg_foreign_server fs on fs.oid = ft.ftserver
+  left join pg_foreign_data_wrapper fdw on fdw.oid = fs.srvfdw
+  left join pg_proc handler on handler.oid = fdw.fdwhandler
```
soedirgo: Let's do inner joins instead so these fields are non-nullable?

soedirgo (approved):

soedirgo (approved):

awaseem (approved):
LGTM

PR #40155, opened 2025-11-04
fix(logout): don't use returnTo on sign-in page if user intentionally logs out

charislam:
When user logs out from the user dropdown, they're redirected to the sign-in page via the withAuth redirect, which means that the returnTo param is used -- but if the user _intentionally_ logged out, this might not be the right behavior, i.e., if they want to log in to another account. Instead, if they chose to log out, we shouldn't use the returnTo param.

Since the /logout page already implements this behavior, I changed the "Log out" button to just route to the /logout page.

## Testing

- [x] Log out using the user dropdown button. There should be no `returnTo` on the log in page.
- [x] Log out using another method (e.g., delete the auth token from browser storage). There should be a `returnTo` param on the log in page.

Resolves FE-1842

### DIFF
--- a/apps/studio/components/interfaces/UserDropdown.tsx
+++ b/apps/studio/components/interfaces/UserDropdown.tsx
@@ -4,9 +4,8 @@ import Link from 'next/link'
 import { useRouter } from 'next/router'
 
 import { ProfileImage } from 'components/ui/ProfileImage'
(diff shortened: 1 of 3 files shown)


awaseem (approved):
Tested both, looks good to me

PR #40156, opened 2025-11-04
ci: add eslint ratcheting

charislam:
Add an ESLint ratchet script and actions.

- Ratchet script runs ESLint and reads a metadata file containing some baselines. For any tracked lint rule, if the number of violations exceeds the baseline, it fails. New rules can be added locally with the `--init --rule <rule-id>` option and pushed to the repo to be tracked.
- Ratchet action runs on each PR to make sure no additional ESLint violations are introduced.
- Ratchet update action runs weekly to decrease the baseline to the new lowest number if possible.

Tracking some of our most common rule warnings. The only one missing is `no-restricted-exports`, because there are legitimate reasons to add default exports (mandatory Next.js convention). We need to figure out if it's possible to configure this rule to ignore specific directories/files before we can start tracking it.

### DIFF
--- a/.github/workflows/studio-lint-ratchet.yml
+++ b/.github/workflows/studio-lint-ratchet.yml
@@ -1,0 +1,45 @@
+name: Ratchet studio lint checks
+
+on:
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'apps/studio/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  ratchet:
+    # Uses larger hosted runner as it significantly decreases build times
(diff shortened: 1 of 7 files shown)


.github/workflows/studio-lint-ratchet.yml
```diff
@@ -0,0 +1,49 @@
...
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
+
+      - name: Install deps
+        run: pnpm install --frozen-lockfile
+
+      - name: Run ratchet script 
+        run: pnpm -filter studio run lint:ratchet
```
ivasilov: This should be `--`, right?

charislam: oops yep!

awaseem (approved):
I'm pumped for this 💪

PR #40159, opened 2025-11-04
fix(table row context menu): hide actions and separators on protected schemas

charislam:
Protected schemas shouldn't have the "Edit row" and "Delete row" actions.

Resolves FE-1942


awaseem (approved):
Tested on preview, no strange issues. LGTM

PR #40184, opened 2025-11-05
Chore: add param routing on panels

fsansalvadore:
## What's new

Add param queryState routing with _nuqs_ when opening side panels and dialogs:

- Database Functions -> `create` / `update` / `delete`
- Database Enumerated Types -> `create` / `update` / `delete`
- Database Indexes -> `create` / `update` / `delete`
- Database Roles -> `create` / `delete`

Screenshot 2025-11-07 at 17 56 02
Screenshot 2025-11-07 at 17 55 13
Screenshot 2025-11-07 at 17 56 21
Screenshot 2025-11-07 at 17 56 31

## Custom `useQueryStateRouting` hook
Hook to abstract repetitive logic that uses _nuqs_ state to control dialogs and panels:
- for common actions on entities, you can specify any term: 
  - `edit` as `?edit=123`
  - `duplicate` as `?duplicate=123` // the entity id is of the entity to duplicate from
  - `delete` as `?delete=123`
- error handling:
  - when an `?edit` or `?duplicate` or `?delete` param is opened but the entity doesn't exist, the dialog/panel is automatically closed and a toast says "[entityName] not found"

Example usage if you need `new`, `edit`, `duplicate` and `delete`:
```tsx
  // Pure nuqs `useStateQuery` for booleans
  const [showCreateFunctionForm, setShowCreateFunctionForm] = useQueryState(
    'new',
    parseAsBoolean.withDefault(false).withOptions({ history: 'push', clearOnDefault: true })
  )

 const { value: typeToEditValue, setValue: setTypeToEditValue } = useQueryStateWithSelect({
    enabled: !!data,
    urlKey: 'edit',
    select: (id) => data?.find((type) => type.id.toString() === id),
    onError: (error) => {
      toast.error(error.message)
    },
  })
```

## Demo
Demo on Database Functions showing `new`, `edit`, `duplicate` and `delete`, browser history, deep linking and error handling:

https://github.com/user-attachments/assets/937dc4b5-38d5-45c5-bd32-4fa93c8ec964

## Testing

Open these pages and perform the associated actions:
- [Database Functions](https://studio-staging-git-chore-add-param-routing-supabase.vercel.app/dashboard/project/_/database/functions?new=true) -> `create` / `update` / `delete`
- [Database Enumerated Types](https://studio-staging-git-chore-add-param-routing-supabase.vercel.app/dashboard/project/_/database/types?new=true) -> `create` / `update` / `delete`
- [Database Indexes](https://studio-staging-git-chore-add-param-routing-supabase.vercel.app/dashboard/project/_/database/indexes?new=true) -> `create` / `update` / `delete`
- [Database Roles](https://studio-staging-git-chore-add-param-routing-supabase.vercel.app/dashboard/project/_/database/roles?new=true) -> `create` / `delete`

What's expected: 
- each action should append `?verb=...` to the url
- deep-linking the url with the param should open the panel/dialog
- browser history should open/close panels/dialogs
- using a wrong id should automatically close the panel and show a "[entityName] not found" toast


apps/studio/components/interfaces/Database/EnumeratedTypes/EnumeratedTypes.tsx
```diff
@@ -36,14 +34,40 @@ export const EnumeratedTypes = () => {
...
+    setSelectedId: setSelectedTypeIdToEdit,
+    entity: typeToEdit,
+    show: showTypeToEdit,
+  } = editQueryState!
+
+  const {
+    setSelectedId: setSelectedTypeIdToDelete,
+    entity: typeToDelete,
+    show: showTypeToDelete,
+  } = deleteQueryState!
```
awaseem: This seems like a bad idea, why force unwrap these vs. just handling the optionals. If these are undefined then best to set default values rather than force unwrapping

apps/studio/components/interfaces/Database/Functions/FunctionsList/FunctionsList.tsx
```diff
@@ -36,26 +35,75 @@ import {
...
+    if (isInlineEditorEnabled) {
+      setEditorPanelInitialPrompt('Create a new database function that...')
+      setEditorPanelValue(`create function function_name()
+returns void
+language plpgsql
+as $$
+begin
+  -- Write your function logic here
+end;
+$$;`)
```
awaseem: nit: Can we create a const for this rather then referencing the function snippet here? it looks a bit messy

apps/studio/components/interfaces/Database/Functions/FunctionsList/FunctionsList.tsx
```diff
@@ -114,12 +162,51 @@ const FunctionsList = ({
...
+    setSelectedId: setSelectedFunctionToDelete,
+    entity: functionToDelete,
+    show: showFunctionToDelete,
+  } = deleteQueryState!
+
+  const {
+    setSelectedId: setSelectedFunctionIdToDuplicate,
+    entity: functionToDuplicate,
+    show: showFunctionToDuplicate,
+  } = duplicateQueryState!
```
awaseem: Again I think forcing to unwrap these and not checking the proper optional types is going to cause unintended issues in the future

apps/studio/components/interfaces/Database/Functions/DeleteFunction.tsx
```diff
@@ -8,7 +8,7 @@ import TextConfirmModal from 'ui-patterns/Dialogs/TextConfirmModal'
 interface DeleteFunctionProps {
   func?: DatabaseFunction
   visible: boolean
-  setVisible: (value: boolean) => void
+  setVisible: (value: string) => void
```
awaseem: Why string? boolean with the logic made much more sense no? This can lead to confusion on the DX side

apps/studio/components/interfaces/Database/Indexes/Indexes.tsx
```diff
@@ -51,6 +49,30 @@ const Indexes = () => {
...
+
+  const { show: showCreateIndex, setShow: setShowCreateIndex } = newQueryState
+
+  const { setSelectedId: setSelectedIndexName, entity: selectedIndex } = editQueryState!
+
+  const {
+    setSelectedId: setSelectedIndexNameToDelete,
+    entity: selectedIndexToDelete,
+    show: showindexToDelete,
+  } = deleteQueryState!
```
awaseem: Same comment, force unwrapping going to cause issues in the future

apps/studio/components/interfaces/Database/Indexes/Indexes.tsx
```diff
@@ -221,14 +246,14 @@ const Indexes = () => {
...
-        visible={selectedIndex !== undefined}
+        visible={!!selectedIndex}
         header={
           <>
             <span>Index:</span>
             <code className="text-sm ml-2">{selectedIndex?.name}</code>
           </>
         }
-        onCancel={() => setSelectedIndex(undefined)}
+        onCancel={() => setSelectedIndexName('')}
```
awaseem: having this be an optional type is much better than using empty string IMO. it forces the developer to think about the situation where we need to check the type and handle cases where it might be undefined. We should witch this back IMO

apps/studio/components/interfaces/Database/Roles/RolesList.tsx
```diff
@@ -42,6 +42,28 @@ export const RolesList = () => {
...
+    transformId: (id) => id.toString(),
+  })
+
+  const { show: isCreatingRole, setShow: setIsCreatingRole } = newQueryState
+
+  const {
+    setSelectedId: setSelectedRoleIdToDelete,
+    entity: roleToDelete,
+    show: showRoleToDelete,
+  } = deleteQueryState!
```
awaseem: Same comment here, need to be careful with forcing unwraps of optionals

apps/studio/components/interfaces/Database/Roles/RolesList.tsx
```diff
@@ -212,9 +234,9 @@ export const RolesList = () => {
       <CreateRolePanel visible={isCreatingRole} onClose={() => setIsCreatingRole(false)} />
 
       <DeleteRoleModal
-        role={selectedRoleToDelete}
-        visible={selectedRoleToDelete !== undefined}
-        onClose={() => setSelectedRoleToDelete(undefined)}
+        role={roleToDelete as unknown as PostgresRole}
+        visible={showRoleToDelete}
+        onClose={() => setSelectedRoleIdToDelete('')}
```
awaseem: These are better as undefined rather than empty strings. Its better dx for users

awaseem (changes requested):
Needs some work around the typing, reach out if you have any questions or concerns

apps/studio/components/interfaces/Database/EnumeratedTypes/EnumeratedTypes.tsx
fsansalvadore: Since you pointed this out (and btw it didn't feel 100% right before, to be honest) I realized that it makes sense to use one hook per `useQueryState`. It adds some slight more config repetition but it keeps everything cleaner and less error-prone. Wdyt?

```tsx
const { show: showCreateFunctionForm, setShow: setShowCreateFunctionForm } = useQueryStateRouting(
    {
      key: 'new',
    }
  )

  const {
    setSelectedId: setSelectedFunctionToEdit,
    entity: functionToEdit,
    show: showFunctionToEdit,
  } = useQueryStateRouting({
    key: 'edit',
    entities: functions,
    isLoading,
    idField: 'id',
    entityName: 'Database Function',
    transformId: (id: number) => id.toString(),
  })

  const {
    setSelectedId: setSelectedFunctionIdToDuplicate,
    entity: functionToDuplicate,
    show: showFunctionToDuplicate,
  } = useQueryStateRouting({
    key: 'duplicate',
    entities: functions,
    isLoading,
    idField: 'id',
    entityName: 'Database Function',
    transformId: (id: number) => id.toString(),
    transformDuplicate: (fn: any) => ({
      ...fn,
      name: `${fn.name}_duplicate`,
    }),
  })

  const {
    setSelectedId: setSelectedFunctionToDelete,
    entity: functionToDelete,
    show: showFunctionToDelete,
  } = useQueryStateRouting({
    key: 'delete',
    entities: functions,
    isLoading,
    idField: 'id',
    entityName: 'Database Function',
    transformId: (id: number) => id.toString(),
  })
```

apps/studio/components/interfaces/Database/Functions/DeleteFunction.tsx
fsansalvadore: I should probably rename this prop which was pre-existing. It was a simple boolean before, but now it should actually remove the param "id" and set it to null.

apps/studio/components/interfaces/Database/EnumeratedTypes/EnumeratedTypes.tsx
fsansalvadore: Removed the support for boolean operations such as "new" as it's basically just plain nuqs...

awaseem (approved):
:chefs-kiss: great work!

PR #40190, opened 2025-11-05
Fix: Update sentry to ignore events that are causing noise

awaseem:

Yes

Adds DenyURLs to deny values

We are getting alerted of other user scripts

## What is the new behavior?

Ignore user script errors

## Additional context

N/A

### DIFF
--- a/apps/studio/instrumentation-client.ts
+++ b/apps/studio/instrumentation-client.ts
@@ -56,15 +56,7 @@ Sentry.init({
   debug: false,
   // [Ali] Filter out browser extensions and user scripts (FE-2094)
   // Using denyUrls to block known third-party script patterns
-  denyUrls: [
-    /extensions?\//i,
-    /^chrome:\/\//i,
-    /^chrome-extension:\/\//i,
-    /^moz-extension:\/\//i,
-    /^safari-extension:\/\//i,
-    /^safari-web-extension:\/\//i,
-    /userscript/i,
-  ],
+  denyUrls: [/userscript/i],
   beforeBreadcrumb(breadcrumb, _hint) {
     const cleanedBreadcrumb = { ...breadcrumb }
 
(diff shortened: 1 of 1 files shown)


apps/studio/instrumentation-client.ts
```diff
@@ -34,23 +34,37 @@ function isHCaptchaRelatedError(event: Sentry.Event): boolean {
 // and has a specific pre_context comment that we can use for filtering.
 // Copied from docs app instrumentation-client.ts
 function isThirdPartyError(frames: Sentry.StackFrame[] | undefined) {
-  if (!frames) return false
+  if (!frames || frames.length === 0) return false
 
   function isSentryFrame(frame: Sentry.StackFrame, index: number) {
     return index === 0 && frame.pre_context?.[0]?.includes('sentry.javascript')
```
charislam:
```suggestion
    return index === 0 && frame.pre_context?.some(line => line.includes('sentry.javascript'))
```

we have a bug here actually, the latest user scripting alerts would've been caught with this

apps/studio/instrumentation-client.ts
```diff
@@ -34,23 +34,37 @@ function isHCaptchaRelatedError(event: Sentry.Event): boolean {
...
 }
 
 Sentry.init({
   dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
   // Setting this option to true will print useful information to the console while you're setting up Sentry.
   debug: false,
+  // [Ali] Filter out browser extensions and user scripts (FE-2094)
+  // Using denyUrls to block known third-party script patterns
+  denyUrls: [
+    /extensions?\//i,
```
charislam: I'm worried this one is far too broad

charislam (approved):

apps/studio/instrumentation-client.ts
awaseem: Lets just do user scripts for now

apps/studio/instrumentation-client.ts
awaseem: 🐐 find!

PR #40192, opened 2025-11-05
fix(table editor): prevent unnecessary update requests

charislam:
Continues on from https://github.com/supabase/supabase/pull/40160, reducing the number of unnecessary requests made when table changes are saved.

Currently, when you open the "Edit table" side panel, make _zero changes_, and click "Save", an unnecessary request runs to:
- Alter the table's comment to its existing comment
- Alter the table's schema to its existing schema
- Alter the table's RLS enable status to its existing RLS enable status

This PR changes the payload so we only alter what has actually changed. It also adds some refactoring to define clearer types and add type safety for the three possible kinds of table save actions (create new, duplicate, update existing).

After this, there are still a few unnecessary column update actions happening on array columns; those will be fixed in a final PR.

## Testing

- [ ] Creating a table works as before
- [ ] Duplicating a table works as before
- [ ] Updating a table with a real update works as before
- [ ] Updating a table with no actual changes does not send a `table-update` request

Towards FE-2066

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
@@ -724,7 +724,6 @@ export const updateTable = async ({
   const primaryKeyColumns = columns
     .filter((column) => column.isPrimaryKey)
     .map((column) => column.name)
-
   const existingPrimaryKeyColumns = table.primary_keys.map((pk: PostgresPrimaryKey) => pk.name)
   const isPrimaryKeyUpdated = !isEqual(primaryKeyColumns, existingPrimaryKeyColumns)
 
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
@@ -394,7 +394,7 @@ export const duplicateTable = async (
     connectionString,
     sql: [
       `CREATE TABLE "${sourceTableSchema}"."${duplicatedTableName}" (LIKE "${sourceTableSchema}"."${sourceTableName}" INCLUDING ALL);`,
-      payload.comment !== undefined
+      payload.comment != undefined
         ? `comment on table "${sourceTableSchema}"."${duplicatedTableName}" is '${payload.comment}';`
         : '',
     ].join('\n'),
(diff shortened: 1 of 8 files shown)


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
```diff
@@ -394,7 +394,7 @@ export const duplicateTable = async (
     connectionString,
     sql: [
       `CREATE TABLE "${sourceTableSchema}"."${duplicatedTableName}" (LIKE "${sourceTableSchema}"."${sourceTableName}" INCLUDING ALL);`,
-      payload.comment !== undefined
+      payload.comment != undefined
```
soedirgo, line 397: Any reason why we want loose equality here?

packages/pg-meta/src/pg-meta-tables.ts
```diff
@@ -139,15 +139,15 @@ const generateEnrichedTablesSql = ({ includeColumns }: { includeColumns?: boolea
...
+  comment?: string | null
 }
 
 function create({ name, schema = 'public', comment }: TableCreateParams): { sql: string } {
   const tableSql = `CREATE TABLE ${ident(schema)}.${ident(name)} ();`
   const commentSql =
-    comment === undefined
-      ? ''
-      : `COMMENT ON TABLE ${ident(schema)}.${ident(name)} IS ${literal(comment)};`
+    comment != undefined
```
soedirgo, line 148: Loose equality?

packages/pg-meta/src/pg-meta-tables.ts
```diff
@@ -232,7 +232,7 @@ $$;
     }
   }
   const commentSql =
-    comment === undefined
+    comment == undefined
```
soedirgo, line 235: Loose equality?

packages/pg-meta/src/pg-meta-tables.ts
charislam: same here

packages/pg-meta/src/pg-meta-tables.ts
charislam: same here

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
charislam: Yep, because if it's null we also don't want to send an update. The type is (unfortunately) `string | undefined | null`, where `undefined` and `null` both represent no change, and `''` represents changing the comment to remove it

soedirgo: Cool, just wanted to check if it's intentional 👍

soedirgo (approved):

alaister (approved):
LGTM!

- Tested as described on preview

kemaldotearth (approved):
LGTM, ran through tests :shipit:

PR #40170, opened 2025-11-05
feat(studio): clearer table row actions

dnywh:
## What kind of change does this PR introduce?

- Interface update

## What is the current behavior?

- Interactive elements within a TableRow → TableCell often have their action hidden

## What is the new behavior?

- Interactive elements within a TableRow → TableCell have a suggestive underline

| Before and After |
| --- |
|  before |
| after |

## Additional context

- This is achieved via a new `.text-link-table-cell` typography style class
- Prior commits used a custom component, but this quickly becomes complex as various scenarios must support
  - Buttons, not just links
  - Some lighter, some darker
  - Various class name additions depending on the situations

Still to come:

- [ ] Design system documentation

PR #40193, opened 2025-11-05
docs(design-system): interactive rows guidance

dnywh:

Documentation

We don’t have any guidance around row interactivity. There are a few complex cases:

- Interactive elements within a TableRow
- Interactive elements with CardContent

And in both of those cases:
- When there is one primary action and 1+ secondary actions
- When there are equal-weight actions(?)
- How `stopPropagation` works

## What is the new behavior?

Guidance and demo components for all of the above.

## Still to come

- [ ] Proper inline demo components (as text explanations aren’t enough—show don’t tell)
  - [ ] Real-world examples (e.g. Security email templates)
- [ ] Guidance around `stopPropagation`, including affordances for misclicks

## Prior art and references
- [Discussion](https://supabase.slack.com/archives/C0429V78ACX/p1762380146001729?thread_ts=1762322376.162839&cid=C0429V78ACX)
- https://github.com/supabase/supabase/pull/40170
- Vercel deployments
- Linear issues

### DIFF
--- a/apps/design-system/content/docs/ui-patterns/navigation.mdx
+++ b/apps/design-system/content/docs/ui-patterns/navigation.mdx
@@ -5,6 +5,55 @@ description: Navigation patterns help users understand where they are and where 
 
 Supabase has a necessarily complex navigation system to handle multiple products and levels of hierarchy. This page introduces those general patterns, best practices, and the components involved.
 
+## Row interaction
+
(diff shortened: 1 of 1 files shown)


dnywh:
Superseded by later design docs and a shared function as described in https://github.com/supabase/supabase/pull/41628. Closing.

PR #40196, opened 2025-11-06
docs(design-system): design system improvements

dnywh:

- Design system documentation

- We have some extraneous pages
- Our empty state documentation misses some important info
  - DataGrid
  - More detail around presentation

## What is the new behavior?

- Removed old pages
- Expanded [empty state documentation and example components](https://design-system-git-dnywh-docsdesign-system-improvements-supabase.vercel.app/design-system/docs/ui-patterns/empty-states)
  - Brings in DataGrid, which in turn requires `react-data-grid`
- Misc improvements
  - Font smoothing to match elsewhere

## Additional context

I wonder if we should wrap the (headless?) `react-data-grid` in our own component so we don’t need to do these sorts of imports. This can be a later improvement.


apps/design-system/config/docs.ts
```diff
@@ -40,11 +40,6 @@ export const docsConfig: DocsConfig = {
           href: '/docs/figma',
           items: [],
         },
-        {
-          title: 'Changelog',
-          href: '/docs/changelog',
-          items: [],
-        },
```
dnywh, line 47: We decided as a team to remove this page

apps/design-system/styles/globals.css
```diff
@@ -37,6 +37,8 @@
     /* font-feature-settings: "rlig" 1, "calt" 1; */
     font-synthesis-weight: none;
     text-rendering: optimizeLegibility;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
```
dnywh, line 41: Matches text rendering on Studio and elsewhere.

apps/design-system/package.json
```diff
@@ -29,6 +29,7 @@
     "next-contentlayer2": "0.4.6",
     "next-themes": "^0.3.0",
     "react": "catalog:",
+    "react-data-grid": "7.0.0-beta.41",
```
dnywh, line 32: Needed for `DataGrid`

apps/design-system/registry/default/example/empty-state-zero-items-data-grid.tsx
```diff
@@ -0,0 +1,47 @@
+import { Users } from 'lucide-react'
+import DataGrid, { Column } from 'react-data-grid'
+import 'react-data-grid/lib/styles.css'
```
dnywh, line 3: Matching how we import `DataGrid` from Studio, where we also import the styles.

We should probably wrap `react-data-grid` in our own UI component eventually, right?

apps/design-system/registry/default/example/empty-state-missing-route.tsx
```diff
@@ -14,7 +14,14 @@ export default function EmptyStateMissingRoute() {
         description={`${bucketId ? `The bucket “${bucketId}”` : 'This bucket'} doesn’t seem to exist.`}
       >
         <Button asChild type="default" className="mt-2">
-          <Link href="/">Head back</Link>
+          <Link
+            href="/"
+            onClick={(e) => {
+              e.preventDefault()
```
dnywh, line 20: Just stopping this button from actually doing anything.

apps/design-system/config/docs.ts
dnywh: And the Figma one

pnpm-lock.yaml
dnywh, line 1: Necessary changes for `react-data-grid` in design-system.

SaxonF (approved):
Reviewed empty states pattern in design-system staging

PR #39911, opened 2025-10-27
Update custom-claims-and-role-based-access-control-rbac.mdx

EliaTolin:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

This pull request makes a small change to the SQL policy definition in the documentation by adding a missing semicolon at the end of the `using (true)` statement. This helps ensure the SQL example is syntactically correct.

## What is the current behavior?

This pull request makes a minor change to the SQL policy definition in the documentation by adding a missing semicolon at the end of the `using (true)` statement. This ensures the SQL syntax is correct.

## What is the new behavior?

This pull request makes a minor update to the SQL policy definition for reading user roles, adding a missing semicolon at the end of the statement for proper SQL syntax.

## Summary by CodeRabbit

* **Documentation**
  * Fixed SQL syntax in guide documentation to ensure database statements are properly formatted and valid.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #40197, opened 2025-11-06
Fix jsonb_set usage for claims and roles

luismtapiab:
These calls need a explicit casting otherwise they cause errors when login. I also took the opportunity to add a missing semicolon in the same snippet.

YES

docs update

Login fails with this response: 
{
  "code":"unexpected_failure",
  "message":"Error running hook URI: pg-functions://postgres/public/custom_access_token_hook"
}

The errors in the logs like this: 
ERROR: function json_set(jsonb, unknown, jsonb) does not exist (SQLSTATE 42883)

## What is the new behavior?
No errors from the hook. Login or not but not because of this.

## Additional context

The missing semicolon was mentioned in #39911 so its a good chance to fix that too. 
Everyday I found an instance where PL/pgSQL linter would reduce so much debugging, looking forward to its release.

### DIFF
--- a/apps/docs/content/guides/database/postgres/custom-claims-and-role-based-access-control-rbac.mdx
+++ b/apps/docs/content/guides/database/postgres/custom-claims-and-role-based-access-control-rbac.mdx
@@ -100,13 +100,13 @@ as $$
 
     if user_role is not null then
       -- Set the claim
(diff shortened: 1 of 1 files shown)


Hallidayo:

PR #40217, opened 2025-11-06
Joshen/depr 170 analytics buckets hook up connect tables functionality

joshenlim:
## Context

Changes here aim to hook up the "Connect table" flow for Analytics Buckets. Some notes up front though
- There's a number of potential BE issues that I stumbled upon while working on this, so it's still pretty iffy end to end
- As such, I don't require a stringent smoke test on the staging preview but just a code review will do
- Priority with the review is to ensure no changes for public users (In this case the "Connect tables" CTA should remain disabled with "Coming soon" tooltip)
image

Just to call out one thing - FE is doing a lot of orchestration here which is understandably a bad practice (and hence why the code here looks pretty complex), we'll aim to shift a lot of the logic to the BE eventually once we're ready and hopefully we can thereafter simplify the code here.

## Priority testing

- [ ] Verify that changes from this PR have no affect on what's on production at the moment
  - "Connect tables" CTA should remain disabled with "Coming soon" tooltip
  - This was previously hardcoded to be disabled, but I'm changing that to rely on the `storageAnalyticsVector` config cat feature flag instead

## Secondary testing

As mentioned, there's some issues I ran into with the BE so I don't expect this to work fully end to end without any issues atm. But here's some basic testing that you can do still

- When clicking "Connect tables"
  - [ ] If replication has not been enabled for the project, but the project does not have access to replication, should see this callout
  Screenshot 2025-11-07 at 02 11 34

  - [ ] If replication has not been enabled for the project, and the project does have access to replication, should see this callout
  Screenshot 2025-11-07 at 02 11 14

- [ ] If replication has already been enabled for the project, should see the table selector as such (it'll only be listing tables from the public schema here)
  image

- [ ] If you try to connect a single table, dashboard will thereafter long poll for namespaces and the iceberg wrapper, which once the data is ready, you should see this
  - This is where I ran into problems locally, it was working initially but somehow the namespaces and wrapper wasn't getting updated anymore
  - I have yet to try selecting multiple tables too
  image

- [ ] You should be able to stop replication for a table (and subsequently start it again)
  image

- You should technically be able to add more tables to the analytics bucket, but there's where I ran into problems too - the iceberg wrapper wasn't getting updated on the BE

- [ ] Just double check that all the other actions are working
  - Can view table in table editor
  - Can query table in SQL editor
  - Can view replication for the bucket


apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
```diff
@@ -78,52 +152,48 @@ export const ConnectTablesDialog = ({ bucketId }: { bucketId?: string }) => {
...
     if (!projectRef) return console.error('Project ref is required')
-    if (!sourceId) return toast.error('Source ID is required')
     if (!bucketId) return toast.error('Bucket ID is required')
+    if (!sourceId) return toast.error('Replication has not been enabled for your project')
 
     try {
+      setIsConnecting(true)
+
+      // Step 1: Create publication
+      setConnectingStep(1)
```
awaseem: Can we remove the magic number? would rather use constant types or maybe enum (strong MAYBE). Make it easier to read as well. thoughts?

awaseem: Same thing in this whole file

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
```diff
@@ -172,104 +228,242 @@ export const ConnectTablesDialog = ({ bucketId }: { bucketId?: string }) => {
...
-        >
-          Connect tables
-        </ButtonTooltip>
-      </DialogTrigger>
+  const onSubmit: SubmitHandler<ConnectTablesForm> = async (values) => {
+    if (isEditingExistingPublication) {
+      onSubmitUpdatePublication(values)
+    } else {
+      onSubmitNewPublication(values)
+    }
```
awaseem, line 306: nit: return early? Thoughts about this?

```suggestion
    if (isEditingExistingPublication) {
      onSubmitUpdatePublication(values)
      return
    }  
    onSubmitNewPublication(values)
```

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
```diff
@@ -172,104 +228,242 @@ export const ConnectTablesDialog = ({ bucketId }: { bucketId?: string }) => {
...
+  }
 
-      <DialogContent>
-        <DialogHeader>
-          <DialogTitle>Connect tables</DialogTitle>
-        </DialogHeader>
+  useEffect(() => {
+    form.reset()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [visible])
```
awaseem, line 312: Could this be a good use case of useLatest that @charislam has been working on?

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
```diff
@@ -172,104 +228,242 @@ export const ConnectTablesDialog = ({ bucketId }: { bucketId?: string }) => {
...
+  const { error } = useReplicationSourcesQuery({ projectRef })
+  const noAccessToReplication =
+    !enablePgReplicate || error?.message.includes('feature flag is required')
+
+  const { mutateAsync: createTenantSource, isLoading: creatingTenantSource } =
+    useCreateTenantSourceMutation()
+
+  const onEnableReplication = async () => {
+    if (!projectRef) return console.error('Project ref is required')
+    await createTenantSource({ projectRef })
```
awaseem, line 428: Do we care to even log this if its undefined. What we change it to the following:
```suggestion
    if (projectRef) { 
        await createTenantSource({ projectRef })
    }
```

apps/studio/data/replication/pipelines-query.ts
```diff
@@ -37,5 +35,23 @@ export const useReplicationPipelinesQuery = <TData = ReplicationPipelinesData>(
...
+      ) {
+        return false
+      }
+
+      if (failureCount < 3) {
+        return true
+      }
+
+      return false
+    },
```
awaseem: I know this isn't your change, but should we refactor this right now to a simple function that can be used in other queries. I see this on the client side as well. Sorry for the extra work 😓

apps/studio/data/replication/sources-query.ts
```diff
@@ -36,5 +36,23 @@ export const useReplicationSourcesQuery = <TData = ReplicationSourcesData>(
...
+      ) {
+        return false
+      }
+
+      if (failureCount < 3) {
+        return true
+      }
+
+      return false
+    },
```
awaseem: Same comment here

apps/studio/data/storage/iceberg-namespace-create-mutation.ts
```diff
@@ -90,7 +90,7 @@ export const useIcebergNamespaceCreateMutation = ({
       await onSuccess?.(data, variables, context)
     },
     async onError(data, variables, context) {
-      if ((data.message = 'Request failed with status code 409')) {
+      if (data.message === 'Request failed with status code 409') {
```
awaseem, line 93: wow this is a good catch, this was doing assignment 😅

awaseem (changes requested):
Let me know your thoughts on the comments

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
joshenlim: hmm if else is more readable in this case imo

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
joshenlim: hmm how would we use useLatest in here though? (i might not have context here)

mainly reseting the form values whenever the modal is opened

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
joshenlim: no strong opinion for this one but for context `projectRef` should never be missing in this case given that its coming from the params

the intention here was that if for some reason it does turn up missing somehow - ideally we don't swallow that error too

it's also just to be consistent with how we've been adding similar guards like this
image

apps/studio/data/storage/iceberg-namespace-create-mutation.ts
joshenlim: hahaha yeah i was just looking at this and went "wait a minute" like the kazoo kid meme

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
awaseem: Thats fair!

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
awaseem: I think that PR isn't merged or done yet, so this is fine right now

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
awaseem: Oh thats a good point, side question what would even cause this to be nil? That would be the case in self hosted?

apps/studio/data/replication/pipelines-query.ts
```diff
@@ -37,5 +36,19 @@ export const useReplicationPipelinesQuery = <TData = ReplicationPipelinesData>(
...
+      if (isFeatureFlagRequiredError) {
+        return false
+      }
+
+      if (failureCount < 3) {
+        return true
+      }
+
+      return false
+    },
```
awaseem: Is this not the same function as `source-query.ts`? Can we reuse it?

awaseem (approved):
LGTM, minor reuse comments

apps/studio/data/replication/pipelines-query.ts
joshenlim: ahh i see what you mean! sorry brain fart 🤦

joshenlim: ahh i see what you mean okay gotcha on this! sorry brain fart 🤦

alaister (approved):
LGTM!

PR #40220, opened 2025-11-06
docs(auth hooks): clarify hook name must match available hooks

tk1ng:

YES

Docs update

Auth hooks documentation doesn't explicitly state that the hook name configured in `config.toml` must correspond to available hooks to prevent errors, which has caused confusion for some users.

## What is the new behavior?

Doc includes mention that hook name must correspond to available hooks.

## Summary by CodeRabbit

* **Documentation**
  * Added clarification note regarding authentication hook configuration requirements, specifying that hook names in configuration files must correspond to available hook types.


silentworks:
You might want to add to this PR the correct hook names too. I was going to do that in my PR but left it out. The docs uses hyphens for hook names while they need to be underscore.

tk1ng:
> You might want to add to this PR the correct hook names too. I was going to do that in my PR but left it out. The docs uses hyphens for hook names while they need to be underscore.

Good call out. Moved the note into an admonition component and added an example for more clarity.

CleanShot 2025-11-06 at 14 11 58

ChrisChinchilla (approved):

PR #40223, opened 2025-11-06
Fix: Double words for errors within Sentry for capture message

awaseem:
I messed up the messages for the logs, so it was doing this double formatting (whoops!) this should fix it and standardize it!

### DIFF
--- a/apps/studio/lib/error-reporting.ts
+++ b/apps/studio/lib/error-reporting.ts
@@ -67,7 +67,7 @@ function handleResponseError(error: ResponseError, context: string) {
   const { code, message, requestPathname } = error
   if (!requestPathname || !code) {
     captureMessage({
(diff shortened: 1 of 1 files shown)


apps/studio/lib/error-reporting.ts
```diff
@@ -77,31 +77,33 @@ function handleResponseError(error: ResponseError, context: string) {
...
+  if (
+    error &&
+    typeof error === 'object' &&
+    'message' in error &&
+    typeof error.message === 'string'
+  ) {
     captureMessage({
-      message: `[CRITICAL][${context}] Failed: ${error.message}`,
+      message: error.message,
       context,
```
awaseem, line 107: Normally I hate this, but Typescript doesn't safely determine error.message is a string if its assigned :(

charislam (approved):

PR #40229, opened 2025-11-07
Clean up RQ retries

joshenlim:
## Context

Just realised that the global `query-client` already handles retries logic to stop retrying if error is `>=400` and `< 500`, and if failure count is >= 3. We had some duplicated logic in the individual react queries that this PR cleans up


alaister (approved):
LGTM!

- Tested on preview

PR #40230, opened 2025-11-07
studio(chore): improve storage bucket presentation

dnywh:

- UI updates

The root page for file, analytics, and vector storage types have some issues:

- Users would like to tap the whole row to enter a bucket
- Secondary actions (e.g. delete bucket) should live inside the bucket, not at its parent
- “Alpha” status is scary and loud

## What is the new behavior?

- The above issues are fixed or improved
- Minor general improvements elsewhere for storage

| Before | After |
| --- | --- |

## Additional context

- I’d like to test virtualised tables, but suppose that can only happen on “MOCKMASTER” staging project after merge. Please let me know if that’s not the case, or if there is a better way.
- The zero results (searched) state for buckets removes the Bucket icon and its TableCell. This is a known layout shift. I felt like this was better than a blank space, but happy to be convinced otherwise.
- Known copywriting inconsistency now in deletion (and similar) dialogs across bucket types. I’ll get to these.


apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -44,26 +30,27 @@ export const AnalyticsBuckets = () => {
...
           </Button>
         }
       >
-        <p className="!leading-normal !mb-0">
-          Expect rapid changes, limited features, and possible breaking updates as we expand access.
+        <p className="!leading-normal !mb-0 text-balance">
+          Analytics buckets are now in private alpha. Expect rapid changes, limited features, and
+          possible breaking updates. Please share feedback as we refine the experience and expand
+          access.
         </p>
```
dnywh, line 65: We discussed rephrasing the “Alpha” flag and its admonition to be more welcoming than scary. Similar change for vector buckets (and the Badge for each)

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -94,9 +81,16 @@ export const AnalyticsBuckets = () => {
               <Table>
                 <TableHeader>
                   <TableRow>
+                    {analyticsBuckets.length > 0 && (
+                      <TableHead className="w-2 pr-1">
+                        <span className="sr-only">Icon</span>
+                      </TableHead>
+                    )}
```
dnywh, line 100: Show a little bucket icon against each row (unless empty state).

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -111,15 +105,20 @@ export const AnalyticsBuckets = () => {
...
                   {analyticsBuckets.map((bucket) => (
-                    <TableRow key={bucket.id}>
+                    <TableRow
+                      key={bucket.id}
+                      className="cursor-pointer h-16"
+                      onClick={(event) => {
+                        const url = `/project/${ref}/storage/analytics/buckets/${encodeURIComponent(bucket.id)}`
+                        if (event.metaKey) window.open(url, '_blank')
+                        else router.push(url)
+                      }}
```
dnywh: We discussed making the entire row tappable.

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -132,31 +131,8 @@ export const AnalyticsBuckets = () => {
...
-                                  setSelectedBucket(bucket)
-                                }}
-                              >
-                                <Trash2 size={12} />
-                                <p>Delete bucket</p>
-                              </DropdownMenuItem>
-                            </DropdownMenuContent>
-                          </DropdownMenu>
+                        <div className="flex justify-end items-center h-full">
+                          <ChevronRight size={14} className="text-foreground-muted/60" />
```
dnywh, line 145: A little affordance to show that the row goes somewhere, rather than just being read-only data. Can remove if we don’t like it.

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -167,14 +143,6 @@ export const AnalyticsBuckets = () => {
...
         </div>
       )}
-
-      {selectedBucket && (
-        <DeleteAnalyticsBucketModal
-          visible={modal === 'delete'}
-          bucketId={selectedBucket.id}
-          onClose={() => setModal(null)}
-        />
-      )}
```
dnywh, line 177: Now happens within bucket, not outside.

apps/studio/components/interfaces/Storage/FilesBuckets/BucketsTable.tsx
```diff
@@ -37,7 +33,7 @@ const BucketsTableUnvirtualized = ({
     <Table
       containerProps={{ containerClassName: 'h-full overflow-auto', className: 'overflow-visible' }}
     >
-      <BucketTableHeader mode="standard" />
+      <BucketTableHeader mode="standard" hasBuckets={buckets.length > 0} />
```
dnywh, line 36: Prop needed to conditionally show cell to match bucket icon, depending on whether the search results are zero or 1+.

apps/studio/components/interfaces/Storage/DeleteBucketModal.tsx
```diff
@@ -117,7 +117,7 @@ export const DeleteBucketModal = ({ visible, bucket, onClose }: DeleteBucketModa
     >
       <DialogContent aria-describedby={undefined}>
         <DialogHeader>
-          <DialogTitle>Confirm deletion of {bucket.id}</DialogTitle>
```
dnywh: Dialog titles should be succinct, hence these little improvements. I’d even be inclined _not_ show the bucket name here, as they have variable length.

All of this can be shown in the dialog body. But that’s a later improvement, I suppose.

apps/studio/components/layouts/PageLayout/PageHeader.tsx
dnywh, line 1: General component updates because it did not handle long titles well. The `secondaryActions` would be pushed off screen if the `title` was too long.

apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx
```diff
@@ -88,11 +130,25 @@ const BucketPage: NextPageWithLayout = () => {
...
+            bucket={bucket}
+            onClose={() => setModal(null)}
+          />
+          <DeleteBucketModal
+            visible={modal === 'delete'}
+            bucket={bucket}
+            onClose={() => setModal(null)}
+          />
+        </>
+      )}
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx:151]: These were removed from the root Files page so now added here, within the bucket page itself (via a dropdown button).

packages/ui/src/components/shadcn/ui/badge.tsx
```diff
@@ -3,34 +3,32 @@ import * as React from 'react'
...
+      brand: 'bg-brand bg-opacity-10 text-brand-600 border border-brand-500',
+      secondary:
+        'bg-secondary bg-opacity-10 hover:bg-secondary/80 border-transparent text-secondary-foreground',
+      outline: 'bg-transparent text border border-foreground-muted',
     },
-    defaultVariants: {
-      variant: 'default',
-      size: 'small',
+    size: {
+      small: 'px-2 py-0.5 text-xs',
```
dnywh, line 20: This is the only visual change: `small` has a `px-2` instead of `px-2.5`. Everything else here is a style cleanup. A lot of duplicity.

packages/ui/src/components/shadcn/ui/badge.tsx
```diff
@@ -3,34 +3,32 @@ import * as React from 'react'
...
-        small: 'px-2.5 py-0.5 text-xs',
-        large: 'px-3 py-0.5 rounded-full text-sm',
-      },
-      dot: {
-        true: '-ml-0.5 mr-1.5 h-2 w-2 rounded-full',
-      },
+const badgeVariants = cva('inline-flex items-center rounded-full font-normal', {
+  variants: {
+    variant: {
+      default: 'bg-surface-75 text-foreground-light border border-strong',
```
dnywh, line 9: Oh, and `bg-surface-75` instead of `bg-surface-200`, as the former stands out more. Folks should use `outline` if they want a subtle (zero) background.

packages/ui/src/components/shadcn/ui/dialog.tsx
```diff
@@ -162,6 +162,8 @@ const DialogTitle = React.forwardRef<
   <DialogPrimitive.Title
     ref={ref}
     className={cn('text-base leading-none font-normal', className)}
+    // [Danny] Make space for the close button
+    style={{ maxWidth: 'calc(100% - 1rem)' }}
```
dnywh: As it says in the comment: the `X` that every dialog should have means text will be visually clipped when long. This prevents that. I don’t believe there is a straightforward way to do `calc` in Tailwind.

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -111,15 +105,20 @@ export const AnalyticsBuckets = () => {
                     </TableRow>
                   )}
                   {analyticsBuckets.map((bucket) => (
-                    <TableRow key={bucket.id}>
+                    <TableRow
+                      key={bucket.id}
+                      className="cursor-pointer h-16"
```
dnywh: The default height of these rows felt a bit puny for now-tappable elements. Is `h-16` the best way to do this? Padding wouldn’t cooperate.

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -111,15 +105,20 @@ export const AnalyticsBuckets = () => {
...
+                      onClick={(event) => {
+                        const url = `/project/${ref}/storage/analytics/buckets/${encodeURIComponent(bucket.id)}`
+                        if (event.metaKey) window.open(url, '_blank')
+                        else router.push(url)
+                      }}
+                    >
+                      <TableCell className="w-2 pr-1">
+                        <BucketIcon size={16} className="text-foreground-muted" />
```
charislam, line 122: You're the expert here :D but IMO the bucket icon looks like a cross between a trash icon and a lock icon, it just doesn't make me think "that's a bucket!"

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -111,15 +105,20 @@ export const AnalyticsBuckets = () => {
                     </TableRow>
                   )}
                   {analyticsBuckets.map((bucket) => (
-                    <TableRow key={bucket.id}>
+                    <TableRow
```
charislam: This needs to be keyboard-focusable because now there's no way to open this without a mouse

apps/studio/components/interfaces/Storage/FilesBuckets/BucketTable.tsx
```diff
@@ -89,23 +80,28 @@ export const BucketTableRow = ({
...
-      <BucketTableCell>
-        <div className="flex items-center gap-2">
-          <Link
-            href={`/project/${projectRef}/storage/files/buckets/${encodeURIComponent(bucket.id)}`}
-            title={bucket.id}
-            className="text-link-table-cell"
-          >
-            {bucket.id}
-          </Link>
+    <BucketTableRow
```
charislam: same here, needs to be keyboard-focusable and work on Enter

apps/studio/components/interfaces/Storage/VectorBuckets/index.tsx
```diff
@@ -126,15 +118,20 @@ export const VectorsBuckets = () => {
                     const created = +bucket.creationTime * 1000
 
                     return (
-                      <TableRow key={id}>
+                      <TableRow
```
charislam: same thing again here

charislam (changes requested):

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
dnywh: Fair point! I will tinker and push updates in a separate PR.

apps/studio/components/interfaces/Storage/FilesBuckets/BucketTable.tsx
dnywh: Good call, thanks @charislam. Made those changes across files, analytics, and vector bucket rows.

- `role="button"` and `tabIndex={0}` for keyboard focus
-  `onKeyDown` handler for Enter key
- basic `focus-visible` styles
- `(event.metaKey || event.ctrlKey)` for new tab

I made a `handleBucketNavigation` in each to extract some shared logic. Let me know what you think!

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -107,12 +107,22 @@ export const AnalyticsBuckets = () => {
                   {analyticsBuckets.map((bucket) => (
                     <TableRow
                       key={bucket.id}
-                      className="cursor-pointer h-16"
+                      role="button"
```
charislam: this removes table semantics on the table row, and screen readers will skip the table body because it has no rows

charislam (approved):
Pushed a commit to fix the a11y -- needed to take a different approach to make this work for screen readers and keyboards. There's an outline clipping issue on the focus state, do you know a workaround for that?

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
dnywh: Good catch. Thanks

dnywh:
> There's an outline clipping issue on the focus state, do you know a workaround for that?

That was bugging me too. I just pushed an inset version.

dnywh:
Thanks for your help @charislam! Some changes since you last reviewed:

- Inset `focus-visible` to avoid clipping
- Increased prominence of `outline` (and removed background as a result)
- Made a shared style for the focus utility classes to reduce repetition and build upon elsewhere
- Jotted down first accessibility guidelines based on this work

Nothing else needed for now, unless you feel like reviewing the accessibility doc. Feel free to push changes directly to this branch, as I’ll likely ask for your input anyway :)

Just waiting on review from @supabase/design and then I’ll merge.

packages/ui/src/components/shadcn/ui/dialog.tsx
SaxonF: Can do `max-w-[calc(100%-1rem)]`

apps/studio/styles/focus.scss
```diff
@@ -0,0 +1,4 @@
+/* Focus styles for keyboard navigation on interactive table rows */
+.table-row-focus {
```
SaxonF: Should we name this something more generically so that it can be applied to any kind of tappable element that isn't a `Button`?

SaxonF: Should we name this something more generically so that it can be applied to any kind of tappable element that isn't a `Button`?

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -132,31 +141,8 @@ export const AnalyticsBuckets = () => {
...
                       <TableCell>
-                        <div className="flex justify-end gap-2">
-                          <Button asChild type="default">
-                            <Link
-                              href={`/project/${ref}/storage/analytics/buckets/${encodeURIComponent(bucket.id)}`}
-                            >
-                              View contents
-                            </Link>
-                          </Button>
-                          <DropdownMenu>
```
SaxonF, line 143: Why not offer bucket actions on both list item and within the detail page?

SaxonF (approved):
Approving to unblock after checking design-system on staging. I did not check Studio. Couple nit suggestions.

packages/ui/src/components/shadcn/ui/dialog.tsx
dnywh: Good shout. Fixing

apps/studio/styles/focus.scss
dnywh: Discussed elsewhere. Changed to `inset-focus` for now since it has opinionated, inset, styles

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
dnywh: @fenos pointed out that:

1. Only files buckets have additional actions (aside from enter and delete bucket).
2. These actions make more sense being available only inside the bucket.

I’m personally impartial. What do you think, should they be available outside? 

If we did bring them back (and assuming the whole row remains tappable), I’d probably look at a pattern like Vercel:

CleanShot 2025-11-11 at 14 05 59@2x

Though it would mean dropping the chevron.

PR #40232, opened 2025-11-07
default filters and state fix

SaxonF:
Adjusts default filters in advisor to show critical and warnings for now until can come up with a better pattern for revealing warnings particularly for "messages".

This also fixes an issue where if you traversed up to a level like account/me, then back down to project, it would run into an issue where sidebar was not being registered correctly / state was outdated. This meant you could not open up the sidebar.


alaister (approved):
LGTM!

- Tested locally

PR #40242, opened 2025-11-07
chore: audit log entitlement

ignaciodob:
This PR adds entitlements checks for the audit log access control and removes the hardcoded checks.

Changes:
- Use the entitlement hook to check for access
- Updated the number of days allowed to reflect the proper values

Testing:
- Head to `org/_/audit` with an Org on the Free Plan and an Org in the Pro Plan
- Assert that you see the following message:
image

- Head to `org/_/audit` with an Org on the Team Plan
- Assert than when picking a timeframe longer than 62 days you see this error message:
image


ecktoteckto (approved):
Checked on Preview

PR #40206, opened 2025-11-06
feat: Create FDW for S3 Vectors buckets

ivasilov:
> [!NOTE]
> Adds AWS S3 Vectors wrapper with docs/nav, and updates dashboard to auto-provision/manage FDW/schema for vector buckets and tables.
> 
> - **Wrappers/Integrations**:
>   - Add `s3_vectors_wrapper` (`S3_VECTORS` handler) with server options, metadata, and docs link.
>   - Create helper hooks `useS3VectorsWrapperExtension` and `useS3VectorsWrapperInstance`.
>   - Improve wrapper creation: optional `source_schema` validation and mode mapping (`schema`/`skip`).
>   - Replace integration overview admonition with Shadcn `Alert`.
> - **Storage (Vector Buckets)**:
>   - Bucket creation: auto-create S3 Vectors FDW (with generated S3 keys) and matching schema when Wrappers extension is available.
>   - Table creation: create index and import foreign schema with `bucket_name` option; success/error toasts.
>   - Deletion: drop foreign tables and FDW on table/bucket delete.
>   - New utils: `getVectorURI`, `getVectorBucketFDWName/FDWSchemaName/S3KeyName`.
>   - Manage UI states for missing/not-installed/needs-upgrade Wrappers; inline setup flows.
> - **FDW SQL/Mutations**:
>   - `fdw-create`/`fdw-import-foreign-schema` support `schemaOptions` and add `strict 'true'` option.
>   - Add `fdw-drop-foreign-table` mutation; relax `fdw-delete` to accept `{ name }`.
> - **Docs**:
>   - Add S3 Vectors to wrappers guides (`s3vectors.md` mapping) and navigation.
>   - New integration overview content for S3 Vectors.
> - **UI**:
>   - `Admonition` supports `childProps`; `FormSectionLabel` uses `cn`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9dd61790c2652d0a569858e9dadc57584b4e8b9d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #40244, opened 2025-11-07
Set up removing tables from namespaces in analytics buckets

joshenlim:
## Context

Add support for removing connected tables from an analytics bucket.

image

Only buckets that aren't replicating can be removed - will need to stop replication first before being allowed to remove

image

Removing a table involves
- Updating the FDW to remove the table
- Deleting the table from the iceberg namespace

Have verified locally the following
- Able to remove a table
- Able to add that table back

### DIFF
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
@@ -1,13 +1,21 @@
-import { snakeCase } from 'lodash'
-import { MoreVertical, Pause, Play } from 'lucide-react'
+import { snakeCase, uniq } from 'lodash'
+import { MoreVertical, Pause, Play, Trash } from 'lucide-react'
 import Link from 'next/link'
 import { useState } from 'react'
 import { toast } from 'sonner'
 
 import { useParams } from 'common'
+import {
+  convertKVStringArrayToJson,
+  formatWrapperTables,
+} from 'components/interfaces/Integrations/Wrappers/Wrappers.utils'
+import { getDecryptedParameters } from 'components/interfaces/Storage/ImportForeignSchemaDialog.utils'
+import { DropdownMenuItemTooltip } from 'components/ui/DropdownMenuItemTooltip'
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
```diff
@@ -117,6 +139,63 @@ export const TableRowComponent = ({
     }
   }
 
+  const onConfirmRemoveTable = async () => {
+    if (!wrapperInstance || !wrapperMeta) return toast.error('Unable to find wrapper')
+
+    try {
+      setIsRemovingTable(true)
+
+      const serverName = `${snakeCase(bucketId)}_fdw_server`
```
ivasilov: You should use `getAnalyticsBucketFDWName_server` here.

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
```diff
@@ -117,6 +139,63 @@ export const TableRowComponent = ({
...
+        ...serverOptions,
+      }
+      const targetSchemas = (formValues['supabase_target_schema'] || '')
+        .split(',')
+        .map((s) => s.trim())
+      const wrapperTables = formatWrapperTables(wrapperInstance, wrapperMeta).filter(
+        (x) => x.table_name !== table.name
+      )
+
+      await updateFDW({
```
ivasilov, line 170: If you're doing all this just to delete a table, you can use this from my PR https://github.com/supabase/supabase/pull/40206/files#diff-d65d8f2c348d632d033616d2e15c0bed1a32ed4416c21892d02ad98511475a10

ivasilov (approved):

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
joshenlim: ooooooh perfect! i'll leave a comment in the code 😄 🙏

PR #40251, opened 2025-11-07
[Documentation] Fix owner_id type

Revadike:
If you don't type cast you'll encounter the following error:
```
Error: ERROR: 42883: operator does not exist: text = uuid HINT: No operator matches the given name and argument types. You might need to add explicit type casts. 
```
This should fix it.


ChrisChinchilla (approved):
Thanks @Revadike !

PR #40259, opened 2025-11-07
Fix: Make the FK editor faster when loading tables 

awaseem:

YES

Bug fix, FK editor is slow on large tables. This should fix loading on selection rather than loading all tables and then the columns. Also created a new Loader

Massive tables hang and timeout

## What is the new behavior?

Tables columns load on selection 

## Additional context

How to test (self hosted and preview)
- [ ] Create FK from the column editor (right click column)
- [ ] Test FK creation on new table
- [ ] Ensure FK show up correctly on the table editor


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnForeignKey.tsx
```diff
@@ -47,7 +47,7 @@ const ColumnForeignKey = ({
     return {
       id: c.id,
       name: c.name,
-      format: column.format || c.format,
+      format: c.format || column.format,
```
awaseem, line 50: Fixes this bug 
Screenshot 2025-11-07 at 3 28 19 PM

e2e/studio/features/table-editor.spec.ts
```diff
@@ -231,7 +231,7 @@ test.describe.serial('table editor', () => {
       page,
       'pg-meta',
       ref,
-      'tables?include_columns=true&included_schemas=public'
+      'tables?include_columns=false&included_schemas=public'
```
awaseem, line 234: This test was waiting for the FK query to run on the column. Its strange because columns are fetched in other ways. Fixed the request param

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ForeignKeySelector/ForeignKeySelector.tsx
```diff
@@ -284,42 +297,87 @@ export const ForeignKeySelector = ({
...
+                    </div>
+                    {fk.columns.length === 0 && (
+                      <Alert_Shadcn_ className="col-span-10 py-2 px-3">
+                        <AlertDescription_Shadcn_>
+                          There are no foreign key relations between the tables
+                        </AlertDescription_Shadcn_>
+                      </Alert_Shadcn_>
+                    )}
+                    {fk.columns.map((_, idx) => (
+                      <Fragment key={`${uuidv4()}`}>
```
charislam: Won't this unmount and remount every fragment on every render? I think keying on `${fk.schema}-${fk.table}-${idx}` should be fine here, as far as I can tell there's no way to rearrange the list

charislam (approved):
Tested on self-hosted and hosted previews

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ForeignKeySelector/ForeignKeySelector.tsx
awaseem: omg yes! Good catch 🐐

PR #40265, opened 2025-11-08
fix(studio): redirect 'Create Account' button to sign-up page instead…

naamyuvraj:
### Problem
The "Create Account" button on the Organization Invite page redirects users to the sign-in page, even if they don't have an account.

### Solution
Created a `signupRedirectLink` and updated the button to use it, so users are now redirected to the sign-up page.

### Testing
- Invited a new user without a Supabase account.
- Clicked "Create Account" in the invite email.
- Confirmed redirection to the sign-up page.


Hallidayo:

PR #40291, opened 2025-11-10
Refactor/use text confirm modal for delete buckets

joshenlim:
Fixes DEPR-204

## Context

Just refactoring all delete bucket modals to use `TextConfirmModal` component instead.

Also updates `TextConfirmModal` to disable the delete CTA until the text string matches for better visual indication that the input value matches the confirmation string


alaister (approved):
LGTM!

- Tested on preview

PR #40293, opened 2025-11-10
Advisor refine

SaxonF:
image

Applies some small visual and content refinements to the Advisor panel. Advisor issue titles can be quite long making it hard to pass. It now displays the type along with entity for Lint issues instead. Also moved a few shared utils and types.

To test:
- Open new advisor panel from toolbar
- View advisor issues
- View messages


joshenlim (approved):

PR #40294, opened 2025-11-10
feat: add stopping pipeline status in ui

imor:
This PR handles the stopping status for a pipeline. The `stopping` state will now be returned by the api.

Companion `etl-api` PR: https://github.com/supabase/etl/pull/443

### DIFF
--- a/apps/studio/components/interfaces/Database/ETL/Pipeline.utils.ts
+++ b/apps/studio/components/interfaces/Database/ETL/Pipeline.utils.ts
@@ -58,6 +58,11 @@ const PIPELINE_STATE_MESSAGES = {
     message: 'Initializing replication. Table status will be available once running.',
     badge: 'Starting',
   },
+  stopping: {
+    title: 'Pipeline stopping',
+    message: 'Stopping replication. Table replication will be paused once stopped',
+    badge: 'Starting',
+  },
   running: {
     title: 'Pipeline running',
     message: 'Replication is active and processing data',
@@ -100,6 +105,8 @@ export const getPipelineStateMessages = (
       return PIPELINE_STATE_MESSAGES.stopped
     case 'started':
       return PIPELINE_STATE_MESSAGES.running
+    case 'stopping':
+      return PIPELINE_STATE_MESSAGES.stopping
     case 'unknown':
       return PIPELINE_STATE_MESSAGES.unknown
     default:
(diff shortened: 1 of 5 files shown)


apps/studio/components/interfaces/Database/ETL/Pipeline.utils.ts
```diff
@@ -58,6 +58,11 @@ const PIPELINE_STATE_MESSAGES = {
     message: 'Initializing replication. Table status will be available once running.',
     badge: 'Starting',
   },
+  stopping: {
+    title: 'Pipeline stopping',
+    message: 'Stopping replication. Table replication will be paused once stopped',
+    badge: 'Starting',
```
joshenlim: should say Stopping here?

apps/studio/components/interfaces/Database/ETL/Pipeline.utils.ts
joshenlim, line 1: JFYI i noticed that we currently (probably) have some optimistic rendering logic in the UI
as stopping a pipeline does currently showing "Stopping" in the UI - will be good to clean that up (i'm trying to find where the optimistic rendering is happening)

joshenlim: mmmm actually sorry! maybe not - the optimistic rendering probably needs to stay given that all these are likely happening async

e.g i'm presuming the API endpoint /pipelines/id/stop doesn't return the pipeline status as Stopping immediately

joshenlim: but all good for now i reckon

joshenlim (approved):
Verified changes on staging preview - LGTM! Checked to refresh the page after clicking "Stop pipeline" to ensure that we're using the pipeline status from the API

apps/studio/components/interfaces/Database/ETL/Pipeline.utils.ts
imor: Right will fix.

imor: I see you fixed it, thanks Joshen.

apps/studio/components/interfaces/Database/ETL/Pipeline.utils.ts
imor: Yes the `stop` endpoint doesn't return the status immediately, it is polled via the `status` endpoint. Although we could start sending an immediate status on `start`/`stop` endpoints for faster rendering on UI.

PR #40295, opened 2025-11-10
Only show replication action if publication has been created

joshenlim:
## Context

Realised that replication actions on a analytics bucket table should only show if there's a corresponding publication created

image

We already have that logic in place by hiding the replication status column in the table UI, but left this one out


alaister (approved):
LGTM!

PR #40305, opened 2025-11-10
Fix table editor not scrollable when no rows

joshenlim:
Fixes FE-2105

## Context

Table editor is not horizontally scrollable when there's no rows due to the empty state being rendered outside of the DataGrid. For context the reason why we did it this way (using absolutely positioned divs) is because rendering the empty state in the DataGrid (the library has some param to do so) makes the empty state not stay in place when scrolling horizontally (the empty state will scroll as you scroll the table) and it's anchored to the grid

## Changes involved

- Add pointer events none to the empty state to allow scroll events to pass through
- Add pointer events auto to text renders (just in case for some reason the user wants to select and highlight the texts in the empty state)
- Buttons are already pointer events auto so verified that they're all still clickable

## To test

- Create a table that has sufficient number of columns (or columns with long names) such that it'll be scrollable in the table editor
- Verify that you can still scroll the table columns in the editor while the empty state is shown

### DIFF
--- a/apps/studio/components/grid/components/grid/Grid.tsx
+++ b/apps/studio/components/grid/components/grid/Grid.tsx
@@ -147,7 +147,7 @@ export const Grid = memo(
           {(rows ?? []).length === 0 && (
             <div
               className={cn(
-                'absolute top-9 p-2 w-full z-[1]',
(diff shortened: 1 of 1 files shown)


awaseem (approved):
Tested in self hosting and staging ✅

PR #40310, opened 2025-11-10
refactor(close confirmation modal): abstract out useConfirmOnClose hook

charislam:
Abstract out a hook, useConfirmOnClose, for the pattern where we show a
confirmation modal on close, conditional on whether the user has made
edits.

### DIFF
--- a/apps/studio/hooks/ui/useConfirmOnClose.tsx
+++ b/apps/studio/hooks/ui/useConfirmOnClose.tsx
@@ -1,69 +1,56 @@
-import { useCallback, useState, type ReactNode } from 'react'
+import { useCallback, useMemo, useState } from 'react'
+import useLatest from '../misc/useLatest'
 
-export type ConfirmOnCloseModalProps = {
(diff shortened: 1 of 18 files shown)


apps/studio/hooks/ui/useConfirmOnClose.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  const [visible, setVisible] = useState(false)
+
+  const confirmOnClose = useCallback(() => {
+    if (checkIsDirty()) {
+      setVisible(true)
+    } else {
+      onClose()
+    }
+  }, [checkIsDirty, onClose])
```
alaister: `useCallback()` is probably negatively impacting performance here since `checkIsDirty` and `onClose` are almost never memoized. We can still use it though if you wrap them both with `useLatest()` and remove them from the dependency array.

alaister (approved):
LGTM!

- Tested locally

PR #40317, opened 2025-11-10
feat(studio): rename reports to observability

kemaldotearth:

YES

Some internal tidy up. Reports becomes Observability to acomodate future changes.


apps/studio/pages/project/[ref]/observability/database.tsx
```diff
@@ -0,0 +1,409 @@
...
+        state.setSelectedDatabaseId(db)
+      }, 100)
+    }
+    if (chart !== undefined) {
+      setTimeout(() => {
+        const el = document.getElementById(chart)
+        if (el) el.scrollIntoView({ behavior: 'smooth', block: 'center' })
+      }, 200)
+    }
+  }, [db, chart, state])
```
Comment (charislam) [inline apps/studio/pages/project/[ref]/observability/database.tsx]:
```suggestion
  const stateSyncedFromUrlRef = useRef(false)
  useEffect(() => {
    if (stateSyncedFromUrlRef.current) return
    stateSyncedFromUrlRef.current = true
  
    if (db !== undefined) {
      setTimeout(() => {
        // [Joshen] Adding a timeout here to support navigation from settings to reports
        // Both are rendering different instances of ProjectLayout which is where the
        // DatabaseSelectorContextProvider lies in (unless we reckon shifting the provider up one more level is better)
        state.setSelectedDatabaseId(db)
      }, 100)
    }
    if (chart !== undefined) {
      setTimeout(() => {
        const el = document.getElementById(chart)
        if (el) el.scrollIntoView({ behavior: 'smooth', block: 'center' })
      }, 200)
    }
  }, [db, chart, state])
```

I think we still want this to run only once, so we need to track its status in a ref to prevent reruns

apps/studio/pages/project/[ref]/observability/postgrest.tsx
```diff
@@ -0,0 +1,162 @@
...
+      }, 100)
+    }
+    if (chart !== undefined) {
+      setTimeout(() => {
+        const el = document.getElementById(chart)
+        if (el) el.scrollIntoView({ behavior: 'smooth', block: 'center' })
+      }, 200)
+    }
+    /* eslint-disable-next-line react-hooks/exhaustive-deps */
+  }, [])
```
Comment (charislam) [inline apps/studio/pages/project/[ref]/observability/postgrest.tsx]:
```suggestion
  const hasStateSyncedFromUrlRef = useRef(false)
  useEffect(() => {
    if (hasStateSyncedFromUrlRef.current) return
    hasStateSyncedFromUrlRef.current = true
   
    if (db !== undefined) {
      setTimeout(() => {
        // [Joshen] Adding a timeout here to support navigation from settings to reports
        // Both are rendering different instances of ProjectLayout which is where the
        // DatabaseSelectorContextProvider lies in (unless we reckon shifting the provider up one more level is better)
        state.setSelectedDatabaseId(db)
      }, 100)
    }
    if (chart !== undefined) {
      setTimeout(() => {
        const el = document.getElementById(chart)
        if (el) el.scrollIntoView({ behavior: 'smooth', block: 'center' })
      }, 200)
    }
  }, [db, state, chart])
```

Let's make this explicit in code instead of disabling the lint (disabling it can cause problems later if someone changes the effect body)

apps/studio/pages/project/[ref]/observability/realtime.tsx
```diff
@@ -0,0 +1,222 @@
...
+      }, 100)
+    }
+    if (chart !== undefined) {
+      setTimeout(() => {
+        const el = document.getElementById(chart)
+        if (el) el.scrollIntoView({ behavior: 'smooth', block: 'center' })
+      }, 200)
+    }
+    /* eslint-disable-next-line react-hooks/exhaustive-deps */
+  }, [])
```
Comment (charislam) [inline apps/studio/pages/project/[ref]/observability/realtime.tsx]:
```suggestion
  const hasStateSyncedFromUrlRef = useRef(false)
  useEffect(() => {
    if (hasStateSyncedFromUrlRef.current) return
    hasStateSyncedFromUrlRef.current = true
  
    if (db !== undefined) {
      setTimeout(() => {
        // [Joshen] Adding a timeout here to support navigation from settings to reports
        // Both are rendering different instances of ProjectLayout which is where the
        // DatabaseSelectorContextProvider lies in (unless we reckon shifting the provider up one more level is better)
        state.setSelectedDatabaseId(db)
      }, 100)
    }
    if (chart !== undefined) {
      setTimeout(() => {
        const el = document.getElementById(chart)
        if (el) el.scrollIntoView({ behavior: 'smooth', block: 'center' })
      }, 200)
    }
  }, [state, db, chart])
```

charislam (approved):
Has most of this code been copied/pasted from elsewhere, and if so, can we delete the originals?

apps/studio/pages/project/[ref]/observability/realtime.tsx
Comment (kemaldotearth) [reply apps/studio/pages/project/[ref]/observability/realtime.tsx]: Nice, I actually wondered if this was going to create a recursive loop or something - the useEffect dependencies have scared me for life.

kemaldotearth:
Waiting on updating some references in docs, then will merge sequentially.

charislam (commented):
There's odd behavior on two menu items:
- The realtime report uses a different side menu (is this the old one)?
- There's a full page reload for custom reports

https://github.com/user-attachments/assets/b89cc2c5-3838-4020-b5fe-6ebcd3e8e0a3

charislam (approved):

PR #40329, opened 2025-11-11
Improve granularity of replication status

joshenlim:
## Context

For analytics buckets, currently the replication status column is a little too broad where it's only checking if the corresponding postgres table of the namespace table exists in the replication publication. However we should also be considering the replication pipeline as well for a more accurate representation and so this PR handles a couple of these scenarios

## Changes involved

- Replication status column will only show if the analytics bucket has a corresponding publication + replication pipeline
  - Previously was only checking for corresponding publication 
- More granular status reflection + tooltips
  - If the table is under the replication publication, and the pipeline is running
    image
  - If there's issues with the replication pipeline itself
    image
  - If the table is not under the replication publication
    image

## To test

- [ ] Minimally sanity check the analytics buckets UI on staging preview - e.g start/stop replication on a table

### DIFF
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
@@ -1,5 +1,5 @@
 import { snakeCase, uniq } from 'lodash'
-import { MoreVertical, Pause, Play, Trash } from 'lucide-react'
+import { Loader2, MoreVertical, Pause, Play, Trash } from 'lucide-react'
 import Link from 'next/link'
 import { useState } from 'react'
 import { toast } from 'sonner'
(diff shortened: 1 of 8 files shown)


joshenlim:
JFYI will need to make a bit of changes here

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
```diff
@@ -74,9 +80,27 @@ export const TableRowComponent = ({
...
+  const inferredPostgresTable = inferPostgresTableFromNamespaceTable({
+    publication,
+    tableName: table.name,
+  })
+  const isTableUnderReplicationPublication = !!inferredPostgresTable
+  const hasReplication = !!pipeline && !!publication
+  const isPipelineRunning = pipelineStatus === 'started'
+  const isReplicating = isTableUnderReplicationPublication && isPipelineRunning
+
+  const getStatusLabel = () => {
```
alaister: Nit: convert this to a useMemo instead of a function call... `{getStatusLabel()}` makes me feel weird lol

alaister (approved):
LGTM!

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
joshenlim: good point! 🙏

PR #40331, opened 2025-11-11
Only update FDW when removing namespace table if table is in FDW instance tables

joshenlim:
## Context

Currently we're updating the FDW instance irregardless. But if the namespace was for example added outside of the dashboard - it wouldn't be connected to a publication / replication / wrapper instance, so in this scenario we only need to delete the namespace table then


alaister (approved):
LGTM!

PR #40332, opened 2025-11-11
Hide cmd k keyboard shortcut if disabled in account pref

joshenlim:
## Context

As per PR title - keyboard shortcut should be hidden if disabled in account pref

image

image

## To test

- [ ] Disable CMD K shortcut in account pref and verify that the keyboard shortcut hides in the top nav


charislam (approved):
LGTM!

PR #40353, opened 2025-11-11
feat(studio): setup soft-delete and easy branch restore

avallete:

- Add "soft deleted branch" section to the branching section

https://github.com/user-attachments/assets/f738288a-e3ee-4719-b0e9-ebf4b147710f

## Summary by CodeRabbit

* **New Features**
  * Display countdown timers showing when branches will be permanently deleted and show "Deletion pending..." if past due
  * Added a dedicated "Scheduled for deletion" section with loader, empty state, list, and restore actions
  * Enable restoring branches scheduled for deletion with success feedback (toast) and UI reset

* **Bug Fixes**
  * Fixed branch list sorting edge case to avoid sporadic dropdown errors


apps/studio/components/layouts/AppLayout/BranchDropdown.tsx
```diff
@@ -99,7 +100,11 @@ export const BranchDropdown = () => {
   } as unknown as Branch
 
   const mainBranch = branches?.find((branch) => branch.is_default)
-  const restOfBranches = branches
+  const [_scheduledForDeletionBranches, activeBranches] = partition(
+    branches,
+    (branch) => branch.deletion_scheduled_at !== undefined
+  )
+  const restOfBranches = activeBranches
```
awaseem: We don't need partition here since we're not even using the results of the first array argument. Something better would be:


```suggestion
  const restOfBranches = branches.filter((branch) => branch.deletion_scheduled_at !== undefined))
```

awaseem (approved):
Minor changes

awaseem:
Tested on staging and it work!

PR #40354, opened 2025-11-11
docs: Add CTAs for ui lib

saltcod:
Add a CTA for the ui library in the next/react/vue/nuxt guides

screenshot-2025-11-11-at-11 38 21


apps/docs/content/guides/getting-started/quickstarts/nextjs.mdx
```diff
@@ -3,6 +3,7 @@ title: 'Use Supabase with Next.js'
 subtitle: 'Learn how to create a Supabase project, add some sample data, and query from a Next.js app.'
 breadcrumb: 'Framework Quickstarts'
 hideToc: true
+uiLibraryCta: true
```
saltcod: I prob shouldn't do this, but not sure the right way. 

Should I import the <UiLibraryCta /> component as an available mdx component? Should I create a $Partial?  Not sure

apps/docs/components/UiLibraryCta.tsx
```diff
@@ -0,0 +1,24 @@
...
+        <div>
+          <div className="flex items-center gap-2">
+            <h3 className="mb-1 font-medium">Drop-in UI components for your Supabase app</h3>
+            <Badge variant="success">New</Badge>
+          </div>
+          <p className="text-sm text-muted-foreground">
+            UI components built on shadcn/ui that connect to Supabase via a single command.
+          </p>
+        </div>
+        <Button type="primary" iconRight={<ArrowRight className="" />}>
```
saltcod: I also have no clue how to make a link look like a button

ChrisChinchilla:
@saltcod A simpler option may be to use an admonition. We'd lose the fancy buttons etc, but it's way simpler, on the other hand, when we roll UI library components properly in, the "new" aspect can also be more applicable and useful.

And if we keep it loosely as is, probably just make it in to a partial. I can sort.

saltcod:
@ChrisChinchilla yeah was thinking an admonition makes more sense actually. Sounds good.

ChrisChinchilla:
@saltcod Here you go, I simplified it a bit. Still slightly confusing having a tutorial with a CTA in the middle telling people that they can look elsewhere. But it's a good short-term fix while we get the tutorials updated

ChrisChinchilla (approved):

PR #40355, opened 2025-11-11
feat(design-system): add <MetricsCard /> component

kemaldotearth:

YES

Abstracting and tidying the metric block we added during auth overview design work. This will likely be used elsewhere to display singular metrics, so componentising.


packages/ui-patterns/src/MetricsCard/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+          'py-4 px-6 flex flex-row items-center justify-between gap-2 space-y-0 pb-0 border-b-0 relative',
+          className
+        )}
+        {...props}
+      >
+        <div className="flex flex-row items-center gap-2">{children}</div>
+        {hasLink && (
+          <Button type="text" size="tiny" className="px-1 text-foreground-lighter" asChild>
+            <Link href={href || ''}>
+              <ExternalLink size={14} strokeWidth={1.5} />
```
charislam:
```suggestion
              <ExternalLink aria-disabled={true} size={14} strokeWidth={1.5} />
              <span className="sr-only">More information</span>
```

packages/ui-patterns/src/MetricsCard/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+        </MetricsCardContext.Provider>
+      </Card>
+    )
+  }
+)
+MetricsCard.displayName = 'MetricsCard'
+
+interface MetricsCardHeaderProps extends React.HTMLAttributes<HTMLDivElement> {
+  hasLink?: boolean
+  href?: string
```
charislam: I don't know if I like having both, it opens up seemingly invalid states like `{ hasLink: true, href: undefined }` or `{ hasLink: false, href: 'some-url' }`. Is there any case where we would want to display the icon button without actually having a link?

packages/ui-patterns/src/MetricsCard/index.tsx
```diff
@@ -0,0 +1,268 @@
...
+const MetricsCardIcon = React.forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDivElement>>(
+  ({ className, ...props }, ref) => (
+    <div ref={ref} className={cn('text-foreground-light', className)} {...props} />
+  )
+)
+MetricsCardIcon.displayName = 'MetricsCardIcon'
+
+interface MetricsCardLabelProps extends React.HTMLAttributes<HTMLDivElement> {
+  hasTooltip?: boolean
+  tooltip?: string
```
charislam: Same here, is there any case where we would want `{ hasTooltip: true, tooltip: undefined }`?

apps/studio/components/ui/Charts/AreaChart.tsx
```diff
@@ -148,12 +148,6 @@ const AreaChart = ({
           />
         </RechartAreaChart>
       </Container>
-      {data && (
```
charislam, line 151: Is this removal intentional? It feels like it makes more sense to show it, at least on our existing charts:

Before:

Image

After:

Image

charislam (approved):

packages/ui-patterns/src/MetricsCard/index.tsx
kemaldotearth: Good shout, removing the `hasLink` and `hasTooltip` props!

apps/studio/components/ui/Charts/AreaChart.tsx
kemaldotearth: 🤦 My bad, removed it from the wrong file. Re-introduced it!

apps/design-system/content/docs/fragments/metrics-card.mdx
```diff
@@ -0,0 +1,21 @@
+---
+title: Metrics Card
```
dnywh: Tiniest of nits: plural on _Metrics_ reads funny to me. It’s a singular _Metric Card_, no?

apps/design-system/registry/default/example/metrics-card-with-icon-link-tooltip.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const diffPercentage = (diff / averageValue) * 100
+
+  return (
+    <div className="w-1/2">
+      <MetricsCard isLoading={!data.length}>
+        <MetricsCardHeader href="https://www.supabase.io">
+          <MetricsCardIcon>
+            <User2 size={16} strokeWidth={1.5} />
+          </MetricsCardIcon>
+          <MetricsCardLabel tooltip="The number of Active Users over the last 24 hours">
```
dnywh:
```suggestion
          <MetricsCardLabel tooltip="The number of active users over the last 24 hours">
```

Or is Active Users something special?

apps/design-system/registry/default/example/metrics-card.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const averageValue = data.reduce((acc, curr) => acc + curr.value, 0) / data.length
+
+  const diff = data[data.length - 1]?.value - data[0]?.value || 0
+  const diffPercentage = (diff / averageValue) * 100
+
+  return (
+    <div className="w-1/2">
+      <MetricsCard isLoading={!data.length}>
+        <MetricsCardHeader href="https://www.supabase.io">
+          <MetricsCardLabel tooltip="The number of Active Users over the last 24 hours">
```
dnywh: Same as above RE casing of _Active Users_

dnywh:
Weird build error on docs:

```
Module not found: Can't resolve './build/Release/queryparser.node'
> 1 | const PgQuery = require('./build/Release/queryparser.node');
    | ^
  2 |
  3 | module.exports = {
  4 |   parseQuery(query) {
```

packages/ui-patterns/src/MetricCard/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+    return `${date.format('MMM D')}, ${displayHour}${period}`
+  }
+
+  return (
+    <div className="bg-black/90 text-white p-2 rounded text-xs">
+      {label && (
+        <div className="text-foreground-light">{formatTimestamp(payload[0].payload.timestamp)}</div>
+      )}
+      <div>{payload[0].value.toLocaleString(undefined, { maximumFractionDigits: 0 })}</div>
+    </div>
```
dnywh, line 212: This top line of text is a bit hard to read on light mode: 

Image

The hardcoded black background and dynamic `foreground-light` combo is probably to blame. How about something like this?


```suggestion
    <div className="bg-black/90 p-2 rounded text-xs">
      {label && <p className="text-white/50">{formatTimestamp(payload[0].payload.timestamp)}</p>}
      <p className="text-white">
        {payload[0].value.toLocaleString(undefined, { maximumFractionDigits: 0 })}
      </p>
    </div>
```

Cleans up a bit of the syntax too. Looks like this:

Image

This is inherited from a few other components, right? I see similar markup in:

- AreaChart
- BarChart

dnywh (approved):
Looks good! Approved but two things to probably investigate:

- [Light mode text legibility](https://github.com/supabase/supabase/pull/40355/files#r2525391721)
	- Non-blocking as I see it predates your work here
- Error on docs build
	- Could be unrelated or inconsequential

apps/design-system/content/docs/fragments/metrics-card.mdx
kemaldotearth: Yeah you're right here, 🤿  in to rename all symbols.

PR #40356, opened 2025-11-11
Fix: updated RLS to use messages instead of formatted error

### LINKED ISSUE #40282: Not showing error message in "Error running SQL query" toast
# Bug report

## Describe the bug

When creating a RLS policy for a table and trying to save the Error running SQL query" toast will appear but no error message is shown. I can see the error in the response (network tab). 

## To Reproduce

1. Go to 'Authentications'
2. Click on 'Policies'
3. Scroll down to 'Create policy'
4. Choose a Policy Template (Make sure it cause an error)
5. The Error toast will appear with no error message

## Expected behavior

The error message should be shown in error toast.

## Screenshots

Image

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

awaseem:

Yes

RLS error messages were not showing up, this fixes it to use the message string

No error messages

## What is the new behavior?

Error messages when modify RLS through editor

## Additional context

- closes #40282

### DIFF
--- a/apps/studio/data/sql/execute-sql-mutation.ts
+++ b/apps/studio/data/sql/execute-sql-mutation.ts
@@ -16,7 +16,6 @@ export type QueryResponseError = {
   code: string
   message: string
   error: string
-  formattedError: string
   file: string
   length: number
   line: string
--- a/apps/studio/components/interfaces/Auth/Policies/PolicyEditorPanel/QueryError.tsx
+++ b/apps/studio/components/interfaces/Auth/Policies/PolicyEditorPanel/QueryError.tsx
@@ -23,7 +23,7 @@ export const QueryError = ({
   setOpen: Dispatch<SetStateAction<boolean>>
 }) => {
   const formattedError =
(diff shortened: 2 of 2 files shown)


apps/studio/data/sql/execute-sql-mutation.ts
```diff
@@ -16,7 +16,6 @@ export type QueryResponseError = {
   code: string
   message: string
   error: string
-  formattedError: string
```
awaseem, line 19: This is not used anywhere. Its kinda strange, we as a team decided to move off this implicitly. See investigation notes here: 

>Wow I went down a deep rabbithole. So formattedError is only used in a single place which is this prompt. Also all the errors in the SQL editor or event the cmd E sql editor don't have formatting. It seems everyone in the code base moved of formatted error

> It only works for self hosted because we handle it in the API, the fix is simple


>N one of them have formatting in platform or self hosted, I guess this just become the default now

charislam (approved):
Tested for both self-hosted and hosted

PR #40367, opened 2025-11-12
chore(studio): improve Storage UI

dnywh:

- UI improvements to Storage
- Fixes DEPR-238

## What is the new behavior?

- More glanceable, clearer, indication that you are inside a bucket
  - Bucket icon via `PageLayout` `icon` prop
  - Static “Buckets” breadcrumb that matches /buckets/ slug
- Nicer Admonition with up-and-coming _New_ variant
  - Actual variant will be added later
- Handle cases where the maximum number of analytics buckets has been reached 

| Before | After |
| --- | --- |

### DIFF
--- a/apps/studio/components/interfaces/Auth/EmailTemplates/EmailTemplates.tsx
+++ b/apps/studio/components/interfaces/Auth/EmailTemplates/EmailTemplates.tsx
@@ -193,10 +193,10 @@ export const EmailTemplates = () => {
                     <div className="flex flex-col md:flex-row md:items-center gap-y-2 md:gap-x-8 justify-between px-2 py-1">
                       <div className="flex flex-col gap-y-0.5">
                         <div className="flex flex-col gap-y-2 items-start">
-                          <Badge variant="success" className="-ml-0.5">
-                            NEW
+                          <Badge variant="success" className="-ml-0.5 uppercase">
+                            New
                           </Badge>
-                          <p className="text-sm">
+                          <p className="text-sm font-medium">
                             Notify users about security-sensitive actions on their accounts
                           </p>
                         </div>
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/SimpleConfigurationDetails.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/SimpleConfigurationDetails.tsx
@@ -37,7 +37,6 @@ export const SimpleConfigurationDetails = ({ bucketName }: { bucketName?: string
             >
               Learn more
             </InlineLink>
-            .
           </ScaffoldSectionDescription>
         </div>
         <CopyEnvButton
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/CreateAnalyticsBucketModal.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/CreateAnalyticsBucketModal.tsx
@@ -192,7 +192,7 @@ export const CreateAnalyticsBucketModal = ({
           type={buttonType}
           className={buttonClassName}
           icon={<Plus size={14} />}
-          disabled={!canCreateBuckets || !icebergCatalogEnabled || disabled}
+          disabled={!canCreateBuckets || !icebergCatalogEnabled || buckets.length >= 2 || disabled}
           style={{ justifyContent: 'start' }}
           onClick={() => setVisible(true)}
           tooltip={{
@@ -203,7 +203,9 @@ export const CreateAnalyticsBucketModal = ({
                 ? 'Analytics buckets are not enabled for your project. Please contact support to enable it.'
                 : !canCreateBuckets
                   ? 'You need additional permissions to create buckets'
-                  : tooltip?.content?.text,
+                  : buckets.length >= 2
+                    ? 'Bucket limit reached'
+                    : tooltip?.content?.text,
             },
           }}
         >
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/CreateAnalyticsBucketModal.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/CreateAnalyticsBucketModal.tsx
@@ -105,7 +105,7 @@ export const CreateAnalyticsBucketModal = ({
 
   const [visible, setVisible] = useState(false)
 
-  const { data: buckets = [] } = useAnalyticsBucketsQuery({ projectRef: ref })
+  const { data: buckets = [], isLoading } = useAnalyticsBucketsQuery({ projectRef: ref })
   const icebergCatalogEnabled = useIsAnalyticsBucketsEnabled({ projectRef: ref })
   const wrappersExtenstionNeedsUpgrading = wrappersExtensionState === 'needs-upgrade'
 
@@ -192,7 +192,13 @@ export const CreateAnalyticsBucketModal = ({
           type={buttonType}
           className={buttonClassName}
           icon={<Plus size={14} />}
-          disabled={!canCreateBuckets || !icebergCatalogEnabled || buckets.length >= 2 || disabled}
+          disabled={
+            !canCreateBuckets ||
+            !icebergCatalogEnabled ||
+            isLoading ||
+            buckets.length >= 2 ||
+            disabled
+          }
           style={{ justifyContent: 'start' }}
           onClick={() => setVisible(true)}
           tooltip={{
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
@@ -100,8 +100,14 @@ export const AnalyticsBuckets = () => {
         <EmptyBucketState bucketType="analytics" />
       ) : (
         <div className="flex flex-col gap-y-4">
-          <ScaffoldHeader className="py-0">
+          <ScaffoldHeader className="py-0 flex flex-row items-baseline gap-x-2">
             <ScaffoldSectionTitle>Buckets</ScaffoldSectionTitle>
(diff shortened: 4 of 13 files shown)


apps/studio/components/interfaces/Auth/EmailTemplates/EmailTemplates.tsx
```diff
@@ -193,10 +193,10 @@ export const EmailTemplates = () => {
                     <div className="flex flex-col md:flex-row md:items-center gap-y-2 md:gap-x-8 justify-between px-2 py-1">
                       <div className="flex flex-col gap-y-0.5">
                         <div className="flex flex-col gap-y-2 items-start">
-                          <Badge variant="success" className="-ml-0.5">
-                            NEW
+                          <Badge variant="success" className="-ml-0.5 uppercase">
+                            New
                           </Badge>
-                          <p className="text-sm">
+                          <p className="text-sm font-medium">
```
dnywh, line 199: Just matching changes elsewhere. Will sweep holistically as part of a later design-system PR.

apps/studio/components/interfaces/Storage/AnalyticsBuckets/CreateAnalyticsBucketModal.tsx
```diff
@@ -203,7 +203,9 @@ export const CreateAnalyticsBucketModal = ({
                 ? 'Analytics buckets are not enabled for your project. Please contact support to enable it.'
                 : !canCreateBuckets
                   ? 'You need additional permissions to create buckets'
-                  : tooltip?.content?.text,
+                  : buckets.length >= 2
+                    ? 'Bucket limit reached'
+                    : tooltip?.content?.text,
```
dnywh: Don’t let the user go through the new bucket creation flow if they’ve hit their limit. Explain via tooltip.

P.S. this `buckets.length >= 2` is hardcoded. Do we store this limit of `2` somewhere?

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -71,8 +100,14 @@ export const AnalyticsBuckets = () => {
...
         <div className="flex flex-col gap-y-4">
-          <ScaffoldHeader className="py-0">
+          <ScaffoldHeader className="py-0 flex flex-row items-baseline gap-x-2">
             <ScaffoldSectionTitle>Buckets</ScaffoldSectionTitle>
+            {analyticsBuckets.length > 0 && (
+              <span className="bg-surface-200 rounded-full px-1.5 py-1 leading-none text-xs text-foreground-lighter tracking-widest">
+                {analyticsBuckets.length}
+                /2
+              </span>
+            )}
```
dnywh, line 120: Conditionally show the limit count when getting close.

Another case of hardcoded `2` limit for max buckets.

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -41,28 +53,45 @@ export const AnalyticsBuckets = () => {
...
           </Button>
-        }
-      >
-        <p className="!leading-normal !mb-0 text-balance">
-          Analytics buckets are now in private alpha. Expect rapid changes, limited features, and
-          possible breaking updates. Please share feedback as we refine the experience and expand
-          access.
-        </p>
+        </div>
       </Admonition>
```
dnywh, line 98: More visually interesting Admonition. Sorry for the noise. Will wrap whatever the result is in the new variant, later.

apps/studio/components/interfaces/Storage/StorageMenuV2.tsx
```diff
@@ -35,7 +35,7 @@ export const StorageMenuV2 = () => {
                   <div className="flex items-center justify-between">
                     <p className="truncate">{config.displayName}</p>
                     {isAlphaEnabled && (
-                      <Badge variant="default" size="small">
+                      <Badge variant="success" size="tiny">
```
dnywh, line 38: `tiny` is a new variant I’ve made for sidebar applications. Uppercase and, as it says, tiny!

apps/studio/pages/project/[ref]/storage/analytics/buckets/[bucketId].tsx
```diff
@@ -16,11 +17,19 @@ const AnalyticsBucketPage: NextPageWithLayout = () => {
   return (
     <PageLayout
       title={bucketId}
+      icon={
+        <div className="shrink-0 w-10 h-10 relative bg-surface-100 border rounded-md flex items-center justify-center">
+          <BucketIcon size={20} className="text-foreground-light" />
+        </div>
+      }
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/analytics/buckets/[bucketId].tsx:24]: We should probably hardcode this outer div on the `PageLayout` component (or its descendant).

apps/studio/pages/project/[ref]/storage/analytics/buckets/[bucketId].tsx
```diff
@@ -16,11 +17,19 @@ const AnalyticsBucketPage: NextPageWithLayout = () => {
...
       breadcrumbs={[
         {
           label: 'Analytics',
           href: `/project/${project?.ref}/storage/analytics`,
         },
+        {
+          label: 'Buckets',
+        },
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/analytics/buckets/[bucketId].tsx:32]: Static breadcrumb just to reinforce that we are on the buckets subdirectory.

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
joshenlim: i'm also thinking whether to include a tooltip here RE this limitation
as a user i might be wondering _why_ is there a 2 bucket limit? Perhaps we could say that this limit is while we're in alpha and may change

joshenlim: very minor nit here hahaha but it doesnt look vertically aligned
image

joshenlim: so something like this perhaps (will push this up)
image

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/SimpleConfigurationDetails.tsx
```diff
@@ -37,7 +37,6 @@ export const SimpleConfigurationDetails = ({ bucketName }: { bucketName?: string
             >
               Learn more
             </InlineLink>
-            .
           </ScaffoldSectionDescription>
```
joshenlim, line 40: Unrelated but I'm wondering whether this inline link might be better as a button beside "Copy all"

apps/studio/components/interfaces/Storage/AnalyticsBuckets/CreateAnalyticsBucketModal.tsx
```diff
@@ -192,7 +192,7 @@ export const CreateAnalyticsBucketModal = ({
           type={buttonType}
           className={buttonClassName}
           icon={<Plus size={14} />}
-          disabled={!canCreateBuckets || !icebergCatalogEnabled || disabled}
+          disabled={!canCreateBuckets || !icebergCatalogEnabled || buckets.length >= 2 || disabled}
```
joshenlim: reckon we'll also need to disable this while buckets is loading then
otherwise users will be able to click on this initially

dnywh: Good catch

joshenlim (approved):

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/SimpleConfigurationDetails.tsx
dnywh: Opted against this because there is already a _Docs_ button up top. Felt like two button-level docs-like buttons would get confusing.

Or did you have another idea for what this button would say, aside from _Docs_?

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
dnywh: Nice, good catch. Also, I just changed the capitalisation to sentence case to match elsewhere. Let me know if you feel strongly about title case for stuff. @kemaldotearth and I are looking at capitalization/copywriting rules in general

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/SimpleConfigurationDetails.tsx
joshenlim: oh thats true - yeah that plus there's starting to be a lot of buttons as well
- Docs
- View replication
- Add tables
- (If schema not synced) Connect a table

agree lets just keep it inline hahah

apps/studio/components/interfaces/Storage/StorageMenuV2.tsx
dnywh: It makes all the sidebar variations the same, and is inching towards something more standard everywhere (future PR to come)

packages/ui/src/components/shadcn/ui/badge.tsx
```diff
@@ -17,6 +17,7 @@ const badgeVariants = cva('inline-flex items-center rounded-full font-normal whi
       outline: 'bg-transparent text border border-foreground-muted',
     },
     size: {
+      tiny: 'px-1.5 py-[3px] text-[10px] leading-none tracking-wide uppercase',
```
fsansalvadore, line 20: `9px` would look even better, but lets stick with 10 😁

fsansalvadore (approved):
👍

PR #40368, opened 2025-11-12
fix: correctly set entrypoint, import map path and static patterns for new functions

laktek:
Previously, these values were not set when a function was created. Only when updating a function you could set those.

Note: We should consider combining the logic / components in `apps/studio/pages/project/[ref]/functions/new.tsx` and `apps/studio/pages/project/[ref]/functions/[functionSlug]/code.tsx` to prevent this from happening in future. I didn't attempt to do that in this PR as that would be better to do by someone who understand's frontend architecture better.


joshenlim (approved):

PR #40381, opened 2025-11-12
Refactor polling logic for Analytics Buckets while syncing with namespace

joshenlim:
## Context

Previously, we implemented long polling when hitting connecting tables to the Analytics Bucket, such that we'd store on the client side what table was being connected and we'd poll for both namespaces and namespace tables until the data is synced (connected postgres tables have a matching namespace table)

However this logic is flawed as the long polling will cease once the user refreshes the browser (and client state is lost)

Opted to refactor the logic here, such that we'd check if the data has been synced once the requests to fetch all analytics bucket relevant data is finished. That way the long polling will still kick in after the user has refreshed.

For context, connecting tables to the bucket just updates the replication publication -> which this data is immediately updated once the POST request completes. So we're just checking that against the data from the namespaces to see if the data has been synced

## Other changes

- Improved loading state by adding a new `GenericTableLoader` component
- Improved loading behaviour as the "state" while in an analytics bucket flickers to "missing extension" for a second at the start
- Added error handling for if the namespaces query fails
- Added tooltip to "Connecting tables" text for clarity
  Screenshot 2025-11-12 at 20 38 08


apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/index.tsx
```diff
@@ -38,7 +39,7 @@ type NamespaceWithTablesProps = {
   wrapperInstance: FDW
   wrapperValues: Record<string, string>
   wrapperMeta: WrapperMeta
-  tablesToPoll: { schema: string; name: string }[]
+  // tablesToPoll: { schema: string; name: string }[]
```
alaister: Did you mean to leave this?
```suggestion
```

alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/index.tsx
joshenlim: ahh need to remove sorry!

PR #40385, opened 2025-11-12
Fix: updated action to include state

awaseem:
- Refactor action into scripts
- Show better states of PRs

### DIFF
--- a/.github/workflows/dashboard-pr-reminder.yml
+++ b/.github/workflows/dashboard-pr-reminder.yml
@@ -16,114 +16,16 @@ jobs:
   check-dashboard-prs:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
       - name: Find Dashboard PRs older than 24 hours
         id: find-prs
         uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
         with:
           script: |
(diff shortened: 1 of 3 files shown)


.github/workflows/dashboard-pr-reminder.yml
```diff
@@ -16,114 +16,16 @@ jobs:
   check-dashboard-prs:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
charislam:
```suggestion
        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
```

Let's use commit hashes on all actions to prevent supply chain attacks via GHA

scripts/actions/find-stale-dashboard-prs.js
```diff
@@ -0,0 +1,214 @@
...
+        )
+      }
+    }
+
+    page++
+  }
+
+  console.log(`Found ${stalePRs.length} stale Dashboard PRs`)
+
+  // Sort by age (oldest first)
```
charislam: This would be a newest first, or is my head all turned around?

charislam (approved):

scripts/actions/find-stale-dashboard-prs.js
awaseem: Nope comment is wrong, it should be newest first

PR #40386, opened 2025-11-12
fix: Small fixes in Analytics and Vector Buckets

ivasilov:
# Summary
- Extend `useProtectedSchemas` so FDW-managed schemas are discovered for every wrapper that defines supabase_target_schema, and surface the FDW type/server metadata alongside internal schemas;
- Tailor ProtectedSchemaWarning messaging for Iceberg and S3 vector wrappers while retaining the generic fallback dialog for other protected schemas;
- Include the target schema when creating a S3 vector wrapper creation to show the schema as protected;
- Fix analytics bucket row navigation so modifier clicks (⌘/Ctrl) work with base paths (present on previews/staging/prod).

Screenshot 2025-11-12 at 17 37 36

Screenshot 2025-11-12 at 17 37 45


awaseem (approved):
LGTM

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -33,7 +34,7 @@ export const AnalyticsBuckets = () => {
   ) => {
     const url = `/project/${ref}/storage/analytics/buckets/${encodeURIComponent(bucketId)}`
     if (event.metaKey || event.ctrlKey) {
-      window.open(url, '_blank')
+      window.open(`${BASE_PATH}${url}`, '_blank')
```
joshenlim, line 52: unrelated but i feel like this should be handled by `Link` - will look into this

joshenlim: ahhh cant make a `tr` as a `a` - no worries then

joshenlim (approved):
Verified changes on preview!

PR #40392, opened 2025-11-12
chore: use entitlements when enabling PITR

ignaciodob:
Depends on https://github.com/supabase/infrastructure/pull/27470

This PR updates the PITR (Point-in-Time Recovery) side panel to use the entitlements system instead of checking for a free plan directly.

## Changes
- Replaces the isFreePlan check with the useCheckEntitlements hook to validate access to PITR variants
- Uses the `pitr.available_variants` entitlement to determine if a user has access to PITR functionality instead of the `isFreePlan` check

## Testing 
- Head to project/_/settings/addons
- Verify Free tier organizations cannot enable PITR addons see this notice:
image

- Verify that Orgs not on the Free Plan don't see the notice above


kevcodez (approved):

awaseem (approved):
LGTM

PR #40395, opened 2025-11-12
fix(cron): optimize queries for cron tables

charislam:
The `cron.job_run_details` table can get very long (millions of rows).
It's only indexed on the `runid` column, and adding a new index is a
non-trivial matter.

On the job run details table, we automatically sort and paginate by the
`start_time` column, so this creates a problem for very large tables.
This PR changes to use the `runid` column instead. Since `runid`s are
sequential, this should correlate pretty closely with `start_time`
anyway, while letting Postgres use the index.

Similarly, on the cron overviews table, the search for the latest run
required a sequential scan on the `job_run_details` table. Changed this
to also use the `runid` index instead.

Resolves FE-2115

### DIFF
--- a/apps/studio/data/database-cron-jobs/database-cron-jobs-infinite-query.ts
+++ b/apps/studio/data/database-cron-jobs/database-cron-jobs-infinite-query.ts
@@ -25,36 +25,25 @@ export type CronJob = {
 // [Joshen] Just to call out that I had AI help me with this, so please let me know if this can be optimized
 const getCronJobSql = ({ searchTerm, page }: { searchTerm?: string; page: number }) =>
   `
-WITH latest_runs AS (
-  SELECT 
-    jobid,
-    status,
-    MAX(start_time) AS latest_run
-  FROM cron.job_run_details
-  GROUP BY jobid, status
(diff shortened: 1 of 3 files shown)


apps/studio/data/database-cron-jobs/database-cron-jobs-infinite-query.ts
```diff
@@ -25,36 +25,25 @@ export type CronJob = {
...
-    latest_run
-  FROM latest_runs lr1
-  WHERE latest_run = (
-    SELECT MAX(latest_run) 
-    FROM latest_runs lr2 
-    WHERE lr2.jobid = lr1.jobid
-  )
-)
-SELECT 
+SELECT
```
charislam, line 28: I'm not 100% sure on this one. It does a nested loop over `job` which makes me a little nervous, but assuming `job` is much smaller than `job_run_details`, it makes much better use of the `job_run_details.runid` index. But if we have one job that clutters up most of the `job_run_details` table, that could be much worse... Or say we have several jobs that were only ever run once at the very beginning of the universe, that could also be very bad... 🤔 

Image

PR #40404, opened 2025-11-12
test: parallelize table editor tests

charislam:
Table editor tests were being run serially because they would otherwise stomp all over each other in setup. But this made them really slow when running locally (which dependent on machine, generally runs ~5 workers). This PR changes them to run in parallel, and makes sure the setup step runs only once by using a file lock.

Also cleans up miscellaneous other sources of flakiness in tests.


awaseem (approved):
LGTM

dnywh (approved):
Tested visual components on staging:

- RowEditor
- TableEditor
- SidePanel
- DeleteConfirmationDialogs
- UnshareSnippetModal

PR #40405, opened 2025-11-13
fix(www): rss feed

dnywh:

Fix.

The blog RSS feed hasn’t been publishing since August even though new blog posts were being published.

## What is the new behavior?

Added RSS feed generation to the `generateStaticContent.mjs` build script:
- Created `getAllBlogPosts()` function to fetch all blog posts (not just latest 10)
- Added RSS generation logic matching the original `rss.tsx` implementation
- RSS feed is now automatically regenerated during the build process

Also:
- ~~Cleaned up generated .xml files from .gitignore (like `sitemap.xml`, `rss.xml`)~~
  - _This was getting cached during build, so I restored `rss.xml`_
- Updated comment in `lib/rss.tsx` to clarify it's only used for customer stories and planetpg feeds

RSS feed should now update automatically on `pnpm run content:build` or `pnpm run build`.

## Additional context

When the blog migrated from Pages Router to App Router, the RSS feed generation mechanism was lost:

- Pages Router used `getStaticProps` to generate RSS feeds at build time
- App Router doesn't have `getStaticProps`, so there was no mechanism to regenerate the RSS feed
- The `generateStaticContent.mjs` build script was fetching blog posts but not generating the RSS feed
- The `lib/rss.tsx` file is still needed for `customers-rss.xml` and planetpg author feeds


apps/www/public/sitemap.xml
dnywh: Deleted and added to .gitignore. I assume we don’t want these generated files in our repo.

dnywh, line 1: Was already .gitignored, I believe.

craigcannon (approved):

fsansalvadore (approved):
Checked and the rss if fixed ✅

PR #40413, opened 2025-11-13
chore(studio): clean up ETL Replication UI

dnywh:

- UI improvements
- Resolves DEPR-245

The ETL Replication page(s) are a bit out of step with the rest of the studio application. Thsi is especially important as it becomes increasingly intertwined with analytics buckets.

## What is the new behavior?

- The ETL Replication page(s) now use standardised components
- Improved copywriting

| Before | After |
| --- | --- |
|  |  |
|  |  |

## Additional context

This is a draft PR.

### DIFF
--- a/apps/studio/components/interfaces/Database/ETL/Destinations.tsx
+++ b/apps/studio/components/interfaces/Database/ETL/Destinations.tsx
@@ -12,8 +12,9 @@ import { fetchReplicationPipelineVersion } from 'data/etl/pipeline-version-query
 import { useReplicationPipelinesQuery } from 'data/etl/pipelines-query'
 import { useReplicationSourcesQuery } from 'data/etl/sources-query'
 import { DOCS_URL } from 'lib/constants'
-import { Button, cn, Input_Shadcn_ } from 'ui'
+import { Button, cn } from 'ui'
 import { GenericSkeletonLoader } from 'ui-patterns'
+import { Input } from 'ui-patterns/DataInputs/Input'
 import { DestinationPanel } from './DestinationPanel/DestinationPanel'
 import { DestinationRow } from './DestinationRow'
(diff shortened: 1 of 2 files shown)


apps/studio/pages/project/[ref]/database/etl/index.tsx
```diff
@@ -20,35 +20,27 @@ const DatabaseReplicationPage: NextPageWithLayout = () => {
...
 
   return (
-    <>
+    <PageLayout
+      size="large"
+      title="ETL Replication"
+      subtitle={
+        enablePgReplicate
+          ? 'Replicate your database to any analytics destination'
+          : 'Send data to other destinations'
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/database/etl/index.tsx:29]: @imor why do we have two different subtitles depending on if `enablePgReplicate` is true?

Is this (new) first subtitle true for both?

> Replicate your database to any analytics destination

The second one is a bit vague in my opinion. Not a huge deal either way.

apps/studio/components/interfaces/Database/ETL/Destinations.tsx
```diff
@@ -95,19 +96,16 @@ export const Destinations = () => {
...
-                placeholder={'Filter destinations'}
-                value={filterString}
-                onChange={(e) => setFilterString(e.target.value)}
-              />
-            </div>
+            <Input
+              size="tiny"
+              icon={<Search size={12} />}
+              className="w-48 pl-8"
+              placeholder="Search for a destination"
```
dnywh, line 103: Language to match same UI in other pages.

apps/studio/pages/project/[ref]/database/etl/index.tsx
Comment (imor) [reply apps/studio/pages/project/[ref]/database/etl/index.tsx:29]: Agree, can be just one subtitle.

dnywh:
Cancelling as larger effort now kicking off.

PR #40418, opened 2025-11-13
Only show analytics and vector buckets for hosted

joshenlim:
## Context

As per PR title - we won't have support for these 2 types of buckets for self-hosted just yet so removing them from the UI for now

## To test

- [ ] Verify locally via studio-local that you only see files buckets for storage


alaister (approved):
LGTM!

- Tested on preview and self-hosted preview

PR #40419, opened 2025-11-13
Default auth users search to email

joshenlim:
## Context

Defaults search column for auth users to email instead of users ID

## To test

- [ ] Open the Auth users page, search column should default to email instead of user ID


alaister (approved):
LGTM!

- Tested on preview

PR #40423, opened 2025-11-13
chore(studio): default opt-in security notifications

dnywh:

- Feature flag tweaks
- Resolves DEPR-134

- The `securityNotifications` flag’s related feature preview is opt-in
	- Meaning all users must explicitly opt-in to see the feature

## What is the new behavior?

- The `securityNotifications` flag’s related feature preview is  opt-out
	- Meaning all users have this feature preview enabled, and must explicitly opt-out to go back to the previous UI
- 

> [!NOTE]
> We need to set the corresponding ConfigCat flag to 20% (or similar) to avoid rolling this out to everyone all at once.

## Additional context

I decided not to update the notice Admonition (with a link back to this feature preview in case folks want to later opt-out) because:

- I don’t think it’s a controversial enough change
- The Admonition is already quite lengthy


joshenlim (approved):
Verified changes on staging! minimally that feature preview is still available

joshenlim:
JFYI the E2E tests were failing on this branch (specifically database.spec) somehow despite multiple re-runs, but i've checked locally
all seems alright
image

PR #40424, opened 2025-11-13
feat: Add drag and drop for sql snippets

KyleQ1:
- Hard for me to test manually.
- Manual ordering between each of the snippets are not supported since it uses inserted_at in the db for ordering. Could be fixed by either changing that timestamp or using a different ordering like position. This could also be applied to the storage buckets too since they are super annoying to use too.

https://github.com/orgs/supabase/discussions/40408


Hallidayo:

PR #40430, opened 2025-11-13
Joshen/depr 262 bucketid route needs to handle if a bucket doesnt exist

joshenlim:
## Context

Handles re-routing logic when landing on a `[bucketId]` page that doesn't exist for analytics and vector buckets. 

## Changes involved
- Refactor `useSelectedBucket` to only consider file buckets, remove analytics bucket logic
- Route back to `/storage/analytics` if landing on an analytics bucket ID that doesn't exist
- Route back to `/storage/vectors` if landing on an vector bucket ID that doesn't exist
- Update files bucket to route back to `/storage/files` if landing on a files bucket ID that doesn't exist

## To test

- [ ] Try landing on a bucket ID that doesn't exist for files, analytics and vector buckets


apps/studio/components/interfaces/Storage/AnalyticsBuckets/useSelectedAnalyticsBucket.ts
```diff
@@ -0,0 +1,20 @@
...
+    data = [],
+    error,
+    isLoading,
+    isSuccess,
+    isError,
+  } = useAnalyticsBucketsQuery({ projectRef: ref }, { enabled: hasIcebergEnabled })
+
+  const bucket = data.find((x) => x.id === bucketId)
+
+  return { data: bucket, isLoading, isSuccess, isError, error }
```
alaister: If possible we should be returning the query result directly instead of returning a new object. That way the react-query typing magic will continue to work.

You can get the same result by using `select` in the react query options to do the `data.find((x) => x.id === bucketId)`.

alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/interfaces/Storage/AnalyticsBuckets/useSelectedAnalyticsBucket.ts
joshenlim: > You can get the same result by using select in the react query options to do the data.find((x) => x.id === bucketId).

oooh TIL will take a look in a bit!

PR #40437, opened 2025-11-13
Fix: await for profile creation on error function

awaseem:
Potentially fixes user creation redirects to new orgs. Awaiting testing

### DIFF
--- a/apps/studio/lib/profile.tsx
+++ b/apps/studio/lib/profile.tsx
@@ -77,7 +77,7 @@ export const ProfileProvider = ({ children }: PropsWithChildren<{}>) => {
     isSuccess,
   } = useProfileQuery({
     enabled: isLoggedIn,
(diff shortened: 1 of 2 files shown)


apps/studio/lib/profile.tsx
```diff
@@ -77,18 +77,19 @@ export const ProfileProvider = ({ children }: PropsWithChildren<{}>) => {
     isSuccess,
   } = useProfileQuery({
     enabled: isLoggedIn,
-    onError(err) {
+    async onError(err) {
       // if the user does not yet exist, create a profile for them
       if (err.message === "User's profile not found") {
-        createProfile()
+        await createProfile()
```
ivasilov: You can revert this, no need for it.

awaseem: ack

PR #40439, opened 2025-11-13
Fix: dropdown positioning for OrganizationProjectSelector

awaseem:
Fixes FE-2114

The 'All Projects' dropdown was appearing at the top of the page instead of below the trigger button. This was caused by the dropdown being rendered without a portal while inside a sticky positioned container.

Solution: Added portal={true} to PopoverContent_Shadcn_ to render the dropdown in a portal at the document root, ensuring correct positioning regardless of parent container CSS.


joshenlim (approved):
Verified changes on preview and also double checked on other areas where we're using OrganizationProjectSelector

PR #40445, opened 2025-11-13
Fix: storage throttle issues when uploading large number of files

awaseem:

YES

There was throttle exceptions that would occur when uploading a large number of files to storage. The reason is we fetched an API key per file, this causes 2 requests per file reaching the throttle limit faster. We now request a single key and track expiry internally

Trying to upload large number of files can run into a throttle exception

## What is the new behavior?

There should be fewer throttle exceptions due to the less number of API calls

## Additional context

How to test:
- [ ] Check to upload files to storage, the larger the better
- [ ] Check self hosted to ensure nothing is broken

### DIFF
--- a/apps/studio/data/api-keys/temp-api-keys-utils.test.ts
+++ b/apps/studio/data/api-keys/temp-api-keys-utils.test.ts
@@ -1,0 +1,298 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import {
+  createTemporaryUploadKey,
+  isTemporaryUploadKeyValid,
+  type TemporaryUploadKey,
+} from './temp-api-keys-utils'
+
+describe('createTemporaryUploadKey', () => {
+  beforeEach(() => {
+    vi.useFakeTimers()
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+  })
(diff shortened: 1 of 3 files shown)


apps/studio/data/api-keys/temp-api-keys-utils.test.ts
```diff
@@ -0,0 +1,298 @@
...
+    // Advance time by 59 seconds (should still be valid - 61 seconds remaining)
+    vi.advanceTimersByTime(59000)
+    expect(isTemporaryUploadKeyValid(key)).toBe(true)
+
+    // Advance time by 2 more seconds (should be invalid - 59 seconds remaining)
+    vi.advanceTimersByTime(2000)
+    expect(isTemporaryUploadKeyValid(key)).toBe(false)
+  })
+
+  it('should return false for a key missing apiKey property', () => {
```
charislam: The description is the opposite of the assertion here

charislam (approved):

apps/studio/data/api-keys/temp-api-keys-utils.test.ts
awaseem: damn AI slop!

PR #40446, opened 2025-11-13
fix(table editor): unnecessary column updates

charislam:
When you use the table editor to edit a table containing array columns, even if you _make no changes_ to the column, an update is sent to alter the array column's type to its original type. This happens because we aren't correctly checking for array column equality when we diff old/new column configs. Fixed this so no update will be sent if the column isn't changed.

## Testing
- [x] Test that actual column updates will work as expected
- [x] Test that not updating an array column will not fire a column update request

Resolves FE-2066

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
@@ -142,7 +142,11 @@ export const generateUpdateColumnPayload = (
     payload.check = check
   }
 
-  if (!isEqual(originalColumn.format, type)) {
+  const originalFormat =
+    originalColumn.data_type === 'ARRAY'
+      ? `${originalColumn.format.replace(/^_/, '')}[]`
+      : originalColumn.format
+  if (!isEqual(originalFormat, type)) {
     payload.type = type
   }
   if (!isEqual(originalColumn.default_value, field.defaultValue)) {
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
```diff
@@ -142,7 +142,11 @@ export const generateUpdateColumnPayload = (
     payload.check = check
   }
 
-  if (!isEqual(originalColumn.format, type)) {
+  const originalFormat =
+    originalColumn.data_type === 'ARRAY'
+      ? `${originalColumn.format.replace(/^_/, '')}[]`
+      : originalColumn.format
+  if (!isEqual(originalFormat, type)) {
```
joshenlim: tiny tiny nit but we can probably just use `===` since these are all just strings - was old code leaning too much into lodash back then 😬 

i'll push up a refactor for this and the other parts in this function 🙏

joshenlim (approved):
Verified changes on staging as per PR instructions

I did find one unrelated bug though
- Edit table -> Change one of the column's name -> Save -> Everything's alright
- Edit table again -> Change the column's name again -> Save -> There's an error
image

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnEditor.utils.ts
```diff
@@ -1,5 +1,5 @@
 import type { PostgresColumn } from '@supabase/postgres-meta'
-import { isEqual, isNull } from 'lodash'
```
awaseem, line 2: UMMM THANK YOU! Less lodash the better IMO

awaseem (approved):
LGTM tested on staging and self hosted. Works fine

awaseem:
> Verified changes on staging as per PR instructions
> 
> I did find one unrelated bug though
> 
>     * Edit table -> Change one of the column's name -> Save -> Everything's alright
> 
>     * Edit table again -> Change the column's name again -> Save -> There's an error
> 
> 
> image

I tested this and this doesn't happen anymore

PR #40454, opened 2025-11-14
fix(studio): table row focus and hover

dnywh:

Fix for Safari

- Only the last table row across all three bucket types (files, analytics, vectors) are selectable by mouse
- No table rows are selectable by keyboard

## What is the new behavior?

- Fixes for the above whilst maintaining existing functionality in Chrome

## Additional context

I believe https://github.com/supabase/supabase/pull/40230/files introduced this bug with the `absolute` `button`.

## To test

Please use (at least) both Safari and Chrome to test changes.

### DIFF
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/CreateAnalyticsBucketModal.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/CreateAnalyticsBucketModal.tsx
@@ -125,6 +125,8 @@ export const CreateAnalyticsBucketModal = ({
 
   const config = BUCKET_TYPES['analytics']
   const isCreating = isEnablingExtension || isCreatingIcebergWrapper || isCreatingAnalyticsBucket
+  const isDisabled =
+    !canCreateBuckets || !icebergCatalogEnabled || isLoading || buckets.length >= 2 || disabled
 
   const form = useForm<CreateAnalyticsBucketForm>({
     resolver: zodResolver(FormSchema),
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/Storage/AnalyticsBuckets/CreateAnalyticsBucketModal.tsx
```diff
@@ -199,6 +199,15 @@ export const CreateAnalyticsBucketModal = ({
...
           }
+          tabIndex={
+            !canCreateBuckets ||
+            !icebergCatalogEnabled ||
+            isLoading ||
+            buckets.length >= 2 ||
+            disabled
+              ? -1
+              : 0
+          }
```
dnywh: This just repeats the same checks as the `disabled` prop above. We should combine the two into a variable.

apps/studio/components/interfaces/Storage/VectorBuckets/index.tsx
```diff
@@ -157,18 +157,23 @@ export const VectorsBuckets = () => {
...
+                        className={cn('relative cursor-pointer h-16', 'inset-focus')}
+                        onClick={(event) => handleBucketNavigation(name, event)}
+                        onKeyDown={(event) => {
+                          if (event.key === 'Enter' || event.key === ' ') {
+                            event.preventDefault()
+                            handleBucketNavigation(name, event)
+                          }
+                        }}
+                        tabIndex={0}
+                      >
```
dnywh, line 170: Removing the absolutely positioned button inside a table cell and making the row itself interactive.

apps/studio/components/interfaces/Storage/VectorBuckets/index.tsx
```diff
@@ -157,18 +157,23 @@ export const VectorsBuckets = () => {
...
                         <TableCell className="w-2 pr-1">
                           <BucketIcon size={16} className="text-foreground-muted" />
                         </TableCell>
                         <TableCell>
                           <p className="whitespace-nowrap max-w-[512px] truncate">{name}</p>
-                          <button
-                            className={cn('absolute inset-0', 'inset-focus')}
```
dnywh, line 167: I think this absolutely positioned button was causing Safari stacking context issues.

apps/studio/components/interfaces/Storage/AnalyticsBuckets/index.tsx
```diff
@@ -163,18 +163,23 @@ export const AnalyticsBuckets = () => {
...
+                      key={bucket.id}
+                      className={cn('relative cursor-pointer h-16', 'inset-focus')}
+                      onClick={(event) => handleBucketNavigation(bucket.id, event)}
+                      onKeyDown={(event) => {
+                        if (event.key === 'Enter' || event.key === ' ') {
+                          event.preventDefault()
+                          handleBucketNavigation(bucket.id, event)
+                        }
+                      }}
+                      tabIndex={0}
```
dnywh, line 175: I believe this is necessary for Safari.

joshenlim (approved):
Verified changes on preview for both chrome and safari!

PR #40456, opened 2025-11-14
chore(studio): filter policies by bucket name

dnywh:

- Chore
- UI improvement

- Each files bucket page has a button (link) titled `Policies (n)` where `n` is the count of policies for that bucket
- Tapping that button link takes the user to a higher-level route `storage/files/policies`
	- But this page is general to files; it does not not focus on that particular bucket’s policies
	- Meaning it can be disorientating

## What is the new behavior?

- Tapping a bucket’s `Policies (n)` button  takes the user to `storage/files/policies?search=bucketname`
- A search (filter) input has been added to this page, and it watches the URL params
- This means that we pre-filter the page to any bucket names that match the search query

## Additional context

- This approach two flaws:
	- Many bucket names may use the same syntax, meaning a search could still show many buckets
		- 	E.g. a bucket named `test` with sibling buckets `test1`, `test2`, etc
	- Schema policies is further buried down the bottom of the Policies page, since having search means we should show all buckets, even those without policies
		- This is especially problematic for the few users with more than a handful of buckets
- That all said, this is a half-step towards something like DEPR-186: policies inside the bucket page itself
	- It’s a low-lift improvement over what we have today


apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx
```diff
@@ -98,7 +98,11 @@ const BucketPage: NextPageWithLayout = () => {
                 ) : undefined
               }
             >
-              <Link href={`/project/${ref}/storage/files/policies`}>Policies</Link>
+              <Link
+                href={`/project/${ref}/storage/files/policies?search=${encodeURIComponent(bucket.name)}`}
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx]: Set the bucket name as a URL param, much like we do in Table Editor → Policies

apps/studio/components/interfaces/Storage/StoragePolicies/StoragePolicies.tsx
```diff
@@ -78,6 +88,50 @@ export const StoragePolicies = () => {
   // Policies under storage.buckets
   const storageBucketPolicies = filter(policies, { table: 'buckets' })
 
+  /**
+   * Filter buckets based on search string
+   * - Filter buckets by name matching the search string
+   * - Show all policies for filtered buckets (policies are not filtered)
```
dnywh, line 96: As in, don’t take the policy name as a search query. Figured bucket name is enough to go from.

apps/studio/components/interfaces/Storage/StoragePolicies/StoragePolicies.tsx
```diff
@@ -78,6 +88,50 @@ export const StoragePolicies = () => {
...
+        filteredUngroupedPolicies: ungroupedPolicies,
+        filteredBucketPolicies: storageBucketPolicies,
+      }
+    }, [
+      data,
+      deferredSearchString,
+      formattedStorageObjectPolicies,
+      ungroupedPolicies,
+      storageBucketPolicies,
+    ])
```
dnywh: May get pushed down if the list of buckets (with their policies) is long. But I think it’s worth the UI price, given the vast majority of users have only 1–3 buckets.

apps/studio/components/interfaces/Storage/StoragePolicies/StoragePolicies.tsx
```diff
@@ -194,22 +248,44 @@ export const StoragePolicies = () => {
...
+            )}
+
             {/* Sections for policies grouped by buckets */}
             <div className="flex flex-col gap-y-4">
-              {buckets.map((bucket) => {
-                const bucketPolicies = get(
-                  find(formattedStorageObjectPolicies, { name: bucket.name }),
-                  ['policies'],
-                  []
-                ).sort((a: any, b: any) => a.name.localeCompare(b.name))
```
dnywh, line 204: Change worth highlighting:
- This prior version used to *only* show buckets that have policies
- The new version now shows all buckets

I made this change because:
- It matches how we do things on the main (table) policies page
- I find it strange to _not_ show all buckets here
	- E.g. what if I want to add a policy?

apps/studio/components/interfaces/Storage/StoragePolicies/StoragePolicies.tsx
```diff
@@ -26,6 +30,12 @@ export const StoragePolicies = () => {
   const { data: project } = useSelectedProjectQuery()
   const { ref: projectRef } = useParams()
 
+  const [searchString, setSearchString] = useQueryState(
+    'search',
+    parseAsString.withDefault('').withOptions({ history: 'replace', clearOnDefault: true })
+  )
+  const deferredSearchString = useDeferredValue(searchString)
```
joshenlim: i've always wondered how to use this hook actually but never properly understood it
initially thought that it could be a replacement for debouncing but it doesn't seem to work that way

apps/studio/components/interfaces/Storage/StoragePolicies/StoragePolicies.tsx
```diff
@@ -78,6 +88,50 @@ export const StoragePolicies = () => {
   // Policies under storage.buckets
   const storageBucketPolicies = filter(policies, { table: 'buckets' })
 
+  /**
+   * Filter buckets based on search string
+   * - Filter buckets by name matching the search string
+   * - Show all policies for filtered buckets (policies are not filtered)
+   */
+  const { filteredBucketsWithPolicies, filteredUngroupedPolicies, filteredBucketPolicies } =
```
joshenlim: im not too sure i follow here but this feels like its unncessarily returning `filteredUngroupedPolicies` and `filteredBucketPolicies`?

given that `filteredUngroupedPolicies` is just `ungroupedPolicies`
and `filteredBucketPolicies` is just `storageBucketPolicies`

only `filteredBucketsWithPolicies` is being computed within the useMemo

apps/studio/components/interfaces/Storage/StoragePolicies/StoragePolicies.tsx
joshenlim: > Prior version used to only show buckets that have policies

Hmm I don't think that's the current behaviour though - checking on staging for a bucket that has no policies it does show up on the policies page

image

apps/studio/components/interfaces/Storage/StoragePolicies/StoragePolicies.tsx
joshenlim: opting to remove this for now - from what i understand useDeferredValue needs to be used together with suspense so no diff atm

we'll eventually also have to do pagination for buckets so we'll need to use debounce then too

joshenlim (approved):

PR #40471, opened 2025-11-14
fix(bucket table): a11y

charislam:
Fix a11y so screen readers can navigate to bucket details from the bucket overview table


awaseem (approved):
LGTM

PR #40473, opened 2025-11-14
feat(mgmt-api): update reports API to use multiple attributes

raulb:

YES

Improve the performance of the reports page by using a new endpoint that aggregates the different attributes.

We're querying the separate attributes with separate infra-monitoring queries, which sometimes results in a "Too many requests" error from infra.

image

## What is the new behavior?

Uses a new endpoint that aggregates multiple attributes reducing the number of queries for those scenarios and reduce load impact when accessing the reports page.

## Additional context

Depends on https://github.com/supabase/infrastructure/pull/27519


raulb:
Added the `do-not-merge` to indicate that's currently blocked by https://github.com/supabase/infrastructure/pull/27519

awaseem:
Going to move this to draft, once the other PR is merged feel free to re-open again

raulb:
@awaseem API PR [shipped](https://github.com/supabase/infrastructure/pull/27519). It could be tested in staging soon, and in production hopefully tomorrow.  I could use a hand with the E2E tests in the meantime, but I'll try to get into those as soon as I can.

awaseem:
@raulb I have a PR for e2e to fix them because they are so flaky. I'll get that merged and then we can take a look here

apps/studio/data/analytics/infra-monitoring-queries.ts
```diff
@@ -12,18 +30,83 @@ export function useInfraMonitoringQueries(
...
+
+    acc[attribute] = {
+      data: dataPoints,
+      format: metadata.format,
+      total: metadata.total,
+      yAxisLimit: metadata.yAxisLimit,
+    }
+
+    return acc
+  }, {})
```
awaseem: Thoughts on writing tests around this method alone? Might be useful to bake the contract in

awaseem (approved):
Looks good on preview, I would just add a test if you can

apps/studio/data/analytics/infra-monitoring-queries.ts
raulb: 710f5e0c3c29721bae2ec7ca999a1cb9b9cab1df

PR #40479, opened 2025-11-14
Fix: Make self hosted tests more reliable locally and in CI

awaseem:
- Blacksmith runner, giving more resources to playwright headless
- Updated to run tests against a new instance of Supabase rather than the preview env
- Used community playwright report PR comment
- Single command in Root to run tests `pnpm run e2e`


.github/workflows/studio-e2e-test.yml
```diff
@@ -27,28 +27,12 @@ permissions:
 jobs:
   test:
     timeout-minutes: 60
-    runs-on: ubuntu-latest
+    runs-on: blacksmith-4vcpu-ubuntu-2404
     # Require approval only for pull requests from forks
     environment: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork && 'Studio E2E Tests' || '' }}
 
-    env:
```
awaseem, line 34: No more testing against Vercel, we run a docker container of Supabase and build a production env with Next js and test against that.

.github/workflows/studio-e2e-test.yml
```diff
@@ -62,36 +46,11 @@ jobs:
...
-          DEPLOY_URL=$(vercel deploy --prebuilt --token=${{ secrets.VERCEL_TOKEN }})
-          echo "Vercel Preview URL: $DEPLOY_URL"
-          echo "DEPLOY_URL=$DEPLOY_URL" >> $GITHUB_OUTPUT
-
       - name: Install Playwright Browsers
         run: pnpm -C e2e/studio exec playwright install chromium --with-deps --only-shell
 
       - name: 🚀 Run Playwright tests against Vercel Preview
         id: playwright
-        continue-on-error: true
```
awaseem, line 89: This was causing the tests output to not show :(

apps/studio/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "dev": "next dev --turbopack -p 8082",
-    "build": "next build && ./../../scripts/upload-static-assets.sh",
+    "build": "next build && if [ \"$SKIP_ASSET_UPLOAD\" != \"1\" ]; then ./../../scripts/upload-static-assets.sh; fi",
```
awaseem, line 8: @ivasilov Added this to skip the asset upload, rather than a new script. Thoughts?

e2e/studio/features/database.spec.ts
```diff
@@ -185,15 +185,23 @@ test.describe.serial('Database', () => {
```
awaseem, line 185: This helps with the flakyness

e2e/studio/playwright.config.ts
```diff
@@ -61,5 +56,9 @@ export default defineConfig({
     ['html', { open: 'never' }],
     ['json', { outputFile: 'test-results/test-results.json' }],
   ],
-  webServer: webServerConfig,
+  webServer: {
+    command: 'pnpm -w run e2e:setup',
+    port: WEB_SERVER_PORT,
+    timeout: WEB_SERVER_TIMEOUT,
+  },
```
awaseem, line 63: When the tests run, we start a server as a pre-req. This will listen to a server on port 8082 and start. Timesout in 10 mins

scripts/generateLocalEnv.js
```diff
@@ -40,3 +40,21 @@ fs.writeFileSync(
...
+  WEB_SERVER_PORT,
+  IS_PLATFORM: 'false',
+}
+
+fs.writeFileSync(
+  './e2e/studio/.env.local',
+  Object.keys(e2eTestEnv)
+    .map((key) => `${key}=${e2eTestEnv[key]}`)
+    .join('\n')
+)
```
awaseem, line 60: Instead of copying one, we generate the API and host for the tests when studio is finished running

package.json
```diff
@@ -29,8 +29,10 @@
     "test:ui-patterns": "turbo run test --filter=ui-patterns",
     "test:studio": "turbo run test --filter=studio",
     "test:studio:watch": "turbo run test --filter=studio -- watch",
-    "e2e:setup": "pnpm setup:cli && NODE_ENV=test NODE_OPTIONS=\"--max-old-space-size=4096\" pnpm run build:studio && NODE_ENV=test pnpm --prefix ./apps/studio start --port 8082",
+    "e2e:setup:cli": "supabase start -x studio && supabase db reset && supabase status --output json > keys.json && node scripts/generateLocalEnv.js",
+    "e2e:setup": "SKIP_ASSET_UPLOAD=1 pnpm e2e:setup:cli && NODE_ENV=test NODE_OPTIONS=\"--max-old-space-size=4096\" pnpm run build:studio && NODE_ENV=test pnpm --prefix ./apps/studio start --port 8082",
     "e2e": "pnpm --prefix e2e/studio run e2e",
```
awaseem, line 34: The only annoying part of this is that if you have an instance running already, these will fail. So you need to stop supabase before running

awaseem:
These tests are failing because of: https://github.com/supabase/supabase/pull/40530

.github/workflows/studio-e2e-test.yml
```diff
@@ -103,89 +62,12 @@ jobs:
...
-            } catch { s.note='No JSON report found or parse error.' }
-            // Generate the markdown for the summary comment.
-            const md = s.note ? `Note: ${s.note}` : `- Total: ${s.total}\n- Passed: ${s.passed||0}\n- Failed: ${s.failed||0}\n- Skipped: ${s.skipped||0}\n- Timed out: ${s.timedOut||0}\n- Interrupted: ${s.interrupted||0}\n- Flaky: ${s.flaky||0}\n- Duration: ${(s.durationMs/1000).toFixed(1)}s`
-            // Set the summary and flaky_count as outputs for subsequent steps.
-            core.setOutput('summary', md)
-            core.setOutput('flaky_count', s.flaky)
-
-      - name: Comment summary on PR
+      - name: Comment Playwright test results on PR
+        uses: daun/playwright-report-comment@5b72497e81ed11f0d0d7f1d375ef0a0c76d5e6fb # v3.10.0
```
charislam:
```suggestion
        uses: daun/playwright-report-comment@be9e270edd5ad86038604d3caa84a819a6ff6fed # v3.10.0
```

minor nit (I don't care which we use :D) but that's the tip of main, 3.10.0 is this one

.github/workflows/studio-e2e-test.yml
```diff
@@ -103,89 +62,12 @@ jobs:
...
         if: always() && github.event_name == 'pull_request'
```
charislam: This will fail on external PRs because the GitHub token for `pull_request` doesn't have write permissions. Easiest thing is just to check for external PRs and skip the comment there. If you really, really want to comment on external PRs, you need a two-step action where the second step runs in a workflow with elevated permissions. See this example from docs:

- https://github.com/supabase/supabase/blob/master/.github/workflows/docs-lint-v2.yml
- https://github.com/supabase/supabase/blob/master/.github/workflows/docs-lint-v2-comment.yml

e2e/studio/playwright.config.ts
```diff
@@ -61,5 +56,9 @@ export default defineConfig({
     ['html', { open: 'never' }],
     ['json', { outputFile: 'test-results/test-results.json' }],
   ],
-  webServer: webServerConfig,
+  webServer: {
+    command: 'pnpm -w run e2e:setup',
```
charislam:
```suggestion
    command: 'pnpm --workspace-root run e2e:setup',
```

Tiny nit, I have a personal preference for long-form flags in committed code b/c it makes it easier to understand what's going on later. Feel free to ignore if you disagree :D

package.json
```diff
@@ -29,8 +29,10 @@
     "test:ui-patterns": "turbo run test --filter=ui-patterns",
     "test:studio": "turbo run test --filter=studio",
     "test:studio:watch": "turbo run test --filter=studio -- watch",
-    "e2e:setup": "pnpm setup:cli && NODE_ENV=test NODE_OPTIONS=\"--max-old-space-size=4096\" pnpm run build:studio && NODE_ENV=test pnpm --prefix ./apps/studio start --port 8082",
+    "e2e:setup:cli": "supabase start -x studio && supabase db reset && supabase status --output json > keys.json && node scripts/generateLocalEnv.js",
```
charislam:
```suggestion
    "e2e:setup:cli": "supabase start --exclude studio && supabase db reset && supabase status --output json > keys.json && node scripts/generateLocalEnv.js",
```

same here, take it or leave it :D

package.json
```diff
@@ -29,8 +29,10 @@
...
+    "e2e:setup": "SKIP_ASSET_UPLOAD=1 pnpm e2e:setup:cli && NODE_ENV=test NODE_OPTIONS=\"--max-old-space-size=4096\" pnpm run build:studio && NODE_ENV=test pnpm --prefix ./apps/studio start --port 8082",
```
charislam, line 33: Is it possible to run on a different port? It would be nice not to have to stop the normal studio dev server to run E2E tests. I know it's not possible with the platform version because the auth callbacks expect the studio to be running on 8082, but I can't think of whether there's a similar constraint for self-hosted

charislam (approved):
This is so exciting 💃🏼 🎊

.github/workflows/studio-e2e-test.yml
awaseem: Oh that is a good point, totally forgot about that! We can just skip this on external PRs

package.json
awaseem: Would we also need to run supabase cli on different ports right? They might cause collisions with everything else running. I can explore this in a following PR

PR #40509, opened 2025-11-17
Database page components

SaxonF:
Swaps out the old PageLayout and Scaffold components with the new PageHeader, PageSection, PageContainer components on project/[ref]/database pages.

This also updates PageHeader component to make use of context to pass down size rather than needing to set on children.

To test:
Visit all database related pages and ensure layouts are not broken.


awaseem:
@SaxonF you have some merge conflicts to resolve

charislam (approved):

PR #40513, opened 2025-11-17
Joshen/fe 2141 swap organization settings page to use paginated projects

joshenlim:
## Context

Last piece to the puzzle of deprecating the usage of the old unpaginated projects endpoint by updating the Organization Team settings page to use the new V2 organization roles endpoints which now returns project refs instead of IDs

## Changes involved
- Use organization roles V2 endpoint, and fully move off usage of v1 projects endpoint
  - Deprecate useProjectsQuery
- Removed UI roles obfuscation logic of hiding roles of other users
  - Given that currently users with project-scoped roles cannot see other members in the organization
    image
  - Was previously there before ^ that was a thing, so we'd only show roles of other users that have access to the same projects as the account's user
- If you have a project-scoped role, you should not be able to see pending invites in organization team page
  - This is the current behaviour but I realised this is conflicting with the fact that project-scoped users cannot see other members in the organization
  - Hence decided to hide this, although I'm checking with API to handle this from the endpoint itself
- Refactors to reduce reliance on paginated org projects endpoint as much as possible
  - We're mainly using the org projects endpoint in some places to check if a project exists based on the role (I can't remember if its already been resolved by BE doesn't track if the ref of a role is valid, e.g if the project got deleted)
  - So there's a chance that the checks might fall short if the org has > 100 projects
- UX improvement for "This member has only access to x projects" callout to make it into a collapsible
  image
  - Realised original UI can get annoying if I'm trying to remove access for multiple projects as the callout grows in height as I remove more projects
  - So I'm not able to just click in one spot to remove multiple projects
- Removed unused query keys `orgProjects` and `readonlyStatusList` under `projectKeys`
- (Unrelated) Add error-handling for fetching configcat flags so client doesn't error out if the request to fetch flags fail

## To test
The unit tests are all passing for this part of the UI which gives me some confidence but reckon still best to go crazy with some stress testing
- [ ] Verify can change org roles of users for Free / Pro plan
- [ ] Verify can change project roles of users for Team / Enterprise plan
  - Go crazy with the changes, remove, add, update, multiple times - let me know if anything breaks!


apps/studio/components/interfaces/Organization/TeamSettings/UpdateRolesPanel/UpdateRolesPanel.tsx
```diff
@@ -167,13 +156,11 @@ export const UpdateRolesPanel = ({ visible, member, onClose }: UpdateRolesPanelP
 
   useEffect(() => {
     if (visible && isSuccessRoles) {
-      const roleConfiguration = formatMemberRoleToProjectRoleConfiguration(
-        member,
-        allRoles,
-        projects ?? []
-      )
+      const roleConfiguration = formatMemberRoleToProjectRoleConfiguration(member, allRoles)
+      console.log(roleConfiguration)
```
alaister: Sneaky one!
```suggestion
```

alaister (approved):
LGTM!

PR #40516, opened 2025-11-17
Fix fetching analytics bucket in ETL

joshenlim:
## Context

Storage endpoint has updated to no longer return analytics buckets in the /buckets endpoint - just updating to use GET /analytics-buckets endpoint instead


alaister (approved):
LGTM!

- Tested on preview

PR #40518, opened 2025-11-17
feat(studio): add comparisons to query perf calls filter

kemaldotearth:

YES

This replaces the minimum call input on Query Performance and brings in the numeric filter with comparisons (e.g. less than, greater than, equals etc).


awaseem (approved):
LGTM works on self hosted and preview

PR #40521, opened 2025-11-17
Chore/add missing error handler for buckets

joshenlim:
## Context

Realised that the bucket home pages for files, analytics and vectors were missing error state handlers - more importantly for files buckets. There's currently a footgun whereby if you were to disable legacy API keys without having created a new set of API keys, the Storage UI would silently fail (since it depends on the API keys on the BE)

PR here adds error state handlers for all 3 types of buckets, and also contextually renders this error for files buckets if the error returned is about missing API keys

image

## To test

Just a code review will do as this error doesn't surface on local/staging, only production so its not possible to reproduce in the preview. Although please do a quick check to verify everything else tho (loading state, success states)


saltcod (approved):
Code looks good.

PR #40523, opened 2025-11-17
Set up feature for ETL + Analytics Buckets private alpha

joshenlim:
## Context

[Updated] 
Sets up the following feature flags for ETL + Analytics Buckets private alpha. Feature flag will hold a list of string values that correspond to org slugs. Note that the flag will mainly be for ETL actually, Analytics Buckets will rely on storage config (as per current behaviour)

- Analytics Buckets (No changes here)
  - Will rely on storage config to verify if iceberg catalog is enabled (This is controlled by Storage's own feature flag)
- ETL (using `etlPrivateAlpha` flag)
  - Show ETL UI organization is within the flag
  - We'll also have a flag `etlPrivateAlphaOverride` that will act as a truthy override, if true will enable for all projects
  - Mainly to make it convenient for staging (So we don't need to manually opt in all projects) + easier transition to public alpha (code change independent)

The respective "Coming soon" UI will render if the project doesn't have access to the private alpha

## To test

Probably a code review is sufficient for this one as I've enabled the override for staging (so all staging projects has access to ETL + Analytics Buckets as per existing behaviour)

For prod - `etlPrivateAlpha` has been set to `none` (for now) while `etlPrivateAlphaOverride` is `false`


alaister (approved):
LGTM!

PR #40536, opened 2025-11-17
fix(export all rows): use cursor pagination if possible

charislam:
Exporting all rows (in CSV, SQL, or JSON format) currently uses offset pagination, which can cause performance problems if the table is large. There is also a correctness problem if the table is being actively updated as the export happens, because the relative row offsets could shift between queries.

Now that composite filters are available in `postgres-meta`, we can change to using cursor pagination on the primary key (or any non-null unique keys) wherever possible. Where this is not possible, the user will be shown a confirmation dialog explaining the possible performance impact.

### Centralizing export logic

Export logic is now centralized in `TableEditor/ExportAllRows.tsx`:
- `useExportAllRowsGeneric` contains all the logic for fetching rows from the database if necessary
- `useExportAllRowsAsCsv`, `useExportAllRowsAsSql`, and `useExportAllRowsAsJson` build upon `useExportAllRowsGeneric`, passing in callbacks to customize the output behavior

### Postgres query and type updates to preserve primary key and unique index information

To cursor paginate by a primary key/non-null unique index, we need to have information on the index and its ordering.
- This data is now retrieved in the `table-editor` query: primary keys used to be retrieved, but the ordering of their columns (important for establishing which sort order can use the index) was not guaranteed, and unique indexes used to not be retrieved at all.
- The data also needs to survive post-conversion. Two fields were added to the `SupaTable` interface to preserve it: `primaryKey` and `uniqueIndexes`.

### Cursor pagination

Cursor pagination logic is in `table-rows-query.ts`:
- `getDefaultOrderByColumn` is now `getPreferredOrderByColumns`, which distinguishes between sorts that can use cursor pagination and those that can't
- If `getPreferredOrderByColumns` returns a sort that can use cursor pagination, it is used in `getAllTableRowsSql`
- Because cursor pagination changes the query filter in every iteration, I also added a `clone` method to `QueryFilter` to prevent ending up with redundant filters like `(id > 50) and (id > 100) and (id > 150)` instead of just `(id > 150)`. 

### Format-to-SQL improvements

- Prevented problems with formatting arrays and JSON-containing arrays to SQL by:
  - Switching to `ARRAY[...]` syntax rather than `{...}`
  - Using safe dollar-quoting of JSON values
- This prevents problems around quote-escaping in complex nested objects.

Resolves FE-1912, FE-66


apps/studio/data/table-rows/table-rows-query.ts
```diff
@@ -27,15 +27,40 @@ export interface GetTableRowsArgs {
...
+    cursorPaginationEligible.push(primaryKeyColumns)
+  }
+
+  const uniqueIndexes = table.uniqueIndexes
+  const cursorFriendlyUniqueIndexes = uniqueIndexes?.filter((index) => {
+    index.every((columnName) => {
+      const column = table.columns.find((column) => column.name === columnName)
+      return !!column && !column.isNullable
+    })
+  })
```
avallete, line 54: **issue**

I think there is a bug here, this should return the result of every shouldn't it ? Otherwise it's always returning `undefined`.

```suggestion
  const cursorFriendlyUniqueIndexes = uniqueIndexes?.filter((index) => {
    return index.every((columnName) => {
      const column = table.columns.find((column) => column.name === columnName)
      return !!column && !column.isNullable
    })
  })
```

avallete (changes requested):

apps/studio/data/table-rows/table-rows-query.ts
charislam: 🤦🏼 Of course, thanks for the catch!

packages/pg-meta/src/query/QueryFilter.ts
```diff
@@ -44,6 +44,22 @@ export class QueryFilter implements IQueryFilter, IQueryModifier {
...
+        ? this.actionValue.map((val) => (typeof val === 'object' ? { ...val } : val))
+        : !!this.actionValue && typeof this.actionValue === 'object'
+          ? { ...this.actionValue }
+          : this.actionValue,
+      this.actionOptions ? { ...this.actionOptions } : undefined
+    )
+    cloned.filters = this.filters.slice()
+    cloned.sorts = this.sorts.slice()
+    return cloned
+  }
```
avallete, line 68: **nitpick**

I wonder if we could use `structuredClone` here ? https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone

```ts
clone(): QueryFilter {
  const clonedData = structuredClone({
    table: this.table,
    action: this.action,
    actionValue: this.actionValue,
    actionOptions: this.actionOptions,
    filters: this.filters,
    sorts: this.sorts,
  })

  const cloned = new QueryFilter(
    clonedData.table,
    clonedData.action,
    clonedData.actionValue,
    clonedData.actionOptions
  )

  cloned.filters = clonedData.filters
  cloned.sorts = clonedData.sorts

  return cloned
}
```

avallete (commented):
LGTM just a nitpick, leaving the approval to a FE team member.

packages/pg-meta/src/query/QueryFilter.ts
charislam: yeah, good call!

you'll need to approve because you're a codeowner on `pg-meta` so it's actually blocked on your specific approval 😅

soedirgo (approved):

avallete (approved):
Sorry for the delay 😅

apps/studio/components/interfaces/TableGridEditor/TableEntity.utils.ts
```diff
@@ -65,3 +66,60 @@ export const formatTableRowsToSQL = (table: SupaTable, rows: any[]) => {
...
+      throw new Error('Unable to generate a unique dollar-quote tag after multiple attempts.')
+    }
+
+    attempts++
+    tag = generateRandomTag()
+  }
+  return `${tag}${str}${tag}`
+}
+
+const formatArrayForSql = (arr: unknown[]): string => {
```
joshenlim, line 103: should add some unit tests for this

apps/studio/components/layouts/TableEditorLayout/ExportAllRows.progress.tsx
```diff
@@ -0,0 +1,104 @@
...
+      id: number
+      name: string
+      value: number
+      totalRows: number
+    }) => {
+      const savedToastId = toastIdsRef.current.get(id)
+
+      const progress = Math.min((value / totalRows) * 100, 100)
+      const newToastId = toast(
+        <SonnerProgress progress={progress} message={`Exporting ${name}...`} />,
```
joshenlim, line 51: am just wondering if there's a better way we can display the progress here if the table is really long

currently looks as though nothing might be happening since its only a spinner (even if we're constantly paging in the background)

apps/studio/components/layouts/TableEditorLayout/ExportAllRows.tsx
```diff
@@ -0,0 +1,431 @@
...
+        console.error(
+          `Export All Rows > Error: %s%s%s`,
+          error.message,
+          error.cause?.message ? `\n${error.cause.message}` : '',
+          error.cause?.stack ? `:\n${error.cause.stack}` : ''
+        )
+        return stopTrackerWithError(entity.id, entity.name)
+      }
+
+      if (exportResult.status === 'require_confirmation') {
```
joshenlim, line 328: confirmation modal shows up with progress toast in this case - reckon we only show the toast after confirming? I'm guessing its the same with the other statuses as well since `startTrackerProgress` is being called prior to the status checks

image

joshenlim: Running into an error exporting a view - we should also include the error message here in the toast i reckon
image

my SQL view definition is just `create view colors_view as select * from colors limit 2000;`

Error:
```Failed to run sql query: ERROR:  42703: column colors_view.ctid does not exist
LINE 1: select * from public.colors_view order by colors_view.id asc nulls last, colors_view.ctid asc nulls last limit 500 offset 0;
                                                                                 ^
HINT:  Perhaps you meant to reference the column "colors_view.id".
```

apps/studio/components/layouts/TableEditorLayout/ExportAllRows.tsx
```diff
@@ -0,0 +1,431 @@
...
+      <ConfirmationModal
+        title="Confirm export"
+        visible={true}
+        onCancel={onCancelExport}
+        onConfirm={onConfirmExport}
+      >
+        <div className="flex items-center gap-4">
+          <AlertTriangle aria-label="warning" className="shrink-0 text-foreground-lighter" />
+          <p className="text-sm text-foreground-lighter">{confirmationMessage}</p>
+        </div>
```
joshenlim: reckon we use an admonition here for consistency - ConfirmationModal here supports a `alert` param for this

apps/studio/components/layouts/TableEditorLayout/ExportAllRows.tsx
```diff
@@ -0,0 +1,431 @@
...
+  }
+
+  const type = table.entity_type
+  if (
+    (type === ENTITY_TYPE.VIEW || type === ENTITY_TYPE.MATERIALIZED_VIEW) &&
+    !bypassConfirmation
+  ) {
+    return {
+      status: 'require_confirmation',
+      reason: `You're exporting a view, which may cause performance issues on very large views. If possible, we recommend exporting the underlying table instead.`,
```
joshenlim: nit on copy - could saw "Exporting a view may cause..."

apps/studio/components/layouts/TableEditorLayout/ExportAllRows.tsx
joshenlim: Updating as such:
image

apps/studio/components/layouts/TableEditorLayout/ExportAllRows.tsx
joshenlim: will update!

apps/studio/components/layouts/TableEditorLayout/ExportAllRows.tsx
joshenlim, line 1: i'm not too sure which part of the logic this is, but JFYI with the this PR i noticed that this changes the behaviour of "Export all rows" in the table editor as we're no longer respecting the filters + sorts that are applied on the table

e.g say I have a countries table with 500 rows
- i have filter for `id < 200`
- i have a sort of `name` descending order

currently clicking "Select all rows in table" then exporting will download the full 500 rows nonetheless with default sort on PK column, rather than the 199 rows that i was expecting

is it still possible though to follow the filter and sorts? if not then perhaps we may need to consider this use case - e.g only use cursor pagination if no filter and sorts applied or something

apps/studio/components/interfaces/TableGridEditor/TableEntity.utils.ts
charislam: There are some tests for `formatTableRowsToSQL` that I've updated to incorporate this: https://github.com/supabase/supabase/blob/3bac0d3a92d1592c0b75fc3364b7953029c6bb38/apps/studio/components/interfaces/TableGridEditor/TableEntity.utils.test.ts#L87 I'd like to keep the testing there as that is the externally exposed function and this one is an internal detail

apps/studio/components/layouts/TableEditorLayout/ExportAllRows.progress.tsx
charislam: Hmm good call. The progress bar is shown if the export is initiated from the grid header, since it has access to the number of rows, but not if it's initiated from the sidebar. Let me see if it's easy to get the row information in there...

apps/studio/components/layouts/TableEditorLayout/ExportAllRows.tsx
charislam: This was totally my bad, I just forgot to pass the param through 🤦🏼

joshenlim (approved):

PR #40542, opened 2025-11-17
Fix: updated studio to only run on paths, otherwise shortcut out

awaseem:
Minor workflow changes to enable e2e


.github/workflows/studio-e2e-test.yml
```diff
@@ -33,28 +22,45 @@ jobs:
...
       - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+        id: filter
+        with:
+          filters: |
+            studio:
+              - 'packages/pg-meta/**'
+              - 'apps/studio/**'
+              - 'apps/ui-library/**'
+              - 'apps/design-system/**'
```
awaseem, line 33: Based on what happened with the form change, I added these because they can lead to unintended consequences

charislam (approved):

PR #40547, opened 2025-11-18
Shift view replication action to table level and enable search via URL params in ETL pipeline status page

joshenlim:
## Context

Realised that "View replication" CTA on Analytics Bucket page can be more granular as there's individual status for each table. 

## Changes involved

- Remove top level "View replication" button on the bucket
- Add "View replication" dropdrop menu item back to the "More" button of each table
  - Links to the pipeline status page with a search param of the inferred postgres' table
- Add support for search URL param in ETL pipeline status page

## To test

- [ ] Verify that the link works for an analytics bucket table that has replication


dnywh (approved):
Tested on staging. Confirming that the link works for an analytics bucket table that has replication!

PR #40549, opened 2025-11-18
Client side plan check for analytics and vector buckets

joshenlim:
## Context

Adds a plan check and call out for both analytics and vector buckets if project is not on a paid plan

image


alaister (approved):
LGTM!

- Tested on preview

PR #40553, opened 2025-11-18
feat: adds link conversation support form

barcofourie:

YES

Feature

Currently if a user created a support query by contacting via email, there is no way to link that conversation to a customer. 

## What is the new behavior?

Adds a simplified support form, to be able to link a conversation sent via email to an organization and user.

image


apps/studio/pages/support/link.tsx
```diff
@@ -0,0 +1,18 @@
+import type { ReactElement } from 'react'
+import { LinkSupportTicketPage } from 'components/interfaces/Support/LinkSupportTicketPage'
+import AppLayout from 'components/layouts/AppLayout/AppLayout'
+import DefaultLayout from 'components/layouts/DefaultLayout'
+import { withAuth } from 'hooks/misc/withAuth'
+import type { NextPageWithLayout } from 'types'
+
+const LinkSupportTicketPageRoute: NextPageWithLayout = () => {
```
awaseem, line 11: We should not render anything is this is self hosted. There is a env Variables called `IS_PLATFORM` and it should just either redirect to 404 or send a message saying not available

apps/studio/components/interfaces/Support/LinkSupportTicketPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+function LinkSupportTicketPageContent() {
+  const router = useRouter()
+  const conversationId = router.query.conversationId as string | undefined
+
+  if (!conversationId) {
+    return (
+      <Admonition
+        type="warning"
+        title="Missing conversation ID"
+        description="Please provide a conversation_id in the URL to link your support ticket."
```
awaseem: The query param is `conversationId` so this error message should match that

apps/studio/components/interfaces/Support/LinkSupportTicketForm.tsx
```diff
@@ -0,0 +1,360 @@
...
+      category: values.category,
+      allow_support_access: values.allowSupportAccess,
+    })
+  }
+
+  const formId = 'link-support-ticket-form'
+
+  return (
+    <Form_Shadcn_ {...form}>
+      <form id={formId} onSubmit={form.handleSubmit(onSubmit)} className="flex flex-col gap-y-6">
```
awaseem: Nit: why assign this to a variable? Lets just do 


```suggestion
  return (
    <Form_Shadcn_ {...form}>
      <form id={'link-support-ticket-form'} onSubmit={form.handleSubmit(onSubmit)} className="flex flex-col gap-y-6">
```

awaseem (changes requested):
Just updates and self hosted support

awaseem (approved):
LGTM

PR #40554, opened 2025-11-18
Add enable selection prop to FormLabel component

joshenlim:
## Context

Just realised that my previous fix [here](https://github.com/supabase/supabase/pull/40552) will break unit tests with the `DeleteBucketConfirmationModal`, and rightfully so as its technically not semantically correct using a `span` for a input label

Opting to do this properly within the ui packages instead by passing a `enableSelection` flag into the `FormLabel` component - which will then render a native `label` element instead of radix's label component and hence will support double clicking on the label to copy and paste

### DIFF
--- a/packages/ui/src/components/shadcn/ui/form.tsx
+++ b/packages/ui/src/components/shadcn/ui/form.tsx
@@ -13,9 +13,9 @@ import {
   useWatch,
 } from 'react-hook-form'
 
+import { AnimatePresence, motion } from 'framer-motion'
 import { cn } from '../../../lib/utils/cn'
 import { Label } from './label'
-import { motion, AnimatePresence } from 'framer-motion'
 
 const Form = FormProvider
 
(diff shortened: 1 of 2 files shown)


awaseem (approved):
LGTM

packages/ui/src/components/shadcn/ui/form.tsx
```diff
@@ -88,15 +88,17 @@ FormItem.displayName = 'FormItem'
 
 const FormLabel = React.forwardRef<
   React.ElementRef<typeof LabelPrimitive.Root>,
-  React.ComponentPropsWithoutRef<typeof LabelPrimitive.Root>
->(({ className, ...props }, ref) => {
+  React.ComponentPropsWithoutRef<typeof LabelPrimitive.Root> & { enableSelection?: boolean }
+>(({ className, enableSelection = false, ...props }, ref) => {
   const { error, formItemId } = useFormField()
 
+  const Comp = enableSelection ? 'label' : Label
```
MildTomato, line 95: requesting changes on this.
this looks wrong to me.

why are we using a raw label?

MildTomato (changes requested):

packages/ui/src/components/shadcn/ui/form.tsx
MildTomato: temp blocking the PR while investigating

MildTomato (approved):
approving.

we need to remove the hack after though

PR #40556, opened 2025-11-18
refactor: Send favorite as top level value for user content

lforst:
Aligns with what we would like to receive in the API

Resolves: https://linear.app/supabase/issue/FE-1885/update-all-user-content-endpoint-sql-snippets-to-use-dedicated


apps/studio/types/userContent.ts
```diff
@@ -113,28 +109,6 @@ export namespace Dashboards {
   }
 }
 
-export namespace SqlSnippets {
```
lforst, line 116: This was unused

joshenlim (approved):
Verified changes on staging preview - checked favoriting behaviour in SQL Editor

Issue #40301, opened 2025-11-10
[Bug] Supavisor / GoTrue Rejecting openssl-generated Keys (VAULT_ENC_KEY/JWT_SECRET) due to Format/Encoding Error - Error: Unknown cipher or invalid key size


luanlongen:
## Describe the problem

The current documentation instructs users to use a strong, random 32-character secret for the JWT_SECRET.

However, when self-hosting, there is a major confusion regarding the required key format and encoding, especially when setting up the initial services like GoTrue (Auth) and PostgREST.

The issue is a severe contradiction in key acceptance:

Failing Key (Strong): When generating a strong key using the standard cryptographic command (openssl rand -hex 32 - 64 hex characters) and placing it as the JWT_SECRET in the .env file, the stack fails to initialize the services correctly, often leading to postgres access.

Erorr: "Unknown cipher or invalid key size"

Working Key (Incorrect Format): Using an arbitrary —such as a full, valid JWT token copied from a generator like jwt.io—resolves the initialization errors and allows the stack to function.

This suggests that the self-hosted environment's services might not be correctly parsing or expecting a raw hexadecimal secret, which is highly misleading and promotes the use of cryptographically weaker or improperly formatted secrets.

## Describe the improvement

The documentation must clarify the exact format and required encoding (e.g., raw text, hexadecimal, or Base64) for the JWT_SECRET variable, and ensure that the key generation instructions align with the requirements of the services that consume this secret.

The section on "Generate API keys" should be updated to address:

Standardize Tool Use: Confirm if openssl rand -hex 32 should be used, and if so, how to ensure the 64-character output is read correctly (e.g., adding a troubleshooting note about using dos2unix .env to prevent CRLF corruption).

Strictly Forbid JWTs as Secrets: Explicitly warn users not to use a full JWT token (the string with dots) as the JWT_SECRET, as it is a security risk and an incorrect configuration for a secret key.

## Additional context

The core configuration should require a symmetric secret key for HMAC-SHA256, not a token that has already been signed. The current behavior on self-hosted makes it challenging to deploy with a cryptographically sound secret.

aantti:
Thanks for the detailed notes here - useful for my ongoing effort of working on the docs for self-hosted.

aantti:
I've recently updated the [guide](https://supabase.com/docs/guides/self-hosting/docker#generate-and-configure-api-keys). Hope it helps. In addition, I'll be adding an option to use newer auth options soon. Closing this one for now.

PR #34171, opened 2025-03-13
chore: fix for #VAULT_ENC_KEY# value configuration

ghyghoo8:
The comment for "your-encryption-key-32-chars-min" of "VAULT_ENC_KEY" is incorrect. Self-Hosting fails when the key exceeds 32 characters due to Supavisor startup failure.

## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

docs update

## What is the current behavior?

https://github.com/orgs/supabase/discussions/33119

## What is the new behavior?
-

## Additional context

Add any other context or screenshots.

PR #40558, opened 2025-11-18
change default VAULT_ENC_KEY environment variable docker/.env.example to clarify exact 32 character requirement

TerraTalpi:
Enhance default value of environment variable `VAULT_ENC_KEY` in [docker/.env.example line 13](https://github.com/supabase/supabase/blob/16fb243141302f5a1f14452b44f2f3d4e458d96e/docker/.env.example#L13). The current default value is the 32 character long string `your-encryption-key-32-chars-min`. However, when configuring a longer secret the supavisor service stops working. With this pull request, I want to set the value to a new exactly 32 character long value to continue immediate functionality upon copying the example file, while changing the hint that it must be exactly 32 characters long. This is trying to prevent cases and issues like this: #33119

YES

Changing [docker/.env.example#13](https://github.com/supabase/supabase/blob/16fb243141302f5a1f14452b44f2f3d4e458d96e/docker/.env.example#L13) from value `your-encryption-key-32-chars-min` to NEW `your-32-character-encryption-key`.

The default value of `VAULT_ENC_KEY` environment variable suggests that the variable must be minimum 32 characters long, but longer values break the `supavisor` service as discussed in #33119.

## What is the new behavior?

The new value of `VAULT_ENC_KEY` states that the encryption key must be 32 characters long.

## Additional context

I actually have two proposals that are exactly 32 characters long to keep a working default example configuration:

- `your-32-character-encryption-key` (submitted in the PR)
- `your-exactly-32-char-encrypt-key`

Of course there are more options, I just picked one that I found to be the most suitable. Feel free to propose other default values or ask me to use the second one and I will gladly do so.

This PR is kind of related to #34171 that has been closed. However, in that PR a longer than 32 character value for `VAULT_ENC_KEY` was proposed which would break default config / lab deployments if left as it is.

Furthermore, this issue is also related to #40301 because it addresses the same issue with the environment variable `VAULT_ENC_KEY`.


avallete (approved):
LGTM

aantti:
Thanks, @TerraTalpi - appreciate it!

aantti (approved):

PR #40573, opened 2025-11-18
chore: Migrate smaller chunks of ReactQuery code to be ReactQuery v5 compatible

ivasilov:


apps/studio/data/logs/unified-logs-infinite-query.ts
```diff
@@ -47,7 +47,7 @@ export const getUnifiedLogsISOStartEnd = (
 }
 
 export async function getUnifiedLogs(
-  { projectRef, search, pageParam }: UnifiedLogsVariables & { pageParam?: PageParam },
+  { projectRef, search, pageParam }: UnifiedLogsVariables & { pageParam: PageParam | null },
```
awaseem, line 50: Why did we change this? the optional type is better IMO

awaseem (approved):
Minor questions

PR #40592, opened 2025-11-18
Fix: Use storage API for downloading files in the Dashboard

awaseem:

YES

Before we used the management API to call the download file method. Due to recent throttles this doesn't work for larger files. Moved to storage API just like uploads

Downloading larger number of files causes exceptions

## What is the new behavior?

Files can be downloaded directly through storage API

## Additional context

- Added E2E tests for storage 

How to test (self hosted and platform)
- Download files from the preview
- Select multiple files and download
- Select folder and download
- Nested folder downloads


apps/studio/components/interfaces/Storage/FilesBuckets/BucketTable.tsx
```diff
@@ -102,6 +102,7 @@ export const BucketTableRow = ({
   return (
     <BucketTableRow
       key={bucket.id}
+      data-bucket-id={bucket.id}
```
awaseem, line 105: Way easier to find the bucket ID when set properly

apps/studio/state/storage-explorer.tsx
```diff
@@ -836,6 +834,9 @@ function createStorageExplorerState({
           { id: toastId, closeButton: false, position: 'top-right' }
         )
 
+        // Get authenticated Supabase client for Storage API access
+        const client = state.supabaseClient ? await state.supabaseClient() : undefined
```
ivasilov, line 822: If you're trying to download 10 files at once, it fetches 10 temp keys. 

You can try and work around this by using `useSupabaseClientQuery`, putting its result in `useLatest` and using that from the `supabaseClient` promise.

Small note, both approaches will probably break if you're downloading a private file and the download time exceeds the temp key expiry time, but that's probably fine.

awaseem: ~Thats not the case anymore. I fixed the API key by adding a temp cache that only fetches a new key after 10 minutes. This was done in #40445~

Scratch that I see what your saying, looking into it

awaseem:
This is still in progress, moving to draft

ivasilov (approved):

PR #40597, opened 2025-11-19
chore(studio): improve project upgrade experience

dnywh:

- UI updates
- Resolves DEPR-188

- Upgrading is ambiguous

## What is the new behavior?

- Brings the upgrade flow out in a review step
- Adds `backingUp` as a first-class status for clarity

## Additional context

The API returns a rounded version number. This makes it ambiguous as to which exact Postgres version a user chocse. We should add a `target_version_display` or `target_version_string` field to the upgrade status response that contains the full version string (e.g., "17.6.1.062"). This would:

- Remove the dependency on cached eligibility data
- Make the display more reliable
- Provide the exact version string that was used for the upgrade

The current fix looks up the full version from cached eligibility data, but if the cache is unavailable, it falls back to just the number.


apps/studio/components/interfaces/Home/ServiceStatus.tsx
```diff
@@ -51,7 +51,7 @@ const iconProps = {
 }
 const LoaderIcon = () => <Loader2 {...iconProps} className="animate-spin" />
 const AlertIcon = () => <AlertTriangle {...iconProps} />
-const CheckIcon = () => <CheckCircle2 {...iconProps} className="text-brand" />
+const CheckIcon = () => <CheckCircle {...iconProps} className="text-brand" />
```
dnywh, line 56: To align further with visual style in UpgradingState.

apps/studio/components/layouts/ProjectLayout/UpgradingState/index.tsx
```diff
@@ -129,9 +129,13 @@ export const UpgradingState = () => {
               <div className="grid w-[480px] gap-4">
                 <div className="relative mx-auto max-w-[300px]">
                   <div className="absolute flex items-center justify-center w-full h-full">
-                    <Settings className="animate-spin" size={20} strokeWidth={2} />
+                    <Settings
+                      className="animate-[spin_4s_linear_infinite]"
```
dnywh: Slow down or remove.

apps/studio/components/layouts/ProjectLayout/UpgradingState/index.tsx
```diff
@@ -80,7 +80,7 @@ export const UpgradingState = () => {
             {isCompleted ? (
               <div className="grid gap-4">
                 <div className="relative mx-auto max-w-[300px]">
-                  <CheckCircle className="text-brand" size={40} strokeWidth={1.5} />
+                  <CheckCircle className="text-brand-link" size={40} strokeWidth={1.5} />
```
dnywh: Should be light mode only.

packages/ui/src/components/shadcn/ui/badge.tsx
dnywh, line 1: I added `forwardRef` to the component because using it `asChild` on a Tooltip (for example) doesn’t work without it. I believe this is in-line with other Shadcn components so shouldn’t be anything too new.

apps/studio/components/interfaces/Settings/Infrastructure/InfrastructureInfo.tsx
dnywh, line 1: Fixes some visual alignment on the badges. Basically, _not_ having `asChild` wrapped each Badge component in a `button` element that caused wonky visual offsets.

PR #40623, opened 2025-11-19
fix(www): swiper arrows not functioning in products pages' examples

### LINKED ISSUE #40622: Arrow navigation buttons are not working in the examples' sections of Database, Auth, and Edge Functions pages.
Like the title says, sometimes the arrows in the examples section don't work

Image

I did some digging into the codebase, and I think I know the cause of the issue:

Swiper’s Navigation module needs real DOM elements in `prevEl` and `nextEl` at the moment the Swiper instance initializes.

Image

But the refs `prevRef.current` and `nextRef.current` are `null` on first render, and assigned after `Swiper` is initialized.

Image

The code attempts to patch the navigation after `init`, but since the refs may not yet be available when `onInit` fires, the navigation is sometimes set to `null`.

Image

marsou001:
Fixes #40622

### DIFF
--- a/apps/www/components/Examples/ExamplesCarousel.tsx
+++ b/apps/www/components/Examples/ExamplesCarousel.tsx
@@ -1,6 +1,6 @@
 import 'swiper/css'
 
-import { useRef } from 'react'
+import { useState } from 'react'
 import { ArrowLeft, ArrowRight } from 'lucide-react'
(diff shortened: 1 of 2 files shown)


7ttp (commented):
closing this up 
as the issue is no longer reprducible! 💚


Thanks for ur efforts towards contributing to supabase!!

PR #40629, opened 2025-11-19
Analytics: Update tracking for Advisor

awaseem:

YES

Add metrics for advisor clicks/views

No metrics 

## What is the new behavior?

Metrics for when advisors are clicked or viewed


apps/studio/components/interfaces/Linter/LintDetail.tsx
```diff
@@ -18,9 +19,31 @@ interface LintDetailProps {
 }
 
 const LintDetail = ({ lint, projectRef, onAskAssistant }: LintDetailProps) => {
+  const track = useTrack()
```
awaseem, line 22: Noob question, but I'm guessing this tracks self hosted and not send these events?

packages/common/telemetry-constants.ts
```diff
@@ -2313,6 +2313,56 @@ export interface LogDrainConfirmButtonSubmittedEvent {
...
+     */
+    advisor_category: 'PERFORMANCE' | 'SECURITY'
+    /**
+     * Specific advisor type/name, e.g. missing_index, no_rls_policy
+     */
+    advisor_type: string
+  }
+  groups: TelemetryGroups
+}
+
```
awaseem, line 2361: This is all copy pasta from the reference above. Let me know if this is wrong

apps/studio/components/interfaces/Linter/LintDetail.tsx
pamelachia: yes it is confirmed these do not send events if it is self hosted

packages/common/telemetry-constants.ts
```diff
@@ -2313,6 +2313,56 @@ export interface LogDrainConfirmButtonSubmittedEvent {
...
+/**
+ * User opened an advisor detail page to view a specific advisor.
+ * This tracks when users engage with advisor recommendations.
+ *
+ * @group Events
+ * @source studio
+ * @page Advisor Panel
+ */
+export interface AdvisorViewedEvent {
+  action: 'advisor_viewed'
```
pamelachia: This should be renamed to advisor_detail_opened and the @page should specify the advisor panel sidebar. As panel can mean anything.

pamelachia (changes requested):
We actually have some very similar events already being sent via posthog for assistant detail related clicks. Can we consolidate them?

apps/studio/components/interfaces/Linter/LintDetail.tsx
```diff
@@ -18,9 +19,31 @@ interface LintDetailProps {
...
   const snap = useAiAssistantStateSnapshot()
   const { openSidebar } = useSidebarManagerSnapshot()
 
+  const handleAskAssistant = () => {
+    // Track advisor_assistant_clicked event
```
pamelachia: Unnecessary comment, there are more across this PR that are just AI bloat IMO. Remove

packages/common/telemetry-constants.ts
```diff
@@ -2313,6 +2313,56 @@ export interface LogDrainConfirmButtonSubmittedEvent {
...
+  properties: {
+    /**
+     * Category of the advisor, e.g. performance, security
+     */
+    advisor_category: 'PERFORMANCE' | 'SECURITY'
```
pamelachia: Let's keep properties to camel case so advisorCategory. I am aware that some of the events are not camel case but i need to have a data cleanup for those events before changing them

pamelachia (dismissed):

awaseem:
This is my fault, the cursor shot was not good so going to move to draft and fix all the issues before re-opening

packages/common/telemetry-constants.ts
```diff
@@ -1624,51 +1606,6 @@ export interface RealtimeExperimentExposedEvent {
...
- * @source studio
- * @page /project/{ref}
- */
-export interface HomeAdvisorFixIssueClickedEvent {
-  action: 'home_advisor_fix_issue_clicked'
-  properties: {
-    /**
-     * Category of the issue (SECURITY or PERFORMANCE)
-     */
-    issue_category: string
```
awaseem, line 1699: Don't understand why we pass in the first value when this is an array, the new events just pass all the values returned from the lint notification

pamelachia: An issue usually has one category. Were there any lints that had multiple categories? I tested and only got one category value returned which is why i only pass the first value to keep the events easier to search

pamelachia (approved):

charislam (approved):

PR #40640, opened 2025-11-20
chore: Remove the download bucket object mutation

ivasilov:
This PR removes the mutation file since we've migrated to using the supabase client for downloading files from Studio.


awaseem (approved):
LGTM

PR #40641, opened 2025-11-20
chore: Clean `onSuccess` and `onError` props on `useQuery`

ivasilov:
This PR removes all `onSuccess` and `onError` props on `useQuery`. They've been deprecated for a while, and they're finally removed in v5.


alaister (approved):
LGTM!

- Spot tested on preview

PR #40642, opened 2025-11-20
chore: Migrate all `isPending` uses in `react-query`

ivasilov:
This PR bumps RQ to the latest v4 which has some forward-compatible changes to make the migration easier:
- renaming `isLoading` to `isPending` which is a mechanical rename (no change in behaviour).
- renaming `remove` to `queryClient.removeQueries`
- Minor type fixes.


alaister (approved):
LGTM!

- Also spot tested on preview

staaldraad (approved):

PR #40644, opened 2025-11-20
ref(etl): Improve updates, rollbacks and errors UI

iambriccardo:
This PR improves the ETL UI in the following ways:
- Removes the rollback UI and just allows full rollback via `Reset Table and Restart` button.
- Improves the error UI by showing a tooltip of the current error (useful for debugging purposes).
- Now all operations on a pipeline, will kickoff a restart (`stop` + `start` in sequence), which is required due to some limitations in K8s.
- The dialogs for updating pipelines are a bit simpler and you can update a pipeline without having to start it (e.g., if the pipeline is stopped).
- The `View Status` button doesn't show anymore the errors of the tables if the pipeline is stopped.

### Screenshots

Update with stopped pipeline
image

Updated with running pipeline
image

New error view for table status
image

New dialog for resetting table
image


joshenlim (approved):
Generally LGTM! Verified changes locally - just pushed up some changes:
- Fix `NewPublicationPanel` save button also triggering onSubmit action on `DestinationPanel` (due to nested forms)
- Fix creating namespace in DestinationPanel missing API key on `useIcerbergCreateMutation`
- Small update to `UpdateVersionModal` to shift version ID changes within a collapsible
  image
- Convert "Show error" button to a dialog instead of tooltip (more breathing space) + use `CodeBlock` component to support copying
  image

- Small styling fix for `ResetTableButton`

PR #40645, opened 2025-11-20
Update UI to de-couple ETL replication context from Analytics Buckets

joshenlim:
## Context

We're scoping down the technicalities of Analytics Buckets by de-coupling it from ETL replication (so Analytics Bucket will just be about the buckets themselves), which helps us simplify a lot of the complexities around the UI + makes the UI behaviour a lot more deterministic and fault tolerant.

Main user flows covered will be: 
- Creating namespaces/tables
- Viewing data from tables
- Deleting namespaces/tables

The decision of whether to fully de-couple ETL from Analytics Bucket is still in the air, so decided to leave all of the relevant code as is for now, in case we'll revisit them again down the road. On that note though, the code might be a little messy but I hope to clean things up bit by bit in subsequent PRs.

## Changes involved

Our primary CTA of creating tables in a bucket will be to do so through an external Iceberg client - so the empty state is a set of instructions for Pyiceberg which should work out of the box.
image

The sample snippet here will create a namespace called `default` and a table `events` within that namespace. (There's no automatic polling so you'll need to refresh the browser here)
image

You can opt to view the data in the Iceberg table via Postgres by following the "Query from Postgres" which then explains + prompts you to do so via Foreign Data Wrappers. At this step you just need to specify which schema you'd want to have the foreign tables in.
image

This then creates the foreign tables and shows you how you can view the data of your iceberg table from
image

From here that's pretty much it in terms of functionality. You can then choose to delete individual iceberg tables (which also cleans up any associated foreign tables), or delete the whole namespace (which will also clean up any associated foreign tables for iceberg tables in the namespace)

Past the empty state, the "Create table" CTA then shifts up to the header section (as per Vector Buckets), where clicking just opens the set of instructions within a Dialog. (The relevancy of these instructions at this state is up for future debate but good for now)

## Known issues

- [ ] We're using the temporary API keys for all iceberg related requests, and it's currently a known issue that we'll run into 403s rather often (Especially when deleting namespaces / tables). Currently being discussed with Auth team, so if you run into any network request issues because of this, best bet is to just try again for now. (e.g Refresh browser)

## To test

- [ ] Validate that the empty state instructions work, let me know if there's any confusion
- [ ] Generally, validate how the UX feels like, if there's any confusion, any copy writing improvements

### DIFF
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/CreateTableInstructions/index.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/CreateTableInstructions/index.tsx
@@ -5,6 +5,7 @@ import CommandRender from 'components/interfaces/Functions/CommandRender'
 import { convertKVStringArrayToJson } from 'components/interfaces/Integrations/Wrappers/Wrappers.utils'
 import { ButtonTooltip } from 'components/ui/ButtonTooltip'
 import CopyButton from 'components/ui/CopyButton'
+import { InlineLink } from 'components/ui/InlineLink'
 import { useProjectSettingsV2Query } from 'data/config/project-settings-v2-query'
 import {
   getDecryptedValues,
@@ -18,6 +19,7 @@ import {
   AccordionItem_Shadcn_,
   AccordionTrigger_Shadcn_,
   Card,
+  CardFooter,
   CardHeader,
   CardTitle,
   CodeBlock,
@@ -226,7 +228,7 @@ export const CreateTableInstructions = ({
           </AccordionContent_Shadcn_>
         </AccordionItem_Shadcn_>
 
-        <AccordionItem_Shadcn_ value="step-3" className="border-0">
+        <AccordionItem_Shadcn_ value="step-3">
           <AccordionTrigger_Shadcn_ className="px-6 py-3 text-sm">
             <div className="flex items-center gap-x-4">
               <div className="w-6 h-6 rounded-full border flex items-center justify-center text-xs font-mono">
@@ -248,6 +250,16 @@ export const CreateTableInstructions = ({
           </AccordionContent_Shadcn_>
         </AccordionItem_Shadcn_>
       </Accordion_Shadcn_>
+
+      <CardFooter className="bg">
+        <p className="text-xs text-foreground-light">
+          Connecting to bucket with other Iceberg clients? Read more in our{' '}
+          <InlineLink href="https://supabase.com/docs/guides/storage/analytics/connecting-to-analytics-bucket">
+            documentation
+          </InlineLink>
+          .
+        </p>
+      </CardFooter>
     </Card>
   )
 }
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/CreateTableInstructions/index.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/CreateTableInstructions/index.tsx
@@ -12,6 +12,7 @@ import {
   useVaultSecretDecryptedValueQuery,
 } from 'data/vault/vault-secret-decrypted-value-query'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
+import { DOCS_URL } from 'lib/constants'
 import { Eye, EyeOff } from 'lucide-react'
 import {
   Accordion_Shadcn_,
@@ -254,7 +255,7 @@ export const CreateTableInstructions = ({
       <CardFooter className="bg">
         <p className="text-xs text-foreground-light">
           Connecting to bucket with other Iceberg clients? Read more in our{' '}
-          <InlineLink href="https://supabase.com/docs/guides/storage/analytics/connecting-to-analytics-bucket">
+          <InlineLink href={`${DOCS_URL}/guides/storage/analytics/connecting-to-analytics-bucket`}>
             documentation
           </InlineLink>
           .
--- a/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/CreateTableInstructions/index.tsx
+++ b/apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/CreateTableInstructions/index.tsx
@@ -1,0 +1,253 @@
+import { useMemo, useState } from 'react'
+
+import { useParams } from 'common'
+import CommandRender from 'components/interfaces/Functions/CommandRender'
+import { convertKVStringArrayToJson } from 'components/interfaces/Integrations/Wrappers/Wrappers.utils'
+import { ButtonTooltip } from 'components/ui/ButtonTooltip'
+import CopyButton from 'components/ui/CopyButton'
+import { useProjectSettingsV2Query } from 'data/config/project-settings-v2-query'
+import {
+  getDecryptedValues,
+  useVaultSecretDecryptedValueQuery,
+} from 'data/vault/vault-secret-decrypted-value-query'
+import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
+import { Eye, EyeOff } from 'lucide-react'
+import {
+  Accordion_Shadcn_,
+  AccordionContent_Shadcn_,
+  AccordionItem_Shadcn_,
+  AccordionTrigger_Shadcn_,
+  Card,
+  CardHeader,
+  CardTitle,
+  CodeBlock,
+} from 'ui'
+import { useAnalyticsBucketWrapperInstance } from '../useAnalyticsBucketWrapperInstance'
+import { getPyicebergSnippet } from './CreateTableInstructions.constants'
+
+export const CreateTableInstructions = ({
+  hideHeader = false,
+  className,
+}: {
+  hideHeader?: boolean
+  className?: string
+}) => {
+  const { ref, bucketId } = useParams()
+  const { data: project } = useSelectedProjectQuery()
+  const [showKeys, setShowKeys] = useState(false)
+  const [isFetchingSecretsOnCopy, setIsFetchingSecretsOnCopy] = useState(false)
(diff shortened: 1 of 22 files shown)


apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/CreateTableInstructions/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+              ]}
+            />
+          </AccordionContent_Shadcn_>
+        </AccordionItem_Shadcn_>
+      </Accordion_Shadcn_>
+
+      <CardFooter className="bg">
+        <p className="text-xs text-foreground-light">
+          Connecting to bucket with other Iceberg clients? Read more in our{' '}
+          <InlineLink href="https://supabase.com/docs/guides/storage/analytics/connecting-to-analytics-bucket">
```
charislam:
```suggestion
          <InlineLink href={`${DOCS_URL}/guides/storage/analytics/connecting-to-analytics-bucket`}>
```

so we don't link to the wrong domain on the Nimbus build -- though is this even available on Nimbus?

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/index.tsx
```diff
@@ -89,6 +102,17 @@ export const NamespaceWithTables = ({
     }
   )
 
+  const connectedForeignTablesForNamespace =
+    (icebergWrapper?.tables ?? []).filter((x) => x.options[0].startsWith(`table=${namespace}.`)) ??
+    []
```
charislam:
```suggestion
    (icebergWrapper?.tables ?? []).filter((x) => x.options[0].startsWith(`table=${namespace}.`))
```

This fallback should be redundant since there's an earlier fallback to `[]`, I think?

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/index.tsx
```diff
@@ -89,6 +102,17 @@ export const NamespaceWithTables = ({
...
+  const tablesWithConnectedForeignTables = connectedForeignTablesForNamespace.reduce((a, b) => {
+    const table = b.options[0].split(`table=${namespace}.`)[1]
+    if (!a.find((x) => x === table)) a.push(table)
+    return a
+  }, [] as string[])
+  const unconnectedTables = tablesData.filter((x) => !tablesWithConnectedForeignTables.includes(x))
```
charislam:
```suggestion
  const tablesWithConnectedForeignTables = connectedForeignTablesForNamespace.reduce((a, b) => {
    const table = b.options[0].split(`table=${namespace}.`)[1]
    a.add(table)
    return a
  }, new Set<string>())
  const unconnectedTables = tablesData.filter((x) => !tablesWithConnectedForeignTables.has(x))
```

nitpick and not terribly important, but looks a little cleaner and requires less iteration with the Set

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
```diff
@@ -243,10 +241,56 @@ export const TableRowComponent = ({
     }
   }
 
+  const connectedForeignTablesInNamespace =
+    (icebergWrapper?.tables ?? []).filter((x) => x.options[0].includes(`table=${namespace}.`)) ?? []
```
charislam:
```suggestion
  const connectedForeignTablesInNamespace =
    (icebergWrapper?.tables ?? []).filter((x) => x.options[0].includes(`table=${namespace}.`))
```

same here

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
```diff
@@ -243,10 +241,56 @@ export const TableRowComponent = ({
...
+
+  const connectedForeignTables =
+    (icebergWrapper?.tables ?? []).filter(
+      (x) => x.options[0] === `table=${namespace}.${table.name}`
+    ) ?? []
```
charislam:
```suggestion
  const connectedForeignTables =
    (icebergWrapper?.tables ?? []).filter(
      (x) => x.options[0] === `table=${namespace}.${table.name}`
    )
```

same here

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
```diff
@@ -349,14 +380,111 @@ export const TableRowComponent = ({
...
+                </DropdownMenuItem>
+              </DropdownMenuContent>
+            </DropdownMenu>
+          </TableCell>
         )}
       </TableRow>
+
+      {/* [Joshen] Render each foreign table associated to the namespace table as its own row */}
+      {connectedForeignTables?.map((x) => (
+        <TableRow>
```
charislam: needs a key

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/TableRowComponent.tsx
```diff
@@ -349,14 +380,111 @@ export const TableRowComponent = ({
...
+          <TableCell className="pl-6">
+            <div className="flex items-center gap-x-2 rounded">
+              <div className="w-4 h-4 rounded-bl-lg border-l-2 border-b-2 border-copntrol -translate-y-1.5" />
```
charislam:
```suggestion
              <div className="w-4 h-4 rounded-bl-lg border-l-2 border-b-2 border-control -translate-y-1.5" />
```

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/UpdateForeignSchemaDialog.tsx
```diff
@@ -0,0 +1,169 @@
...
+}) => {
+  const { ref: projectRef, bucketId } = useParams()
+  const { data: project } = useSelectedProjectQuery()
+
+  const [isOpen, setIsOpen] = useState(false)
+
+  const { icebergWrapper } = useAnalyticsBucketAssociatedEntities({ bucketId })
+  const connectedForeignTablesForNamespace =
+    (icebergWrapper?.tables ?? []).filter((x) => x.options[0].startsWith(`table=${namespace}.`)) ??
+    []
```
charislam:
```suggestion
  const connectedForeignTablesForNamespace =
    (icebergWrapper?.tables ?? []).filter((x) => x.options[0].startsWith(`table=${namespace}.`))
```

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/UpdateForeignSchemaDialog.tsx
```diff
@@ -0,0 +1,169 @@
...
+                      <TooltipTrigger>
+                        <span className={InlineLinkClassName}>{tables.length} tables</span>
+                      </TooltipTrigger>
+                      <TooltipContent className="max-w-64 text-center" side="bottom">
+                        {tables.join(', ')}
+                      </TooltipContent>
+                    </Tooltip>{' '}
+                    that have yet to be included in the Iceberg Foreign Data Wrapper. The wrapper
+                    can be updated to expose these tables as foreign tables where you then can query
+                    data from.
```
charislam:
```suggestion
                    that aren't included in the Iceberg Foreign Data Wrapper. Update the wrapper to
                    create foreign tables for all unexposed tables. This will let you query the tables
                    from Postgres.
```

charislam (commented):
I can't seem to get the default `main.py` script working. Tried several times with two different buckets and I keep getting a 403 on `/storage/v1/iceberg/v1/config`

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/CreateTableInstructions/index.tsx
joshenlim: hmm good question - unsure atm but i dont think it should be

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/index.tsx
joshenlim: oh great catch hahaha 🤦 overly defensive coding 😂

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/NamespaceWithTables/index.tsx
joshenlim: very neat! i like

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/UpdateForeignSchemaDialog.tsx
joshenlim: 🙏 🙏  thank you for this hahaha i struggled with the copy writing 😂

alaister (approved):
LGTM!

PR #40659, opened 2025-11-20
Fix: updated expiry to be 60 seconds

awaseem:

YES

Bug fix: We found out the management API is not respecting the expiry that well (aside from realtime).

Failure of downloads/uploads due to tokens expiring

## What is the new behavior?

Files should download correctly

### DIFF
--- a/apps/studio/data/api-keys/temp-api-keys-utils.ts
+++ b/apps/studio/data/api-keys/temp-api-keys-utils.ts
@@ -23,7 +23,7 @@ export function isTemporaryApiKeyValid(
 
   const now = Date.now()
   const timeRemaining = key.expiryTime - now
-  return timeRemaining > 60000 // More than 60 seconds remaining
+  return timeRemaining > 20_000 // More than 20 seconds remaining
 }
 
 const checkOrRefreshTemporaryApiKey = async (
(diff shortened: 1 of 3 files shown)


ivasilov (approved):

apps/studio/data/api-keys/temp-api-keys-utils.ts
```diff
@@ -47,9 +48,7 @@ const checkOrRefreshTemporaryApiKey = async (
...
 export function getOrRefreshTemporaryApiKey(projectRef: ProjectRef): Promise<TemporaryApiKey> {
   const existingKey = projectApiKeys.get(projectRef)
-  if (!existingKey) {
-    const data = checkOrRefreshTemporaryApiKey(projectRef, undefined)
-    projectApiKeys.set(projectRef, data)
-  }
-  return projectApiKeys.get(projectRef)!
+  const data = checkOrRefreshTemporaryApiKey(projectRef, existingKey)
+  projectApiKeys.set(projectRef, data)
+  return data
```
awaseem, line 53: @ivasilov it would only check if the Key is expired if the key didn't exist haha. We need to check it each time we get the key

PR #40660, opened 2025-11-20
fix: Fix the Edit policy panel work

ivasilov:
This PR fixes an issue where the Edit policy panel didn't work because of weird React issue.


awaseem (approved):
Tested on staging

PR #40667, opened 2025-11-20
fix(edge functions secrets): allow read-only role to read secrets metadata

charislam:
According to our API permissions model, the read-only role can also read secrets metadata (the actual secret itself is not viewable by anybody). Updating the frontend UI to match.

Testing:
- [ ] Secrets display looks the same for non-readonly role
- [ ] Secrets metadata now viewable for readonly role

Resolves FE-2118


awaseem (approved):
Tested locally with Read only user

PR #40669, opened 2025-11-21
Auth pages new page components

SaxonF:
This is a follow up to https://github.com/supabase/supabase/pull/40509 which replaces PageLayout and Scaffold components with our new page components.

This PR focuses on auth _pages_ first since auth has a large surface area, then will move to auth _components_.

To test:
Navigate through all project auth pages within Studio to ensure everything works as expected. Users will ideally not be able to notice the difference other than some consistency updates.

PR #40670, opened 2025-11-21
Auth interface components new page components

SaxonF:
This is a follow up to https://github.com/supabase/supabase/pull/40669 which replaces PageLayout and Scaffold components with our new page components.

This PR focuses on auth _interface components_ .

To test:
Navigate through all project auth pages within Studio to ensure everything works as expected. Users will ideally not be able to notice the difference other than some consistency updates.


awaseem:
Tried testing this and found the following: 
- Clicking on the breadcrumbs from email template gives you a 404 
Screenshot 2025-11-26 at 9 33 58 AM

Looking more deeply all the breadcrumbs from the auth page are broken on studio. Should fix that

awaseem (changes requested):
Request changes, PR breaks breadcrumbs

awaseem (approved):
LGTM

PR #40672, opened 2025-11-21
Fix transfer project button

joshenlim:
## Context

Transfer project button was staying disabled despite selecting an organization, bug was introduced in this [PR](https://github.com/supabase/supabase/pull/40642/files)

Realised this was because of the dependency array in the useEffect having `selectedOrg` in it, which was causing it to just get reset each time it changed within the `if (isOpen)` conditional block

Opting to only have `isOpen` within the dependency array

## To test

- [ ] Verify that you can transfer a project to another organization


alaister (approved):
LGTM!

- Tested on preview

PR #40673, opened 2025-11-21
Move chat state to project context

SaxonF:
This moves Assistant chat state out of AIAssistant and into our ai-assistant-chat by creating instances for each active chat. This has a few benefits:
- It allows us to close the chat sidebar without losing streaming
- It allows for multiple live chats at once
- It allows us to send messages to chats from outside Assistant. This is perhaps the most important because we currently just set the initial input of the Assistant which requires an extra step from the user even when we know the prompt is a good enough conversation starter.

This PR adjusts the fix with assistant button in lint detail so that it submits the message instead of just setting the input.

To test:
- Create an Assistant chat and send a message
- Create a new chat while the existing one is streaming and submit a message
- Close the chat sidebar and navigate around a project
- Open the sidebar and confirm both chats have a final response
- Ask the assistant to execute a query, click confirm on the resulting query, and make sure the conversation continues after the result has come back from the query
- Switch projects and ensure that all works as it should while Assistant sidebar is open
- Stop streaming, edit message, and delete message actions all still work


apps/studio/state/ai-assistant-state.tsx
```diff
@@ -251,6 +331,19 @@ export const createAiAssistantState = (): AiAssistantState => {
       }
       state.activeChatId = chatId
 
+      // Create new chat instance
+      // Create new chat instance
```
alaister, line 334: Extra comment
```suggestion
```

apps/studio/state/ai-assistant-state.tsx
```diff
@@ -382,6 +492,18 @@ export const createAiAssistantState = (): AiAssistantState => {
           state.newChat()
         }
       }
+
+      // Initialize chat instance for the active chat
+      if (state.activeChatId && state.chats[state.activeChatId]) {
+        if (!state.chatInstances[state.activeChatId]) {
```
alaister: Could these two `if` statements be combined?

alaister (approved):
LGTM!

- Tested on preview

Couple of tiny nits:

PR #40679, opened 2025-11-21
ref(etl): Rework docs to avoid using ETL everywhere

iambriccardo:
This PR improves the ETL docs by talking about manual and managed destinations and improves docs by using tabs instead of separate pages, aligns terminologies and cleans up the structure.


iambriccardo:
Paused for now.

PR #40699, opened 2025-11-21
fix: custom report block is firing SQL queries twice

awaseem:

YES

Bug fix: the report block is firing 2 requests per SQL snippet. To fix it added a ref to keep track of the SQL query and ensure it doesn't fire again

## Testing

- [ ] Create a custom report with SQL snippets 
- [ ] Ensure when you refresh the report, the SQL query run
- [ ] On a fresh page load, it should only run the query once


apps/studio/components/interfaces/Reports/ReportBlock/ReportBlock.tsx
```diff
@@ -66,81 +73,65 @@ export const ReportBlock = ({
...
-  const sqlHasChanged = useChangedSync(sql)
-  const isRefreshingChanged = useChangedSync(isRefreshing)
-  if (sqlHasChanged || (isRefreshingChanged && isRefreshing)) {
-    runQuery('select', sql)
-  }
+  useEffect(() => {
+    if (isRefreshing) {
+      refetch()
+    }
+  }, [isRefreshing, refetch])
```
awaseem, line 134: @ivasilov with the refetching, I had to use the parents isRefreshing and then fire. Adding isRefreshing to the query key would fire per change `false` + `true`

Any ideas if this is right? Or anything we can do to make it better

apps/studio/components/interfaces/Reports/ReportBlock/ReportBlock.tsx
```diff
@@ -66,81 +73,65 @@ export const ReportBlock = ({
...
+    if (executeSqlError) {
+      const errorMessage = String(executeSqlError).toLowerCase()
+      const isReadOnlyError =
+        errorMessage.includes('read-only transaction') ||
+        errorMessage.includes('permission denied') ||
+        errorMessage.includes('must be owner')
+
+      if (isReadOnlyError) {
+        setIsWriteQuery(true)
+      }
```
awaseem, line 126: Logic is so confusing on this, but it still excutes write statements. The good thing is that its only the reader connection string so it'll fail.

This logic is still needed to set the isWriteBoolean

apps/studio/components/interfaces/Reports/ReportBlock/ReportBlock.tsx
```diff
@@ -66,81 +73,65 @@ export const ReportBlock = ({
...
-      return true
+      return executeSql({
+        projectRef,
+        connectionString,
+        sql,
+      })
     },
-    [projectRef, readOnlyConnectionString, postgresConnectionString, executeSql]
-  )
+    enabled: !isLoadingContent && contentError == null,
```
awaseem, line 110: Only fire when values are in a good state

apps/studio/components/interfaces/Reports/ReportBlock/ReportBlock.tsx
```diff
@@ -66,81 +73,65 @@ export const ReportBlock = ({
...
-          ? postgresConnectionString
-          : readOnlyConnectionString ?? postgresConnectionString
+    refetch,
+  } = useQuery({
+    queryKey: sqlKeys.query(projectRef, [
+      item.id,
+      sql,
+      readOnlyConnectionString,
+      postgresConnectionString,
+    ]),
```
awaseem, line 93: Is this the best Query key to use here @ivasilov ?

ivasilov (approved):

apps/studio/components/interfaces/Reports/ReportBlock/ReportBlock.tsx
ivasilov: Yeah, I think it's ok, the sql string will be unique.

PR #40703, opened 2025-11-22
feat: persistent branching entitlement

ignaciodob:
Add entitlement checks and upgrade prompts for the persistent branching feature in the branch management UI.

## Changes
- Add `useCheckEntitlements` hook to verify access to persistent branching
- Display upgrade prompt when users don't have access to persistent branches (platform only)
- Disable branch mode switching for preview branches when user lacks persistent branching entitlement
- Show appropriate tooltip message directing users to upgrade plan

##  Testing
To properly test all the functionality you'll need to run this in your local stack (with https://github.com/supabase/infrastructure/pull/27713 applied)

### Persistent branchining disabled in the Pro Plan
- Disable the feature by editing the Entitlement catalog `PERSISTENT_BRANCHES` feature to `enabled: false` for the Pro Plan
- Head to the manage branches page with an Org on the Pro Plan and at least one preview branch
- Assert that you see an upgrade message and button in the Persistent branches section
- Assert that you can't "Switch to persistent" a preview branch (option should be disabled with a tooltip)

### Not having access still lets you switch a persistent branch to preview
- Enable the feature by editing the Entitlement catalog `PERSISTENT_BRANCHES` feature to `enabled: true` for the Pro Plan
- Switch a preview branch to persistent
- Assert that the branch is now persistent
- Disable the feature by editing the Entitlement catalog `PERSISTENT_BRANCHES` feature to `enabled: false` for the Pro Plan
- Refresh the page and assert that the persistent branch in the Persistent branches section is still visible
- Assert that you can switch the branch to preview.

### Persistent branchining enabled in the Pro Plan
- Enable the feature by editing the Entitlement catalog `PERSISTENT_BRANCHES` feature to `enabled: true` for the Pro Plan
- Assert that you're able to switch a branch to persistent without issues


apps/studio/components/interfaces/BranchManagement/Overview.tsx
```diff
@@ -65,6 +66,10 @@ export const Overview = ({
     (branch) => branch.persistent
   )
   const { ref: projectRef } = useParams()
+  const { data: selectedOrg } = useSelectedOrganizationQuery()
+
+  const { hasAccess: hasAccessToPersistentBranching, isLoading: isLoadingEntitlement } =
+    useCheckEntitlements('persistent_branching')
```
kevcodez, line 74: Might need to change the naming here, but code LGTM

kevcodez (approved):

PR #40704, opened 2025-11-22
feat(studio): show info to unauthorized user in github integration

myerekapan:

YES

Improved UX for GitHub integration settings

When users lack permissions to update or create the GitHub integration, the form fields look disabled (with a different cursor), but there's no explanation of why they're disabled. I personally had to guess that it was a permissions issue.

## What is the new behavior?

Added tooltips to all disabled form fields that inform users they need additional permissions to update or create the GitHub integration. 

## Additional context

I used the `Tooltip`, `TooltipTrigger`, and `TooltipContent` components from the UI library because this matches the pattern used in other parts of the codebase (`ProjectComplianceMode.tsx`) for permission-based tooltips on disabled form controls.

I also considered using the `InfoTooltip` component but this component already has a built-in trigger (info icon) and can't wrap existing form controls (I think so but correct me if I'm wrong )

**Question for reviewers:** Should I switch to using `InfoTooltip` icons next to each form field label instead of the current approach?


Hallidayo:

PR #40723, opened 2025-11-24
fix(studio): entity type color contrast

dnywh:

UI fix

- Many entity type icons aren’t legible on light mode
- Similar issue on plan upgrade sheet

## What is the new behavior?

- Minor updates to the above to increase contrast

## Additional context

| Before | After |
| --- | --- |
| Supabase | Supabase |


joshenlim (approved):

PR #40724, opened 2025-11-24
chore(studio): improve inline code styling

dnywh:

- Copywriting and UI improvements
- Design documentation

- Our confirmation dialogs use slightly redundant language
- Some of our code references are visually hard to distinguish from plain text

## What is the new behavior?

- Fixes to the above
- Related [design documentation](https://design-system-git-dnywh-choreconfirm-language-c-ec7325-supabase.vercel.app/design-system/docs/typography)

## Additional context

| Before | After |
| --- | --- |
| Verbose language, hard-to-distinguish `wrappers` | Fixes to both of these issues |

Note that I did not do a full find-and-replace of “Confirm to enable” → “Enable” strings. I just changed the few that:

1. Were most egregious
2. Were easiest to test


apps/design-system/config/docs.ts
```diff
@@ -36,9 +41,9 @@ export const docsConfig: DocsConfig = {
           items: [],
         },
         {
-          items: [],
-          href: '/docs/ui-patterns/accessibility',
```
dnywh, line 40: Was wrongly ordered. The `href` was also wrong, since this doesn’t live in `ui-patterns`.

apps/design-system/content/docs/typography.mdx
dnywh, line 1: Just a first pass so we have _something_ about these typography classes. See [Primer](https://primer.style/product/primitives/typography/) for a great example of what I envision this working towards.

apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
dnywh, line 1: This is the one pictured in the PR description.

joshenlim:
i'm actually inclined to apply `text-code-line` for all instances of the `code` element actually
tbh there's a currently an issue whereby the padding for the `code` element only shows up if the parent has the `prose` class name which I'm not sure if its intentional - almost all the time i use the `code` element i'd add the `prose` class name to the `parent` just to get the styling

joshenlim (approved):
Skimmed staging preview for changes!

PR #40738, opened 2025-11-24
fix(migrations): ensure pgmq extension consistency during resets

### LINKED ISSUE #4492: PGMQ queue creation fails on preview branch reset: relation "pgmq.q_my_queue_msg_id_seq" does not exist
# Bug report

- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When resetting a Supabase preview branch, PGMQ queue creation fails with a sequence-related error. The pgmq.create() call for a queue (e.g., my_queue) references a sequence (pgmq.q_my_queue_msg_id_seq) that doesn't exist, causing the migration to fail.

## To Reproduce

1. Create a preview branch with migrations that create PGMQ queues:
```sql
   CREATE EXTENSION IF NOT EXISTS pgmq;
   SELECT pgmq.create('my_queue');
   SELECT pgmq.create('my_queue_dlq');
```
2. Let the branch build successfully (first build works).
3. Reset the preview branch via the Supabase dashboard.
- Go to the manage preview branches page -> https://supabase.com/dashboard/project/{your-project-id}/branches
- Click on 'Reset branch' under the menu on the right of your branch name
4. The branch rebuilds and the migration fails.
5. View the logs of the latest build
6. See error

## Expected behavior

- Queues should be created successfully on branch reset.
- PGMQ extension state should be consistent after reset.
- No sequence-related errors should occur.

## Actual behavior
- Migration fails with:
```sql
  ERROR: relation "pgmq.q_my_queue_msg_id_seq" does not exist (SQLSTATE 42P01)
  At statement: 786
  SELECT pgmq.create('my_queue_dlq')
```
- The first queue (my_queue) may be created, but the second queue (my_queue_dlq) fails.
- This only occurs on branch reset, not on first build or new commits.

## System information

- Supabase: Preview branch reset functionality
- PGMQ extension: Latest version (via CREATE EXTENSION IF NOT EXISTS pgmq)
- Migration file: Baseline migration or any migration creating PGMQ queues

7ttp:
when preview branches are reset, the pgmq extension gets left in a partial state where extension metadata exists but actual database objects (sequences, tables, functions) are missing. this causes `relation "pgmq.q_*_msg_id_seq" does not exist` errors when trying to create queues.

## adds

adds a pre-migration check that detects partial pgmq state and performs a clean rebuild:
- checks if pgmq extension or schema exists
- if found, drops both with cascade (removes all dependent objects)
- recreates extension in clean state
- if not found, skips (no-op)

this is idempotent and only affects pgmq-enabled databases.

fixes supabase/cli#4492

### DIFF
--- a/apps/studio/lib/api/self-hosted/migrations.ts
+++ b/apps/studio/lib/api/self-hosted/migrations.ts
@@ -21,6 +21,23 @@ alter table supabase_migrations.schema_migrations add column if not exists name 
 
 commit;`
 
+const ensurePgmqConsistency = () => `
+do $$
+declare
+  ext_exists boolean;
+  schema_exists boolean;
+begin
+  select exists(select 1 from pg_extension where extname = 'pgmq') into ext_exists;
+  select exists(select 1 from pg_namespace where nspname = 'pgmq') into schema_exists;
+  
+  if ext_exists or schema_exists then
+    execute 'drop extension if exists pgmq cascade';
+    execute 'drop schema if exists pgmq cascade';
+    execute 'create extension pgmq';
+  end if;
+end $$;
+`
+
 const applyAndTrackMigrationsQuery = (query: string, name?: string) => {
   // Escapes literals using postgres dollar quoted string
   const dollar = `$${makeRandomString(20)}$`
@@ -99,6 +116,15 @@ export async function applyAndTrackMigrations<T = unknown>({
 
   if (initializeResponse.error) {
     return initializeResponse
+  }
+
+  const cleanupResponse = await executeQuery<void>({
+    query: ensurePgmqConsistency(),
+    headers,
+  })
+
+  if (cleanupResponse.error) {
+    return cleanupResponse
   }
 
   const applyAndTrackResponse = await executeQuery<T>({


kallebysantos:
@7ttp thanks for your contributions 💚

I'm not really sure if this PR will solve https://github.com/supabase/cli/issues/4492, since its a Platform issue, as well the changed code will never be executed on it, cause [it's scoped to Self-Host](https://github.com/supabase/supabase/pull/40738/files#diff-00b931228e37e3ae7f7aafa538e2af116a19c648a142e39fd3e46371810f1a1cR110) 

Also, at this moment database branching is not possible when self-hosting 😕
Its also requires a [Pro plan tier](https://supabase.com/docs/guides/platform/manage-your-usage/branching#pricing) or greater to enable it

7ttp:
@kallebysantos you're right, my change was in the wrong place.

the issue only happens on platform during preview branch resets, but i modified self-hosted/migrations.ts, which never runs there. the fix needs to live inside the platform’s branch-reset logic, not in studio’s self-hosted migration flow. this PR wouldn’t resolve supabase/cli#4492.

thanks for pointing it out. i misunderstood the separation between the self-hosted and platform code paths.

PR #40739, opened 2025-11-24
Joshen/depr 253 tighten field validation for bucket name and table name

joshenlim:
## Context

Updates the validation for bucket names for Analytics and Vector Buckets

Analytics:
- Bucket names must be between 3 and 63 characters long.
- Bucket names can consist only of lowercase letters, numbers, and hyphens (-).
- Bucket names must begin and end with a letter or number.
- Bucket names must not contain any underscores (_) or periods (.).
- Bucket names must not start with any of the following reserved prefixes: 
  - `xn--`
  - `sthree-`
  - `amzn-s3-demo-`
  - `aws`
- Bucket names must not end with any of the following reserved suffixes:
  - `-s3alias`
  - `--ol-s3`
  - `--x-s3`
  - `--table-s3`

Vectors:
- Vector bucket names must be between 3 and 63 characters long.
- Vector bucket names can consist only of lowercase letters (a-z), numbers (0-9), and hyphens (-).
- Vector bucket names must begin and end with a letter or number.

## To test

- [ ] Just double check that the above validation rules are working in the dashboard with appropriate UX response (e.g error messages)


alaister (approved):
LGTM!

- Tested on preview

PR #40748, opened 2025-11-24
fix: update wording in support ticket confirmation

hf:
I asked a friend to create a support ticket, and he sent me back the project ref. Then he sent me the screenshot. The wording is definitely confusing and can be easily interpreted to mean that is the ticket ID.

### DIFF
--- a/apps/studio/components/interfaces/Support/Success.tsx
+++ b/apps/studio/components/interfaces/Support/Success.tsx
@@ -47,7 +47,7 @@ export const Success = ({
         {selectedProject !== NO_PROJECT_MARKER && (
           <p className="text-sm text-foreground-light">
             Your ticket has been logged for the project{' '}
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Support/Success.tsx
```diff
@@ -47,7 +47,7 @@ export const Success = ({
         {selectedProject !== NO_PROJECT_MARKER && (
           <p className="text-sm text-foreground-light">
             Your ticket has been logged for the project{' '}
-            <span className="text-foreground">{projectName}</span>, reference ID:{' '}
+            <span className="text-foreground">{projectName}</span> with project ID:{' '}
```
MildTomato:
```suggestion
            <span className="text-foreground">{projectName}</span> with Project ID:{' '}
```

MildTomato (approved):

PR #40751, opened 2025-11-24
feat: make publishable, secret api keys page front-and-center

hf:
In an effort to push publishable and secret key adoption, their page is the first one when opening the API Keys page. The tab titles have changed to be clearer, and the `/new` page now redirects to the main page.


awaseem (approved):
Tested on staging. This doesn't exist on self hosted

PR #40752, opened 2025-11-24
feat: jwt signing keys front-and-center

hf:
To push JWT signing key adoption and upcoming changes where new projects receive an asymmetric JWT signing key by default, the page is placed front-and-center with the legacy page being secondary.


awaseem (approved):
Tested on staging! redirect works as well

PR #40769, opened 2025-11-25
ref(etl): Rename ETL Replication to Replication and update docs

iambriccardo:


joshenlim (approved):
Minimally did a quick smoke test on staging preview - looks alright
Also checked code locally to verify if that's any trailing `etl` routes - LGTM!

PR #40775, opened 2025-11-25
fix(studio): correct cmdk navigation routes for database settings

pamelachia:
## Summary
Fixes incorrect navigation routes in the command menu (cmd+k) for database-related settings. The routes were pointing to `/settings/general` which no longer exists for these items - they should navigate to `/database/settings` instead. Also fixes the connection string command to properly open the connect modal using query params.

## Changes
- Updated database password route from `/settings/general#database-password` to `/database/settings#database-password`
- Changed connection string from a route to an action that opens the connect modal via `showConnect` query param
- Updated connection pooling route from `/settings/general#connection-pooling` to `/database/settings#connection-pooler`
- Updated SSL configuration route from `/settings/general#ssl-configuration` to `/database/settings#ssl-configuration`

## Testing
**Quick test:**
1. Open command menu (cmd+k) and search for "database password", "connection string", "connection pooling", or "SSL configuration"
2. Verify each command navigates to the correct location in database settings

## Linear
- fixes GROWTH-554


alaister (approved):
LGTM!

- Tested on preview

PR #40777, opened 2025-11-25
[FE-2111] chore: swap custom domains reverify to query

alaister:
Some users are seeing 429s on the reverify endpoint. This PR aims to address that by swapping the endpoint to use a query instead of a mutation, and increasing the refetch interval to 10s to match the `/custom-domains` endpoint.

Benefits:
- Will respect `Retry-After` header if too many requests are sent
- Less calls to `/custom-domains` after each reverify attempt (It will now only call if the reverify data is different, thanks to `useEffect`)
- Reverify button will show background refreshes automatically

To test:
- Go through the custom domains addition process on a project


awaseem (approved):
Nice work, will respond back to customer

PR #40783, opened 2025-11-25
Patch lints

joshenlim:
## Context

Realised that if `exposedSchemas` is not provided, this breaks the SQL under `enrichLintsQuery`

Should conditionally render the `set local pgrst.db_schemas` statement


mattrossman (approved):
Confirming this fixes errors from `get_advisors` in local MCP


| performance | security |
|--------|--------|
| CleanShot 2025-11-25 at 09 48 04@2x | CleanShot 2025-11-25 at 09 47 50@2x |

ivasilov (approved):

PR #40796, opened 2025-11-25
feat: Edge functions search and sorting

awaseem:

YES

Users can now search functions by name and sort them by name, created_at, and updated_at

The API doesn't support this so its all client side for now. Will change later

## Demo

https://github.com/user-attachments/assets/3f97e727-a75a-428d-843f-64ca9c0d4e78

### DIFF
--- a/apps/studio/components/interfaces/EdgeFunctions/EdgeFunctionsSortDropdown.tsx
+++ b/apps/studio/components/interfaces/EdgeFunctions/EdgeFunctionsSortDropdown.tsx
@@ -11,13 +11,10 @@ import {
   DropdownMenuTrigger,
 } from 'ui'
 
-export type EdgeFunctionsSort =
-  | 'name:asc'
-  | 'name:desc'
-  | 'created_at:asc'
-  | 'created_at:desc'
-  | 'updated_at:asc'
-  | 'updated_at:desc'
+export type EdgeFunctionsSortColumn = 'name' | 'created_at' | 'updated_at'
+export type EdgeFunctionsSortOrder = 'asc' | 'desc'
+
+export type EdgeFunctionsSort = `${EdgeFunctionsSortColumn}:${EdgeFunctionsSortOrder}`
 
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/EdgeFunctions/EdgeFunctionsSortDropdown.tsx
```diff
@@ -0,0 +1,76 @@
+import { ArrowDownNarrowWide, ArrowDownWideNarrow } from 'lucide-react'
```
awaseem, line 1: Just cleaner to have its own component

apps/studio/pages/project/[ref]/functions/index.tsx
```diff
@@ -38,6 +54,33 @@ const EdgeFunctionsPage: NextPageWithLayout = () => {
...
+        return a.name.localeCompare(b.name) * orderMultiplier
+      }
+      if (sortCol === 'created_at') {
+        return (a.created_at - b.created_at) * orderMultiplier
+      }
+      if (sortCol === 'updated_at') {
+        return (a.updated_at - b.updated_at) * orderMultiplier
+      }
+      return 0
+    })
```
Comment (awaseem) [inline apps/studio/pages/project/[ref]/functions/index.tsx:87]: this is a bit better when using strings, makes the types harder to mess up and make the implementation easier

kemaldotearth (approved):
Would love this as a standard for most every table like this! LGTM 🚢

apps/studio/pages/project/[ref]/functions/index.tsx
```diff
@@ -89,46 +178,36 @@ const EdgeFunctionsPage: NextPageWithLayout = () => {
...
-              {IS_PLATFORM && <DeployEdgeFunctionButton />}
-            </PageHeaderAside>
-          </PageHeaderMeta>
-        </PageHeader>
-
-        {page}
-      </div>
-    )
-  }
-
```
Comment (awaseem) [inline apps/studio/pages/project/[ref]/functions/index.tsx:127]: Super unneeded array causing so many re-renders.

PR #40798, opened 2025-11-25
docs: Use getClaims instead of getUser in Next.js Auth prompt

jnschaeffer:

YES

This PR updates the Next.js Supabase Auth AI tools prompt to use `supabase.auth.getClaims()` from the Supabase SDK instead of `supabase.auth.getUser()`. This is generally good practice, as `getClaims` does not make network calls unless the user's session is invalid or uses a symmetric signing key.

This prompt instructs the AI tool to use `getUser`, which always makes a network call to the project's Auth API.

## What is the new behavior?

The prompt instructs the AI tool to use `getClaims`, which is generally faster as long as a user has migrated their JWT signing keys.

### DIFF
--- a/examples/prompts/nextjs-supabase-auth.md
+++ b/examples/prompts/nextjs-supabase-auth.md
@@ -149,17 +149,16 @@ export async function middleware(request: NextRequest) {
   )
 
   // Do not run code between createServerClient and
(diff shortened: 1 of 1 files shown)


Hallidayo:

PR #40800, opened 2025-11-25
Security definer materialized view

### LINKED ISSUE #40799: "Security Definer view" error on a materialized view exposed via API.
I don't know if it was intentional or not, but it seems confusing adding the label "Security Definer view" when the error has nothing to do with it.

In fact, the error "Materialized view in API" seems to be completely unrelated.

Image

marsou001:
Fixes #40799 

Now, the correct error will be shown for materialized views accessible to `anon` and `authenticated` roles via Data API, instead of the unrelated "Security Definer view" error.

For convenience, users can immediately fix the issue by running the correct SQL query to revoke select from `anon` and `authenticated`, just by clicking on "confirm" on the confirmation modal I just created.

https://github.com/user-attachments/assets/f3ce9353-4ad0-4063-bf33-0b403f5fa87a

Before

materialized

Edit:

Thanks to the review by @saltcod, I now realise that revoking access from `anon` and `authenticated` users might not be the optimal solution since it would break many projects relying on the access to materialized views.

After pondering on the possible solutions, I figured there isn't an easy one, so I did away with the autofix button, and instead created a dialog explaining three possible options for the user, with a sample query under each one for convenience.

## Summary by CodeRabbit

* **New Features**
  * Added clearer warnings when materialized views are accessible through the API.
  * Added a dialog explaining how to review and revoke API access for materialized views.
  * Added guidance and code examples for restricting access to API roles.
  * Added a link to the Security Advisor for additional information.

* **Improvements**
  * Updated materialized view tooltips to accurately describe API accessibility.


kallebysantos:
Hi @marsou001  💚
Thanks for your contributions, may you run `pnpm run format` to pass CI?

marsou001:
Sorry, my bad! I really thought I ran the test
On it!

marsou001:
@kallebysantos I ran the test, it has passed!

saltcod:
Hey @marsou001 

Thanks so much for this! 
Had a look think we should remove the autofix. 

It would break anyone who is relying on accessing their Materialized view over Data APIs
There's unfortunately no good option to autofix that issue. 
The best options are:
- put a function in front if it and apply a security rule equivalent to RLS in front of the function
- put a view in front of the materialized view and apply a rules like where `uid = auth.uid()` directly in the view

Neither of these can be done automatically. 

Would you mind updating the PR with these options?

marsou001:
Thank you for the review!

You're right, I didn't take into account that some users might need the access to their Materialized views via Data API.

I'll remove the autofix and update the modal's message to showcase those two option you mentioned. The autofix is supposed to be automatic and simple that it can be applied with a button's click.

charislam (approved):

PR #40809, opened 2025-11-26
Fix Users page not invalidating data when switching search mode

joshenlim:
## Context

Realised that users page wasn't invalidating the data when switching search modes. Can be reproduced by:
- Ensure that page starts on "Email address" as search mode
- Search for "example.com"
- Should return 0 results since the optimized search does a prefix search
- Now switch to unified search
- Still 0 data (although notice that the count updates)

Happening due to query key for users-infinite-query not considering `column` + we're unnecessarily passing a query key into `executeSql` (might have been incorrectly set during the RQ v5 migration)

## To test

- [ ] Verify the above steps and ensure that there's data at the last step when switching to unified search


alaister (approved):
LGTM!

- Tested on preview

PR #40811, opened 2025-11-26
Fix URL when cmd/ctrl clicking files and vector bucket row

joshenlim:
## Context

As per PR title - cmd/ctrl clicking on the bucket row is supposed to open the bucket in a new tab, but it currently leads to a 404 due to missing `BASE_PATH`


alaister (approved):
LGTM!

- Tested on preview

PR #40813, opened 2025-11-26
Chore/fix storage upload sanitize duplicate name

joshenlim:
## Context

Changes here cover an edge case to have `sanitizeNameForDuplicateInColumn` extract the extension of a file if the file has `.` in its name - e.g `file.md.mdx`

In this case - the file name should be `file.md` and the extension is `mdx`

## To test
Just an example:
- [ ] Have a `md` file named `test.md.mdx`
  image
- [ ] Upload that file to a bucket twice in the storage explorer
- [ ] First file should retain the original name - `test.md.mdx.md` (when you name a file with multiple `.`s, MacOS treats that as the file name, and the extension is "hidden")
- [ ] Second file should be `test.md.mdx (1).md`


alaister (approved):
LGTM!

- Tested on preview

PR #40824, opened 2025-11-26
fix(studio): realtime metrics

edgurgel:

YES

- Connected clients is now showing the number of connections at the time.
- Broadcast events
- Presence events
- Postgres changes events
- Rate of Channel joins (unchanged)
- Message payload size : median of payload size
- Broadcast From Database Replication Lag:  median replication from commit to broadcast
- (Read) Private Channel Subscription RLS Execution Time (might need a better name here): median time RLS execution to subscribe
- (Write) Private Channel Subscription RLS Execution Time (might need a better name here): median time RLS execution to publish


awaseem (approved):
Tested in staging, Looks good!

PR #40828, opened 2025-11-26
ref(etl): Add better badges

iambriccardo:


apps/studio/components/layouts/DatabaseLayout/DatabaseMenu.utils.tsx
```diff
@@ -110,6 +112,7 @@ export const generateDatabaseMenu = (
                 name: 'Replication',
                 key: 'replication',
                 url: `/project/${ref}/database/replication`,
+                label: enablePgReplicate ? 'PRIVATE ALPHA' : 'ENROLL',
```
joshenlim: could stick to alpha (no need to specify private or public) since only those who has access will see this
otherwise no need a badge (can remove enroll)

```suggestion
                label: enablePgReplicate ? 'ALPHA' : undefined,
```

joshenlim (approved):
Verified changes locally!

PR #40830, opened 2025-11-26
fix(studio): assistant undefined connection string on self-hosted

alaister:
Self hosted requires a connection string (even a blank string). This change reverts a small change made in https://github.com/supabase/supabase/pull/40673 which broke the assistant on self hosted.

To test:
- make sure assistant functions on the self-hosted preview


aantti:
Fixes #40831

awaseem (approved):
LGTM

PR #40832, opened 2025-11-26
fix(studio): z-index of layout sidebar

kemaldotearth:

YES

We had this (hopefully) one off bug of z-index's clashing. The last column on our Extensions table inside Database is also `fixed` and had a higher `z-index` priority. I made sure for the `<ScrollTable />` component, the last fixed column has a lower `z-index` than dialog/modals and anything that requires overlaying the page, like the layout sidebar.

See bug here:

https://github.com/user-attachments/assets/c4515dae-25f6-4e72-9850-7ec7723a226b


joshenlim (approved):

packages/ui/src/components/ShadowScrollArea/index.tsx
dnywh, line 1: I feel like we’re close to a breaking point with all our stacking values. Or at least, issues like this will keep cropping up.

Maybe it’s time to either write up a Supabase stacking context list in the design system. Or explore [alternatives like `isolation`](https://www.joshwcomeau.com/css/stacking-contexts/#airtight-abstractions-with-isolation-6).

Or maybe that’s all overkill and it’s more efficient to squash bugs like this when we see them...

dnywh (approved):
Verified on staging!

packages/ui/src/components/ShadowScrollArea/index.tsx
kemaldotearth: Perhaps like the utility classes you were doing for text, we can attempt something with `isolate` or just have 3-4 layers classes and designate uses for them? I could imagine like `base`, `message`, `full-screen` (better naming than that obviously and stuff that doesn't clash with TW primitives). 

I guess it'll always be tricky with `z-index` inside different components and areas. Though perhaps we could set ourselves a hard rule of never going above `z-20` for components and if you find yourself needing different depths, use arbitrary classes like `z-[21]` etc.

Probably frontend has an idea on how they'd like to approach this :)

PR #40843, opened 2025-11-26
fix(editor): close previous and next page modals on confirm

### LINKED ISSUE #40842: In studio editor pagination, next and previous page modals keep showing "loading" state
## Describe the bug

If you select some rows and then try to go to next or previous page, a confirmation modal will popup to ask if you're sure you want to leave this page, because the above mentioned rows will be unselected.

If you click confirm, the rows will indeed be unselected, and the navigation to the next or previous page will be successful, but the confirmation modal won't close, and "confirm" and "cancel" buttons will still show the "loading" state (spinning icon), unless you close the modal manually.

## To Reproduce

1. Go to `/dashboard/project/${ref}/editor/$[id}
2. Select a few rows
3. On the pagination on the footer, navigate to next or previous page.
4. Upon the opening of the confirmation modal, click "confirm".
5. The modal will be stuck in the "loading" state unless you close it manually, but the navigation will be successful.

## Expected behavior

Either that the confirmation modal closes on its own, or the "loading" state stops and some UI component showing that the pagination is successful.

Personally, I prefer the first approach.

## Screenshots

https://github.com/user-attachments/assets/f9ae0cae-8042-4dce-a671-f27aeefc4889

marsou001:
Fixes #40842 

Before:

https://github.com/user-attachments/assets/14e326f2-57c7-4861-b4f1-07494a669857

After:

https://github.com/user-attachments/assets/93112996-91ee-466c-8664-86a10caa68d8

As I mentioned in the issue above, I prefered closing the modal after success, but after looking at the code of `ConfirmationModal`, I realized that was the only option, since without supplying a `loading` prop to the component, it will [hang in the `loading_= true` state](https://github.com/supabase/supabase/blob/master/packages/ui-patterns/src/Dialogs/ConfirmationModal.tsx#L71), [unless it is closed](https://github.com/supabase/supabase/blob/master/packages/ui-patterns/src/Dialogs/ConfirmationModal.tsx#L76).


awaseem:
Nice this is a great fix!

marsou001:
Happy to help!

I just noticed that when typing a page number directly while some rows are selected, no confirmation modal is shown.

I'm thinking I should consolidate all confirmation modals to just one that will handle all page changes, whether by arrows (previous, next) or directly through the input.

awaseem (approved):
LGTM

PR #40849, opened 2025-11-26
fix: Refactor the temp API keys creation for oAuth Server apps

ivasilov:
This PR fixes a bug with using temp API keys for oAuth Server apps. 

Previously, the temp API key would expire and break the UI. This is now fixed.


apps/studio/state/storage-explorer.tsx
awaseem: @ivasilov should we replace this with your new method?

awaseem (commented):
Minor comments

awaseem (approved):
LGTM. This is good, but if you decide to refactor more ping me so I test

apps/studio/state/storage-explorer.tsx
ivasilov: You're right, i've replaced it.

PR #40857, opened 2025-11-27
table create performance

SaxonF:
Increases performance when creating tables by combining all actions into a single executeSql.


apps/studio/data/tables/table-retrieve-query.ts
```diff
@@ -34,7 +35,7 @@ export type RetrieveTableResult = Awaited<ReturnType<typeof getTable>>
 export type RetrieveTableError = ResponseError
 export type RetrievedTableColumn = NonNullable<RetrieveTableResult['columns']>[number]
 
-export const useTablesQuery = <TData = RetrieveTableResult>(
+export const useTableQuery = <TData = RetrieveTableResult>(
```
joshenlim, line 38: realised this was a very confusing name as we already have a `useTablesQuery`, but this one retrieves a single table

joshenlim (approved):
Verified changes locally and also went through the generated test - looks good!

PR #40858, opened 2025-11-27
Analytics bucket support creating bucket tables in dashboard

joshenlim:
## Context

Dependent on Infra PR: https://github.com/supabase/infrastructure/pull/27868 
NOTE: Ensure that API changes have been deployed before merging (From 01 Dec onwards)

Adds support to create analytic bucket tables via the dashboard - changes are behind a feature flag (`analyticsBucketsTableCreation`)
Also updates all iceberg related react queries to use the new management API endpoints, so we're no longer directly pinging the iceberg catalog (which also does away with the need to use temporary API keys here)

## Changes involved
- "Create table" CTA now opens a sheet instead of the Pyiceberg instructions dialog, although that action is still available via a dropdown menu as such:
  image

- This opens up a sheet where you can create a bucket table with the relevant fields:
  image

- New namespaces can also be created via this sheet:
  image

- For column types that require additional parameters like `decimal` and `fixed` - we'll render the fields for those additional parameters as such:
  image

- All other iceberg operations should still be functional, e.g: Deleting namespace, deleting namespace tables, etc

- Also added "Insert data" CTA for each foreign table row which opens up a dialog on how to insert data into the namespace table via the foreign table (Arguarbly not a full copy paste UX as the dashboard doesn't have context of what columns there are within the namespace table, so users have to figure it out e.g via a select statement)
  image
  image

## To test
Probably can try on a new project:
- [ ] Verify that you can create an analytics bucket table
  - [ ] Into an existing namespace
  - [ ] Into a new namespace
- [ ] Verify that the rest of flow works (e.g "Query from Postgres", "Update schema tables")
- [ ] Verify that the "Insert data" CTA works


alaister (approved):
LGTM!

- Tested on preview

PR #40860, opened 2025-11-27
Clean up security notifications feature flag

joshenlim:
## Context

We're completing the incremental roll out of the Auth Security Notifications feature so removing the feature flag and having the UI as the default now. This means that "Emails" will be under the "Notifications" Section in the Auth menu from here


alaister (approved):
LGTM!

- Tested on preview

PR #40865, opened 2025-11-27
Add config cat custom attribute is_staff

joshenlim:
## Context

As per PR title - to better allow us to create a config cat segment for just internal staff


packages/common/configcat.ts
```diff
@@ -64,8 +64,12 @@ async function getClient() {
   }
 }
 
-export async function getFlags(userEmail: string = '', customAttributes?: Record<string, string>) {
+export async function getFlags(userEmail: string = '', _customAttributes?: Record<string, string>) {
```
alaister: very nit, but since `customAttributes` is public, the underscore makes it seem like a private property that shouldn't be used. I'd opt to make this one without the underscore.
```suggestion
export async function getFlags(userEmail: string = '', customAttributes?: Record<string, string>) {
```

packages/common/configcat.ts
```diff
@@ -64,8 +64,12 @@ async function getClient() {
...
   const client = await getClient()
+  const customAttributes = {
+    ..._customAttributes,
```
alaister: and this one with the underscore (you'll need to modify the code below to use this internal variable too.
```suggestion
  const _customAttributes = {
    ...customAttributes,
```

alaister (approved):
LGTM!

PR #40875, opened 2025-11-27
fix: sentry errors

awaseem:

YES

Lots of sentry errors that need fixing, mostly due to `any` footguns. Will create a ratchet rule to clean that up

### DIFF
--- a/apps/studio/components/grid/components/editor/BooleanEditor.tsx
+++ b/apps/studio/components/grid/components/editor/BooleanEditor.tsx
@@ -32,7 +32,7 @@ export const BooleanEditor = <TRow, TSummaryRow = unknown>({
       size="small"
       onBlur={onBlur}
       onChange={onChange}
-      defaultValue={value == null ? 'null' : value.toString()}
+      defaultValue={!!value ? value.toString() : 'null'}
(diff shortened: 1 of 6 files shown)


apps/studio/components/grid/components/editor/BooleanEditor.tsx
```diff
@@ -32,7 +32,7 @@ export const BooleanEditor = <TRow, TSummaryRow = unknown>({
       size="small"
       onBlur={onBlur}
       onChange={onChange}
-      defaultValue={value === null ? 'null' : value.toString()}
+      defaultValue={value == null ? 'null' : value.toString()}
```
joshenlim: hmm any reason why we're shifting from `===` to `==`?

joshenlim: ahh found the Sentry issue - `TypeError: Cannot read properties of undefined (reading 'toString')`

this is just personal preference - i'd like to ideally use `===` as much as possible throughout the repo and not mix with `==`
reckon we could probably do `!!value` check here instead

joshenlim (approved):
Did some basic smoke testing for the changed files on staging preview - LGTM!

PR #40602, opened 2025-11-19
Feature/lint action

SaxonF:
image

We want to start gradually making it easier for people to resolve advisor issues without having to leave advisor. We already do some aspect of this for index advisor. The first step is allowing people to enable RLS on a table directly from the advisor for public rls issues. 

This PR introduces an optional action prop to the lint info map which renders a component to perform some sort of an action. In this case just a simple enable button with a confirm modal, copying the logic from our policies page.

PR #40881, opened 2025-11-28
Generate policies experiment

SaxonF:
image

We want to test whether including the ability to generate policies as part of table creation has a positive impact on RLS enable and creation rates without introducing much more friction.

The way it works:
- If user is part of experiment we show an policies section in table create/edit
- A user can generate starting policies when creating new tables
- Existing tables show any existing policies but can't generate for now
- For generation, if the new table or any of its foreign relationships (recursively) have a relationship with auth.users than we use that to generate programmatically, if not we use AI but only if permissions allow.

Also:
- I've removed realtime and RLS toggles from table create as part of this work. The primary reason was to reduce complexity of the form. We want to encourage RLS being enabled AND we want people to move towards broadcasting from database so removing these seems aligned with our other work.

To test:
- Will need to enable experiment `tableCreateGeneratePolicies`
- Create new table with a foreign key assuming there is a relationship with auth.users somewhere
- Hit generate and see policies instantly
- Create a new table with no foreign keys OR a more complicated permissions set up
- Hit generate and notice it takes a while to generate (hitting ai policy endpoint)
- Save table and make sure policies are generated
- Edit a table that has existing policies and notice they appear in panel
- Edit a table that has RLS disabled and notice an alert to enable

## Summary by CodeRabbit

* **New Features**
  * Automatic RLS policy generation (programmatic via FK paths and AI-assisted) integrated into table create/duplicate/update flows with optional preview and failure summaries.
  * New RLS management UI: policy list with SQL previews, empty-state AI generator, and Toggle RLS control.

* **Backend / API**
  * Added AI-backed policy generation endpoint and client for SQL policy suggestions.

* **Tests**
  * Comprehensive tests covering programmatic, AI, and orchestration flows.

* **Telemetry**
  * New events and a has_generated_policies flag to track exposure, generation, creation, and removal of generated policies.


apps/studio/components/interfaces/Auth/Policies/Policies.utils.ts
```diff
@@ -151,3 +155,277 @@ export const createPayloadForUpdatePolicy = (
...
+        target_column_name: fk.target_columns[i],
+      }))
+    )
+}
+
+/**
+ * BFS to find shortest path from table to auth.users via foreign key relationships.
+ * Returns null if no path exists within maxDepth.
+ */
+const findPathToAuthUsers = (
```
joshenlim: this particular method feels overly complicated at first glance and i can't really tell what's it meant to do

what does `path` mean here?
`const path = findPathToAuthUsers(table, foreignKeyConstraints)`

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,361 @@
...
+  generatedPolicies = [],
+  onGeneratedPoliciesChange,
+}: RLSManagementProps) => {
+  const { data: project } = useSelectedProjectQuery()
+  const [isGenerating, setIsGenerating] = useState(false)
+  const [generateFailed, setGenerateFailed] = useState(false)
+  const { includeSchemaMetadata } = useOrgAiOptInLevel()
+  const track = useTrack()
+  const isExistingTable = !!table && !isNewRecord && !isDuplicating
+  const rlsEnabled = isRlsEnabled ?? true
```
joshenlim: this should default to `false` -> the consequence of incorrectly defaulting to `true` is greater as it's incorrectly indicating that the table is secure

pamelachia:
For some reason, I can't reach a point to generate policies and keep getting this even after adding new columns to the new table (since the error says I need to update my table schema). I then proceeded to rage click the button 😓 how should i go about this so the policies are actually generated?  

Screenshot 2025-12-02 at 4 40 45 AM

SaxonF:
@pamelachia issue with staging for new endpoint, looking into it

pamelachia (approved):
All good on a telemetry pov. Can infer abandonment from users who clicked "generate" but never fired "created".

apps/studio/components/interfaces/Auth/Policies/Policies.utils.ts
```diff
@@ -151,3 +155,277 @@ export const createPayloadForUpdatePolicy = (
...
+  )
+
+  const queue: { table: { schema: string; name: string }; path: Relationship[] }[] = [
+    { table: startTable, path: [] },
+  ]
+  const visited = new Set<string>()
+  visited.add(`${startTable.schema}.${startTable.name}`)
+
+  while (queue.length > 0) {
+    const { table, path } = queue.shift()!
```
joshenlim: try to avoid implicit non-null assertions with `!` - opt to add checks instead

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
```diff
@@ -129,6 +134,7 @@ export const SidePanelEditor = ({
   const snap = useTableEditorStateSnapshot()
   const tabsSnap = useTabsStateSnapshot()
   const [_, setParams] = useUrlState({ arrayKeys: ['filter', 'sort'] })
+  const { selectedSchema } = useQuerySchemaState()
```
joshenlim: unused - can remove

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
```diff
@@ -165,10 +171,18 @@ export const SidePanelEditor = ({
     projectRef: project?.ref,
     connectionString: project?.connectionString,
   })
+  const { data: allTables } = useTablesQuery({
```
joshenlim: unused - can remove

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,401 @@
...
+        ? 'Row Level Security has been enabled for this table.'
+        : 'Row Level Security has been disabled for this table.'
+    )
+  }
+
+  const handleRlsToggleError = (error: ResponseError) => {
+    toast.error(error.message ?? 'Unable to update Row Level Security for this table.')
+  }
+
+  const renderEnableRlsButton = () => {
```
joshenlim: i'd opt to avoid declaring functions to render UI like this from a readability POV - could either declare a new component if the logic is a bit complex, or just declare directly in the final return statement

otherwise i'd need to jump around in the code to see what is being rendered 😅

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,401 @@
...
+        onSuccess={handleRlsToggleSuccess}
+        onError={handleRlsToggleError}
+        className="w-fit mt-4"
+      />
+    )
+  }
+
+  const disablePoliciesList = isExistingTable && !rlsEnabled
+
+  const renderPolicies = () => {
```
joshenlim: this could probably be declared in its own component for e.g

joshenlim: or all these could be in `PolicyList`

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,401 @@
...
+    if (!hasPolicies) {
+      if (isNewRecord && !isDuplicating) {
+        const getFailureMessage = () => {
+          if (!generateFailed) {
+            return {
+              title: 'Generate starting policies',
+              description:
+                'Starter policies are generated from your table relationships to auth.users—or, if none exist, using AI.',
```
joshenlim: extra `-` between auth.users and or

joshenlim: could use text inline code for auth.users too

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,401 @@
...
+          description="Your table is currently accessible by anyone on the internet. We recommend enabling RLS to restrict access."
+          actions={renderEnableRlsButton()}
+        />
+      )}
+
+      <Card
+        aria-disabled={disablePoliciesList}
+        className={cn(disablePoliciesList && 'opacity-50 pointer-events-none')}
+      >
+        {renderPolicies()}
```
joshenlim: this is quite confusing here - `renderPolicies` is doing too much IMO as it's also controlling the empty state, but within this return statement its no longer the empty state

joshenlim: shouldn't this just render PolicyList directly?

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,401 @@
...
+
+  return (
+    <div>
+      <div className="flex items-center mb-4 gap-2">
+        <div className="flex-1">
+          <h3>Policies</h3>
+          <p className="text-sm text-foreground-lighter">
+            Set rules around who can read and write data to this table
+          </p>
+        </div>
```
joshenlim: we're kinda duplicating this header - could be a parent component

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,401 @@
...
+        <AIOptInModal visible={isOptInModalOpen} onCancel={() => setIsOptInModalOpen(false)} />
+      </div>
+    )
+  }
+
+  return (
+    <div>
+      <div className="flex items-center mb-4 gap-2">
+        <div className="flex-1">
+          <h3>Policies</h3>
```
joshenlim: nit: for consistency based on the other sections in the table editor, could use h5 instead

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,401 @@
...
+  generatedPolicies = [],
+  onGeneratedPoliciesChange,
+}: RLSManagementProps) => {
+  const { data: project } = useSelectedProjectQuery()
+  const [isGenerating, setIsGenerating] = useState(false)
+  const [generateFailed, setGenerateFailed] = useState(false)
+  const [isOptInModalOpen, setIsOptInModalOpen] = useState(false)
+  const { includeSchemaMetadata } = useOrgAiOptInLevel()
+  const track = useTrack()
+  const isExistingTable = !!table && !isNewRecord && !isDuplicating
```
joshenlim, line 50: would this be the same as the inverse of the prop `isNewRecord`?
feels like we can remove the use of this variable here

apps/studio/components/ui/ToggleRlsButton.tsx
joshenlim, line 1: do we foresee that we'll be using this component anywhere else? it feels specific to RLSManagement atm so i'm not sure if we need to declare this component in `ui`

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,401 @@
...
+
+  const renderEnableRlsButton = () => {
+    if (!project || !table || !isExistingTable) return null
+    return (
+      <ToggleRlsButton
+        tableId={table.id}
+        type="default"
+        size="tiny"
+        schema={table.schema ?? schema}
+        tableName={table.name}
```
joshenlim: we could just pass the table itself, rather than passing individual properties

apps/studio/components/ui/ToggleRlsButton.tsx
joshenlim: reckon this component can be simplified btw - e.g `showDefaultConfirmModal` isn't being passed as a prop from where it's declared

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/TableEditor/RLSManagement/RLSManagement.tsx
```diff
@@ -0,0 +1,401 @@
...
+  onGeneratedPoliciesChange,
+}: RLSManagementProps) => {
+  const { data: project } = useSelectedProjectQuery()
+  const [isGenerating, setIsGenerating] = useState(false)
+  const [generateFailed, setGenerateFailed] = useState(false)
+  const [isOptInModalOpen, setIsOptInModalOpen] = useState(false)
+  const { includeSchemaMetadata } = useOrgAiOptInLevel()
+  const track = useTrack()
+  const isExistingTable = !!table && !isNewRecord && !isDuplicating
+  const rlsEnabled = isRlsEnabled ?? false
```
joshenlim: could declare `isRlsEnabled` to have a default value of `false` so we don't need an additional variable

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
```diff
@@ -591,6 +593,7 @@ export const createTable = async ({
         method: 'table_editor',
         schema_name: payload.schema,
         table_name: payload.name,
+        has_generated_policies: hasGeneratedPolicies,
```
joshenlim: i find it odd that we're just passing `hasGeneratedPolicies` just for the `sendEvent`

reckon it'll be better that we add the policy creation logic here as part of `createTable`?

apps/studio/data/ai/sql-policy-mutation.ts
```diff
@@ -0,0 +1,75 @@
...
+  const result = await fetchPost<SqlPolicyGenerateResponse>(`${BASE_PATH}/api/ai/sql/policy`, {
+    tableName,
+    schema,
+    columns,
+    projectRef,
+    connectionString,
+    orgSlug,
+    message,
+  })
+  if ('error' in result) throw result
```
joshenlim: should be throw result.error?

joshenlim:
Pushed up a round of changes which addresses all the feedback that I've left ^
i'll check in again tomorrow to do a round of testing 🙂 🙏

apps/studio/components/ui/ToggleRlsButton.tsx
SaxonF: @joshenlim this was mostly taken from @charislam work on this PR https://github.com/supabase/supabase/pull/40602 . In that PR it's also used for policies, lint, table editor I think.

joshenlim:
Done a first round of review for the following:

Hosted
- If flag is on
  - [x] Generate policies via AI (No columns with FK to auth.users)
  - [x] Create table with generated AI policies
  - [x] Tested error handling for AI policy generation
  - [x] Generate programmatic policies (Have a column with an FK to auth.users)
  - [x] Create table with generated programmatic policies 
  - [x] Tested error handling for policy creation during create table step
- If flag is off: 
  - [x] Everything should be status quo

Self Hosted
- [x] Everything should be status quo

@SaxonF just reiterating here RE a UX issue:
It’s currently not clear from the Table Editor side panel now that no policies = no data can be queried (which has been a common footgun) given that we're removing this callout (esp if we're moving towards secure by default)
image

Reckon we should address this before making this live for users

apps/studio/components/interfaces/Auth/Policies/Policies.utils.ts
```diff
@@ -151,3 +155,281 @@ export const createPayloadForUpdatePolicy = (
...
+      (r) =>
+        `join ${r.target_table_schema}.${r.target_table_name} on ${r.target_table_schema}.${r.target_table_name}.${r.target_column_name} = ${r.source_schema}.${r.source_table_name}.${r.source_column_name}`
+    )
+    .join('\n  ')
+
+  return `exists (
+  select 1 from ${firstTarget}
+  ${joins}
+  where ${firstTarget}.${first.target_column_name} = ${source}.${first.source_column_name}
+  and ${last.source_schema}.${last.source_table_name}.${last.source_column_name} = (select auth.uid())
```
alaister: Do these columns need to wrapped in `"`

joshenlim: great point - we should use `ident` from `pg-meta`

apps/studio/components/interfaces/Auth/Policies/Policies.utils.ts
```diff
@@ -151,3 +156,287 @@ export const createPayloadForUpdatePolicy = (
...
+    })
+    .join('\n  ')
+
+  return `exists (
+  select 1 from ${firstTarget}
+  ${joins}
+  where ${firstTarget}.${ident(first.target_column_name)} = ${source}.${ident(first.source_column_name)}
+  and ${ident(last.source_schema)}.${ident(last.source_table_name)}.${ident(last.source_column_name)} = (select auth.uid())
+)`
+}
```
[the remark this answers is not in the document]
joshenlim: haha good catch :good-clanker:

joshenlim (approved):

PR #40890, opened 2025-11-28
feat: support new platform plan

kevcodez:
Forward compatible changes to support new platform plan (similar handling to Enterprise)


ecktoteckto (approved):
Tested with localstack

PR #40900, opened 2025-11-28
feat: table heading sorting

fsansalvadore:

Extend table with TableHeadSort to add sorting to table headers.

## What is the new behavior?

https://github.com/user-attachments/assets/751523f5-c08e-457d-8d89-4a9acc0d9332

### DIFF
--- a/apps/studio/components/interfaces/Auth/OAuthApps/OAuthAppsList.tsx
+++ b/apps/studio/components/interfaces/Auth/OAuthApps/OAuthAppsList.tsx
@@ -31,6 +31,7 @@ import {
   TableCell,
   TableHead,
   TableHeader,
+  TableHeadSort,
   TableRow,
 } from 'ui'
 import ConfirmationModal from 'ui-patterns/Dialogs/ConfirmationModal'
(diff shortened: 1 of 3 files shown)


apps/studio/components/interfaces/Auth/OAuthApps/OAuthAppsList.tsx
```diff
@@ -229,26 +292,54 @@ export const OAuthAppsList = () => {
...
+                    <TableHeadSort column="name" currentSort={sort} onSortChange={handleSortChange}>
+                      Name
+                    </TableHeadSort>
+                  </TableHead>
                   <TableHead>Client ID</TableHead>
-                  <TableHead>Client Type</TableHead>
-                  <TableHead>Registration Type</TableHead>
-                  <TableHead>Created</TableHead>
+                  <TableHead>
+                    <TableHeadSort
```
kemaldotearth, line 299: Small question, does the order of the label and the sorting matter or does it always appear on the right side of the header column label?

kemaldotearth (approved):
:shipit: LGTM. Only thing I'd add is, do we want to handle DataGrid's separately? we probably need to standardise the sorting there too.

This is much needed, already going to add it to scoped pat's.

apps/studio/components/interfaces/Auth/OAuthApps/OAuthAppsList.tsx
fsansalvadore: Always on the right here. Would you like to have it somewhere else in some cases?

PR #40902, opened 2025-11-28
chore: add realtime troubleshoot guides

filipecabaco:

YES

Add 3 troubleshooting guides:
* Using logger to debug Realtime issues
* Heartbeat explanation on how to use callbacks and webworkers
* How to handle too many channel errors and potential causes

### DIFF
--- a/apps/docs/content/troubleshooting/realtime-debugging-with-logger.mdx
+++ b/apps/docs/content/troubleshooting/realtime-debugging-with-logger.mdx
@@ -152,7 +152,7 @@ channel.subscribe((status, err) => {
       payload: { text: 'Hello' },
     })
   } else {
-    console.error({status, err})
+    console.error({ status, err })
   }
 })
 
--- a/apps/docs/content/troubleshooting/realtime-debugging-with-logger.mdx
+++ b/apps/docs/content/troubleshooting/realtime-debugging-with-logger.mdx
@@ -15,6 +15,7 @@ The Realtime client provides built-in logging to help you debug connection issue
 - **`logLevel`**: Controls which messages the server logs (`info`, `warn`, or `error`)
 
 By default, logging is disabled. When enabled, you can see:
+
 - Messages sent and received
 - Connection events (connect, disconnect, errors)
 - Heartbeat status
@@ -32,8 +33,8 @@ const supabase = createClient(SUPABASE_URL, SUPABASE_KEY, {
   realtime: {
     logger: (kind, msg, data) => {
       console.log(`${kind}: ${msg}`, data)
-    }
-  }
+    },
+  },
 })
 ```
 
@@ -45,8 +46,8 @@ const supabase = createClient(SUPABASE_URL, SUPABASE_KEY, {
     logLevel: 'info', // 'info' | 'warn' | 'error'
     logger: (kind, msg, data) => {
       console.log(`[${kind}] ${msg}`, data)
-    }
-  }
+    },
+  },
 })
 ```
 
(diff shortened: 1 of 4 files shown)


apps/docs/content/troubleshooting/realtime-debugging-with-logger.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+await channel.subscribe()
+
+// Send a message
+await channel.send({
+  type: 'broadcast',
+  event: 'message',
+  payload: { text: 'Hello' },
+})
+
```
edgurgel, line 158: await subscribe? 🤔 

Should we also do 

```
  if (status === 'SUBSCRIBED') { 
```

and then send?

Or do you want to send regardless? If so then should it be `httpSend` ?

apps/docs/content/troubleshooting/realtime-too-many-channels-error.mdx
```diff
@@ -0,0 +1,242 @@
...
+The most common cause is accidentally creating channels without cleaning them up, especially in React applications. This happens when:
+
+{/* supa-mdx-lint-disable Rule003Spelling */}
+- Components create channels on every render without unsubscribing
+- `useEffect` runs multiple times due to missing or incorrect dependencies
+- Components unmount without cleaning up their channels
+- Development mode in React (StrictMode) causes effects to run twice
+{/* supa-mdx-lint-enable Rule003Spelling */}
+
+Each time you call `supabase.channel('topic').subscribe()`, a new channel is created unless you properly clean it up.
```
edgurgel, line 31: Is this completely true? If it's the exact same topic name it won't create a new channel though:

```typescript
  channel(topic: string, params: RealtimeChannelOptions = { config: {} }): RealtimeChannel {
    const realtimeTopic = `realtime:${topic}`
    const exists = this.getChannels().find((c: RealtimeChannel) => c.topic === realtimeTopic)

    if (!exists) {
      const chan = new RealtimeChannel(`realtime:${topic}`, params, this)
      this.channels.push(chan)

      return chan
    } else {
      return exists
    }
  }

```

apps/docs/content/troubleshooting/realtime-too-many-channels-error.mdx
```diff
@@ -0,0 +1,242 @@
...
+    }
+  }, []) // Empty dependencies because supabase is stable
+
+  return <div>Chat</div>
+}
+```
+
+**How to debug channel creation**
+
+Check how many channels your app has created:
```
edgurgel: nice

apps/docs/content/troubleshooting/realtime-too-many-channels-error.mdx
filipecabaco: I think that it's due to the useEffect where one websocket will still be active while other will be created which leads to this weirdness as the loaded code is different (that two refresh from Strict Mode from react means different JS content if I'm not mistaken)

apps/docs/content/troubleshooting/realtime-debugging-with-logger.mdx
filipecabaco: honestly await subscribed it's something we should look into ... but yeah not valid right now need to go back and fix this one

ChrisChinchilla:
@filipecabaco You're not the only one to not do this, but added some page structure as it helps human and machine navigation of the page.

ChrisChinchilla (approved):

PR #40903, opened 2025-11-28
feat(reports): add request country map

jordienr:
CleanShot 2025-12-01 at 16 35 03@2x


apps/studio/components/interfaces/Reports/renderers/ApiRenderers.tsx
```diff
@@ -385,3 +400,238 @@ const RouteTdContent = (datum: RouteTdContentProps) => (
...
+
+  const countsByIso2 = buildCountsByIso2(props.data)
+  const max = Object.values(countsByIso2).reduce((m, v) => (v > m ? v : m), 0)
+  const { resolvedTheme } = useTheme()
+  const theme = resolvedTheme === 'dark' ? MAP_CHART_THEME.dark : MAP_CHART_THEME.light
+
+  if (!!props.error) {
+    const error = (
+      typeof props.error === 'string' ? { message: props.error } : props.error
+    ) as ResponseError
```
charislam: Should we be a bit more defensive about the typing here? `AlertError` expects `error` to be `{ message: string } | null`, so it could potentially crash if this cast is wrong

charislam (approved):

Issue #17823, opened 2023-09-30
Careers Page is slow and theme toggling shows error


mozi47:
# Bug report

The Careers Page is slow, and takes time to load.

Also when toggling dark and light the on Careers Page, this error appears.  
![image](https://github.com/supabase/supabase/assets/20722868/052581e2-8688-489e-88cb-efc568fa034d)

nwaliaez:
Hey @mozi47 
I tried it worked for me
Steps - 
Open link > https://supabase.com/careers > Scroll to the very bottom > click on theme icon on the bottom right side.
Could you please help if I am missing something ?

shareef99:
I am using Microsoft Edge and the career page is really slow.

mozi47:
On mobile it works fine. Try on desktop. And try on edge and firefox. On chrome sometime it works and some time not.

miguelcarvalho13:
The slowness when scrolling the page up/down also happens to me on a windows laptop with an integrated gpu (both Chrome and Edge). My bet was that the globe component could be the root cause of the issue. So I tried running the careers page locally without the cobe globe and it indeed runs way more smoothly. Another thing I've noticed is that running the careers page with the globe but without the markers also feels ok.

There is an issue in the cobe repository related to that: https://github.com/shuding/cobe/issues/78.

spirit-124:
Hey @mozi47 I don't  know if  I am mistaken at something but it did worked for me.

I tried it By:-
Open link > https://supabase.com/careers > Scroll to the very bottom > click on theme icon on the bottom right side.

hichemfantar:
Can confirm that switching themes breaks the website with a client side error.

burggraf:
I've tested this on both chrome and edge and can't duplicate the error.  Can you please confirm it's still happening to you?

mozi47:
Yes the issue still exist. May be I havent add detail information about the error.

On desktop, on chrome browser
Navigatge to career page 
Scroll down to the footer section
Change the theme and it will show the 404 error.

The page load is also slow.

mozi47:
@nwaliaez @burggraf 

Sorry to record on a mobile phone. for some reason, obs was not working properly,

On Edge:

https://github.com/supabase/supabase/assets/20722868/a5ce29b0-386b-40d4-b416-c8977de6dc9b

On Chrome:

https://github.com/supabase/supabase/assets/20722868/00e5f006-2cd6-4bc3-bbaa-54e0dcd48414


[...]

PR #40908, opened 2025-11-29
fix: update cobe version to reduce lag on /career and remove height/width assignment per frame

Rpaudel379:
Update cobe version to reduce lag on `/career` and remove height/width assignment per frame

YES

This PR reduces lag and high GPU usage on the `/careers` page by:
- updating the `cobe` package to `0.6.5` which does not freeze.
- remove redundant height and width assignments on every frame during globe rendering

- Currently `/careers` page freezes and consumes high GPU resource. The current `cobe` version  `0.6.2` appears to cause lagging on some devices i.e larger screens while on mobile it runs smoothly. 
- Also, height and width are being applied on each frame rendering, which is not necessary and likely contributing to freeze.

## What is the new behavior?

- Updating version to `0.6.5` or to a lower version like `0.4.2` significantly reduces lag. 
- Remove code that was assigning height and width on each frame render.
- `/careers` page now runs much more smoothly for affected users.

## Additional context

This issue appears inconsistently, only some users experience lag on `/careers`

There is one closed issue for this which never resolved.
- https://github.com/supabase/supabase/issues/17823

There is also an existing PR to fix this:
https://github.com/supabase/supabase/pull/40131
but that does not resolves the unnecessary per frame width and height updates

## Summary by CodeRabbit

* **Refactor**
  * Optimized Globe component rendering performance.

* **Chores**
  * Updated cobe package dependency to the latest minor version.


stylessh:
hey, can you remove that big lockfile diff? should be like ±10 lines of code with the updated lib.

Rpaudel379:
> hey, can you remove that big lockfile diff? should be like ±10 lines of code with the updated lib.

Hi, I have removed `pnpm-lock.yaml`, its no longer part of the diff. Only the intended code remains. Thanks!

stylessh (approved):
Thanks a lot for the contribution!

Rpaudel379:
> Thanks a lot for the contribution!

Thanks! Appreciate the review

PR #40932, opened 2025-12-01
Fix clone object assistant

SaxonF:
Fixes an issue that prevents queries from being confirmed if page is refreshed between receiving the message and running the query. 

`Proxy object could not be cloned.`

This re-appeared after transitioning to managing chat instances at the project state level.

To test:
- Without the branch, generate a query and ask assistant to execute it
- Refresh the page
- Open assistant and try and run the query and notice error
- Do the same with the branch and notice it completes successfully

Note that this also removes an unused cached results function from state. We no longer store results after running.

### DIFF
--- a/apps/studio/state/ai-assistant-state.tsx
+++ b/apps/studio/state/ai-assistant-state.tsx
@@ -18,7 +18,7 @@ type SuggestionsType = {
   prompts?: { label: string; description: string }[]
 }
 
-export type AssistantMessageType = MessageType & { results?: { [id: string]: any[] } }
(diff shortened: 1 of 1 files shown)


joshenlim (approved):
Hmm I think i might be missing something as I'm not able to reproduce this on staging, but the code changes here LGTM nonetheless and also did a smoke test on staging preview on the assistant - approving to unblock 🙏 ✅

joshenlim:
Ahhhhhh this error surfaces only in the console log and not a UI error

Verified on staging preview that the console error no longer shows up!

PR #40937, opened 2025-12-01
chore(studio): simplify feedback form

dnywh:

UI updates. Resolves DEPR-157.

Our feedback form is highly defensive. It has several “escape hatches” for users who are actually seeking assistance rather than leaving product feedback.

This is problematic for two reasons:

1. What should be a simple feedback flow is multi-step and has a lot of markup.
2. Most importantly, all of the escape hatches take folks straight to creating a support ticket. Discord or the docs may be a better place for them.

Lastly, the escape hatch that shows a “This looks like a support ticket” admonition is quite slow to appear.

## What is the new behavior?

Removed:
- Multi-step `stage` UI to skip straight to the feedback form
- Direct link to Support ticket and Docs from feedback form

Added:
- More indicative placeholder text to denote feedback (not support) context
- Client-side detection of common support words and phrases for immediate catching via the admonition
- Cleaner reference to Help menu for Support, Docs, Discord, etc after feedback is submitted

| Before | After |
| --- | --- |
| Inventory  Nuts  Bolts  Supabase | _Not applicable. This step is skipped in the new version._ |

## Additional context

I initially lowered the `debouncedFeedback` interval from `500` to `300` but put it back after the client-side validation. That validation however caught the bulk of my testing, much sooner. I reset the debounce as a result.

### To test

- [ ] [classify.ts](https://github.com/supabase/supabase/blob/776701e41c22f76d0f8286c9acecbf7566e3001b/apps/studio/pages/api/ai/feedback/classify.ts) seems to throw the below console warning. Are we unnecessarily using a reasoning model for a simple word search?

> AI SDK Warning: The "temperature" setting is not supported by this model - temperature is not supported for reasoning models


apps/studio/components/layouts/ProjectLayout/LayoutHeader/FeedbackDropdown/FeedbackWidget.tsx
```diff
@@ -35,6 +34,81 @@ interface FeedbackWidgetProps {
...
+    /support ticket/i,
+    /contact support/i,
+    /get help/i,
+    /need assistance/i,
+    /require assistance/i,
+    /technical support/i,
+  ]
+
+  return supportPatterns.some((pattern) => pattern.test(lowerText))
+}
```
dnywh: I suppose [classify.ts](https://github.com/supabase/supabase/blob/776701e41c22f76d0f8286c9acecbf7566e3001b/apps/studio/pages/api/ai/feedback/classify.ts) could also be a home for the client-side pattern detection. I felt this was neater, as classification remains purely focused on ML classification.

apps/studio/components/layouts/ProjectLayout/LayoutHeader/FeedbackDropdown/FeedbackWidget.tsx
```diff
@@ -201,29 +279,11 @@ export const FeedbackWidget = ({ onClose }: FeedbackWidgetProps) => {
...
-            </SupportLink>{' '}
-            or{' '}
-            <a href={`${DOCS_URL}`} target="_blank" rel="noreferrer">
-              <span className="cursor-pointer text-brand-link transition-colors hover:text-brand-600">
-                see docs
-              </span>
-            </a>
-            .
-          </p>
-        </div>
```
dnywh, line 226: We don’t want to link straight to support ticket creation. Our now more “snappy” admonition is a better fallback in the case that the user writes support-type language.

apps/studio/components/layouts/ProjectLayout/LayoutHeader/FeedbackDropdown/FeedbackWidget.tsx
```diff
@@ -312,24 +372,19 @@ const ThanksMessage = ({ onClose }: { onClose: () => void }) => {
...
-          <p className="text-sm ">
-            We do not always respond to feedback. If you require assistance, please contact support
-            instead.
-          </p>
+          <div className="text-center flex flex-col">
+            <p className="text-foreground text-base">Your feedback has been sent. Thanks!</p>
+            <p className="text-sm text-balance">
+              We don’t always respond to feedback. If you require assistance, please contact support
+              via the <HelpCircle className="inline-block" size={12} aria-label="Help" /> menu
+              instead.
```
dnywh, line 306: It’s disorientating to have one dropdown open another, hence the reference to another menu via `HelpCircle`.

apps/studio/components/layouts/ProjectLayout/LayoutHeader/FeedbackDropdown/FeedbackWidget.tsx
```diff
@@ -158,33 +236,33 @@ export const FeedbackWidget = ({ onClose }: FeedbackWidgetProps) => {
     <ThanksMessage onClose={onClose} />
   ) : (
     <>
-      <div>
+      <>
         <div className="px-5 pb-4">
           <TextArea_Shadcn_
-            placeholder="It would be great if..."
-            rows={5}
+            placeholder="My idea for improving Supabase is..."
```
dnywh: More indicative placeholder text. More clearly marks this textarea as being for product feedback.

apps/studio/components/layouts/ProjectLayout/LayoutHeader/FeedbackDropdown/FeedbackDropdown.tsx
```diff
@@ -37,30 +32,7 @@ export const FeedbackDropdown = ({ className }: { className?: string }) => {
...
-                <span className="grid gap-1 text-center">
-                  <Lightbulb size="28" className="mx-auto text-warning" />
-                  <span className="text-base">Idea</span>
-                  <span className="text-xm text-foreground-lighter">to improve Supabase</span>
-                </span>
-              </Button>
-            </div>
-          </div>
-        )}
-        {stage === 'widget' && <FeedbackWidget onClose={() => setIsOpen(false)} />}
```
dnywh, line 63: Removed staging step in favour of below.

apps/studio/components/layouts/ProjectLayout/LayoutHeader/FeedbackDropdown/FeedbackWidget.tsx
joshenlim: should be alright - so we're doing 2 levels of classification here
client side as a faster route and AI as a fallback

apps/studio/components/layouts/ProjectLayout/LayoutHeader/FeedbackDropdown/FeedbackDropdown.utils.ts
```diff
@@ -57,3 +57,78 @@ export const uploadAttachment = async ({ image, userId }: UploadAttachmentArgs)
...
+ * Client-side heuristic to quickly detect obvious support requests
+ * This provides immediate feedback before the AI classification completes
+ */
+export function isLikelySupportRequest(text: string): boolean {
+  if (!text || text.trim().length === 0) return false
+
+  const lowerText = text.toLowerCase()
+
+  // Common support request patterns
+  const supportPatterns = [
```
joshenlim, line 71: rekcon its better to remove nouns from the support patterns here

e.g "i'm having trouble" vs "having trouble"

would capture more instances i think - otherwise there'll be too many cases to (attempt to) catch, given this is a best guess sort of approach

joshenlim (approved):
Verified changes on staging - think it's alright, although it's hard to ascertain whether `isLikelySupportRequest` will be able to capture intentions accurately (including false positives). `classify` takes a while before the request completes which users might have already hit "Send" before the request finishes

apps/studio/components/layouts/ProjectLayout/LayoutHeader/FeedbackDropdown/FeedbackDropdown.utils.ts
dnywh: Good thinking

dnywh:
> Verified changes on staging - think it's alright, although it's hard to ascertain whether `isLikelySupportRequest` will be able to capture intentions accurately (including false positives). `classify` takes a while before the request completes which users might have already hit "Send" before the request finishes

Thanks @joshenlim! We could lower the debounce interval for `classify` as a next step, if need be.

PR #40940, opened 2025-12-01
Joshen/fe 2180 right side sidebars too wide on laptop displays

joshenlim:
## Context

We've noticed this for a while now - table editor side nav width seems to be default to being wider than expected despite us setting the `defaultSize` to match the `minSize`

Realised that this was because we weren't passing valid props to the `ResizeablePanel`s components, in which the component will then fail to render properly (i.e. doesn't respect the `defaultSize` param being passed). TLDR the `defaultSize`s of the panels must match to render properly, e.g if panel 1's size is 20, panel 2 must be 80.

Mitigating this bit of a footgun by declaring the "main" panel's props as `const`s, then derivingt the "secondary" panel's props with `100 - x`

## To test
- [ ] Verify that default width for table editor side nav is a bit shorter + can be resized
  - Was previously defaulting to 33% (max), now should be properly defaulted to 20% or `w-64`
- [ ] Verify that default width for global panels (e.g inline SQL editor, or AI assistant, or advisor center is a bit shorter)
  - Was previously defaulting to 50% (max), now should be properly defaulted to 30%


dnywh (approved):
Tested all mentioned areas on staging. Looks good!

alaister (approved):
LGTM!

- Tested on preview

PR #40941, opened 2025-12-01
Remove GOING DOWN to determine if project is pausing

joshenlim:
## Context

We were using the project's `GOING_DOWN` status to determine is the project is pausing which I think might be old code.
Have verified that project's status will be swapped to `GOING_DOWN` when we're deleting the project ([ref](https://github.com/supabase/infrastructure/blob/develop/api/packages/core/src/lib/projects/project.ts#L667)) and that project's status will be swapped to `PAUSING` when we're pausing a project ([ref](https://github.com/supabase/infrastructure/blob/develop/api/packages/core/src/lib/projects/project-pause.ts#L51))

So safe to remove the check for `GOING_DOWN` in FE


alaister (approved):
LGTM!

- Tested on preview

PR #40947, opened 2025-12-01
fix(studio): boolean editor in table editor

alaister:
With the current version of the code, `false` gets converted to `null`. This is an alternative fix to https://github.com/supabase/supabase/pull/40875


joshenlim (approved):
Code changes LGTM! Approving to unblock but jfyi i havent tested on staging preview 🙏

saltcod (approved):

saltcod:
looks good on staging

PR #40950, opened 2025-12-01
Fix resizeable panels

joshenlim:
## Context

Bug was introduced [here](https://github.com/supabase/supabase/pull/40940) - attempted to resolve some console warnings from resizeable panel in `DefaultLayout` as we were dynamically rendering the other panel, by adjusting the size of the panel dynamically instead. While it was alright on normal viewports, because of the responsive tailwind classes applying a fixed width to the panels, hence why the panels ended up persisting on smaller viewports.

Am hence reverting to just keep the dynamic rendering - console warnings are likely safe to ignore either way in this case 🙏 

Fixes https://github.com/supabase/supabase/issues/40945

## To test

- [ ] Verify that the panel which houses inline SQL editor, assistant AI, or advisor center doesn't load with no content on smaller viewports
- [ ] Verify that the panels should only load when being opened from buttons in the top nav

### DIFF
--- a/apps/studio/components/layouts/ProjectLayout/LayoutSidebar/index.tsx
+++ b/apps/studio/components/layouts/ProjectLayout/LayoutSidebar/index.tsx
@@ -27,9 +27,9 @@ export const LayoutSidebar = ({
         id="panel-side"
         key={activeSidebar?.id ?? 'default'}
         order={order}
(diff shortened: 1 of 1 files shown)


saltcod (approved):
checked staging and looks to fix mobile issue

PR #40955, opened 2025-12-01
feat(auth): allow opting in to creating indexes on `auth.users`

fadymak:
Allows users to opt-in to improved users search performance and filtering/sorting experience by creating the necessary indexes on the `auth.users` table.

1. The users will be shown a confirmation modal to provide details about the process
2. Once confirmed, the Auth configuration is updated (`INDEX_WORKER_ENSURE_USER_SEARCH_INDEXES_EXIST`)
3. The `indexworker` process obtains an advisory lock and begins the process of creating the necessary indexes (`CONCURRENTLY`)
4. The user can monitor the process in the logs

To be able to opt-in:

1. The `auth.users` table must have less than 10k users (this will gradually be increased)
2. The feature flag `improvedUserSearch` must be enabled
3. `INDEX_WORKER_ENSURE_USER_SEARCH_INDEXES_EXIST` must be `false`

**Confirmation modal**

confirmation

**Opt-in**

opt-in

**In-progress**

in-progress


packages/common/constants/user-search-indexes.ts
```diff
@@ -0,0 +1,10 @@
+export const USER_SEARCH_INDEXES = [
+  'idx_users_email',
```
ivasilov: `pg-meta` may be a better place for this array.

apps/studio/data/auth/user-search-indexes-query.ts
```diff
@@ -0,0 +1,57 @@
+import { getIndexStatusesSQL } from '@supabase/pg-meta/src/sql/studio/get-index-statuses'
```
ivasilov: Export the new SQL code from the `index` file in `pg-meta` and just import `@supabase/pg-meta`

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -377,7 +460,26 @@ export const UsersV2 = () => {
...
+              icon={<WandSparklesIcon />}
+              className="mr-6"
+              onClick={() => setShowCreateIndexesModal(true)}
+              loading={isUpdatingAuthConfig}
+            >
+              Create indexes
+            </Button>
+          )}
+
+          {indexWorkerInProgress && (
```
ivasilov: I think this would work better as an admonition above the table. It should also have link to `/project/${projectRef}/logs/explorer?q=${encodeURI(INDEX_WORKER_LOGS_SEARCH_STRING)}`.

apps/studio/data/auth/user-search-indexes-query.ts
fadymak: Done ✅ We've got a few places that specify the complete import path to the SQL so I wasn't sure if that was the pattern we wanted to continue with.

packages/common/constants/user-search-indexes.ts
fadymak: Sounds good, applied in https://github.com/supabase/supabase/pull/40955/commits/a62fd0039d2372a0d480f83c2a58b632f5b78f93 👍

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
fadymak: Added alert at the top of the table (see PR description).

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -325,6 +353,74 @@ export const UsersV2 = () => {
...
+    userSearchIndexes?.length === pgMeta.USER_SEARCH_INDEXES.length &&
+    userSearchIndexes?.every((index) => index.is_valid && index.is_ready)
+
+  /**
+   * We want to show the improved search when:
+   * 1. The feature flag is enabled for them
+   * 2. The user has opted in (authConfig.INDEX_WORKER_ENSURE_USER_SEARCH_INDEXES_EXIST is true)
+   * 3. The required indexes are valid and ready
+   */
+  const _showImprovedSearch =
```
ivasilov, line 398: This is not needed, right?

ivasilov (approved):

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
fadymak: I have a followup PR that's based on this branch that will use this to display the improved UI

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -57,8 +69,23 @@ import {
...
+import { useUserIndexStatusesQuery } from 'data/auth/user-search-indexes-query'
+import { useAuthConfigUpdateMutation } from 'data/auth/auth-config-update-mutation'
+import { useIndexWorkerStatusQuery } from 'data/auth/index-worker-status-query'
+import { InlineLink } from 'components/ui/InlineLink'
+import Link from 'next/link'
 
 const SORT_BY_VALUE_COUNT_THRESHOLD = 10_000
+const IMPROVED_SEARCH_COUNT_THRESHOLD = 10_000
+
+const INDEX_WORKER_LOGS_SEARCH_STRING = `select id, auth_logs.timestamp, metadata.level, event_message, metadata.msg as msg, metadata.error
```
avallete, line 82: **nitpick**

Would push this to the pg-meta package, goal being to have a central place to look for SQL queries and have them properly tested.

packages/pg-meta/src/sql/studio/get-index-statuses.ts
```diff
@@ -0,0 +1,21 @@
...
+]
+
+export const getIndexStatusesSQL = () => {
+  return `SELECT c.relname as index_name, i.indisvalid as is_valid, i.indisready as is_ready
+    FROM pg_index i
+    JOIN pg_class c ON c.oid = i.indexrelid
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    WHERE n.nspname = 'auth'
+    AND c.relname IN (${USER_SEARCH_INDEXES.map(literal).join(', ')});`
+}
```
avallete, line 21: **nitpick**

Would be good to have at least some testing coverage like we do here: https://github.com/supabase/supabase/blob/eceb78199c2419692992922a79849e22e5dde3e7/packages/pg-meta/test/sql/studio/get-users-count.test.ts

packages/pg-meta/src/sql/studio/get-index-worker-status.ts
```diff
@@ -0,0 +1,11 @@
...
+
+const INDEX_WORKER_ADVISORY_LOCK_KEY = 'auth_index_worker'
+
+export const getIndexWorkerStatusSQL = () => {
+  return `SELECT EXISTS (
+    SELECT 1 FROM pg_locks
+    WHERE locktype = 'advisory'
+    AND (classid::bigint << 32 | objid::bigint) = hashtext('${INDEX_WORKER_ADVISORY_LOCK_KEY}')::bigint
+  ) as is_in_progress;`
+}
```
avallete, line 11: **nitpick**

Same, some test would be good.

avallete (approved):
A few nitpicks, non blocking, can be addressed in a follow up if necessary.

apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
fadymak: This particular query is not used by `pg-meta` — we're using it to pre-fill the log search for users to check in on the index creation progress/errors.

packages/pg-meta/src/sql/studio/get-index-statuses.ts
fadymak: These queries were ported over from the indexworker and have quite extensive coverage there since we can simulate the different states of the indexes (ref: https://github.com/supabase/auth/blob/master/internal/indexworker/indexworker_test.go)

I could definitely look into getting some of these index health queries added in a followup PR so we can have coverage here as well 👌 Thanks!

PR #40960, opened 2025-12-01
fix: added logic to check and parse valid day.js

awaseem:

YES

Bug fix: migration files are have issues that crash the dashboard when not in the proper date format. This shows unknown and handles this better 

## Demo

Screenshot 2025-12-01 at 2 57 09 PM
Screenshot 2025-12-01 at 2 37 53 PM

### DIFF
--- a/apps/studio/components/interfaces/HomeNew/ActivityStats.tsx
+++ b/apps/studio/components/interfaces/HomeNew/ActivityStats.tsx
@@ -62,7 +62,7 @@ export const ActivityStats = () => {
       return [versionDayjs.fromNow(), versionDayjs.toISOString()]
     }
 
-    return [version, version]
+    return [undefined, undefined]
   }, [latestMigration])
 
   return (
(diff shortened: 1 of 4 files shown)


apps/studio/components/interfaces/HomeNew/ActivityStats.tsx
```diff
@@ -53,6 +57,22 @@ export const ActivityStats = () => {
...
+      return [versionDayjs.fromNow(), versionDayjs.toISOString()]
+    }
+
+    return [undefined, undefined]
+  }, [latestMigration])
+
+  const hasValidVersion = versionLabel && versionTimestamp
+
+  const versionLabelText =
+    migrationsData && migrationsData.length > 0 ? 'Unknown version' : 'No migrations'
```
joshenlim: reckon just Unknown might be good here - since the label is "Last migration", might be confusing to mention "version"

joshenlim (approved):
Verified changes locally by hardcoding values in `migration-query`

PR #40966, opened 2025-12-02
Feat(www)/contribute

saltcod:
Requires the following two secrets in `apps/www/.env`:

- NEXT_PUBLIC_SUPABASE_CONTRIBUTE_URL
- NEXT_PUBLIC_SUPABASE_CONTRIBUTE_PUBLISHABLE_KEY

urls to review: 
- /contribute
- /contribute/about
- /contribute/[id]


apps/www/app/contribute/t/[id]/page.tsx
```diff
@@ -0,0 +1,42 @@
...
+    follow: true,
+  },
+}
+
+// eslint-disable-next-line no-restricted-exports
+export default async function ThreadPage({ params }: { params: Promise<{ id: string }> }) {
+  const { id } = await params
+
+  return (
+    <ContributeGuard>
```
Comment (saltcod) [inline apps/www/app/contribute/t/[id]/page.tsx:22]: Added a feature flag for this page so we can get it out without continuing to pile onto this already big PR. Will remove once's it's launched.

kemaldotearth (approved):
Only 1 question:

The filter is only on "enter" at the moment? At least for me, only works if I press enter. Initially I typed and expected to see the list filtering. Maybe some kind of indicator (and this goes globally for studio and everywhere else) within the input? I'm not sure what, but it was a bit jarring. This could be a _me_ thing though. This can be a follow up either way...

Otherwise :shipit:

PR #40968, opened 2025-12-02
feat(studio): recommended schemas for extensions

dnywh:

UI update that resolves STORAGE-346.

- The `wrappers` extension can be installed in any extension, even though things break when it _isn’t_ installed inside of the `extensions` schema

## What is the new behavior?

- New `extensionsWithRecommendedSchemas` list which includes `wrappers`
- Anything in this list will show the following `descriptionText` under the select element

| Before | After |
| --- | --- |
| Metals  Nuts  Bolts  Supabase | Metals  Nuts  Bolts  Supabase |

## Additional context

We use a client-side list for recommended schemas instead of one on the backend because:

- The `extensions` schema is user-managed and may never exist
- FDW are agnostic where `wrappers` are installed
 
Some (advanced) developers might have good reason to enable that extension in another schema. More context [here](https://linear.app/supabase/issue/STORAGE-346/prevent-installing-wrappers-in-supabase-owned-schemas#comment-a34f9d01).

### DIFF
--- a/apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
+++ b/apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
@@ -26,6 +26,11 @@ import { Admonition } from 'ui-patterns'
 
 const orioleExtCallOuts = ['vector', 'postgis']
 
+// Extensions that have recommended schemas (rather than required schemas)
+const extensionsWithRecommendedSchemas: Record<string, string> = {
+  wrappers: 'extensions',
+}
+
 interface EnableExtensionModalProps {
   visible: boolean
   extension: PostgresExtension
@@ -88,6 +93,20 @@ const EnableExtensionModal = ({ visible, extension, onCancel }: EnableExtensionM
       cancel = true
     }
   }, [visible, extension.name])
+
+  const getSchemaDescriptionText = (extensionName: string, schema: string | null | undefined) => {
+    // Prioritize defaultSchema (required/forced) over recommended schema
+    if (schema) {
+      return `Extension must be installed in the “${schema}” schema.`
+    }
+
+    const recommendedSchema = extensionsWithRecommendedSchemas[extensionName]
+    if (recommendedSchema) {
+      return `Use the “${recommendedSchema}” schema for full compatibility with related features.`
+    }
+
+    return undefined
+  }
 
   const validate = (values: any) => {
     const errors: any = {}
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
```diff
@@ -89,6 +94,20 @@ const EnableExtensionModal = ({ visible, extension, onCancel }: EnableExtensionM
     }
   }, [visible, extension.name])
 
+  const getSchemaDescriptionText = (extensionName: string, schema: string | null | undefined) => {
+    // Prioritize defaultSchema (required/forced) over recommended schema
+    if (schema) {
+      return `Extension must be installed in the “${schema}” schema.`
+    }
```
dnywh, line 101: Since `defaultSchema` forces an option, there’s no point in supporting both that and `recommendedSchema`. They are incompatible, and the former should be preferenced.

apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
```diff
@@ -89,6 +94,20 @@ const EnableExtensionModal = ({ visible, extension, onCancel }: EnableExtensionM
...
+
+    const recommendedSchema = extensionsWithRecommendedSchemas[extensionName]
+    if (recommendedSchema) {
+      return `Use the “${recommendedSchema}” schema for full compatibility with related features.`
```
dnywh, line 105: Some other ideas:

- Enabling this extension in the “extensions” schema is recommended for full compatibility.
- We recommend enabling this in the “extensions” schema.
- This extension works best when you enable it in the “extensions” schema.
- Some features expect this to be in the “extensions” schema. We recommend enabling it there for reliable behavior.

I feel like I’ve struck the best balance of brevity, reason, and action but feel free to suggest alternatives.

apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
```diff
@@ -165,13 +184,14 @@ const EnableExtensionModal = ({ visible, extension, onCancel }: EnableExtensionM
...
                     label="Select a schema to enable the extension for"
-                    descriptionText={`Extension must be installed in ${defaultSchema}.`}
+                    descriptionText={getSchemaDescriptionText(extension.name, defaultSchema)}
                   />
                 ) : (
                   <Listbox
                     size="small"
                     name="schema"
                     label="Select a schema to enable the extension for"
+                    descriptionText={getSchemaDescriptionText(extension.name, null)}
```
dnywh, line 194: I’ve made a reusable function `getSchemaDescriptionText` because an inline ternary is hard to read and not extensible for when we (should) add more recommended schemas in the future.

apps/studio/components/interfaces/Storage/VectorBuckets/CreateVectorTableSheet.tsx
```diff
@@ -327,7 +327,9 @@ export const CreateVectorTableSheet = ({ bucketName }: CreateVectorTableSheetPro
             <SheetSection className="space-y-4">
               <div className="flex items-center justify-between">
                 <label className="text-sm text-foreground">Metadata keys</label>
-                <DocsButton href={`${DOCS_URL}/guides/storage/vector/introduction`} />
+                <DocsButton
+                  href={`${DOCS_URL}/guides/storage/vector/storing-vectors#metadata-best-practices`}
+                />
```
dnywh, line 332: Sorry! Just snuck in a docs URL update.

ivasilov (approved):

PR #40765, opened 2025-11-25
Support copying by clicking string in TextConfirmationModal

joshenlim:
## Context

Nit UX improvement to support copying the confirmation string by clicking in `TextConfirmationModal`

image

PR #40971, opened 2025-12-02
Revert "Support copying by clicking string in TextConfirmationModal"

joshenlim:
Reverts supabase/supabase#40765

## To test
- [ ] Verify that you should still be able to double click on the confirmation text in the label to select it still


alaister (approved):
LGTM!

- Tested on preview

PR #40973, opened 2025-12-02
Fix sidebar param

SaxonF:
Fixes an issue where we are trying to open sidebars before they are registered e.g. via query params. This queues them up to be opened once they are registered.


joshenlim (approved):
Verified that changes work - although we should move to `nuqs` for this URL state handling

pamelachia (approved):

PR #40975, opened 2025-12-02
fix: compute upgrades for v3

kevcodez:
Currently not working as it tries to modify disk which is not supported for v3. PR unblocks compute size changes for v3 projects


sweatybridge (approved):

ecktoteckto (approved):
Tested locally

PR #40981, opened 2025-12-02
Patch org audit logs

joshenlim:
## Context

There's some issues with the both organization audit logs and profile audit logs:
- Dashboard is using the `project_ref` property from the audit logs to determine if the log's target is a project
- However, we noticed that each audit log has "duplicate" properties - `project_ref` and `ref`, and `org_slug` and `slug`
  image
- This leads to some discrepancy as FE was originally using `project_ref` which is now returning `null`, whereas `ref` has the right value instead
- A couple more factors that caused us to overlook this includes:
  - API types are also not properly typed
  - FE isn't receiving the right types from the API types codegen
    image
- Note that this is merely a patch and the proper fix needs to be done on the API end

## To test
- [ ] Verify that the audit logs on both org and account shows the right target if its supposed to be targeting a project (currently org audit logs defaults to showing the organization as the target, while profile audit logs default to just `-`)

### DIFF
--- a/apps/studio/components/interfaces/Account/AuditLogs.tsx
+++ b/apps/studio/components/interfaces/Account/AuditLogs.tsx
@@ -69,7 +69,9 @@ export const AuditLogs = () => {
     )
     ?.filter((log) => {
       if (filters.projects.length > 0) {
-        return filters.projects.includes(log.target.metadata.project_ref || '')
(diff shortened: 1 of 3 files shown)


jordienr (approved):
working now, LGTM!

Image

PR #40983, opened 2025-12-02
Disable 'New table' button in schema visualizer page if schema is pro…

### LINKED ISSUE #40982: Prevent users from adding tables to protected schemas through Schema Visualizer page
## Describe the bug

At schema visualizer page, if a schema doesn't have any table, it shows an "New Table" button that redirects to the editor page and opens the table side panel right after navigation.

This "New Table" button is present whether the schema is protected or not.

Image

Now, if the user actually goes ahead with the table creation process, an error will popup because the user is not allowed to create tables inside a protected schema like `graphql` or `realtime`, but when I tried it on `extensions`, it actually created the table...

Image

marsou001:
Fixes #40982 

I checked if the selected schema is protected, and if so, make "New Table" button disabled with a tooltip explaining why.

Screenshot (1)

This effectively prevents the navigation to the editor table and the opening of the table editor on the protected schema.

Still doesn't fix the issue of `extensions` schema though, I'll try to fix in another PR.

Edit: I made a change to check if the user is allowed to create tables as well, otherwise the button will be disabled!

Screenshot (4)

### DIFF
--- a/apps/studio/components/interfaces/Database/Schemas/SchemaGraph.tsx
+++ b/apps/studio/components/interfaces/Database/Schemas/SchemaGraph.tsx
@@ -14,11 +14,13 @@ import { LOCAL_STORAGE_KEYS, useParams } from 'common'
 import AlertError from 'components/ui/AlertError'
 import { ButtonTooltip } from 'components/ui/ButtonTooltip'
 import SchemaSelector from 'components/ui/SchemaSelector'
+import { ProtectedSchemaWarning } from 'components/interfaces/Database/ProtectedSchemaWarning'
 import { useSchemasQuery } from 'data/database/schemas-query'
 import { useTablesQuery } from 'data/tables/tables-query'
 import { useLocalStorage } from 'hooks/misc/useLocalStorage'
 import { useQuerySchemaState } from 'hooks/misc/useSchemaQueryState'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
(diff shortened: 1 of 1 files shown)


awaseem:
@marsou001 it would be better just to hide the button and update the message inside the "No tables in schema" prompt

awaseem:
We can also change the title or even hide it so its not confusing. So in the case you can't create a table its just a simple message. Thoughts?

marsou001:
My reasoning for using the button tooltip is that the user is aware that he/she can add tables, just not in this case, whether because the schema is protected, or because the user doesn't have the permission to create them.

But your idea is a good option as well, I thought about it too. If you prefer it, I can hide the button and update the message (I prefer to keep the title, I think users should know that the table is empty).

awaseem:
Lets remove and see if anyone yells at us 😆 

>But your idea is a good option as well, I thought about it too. If you prefer it, I can hide the button and update the message (I prefer to keep the title, I think users should now that the table is empty).

lets do this!

marsou001:
Just done it, but before pushing the changes, just one question: if the schema is protected, should I just display a simple one-line message, or should I stick to the coding standard and use `ProtectedSchemaWarning` component like the rest of the codebase?

Edit: I went with the simple one-line message.

awaseem (approved):
Nice work and thanks for the contribution

awaseem (approved):

awaseem:
Nice work, and as always thank you for the contribution 💚

marsou001:
Happy to help! Thank you for the review 😊

PR #40988, opened 2025-12-02
Update humans.txt

hunleyd:
Add Anna as per her Slack request


jchancojr (approved):
LGTM other than check warnings

PR #40992, opened 2025-12-02
fix: bucket delete wait for all files to be deleted

awaseem:

YES

Bug fix: poll till files are deleted then remove the bucket. Otherwise files get deleted (async operation) and bucket fails to delete. Adds a bit of delay

Workaround until we have a better endpoint ready

### DIFF
--- a/apps/studio/data/storage/bucket-util.ts
+++ b/apps/studio/data/storage/bucket-util.ts
@@ -1,0 +1,35 @@
+import { listBucketObjects } from './bucket-objects-list-mutation'
+
+const DEFAULT_INTERVAL_MS = 3000
+const DEFAULT_MAX_ATTEMPTS = 60
+
+export async function pollUntilBucketEmpty({
+  projectRef,
+  bucketId,
+  intervalMs = DEFAULT_INTERVAL_MS,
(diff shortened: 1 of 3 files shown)


apps/studio/data/storage/bucket-util.ts
```diff
@@ -0,0 +1,35 @@
+import { listBucketObjects } from './bucket-objects-list-mutation'
```
awaseem, line 1: utility if we need to change it for other products

charislam (approved):

apps/studio/data/storage/bucket-util.ts
```diff
@@ -0,0 +1,35 @@
...
+    })
+
+    if (objects.length === 0) {
+      return
+    }
+
+    await new Promise((resolve) => setTimeout(resolve, intervalMs))
+  }
+
+  throw new Error('Timeout waiting for bucket to empty')
```
ivasilov: This will be shown to the user in a toast, right? The message could be better and have "please try again in few minutes"

awaseem: Done!

Issue #40113, opened 2025-11-04
Self hosted: "realtime" pod crashing


agardnerIT:
# Bug report

I'm trying to follow [the quickstart](https://supabase.com/docs/guides/self-hosting/docker) and, after this comment, things look more healthy. But now the UI isn't appearing on `8000`.

Digging deeper I realised that of the 13 services, only 10 are starting:

`imgproxy`, `auth`, `rest`, `storage`, `meta`, `functions`, `analytics`, `db`, `vector` and `supavisor`.

I see `realtime` trying to start and crashing repeatedly with this error:

```
Setting RLIMIT_NOFILE to 10000
+ ulimit -n
+ '[' '!' -z 10000 ']'
+ echo 'Setting RLIMIT_NOFILE to 10000'
+ ulimit -Sn 10000
+ export ERL_CRASH_DUMP=/tmp/erl_crash.dump
+ ERL_CRASH_DUMP=/tmp/erl_crash.dump
+ '[' false = true ']'
+ [[ -n '' ]]
+ echo 'Running migrations'
+ sudo -E -u nobody /app/bin/migrate
Running migrations
ERROR! Config provider Config.Reader failed with:
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not an already existing atom

    :erlang.binary_to_existing_atom("True")
    /app/releases/2.57.2/runtime.exs:62: (file)
    (elixir 1.18.4) src/elixir.erl:386: :elixir.eval_external_handler/3
    (stdlib 6.2.1) erl_eval.erl:919: :erl_eval.do_apply/7
    (stdlib 6.2.1) erl_eval.erl:663: :erl_eval.expr/6
    (stdlib 6.2.1) erl_eval.erl:271: :erl_eval.exprs/6
    (elixir 1.18.4) src/elixir.erl:364: :elixir.eval_forms/4
    (elixir 1.18.4) lib/module/parallel_checker.ex:120: Module.ParallelChecker.verify/1

[...]

PR #40994, opened 2025-12-02
Update SEED_SELF_HOST and RUN_JANITOR to string values

### LINKED ISSUE #40993: Realtime not seeding with podman/docker-compose
# Bug report

## Describe the bug

Running `podman-compose` on docker-compose.yml causes the realtime-dev.supabase-realtime container to fail healthcheck due to a 404 since the initial tenant was never seeded upon initialization. 

However, using `podman run` for each individual container in order of dependencies (vector, db, analytics, supavisor, realtime) does seed. 

Differences in logs show that using `run` the SEED_SELF_HOST returns 'true' whereas in the `compose` log, it returns 'True' which breaks the conditional. 

## To Reproduce

execute docker/podman-compose

## Expected behavior

realtime seeds the initial tenant

## Additional Details

After hours spent ensuring it wasn't a podman/docker discrepancy (although, I'm not sure if docker handles case-sensitivity differently) or an issue with another container, I ran the log outputs through Claude for quick sanity check. Lo and behold, it worked, and I tested with the quotes and all the containers ended up healthy.

From Claude: 

> Looking at your logs and configuration, I can identify the key difference:
> 
> ### Root Cause
> 
> The issue is in the `realtime` service configuration in `anythingllm.yml`. The seeding process is controlled by the `SEED_SELF_HOST` environment variable, but more importantly, it only runs when a specific condition is met in the container's entrypoint script.
> 
> ### The Difference
> 
> In your **`test_realtime.sh`** (manual run):
> ```bash
> -e SEED_SELF_HOST=true
> ```
> 
> In your **`anythingllm.yml`** (compose):
> ```yaml
> SEED_SELF_HOST: true
> ```
> 
> Both set the variable correctly. However, examining the logs reveals the actual issue:
> 
> #### Manual Run Log (test_realtime.sh):
> ```
> + '[' true = true ']'
> + echo 'Seeding selfhosted Realtime'
> + sudo -E -u nobody /app/bin/realtime eval 'Realtime.Release.seeds(Realtime.Repo)'
> Seeding selfhosted Realtime
> ```
> 
> #### Compose Run Log:
> ```
> + '[' True = true ']'    # Note: capital 'T'
> + echo 'Starting Realtime'
> Starting Realtime
> ```
> 
> ### The Problem
> 
> The condition check in the realtime container's entrypoint script uses **case-sensitive comparison**:
> ```bash
> if [ "$SEED_SELF_HOST" = "true" ]; then
> ```
> 
> In your compose file, something is converting the boolean `true` to `True` (capital T). This is likely because YAML boolean values are being processed differently than string values.
> 
> ### Solution
> 
> In your `anythingllm.yml`, change the `SEED_SELF_HOST` value from a boolean to an explicit string:
> 
> ```yaml
> realtime:
>   container_name: ${POOLER_TENANT_ID}.supabase-realtime
>   image: supabase/realtime:v2.65.3
>   # ... other config ...
>   environment:
>     # ... other variables ...
>     SEED_SELF_HOST: "true"  # Quote it to force string type
>     RUN_JANITOR: "true"     # Same for this one
> ```
> 
> By explicitly quoting the value as `"true"`, you ensure it's treated as a string literal rather than a YAML boolean, preventing the capitalization that's causing the condition to fail.

## System information

- OS: Fedora 42
- Version of main as of 11-30-2025

imdaveho:
Changed SEED_SELF_HOST and RUN_JANITOR values to strings. Prevents case-sensitive type issues referenced in https://github.com/supabase/supabase/issues/40993

YES

Bug/hot fix,

https://github.com/supabase/supabase/issues/40993

## What is the new behavior?

Seeds the database

## Additional context

See https://github.com/supabase/supabase/issues/40993


aantti:
Similar issue - #40113

aantti (approved):

saltcod (approved):

imdaveho:
@aantti -- it just occurred to me that the use of strings in compose for boolean values might not be the best choice here. The better approach might be to adjust the initialization script that for SEED_SELF_HOST for boolean type not a string type. This way things like 1, true, and "true" all might work. But again, this works for now...

PR #41004, opened 2025-12-03
Update how-to-change-max-database-connections.mdx

sudayavarman:
* Updating the troubleshooting guide with a small warning statement on changing max_connections leading to database restart.


sudayavarman (commented):
* Fixing the Prettier run failures due to filename not adhering to prettier rules

PR #41010, opened 2025-12-03
fix(studio): delegate check for expired token to gotrue

charislam:
Gotrue already does the work of using the cached token if possible/refreshing if about to expire (and uses a different expiry period from what we have -- 90 seconds vs. 30 seconds). We can delegate this work to Gotrue to avoid the expiry check mismatch

Related to FE-2179


alaister (approved):
LGTM!

- Verified on staging

PR #41013, opened 2025-12-03
Joshen/fe 2194 disk management form resets storage type on review

joshenlim:
## Context

Realised that the disk configuration page was missing a couple of disabled states

- If organization has spend cap enabled, disk size, storage type, throughput and iops input fields were still editable
- The check for large compute to adjust storage type, throughput and iops was missing too
  image

## To test

- [ ] If organization has spend cap enabled, disk size, storage type, throughput and iops input fields should now be disabled
- [ ] You should see the `Adjusting disk configuration requires LARGE Compute size or above` admonition under "Advanced disk settings" if project is not on a large compute. disk size, storage type, throughput and iops input fields should be disabled too


alaister (approved):
LGTM!

- Tested on preview

PR #41016, opened 2025-12-03
fix(docs): remove unnecessary network restriction section

avallete:
Branching executors are now part of the supabase systems allowed lists


sweatybridge (approved):

sweatybridge:
thanks for catching it

PR #41018, opened 2025-12-03
refactor: infra queries to use `attributes`

raulb:

YES

This PR refactors the infrastructure monitoring query code reducing duplication and unifying the API request to always be `attributes`.

There's duplicated code and in some situations we request for `infra-monitoring?attributes` and in some others `infra-monitoring?attribute` which makes some code a bit redundant. 

## What is the new behavior?

> [!NOTE]  
> No functional changes.

**Code changes**

- Removed the separate useInfraMonitoringQuery hook and getInfraMonitoring function that handled a single monitoring query
- Consolidated all infrastructure monitoring queries into a unified `useInfraMonitoringAttributesQuery` hook that handles multi-attribute requests
- Moved interval selection logic from the query layer to the consumer `InfrastructureActivity.tsx`, where it can be computed dynamically based on user-selected date ranges
- Simplified query types by removing intermediate InfraMonitoringData and InfraMonitoringVariables types
- Interval is now computed in the component (defaults to 1d, switches to 1h for date ranges ≤48 hours) rather than hardcoded in the query layer
- All queries now use the unified multi-attribute endpoint with explicit parameter passing

## Additional context

Working on https://github.com/supabase/platform/pull/28021, I noticed we were still making these single attribute requests.

This is take II of https://github.com/supabase/supabase/pull/40473.

### DIFF
--- a/apps/studio/data/analytics/infra-monitoring-queries.ts
+++ b/apps/studio/data/analytics/infra-monitoring-queries.ts
@@ -9,11 +9,9 @@ import {
   InfraMonitoringMultiData,
   InfraMonitoringError,
   useInfraMonitoringAttributesQuery,
-  useInfraMonitoringQuery,
 } from './infra-monitoring-query'
 
 const DEFAULT_DATE_FORMAT = 'HH:mm DD MMM'
-const DEFAULT_ATTRIBUTE: InfraMonitoringAttribute = 'max_cpu_usage'
 
 type InfraQueryResult = Pick<
   UseQueryResult<AnalyticsData, InfraMonitoringError>,
@@ -32,22 +30,8 @@ export function useInfraMonitoringQueries(
 ) {
   const shouldFetch = data === undefined && isVisible
   const hasAttributes = attributes.length > 0
-  const isSingleAttribute = attributes.length === 1
-  const primaryAttribute = attributes[0] ?? DEFAULT_ATTRIBUTE
 
-  const singleAttributeQuery = useInfraMonitoringQuery(
-    {
-      projectRef: ref as string,
-      attribute: primaryAttribute,
-      startDate,
-      endDate,
-      interval,
-      databaseIdentifier,
-    },
-    { enabled: shouldFetch && isSingleAttribute && hasAttributes }
-  )
-
-  const multiQuery = useInfraMonitoringAttributesQuery(
+  const query = useInfraMonitoringAttributesQuery(
     {
       projectRef: ref as string,
       attributes,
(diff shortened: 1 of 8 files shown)


apps/studio/data/analytics/infra-monitoring-queries.ts
```diff
@@ -32,22 +30,8 @@ export function useInfraMonitoringQueries(
 ) {
   const shouldFetch = data === undefined && isVisible
   const hasAttributes = attributes.length > 0
-  const isSingleAttribute = attributes.length === 1
-  const primaryAttribute = attributes[0] ?? DEFAULT_ATTRIBUTE
 
-  const singleAttributeQuery = useInfraMonitoringQuery(
```
charislam, line 38: There's a breaking change here -- `singleAttributeQuery` used to be returned without mapping the response, so the assumptions of that mapping function are incorrect. This breaks charts like IO usage that use only a single attribute

charislam (approved):
Had to push some fixes because the single attribute response still comes back as a different shape from the multiple attribute response, otherwise LGTM

charislam:
Actually just realized the the single branch was intended to be removed haha: https://github.com/supabase/platform/blob/7d7580c81a74ed8cca6ed2dd2c03269f1d6508c3/api/apps/mgmt-api/src/routes/platform/projects/ref/infra-monitoring.controller.ts#L178

But we'll still need to handle the transition:
- FE stops sending attribute
- FE needs to handle both single and multiple response shapes in intermediate phase 🤔 
- API stops sending single response shape

raulb:
> Actually just realized the the single branch was intended to be removed haha: [supabase/platform@`7d7580c`/api/apps/mgmt-api/src/routes/platform/projects/ref/infra-monitoring.controller.ts#L178](https://github.com/supabase/platform/blob/7d7580c81a74ed8cca6ed2dd2c03269f1d6508c3/api/apps/mgmt-api/src/routes/platform/projects/ref/infra-monitoring.controller.ts#L178)
> 
> But we'll still need to handle the transition:
> 
> * FE stops sending attribute
> * FE needs to handle both single and multiple response shapes in intermediate phase 🤔
> * API stops sending single response shape

@charislam Thanks for looking! This is what I'm doing with this PR right? The API currently accepts both single `attribute` and `attributes`. Once this PR ships, I could remove the API endpoint (I'm planning to leave it running in production for a couple of weeks to make sure there's no real usage. This is how I noticed I still needed to do this clean up 😂).

Let me know if I'm missing something here, or if you think I still need to do something here.

raulb:
> Had to push some fixes because the single attribute response still comes back as a different shape from the multiple attribute response, otherwise LGTM

Actually, did I miss those changes by force-pushing? Hopefully not!

raulb:
@charislam just looked into this more closely. I think you're right. I think the API still need to be modified before we ship this one. Doing that now

apps/studio/data/analytics/infra-monitoring-queries.ts
raulb: Good catch. Made some changes here, let me know what you think https://github.com/supabase/supabase/pull/41018/commits/ed828a9c5250267e6636f6edc13a01a2e73b99f4.

charislam (approved):

PR #41019, opened 2025-12-03
[FE-2176] fix(studio): add missing sign up enabled flag for nimbus

alaister:
To test:
- make sure sign up is still button is still available on staging


joshenlim (approved):
Verified on staging preview org invite that create account button is visible

PR #41024, opened 2025-12-03
blog: add pagination

stylessh:

YES


apps/www/app/api-v2/blog-posts/route.ts
```diff
@@ -0,0 +1,97 @@
+import { NextRequest, NextResponse } from 'next/server'
+import { getSortedPosts } from 'lib/posts'
+import { getAllCMSPosts } from 'lib/get-cms-posts'
+
+export const revalidate = 30
+
+export async function OPTIONS() {
+  return new Response(null, {
+    status: 200,
+    headers: corsHeaders,
```
kemaldotearth: This seems to be throwing a type error on the linters, looks like it's missing a definition.

CoolAssPuppy (approved):

kemaldotearth (changes requested):
Other than the `corsHeaders` type error and the conflict in the RSS file, seems to be working much better than before. 

I'm wondering if the threshold is triggering the load more a bit too early, but no strong opinion there. Much less req in terms of fetching more items.

stylessh:
yes i can adjust the threshold

apps/www/app/api-v2/blog-posts/route.ts
stylessh: fixed, this is not used anymore

kemaldotearth (approved):
Approving so as not to block this, but there is still a type error for the `corsHeader` in your API route. Doesn't seem to have any actual effect on staging at least 🤨

PR #41026, opened 2025-12-03
feat(design-system): charts defrag

kemaldotearth:

YES

This is part of our design defrag topic. I am pruning through our Design System charts, auditing what we do or don't use and looking at making our Report charts (well, any charts) composable and easy to use. 

**List of changes:**
- Remove all the unused chart examples from our design system.
- Standardise all Chart patterns from Studio into composable components.
- Merge Metric Card components into Chart.
- Better documentation for the Design System.
- Better examples of representative Charts inside Studio.

I am holding off on replacing charts inside Studio to begin with. This takes care of documentation and suggested best practices/compositions of charts going forward.


apps/design-system/registry/default/block/chart-composed-states.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <ChartTitle tooltip="This is a tooltip">Response Errors</ChartTitle>
+            <ChartActions actions={actions} />
+          </ChartHeader>
+          <ChartContent
+            loadingState={<ChartLoadingState />}
+            disabledState={
+              <ChartDisabledState
+                icon={<Badge variant="success">Pro</Badge>}
+                label="API Processing Time"
+                description="This chart is available from pro plan and above"
```
dnywh:
```suggestion
                description="This chart is available on the Pro plan and above"
```

apps/design-system/registry/default/block/chart-composed-states.tsx
```diff
@@ -0,0 +1,92 @@
...
+        alert('Ta da! You clicked me! 🎉')
+      },
+      icon: <ExternalLink size={12} />,
+    },
+  ]
+
+  const disabledActions = [
+    {
+      label: 'Upgrade to Pro',
+      href: '/pro',
```
dnywh: Not that big of a deal, but worth mentioning: having this `href` means that tapping the button redirects the user from the page. We could omit the `href` entirely here (or comment it out so it’s still indicative to viewers + LLMs?).

I swear there was a way to set a `href` value that was ignored by browsers. But now I can’t find it. It was something like:

```html
<a href="#.">Link</a>
```

apps/design-system/__registry__/default/block/chart-composed-table.tsx
dnywh, line 1: What’s this one for?

dnywh: Assuming it’s placeholder for a future PR.

apps/design-system/eslint.config.cjs
```diff
@@ -1,4 +1,12 @@
 const { defineConfig } = require('eslint/config')
 const supabaseConfig = require('eslint-config-supabase/next')
 
-module.exports = defineConfig([supabaseConfig])
+module.exports = defineConfig([
+  supabaseConfig,
+  {
+    files: ['registry/**/*.tsx', '__registry__/**/*.tsx'],
+    rules: {
+      'no-restricted-exports': 'off',
```
dnywh, line 9: My understanding of this:
- We prefer named exports in `studio` etc
- But this is an exception, since we have a registry build process
- This build process relies on `React.lazy()` which in turn relies on default exports

Is that the case? Just curious about how this works.

dnywh (approved):
Reviewed design system changes and minimally tested locally. Left a few non-blocking comments/questions. Looks great!

apps/design-system/eslint.config.cjs
kemaldotearth: Yeah that's the issue, we prefer the named exports, but `React.lazy()` doesn't seem to play nice with them here. Feels like a separate issue. It started giving me a bunch of linter errors for default exports across all our registry, so put this in there for the time being. 

Maybe we can isolate to fix this post?

apps/design-system/__registry__/default/block/chart-composed-table.tsx
kemaldotearth: Is it not rendering for you? There's a component with a table footer.

Screenshot 2025-12-09 at 13 22 53

jordienr (approved):

PR #41031, opened 2025-12-03
feat(log drains): add s3 form FE-2101

jordienr:
- adds s3 log drain form
- fixes issue tabbing over inputs

to test you can create an s3 bucket in a aws account, create some credentials and try it, however I've tested it already and the logs reach S3 just fine. 

## Summary by CodeRabbit

* **New Features**
  * Added Amazon S3 as a new log drain destination with configuration for bucket, region, access credentials, and batch timeout settings.

* **Documentation**
  * Updated guide documentation to include the new S3 destination alongside existing options and standardized Datadog terminology.

* **Improvements**
  * Simplified error message rendering in form layouts for improved performance.


packages/common/telemetry-constants.ts
```diff
@@ -2281,7 +2281,7 @@ export interface LogDrainConfirmButtonSubmittedEvent {
     /**
      * Type of the destination confirmed
      */
-    destination: 'webhook' | 'datadog' | 'loki' | 'sentry'
+    destination: string
```
awaseem: Should we not keep these as hard typed? Moving to string seems backwards

awaseem (approved):
LGTM One minor comment

packages/common/telemetry-constants.ts
jordienr: fixed and added missing types 🚢

pamelachia (approved):

PR #41034, opened 2025-12-03
chore(studio): add dev dependency on @typescript-eslint/eslint-plugin

charislam:
The plugin was added but not in dependencies, which caused problems with IDE language server resolution.


awaseem (approved):
LGTM

PR #41041, opened 2025-12-04
fix(studio): send new signups to org creation directly

dnywh:
## What kind of change does this PR introduce?

Bug fix. Resolves DEPR-318.

## What is the current behavior?

New users who confirm their email land on `/sign-in`, then `/organizations`, then get bounced to `/new` via a `useEffect`. Cancelling org creation with zero orgs sends them back to `/organizations`, which immediately redirects into `/new` again.

## What is the new behavior?

- Signup email verification redirects to `/new` directly
- `/organizations` with zero orgs shows the existing empty state instead of force-redirecting

### One-time setup

Assuming you don’t already have a staging account with **zero** orgs:

1. On **supabase.green**, sign up with a fresh email and confirm it
2. Stop at org creation. Do **not** create an org

### On this PR

Using the [studio-staging preview](https://studio-staging-git-dnywh-fixremove-org-redirect-supabase.vercel.app/) from Vercel checks:

4. Sign in on the preview with that account
5. Open `/dashboard/organizations`. Expect the **Create an organization** empty state, with no redirect to `/new`
6. Open `/dashboard/new`, click **Cancel**. Expect to land on `/organizations` and stay there

### Compare on supabase.green

Optional. Just to show what happens currently on `master`:

7. Repeat steps 3–5 on **supabase.green**. `/organizations` should bounce to `/new`, and **Cancel** should send you back into org creation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signup redirects by preserving valid destinations and relevant query parameters.
  * Added safer fallback behavior for missing, invalid, or unsupported destinations.
  * Improved handling of signup redirects provided in multiple formats.
  * Prevented automatic redirection from the organizations page when no organizations exist.

* **Style**
  * Updated the organizations page title capitalization for consistency.

PR #41044, opened 2025-12-04
chore(studio): better empty states for new users

dnywh:

UI update that partially resolves DEPR-228.

A user who is creating their first organisation, project, and content gets some inconsistent empty states. Consistent empty states would make the onboarding experience slightly smoother and overall experience more polished.

## What is the new behavior?

Improved empty states across:

- Organizations
- Projects

Minor design docs content reshuffle and icon addition.

| Before | After |
| --- | --- |
| Supabase | Supabase |
| Supabase | Supabase |
| Supabase | Supabase |

## Additional context

I’ve improved and implemented `NoOrganizationsState` in preparation of removing the redirect away from it (see https://github.com/supabase/supabase/pull/41041 for proposed fix). It’s still shown when a user goes to `/organizations` _if_ they pass an invalid project name. For example, going to:

```
localhost:8082/org/made-up-org-name/billing
```

...when the user **has no organizations** will take them to `/organizations`. This PR improves the presentation there.

I’ve updated the admonition in the project creation flow to suit:

```
localhost:8082/new/made-up-org-name
```

This one needs to be tested in a project with 1+ other organizations.


packages/icons/src/raw-icons/box-plus.svg
dnywh, line 1: New custom icon for projects, based on Lucide’s [box](https://lucide.dev/icons/box), which we already use to denote projects.

apps/design-system/content/docs/ui-patterns/empty-states.mdx
dnywh, line 1: Just reordering content to flow better.

apps/design-system/content/docs/icons.mdx
```diff
@@ -55,6 +55,7 @@ Follow these steps to add a new custom icon to the Supabase icon library.
 
    - Exported at 24x24px with `viewBox="0 0 24 24"`
    - Uses `stroke="currentColor"` for strokes (no hardcoded colors)
+   - Uses `stroke-width="1.5"` unless there is a good reason to deviate
```
dnywh, line 58: Lucide tends to have a `stroke-width` of `2` when exporting an SVG from their site, whereas our versions tend to use `1.5`. Better to align, I think.

dnywh: I believe this is overridden by the prop so not a huge deal.

apps/studio/components/interfaces/Home/ProjectList/EmptyStates.tsx
dnywh, line 1: Reusing our fancy component instead of making bespoke ones.

apps/studio/components/interfaces/Organization/OrgNotFound.tsx
dnywh, line 1: Copy improvements and switch to using latest alert type component (Admonition).

apps/studio/components/interfaces/SignIn/LastSignInWrapper.tsx
dnywh, line 1: Unrelated mini fix from earlier Badge work, sorry!

| Before | After |
| --- | --- |
| Image | Image |

jordienr (approved):

PR #41057, opened 2025-12-04
perf: preconnect to API domain

kevcodez:
This change adds a <link rel="preconnect"> for our API domain. Because the application makes several client-side requests to https://api.supabase.com shortly after page load, the browser currently has to perform DNS lookup, TCP handshake, and TLS negotiation only when the first API call happens.

By preconnecting, we let the browser complete these expensive operations ahead of time, in parallel with page rendering. This reduces the latency of the first API request and improves overall perceived performance, especially for users on slower networks or with high latency to the API host.


alaister (approved):
LGTM!

- Tested on preview

PR #41067, opened 2025-12-04
Revert "Revert "perf: preconnect to API domain""

alaister:
Reverts supabase/supabase#41063

Also switches to `crossOrigin="use-credentials"` since we're including credentials in our API requests.

To test:
- Make sure nothing breaks
- You can check in the Safari JS console as it logs when it pre-connects successfully


joshenlim (approved):
Minimally checked staging preview

PR #41077, opened 2025-12-04
fix: Refactor the foreign tables query to use FE pg-meta

ivasilov:
This PR fixes an issue where foreign tables would be listed in all schemas.

### DIFF
--- a/apps/studio/data/foreign-tables/foreign-tables-query.ts
+++ b/apps/studio/data/foreign-tables/foreign-tables-query.ts
@@ -1,8 +1,8 @@
+import pgMeta from '@supabase/pg-meta'
+import { PostgresView } from '@supabase/postgres-meta'
 import { useQuery } from '@tanstack/react-query'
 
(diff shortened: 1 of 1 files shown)


apps/studio/data/foreign-tables/foreign-tables-query.ts
```diff
@@ -16,29 +16,17 @@ export async function getForeignTables(
...
-      path: { ref: projectRef },
-      query: {
-        included_schemas: schema || '',
-        include_columns: true,
-      } as any,
+  const { result } = await executeSql(
+    {
+      projectRef,
+      connectionString,
+      sql: pgMeta.foreignTables.list(schema ? { includedSchemas: [schema] } : undefined).sql,
```
awaseem, line 23: If the value is undefined, `.sql` reference will fail FYI

awaseem: Ignore this, got confused with the assignment

awaseem (approved):
LGTM

PR #41080, opened 2025-12-04
fix(pg-meta): pair composite foreign key columns correctly

### LINKED ISSUE #41068: Foreign key relation displays wrong table when composite foreign keys exist (intermittent)
# Bug report

- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/orgs/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

This is the exact same issue as #23485, which has been waiting for feedback since 2024.

The Table Editor displays incorrect foreign key relationships in the UI:
- Tooltip shows the wrong target table (displays the source table instead of the actual FK target)
- Clicking to select a related record opens the wrong table picker

I've verified via SQL that the database schema is correct - this is purely a UI bug.

**The bug is intermittent and severely impacts multi-tenancy implementations:** Some tables in my database display FKs correctly, others don't. The issue occurs when:
- Tables have composite primary keys or composite unique constraints (e.g., `id` + `tenant_id`)
- Tables have simple primary keys but unique constraints combining any column with `tenant_id`
- Other tables reference these with simple FKs

This makes multi-tenancy patterns (where most tables need `tenant_id` in unique constraints) essentially broken in the Table Editor UI.

## To Reproduce

Same as described in #23485. The bug occurs when:
1. You have tables with composite foreign keys
2. Other tables have simple FKs that reference columns involved in those composite FKs
3. The UI incorrectly displays the FK target table and opens wrong pickers

## Expected behavior

- UI should display the correct target table name from the actual database schema
- Clicking to select should open the correct table picker
- Behavior should be consistent across all tables

## Screenshots

[I can provide screenshots if needed - the bug looks identical to #23485]

## System information

- OS: Windows/macOS/Linux
- Browser: Opera
- Supabase: Cloud hosted

## Additional context

**Reproducibility:** Non-deterministic. I've tried dropping and recreating the foreign keys multiple times:
- On some tables, it eventually started working after several attempts
- On other tables, the bug persists regardless of how many times I recreate the constraint

The bug makes the Table Editor unusable for data entry on affected tables. Current workaround is using SQL Editor directly.

This needs attention as it's been affecting users since at least 2024 (#23485).

7ttp:
## TL;DR

Correctly pairs composite foreign key columns when loading table metadata.

## What's hurting?

The tables introspection query matched every source column in a composite foreign key with every target column. For `(user_id, tenant_id) → (id, tenant_id)`, it returned four relationships instead of the correct two, causing incorrect relationship metadata in the Table Editor...

## Now fixed

Source and target columns are paired by ordinal position using a lateral multi-array `unnest`.
 Regression coverage now verifies adversarial column ordering, and the existing performance guard exercises thousands of composite foreign keys...

PS:  local stress test found no performance regression or unexpected sequential scans.

## Ref

- Closes #41068

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected foreign-key relationship detection for composite keys, ensuring source and target columns are paired accurately.
  * Improved catalog relationship queries to remain within performance limits for composite-key tables.


7ttp:
@awaseem please review, when available!! :D

packages/pg-meta/src/sql/tables.ts
```diff
@@ -60,16 +60,17 @@ FROM
...
       join pg_class csa on sa.attrelid = csa.oid
       join pg_namespace nsa on csa.relnamespace = nsa.oid
-    ) on sa.attrelid = c.conrelid and sa.attnum = any (c.conkey)
+    ) on sa.attrelid = c.conrelid and sa.attnum = cols.source_attnum
     join (
       pg_attribute ta
       join pg_class cta on ta.attrelid = cta.oid
       join pg_namespace nta on cta.relnamespace = nta.oid
-    ) on ta.attrelid = c.confrelid and ta.attnum = any (c.confkey)
+    ) on ta.attrelid = c.confrelid and ta.attnum = cols.target_attnum
```
avallete, line 116: Can we stress test this under load (lot of tables/relations) to make sure it doesn't degrade performances of the query too much ?

Similar to:

https://github.com/supabase/supabase/pull/48148/changes#diff-b2e0a37dfc4176d536d8d83275b7ac22f1302bd880e0e49444b54a7f5f66d1b3

https://github.com/supabase/supabase/pull/48148/changes#diff-dfdf033c20b74049e513375093802a54e154730a982e232299b3917f53ebe3b7

Just to make sure we don't blow things up with additional seq-scans, etc...

packages/pg-meta/src/sql/tables.ts
```diff
@@ -103,16 +103,17 @@ FROM
       ta.attname as target_column_name
     from
       pg_constraint c
+    cross join lateral unnest(c.conkey, c.confkey) with ordinality as cols(source_attnum, target_attnum, ord)
```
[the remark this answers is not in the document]
7ttp: Valid and resolved in commit f871ac5dc8

The regression now asserts the two ordinal pairs with adversarial column order, and the full pg-meta suite passes...

packages/pg-meta/src/sql/tables.ts
7ttp: added in latest commit

avallete (approved):

PR #41090, opened 2025-12-04
feat(studio): use bucket pagination for api docs

charislam:
API docs V2 updated to use pagination for fetchign storage buckets.

Also some refactoring to split up large components for docs V2 interfaces.

You'll need to turn on the API Documentation feature preview to test this (storage buckets are not shown on the legacy API Docs pages).

Towards FE-2118


awaseem (approved):
Good clean up, thanks!

PR #41107, opened 2025-12-05
chore: entitlements feature catalog

ignaciodob:
### Changes
- Add type-safe `Feature` constant for entitlement feature keys, derived from backend API types
- Update `useCheckEntitlements` hook to accept `FeatureKey` type instead of arbitrary strings

### DIFF
--- a/apps/studio/data/entitlements/entitlements.constants.ts
+++ b/apps/studio/data/entitlements/entitlements.constants.ts
@@ -3,9 +3,9 @@ import type { components } from 'api-types'
 export type FeatureKey = components['schemas']['ListEntitlementsResponse']['features'][number]
 
 export const Feature = {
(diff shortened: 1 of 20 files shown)


ignaciodob (commented):
I've checked every file - it looks okay. A second look wouldn't hurt 🙏

apps/studio/data/entitlements/entitlements.constants.ts
```diff
@@ -0,0 +1,21 @@
+import type { components } from 'api-types'
+
+export type FeatureKey =
+  components['schemas']['ListEntitlementsResponse']['entitlements'][number]['feature']['key']
+
+export const Feature = {
```
kevcodez: There is no need to redefine all keys again, you can simply use the type from the generated api types and type out the string in a type safe way with auto completion

kevcodez (changes requested):

kevcodez (approved):

PR #41114, opened 2025-12-05
chore(studio): bump lint rule to error

charislam:
We've removed all violation of @tanstack/query/no-deprecated-options, so we can bump that lint rule up to an error.


awaseem (approved):
LGTM

ivasilov (approved):

PR #41116, opened 2025-12-05
fix(studio): scope indexes tab to just select queries

kemaldotearth:

YES

As we only scope Index Advisor to `SELECT` queries, the Indexes tab was hanging on some fluff queries from `pg_stat_statements`.  This hides the tab on those queries until we add a more robust check.


awaseem (approved):
LGTM

PR #41121, opened 2025-12-05
fix(studio): add PG_META_CRYPTO_KEY to .env example

pinkbit256:

YES

If the PG_META_CRYPTO_KEY variable is not set in the .env file, the system will fall back to the default key (SAMPLE_KEY) using:
```ts
export const ENCRYPTION_KEY = process.env.PG_META_CRYPTO_KEY || 'SAMPLE_KEY'
```
This default value can lead to connection issues for the meta service.

It will use SAMPLE_KEY by default. 

## What is the new behavior?
Read from .env file.

## Additional context
None

### DIFF
--- a/apps/studio/.env
+++ b/apps/studio/.env
@@ -1,5 +1,6 @@
 STUDIO_PG_META_URL=http://localhost:8000/pg
 POSTGRES_PASSWORD=your-super-secret-and-long-postgres-password
+PG_META_CRYPTO_KEY=your-encryption-key-32-chars-min
 
 DEFAULT_ORGANIZATION_NAME=Default Organization
 DEFAULT_PROJECT_NAME=Default Project


aantti:
Thanks, @hi-jennie - can you possibly describe what kind of a setup you're addressing with this? Normally, this key has to be set to something else rather than a default placeholder password. Are you self-hosting? Developing a local version of Studio? Something else?

pinkbit256:
> Thanks, @hi-jennie - can you possibly describe what kind of a setup you're addressing with this? Normally, this key has to be set to something else rather than a default placeholder password. Are you self-hosting? Developing a local version of Studio? Something else?

yes, I’m self-hosting.
In the Docker setup for Studio, this key isn’t present in the provided `.env`, so it falls back to the default  value.
The issue I’m seeing is that when Meta uses that default value to talk to the Studio service, Studio doesn’t accept it (auth/verification fails), so the integration doesn’t work.
Once I explicitly set this key in `.env` (overriding the default) and used the same value on the Meta side, Meta → Studio started working and Studio could be set up successfully.
So it seems this variable needs to be explicitly configured for self-hosted deployments, rather than relying on the default value.

aantti:
Correct, but also - it's been added to `docker-compose.yml` when it was introduced with the MCP server related changes earlier in October:

https://github.com/supabase/supabase/blob/bfe3ea06df7af0f145388ae8b885deaea43f7fce/docker/docker-compose.yml#L39

https://github.com/supabase/supabase/blob/bfe3ea06df7af0f145388ae8b885deaea43f7fce/docker/.env.example#L14

I've also included it to the recent changes to self-hosted guide:

https://supabase.com/docs/guides/self-hosting/docker#configure-other-keys-and-important-urls

I'm not too sure if the default `PG_META_CRYPTO_KEY` should also be added to the Studio dev environment, though. Will leave it for the frontend team to assess. Your proposed changes, @hi-jennie, seem to be about a local development / changes to Studio more so than when self-hosting with Docker Compose and pre-built images. If I'm missing something here, please let me know.

Cc @awaseem @ivasilov 

Upd: I think, you might be referring to the CLI/local-dev configuration, though, @hi-jennie ? CLI doesn't seem to configure `PG_META_CRYPTO_KEY` explicitly, but it's the same `SAMPLE_KEY` in both [postgres-meta code](https://github.com/supabase/postgres-meta/blob/185eac0e5dc74d21eec4675068ecaedaeecf7c77/src/server/constants.ts#L9) and Studio:

https://github.com/supabase/supabase/blob/bfe3ea06df7af0f145388ae8b885deaea43f7fce/apps/studio/lib/api/self-hosted/constants.ts#L3

This doesn't seem very transparent, though. Cc @sweatybridge @avallete

kekekuli:
My suggestion is to add the same placeholder to apps/studio/.env, emit a prominent warning when the key falls back to SAMPLE_KEY, and specifically point out a possible crypto key mismatch when a connection fails, as #50628 described.
>But why does Docker Compose use a default placeholder password as the crypto key 😂? And SAMPLE_KEY is also a placeholder. I'm a bit confused here.

PR #41149, opened 2025-12-08
fix: use deno2 entrypoint when it's available

laktek:

YES

This uses entrypoint in eszip bundle if it's availble (only for deno2 bundles)


joshenlim (approved):
Minimally tested staging preview creating + updating edge functions

PR #41150, opened 2025-12-08
Small advisor refinements

SaxonF:
image

Some small UI refinements to advisor section on new home

### DIFF
--- a/apps/studio/components/interfaces/HomeNew/AdvisorSection.tsx
+++ b/apps/studio/components/interfaces/HomeNew/AdvisorSection.tsx
@@ -1,6 +1,5 @@
 import { BarChart, Shield } from 'lucide-react'
 import { useCallback, useMemo } from 'react'
-import ReactMarkdown from 'react-markdown'
 
 import { useParams } from 'common'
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/HomeNew/AdvisorSection.tsx
```diff
@@ -138,8 +139,10 @@ export const AdvisorSection = ({ showEmptyState = false }: { showEmptyState?: bo
                     />
                   </CardHeader>
                   <CardContent className="p-6 pt-16 flex flex-col justify-end flex-1 overflow-auto">
-                    {lint.detail ? lint.detail.substring(0, 100) : lint.title}
-                    {lint.detail && lint.detail.length > 100 && '...'}
+                    <h3 className="mb-1">{lint.title}</h3>
+                    <ReactMarkdown className="leading-6 text-sm text-foreground-light">
```
joshenlim: reckon we use our Markdown component which is also using ReactMarkdown inside, just with custom styling - e.g for code elements

joshenlim (approved):
Verified changes! Just a minor feedback but wondering if we should color the "shield" icon in the cards to match the warning level, for consistency with the advisor panel

image

PR #41152, opened 2025-12-08
Add columns to database indexes UI

joshenlim:
## Context

For the database indexes page, include a new column "Columns" to show which columns of the table the index applies to

Am also shifting dashboard specific queries out of the `pg-meta` package into dashboard within `data/sql/queries` and assigning postgres team as the codeowners there (based on request from postgres team)

Main changes for this PR are from the files
- `get-indexes.ts`
- `indexes-query.ts`
- `Indexes.ts`

## Before
image

## After
image

## Changes involved
- Update + shift SQL for retrieving database indexes to pg-meta
- Hide "Schema" column (consistent with database/tables page + there's already a schema selector)
- Add "Columns" column

## To test

- [ ] Verify that you can view the columns which each database index applies to


saltcod (approved):
looks great!

apps/studio/data/database-indexes/indexes-query.ts
```diff
@@ -1,31 +1,20 @@
...
-  tablename as "table",
-  indexname as "name",
-  indexdef as "definition"
-FROM pg_indexes
-WHERE schemaname = '${schema}';
-`.trim()
-
-  return sql
-}
-
```
soedirgo, line 22: Actually @avallete and I agreed that we should keep Studio-specific queries in the Studio codebase, and only put general purpose queries in pg-meta ([ref](https://supabase.slack.com/archives/C085L4EFUGH/p1762284188642459))

soedirgo: I'm guessing [this query](https://github.com/supabase/supabase/blob/master/packages/pg-meta/src/pg-meta-indexes.ts) isn't enough hence why we're using a separate query?

joshenlim: ahh okay gotcha on this - I can shift the queries out, but will still have team-postgres as code owners in the new folder then 🙏🙂

>I'm guessing [this query](https://github.com/supabase/supabase/blob/master/packages/pg-meta/src/pg-meta-indexes.ts) isn't enough hence why we're using a separate query?

yup - the additional info that we needed here is to list the columns that the indexes applies to

soedirgo: That's perfect, thanks! For ease of reviewing/codeowner setup, maybe we can move these to `apps/studio/data/sql` and update the codeowners file? Since `apps/studio/data` currently also contains Mgmt API handlers

soedirgo (approved):

PR #41162, opened 2025-12-08
feat(design-system): add Field component and docs

fsansalvadore:
New _Field_ documentation in the design system:
https://design-system-git-feat-ds-field-supabase.vercel.app/design-system/docs/components/field

### DIFF
--- a/apps/design-system/registry/default/example/field-slider.tsx
+++ b/apps/design-system/registry/default/example/field-slider.tsx
@@ -1,0 +1,30 @@
+'use client'
+
+import { useState } from 'react'
+import { Slider } from 'ui'
+import { Field, FieldTitle, FieldDescription } from 'ui/src/components/shadcn/ui/field'
+
+export default function FieldSlider() {
+  const [value, setValue] = useState([200, 800])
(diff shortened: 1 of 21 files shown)


apps/design-system/registry/default/example/field-slider.tsx
dnywh, line 1: Missing knob on right-hand side:

| Shadcn | Supabase |
| --- | --- |
| Image | Image |

dnywh (approved):
Nice one. Approved to unblock. One issue with Slider component, as mentioned above. Would also be nice to have some more clarity around `FormItemLayout` vs `FormLayout`, plus things that _look_ like forms but aren’t really (i.e. [Layout](https://design-system-git-feat-ds-field-supabase.vercel.app/design-system/docs/ui-patterns/layout)).

apps/design-system/registry/default/example/field-slider.tsx
fsansalvadore: I see this is actually an issue with the Slider component, will fix in a separate PR 👍

PR #41163, opened 2025-12-08
ref: Remove unused queries/mutations

kamilogorek:
All those removed queries and mutations were not wired up to anything and were not called in at least last 2 months. Can be safely removed.


awaseem (approved):
LGTM

PR #41165, opened 2025-12-08
feat: add auth db connection strategy linter

hf:
Adds a new informational performance lint rule that checks the Auth server's connection pooling strategy. If it's not percent then it shows up on projects that aren't on free plan and have at least one user.

image


alaister (approved):
LGTM!

PR #41172, opened 2025-12-08
Improve experiment telemetry for generate policies A/B test

seanoliver:
## Problem

The generate policies experiment telemetry wasn't set up in a way that makes PostHog analysis straightforward. We learned from the table quickstart experiment that having concrete exposure and conversion events with consistent `experiment_id` and `variant` properties makes analyzing results way cleaner directly in PostHog.

## Changes

- Updated the exposure event to use `experiment_id` and `variant` props instead of just an `enabled` boolean
- Added a dedicated conversion event that fires when a user in the experiment creates a table
- Both events share the same structure so we can easily build funnels in PostHog

The conversion event is technically redundant with `table_created`, but having a dedicated experiment event is temporary (just for the life of this A/B test, same as the exposure event) and makes the analysis much cleaner since we don't have to filter/join against a more generic event.

## Testing

Verified TypeScript compiles, checked that both track calls match their type definitions. The exposure event fires via the existing hook when the table editor renders for new tables, and the conversion event fires after successful table creation.


seanoliver:
@joshenlim @SaxonF Let me know if this is safe to merge into the feature branch!

awaseem (approved):
LGTM

4L3k51 (approved):

PR #41177, opened 2025-12-09
chore: use simplified support form for platform orgs

kevcodez:


joshenlim (approved):
Verified changes locally!

PR #41179, opened 2025-12-09
chore(studio): improve projects presentation

dnywh:

UI improvements. Resolves DEPR-269.

Viewing projects as a list (not cards) reveals:

- Full-height table despite most people having only a handful of projects and our pagination handling not using this style elsewhere (e.g buckets)
- Double border between `TableHeader` and `TableBody`
- Clipped top edge of the `TableHeader`
- Minor UI rough edges like project status presentation

## What is the new behavior?

Fixes to all of the above including:

- Removal of full-height table container which also fixes the clipping issue
- Removal of duplicate loading state which also fixes the double border
- Polish on table presentation including consistent typography
- Polish on integrations such as Vercel and GitHub
- Polish on project status presentation

| Before | After |
| --- | --- |
| Supabase | Supabase |
|  Supabase | Supabase |
| Supabase | Supabase |

## Summary by CodeRabbit

* **New Features**
  * Empty states now render contextually inside table cells and support customizable labels/descriptions and reset actions.
  * Infinite-scroll "Load more" placeholders added for both card and table views.

* **UI**
  * Project cards and table rows updated to a stacked title layout with refreshed integration badges and tighter spacing.
  * Table headers animate color when no-results are shown.

* **Documentation**
  * Docs expanded to describe NoFilterResults and NoSearchResults patterns.

* **Refactor**
  * Badge component updated to support forwarded refs.


apps/studio/components/interfaces/Home/ProjectList/EmptyStates.tsx
```diff
@@ -39,33 +39,36 @@ export const Header = () => {
 export const NoFilterResults = ({
   filterStatus,
   resetFilterStatus,
+  withinTableCell = false,
   className,
 }: {
   filterStatus: string[]
   resetFilterStatus?: () => void
+  withinTableCell?: boolean
```
dnywh: Matching `NoSearchResults`. I wonder if we should combine these two.

apps/studio/components/ui/NoSearchResults.tsx
```diff
@@ -2,26 +2,29 @@ import { Button, cn } from 'ui'
...
 export interface NoSearchResultsProps {
   searchString: string
+  withinTableCell?: boolean
   onResetFilter?: () => void
   className?: string
 }
 
 export const NoSearchResults = ({
   searchString,
+  withinTableCell = false,
```
dnywh, line 14: Otherwise we are constantly a bunch of classes via `className` whenever this component is used inside. table. This way we can simply render those classes when it is “standalone”.

I think a clearer prop would be the inverse of how I’ve done it: `standalone={true}`. But this is a move towards that without needing to touch all 15 instances.

apps/design-system/content/docs/ui-patterns/empty-states.mdx
```diff
@@ -42,6 +42,11 @@ A [Table](../components/table) instance with zero results should display a singl
 
 <ComponentPreview name="empty-state-zero-items-table" peekCode wide />
 
+Studio contains two pre-built components to handle these cases consistently:
+
+- NoFilterResults
+- NoSearchResults
+
```
dnywh, line 49: I only discovered these two today despite having done a bunch of equivalents manually for a while. Even though these two aren’t shared components, I think they are worth mentioning in the design system so others (and LLMs) know they exist.

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
```diff
@@ -105,6 +95,8 @@ export const ProjectList = ({ organization: organization_, rewriteHref }: Projec
   const noResultsFromStatusFilter =
     filterStatus.length > 0 && isSuccessProjects && orgProjects.length === 0
 
+  const noResults = noResultsFromStatusFilter || noResultsFromSearch
```
dnywh, line 91: Useful for below check on `TableHead` instances

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
```diff
@@ -161,35 +153,40 @@ export const ProjectList = ({ organization: organization_, rewriteHref }: Projec
...
+              <TableHead className={noResults && 'text-foreground-muted'}>
+                Created
+              </TableHead>
             </TableRow>
           </TableHeader>
           <TableBody>
             {noResultsFromStatusFilter ? (
-              <TableRow>
-                <TableCell colSpan={5} className="p-0">
+              <TableRow className="[&>td]:hover:bg-inherit">
```
dnywh, line 155: Prevent a hover state if there are no actual rows, as per [design system](https://supabase-design-system.vercel.app/design-system/docs/ui-patterns/empty-states#zero-results).

dnywh: Would be cool to systemise this somehow.

apps/studio/components/interfaces/Home/ProjectList/ProjectTableRow.tsx
dnywh, line 1: Match nicer styling set out in `ProjectCard`.

apps/studio/components/ui/ComputeBadgeWrapper.tsx
```diff
@@ -79,7 +79,7 @@ export const ComputeBadgeWrapper = ({
   return (
     <HoverCard onOpenChange={() => setOpenState(!open)} openDelay={280}>
       <HoverCardTrigger asChild className="group" onClick={(e) => e.stopPropagation()}>
-        <div>
+        <div className="flex items-center">
```
dnywh, line 82: Required in order to remove random margins. See prod for extraneous `mt-6` on this instance because of lack of sizing information.

apps/studio/pages/org/[slug]/index.tsx
```diff
@@ -20,17 +21,24 @@ const ProjectsPage: NextPageWithLayout = () => {
...
+                Set up multi-factor authentication (MFA) on your account to access this
+                organization’s projects.
+              </>
+            }
+            actions={
+              <Button asChild type="default">
+                <Link href="/account/security">Set up MFA</Link>
+              </Button>
+            }
+          />
```
Comment (dnywh) [inline apps/studio/pages/org/[slug]/index.tsx:40]: Slightly nicer Admonition presentation and copy.

apps/studio/pages/org/[slug]/index.tsx
```diff
@@ -20,17 +21,24 @@ const ProjectsPage: NextPageWithLayout = () => {
...
         ) : (
-          // [Joshen] Very odd, but the h-px here is required for ProjectList to have a max
-          // height based on the remaining space that it can grow to
-          <div className="flex flex-col gap-y-4 flex-grow h-px">
```
Comment (dnywh) [inline apps/studio/pages/org/[slug]/index.tsx:33]: Removing this fixes:

- Clipping on top of card
- Full-height project table list

Was that last one intentional @joshenlim? If so, are you happy to go back to a table that is only as big as its contents? It’s the pattern we’ve reverted to elsewhere, like on file buckets.

packages/ui/src/components/shadcn/ui/badge.tsx
```diff
@@ -28,12 +28,16 @@ export interface BadgeProps
...
     VariantProps<typeof badgeVariants> {}
 
-function Badge({ className, variant = 'default', children, ...props }: BadgeProps) {
-  return (
-    <div className={cn(badgeVariants({ variant }), className)} {...props}>
-      {children}
-    </div>
-  )
-}
+// Forward refs in order to allow tooltips to be applied to the badge
```
dnywh, line 31: As it says in the comment. In other words:

I couldn’t use Badge `aschild` in a Tooltip without this. The alternative is to wrap `Badge` in a `span` but that felt messy.

apps/studio/components/interfaces/Home/ProjectList/EmptyStates.tsx
joshenlim: makes sense i think! can extend NoSearchResults to handle custom text

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
```diff
@@ -159,38 +138,38 @@ export const ProjectList = ({ organization: organization_, rewriteHref }: Projec
...
-              <TableHead>Status</TableHead>
-              <TableHead>Compute</TableHead>
-              <TableHead>Region</TableHead>
-              <TableHead>Created</TableHead>
-            </TableRow>
-            <TableRow className="!border-b-0">
-              <TableCell colSpan={5} className="p-0">
-                <LoadingLine loading={isFetching} />
-              </TableCell>
+              <TableHead className={cn(noResults && 'text-foreground-muted')}>Project</TableHead>
```
joshenlim, line 146: very nice touch ✨

joshenlim (approved):
Verified changes locally - double checked load more behaviour too

apps/studio/components/interfaces/Home/ProjectList/ProjectList.tsx
dnywh: Thanks!

kemaldotearth (approved):
YES, YES & YES.

PR #41180, opened 2025-12-09
Use timestamp info for api keys last used

joshenlim:
## Context

Came across the API keys page and realised this can be improved slightly.

- Update column "Last seen" to "Last used"
- Show a loader when loading last used data
- Use `TimestampInfo` component to render Last used value

## To test

I don't really have usage data on staging/local to test this though, but I did hardcode data locally to test if that's an option
Can hardcode the value returned from `useLastSeen` in `SecretAPIKeys.tsx` to return for e.g 
```{hash: the API key's hash value, timestamp: 1765206756196 }```

Once changes are through though, we can test on the Misc Use prod project


alaister (approved):
LGTM!

- Tested on preview

PR #41183, opened 2025-12-09
fix(ui): update Slider component

fsansalvadore:
Update Slider to support double knobs for `min` and `max` range.

Screenshot 2025-12-09 at 09 40 45


kemaldotearth (approved):
:shipit:

PR #41199, opened 2025-12-09
feat(auth): improved user search

fadymak:
Improved user search is only shown when a user has opted-in and the indexes have successfully been created (see https://github.com/supabase/supabase/pull/40955).

Changes when improved user search is **enabled**:

1. Default to `created_at desc` sorting on page load (uses `idx_users_created_at_desc` index)
2. Remove the `Unified search` option from the sort by dropdown
3. Introduce searching on the `raw_user_metadata.name` column (uses `idx_users_name` index)
4. Simplified generated SQL query to leverage the new indexes better
    - User ID search is always treated as an exact match search (using btree index) e.g. `= 'UUID'`
    - Email searches will always use a range search (using btree index) for exact and prefix searches + ordering
    - Phone searches will always use a range search (using btree index) e.g. `>= 123 AND < 124`
    - Name will now only perform a range search on `raw_user_metadata.name` using btree index instead of multiple metadata fields (e.g. `display_name`, `full_name`, etc)
5. Use cursor-based pagination instead of offset-limit

Changes that apply regardless of improved user search:

1. Fixed phone number queries to not include leading `+` as phone numbers are stored without it in the database

Screenshot 2025-12-09 at 17 04 22
Screenshot 2025-12-09 at 17 04 05

## Summary by CodeRabbit

* **New Features**
  * Optional "improved search" opt‑in with a Name filter option
  * Cursor‑based pagination for faster, more accurate search results
  * Search input placeholder and filter dropdown adapt to improved search state
  * Opt‑in and progress UI for enabling enhanced search indexes

* **Bug Fixes**
  * More reliable user display name resolution from metadata

* **Chores**
  * Default user sort order updated to created_at:desc


apps/studio/components/interfaces/Auth/Users/UsersV2.tsx
```diff
@@ -195,6 +199,75 @@ export const UsersV2 = () => {
   const totalUsers = totalUsersCountData?.count ?? 0
   const isCountWithinThresholdForSortBy = totalUsers <= SORT_BY_VALUE_COUNT_THRESHOLD
 
+  const isImprovedUserSearchFlagEnabled = useFlag('improvedUserSearch')
```
fadymak, line 205: This section is existing code that was reordered to use the query results in dependencies.

hf (approved):

ivasilov (approved):

hf (approved):

packages/pg-meta/src/sql/studio/get-users-paginated.ts
```diff
@@ -139,3 +164,151 @@ from
...
+  const sortOrder = order ?? 'desc'
+
+  // Cursor-based pagination: fetch rows after the cursor position
+  if (cursor) {
+    const operator = sortOrder === 'desc' ? '<' : '>'
+    // When sorting by id, no need for a composite cursor since id is already unique
+    if (sortOn === 'id') {
+      conditions.push(`id ${operator} '${cursor.id}'::uuid`)
+    } else {
+      conditions.push(`("${sortOn}", id) ${operator} ('${cursor.sort}', '${cursor.id}'::uuid)`)
```
avallete, line 260: **nitpick**

Can be addressed in a follow up PR, but should try to avoid manual escaping as much as possible

Instead
```suggestion
      conditions.push(`(${ident(sortOn)}, id) ${operator} (${literal(cursor.sort)}, ${literal(cursor.id)}::uuid)`)
```

There is a bunch of other place where this could be applied.

avallete (approved):
Just a minor nitpick, that can be addressed in a follow up PR.

packages/pg-meta/src/sql/studio/get-users-paginated.ts
fadymak: Thanks @avallete! I wasn't aware of the literal and ident helpers 🙇  I've made the changes — there are quite a few instances in this file so I'll push them up as a separate PR to keep this one from growing

packages/pg-meta/src/sql/studio/get-users-paginated.ts
```diff
@@ -139,3 +164,140 @@ from
...
+    }
+  }
```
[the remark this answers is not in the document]
fadymak: Will address in followup PR (see https://github.com/supabase/supabase/pull/41199#discussion_r2645278036)

PR #41201, opened 2025-12-09
fix: temp remove api key queries

Ziinc:
temporarily removes api key queries


awaseem:
Re-doing this with a flag. Closing

PR #41205, opened 2025-12-09
fix: updated flag check for undefiend

awaseem:

YES

Minor fix for flag service

### DIFF
--- a/apps/studio/components/interfaces/APIKeys/SecretAPIKeys.tsx
+++ b/apps/studio/components/interfaces/APIKeys/SecretAPIKeys.tsx
@@ -79,7 +79,7 @@ export const SecretAPIKeys = () => {
     isError: isErrorApiKeys,
   } = useAPIKeysQuery({ projectRef, reveal: false }, { enabled: canReadAPIKeys })
 
-  const hideApiKeyLastUsed = useFlag('HideApiKeyLastUsed')
+  const hideApiKeyLastUsed = useFlag('HideApiKeyLastUsed') ?? true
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/APIKeys/SecretAPIKeys.tsx
```diff
@@ -79,7 +79,7 @@ export const SecretAPIKeys = () => {
     isError: isErrorApiKeys,
   } = useAPIKeysQuery({ projectRef, reveal: false }, { enabled: canReadAPIKeys })
 
-  const hideApiKeyLastUsed = useFlag('HideApiKeyLastUsed')
+  const hideApiKeyLastUsed = useFlag('HideApiKeyLastUsed') ?? true
```
awaseem, line 82: null collase, if value is false then false takes over rather than ||

awaseem:
updated the flag recently, so it might take some time to cache

apps/studio/components/interfaces/APIKeys/SecretAPIKeys.tsx
jordienr: I'd prefer a positive boolean instead (showWhatever) so it's easier to read

jordienr (approved):
Lgtm but can't test properly right now

apps/studio/components/interfaces/APIKeys/SecretAPIKeys.tsx
awaseem: Can follow up with that after, just want to close the leak 😬

PR #41210, opened 2025-12-09
Simplify realtime experiment to rely on PostHog targeting

seanoliver:
The original PR (#39748) had the experiment targeting "new projects" (created in the last 7 days) with date logic baked into the hook. We actually want to target new *users* instead, and it makes more sense to handle that targeting entirely in PostHog rather than duplicating cohort logic in the codebase.

**Why user-level targeting?**

We learned from the table quickstart experiment that project-level targeting causes problems: users with multiple projects can get bucketed into different variants as they switch between projects, which breaks the analytics. By targeting at the user level in PostHog, each user gets a consistent variant across all their projects.

This simplifies `useRealtimeExperiment` pretty significantly - we removed all the dayjs date math, the `isNewProject` tracking, and the `projectInsertedAt` prop that call sites had to pass. The hook now just checks the PostHog flag and fires an exposure event.

**Changes**

- Removed date-based targeting from `useRealtimeExperiment` (PostHog handles user cohort targeting)
- Removed `projectInsertedAt` prop from hook - call sites no longer need to pass project data
- Added `experiment_id` to the exposure event for easier querying
- Removed `days_since_project_creation` from telemetry since we're not tracking that anymore
- Cleaned up verbose JSDoc comments

The conversion event is `table_realtime_enabled` which was already instrumented.

**Testing**

Verified types pass and the hook still returns the correct variant based on the PostHog flag. The exposure event fires once per session when viewing a table.

## Summary by CodeRabbit

* **Refactor**
  * Simplified realtime experiment variant selection by removing date-based "new project" gating.
  * Kept realtime enablement logic intact while reducing inputs to the experiment decision.

* **Chores**
  * Updated experiment telemetry: added an experiment identifier and removed legacy project-age metric; exposure tracking now handles errors more robustly.

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
@@ -113,7 +113,6 @@ export const GridHeaderActions = ({ table, isRefetching }: GridHeaderActionsProp
   const isRealtimeEnabled = realtimeEnabledTables.some((t) => t.id === table?.id)
 
   const { activeVariant: activeRealtimeVariant } = useRealtimeExperiment({
-    projectInsertedAt: project?.inserted_at,
     isTable,
     isRealtimeEnabled,
   })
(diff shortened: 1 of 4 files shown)


4L3k51 (approved):

alaister (approved):
LGTM!

PR #41218, opened 2025-12-10
Enable or disable Data API access per table

SaxonF:
image

This adds an enable/disable switch to table creation/editing panel via the table editor. The purpose is to give users more control and visibility over which tables are exposed via the Data API. This works in conjunction with exposed schemas so the setting is disabled if a schema is not exposed via the API.

Under the hood this just grants or revokes all privilages to anon and authenticated roles. 

Note: We need to investigate other possible side effects of revoking access beyond API.

To test:
1. Create a new table to public schema without touching the new setting
2. Notice its enabled by default ie no badge in table list in editor
3. Create a new table to public schema and disable setting
4. Notice badge shows on that table
5. Create a new schema
6. Create a new table under that schema and notice switch is disabled
7. Add new schema to exposed schemas list in API settings
8. Return and notice you can enable/disable API access

## Summary by CodeRabbit

* **New Features**
  * Added per-table API access privileges management with role-based control (anonymous and authenticated users)
  * New API access toggle in table editor for enabling/disabling Data API exposure
  * Visual globe indicator shows tables exposed via the Data API in table list

* **Bug Fixes**
  * Fixed multi-select component closing behavior to properly close on outside clicks

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
@@ -4,6 +4,10 @@ import { isEmpty, isUndefined, noop } from 'lodash'
 import { useState } from 'react'
 import { toast } from 'sonner'
 
+import { useTableApiAccessPrivilegesMutation } from '@/data/privileges/table-api-access-mutation'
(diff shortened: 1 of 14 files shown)


charislam:
@coderabbit full review

charislam (approved):

PR #41219, opened 2025-12-10
fix: go to blog anchor links on initial page load

w3b6x9:


apps/www/app/blog/[slug]/BlogPostClient.tsx
```diff
@@ -29,13 +29,66 @@ export default function BlogPostClient(props: BlogPostPageProps) {
...
+  const hasScrolledToHash = useRef(false)
 
   const { data: livePreviewData, isLoading: isLivePreviewLoading } = useLivePreview({
     initialData: props.blog,
     serverURL: CMS_SITE_ORIGIN || 'http://localhost:3030',
     depth: 2,
   })
 
+  // Handle scrolling to anchor link on initial page load
+  // This is needed because the content is dynamically loaded and the browser's
```
Comment (MildTomato) [inline apps/www/app/blog/[slug]/BlogPostClient.tsx:41]: > This is needed because the content is dynamically loaded and the browser's

it shouldn't be

PR #41227, opened 2025-12-10
fix: Hide favourites and share snippets on self-hosted variant

ivasilov:


alaister (approved):
LGTM!

- Tested on both previews

PR #41228, opened 2025-12-10
feat(reports): support new database report granularity FE-2221

jordienr:
- 7 day or less = 10min ganularity
- 24h or less = 1min granularity


alaister (approved):
LGTM!

PR #41247, opened 2025-12-10
chore: use shadcn's radio component in create function form

marsou001:
As per @ivasilov, the radio form control used to determine the security setting (Security Definer vs Security Invoker) of the function to be created is [deprecated and ought to be replaced](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Database/Functions/CreateFunction/index.tsx#L345), which I did using Shadcn's radio component.

I made sure the UI didn't look different.


Hallidayo:

PR #41254, opened 2025-12-11
Fix creating users not invalidating users list and count

joshenlim:
## Context

Creating a new user from the auth users UI in the dashboard doesn't invalidate the list of users at the moment - I'm actually not too sure when this broke, but it was due to the react query keys.

`create-user-mutation` will invalidate the `usersinfinite` key without params, but because the key itself doesn't have a `.filter(Boolean)`, the invalidation from the mutation is hence trying to invalidate `['projects', projectRef, 'users-infinite', undefined]` then.

Adding a `.filter(Boolean)` should do the trick, including for user counts

## To test

- [ ] Verify that creating a new user in the auth users UI should invalidate the list of users

## Summary by CodeRabbit

* **New Features**
  * Added user search and indexing capabilities for improved user management.
  * Enhanced authentication configuration update workflows.

* **Improvements**
  * Strengthened type safety in user creation forms.
  * Improved data consistency through enhanced cache management.


awaseem (approved):
LGTM nice fix

PR #41264, opened 2025-12-11
chore: Refactor the generation of ids for snippets

ivasilov:
## Summary by CodeRabbit

* **New Features**
  * Editor and reports now use generated, consistent snippet titles and deterministic IDs for new snippets.
  * Editor now accepts a snippet name when creating snippets.
  * Platform-only menu item hidden when not available.

* **Refactor**
  * Unified snippet creation so new snippets use a single, consistent identifier source across editor, reports, drag/drop and templates.
  * Removed client-side random ID generation in favor of deterministic/name-derived IDs.

* **Tests**
  * Added tests for deterministic identifier generation.


apps/studio/components/interfaces/SQLEditor/SQLEditor.tsx
```diff
@@ -121,10 +123,13 @@ export const SQLEditor = () => {
...
   // shows up when clicking on the SQL Editor while being in the SQL editor on a random snippet.
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  const generatedId = useMemo(() => uuidv4(), [urlId])
+  const [generatedNewSnippetName, generatedId] = useMemo(() => {
+    const name = generateSnippetTitle()
+    return [name, generateUuid([name])]
+  }, [urlId])
+
   // the id is stable across renders - it depends either on the url or on the memoized generated id
   const id = !urlId || urlId === 'new' ? generatedId : urlId
```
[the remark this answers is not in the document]
ivasilov: Fixed.

apps/studio/components/interfaces/SQLEditor/MonacoEditor.tsx
```diff
@@ -189,15 +190,16 @@ const MonacoEditor = ({
...
-          id,
-          name: untitledSnippetTitle,
+          idOverride: id,
+          name: snippetName,
           sql: value,
           owner_id: profile?.id,
           project_id: project?.id,
         })
         snapV2.addSnippet({ projectRef: ref, snippet })
+        snapV2.addNeedsSaving(snippet.id)
```
joshenlim: hmm just to check if this one's needed? cause i think the `useEffect` below [here](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/SQLEditor/MonacoEditor.tsx#L208) will also trigger the save for the snippet, so it feels like a duplicate action

can confirm the following: lets say im writing a new snippet and im typing fast, we'll be sending a PUT content request when the snippet gets created, and after i stop typing, we'll send another PUT request after the debounce duration

it's a nit really, but am thinking we can remove this to save us one API call

joshenlim (approved):
Verified changes on staging preview - behaviour seems to be status quo
Also check other areas of the dashboard that creates snippets e.g Inline SQL Editor and AI Assistant

Also checked locally for self-host / CLI mode, there's some existing odd UX for snippets in self hosted (e.g tabs - given that snippets aren't saved yet, but the tabs persist due to local storage), but those are unrelated to the changes here - can address separately

apps/studio/components/interfaces/SQLEditor/MonacoEditor.tsx
ivasilov: The call fixes a bug where a single character snippet would not be saved. By adding `snapV2.addNeedsSaving`, the snippet will be saved after the first character. You can easily replicate the bug by:
1. create a new snippet
2. write a single character
3. refresh the page and you'll see `Unable to find snippet with ID`

I agree that it's an edge case. I would agree on removing it.

PR #41270, opened 2025-12-11
fix(docs): make code blocks overflow scrollable.

mabouguerra:

YES

Bug fix

In the docs, text overflow in code blocks is hidden and cannot be scrolled, thus making them unusable on mobile.

Screenshot 2025-12-11 at 16 49 42

## What is the new behavior?

Code blocks now have `overflow: auto` so that all their content can be accessed on smaller devices.

Screenshot 2025-12-11 at 16 51 00

## Summary by CodeRabbit

* **Style**
  * Code blocks now support horizontal and vertical scrolling for content overflow.

### DIFF
--- a/apps/docs/features/ui/CodeBlock/CodeBlock.tsx
+++ b/apps/docs/features/ui/CodeBlock/CodeBlock.tsx
@@ -64,7 +64,7 @@ export async function CodeBlock({
         'group',
         'relative',
         'not-prose',
-        'w-full overflow-hidden',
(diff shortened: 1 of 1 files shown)


7ttp (commented):
Thanks for submitting this! 💚
However, it looks like this change has already been `merged` into `master`.
The conflict markers show that `master` already has `overflow-x-auto`, which means this fix is already live.
Closing this as the issue is already resolved. 
Thanks for your contribution though! 🙏

PR #41230, opened 2025-12-10
docs: byom mcp server without auth

Rodriguespn:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

Docs update

## What is the current behavior?

There is no documentation for deploying custom MCP servers on Supabase Edge Functions. The existing MCP page only covers the Supabase MCP server.

## What is the new behavior?

Adds a comprehensive guide for building and deploying remote MCP servers on Supabase Edge Functions, without authentication:

- **New docs page**: `apps/docs/content/guides/getting-started/byom.mdx` - Complete tutorial covering prerequisites, project setup, creating an MCP server with the official TypeScript SDK, local development, testing with MCP Inspector, and production deployment
- **Example code**: `examples/edge-functions/supabase/functions/mcp/simple-mcp-server/` - Ready-to-use MCP server implementation with Hono routing and tool registration
- **Navigation updates**:
  - Renamed "Model context protocol (MCP)" to "Supabase MCP server"
  - Added new "Deploy MCP servers" entry under AI Tools section

## Preview Link
https://docs-git-docs-byom-mcp-server-supabase.vercel.app/docs/guides/getting-started/byom

Closes https://linear.app/supabase/issue/AI-305/byom-deploy-mcp-servers-without-authentication-one-pager

PR #41274, opened 2025-12-11
docs(update): address review comments for MCP server guide

Rodriguespn:
This PR addresses the review comments from PR #41230, which was accidentally merged before being fully reviewed.

[Docs Preview](https://docs-git-docs-byom-mcp-server-supabase.vercel.app/docs/guides/getting-started/byo-mcp)

Refs: #41230

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the navigation link to the MCP server setup guide.

* **Documentation**
  * Revised MCP server guide: updated naming, routing, file paths, and local/production instructions.
  * Added an admonition about unauthenticated MCP requests and forthcoming OAuth 2.1 support.
  * Clarified inspector and local endpoint guidance.

* **Chores**
  * Updated example project dependency versions.


apps/docs/content/guides/getting-started/byo-mcp.mdx
```diff
@@ -68,14 +70,15 @@ import { StreamableHTTPTransport } from '@hono/mcp'
 import { Hono } from 'hono'
 import { z } from 'zod'
 
+// Edge Functions route to /<functionName>/*, so set the base path accordingly.
```
gregnr: Wouldn't this actually be `/functions/v1/<functionName>` in practice? I might be missing something.

apps/docs/content/guides/getting-started/byo-mcp.mdx
```diff
@@ -158,18 +115,18 @@ supabase start
...
 
 ```
-http://localhost:54321/functions/v1/simple-mcp-server/mcp
+http://localhost:54321/functions/v1/mcp
 ```
 
 <Admonition type="note">
 
-The `--no-verify-jwt` flag disables JWT verification at the Edge Function layer. This is required because MCP authentication is handled by the MCP server itself, not by Supabase's standard JWT validation.
+The `--no-verify-jwt` flag disables JWT verification at the Edge Function layer so your MCP server can accept unauthenticated requests. If your mcp server needs authentication, keep your function private until MCP auth support on Edge Functions is available.
```
gregnr: > keep your function private

We might want to expand on this. Do we mean to omit this flag? Tbh I'm not sure we actually can support authentication even as a fallback until we integrate it properly with project level oauth.

Rodriguespn: My initial though is to omit the flag and the /mcp endpoint is somehow protected. This is useful for MCP servers that are used internally inside orgs for example so I see some value on suggesting this. Of course the best case scenario is to support oauth 2.1

Rodriguespn: I agree that it could be more explicit. Let me iterate this

apps/docs/content/guides/getting-started/byo-mcp.mdx
Rodriguespn: You're right, the actual URL path is `/functions/v1/<functionName>`, not `/<functionName>/*`.
Edge Functions automatically strip the `/functions/v1/<functionName>` prefix before passing the request to the function handler, so `basePath` isn't needed. I included it because it was required when testing with `mcp-lite` last month, but it's not needed for this setup.
I've removed the misleading comment and all `basePath` references from the tutorial and example. The code is now just 
```ts
const app = new Hono().
```
Thanks for catching that 🙌

apps/docs/content/guides/getting-started/byo-mcp.mdx
Rodriguespn: @gregnr what do you think?

> If you need to protect your MCP server, you can omit the `--no-verify-jwt` flag to enable JWT verification at the Edge Function layer. This requires clients to provide a valid Supabase JWT token. However, this doesn't provide full MCP authentication. For that, wait for OAuth 2.1 support on Edge Functions (coming soon).

apps/docs/content/guides/getting-started/byo-mcp.mdx
```diff
@@ -244,7 +201,7 @@ server.registerTool(
 
 <Admonition type="caution">
 
-MCP authentication is not yet supported on Edge Functions. For now, MCP servers deployed on Supabase Edge Functions are publicly accessible.
+Full MCP authentication (OAuth 2.1) is not yet supported on Edge Functions. For basic protection, you can omit the `--no-verify-jwt` flag to require Supabase JWT tokens.
 
 </Admonition>
```
gregnr: Perhaps we can omit this for now

apps/docs/content/guides/getting-started/byo-mcp.mdx
gregnr: The more I look at this the more I think we should just say its not supported for now. So something like:

> The `--no-verify-jwt` flag disables JWT verification at the Edge Function layer so your MCP server can accept unauthenticated requests. Authenticated MCP support is coming soon.

gregnr (approved):
Approving pending the last 2 comments

ChrisChinchilla:
I am happy with this after addressing the review comments too

Rodriguespn:
Thx @gregnr and @ChrisChinchilla for the comments. Will merge this once I have a green pipeline

PR #41284, opened 2025-12-11
fix(studio): use older supported syntax rather than Map.groupBy

charislam:
Some users' browsers do not support Map.groupBy (used in Storage.utils.ts)

## Summary by CodeRabbit

* **Refactor**
  * Optimized the internal logic for grouping storage policies by bucket.


awaseem (approved):
LGTM

PR #41320, opened 2025-12-12
fix(studio): show correct column error on composite foreign keys

### LINKED ISSUE #41319: Foreign key side panel displays wrong columns on foreign keys misconfiguration.
## Describe the bug

If you create a foreign key relation and the setup of the second pair or more is not right, like if the source type and target type don't match, you get an error warning that the foreign key can't be edited, or that the source column's type will be changed later to match the target type if the column is new.

The problem is when you have a composite foreign key relation. If you add more than one pair (source column, target column) and the setup of the second pair or more is not right, the error UI shows the first pair's columns, instead of the names of the mismatching columns.

## To Reproduce

1. Go to `/dashboard/project/${projectRef}/editor/${tableId}`

2. Choose a column and click "edit column"

3. Click "add foreign key"

4. Add two pairs to form a composite relation.

5. The second pair should have mismatching types to see the error.

6. The error will display the wrong columns (columns of the first pair).

## Screenshots

Image

I did make a video but its size was way beyond the allowed 10mb, so I chose the screenshot.

The columns of the first pair (`id`, and `id`) are both of type `int8`, but the second pairs columns don't match (`uuid` for `mio` and `timestampz` for `created_at`).

Notice that the types are correctly displayed but not the corresponding columns.

marsou001:
Fixes #41319 

All non-correctly configured columns in foreign key selector panel are correctly displayed, along with their corresponding types.

## Summary by CodeRabbit

* **Refactor**
  * Strengthened type safety for foreign-key selection and validation; error objects now include both source and target column references and their types.
  * Validation and error/notice handling updated to use the expanded error shapes.
  * UI error notices now surface source/target names and types directly for clearer, more accurate messages.

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ForeignKeySelector/ForeignKeySelector.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ForeignKeySelector/ForeignKeySelector.tsx
@@ -25,7 +25,7 @@ import { ActionBar } from '../ActionBar'
 import { NUMERICAL_TYPES, TEXT_TYPES } from '../SidePanelEditor.constants'
 import type { ColumnField } from '../SidePanelEditor.types'
 import { FOREIGN_KEY_CASCADE_OPTIONS } from './ForeignKeySelector.constants'
-import type { ForeignKey } from './ForeignKeySelector.types'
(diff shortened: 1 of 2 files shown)


awaseem (approved):
Tested in staging and self hosted. Working great, plus thank you cleaning the types!

PR #41322, opened 2025-12-12
fix(telemetry): pass PostHog distinct_id in identify request

seanoliver:
## Problem

Our first-touch attribution data in PostHog stopped working after a Dec 3rd change. The root cause: the backend identify endpoint generates its own `anonymous_id` cookie value instead of using the actual PostHog JS SDK's `distinct_id`.

When PostHog's `alias()` is called, it links this server-generated ID (which has no attribution data) to the authenticated user, instead of linking the real PostHog anonymous profile (which contains `$initial_referrer`, `$initial_utm_source`, etc.).

## Changes

- Added `getDistinctId()` method to `PostHogClient` to expose the underlying `posthog.get_distinct_id()` value
- Updated `useTelemetryIdentify` to pass this value as `anonymous_id` in the identify request
- Updated the `TelemetryIdentifyBodyV2` type to include `anonymous_id` (will be regenerated when backend PR merges)

This is the frontend counterpart to https://github.com/supabase/platform/pull/28375 which adds support for accepting `anonymous_id` in the backend identify endpoint.

## Testing

- Verified TypeScript passes
- Will need manual testing after both PRs are deployed: confirm that logging in preserves first-touch attribution data on the user profile in PostHog

Fixes GROWTH-567

## Summary by CodeRabbit

* **Chores**
  * Telemetry identify payloads now include an optional anonymous identifier when available to better preserve first-touch attribution and link analytics across systems.
  * Client initialization flow hardened to prevent double-initialization and ensure pending telemetry (groups, identifications, events) is applied once ready.
  * Backend identify calls now attach the client anonymous ID when present to improve cross-system attribution.

### DIFF
--- a/packages/api-types/types/platform.d.ts
+++ b/packages/api-types/types/platform.d.ts
@@ -9418,6 +9418,8 @@ export interface components {
       reset_project?: boolean
     }
     TelemetryIdentifyBodyV2: {
+      /** PostHog JS SDK's distinct_id - used for aliasing anonymous → authenticated users */
+      anonymous_id?: string
       organization_slug?: string
       project_ref?: string
       user_id: string
--- a/packages/common/posthog-client.ts
+++ b/packages/common/posthog-client.ts
@@ -156,6 +156,25 @@ class PostHogClient {
       console.error('PostHog reset failed:', error)
     }
   }
+
+  /**
+   * Returns the current PostHog distinct_id for the user.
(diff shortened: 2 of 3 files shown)


seanoliver:
Good catch @coderabbitai! Fixed in eb9300131a - added a separate `ready` flag that's set in the `loaded` callback, and `getDistinctId()` now checks `this.ready` instead of `this.initialized`.

charislam (approved):

PR #41339, opened 2025-12-15
Fix deleting a user shows User not found toast error

joshenlim:
## Context

Deleting a user will render a "User not found" error toast, along with the deletion success toast

This is due to the `useQueryStateWithSelect` hook which i've started an internal discussion on how to improve the DX - the changes here fixes the immediate issue, but we may change the fix depending on the outcome of the discussion

## To test

- [ ] Verify that there's no error toast when deleting a user in the Auth Users page

## Bug Fixes
* Enhanced user data loading reliability by tightening query activation conditions to confirm data presence
* Corrected callback execution sequence in user deletion operations to maintain proper ordering before cache updates


awaseem (approved):
Clean and simple!

PR #41341, opened 2025-12-15
Fix replica panel closing when clicking new replica CTA from database selector

joshenlim:
## Context

While on the settings -> infrastructure page, if you hit the "Deploy a new replica" button, it opens the replica panel alright. However if you scroll down to "Infrastructure Activity" and click the database selector -> "Create a new read replica" CTA, the replica panel seems to open then close immediately.

Am thinking this is happening due to a shift in browser focus when opening the panel from a "portal" based component (e.g Dialog, Sheet, Dropdown, etc), causing the `onCancel` method of the `SidePanel` component within `DeployNewReplicaPanel.tsx` to be called (it happens even if I use the new `Sheet` component)

Adding a `timeout` when clicking on the "Create a new read replica" CTA within `DatabaseSelector` will do the trick by opening the panel only after the focus is shifted - fixes the immediate problem although ideally I'd wanna figure out a better solution for this

## Changes involved
- [ ] Add `timeout` when clicking "Create new read replica" in `DatabaseSelector`
- [ ] Remove `export default` from `DatabaseSelector` and update import statements from other components
- [ ] Clean up `PreviewFilterPanelWithUniversal.tsx` -> just saw that it had an unused method

## To test

- [ ] As per PR description above, try opening the replica panel from the infrastructure settings page in (1) the node diagram and (2) infrastructure activity
  - It's not bound to just infrastructure activity btw, so could be SQL editor, Connect UI - anywhere that renders the DatabaseSelector component

## Summary by CodeRabbit

* **Refactor**
  * Refactored the Database Selector component to streamline the replica creation flow and improve internal component structure.


awaseem (approved):
LGTM tested on staging

PR #41355, opened 2025-12-15
feat(studio): endpoint to retrieve incident status

charislam:
Incident status banners/alerts should be automatically determined from the status page, rather than being manually configured.

Adds a new endpoint, /api/incident-status, to return information about active incidents. This information is fetched from the StatusPage API. The endpoint is cached to reduce number of third-party API requests.

**Testing**

Because Vercel deployment protection is on, the API can't be accessed on the preview site. You'll have to test locally.

Post-merge, we should verify that CDN caching works as expected before UI features are built on top of this.

Resolves FE-2198

## Summary by CodeRabbit

* **New Features**
  * Added an incident status API that returns concise active-incident summaries, supports HEAD/GET with 5-minute caching, and includes robust error handling and logging for invalid responses.

* **Chores**
  * Updated studio build environment to provide Statuspage credentials for incident fetching.
  * Whitelisted the incident-status hosted endpoint so it’s reachable from hosted builds.


apps/studio/pages/api/incident-status.ts
```diff
@@ -0,0 +1,156 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import z from 'zod'
+
+import { IS_PLATFORM } from 'common'
+
+class InternalServerError extends Error {
```
awaseem: Wonder if we should put this in some sort of shared util? Other handlers can use this

awaseem (approved):
Very minor refactor comment

apps/studio/pages/api/incident-status.ts
```diff
@@ -0,0 +1,156 @@
...
+      })
+      return true
+    }
+
+    return scheduledTime <= now
+  })
+
+  return activeIncidents.map((incident) => ({
+    id: incident.id,
+    name: incident.name,
```
charislam, line 103: note to self: also pass status through

PR #41378, opened 2025-12-15
fix(studio): rework global storage size validation

charislam:

YES

Bug fix

The global storage size validation depends on an unpaginated buckets query to determine whether it is lower than any individual bucket's cutoff. This causes a problem for users with tens of thousands of buckets.

## What is the new behavior?

There's a bit of a UX/performance problem here, because in order to determine whether any bucket's `file_size_limit` exceeds the global setting, we need to get the max `file_size_limit` of `storage.buckets` -- however, that column is not indexed.

My workaround is:
- Below a certain threshold (10,000) buckets, the query for max `file_size_limit` is automatically run on form submit.
- Above that threshold, the user must confirm whether they want to run the query. They're still allowed to change the storage config without running it -- this does open a loophole where they can have a global storage setting lower than an individual bucket's file size limit, but though this is a potentially confusing situation, it's not strictly an error. Open to other suggestions for how to balance UX/performance here.

## Additional context

Towards FE-2118

## Summary by CodeRabbit

* **New Features**
  * Storage size-limit validation UI with automatic checks for small projects and a manual "Validate size limit" confirmation flow that reports buckets exceeding the configured file-size limit.

* **User Interface**
  * New error component that highlights the primary affected bucket (linked when available), shows its formatted limit, and offers a tooltip listing other impacted buckets with a summary when more exist.

* **Utilities**
  * Helpers to format byte sizes and encode/decode bucket-limit messages plus a threshold constant driving auto vs. confirm behavior.

* **Data**
  * New queries/estimators to fetch largest bucket limits and live-table estimates.

* **Tests**
  * Tests added for bucket-limit message encoding/decoding.


soedirgo (approved):

joshenlim:
Thinking we should have the save button in a loading state upon clicking while the validation is checking - otherwise it looks like nothing happened until the error message popped up

https://github.com/user-attachments/assets/6f8d46a0-f958-482a-aafd-3cee706925c7

apps/studio/components/interfaces/Storage/StorageSettings/StorageFileSizeLimitErrorMessage.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const [primaryBucket, ...remainingBuckets] = bucketLimitErrorBuckets
+  const otherBuckets = remainingBuckets.slice(0, LARGEST_SIZE_LIMIT_BUCKETS_COUNT - 1)
+
+  const primaryBucketLimit = formatBytesForDisplay(primaryBucket.limit)
+
+  const primaryBucketLink =
+    projectRef !== undefined ? (
+      <InlineLink
+        href={`/project/${projectRef}/storage/files/buckets/${primaryBucket.name}`}
```
joshenlim: could link to open the edit modal directly

```suggestion
        href={`/project/${projectRef}/storage/files/buckets/${primaryBucket.name}?edit=true`}
```

apps/studio/components/interfaces/Storage/StorageSettings/StorageFileSizeLimitErrorMessage.tsx
```diff
@@ -0,0 +1,105 @@
...
+                </span>
+              </TooltipTrigger>
+              <TooltipContent side="bottom">
+                <ul>
+                  {otherBuckets.map(({ name, limit }) => {
+                    const formattedLimit = formatBytesForDisplay(limit)
+                    return (
+                      <li key={name} className="hover:underline underline-offset-2">
+                        {projectRef !== undefined ? (
+                          <Link href={`/project/${projectRef}/storage/files/buckets/${name}`}>
```
joshenlim: likewise for here - could link to open the edit dialog directly

```suggestion
                          <Link href={`/project/${projectRef}/storage/files/buckets/${name}?edit=true`}>
```

apps/studio/components/interfaces/Storage/StorageSettings/StorageFileSizeLimitErrorMessage.tsx
```diff
@@ -0,0 +1,105 @@
...
+              <TooltipTrigger asChild>
+                <span
+                  className={cn(
+                    'underline underline-offset-2',
+                    'decoration-dotted decoration-destructive-500',
+                    'hover:decoration-destructive transition',
+                    'cursor-default'
+                  )}
+                >
+                  {showOtherCount && `+${otherBuckets.length} `}other buckets
```
joshenlim: very very minor nit sorry 😅 
could conditionally render the plural for bucket based on otherBuckets

should be singular if otherBuckets.length === 1

joshenlim:
just adding a small text here in the case of auto validation is `confirm` to kinda signal why the user should validate the size limit

image

joshenlim (approved):
Verified changes on local! I think these are alright - relatively edge case as well since our threshold is 10 buckets, so in most cases the auto validation will kick in

PR #41381, opened 2025-12-16
Prevent ForeignRowSelector from firing multiple saves when clicked

joshenlim:
## Context

When updating a foreign row reference in the Table Editor directly from the grid, when clicking on the row multiple times, the `onSelect` prop in `ForeignRowSelector` gets triggered multiple times too, which in this case results in the `SidePanelEditor` calling `onSaveForeignRow` multiple times, and multiple success toasts will hence show up
image

## Changes involved

- Support async saving for `ForeignRowSelector` by passing in a `isSaving` prop, which will prevent the `onSelect` from getting called when a row is clicked if `isSaving` is true
  - This will prevent multiple saves in this particular scenario
- Also removed the `ActionBar` in `ForeignRowSelector`, shifts the close panel action to the header and added a saving indicator. This also gives more real estate to display the rows too
  image

## To test

- [ ] Verify that when updating a foreign row directly from the grid, clicking the row multiple times quickly shouldn't trigger multiple save requests

## Summary by CodeRabbit

* **New Features**
  * Added a saving indicator and close button to the foreign row selector panel for clearer save-state feedback.

* **Improvements**
  * Row selection is disabled while saving to prevent accidental changes.
  * Schema list is now sorted alphabetically and safely handles missing schema data.
  * Saving state is propagated to the selector so the UI reliably reflects in-progress updates.


apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ForeignKeySelector/ForeignKeySelector.tsx
```diff
@@ -62,14 +62,20 @@ export const ForeignKeySelector = ({
...
+    typeNotice?: { sourceType: string; targetType: string }[]
+  }>({})
+  const hasTypeErrors = (errors?.types ?? []).filter((x) => x !== undefined).length > 0
+  const hasTypeNotices = (errors?.typeNotice ?? []).filter((x) => x !== undefined).length > 0
+
+  const { data: schemas = [] } = useSchemasQuery({
     projectRef: project?.ref,
     connectionString: project?.connectionString,
   })
+  const sortedSchemas = schemas.sort((a, b) => a.name.localeCompare(b.name))
```
ivasilov: this should be `[...schemas].sort`. The current approach sorts the data inside the RQ cache which may cause issues.

Or sort it in RQ fetch function.

joshenlim: gotcha on this! pushed a fix

alaister (approved):
LGTM!

- Tested on preview

PR #41384, opened 2025-12-16
Use totalRows if available from params to validate MAX_EXPORT_ROW_COUNT

joshenlim:
Fixes FE-2276

## Context

When exporting all rows from the table editor, we currently have a limit of 500,000 rows to support exporting via browser - above that we show a toast error to do so via `pg_dump`

However, atm if your table has > 500k rows, but your current view has some filters applied such that the row count with the filters is < 500k rows, the UI prevents you from exporting the rows with that same toast error (which is incorrect)

The validation in the UI currently only considers the table's `live_row_estimate` ([ref](https://github.com/supabase/supabase/blob/9274b81bd76a9fa74e51d1c8577185e4db43fb29/apps/studio/components/layouts/TableEditorLayout/ExportAllRows.tsx#L115)) but we should also be considering the `totalRows` given that we pass that as a param (and `totalRows` should take precedence)

## To test

- [ ] Verify that for a table with > 500k, you can still export the data if you've applied filters and the row count is below 500k
  - Can also test locally to reduce that limit to something smaller like 1k

## Summary by CodeRabbit

* **Improvements**
  * Implemented a 500,000 row maximum limit for bulk data exports in CSV, SQL, and JSON formats to ensure system reliability
  * Enhanced error handling with improved user messaging that includes direct documentation references when export size limits are exceeded
  * Consolidated export validation logic for better consistency and clearer user feedback

### DIFF
--- a/apps/studio/components/grid/components/header/Header.tsx
+++ b/apps/studio/components/grid/components/header/Header.tsx
@@ -1,6 +1,5 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { ArrowUp, ChevronDown, FileText, Trash } from 'lucide-react'
-import Link from 'next/link'
 import { ReactNode, useState } from 'react'
 import { toast } from 'sonner'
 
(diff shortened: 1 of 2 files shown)


awaseem (approved):
LGTM tested locally!

PR #41402, opened 2025-12-16
fix(studio): support for language feature in Storage.utils.ts

charislam:
Iterator.prototype.map() is only newly baseline available, which is causing problems with older browsers.

Switched to Array.from(<iterable>) since that (and Map as an iterable) date from 2015/2016

## Summary by CodeRabbit

* **Refactor**
  * Improved internal implementation of storage policy grouping logic with no impact to functionality or user experience.


awaseem (approved):
LGTM

PR #41405, opened 2025-12-16
chore: Migrate `studio` to use `ui-patterns/shimmeringLoader`

ivasilov:
This PR deletes `ShimmeringLoader` from `studio` and migrates it to use `ui-patterns/ShimmeringLoader`.


packages/ui-patterns/package.json
```diff
@@ -50,6 +50,18 @@
       "import": "./src/Banners/index.ts",
```
awaseem, line 50: Guessing some of these are new because of script ran?

awaseem (approved):
LGTM

packages/ui-patterns/package.json
ivasilov: Yup, other devs added components but haven't run the script.

staaldraad (approved):

PR #41407, opened 2025-12-16
chore: remove tierKey from backups

ignaciodob:
Just removes a leftover plan check from backups

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed backup empty state message to display consistently for all users when no backups are available, improving the backup management experience.

### DIFF
--- a/apps/studio/components/interfaces/Database/Backups/BackupsList.tsx
+++ b/apps/studio/components/interfaces/Database/Backups/BackupsList.tsx
@@ -49,7 +49,6 @@ export const BackupsList = () => {
     },
   })
 
-  const planKey = backups?.tierKey ?? ''
   const sortedBackups = (backups?.backups ?? []).sort(
     (a, b) => new Date(b.inserted_at).valueOf() - new Date(a.inserted_at).valueOf()
   )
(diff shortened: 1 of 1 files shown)


kanishkdudeja (approved):

apps/studio/components/interfaces/Database/Backups/BackupsList.tsx
```diff
@@ -74,7 +73,7 @@ export const BackupsList = () => {
   return (
     <>
       <div className="space-y-6">
-        {sortedBackups.length === 0 && planKey !== 'FREE' ? (
+        {sortedBackups.length === 0 ? (
```
kevcodez, line 76: Don't we need to check for the entitlement here when removing that?

ignaciodob: We have an early return above which ensures that this code block is never reached if `!hasAccess`, so we no longer need this :)

kevcodez (approved):

kevcodez (approved):

ecktoteckto (approved):

PR #41416, opened 2025-12-16
feat(studio): show icon on tables exposed by data api

charislam:

YES

Feature

No labelling for tables exposed via Data API

## What is the new behavior?

Show an icon in the Table Editor for tables that are exposed via the Data API

### Testing

In the public schema, ensure that tables/views have the globe indicator.

In the public schema, ensure that tables with access revoked from `anon` and `authenticated` do not have the globe indicator. You can create one like this:

```sql
create table not_exposed (
  id uuid primary key default gen_random_uuid()
);

revoke all on not_exposed from anon;
revoke all on not_exposed from authenticated;
```

Switch to a non-exposed schema, like `auth`, and ensure that tables do not have the globe indicator.

### Tracking

Resolves FE-2283

## Summary by CodeRabbit

* **New Features**
  * Added a globe badge indicator to display tables exposed via the Data API, with feature flag control for gradual rollout.

* **Bug Fixes**
  * Corrected privilege cache invalidation after table creation and duplication operations to ensure accurate permission state.

* **Style**
  * Updated the RLS disabled button styling from warning to danger intent for better visual clarity.


apps/studio/lib/type-helpers.ts
```diff
@@ -1 +1,3 @@
 export type PlainObject<Value = unknown> = Record<string | number | symbol, Value>
+
+export type Prettify<T> = { [K in keyof T]: T[K] } & {}
```
joshenlim, line 3: just a nit but wondering what this does? does it make the type into a flat object?

apps/studio/lib/data-api-types.ts
joshenlim, line 1: just a nit but i reckon these could all live within `table-api-access-query` as its only being used there (also contextual to that file too)

probably don't need to export all the variables here too thereafter as i don't think i see them being used in anywhere else except in table-api-access-query

joshenlim:
small issue found but newly created tables in an exposed schema won't show the globe icon - might just be missing an invalidation on table creation

joshenlim:
this is totally unrelated, but realised that the CTA here should perhaps be primary, since we're trying to push for secure by default
image

SaxonF:
@copilot make the following changes to [apps/studio/components/layouts/TableEditorLayout/EntityListItem.tsx](https://github.com/supabase/supabase/pull/41416/changes#diff-e57e0dd46c4e487052adf78e587b6a300372ae01e44c84c0a5b6b006be9ba62c)

- When table has API access, has RLS enabled and also has at least 1 policy the tooltip content should be "This table can be accessed by anyone via the Data API as RLS is disabled"
- When table has API access, has RLS enabled but has no policies it should be "This table can be accessed via the Data API but no RLS policies exist so no data will be returned"
- When table has API access but RLS is disabled it shows the unrestricted badge with adjusted tooltip content copy "This table can be accessed by anyone via the Data API as RLS is disabled"

apps/studio/lib/type-helpers.ts
charislam: It makes intersections of object types easier to read: https://www.totaltypescript.com/concepts/the-prettify-helper

apps/studio/lib/data-api-types.ts
charislam: Hmm I put them there in preparation for future PRs because I think they'll be reused in the next step (add Data API access toggling to Table Editor sidebar) but let me double check whether that's likely to be true

charislam: Yeah, we'll be reusing these shortly

joshenlim: gotchaaa yup all good then! 🙂 🙏

apps/studio/lib/type-helpers.ts
joshenlim: very cool! TIL hahaha thanks so muchh 🙏 🙏

joshenlim (approved):

PR #41418, opened 2025-12-16
Clean up dead telemetry code after client-side migration

seanoliver:
## Context

Back when we had server-side page telemetry (the `/platform/telemetry/page` endpoint), `handlePageTelemetry` accepted a `telemetryDataOverride` parameter that let callers pass extra data to the backend. We also had cookie-based logic to persist telemetry data across page loads.

Now that page telemetry goes directly to PostHog client-side (see GROWTH-562/563), that backend endpoint is gone and this code is dead.

## Changes

- Remove the unused `telemetryDataOverride` parameter from `handlePageTelemetry`
- Simplify the cookie handling — we still clear the legacy cookie if it exists, but no longer parse it since we don't send data anywhere
- Remove the try/catch block that was handling JSON parsing errors for the cookie data

## Notes

The type `TelemetryPageBodyV2` that this parameter referenced still exists in `api-types` (it'll get cleaned up whenever someone next runs codegen against the updated backend spec). This is just the frontend cleanup portion.

---

Closes GROWTH-564

## Summary by CodeRabbit

* **Refactor**
  * Simplified telemetry behavior: removed legacy cookie-based telemetry path, now clears any old telemetry cookie and uses a single, consistent tracking flow to improve reliability and reduce duplicate or stale telemetry data.

### DIFF
--- a/packages/common/telemetry.tsx
+++ b/packages/common/telemetry.tsx
@@ -50,8 +50,7 @@ export function handlePageTelemetry(
     [key: string]: unknown
   },
   slug?: string,
-  ref?: string,
-  telemetryDataOverride?: components['schemas']['TelemetryPageBodyV2']
+  ref?: string
 ) {
   // Send to PostHog client-side (only in browser)
   if (typeof window !== 'undefined') {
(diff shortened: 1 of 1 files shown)


joshenlim (approved):
Changes LGTM!

seanoliver:
Closing in favor of GROWTH-573: this PR removes the telemetry cookie override path, but we actually need to restore/use the first-touch snapshot (and extend to non-GDPR) to fix internal-referrer misattribution. We'll revisit cleanup/hardening after the core fix lands.

PR #41423, opened 2025-12-17
Fix table quickstart experiment multi-bucket exposure issue

seanoliver:
## Problem

We noticed ~20% of exposures in the table quickstart experiment ended up in multiple buckets. After digging in, turns out this wasn't a PostHog bucketing issue - the flag evaluation is correctly user-level based on gotrue_id.

The actual culprit was our exposure tracking. We were using a React ref (`useRef`) to dedupe the `table_quickstart_opened` event, but refs reset whenever the component remounts. So if someone visited multiple new projects (or just refreshed the page), we'd fire duplicate exposure events.

On top of that, we had a 7-day project age filter that added unnecessary complexity - same user could see the experiment on new projects but not old ones, creating weird exposure patterns.

Additionally, the localStorage key was browser-global and allowlisted, meaning on shared devices, only the first user to log in would have their exposure tracked - subsequent users would be silently skipped, further skewing experiment metrics.

## Changes

- Removed the `isNewProject` check that restricted the experiment to projects < 7 days old
- Swapped `useRef` for `useLocalStorageQuery` so exposure tracking persists across sessions/navigation
- **Scoped exposure key to user ID** (`TABLE_QUICKSTART_EXPOSURE_TRACKED(userId)`) to prevent shared device contamination
- Added `profile?.id` guard to `shouldTrackExposure` to ensure only authenticated users are tracked
- Removed key from allowlist so it clears on logout (prevents cross-user contamination)

## Testing

- Verified TypeScript type checking passes with zero errors
- Confirmed ESLint passes with no new violations
- Tested that localStorage key is scoped per user: `table-quickstart-exposure-tracked-{userId}`
- Verified exposure tracking fires exactly once per authenticated user
- Confirmed key clears on logout (no longer in allowlist)

GROWTH-572

### DIFF
--- a/apps/studio/components/layouts/Tabs/NewTab.tsx
+++ b/apps/studio/components/layouts/Tabs/NewTab.tsx
@@ -104,7 +104,7 @@ export function NewTab() {
 
   const shouldTrackExposure =
     editor !== 'sql' &&
-    profile?.id &&
+    !!profile?.id &&
     tableQuickstartVariant !== false &&
     tableQuickstartVariant !== undefined
 
--- a/apps/studio/components/layouts/Tabs/NewTab.tsx
+++ b/apps/studio/components/layouts/Tabs/NewTab.tsx
@@ -1,13 +1,12 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
-import dayjs from 'dayjs'
 import { partition } from 'lodash'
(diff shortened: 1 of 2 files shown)


joshenlim (approved):
Changes LGTM! Verified locally

PR #41430, opened 2025-12-17
chore: Align profile endpoint usage with api schema corrections

lforst:
Just a small change where we align the usage of the create profile endpoint data with a small schema correction on the API: https://github.com/supabase/platform/pull/28362

Also, because I bumped the schema and we removed some telemetry endpoints, I removed an unused parameter from the telemetry connection because the type was no longer available.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form reset so optional profile fields restore correctly after updates.

* **Updates**
  * Profile fields (first name, last name, mobile, free project limit) can be nullable/left empty.
  * Organization entitlements route consolidated and response adds auth.mfa_web_authn.
  * New external provider configuration fields added (client id, secret, enabled, email optional).
  * Removed two telemetry page endpoints from the public API.

* **Refactor**
  * Simplified page telemetry handling.


joshenlim (approved):
Verified changes on staging preview (account pref -> update profile)

PR #41447, opened 2025-12-17
Fix first-touch referrer for initial $pageview

seanoliver:
## Problem
We’re seeing a bunch of “first pageviews” in PostHog where the referrer shows up as `supabase.com`, which basically means we’re capturing the first-touch referrer too late (after we’ve already navigated around internally). The code already tries to save an initial telemetry snapshot in a cookie for consented flows, but we never actually use that cookie when sending the initial `$pageview`.

## What happened
At some point we introduced the `supabase-telemetry-data` cookie as a way to hold onto `document.referrer` until consent is granted, but the read path only grabbed the cookie string and then immediately fell back to `getSharedTelemetryData()` (current document state). So if someone lands from Google, clicks into `/dashboard`, then consents, the “first pageview” picks up the internal referrer.

## Changes
- Always write a one-time `supabase-telemetry-data` cookie on initial load (same un-hardened payload as today).
- When we emit the initial `$pageview`, parse that cookie and use it as the source of truth for the first-touch fields instead of recomputing from the current document state.
- Explicitly set `$referrer` / `$referring_domain` from the first-touch snapshot so PostHog doesn’t infer an internal referrer.
- Clear the cookie after we’ve consumed it (and on opt-out), using the right prod domain/path attributes so it actually gets removed.

GROWTH-573

## Summary by CodeRabbit

* **Refactor**
  * Centralized telemetry handling with environment-aware cookie options and unified clear behavior.

* **Changes**
  * Telemetry now uses a single enabled flag; when enabled a one-time cookie is written and used to enrich the initial pageview (includes referrer and referring domain). Page path is detected from the browser location.

* **Bug Fixes**
  * Cookie read/write and clearing behave consistently and are cleared after consumption.

* **Chores**
  * Shared telemetry utilities exported for reuse.


packages/common/telemetry.tsx
```diff
@@ -43,28 +47,117 @@ export const TelemetryTagManager = () => {
...
+      ...(getParam('rdt_cid') && { rdt_cid: getParam('rdt_cid') }), // Reddit Ads
+      ...(getParam('ttclid') && { ttclid: getParam('ttclid') }), // TikTok Ads
+      ...(getParam('twclid') && { twclid: getParam('twclid') }), // X Ads (Twitter)
+      ...(getParam('li_fat_id') && { li_fat_id: getParam('li_fat_id') }), // LinkedIn Ads
+    }
+
+    return {
+      ...utmProps,
+      ...clickIdProps,
+      first_touch_url: url.href,
```
charislam, line 89: This should be sanitized of hash param to prevent accidentally logging credentials

seanoliver: Good catch and agreed! I pushed 8e0b86f137 to strip the URL fragment (`#...`) before we persist `page_url` (telemetry cookie) and before we derive `first_touch_*` props, so we don’t accidentally capture token-bearing fragments.

charislam (approved):

PR #41451, opened 2025-12-18
chore(www): clarify image paths

dnywh:

Frontmatter name change.

We repeatedly mistake `thumb` for `image` and visa versa, meaning the wrong images are used for Open Graph and in-site thumbnails on blog posts. Events and case studies use the same naming convention too.

## What is the new behavior?

These two bits of frontmatter are renamed for clarity:

  - Blog posts: `imgThumb` + `imgSocial`

That mapping for blog posts:

- `thumb` is now `imgThumb`
- `image` is now `imgSocial`

These related bits remain as-is:

  - Events
  - Case studies

The [www/README.md](https://github.com/supabase/supabase/blob/dnywh/chore/blog-image-frontmatter/apps/www/README.md#best-practices) file has been expanded to clarify all of the above. It now also provides instructions on image optimisation.

## To test

A lot of files were touched here. Please help make sure:

- [ ] The CMS works as intended. This is the **biggest unknown**.
- [x] All blog posts render the correct image as their on-site thumbnail and Open Graph image. You can test the latter by firing up a draft iMessage. Online Open Graph services like Facebook cache images, so aren’t reliable.
- [x] All events render their correct images
- [x] All case studies render their correct images
- [x] All customer stories render their correct images ([known issue](https://supabase.slack.com/archives/C072FL5KKKP/p1768888063209359?thread_ts=1768885681.502169&cid=C072FL5KKKP), predates this work)


apps/docs/public/img/supabase-auth-cover.png
dnywh, line 1: Duplicate of file from `www`. We seem to have a few of these. One for another time...

apps/www/app/api-v2/cms-posts/route.ts
dnywh, line 1: Please review this to make sure it works as intended!

apps/www/data/Developers.tsx
```diff
@@ -44,7 +44,7 @@ export const data = {
                 height="14.72"
                 rx="1.92"
                 stroke="currentColor"
-                stroke-width="1.28"
+                strokeWidth="1.28"
```
dnywh, line 46: Unrelated fix, sorry. Was just causing a Next error.

apps/www/public/images/blog/2025-12-17-building-chatgpt-apps-with-supabase/add-supabase-mcp-server.png
dnywh, line 1: This and other image swap-outs are unrelated, sorry. Just saw some images that needed downsizing.

apps/www/components/CustomerStories/index.tsx
dnywh, line 1: Unrelated change for capitalisation. Sorry.

fsansalvadore (approved):
All good, looking at the previews.
But could we use `imgThumb` instead of `imgSite` pls? 🙏

I think it would be a bit more descriptive and also recall the previous naming which would make it easier to find.

apps/www/app/api-v2/cms-posts/route.ts
stylessh: this looks alright, keep in mind that this is not being used at the moment and will be deprecated soon. so if build is not breaking i say it's good.

stylessh (approved):

PR #41466, opened 2025-12-18
fix: Clear the feedback text area after sending

ivasilov:
The text area in the feedback widget kept the same value after sending the feedback. The value was stored in local storage and loaded upon reopening. 

This PR clears the value from local storage when the feedback has been sent.

## Release Notes

* **Bug Fixes**
  * Feedback data is now properly cleared from local storage after successful submission.

* **Style**
  * Refined layout and spacing of the feedback submission confirmation message.


awaseem (approved):
Fix is good, just a weird issue with that screen I mentioned.

PR #41489, opened 2025-12-19
fix(docs)bad-links

smeubank:
These links are not routing correctly. It was also discussed that such a UI feature in docs should be discussed an implemented in way that could be scaled out more anyway.

YES

Bug fix docs

404

## What is the new behavior?

no 404, no links

## Additional context

## Summary by CodeRabbit

* **Documentation**
  * Removed trailing "Back to the Metrics API landing" navigation links from the Grafana Cloud, Grafana self‑hosted, and vendor‑agnostic metrics guides. This simplifies end-of-guide navigation and removes the duplicate back-link at the bottom of those pages.


ChrisChinchilla (approved):
I think they are missing the preceding "/docs"…

We badly need a link checker to catch all of these. Quite easy to do, so on it. But also I kind of agree that a scalable solution might be better. I am happy to keep these as now with correct links if you want to

smeubank:
> I think they are missing the preceding "/docs"…

yes it's an easy fix, but i was thinking as part of the O11y product docs work, i can do better to outline docs UI components that would be nice to build out. Nav buttons, cards, accordions, etc. It can be part of what we get into when you have time to revist the other PR

PR #41494, opened 2025-12-19
fix(studio): TSV file upload

marsou001:

YES

Bug fix

## What is the new behavior?

Resolves #41493 

- TSV files are fully supported now.
- Users can insert tabs inside "Paste text" section, or press Tab 2 times to focus on next element.
- Upload file error messaging includes TSV.

## Additional context

4 things I want to mention:

**First**: The issue can only be replicated on some machines due to the nature of the bug. The browser gets the mime type from the local machine's registry, which doesn't have a match for `text/tab-separated-values` in my case. The result is that `file.type` is an empty string. I checked it myself:

mime

This is confirmed by [this Stackoverflow thread](https://stackoverflow.com/questions/51724649/mime-type-of-file-returning-empty-in-javascript-on-some-machines). The conclusion is that `file.type` is not a safe mechanism to determine the whether the file is a true CSV or TSV file, extensions are more robust.

**Second**: when a table is empty and users drag a file before dropping it, a red text appears that says "Only CSV files are accepted" if the file is invalid. I tried to include the new context into account, but turns out browsers don't expose file information like file name and content before drop.

So, while the file is being drag, the only information we are allowed to obtain from it is it's mime type, which is, as I said earlier, not reliable enough. That's why I included the empty `file.type` string as a valid case among `UPLOAD_FILE_TYPES`

**Third**: I noticed that the textarea in "Paste text" section doesn't support tab insertion (tabs can be pasted though), so I fixed, and added a note saying that users should press tab 2 times to focus out of the textarea.

**Fourth**: empty files' upload are not handled. If an empty file has been uploaded, a toast popup with a generic error message appears, but no details are shown under the textarea. I just care of the issue by changing the toast message to "Couldn't import file because it is empty".

## Summary by CodeRabbit

* **New Features**
  * Accept TSV files in addition to CSV.
  * Auto-insert a Tab character when pressing Tab in the spreadsheet text input.

* **Bug Fixes**
  * Improved file validation for missing MIME metadata and zero-byte files.
  * Updated drag-and-drop and import messages to list CSV and TSV.

* **Improvements**
  * Converted spreadsheet input to a controlled field and clarified usage tips.

### DIFF
--- a/apps/studio/components/grid/components/grid/Grid.tsx
+++ b/apps/studio/components/grid/components/grid/Grid.tsx
@@ -180,7 +180,7 @@ export const Grid = memo(
                           isValidFileDraggedOver ? (
                             'Drop your CSV file here'
                           ) : (
-                            <span className="text-destructive">Only CSV files are accepted</span>
+                            <span className="text-destructive">Only CSV and TSV files are accepted</span>
                           )
                         ) : (
                           'This table is empty'
--- a/apps/studio/components/grid/components/grid/Grid.tsx
+++ b/apps/studio/components/grid/components/grid/Grid.tsx
@@ -180,7 +180,9 @@ export const Grid = memo(
                           isValidFileDraggedOver ? (
                             'Drop your CSV file here'
                           ) : (
-                            <span className="text-destructive">Only CSV and TSV files are accepted</span>
+                            <span className="text-destructive">
+                              Only CSV and TSV files are accepted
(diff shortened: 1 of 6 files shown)


7ttp (commented):
currently the team, wants to support CSV only
closing this up for now!...

> Thanks for ur efforts, towards contributing to Supabase !! 💚

PR #41503, opened 2025-12-19
fix: remove entity fetching as it is not needed anymore

awaseem:

YES

The original reason why this was because of the following: https://github.com/supabase/supabase/pull/16382

That is not the case anymore, the AI assistant does its own execute SQL and has context of any tables created. I tested the original instructions and this execute code is never called, we can remove this safely and increase performance

## Refactor
- Simplified the SQL Editor component by removing schema metadata fetching operations and streamlining data dependency handling. This refactoring reduces component initialization complexity and improves code clarity while maintaining all core SQL editing functionality.


charislam (approved):
Nice!

PR #41527, opened 2025-12-22
chore(studio): link directly to Postgres upgrade blockers

dnywh:

- UI improvement
- Resolves DEPR-104
- Follows up from DEPR-80
- References DEPR-183
- [x] Requires these [these backend changes to be merged](https://github.com/supabase/platform/pull/28528/)

Some users who could upgrade their project’s Postgres version can’t because they have things to fix first. Examples include:

- A) Deprecated extensions
- B) Functions that use those deprecated extensions

https://github.com/supabase/supabase/pull/38904 brought in direct links to A, so users could just go straight to the thing blocking upgrade and remove it. We couldn’t do this for B because the backend didn’t provide the necessary information, like function name. I did my best to give the user enough info to do it themselves in https://github.com/supabase/supabase/pull/39189.

@soedirgo has recently made the [backend changes](https://github.com/supabase/platform/pull/28528/) to expose the name of problematic functions and their extensions. This unblocks B from behaving like A.

## What is the new behavior?

In that same piece of work, @soedirgo exposed a bunch of possible validation errors. I’ve therefore expanded `ObjectsToBeDroppedWarning` to handle and communicate a myriad of issues, not just problematic functions.

Here’s the gist of what’s new:

- Everything in `validation_errors` from the backend is caught
- These errors are listed in an Admonition
- When appropriate, a _Manage_ button link is provided to the relevant area to fix this issue

| Before | After |
| --- | --- |
| Infrastructure  Supabase | Infrastructure  Supabase |
| Infrastructure  Supabase | Infrastructure  Supabase |

## Additional context

The [required backend changes](https://github.com/supabase/platform/pull/28528/) have already been merged to master.

## To test

`plv8` and `plcoffee` are two deprecated extensions you can use to test this work against. Make a simple “hello world” function for each to test the `function_using_obsolete_lang` warning.

Please ensure the other types of `validation_errors` are also up to scratch. You can find the full list [here](https://github.com/supabase/platform/pull/28528/changes#diff-53c5a34856baa0072b9f5fa2b91becc11f33bbef7f51b36873388c718180b44dR3061) and [here](https://github.com/supabase/supabase/blob/ea6ba04828594334aa7ea07436721f3e700a8a05/apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx).

## Summary by CodeRabbit

* **New Features**
  * Consolidated validation error list for infrastructure upgrades with richer error details and contextual “Manage” actions.

* **Bug Fixes**
  * More robust upgrade eligibility handling to prevent missing-data edge cases.

* **UI/UX**
  * Replaced legacy alert blocks with a unified admonition-style presentation and updated header casing for clarity.

* **Chores**
  * Added a deprecation note directing use of the new admonition pattern for notices.


dnywh:
@coderabbitai generate docstrings

soedirgo (approved):
Tested on local 👍

apps/studio/data/config/project-upgrade-eligibility-query.ts
```diff
@@ -7,16 +7,26 @@ import { useProjectDetailQuery } from 'data/projects/project-detail-query'
...
 import type { ResponseError, UseCustomQueryOptions } from 'types'
 import { configKeys } from './keys'
+import type { ValidationError } from './types'
 
 export type ProjectUpgradeTargetVersion = { postgres_version: string; release_channel: string }
 export type ProjectUpgradeEligibilityVariables = { projectRef?: string }
 export type ProjectUpgradeEligibilityResponse =
-  components['schemas']['ProjectUpgradeEligibilityResponse']
+  components['schemas']['ProjectUpgradeEligibilityResponse'] & {
+    validation_errors?: ValidationError[]
```
joshenlim: just to check - these validation_errors are coming from the API? if so - ideally these are directly from the generated API types then rather than defining them manually in the dashboard repo

joshenlim: pushed up the fixes for this!

joshenlim (approved):
Verified changes on staging preview - although just a feedback here, wondering if the [descriptions](https://github.com/supabase/supabase/blob/dnywh/chore/link-to-dropped-objects/apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx#L73) should be more actionable. cc @dnywh 

e.g here could mention "Remove extension before upgrade postgres", remove the guess work after clicking "Manage"
image

`function_using_obsolete_lang` as well - the description says `Function uses obsolete language: ${error.lang_name}`, but I'm not sure what I need to do? Do I need to delete that function? or change the language?

Other examples like `indexes_referencing_ll_to_earth` or `unlogged_table_with_persistent_sequence` - I'm not entirely sure what the required actions are before I can upgrade Postgres. (Although these might be out of scope from this PR since this PR is specifically for functions and extensions)

Approving nonetheless since this is already an improvement

soedirgo:
Hmm agree; for the others they need to either be dropped or be updated to not be offending (e.g. index needs to not reference `ll_to_earth()`)

joshenlim:
a bit more feedback RE the validation error + "manage" CTA (but we can address these separately really, don't wanna hold back the changes here)

- `active_replication_slot`: there's no manage link here, so I'm not sure how to "drop the active replication slot" from a user's POV
- `unsupported_fdw_handler`: Manage link could be the integrations page? Ideally could even detect which wrapper

joshenlim (approved):
Double approve!

apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
```diff
@@ -1,146 +1,178 @@
...
+        return `/project/${projectRef}/database/extensions?filter=${encode(error.extension_name)}`
+      case 'indexes_referencing_ll_to_earth':
+        return `/project/${projectRef}/database/indexes?schema=${encode(error.schema_name)}&search=${encode(error.index_name)}`
+      case 'unlogged_table_with_persistent_sequence':
+        return `/project/${projectRef}/database/tables?schema=${encode(error.schema_name)}&search=${encode(error.table_name)}`
+      case 'user_defined_objects_in_internal_schemas':
+        return error.obj_type === 'function'
+          ? `/project/${projectRef}/database/functions?schema=${encode(error.schema_name)}&search=${encode(error.obj_name)}`
+          : error.obj_type === 'table'
+            ? `/project/${projectRef}/database/tables?schema=${encode(error.schema_name)}&search=${encode(error.obj_name)}`
```
dnywh: These two `/database/tables/` links include URL query parameters that won't work until https://github.com/supabase/supabase/pull/42022 is merged.

It fails silently, which is why I’ve still included it in this PR.

apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
```diff
@@ -1,146 +1,178 @@
...
+            : null
+      case 'active_replication_slot':
+        return null
```
dnywh, line 98: Would be better to link to docs or a dialog with an explanation on how to fix this. I.e. updating the replication FAQ page to include a section on managing replication slots.

One for another day, though.

apps/studio/components/interfaces/Settings/Infrastructure/UpgradeWarnings.tsx
```diff
@@ -1,146 +1,178 @@
...
+      case 'unsupported_extension':
+        return `/project/${projectRef}/database/extensions?filter=${encode(error.extension_name)}`
+      case 'indexes_referencing_ll_to_earth':
+        return `/project/${projectRef}/database/indexes?schema=${encode(error.schema_name)}&search=${encode(error.index_name)}`
+      case 'unlogged_table_with_persistent_sequence':
+        return `/project/${projectRef}/database/tables?schema=${encode(error.schema_name)}&search=${encode(error.table_name)}`
+      case 'user_defined_objects_in_internal_schemas':
+        return error.obj_type === 'function'
+          ? `/project/${projectRef}/database/functions?schema=${encode(error.schema_name)}&search=${encode(error.obj_name)}`
+          : error.obj_type === 'table'
```
dnywh: Is this check redundant? I think the `obj_type` is only either a `function` or `table`?

dnywh: From the API types, `obj_type` is `'table' | 'function'`, so it can only be those two. The nested ternary covers both, and the final null is unreachable.

Since refactored to use conditional returns for readability.

joshenlim (approved):
Triple ✅ !

PR #41545, opened 2025-12-23
Filter pattern refactor

SaxonF:
image

Improvements made to the FilterBar component within ui-patterns. As part of this PR I've also removed the old test temporarily and will re-add a new test within the table editor in a follow up PR. This PR focuses entirely on ui-patterns and design-system so leaves Studio untouched.

## Summary by CodeRabbit

* **New Features**
  * Filter Bar now accepts custom actions (including AI-based actions) and an isLoading state
  * New advanced table header and a redesigned filter popover with AI-assisted filter generation and richer date options

* **Improvements**
  * Operator labels are human-friendly (e.g., "Equals")
  * Filter Bar refactored to a composable, context-driven structure for easier customization

* **Docs**
  * Documentation updated to an actions-based API and simplified prop tables

* **Tests**
  * Added comprehensive filter helpers unit tests


apps/studio/components/grid/components/header/filter/FilterPopoverNew.tsx
```diff
@@ -0,0 +1,277 @@
...
+  const { filters: urlFilters, onApplyFilters } = useTableFilter()
+  const snap = useTableEditorTableStateSnapshot()
+
+  // Local state for immediate UI updates
+  const [localFilters, setLocalFilters] = useState<Filter[]>(urlFilters)
+  const debouncedLocalFilters = useDebounce(localFilters, 500)
+
+  useEffect(() => {
+    onApplyFilters(debouncedLocalFilters)
+  }, [debouncedLocalFilters])
```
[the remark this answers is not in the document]
awaseem: Agreed with this, why is the function missing here?

apps/studio/components/grid/components/header/filter/FilterPopoverNew.tsx
```diff
@@ -0,0 +1,277 @@
...
+        }
+      }
+
+      // For other columns, keep a simple text-based filter
+      return {
+        label: column.name,
+        name: column.name,
+        type: 'string' as const,
+        operators: stringOperators,
+      }
```
awaseem: This map function is MASSIVE, it would be better separated out for the following:
- You can then write a small unit test to test all the branching statements
- Its way easier to read and understand how the transform works on a single value rather than an array

apps/studio/components/grid/components/header/filter/FilterPopoverNew.tsx
```diff
@@ -0,0 +1,277 @@
...
+            const hasValue = typeof (option as any)?.value === 'string'
+            const hasLabel = typeof (option as any)?.label === 'string'
+            if (hasValue && hasLabel) return (option as any).value || (option as any).label
+            if (hasLabel) return (option as any).label
+            return null
+          })
+          .filter((value): value is string => Boolean(value))
+      : undefined,
+  }))
+}
```
awaseem, line 116: All these methods can be moved to a .utl.ts file and be unit tested. We should refactor these methods out of this file and test the values

apps/studio/components/grid/components/header/HeaderNew.tsx
```diff
@@ -0,0 +1,504 @@
...
+                    />
+                  </div>
+                  <div>
+                    <p>Import data from CSV</p>
+                    <p className="text-foreground-light">Insert new rows from a CSV</p>
+                  </div>
+                </DropdownMenuItem>,
+              ]
+            : []),
+        ]}
```
awaseem, line 208: Confused by whats happening here? Why check the value to see if its undefined and then assign it to be the first index inside the array then spread the array? Just check if it exists and mount the component no?

apps/studio/components/grid/components/header/HeaderNew.tsx
```diff
@@ -0,0 +1,504 @@
+import { PermissionAction } from '@supabase/shared-types/out/constants'
```
awaseem, line 1: nit: There are multiple components in this file that can be spread out into separate components. Adding multiple react components in the same file makes it noisy and hard to read, so really recommend separating this

apps/studio/data/ai/sql-filter-mutation.ts
```diff
@@ -0,0 +1,62 @@
...
+  try {
+    body = await response.json()
+  } catch {}
+
+  if (!response.ok) {
+    throw new ResponseError(body?.error || body?.message, response.status)
+  }
+
+  return body as SqlFilterGenerateResponse
+}
```
awaseem: This isn't right, we shouldn't parse the body before checking the response type. This seems like a way to skip the undefined case where there is an exception thrown and the value of the body is incorrect. We should handle this gracefully

apps/studio/pages/api/ai/sql/filter-v1.ts
```diff
@@ -0,0 +1,226 @@
...
+
+        Rules:
+        - Use only the provided property names and operators for each property.
+        - Prefer logical operator "AND" unless the user explicitly asks for "OR".
+        - When unsure, default to simple equality comparisons with reasonable values.
+        - Values should respect property types: booleans must be true/false, dates should be ISO date strings (YYYY-MM-DD), and numbers must be numbers.
+        - If options are provided for a property, choose from those values when appropriate.
+
+        User request: "${prompt}"
+      `,
```
awaseem, line 79: nit: I really like the pattern of separating the prompt generation and even AI calls to its own .at.ts file. Here is an example https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/QueryPerformance/QueryPerformance.ai.ts#L29-L30

apps/studio/pages/api/ai/sql/filter-v1.ts
```diff
@@ -0,0 +1,226 @@
...
+    .map((operator) => {
+      if (typeof operator === 'string') return operator
+      if (operator?.value) return operator.value
+      if (operator?.label) return operator.label
+      return null
+    })
+    .filter((value): value is string => Boolean(value))
+
+  return serialized.length > 0 ? serialized : ['=']
+}
```
awaseem: nit: separate and refactor these into its own file and add tests for these. Its a good way to know that the logic will work and serialize the right operators

packages/ui-patterns/src/FilterBar/FilterBarContext.tsx
```diff
@@ -0,0 +1,309 @@
...
+  handleInputFocus: (path: number[]) => void
+  handleOperatorFocus: (path: number[]) => void
+  handleGroupFreeformFocus: (path: number[]) => void
+  handleInputBlur: () => void
+  handleGroupFreeformChange: (path: number[], value: string) => void
+  handleLabelClick: (path: number[]) => void
+  handleLogicalOperatorChange: (path: number[]) => void
+
+  // Options cache
+  propertyOptionsCache: Record<string, { options: any[]; searchValue: string }>
```
awaseem: Trying to avoid `any` here, I know this is coming from FilterCondition but it might be a good place to clean this up. Any just causes so many headaches for developers trying to debug this in the future and since we are modifying these files we should try to clean this up for others

packages/ui-patterns/src/FilterBar/FilterCondition.tsx
```diff
@@ -1,72 +1,73 @@
...
+    if (!isOperatorActive) {
+      setHasTypedOperator(false)
+    }
+  }, [isOperatorActive])
+
+  useEffect(() => {
+    if (!isActive) {
+      setHasTypedValue(false)
+    }
+  }, [isActive])
```
awaseem: These seem to be missing the functions as its deps

packages/ui-patterns/src/FilterBar/useCommandHandling.ts
```diff
@@ -121,6 +74,32 @@ export function useCommandHandling({
...
+      const updatedFilters = addFilterToGroup(activeFilters, currentPath, selectedProperty)
+      onFilterChange(updatedFilters)
+      const newPath = [...currentPath, group.conditions.length]
+
+      setTimeout(() => {
+        setActiveInput({ type: 'value', path: newPath })
+      }, 0)
+    },
+    [activeFilters, onFilterChange, setActiveInput]
+  )
```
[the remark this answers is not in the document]
awaseem: Agreed here, this should be changed as they are identical

awaseem (changes requested):
This needs a good clean up, just go through comments + coderrabbit as its found some good issues as well

apps/studio/components/grid/components/header/HeaderNew.tsx
SaxonF: @awaseem this is just logic taken from original Header (HeaderNew is a copy simply because we re-order a few things). Can we treat this as separate issue?

apps/studio/components/grid/components/header/HeaderNew.tsx
SaxonF: Similar to [this comment](https://github.com/supabase/supabase/pull/41545#discussion_r2644675800) could we separate into separate PR and solve both old and new at once?

apps/studio/data/ai/sql-filter-mutation.ts
SaxonF: @awaseem this follows same pattern as our other ai queries / mutations . I've pushed a change though let me if that is preferred and we can update others.

apps/studio/pages/api/ai/sql/filter-v1.ts
SaxonF: For this I think we should look at the new Agent abstraction in AI SDK https://vercel.com/blog/ai-sdk-6#agents

apps/studio/components/grid/components/header/filter/FilterPopoverNew.tsx
```diff
@@ -0,0 +1,199 @@
...
+  FilterProperty,
+  isGroup,
+  SerializableFilterProperty,
+  updateGroupAtPath,
+} from 'ui-patterns'
+import { columnToFilterProperty } from './FilterPopoverNew.utils'
+
+export interface FilterPopoverProps {
+  portal?: boolean
+}
```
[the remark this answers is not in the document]
awaseem: This seems to be true, we should remove the portal prop since its not used. I guess the whole type is no longer needed

apps/studio/components/grid/components/header/filter/FilterPopoverNew.tsx
```diff
@@ -0,0 +1,199 @@
...
+    options: Array.isArray(property.options)
+      ? property.options
+          .map((option) => {
+            if (typeof option === 'string') return option
+            const hasValue = typeof (option as any)?.value === 'string'
+            const hasLabel = typeof (option as any)?.label === 'string'
+            if (hasValue && hasLabel) return (option as any).value || (option as any).label
+            if (hasLabel) return (option as any).label
+            return null
+          })
```
awaseem: This can be cleaned up to now use any. Here is a simple example to make this easier:

```
type OptionLike =
  | string
  | { value?: unknown; label?: unknown }
  | null
  | undefined

const isObject = (v: unknown): v is Record<string, unknown> =>
  typeof v === "object" && v !== null

const toOptionText = (option: OptionLike): string | null => {
  if (typeof option === "string") return option
  if (!isObject(option)) return null

  const value = option.value
  const label = option.label

  if (typeof value === "string" && value) return value
  if (typeof label === "string" && label) return label
  return null
}

// usage
...options.map(toOptionText)
```

apps/studio/components/grid/components/header/filter/FilterPopoverNew.tsx
```diff
@@ -0,0 +1,199 @@
...
+            {
+              label: 'Pick a date...',
+              component: (props: CustomOptionProps) => <DatePickerOption {...props} />,
+            },
+          ],
+        }
+      }
+      return property
+    })
+  }, [snap.table?.columns])
```
awaseem: This is used in multiple places, should we use a memo colums value instead? 

```
const columns =snap.table?.columns ?? []
```

Can also memoize it

packages/ui-patterns/src/FilterBar/useCommandHandling.ts
```diff
@@ -135,46 +96,69 @@ export function useCommandHandling({
...
     },
-    [activeInput, filterProperties, activeFilters, onFilterChange, onFreeformTextChange]
+    [
+      activeInput,
+      filterProperties,
+      activeFilters,
+      onFreeformTextChange,
+      handlePropertySelection,
+      handlePropertySelection,
+    ]
```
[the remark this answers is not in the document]
awaseem: Agree with this

awaseem (approved):
Minor clean up required here + comments from code rabbit are good as well. Approving since a bulk of the changes are completed.

You'll def need to fix the usage of any to merge the PRs, we have new rules to stop the usage

PR #41572, opened 2025-12-24
fix(studio): tab focus on generate password button

dnywh:

- Bug fix

- The _Generate a password_ button (that looks like `InlineLink`) in the _New project_ form is not focusable by keyboard
- The other ones (`InlineLinks`) around it _are_ focusable by keyboard
- The _Generate a password_ button uses a `span` but a `button` would be more semantic and fix this issue

## What is the new behavior?

- The _Generate a password_ button now uses a `button` element, not a `span`
- Minor cleanup of redundant `className` elsewhere in form

## Additional context

No visual changes. Here’s the bit of UI in question:

| New Project |
| --- |
|  |

Spotted by @jahrlin.

## Summary by CodeRabbit

* **Improvements**
  * Updated links in the compute size selector to open in the same tab with standard styling instead of launching external windows for a more seamless navigation experience.
  * Enhanced the password strength bar with improved semantic structure for better accessibility when generating strong passwords.

### DIFF
--- a/apps/studio/components/interfaces/ProjectCreation/ComputeSizeSelector.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/ComputeSizeSelector.tsx
@@ -38,21 +38,11 @@ export const ComputeSizeSelector = ({ form }: ComputeSizeSelectorProps) => {
               <>
                 <p>
                   The size for your dedicated database. You can change this later. Learn more about{' '}
-                  <InlineLink
-                    target="_blank"
-                    rel="noopener noreferrer"
-                    className="text-inherit hover:text-foreground transition-colors"
-                    href={`${DOCS_URL}/guides/platform/compute-add-ons`}
-                  >
+                  <InlineLink href={`${DOCS_URL}/guides/platform/compute-add-ons`}>
                     compute add-ons
                   </InlineLink>{' '}
                   and{' '}
-                  <InlineLink
-                    target="_blank"
-                    rel="noopener noreferrer"
-                    className="text-inherit hover:text-foreground transition-colors"
-                    href={`${DOCS_URL}/guides/platform/manage-your-usage/compute`}
-                  >
+                  <InlineLink href={`${DOCS_URL}/guides/platform/manage-your-usage/compute`}>
                     compute billing
                   </InlineLink>
                   .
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/ProjectCreation/ComputeSizeSelector.tsx
```diff
@@ -38,21 +38,11 @@ export const ComputeSizeSelector = ({ form }: ComputeSizeSelectorProps) => {
               <>
                 <p>
                   The size for your dedicated database. You can change this later. Learn more about{' '}
-                  <InlineLink
-                    target="_blank"
-                    rel="noopener noreferrer"
-                    className="text-inherit hover:text-foreground transition-colors"
```
dnywh, line 44: The `target` and `rel` tags are all set on the component itself, and are therefore duplicative here.

_(I’m pretty sure it was me who added it in the first place...)_

apps/studio/components/ui/PasswordStrengthBar.tsx
```diff
@@ -1,4 +1,5 @@
 import { PASSWORD_STRENGTH_COLOR, PASSWORD_STRENGTH_PERCENTAGE } from 'lib/constants'
+import { InlineLinkClassName } from './InlineLink'
```
dnywh, line 2: Use the exact same styles as the inline links elsewhere in the form.

apps/studio/components/ui/PasswordStrengthBar.tsx
```diff
@@ -39,12 +40,9 @@ const PasswordStrengthBar = ({
...
           : 'This is the password to your Postgres database, so it must be strong and hard to guess.') +
           ' '}
-        <span
-          className="text-inherit underline hover:text-foreground transition-colors cursor-pointer"
-          onClick={generateStrongPassword}
-        >
+        <button type="button" className={InlineLinkClassName} onClick={generateStrongPassword}>
           Generate a password
-        </span>
+        </button>
```
dnywh, line 45: Using a `button` fixes the focus issue.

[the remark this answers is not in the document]
dnywh: About that reset:
```
bg-transparent border-none p-0 inline
```

That’s already the case, at least visually. Those styles do nothing.

About the `.inset-focus`: that’s for tables, hence `inset`. I’ll do a separate sweep for buttons.

ivasilov (approved):

PR #41634, opened 2025-12-29
fix Log Explorer run query inconsistency

jordienr:
- Fixes Logs Explorer so “Last hour/3 days/...” helpers always recompute on every run (button or Cmd+Enter) and don’t get frozen into static timestamps.
- Ensures keyboard shortcut and Run button use the same run logic.

- Made date-picker state explicit: page owns datePickerValue and passes it to LogsQueryPanel, keeping helper vs custom ranges in sync.
- Helper ranges stay dynamic: on run, we recompute timestamps; we clear URL timestamps and avoid persisting them when a helper is active.
- Custom ranges stay static: we persist their timestamps to URL state on selection/run.
- Both Run button and Cmd+Enter now share the same run handler (CodeEditor uses a ref to the latest callback), so behavior is consistent.
- Added pure date-range helpers and tests to ensure helper ranges recalc relative to “now” while custom ranges remain fixed.

## To test

- go to log explorer
- with Last hour selected → click run query a few times → always fetches latest results
- with last hour selected → press ⌘ + Enter a few times → always fetches latest results
- with a date range selected → Click run a few times → shows the same results all the time (does nothing basically)
- with a date range selected → press ⌘ Enter a few times → shows the same results all the time (does nothing basically)

## Summary by CodeRabbit

* **New Features**
  * Date picker now uses a single controlled value and supports helper-based ranges for log queries.

* **Refactor**
  * Centralized date-range resolution and reactive synchronization of date picker state.
  * Run Query action updated to always invoke the latest callback.
  * Selected-log handling in the logs table improved for more consistent selection behavior.

* **Tests**
  * Added/updated tests covering date-range resolution and query parameter construction.

### DIFF
--- a/apps/studio/components/interfaces/Settings/Logs/LogsQueryPanel.tsx
+++ b/apps/studio/components/interfaces/Settings/Logs/LogsQueryPanel.tsx
@@ -1,7 +1,6 @@
-import dayjs from 'dayjs'
 import { BookOpen, Check, ChevronDown, Copy, ExternalLink, X } from 'lucide-react'
 import Link from 'next/link'
-import { ReactNode, useState } from 'react'
+import { ReactNode, useEffect, useState } from 'react'
 
 import { IS_PLATFORM } from 'common'
 import Table from 'components/to-be-cleaned/Table'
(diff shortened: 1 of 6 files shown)


awaseem (approved):
tested in staging ✅

PR #41642, opened 2025-12-30
chore(studio): Use session access token for GraphiQL UI

joshenlim:
## Context

Realised that GraphiQL UI was broken with the new API keys, showing an error of `Expected 3 parts; got 1` on `POST /graphql` as we were using the new API keys for the `x-graphql-authorization` header

Also noticed that we're using the session access token for the `Authorization` header [here](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Integrations/GraphQL/GraphiQLTab.tsx#L24) and [here](https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Integrations/GraphQL/GraphiQLTab.tsx#L67) so opting to do the same for the `x-graphql-authorization` header - although I'm not entirely sure if this is advised (should both `Authorization` and `x-graphql-authorization` headers be using the new API keys instead?)

The changes here does make the GraphiQL UI work though

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified GraphQL authentication flow by streamlining authorization header construction.

* **Refactor**
  * Removed legacy API key-based authorization path; consolidated authentication to use a single token-based mechanism for improved reliability.


alaister (approved):
LGTM!

- Tested on preview

PR #41706, opened 2026-01-05
fix(studio): convert tableCreateGeneratePolicies flag to multivariate for PostHog experimentation

seanoliver:
## Problem

Turns out PostHog only supports running experiments on multivariate feature flags, not boolean ones. We originally implemented `tableCreateGeneratePolicies` as a simple true/false flag, which means we can't actually get proper experiment results from it.

This is blocking us from getting data we need to make business decisions around the generate policies feature, so we need to fix this ASAP.

## Changes

Updated the flag from boolean to multivariate (`"control"` / `"variation"`):

- **useTableCreateGeneratePolicies.ts** - Changed flag type to string, added a `isValidExperimentVariant()` type guard, updated logic to check for `'variation'` instead of truthy
- **SidePanelEditor.tsx** - Updated conversion event to use the type guard
- **telemetry-constants.ts** - Updated variant types from `'control' | 'treatment'` to `'control' | 'variation'`

Also made the exposure/conversion events only fire when the flag returns a valid experiment value (`"control"` or `"variation"`). If PostHog returns undefined/null/false for whatever reason, we skip tracking since those users aren't actually in the experiment.

## Testing

- Verified TypeScript compiles without errors
- The flag has already been updated in PostHog to return the new multivariate values

---

Fixes GROWTH-580

## Summary by CodeRabbit

* **Refactor**
  * Overhauled experiment variant handling and validation to use explicit variant strings.
  * Changed the feature flag type from boolean to string for clearer semantics.
  * Exposure and conversion gating now validate variant values before tracking.
  * Telemetry reports the raw variant label and updated variant naming (treatment → variation) for experiment events.

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
@@ -26,6 +26,7 @@ import { RetrieveTableResult } from 'data/tables/table-retrieve-query'
 import { getTables } from 'data/tables/tables-query'
 import { useSelectedOrganizationQuery } from 'hooks/misc/useSelectedOrganization'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
+import { isValidExperimentVariant } from 'hooks/misc/useTableCreateGeneratePolicies'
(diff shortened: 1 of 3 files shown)


pamelachia (approved):

alaister (approved):
LGTM!

PR #41730, opened 2026-01-06
chore(studio): Remove coming soon notices for analytics and vector buckets + add callout for region availability for vector buckets

joshenlim:
## Context

As per PR title - Analytics and Vector buckets are automatically available for users once the project is on a paid plan so there's no longer a need to:
- Show a coming soon UI, just the upgrade UI will do
- Manually check the subscription, just relying on the storage config will do

Also added a callout for the region availability for vector buckets:
image

## Summary by CodeRabbit

* **New Features**
  * Vector buckets now use region-based availability with an enhanced empty state listing supported regions and guidance.
  * Analytics storage now directs users to upgrade/plan options instead of a coming‑soon message.
  * General storage empty states improved for clearer availability status and next-step guidance.

* **Refactor**
  * Internal layout and gating logic simplified for more consistent user flows.


jordienr (approved):
looking gud ✅ 

agree with the coderabbit nitpicks but not necessary imo

PR #41783, opened 2026-01-08
chore(studio): show oauth server endpoints in oauth server settings

fsansalvadore:
Show OAuth server endpoints in oauth server settings page.

Preview: [OAuth Server settings](https://studio-staging-git-chore-show-oauth-server-endpoints-supabase.vercel.app/dashboard/project/_/auth/oauth-server)

Screenshot 2026-01-09 at 12 00 31

## Summary by CodeRabbit

* **New Features**
  * Added an OAuth endpoints table showing Authorization, Token, JWKS, and Discovery/OpenID URLs with copy-to-clipboard and a masked preview mode.
  * Inline preview of the Authorization URL when an authorization path is set.

* **Improvements**
  * Reorganized OAuth server settings for clearer enable/disable flow, conditional field visibility, and disable confirmation.
  * Dynamic loading of the endpoints table, improved loading skeletons, layout refinements, and form reset to reflect saved defaults.


apps/studio/components/interfaces/Auth/OAuthApps/OAuthEndpointsTable.tsx
joshenlim, line 1: hmmm this one's more of a feedback but wondering if we could use `FormItemLayout` and `Input`  here for consistency? so you wouldn't need to create a `CopyableCell` component too.

Will look something like this:
```
<Card>
  <CardContent className="flex flex-col gap-y-4">
    {endpoints.map((endpoint) => {
      const url = `${baseUrl}${endpoint.path}`
      return (
        <FormItemLayout layout="horizontal" isReactForm={false} label={endpoint.name}>
          <Input readOnly copy value={url} />
        </FormItemLayout>
      )
     })}
  </CardContent>
</Card>
```

image

apps/studio/pages/project/[ref]/auth/oauth-server.tsx
Comment (joshenlim) [inline apps/studio/pages/project/[ref]/auth/oauth-server.tsx:1]: (unrelated feedback)

just wondering if we wanna remove "Learn more" from the description here - given that we've got the docs button up top and they both link to the same docs page. so the description wouldn't need to wrap
image

apps/studio/components/interfaces/Auth/OAuthApps/OAuthEndpointsTable.tsx
fsansalvadore: Makes sense, update it to use the FormItemLayout + Input components 👍

apps/studio/pages/project/[ref]/auth/oauth-server.tsx
Comment (fsansalvadore) [reply apps/studio/pages/project/[ref]/auth/oauth-server.tsx:1]: Yep that link is indeed redundant, removed

apps/studio/components/interfaces/Auth/OAuthApps/OAuthEndpointsTable.tsx
```diff
@@ -0,0 +1,71 @@
...
+    },
+    {
+      name: 'Discovery endpoint',
+      path: '/.well-known/oauth-authorization-server/auth/v1',
+    },
+    {
+      name: 'OIDC discovery',
+      path: '/auth/v1/.well-known/openid-configuration',
+    },
+  ]
```
cemalkilic, line 54: Instead of hardcoding the endpoints here, we can make a call to `<project_url>/auth/v1/.well-known/openid-configuration` and parse these values from the response.

Also no need to show the `Discovery endpoint` (`/.well-known/oauth-authorization-server/auth/v1`)

cemalkilic (commented):
The endpoint list is being visible immediately after I toggled the "Enable OAuth Server" in the `dashboard/project/_/auth/oauth-server` page.

IMHO it's better to list it after user clickes "Save changes"

apps/studio/components/interfaces/Auth/OAuthApps/OAuthEndpointsTable.tsx
fsansalvadore: ok, made and pushed the changes @cemalkilic. 

So there are a few loading states. When the value is toggle but not yet saved, the endpoints table is shown with blurred data, skeletons show when values are loading, and the endpoints are slightly dimmed if the user switches off the oauth server toggle but has not saved yet.

To me it feels like good visual feedback of what's going on during each step now.

Wdyt?

https://github.com/user-attachments/assets/06c02c1a-394e-471a-b14e-8b846dcf4e77

fsansalvadore: @cemalkilic @joshenlim is this good to go?

cemalkilic: hey @fsansalvadore , pardon me missed the notification earlier.

I think looks good in general, maybe one thing to improve: when the oauth server toggle is disabled, we can remove the "OAuth Endpoints" section completely (before clicking "Save")? wdyt?

https://github.com/user-attachments/assets/7a031784-dd42-4dbf-a433-3091a9e01d42

awaseem:
Is this even needed still @fsansalvadore? my guess is not so I'm moving to draft for now

apps/studio/components/interfaces/Auth/OAuthApps/OAuthEndpointsTable.tsx
fsansalvadore: @cemalkilic ok hiding the endpoints when disabling the server and not yet save. But endpoints will only show once the form is saved, when toggling from `false` to `true`.

awaseem (approved):
LGTM

PR #41785, opened 2026-01-08
Remove check for sourceId for rendering add destination button

joshenlim:
## Context

The "Add destination" button should no longer require a check for `sourceId` if `unifiedReplication` is on, since `sourceId` is ETL specific, while destinations will now include Read Replicas

## Summary by CodeRabbit

* **Improvements**
  * The "Add destination" button in Database Replication now appears when unified replication is active as well as under previous conditions (e.g., when a source is present). This makes it easier to configure replication destinations across more scenarios and streamlines the replication setup experience.

### DIFF
--- a/apps/studio/components/interfaces/Database/Replication/Destinations.tsx
+++ b/apps/studio/components/interfaces/Database/Replication/Destinations.tsx
@@ -190,7 +190,7 @@ export const Destinations = () => {
             />
           </div>
           <div className="flex items-center gap-x-2">
-            {(unifiedReplication || (!unifiedReplication && !!sourceId)) && (
+            {(unifiedReplication || !!sourceId) && (
               <Button
                 type="default"
                 icon={<Plus />}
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Database/Replication/Destinations.tsx
```diff
@@ -191,7 +190,7 @@ export const Destinations = () => {
             />
           </div>
           <div className="flex items-center gap-x-2">
-            {!!sourceId && (
+            {(unifiedReplication || (!unifiedReplication && !!sourceId)) && (
```
alaister: nit simplify:
```suggestion
            {(unifiedReplication || !!sourceId) && (
```

truth table:
unifiedReplication | sourceId | Original | Simplified
-- | -- | -- | --
false | falsy | false | false
false | truthy | true | true
true | falsy | true | true
true | truthy | true | true

alaister (approved):
LGTM!

one small one:

apps/studio/components/interfaces/Database/Replication/Destinations.tsx
joshenlim: oh nice! hahaha thanks for this 🙏 🙏

PR #41808, opened 2026-01-09
Use new action runs endpoints

joshenlim:
## Context

This is only for Gitless branching - merge attempts get stuck at "Creating project" in the logs. This is due to the wrong endpoint being used as we now need to use `v1` endpoints as opposed to the `platform` endpoints.

Changes in this PR fixes that such that the workflow logs populate correctly.

## Other changes involved
- Remove `workflow-runs` react query -> Branching now uses the `actions` react query
  - Also refactored the queries in `actions` to be more consistent with how we write our RQs
- (Unrelated) Unify usage of `DiffEditor` components to import from `ui` instead of `monaco` directly
  - `DiffEditor` in `ui` consolidates the options, rather than declaring the options in separate places whenever we render `DiffEditor`
  - Affects SQL Editor  + AI Editor (Used in edge functions)
  - Also fixes the `TextModel got disposed before DiffEditorWidget model got reset` when switching between "Database" and "Edge Functions" in the branch merge page

## To test
- [ ] Try merging a gitless branch - logs should populate nicely, and not get stuck at "Creating project" (refer to demo video in [Linear](https://linear.app/supabase/issue/FE-2348/gitless-branch-merges-stuck-in-creating-project-status) for reference)

## Summary by CodeRabbit

* **Refactor**
  * Clearer workflow status states and updated success/failure/running labels and icons for more consistent UI feedback.
  * Consolidated workflow run and logs fetching for more reliable and accurate run status and log display.
  * Diff views standardized to a consistent, read-only diff editor for SQL and function diffs.

* **Chores**
  * Components updated to named exports for module consistency.


apps/studio/data/workflow-runs/workflow-runs-query.ts
joshenlim, line 1: no longer used

apps/studio/components/interfaces/BranchManagement/WorkflowLogsCard.tsx
```diff
@@ -1,19 +1,20 @@
...
+// interface WorkflowRun {
+//   id: string
+//   status: string
+//   branch_id?: string
+//   check_run_id?: number | null
+//   created_at?: string
+//   updated_at?: string
+//   workdir?: string | null
+//   git_config?: unknown
+// }
```
joshenlim: should be obtained from API generated types, now exported in workflow-run-query

awaseem (changes requested):
Put some comments in the slack thread, Just need to test to make sure this is working correctly

awaseem (approved):
LGTM

PR #41747, opened 2026-01-06
fix: added portal prop for weird popover placement

awaseem:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

Bug fix: Portal for download was WAYYY OUT THERE. Now its in the right place

## Summary by CodeRabbit

* **Refactor**
  * Enhanced popover rendering behavior in the Edge Functions details section. Optimizations to the rendering mechanism improve visual presentation, positioning accuracy, and overall display consistency. Layout properties including alignment and padding configuration remain unchanged, ensuring the interface maintains visual stability and consistency while benefiting from the improved rendering approach.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #41818, opened 2026-01-09
Popover ui component to use portal by default, remove all manual declaration of portal for PopoverContent in dashboard

joshenlim:
## Context

Noticed that `PopoverContent` is the only UI component that allows dynamic usage of `portal` - this has historically caused quite a bit of confusion + sporadic bugs where the PopoverContent is misaligned like [here](https://github.com/supabase/supabase/pull/41747). (And 99% (if not all) of the time the fix was just to add a `portal` param to `PopoverContent`)

Ideally the portal behaviour is consistent across our UI components so that it's easier for us to understand similar positioning issues (if any pop up in the future), so the changes here are mainly to have `PopoverContent` use portal by default.

## Changes involved
- `PopoverContent` to use `Portal` by default
- Remove all manual passing of `portal` parameter into `PopoverContent` from the dashboard

## To test
Have manually  checked for both Chrome and Safari for all UI areas that are affected and all actually looks good, but it'll be good to get a second pair of eyes to make sure that I didn't overlook any parts 🙏 

It's quite a broad area to test though, but to simplify the testing, it'll be great to test
- [x] On both chrome and safari
- [x] PopoverContent when rendered in the dashboard on its own (e.g Table Editor filter + sorts, etc)
- [x] PopoverContent when rendered in a sheet (e.g Table side panel editor, Foreign key selector panel filter + sorts, etc))
- [x] PopoverContent when rendered in a Dialog (e.g Move query Modal, Connect Modal, etc)

## Summary by CodeRabbit

* **Refactor**
  * Unified popover rendering across the app to a single, consistent behavior (removed per-instance portal control).
  * Simplified popover APIs and reduced portal-related options.

* **Style**
  * Minor UI tweaks: smaller input text in some selectors and adjusted popover alignment for improved visual consistency.

* **Breaking**
  * A few selector components now expose updated props/exports — update integrations if you relied on removed portal props or default exports.

### DIFF
--- a/apps/studio/components/interfaces/Functions/EdgeFunctionDetails/EdgeFunctionTesterSheet.tsx
+++ b/apps/studio/components/interfaces/Functions/EdgeFunctionDetails/EdgeFunctionTesterSheet.tsx
@@ -428,7 +428,7 @@ export const EdgeFunctionTesterSheet = ({ visible, onClose }: EdgeFunctionTester
                 <RoleImpersonationStateContextProvider
                   key={`role-impersonation-state-${projectRef}`}
                 >
-                  <RoleImpersonationPopover portal={false} disallowAuthenticatedOption={true} />
+                  <RoleImpersonationPopover disallowAuthenticatedOption />
                 </RoleImpersonationStateContextProvider>
                 <Button
                   type="primary"
--- a/apps/studio/components/grid/components/editor/DateTimeEditor.tsx
+++ b/apps/studio/components/grid/components/editor/DateTimeEditor.tsx
@@ -76,7 +76,7 @@ function BaseEditor<TRow, TSummaryRow = unknown>({
           {value === null ? 'NULL' : value}
         </div>
       </PopoverTrigger_Shadcn_>
-      <PopoverContent_Shadcn_ portal align="start" className="p-0 rounded-none w-64">
+      <PopoverContent_Shadcn_ align="start" className="p-0 rounded-none w-64">
         <BlockKeys
           ignoreOutsideClicks
           value={inputValue}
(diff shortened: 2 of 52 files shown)


awaseem (approved):
LGTM I tested all the previous broken popovers and they seem to be working. Its a way cleaner unlock as well

apps/studio/components/interfaces/Functions/EdgeFunctionDetails/EdgeFunctionTesterSheet.tsx
```diff
@@ -428,7 +428,7 @@ export const EdgeFunctionTesterSheet = ({ visible, onClose }: EdgeFunctionTester
                 <RoleImpersonationStateContextProvider
                   key={`role-impersonation-state-${projectRef}`}
                 >
-                  <RoleImpersonationPopover portal={false} disallowAuthenticatedOption={true} />
+                  <RoleImpersonationPopover disallowAuthenticatedOption />
```
dnywh, line 431: Noticed this was super wide during testing. Unrelated to this PR but the fix would be something like:
```
<PopoverContent_Shadcn_ className="p-0 w-[420px] overflow-hidden" side="bottom" align={align}>
```

...in [`RoleImpersonationPopover`](https://github.com/supabase/supabase/blob/5808dc5a6568c7524b4671d1e775ca5b530361ce/apps/studio/components/interfaces/RoleImpersonationSelector/RoleImpersonationPopover.tsx#L72). That whole component could be redone, though!

dnywh: Never mind, this seems to be used also on Table Editor with 3 roles, so needs the width. `min-w` doesn’t do the trick.

dnywh (approved):
Tested about half of these instances locally. Nothing out of place.

apps/studio/components/interfaces/Functions/EdgeFunctionDetails/EdgeFunctionTesterSheet.tsx
joshenlim: i think you might be looking for `w-min` which actually does the trick hahaha - will add that fix in! 🙂 🙏 great catch!

joshenlim: image

PR #41827, opened 2026-01-10
Studio: Enable PostHog toolbar via CSP frame-src

seanoliver:

YES

Bug fix

The PostHog toolbar works fine on www and docs, but in Studio it's blocked by CSP. When you try to open it, DevTools shows:

> Refused to frame 'https://ph.supabase.com/...' because it violates the following Content Security Policy directive: "frame-src..."

This happens because Studio has its own strict CSP configuration in `apps/studio/csp.js`. We already allow PostHog scripts to run (it's in `SCRIPT_SRC_URLS`), but the toolbar also needs frame embedding permission to load its UI.

## What is the new behavior?

Added `POSTHOG_URL` to the `FRAME_SRC_URLS` array in Studio's CSP config, but only for dev/staging environments. Now the toolbar loads properly in Studio during local development and on staging, just like it does in www/docs.

The toolbar is restricted to non-production environments since it's primarily used for:
- Debugging events and payloads while navigating Studio during development
- Testing PII obfuscation and catching regressions in staging
- Verifying feature flags and experiment instrumentation during QA
- Correlating sessions during development/testing

Production environments can still collect PostHog analytics (script execution works fine) - they just won't have the interactive toolbar overlay. If needed for production debugging, the PostHog web UI provides full event inspection capabilities.

## Additional context

Change in `apps/studio/csp.js` conditionally adds PostHog to frame sources using the existing `isDevOrStaging` check (which includes local, preview, and staging environments). This keeps the production CSP tighter while supporting the primary developer workflow.

GROWTH-576

## Summary by CodeRabbit

* **Chores**
  * Updated Content Security Policy configuration to include support for additional tracking services in development and staging environments.

### DIFF
--- a/apps/studio/csp.js
+++ b/apps/studio/csp.js
@@ -113,7 +113,12 @@ module.exports.getCSP = function getCSP() {
     STAPE_URL,
     POSTHOG_URL,
   ]
-  const FRAME_SRC_URLS = [HCAPTCHA_ASSET_URL, STRIPE_JS_URL, STAPE_URL]
+  const FRAME_SRC_URLS = [
+    HCAPTCHA_ASSET_URL,
+    STRIPE_JS_URL,
+    STAPE_URL,
+    ...(isDevOrStaging ? [POSTHOG_URL] : []),
+  ]
   const IMG_SRC_URLS = [
     SUPABASE_URL,
     SUPABASE_COM_URL,


charislam:
@seanoliver do you need the PH toolbar in prod as well, or is it possible to rely on it only in staging?

seanoliver:
> @seanoliver do you need the PH toolbar in prod as well, or is it possible to rely on it only in staging?

Great question. You're right we shouldn't need the PH toolbar in prod. Let me adjust this so that it's only included in staging / localhost.

staaldraad (approved):

PR #41833, opened 2026-01-10
fix(studio): add postgres SSL support for self-hosted instance

chloyka:
Yes

Fix

When self-hosting Supabase Studio with a PostgreSQL database that requires SSL connections (e.g., `hostnossl all all all reject` in pg_hba.conf), the connection fails with:
pg_hba.conf rejects connection for host "x.x.x.x", user "postgres", database "postgres", no encryption

**This happens because:**
1. The `getConnectionString()` function in `lib/api/self-hosted/util.ts` doesn't include `sslmode` parameter
2. There's no way to configure SSL mode via environment variables

## What is the new behavior?
- Added `POSTGRES_SSL` env var to configure `sslmode` parameter in connection strings
- When `POSTGRES_SSL=require` is set, connection strings include `?sslmode=require`
- Also documented existing but undocumented env vars: `POSTGRES_USER_READ_WRITE`, `POSTGRES_USER_READ_ONLY`

## New Environment Variables
| Variable | Default | Description |
|----------|---------|-------------|
| `POSTGRES_SSL` | (empty) | PostgreSQL sslmode (disable, allow, prefer, require, verify-ca, verify-full) |

## Example Usage
```bash
# .env
POSTGRES_HOST=db
POSTGRES_PORT=5432
POSTGRES_DB=postgres
POSTGRES_PASSWORD=secret
POSTGRES_SSL=require
POSTGRES_USER_READ_WRITE=supabase_admin
```

## Additional context
This is especially important for:
- Kubernetes deployments with Zalando PostgreSQL Operator (enforces SSL)
- Any self-hosted setup with SSL-only PostgreSQL

## Summary by CodeRabbit

* **New Features**
  * Added support for self-hosted PostgreSQL connection configuration, including settings for host, port, database name, and SSL mode alongside dedicated read-write and read-only user credentials.

* **Documentation**
  * Updated documentation with reference guide for self-hosted PostgreSQL environment configuration variables and their defaults.


chloyka:
up

PR #41837, opened 2026-01-11
Fix crawler smoke test for function pages

p0sadas:

YES

Bug fix (test stability)

The crawler smoke test for function pages fails even though the production
documentation is correct.

The test expects the description to be the text immediately after the `<h1>`,
but the current docs layout renders the "Parameters" heading first, which causes
the test to receive `"Parameters"` instead of the actual description.

This makes CI fail with a false negative.

## What is the new behavior?

The test now validates semantic content instead of relying on brittle DOM
structure.

It checks for the presence of important sections such as:
- `parameters`
- `examples`

This keeps the crawler validation meaningful while making it resilient to
layout changes in the docs.

## Additional context

This change aligns the smoke test with how users actually consume the docs,
and prevents CI from failing due to harmless UI or layout updates.

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability for documentation pages by enhancing heading verification logic with more flexible matching patterns.

### DIFF
--- a/apps/docs/app/api/crawlers/route.smoke.test.ts
+++ b/apps/docs/app/api/crawlers/route.smoke.test.ts
@@ -53,21 +53,18 @@ describe('prod smoke test: crawler pages return correct data', () => {
     const text = await result.text()
 
     const $ = load(text)
+
     const h1 = $('h1').text()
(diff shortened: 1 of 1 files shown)


Hallidayo:

Issue #41841, opened 2026-01-11
storage.download() returns stale/cached data after file updates


kanishkjain-superbuilders:
When using supabase.storage.from('bucket').download(path) in Supabase Edge Functions, updated
   files are not reflected immediately. The method appears to have HTTP-level caching that     
  serves stale data.                                                                           
                                                                                               
  Reproduction:                                                                                
  1. Upload a JSON file to storage                                                             
  2. Call download() to fetch it                                                               
  3. Update the file in storage (via dashboard or API)                                         
  4. Call download() again → still returns old content                                         
                                                                                               
  Workaround:                                                                                  
  Using createSignedUrl() + manual fetch with cache: 'no-store' returns fresh data:            
                                                                                               
  const { data: urlData } = await supabase.storage                                             
    .from('bucket')                                                                            
    .createSignedUrl(path, 60);                                                                
                                                                                               
  const response = await fetch(urlData.signedUrl, { cache: 'no-store' });                      
  const blob = await response.blob();                                                          
                                                                                               
  Expected behavior:                                                                           
  download() should return the current file contents, not cached data.                         
                                                                                               
  Environment:                                                                                 
  - Supabase Edge Functions (Deno)                                                             
  - @supabase/supabase-js v2

7ttp:
Fixed in https://github.com/supabase/supabase-js/pull/2090
Closing this up!

PR #41855, opened 2026-01-12
fix storage.download() returns stale/cached data after file updates

Adityakk9031:
Issue
storage.download() returns stale data because it uses the default fetch behavior which may cache responses, and the method currently provides no way to pass Cache-Control headers or options to opt-out of this caching (e.g. no-store). This is particularly problematic in environments like Supabase Edge Functions (Deno).

Solution
Updated the documentation in 
apps/docs/content/guides/storage/quickstart.mdx
 to provide a verified workaround. The guide now includes an example of using createSignedUrl combined with fetch(..., { cache: 'no-store' }) to ensure fresh data retrieval.

Includes inline comments to clarify URL validity and blob usage, incorporating review feedback for better developer experience.

closes:#41841

## Summary by CodeRabbit

* **Documentation**
  * Updated Storage Quickstart (JavaScript) download section with an admonition about cached/stale downloads. Adds guidance on obtaining the freshest file by using signed download links and fetching with cache control to bypass caches, and includes a code example demonstrating the approach.

### DIFF
--- a/apps/docs/content/guides/storage/quickstart.mdx
+++ b/apps/docs/content/guides/storage/quickstart.mdx
@@ -204,11 +204,12 @@ To ensure you get the latest version, create a signed URL and fetch it with `cac
 ```js
 const { data } = await supabase.storage
   .from('avatars')
-  .createSignedUrl('public/avatar1.png', 60)
+  .createSignedUrl('public/avatar1.png', 60) // URL valid for 60 seconds
 
 if (data) {
   const response = await fetch(data.signedUrl, { cache: 'no-store' })
   const blob = await response.blob()
+  // Use blob as needed (e.g., create object URL, read as text, etc.)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mandarini:
Hi @Adityakk9031! 

Thank you for documenting a workaround for this issue. However, we're going to close this PR because we're fixing the root cause in the SDK instead.

## Why We're Not Merging This

While your workaround technically works, it has some issues:

1. **Misuse of `createSignedUrl()`** - This method is for private buckets, but your example path suggests a public file
2. **Bypasses SDK features** - Manual `fetch()` loses error handling, authentication, and RLS enforcement
3. **Missing context** - The issue is specific to Edge Functions (Deno runtime), not universal
4. **Overly complex** - Requires two API calls instead of one
5. **Wrong layer** - This is an SDK limitation, not something users should work around

## The Proper Fix

We're fixing this in the SDK itself by exposing fetch parameters in the `download()` method: https://github.com/supabase/supabase-js/pull/2090

Once released, the clean solution will be:

```js
// Edge Functions - bypass cache
const { data, error } = await supabase.storage
  .from('avatars')
  .download('file.png', {}, { cache: 'no-store' })

// Default (browsers/Node.js)
const { data, error } = await supabase.storage
  .from('avatars')
  .download('file.png')
```

And I'll make sure to document this.

Thank you very much for contributing to Supabase! Your contributions is what makes our tools better 💚

PR #41871, opened 2026-01-13
fix(studio): function form reset only after open or change

dnywh:

- Fix
- Resolves DEPR-323

1. Navigate to [Database Functions](http://localhost:8082/project/_/database/functions)
2. Duplicate an existing function, which should open a sheet
3. Add 1+ new arguments, leave the sheet open
4. Blur focus (open a new tab, different window, wait a minute)
5. Come back and notice how the new, unsaved, arguments disappear

This bug exists because the `useEffect` depends on the entire `func` object, which changes often and therefore resets the form.

## What is the new behavior?

- Changed the dependency array from `[visible, func]` to `[visible, func?.id]` to track the function ID instead of the object reference
- Only resets the form when:
	- The panel opens (first time), or
	- The function ID changes (switching to a different function)

## To test

Please run through steps 1–5 on both `master` (or prod) vs this branch to see the difference.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form state management in the database function creation panel: the form now reliably resets when the panel opens/closes and correctly repopulates when switching between different functions, preventing stale or incorrect values from appearing.

### DIFF
--- a/apps/studio/components/interfaces/Database/Functions/CreateFunction/index.tsx
+++ b/apps/studio/components/interfaces/Database/Functions/CreateFunction/index.tsx
@@ -139,17 +139,7 @@ export const CreateFunction = ({
   }
 
   useEffect(() => {
-    if (!visible) {
-      // Reset tracking when panel closes
-      lastInitializedFuncIdRef.current = undefined
-      return
-    }
-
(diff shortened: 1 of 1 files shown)


joshenlim (approved):

PR #41898, opened 2026-01-14
fix(studio): remove wrongly semicolon from export cli command

kemaldotearth:

YES

This should fix the export CLI command in the Table Editor, removing the falsely placed semicolin inside the parenthesis. Works as expected now generating a `.csv` file with `psql`. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Exported SQL queries now have trailing semicolons removed so CSV/COPY exports use clean, consistent query strings.

* **Style**
  * Code block string color adjusted to a theme-aware value for improved readability while maintaining dark-mode appearance.


NARUDESIGNS:
@kemaldotearth looks good, thanks! 🙌🏼

apps/studio/components/grid/components/header/ExportDialog.tsx
```diff
@@ -75,6 +75,8 @@ export const ExportDialog = ({
         )
     : ''
 
+  const query = queryWithSemicolon.replace(/;\s*$/, '')
```
NARUDESIGNS, line 78: Yea, the query looks good now, thanks again.
I'm curious tho, about how the ; was added to the query in the first place. I see you're simply replacing it with an empty string here which works, anyway.

kemaldotearth: Added to guarantee it gets stripped out. There's something like `isFinal` prop on the `toSql()` function which adds the semicolon. This ensures we don't add it, but there might be a better way of doing it 🤔

saltcod (approved):

packages/ui/src/components/CodeBlock/CodeBlock.utils.ts
```diff
@@ -45,7 +45,7 @@ export const monokaiCustomTheme = (isDarkMode: boolean) => {
       color: '#bf79db',
     },
     'hljs-string': {
-      color: '#3ECF8E',
+      color: isDarkMode ? '#3ECF8E' : '#0B6C50',
```
dnywh: Could reuse `text-brand-link`? I.e:


```suggestion
      color: 'hsl(var(--brand-link))',
```

dnywh: It also makes the green stand out a wee bit better on light mode. Hopefully still enough contrast against the white background

dnywh (approved):
Approved with minor suggestion 🦟

PR #41922, opened 2026-01-15
normalize cpu usage chart FE-2354

jordienr:
Normalize CPU usage chart when its above 100% due to multiple cores. 

## Before
CleanShot 2026-01-15 at 16 43 03@2x

## After
CleanShot 2026-01-15 at 16 43 59@2x

## To test:
- Using something like the tweak chrome extension, change the return payload of the cpu-usage request and change some values so its above 100%, reload the page and you should see the data normalized to a 0-100% range

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CPU usage metrics display in charts with proper data normalization to ensure metrics scale correctly and remain within expected ranges.

* **Tests**
  * Added test coverage for CPU usage data normalization behavior.


awaseem (approved):
Tested on staging

PR #41984, opened 2026-01-19
chore(studio): remove homeNew feature flag helper

pamelachia:
## Summary

- Deletes the `apps/studio/lib/featureFlags/` folder and `homeNew.ts` helper
- Inlines the flag check as `usePHFlag('homeNew') === 'new-home'` at 3 call sites

## Context

The `isHomeNewVariant()` helper was only useful during migration from boolean to PostHog string variants. Now that migration is complete, the helper just does `value === 'new-home'` - trivial enough to inline directly.

## Test plan

- [x] TypeScript compiles without errors related to these changes
- [x] No remaining references to the deleted helper (verified with grep)
- [x] Tested both variants in vercel preview and the homepage loads fine

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal feature flag evaluation logic to reduce code complexity and improve maintainability.


alaister (approved):
LGTM!

- Tested on preview

Issue #37842, opened 2025-08-11
Wrong server port for local postgres connection


main-uk:
## Describe the bug

The generated documentation for the local instance displayed the wrong server port

## To Reproduce

1. Run an local instance of Supabase
2. Click on 'Connect'
3. Click on 'View Parameters'
4. Scroll down to 'port'
5. Port is displayed as 5432 instead of 54322
## Expected behavior

The port should be 54322 as per the default toml configuration

## Screenshots

<img width="1728" height="1080" alt="Image" src="https://github.com/user-attachments/assets/61c040bc-453c-4b68-91a5-2627e32bea56" />

aimanafzal:
Please assign this to me. 
I have found the issue. @main-uk @neiled @laktek @madx @RandalSchwartz

RohanHandore:
Hi! I've created a PR to fix this: #41991

The fix updates all hardcoded port 5432 values to use the correct 54322 port for local Supabase CLI instances. Tested locally and working correctly. Ready for review!

aantti:
Appreciate it @RohanHandore! Let's try to finalize :)

aantti:
Sadly, have to postpone a solution for this one. Keeping it open with the intent of adding further improvements. More context in PR #41991 and PR #42220 😢

aantti:
Should have been solved via PR #46217 :)

PR #41991, opened 2026-01-19
fix(studio): correct local postgres port from 5432 to 54322

RohanHandore:

YES

Bug fix

When running a local Supabase instance, the "View Parameters" documentation displays the wrong PostgreSQL port number (5432 instead of 54322).

Issue: #37842

## What is the new behavior?

The port now correctly displays as 54322 for local Supabase CLI instances, matching the default port configured in `supabase/config.toml`.

## Additional context

- Updated `POSTGRES_PORT` constant default from 5432 to 54322
- Updated project settings to use `POSTGRES_PORT` constant instead of hardcoded value
- Fixed databases API endpoint to return correct port
- Fixed project props API endpoint to return correct port

Tested locally and verified that the connection string and View Parameters now show port 54322.
image

## Summary by CodeRabbit

* **Chores**
  * Ensure PostgreSQL port is read as a numeric environment value when provided, defaulting to 5432, and applied consistently across settings and APIs.
  * Introduce a safe DB host resolver that derives the host from SUPABASE_PUBLIC_URL when available, falling back to 127.0.0.1 on missing or invalid values.


aantti:
Thanks, much, for submitting this one, @RohanHandore ! It's a great start :) It would actually be super useful if this was a complete fix, though. Here's what has to be addressed.

This change assumes Studio only runs locally when started by [CLI/local-dev](https://supabase.com/docs/guides/local-development), but it also can be part of [self-hosted Supabase](https://supabase.com/docs/guides/self-hosting) instance.

Self-hosted configuration already includes host and port that would be convenient to use in the connect component:

https://github.com/supabase/supabase/blob/c51fa00db2e919eda786bb551f8e85ea863406cd/docker/docker-compose.yml#L35

A proper fix for CLI would ideally include changes somewhere around:

https://github.com/supabase/cli/blob/c4b381a7ffcd5ee8a7df8cdd44564d394b186380/internal/start/start.go#L1155

to include host and port as env-vars to the Studio container. The port in particular can be (re)defined in config.toml, it's not necessarily always `54322` or so is my understanding.

So basically, for CLI - I'd recommend submitting a PR for the CLI codebase too 🙈

If you'd also have time to fix the "connect" UI in [self-hosted Supabase](https://supabase.com/docs/guides/self-hosting) - do you think you could amend this PR and test it against a Docker Compose instance? That'd be massive and greatly appreciated.

RohanHandore:
Thanks for the feedback, @aantti! I understand now... Studio needs to handle both CLI and self hosted contexts.

I'll work on:
1. **CLI fix**: Submit a PR to the CLI repo to pass the port from config.toml as an env var to Studio
2. **Self-hosted fix**: Update this PR to properly read from `POSTGRES_PORT` env var for Docker Compose instances

Let me start with the self-hosted fix in this PR first, then move to the CLI repo. I'll test against Docker Compose to make sure it works correctly.

Thanks for the guidance! 🙏

RohanHandore:
Hi @aantti, thanks for the feedback! I have a question to make sure I implement this correctly ... Should I update this PR first with the self-hosted fix, then create a separate PR for the CLI changes? Or would you prefer both fixes together?

aantti:
@RohanHandore if you could amend this PR with the focus of making it work for both self-hosted and CLI first, that'd be great. My current thinking:

- Use `POSTGRES_PORT` environment variable otherwise still use the default `5432` (I think this is what the existing code does, so no change needed?)
- Use `SUPABASE_PUBLIC_URL` if available - to extract the hostname part - and use it for `db_host` instead of currently hardcoded `127.0.0.1`. For CLI this will just remain `127.0.0.1` and for self-hosted it's going to be what was defined in the [.env file](https://github.com/supabase/supabase/blob/master/docker/.env.example)

Then you could add a separate PR in the CLI repo to pass `POSTGRES_PORT` as part of Studio container initialization.

I don't believe you'll need it as part of this change, but just in case - `!IS_PLATFORM` is used in Studio to determine if it's pointing to a local instance.

Does it make sense? 🙈

RohanHandore:
Hi @aantti! I've updated the PR with the changes you requested:

✅ **POSTGRES_PORT**: Now uses env var if available, otherwise defaults to 5432 (CLI will pass 54322 via env var later)

✅ **db_host**: Added `getDbHost()` function that:
   - Extracts hostname from `SUPABASE_PUBLIC_URL` if available (for self-hosted)
   - Falls back to `127.0.0.1` if not set (for CLI/local instances)

✅ **Updated all locations**: Changed all hardcoded `db_host` values to use `getDbHost()`:
   - `apps/studio/lib/api/self-hosted/settings.ts`
   - `apps/studio/pages/api/platform/projects/[ref]/databases.ts`
   - `apps/studio/pages/api/platform/props/project/[ref]/api.ts`

The fix now works for both CLI and self-hosted contexts as requested. Ready for review! 🚀

aantti:
Thanks, will check!

aantti:
@RohanHandore Can you possibly run prettier for constants.ts and update? :)

RohanHandore:
@aantti Done! I've run prettier on `constants.ts` and pushed the update. Thanks for catching that! 👍

aantti:
@RohanHandore Appreciate you spending time on this. Embarrassingly, while reviewing it and thinking a bit more about what could / should be fixed for both CLI and self-hosted, I've realized it's actually quite a bit more convoluted that I was thinking. I do really value your time and effort of trying to contribute this one, but it looks like there's no easy and quick fix for either CLI or self-hosted.

I've tried to think a bit more of it (via Claude Code, tbh) - and here's a [draft "fix"](https://github.com/supabase/supabase/pull/42220) that I'm gonna test and see if it makes sense to go that route.

My apologies if this sounds disappointing - I should've spent more time on the initial analysis.

aantti:
@RohanHandore I'm sorry, I'll be closing this one. There's some revamping going on currently around this kind of configuration UI and improving it for self-hosted is currently of a lower priority internally. We'll get back to fixing it for self-hosted but not momentarily. Once again, I really appreciate you effort - it helped to see this problem in its entirety.

aantti:
Ftr, see PR #46217.

PR #42005, opened 2026-01-20
docs: document Prisma v7+ migration workaround

### LINKED ISSUE #41621: Prisma no longer supports `url` and `directUrl` in schema.prisma, forcing DB connection to be defined in code and breaking Supabase migration flow
In the latest Prisma version (7.2.0), the database connection can no longer be defined inside schema.prisma using DATABASE_URL.

Previously, Prisma allowed defining the database connection directly in the schema file:

```prisma
datasource db {
  generator client {
    provider = "prisma-client-js"
  }
  
  datasource db {
    provider  = "postgresql"
    url       = env("DATABASE_URL")
    directUrl = env("DIRECT_URL")
  }
 }
```

However, this no longer works in the current version. The schema now requires the datasource to not include url and directUrl, and the database connection must instead be defined directly when instantiating PrismaClient in application code.

Example of the current required setup:
``` typescript
import { PrismaClient } from "@/generated/prisma";

const prisma = new PrismaClient({
  accelerateUrl: process.env.DATABASE_URL!,
});
```

Since Supabase still requires two separate database URLs, one for application runtime (pooled connection) and another for migrations (direct connection), this change creates a significant workflow issue.

Currently, `prisma.config.ts` relies on `DATABASE_URL`:
```typescript
datasource: {
  url: process.env["DATABASE_URL"],
}
```

However, when running:
```bash
npx prisma migrate
```
the command fails to migrate (stuck) because Supabase does not allow migrations using the pooled connection. Migrations must use the direct connection URL instead.

As a result, developers are forced to manually replace `DATABASE_URL` with `DIRECT_URL` in the configuration whenever running migrations. This is error-prone, breaks automation, and defeats the purpose of having separate connection types.

This behavior makes Prisma’s current workflow incompatible with Supabase’s recommended connection setup

whoistausif:
## What this PR does
Adds documentation for Prisma v7+ users where `directUrl` is no longer
supported in `schema.prisma`.

## Why
Prisma v7.2.0 breaks Supabase migration workflows when using pooled
connections. This PR documents a safe workaround.

## Related Issue
Fixes #41621


1Ghasthunter1:
@ChrisChinchilla is there a work-around/alternative instead of just closing this?

ChrisChinchilla:
@1Ghasthunter1 Possibly, but I closed this as the PR was empty

PR #42022, opened 2026-01-21
chore(studio): add URL state management to database tables page

dnywh:

Minor improvement.

Using the text input to search for a specific database table (`/database/tables`) does not add a URL query parameter. Similar pages like `/database/indexes` does this.

This is useful for when linking to this page and wanting to target a specific table. E.g. https://github.com/supabase/supabase/pull/41527/

## What is the new behavior?

- Replace deprecated `Input` with one from `DataInputs/Input`
- Add URL state management
- Syncs the search value with the URL query parameter

## To test

- [ ] Make sure search on `/database/tables` works as it did before
- [ ] See if the URL param updates as this input value is added, changed, and removed

## Summary by CodeRabbit

* **New Features**
  * Search filter now persists in the URL, enabling users to share filtered search states and maintain their search term when returning to the page.

### DIFF
--- a/apps/studio/components/interfaces/Database/Tables/TableList.tsx
+++ b/apps/studio/components/interfaces/Database/Tables/TableList.tsx
@@ -16,6 +16,7 @@ import {
 } from 'lucide-react'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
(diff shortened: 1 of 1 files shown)


joshenlim (approved):
Verified changes on preview!

PR #42042, opened 2026-01-21
Joshen/fe 2287 summary nodegraph in replication page

joshenlim:
## Context

Related to unifying read replicas into database/replication - adds a read-only summary node graph to the top of the page

- Each replica / destination will be it's own node
  - Opting to show _all_ nodes for now, we can consider further down whether we want to consolidate nodes (e.g "+ 4 more targets" or something to prevent rendering too many nodes although I don't think that'll be a problem for now)
- The animated edges are meant to be indicative of an active replication
  - e.g if the replica status is not "active",  the edge will be a static dashed line with a slightly lower opacity

https://github.com/user-attachments/assets/72fdbae4-aa2b-457c-9a50-74fb0161ff18

## Summary by CodeRabbit

* **New Features**
  * Added a visual, read-only replication topology diagram showing primary, read replicas, and destinations with status indicators, region labels, and connection handles.

* **Improvements**
  * Integrated the diagram into the replication page as a full-width section and refreshed layout/spacing for clearer organization.
  * More efficient, data-driven status refreshing that stops when replicas are stable to reduce unnecessary updates.


apps/studio/components/interfaces/Database/Replication/Destinations.tsx
```diff
@@ -58,9 +58,7 @@ export const Destinations = () => {
     isError: isDatabasesError,
     isSuccess: isDatabasesSuccess,
     refetch: refetchDatabases,
-  } = useReadReplicasQuery({
-    projectRef,
-  })
+  } = useReadReplicasQuery({ projectRef }, { refetchInterval: statusRefetchInterval })
```
joshenlim, line 56: JFYI i think i can simplify the long polling logic here to not depend on the replica status query actually

will look into this separately

apps/studio/components/interfaces/Database/Replication/ReplicationDiagram/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+          style: {
+            opacity: isReplicating ? 1 : 0.4,
+            strokeDasharray: isReplicating ? undefined : '5 5',
+          },
+        }
+      }),
+    ]
+  }, [destinations, pipelinesData?.pipelines, projectRef, queryClient, readReplicas])
+
+  const nodeTypes = useMemo(
```
alaister: Very much a nit, but this is static right? So you could just move it outside of the component

alaister (approved):
LGTM!

- Tested on preview

apps/studio/components/interfaces/Database/Replication/ReplicationDiagram/index.tsx
joshenlim: oh yup! that's fair

PR #42043, opened 2026-01-21
fix: handle errors better in index advisor

awaseem:

YES

fix: index advisor would show a unhelpful error when analyzing queries that referenced deleted entries. This should handle that case and show a more helpful error message

## How to test

- Create a table with zero indexes
- Use a column and run SQL statements 
- See if the query showes up in query performance (better to search)
- Delete the table
- Check index advisor for this error

## Demo
Screenshot 2026-01-21 at 10 38 21 AM

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented abrupt errors when index recommendations are missing; the system now logs issues and returns a safe state so the app remains responsive.
* **UI/UX**
  * Shows a clear "Index recommendations not available" notice when advisor data is absent, while preserving existing empty/result flows and messaging.

### DIFF
--- a/apps/studio/components/interfaces/QueryPerformance/QueryIndexes.tsx
+++ b/apps/studio/components/interfaces/QueryPerformance/QueryIndexes.tsx
@@ -1,10 +1,11 @@
-import { Check, Table2, Lightbulb, X } from 'lucide-react'
+import { Check, Table2, Lightbulb } from 'lucide-react'
 import { useState, useEffect } from 'react'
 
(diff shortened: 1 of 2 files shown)


kemaldotearth (changes requested):
Very nice, small nit to change, could we use the same info icon from our admonitions? Perhaps an admonition works a bit better than alert here as it's informational? I guess we can kill the border of the admonition in that side panel context.

Image

See [here](https://supabase-design-system.vercel.app/design-system/docs/fragments/admonition) for ref, otherwise good stuff!

kemaldotearth (approved):

PR #42056, opened 2026-01-22
chore(studio): update upgrade to pro admonition button variants

dnywh:

- UI improvement
- Resolves DEPR-117

- The file storage settings `UpgradeToPro` admonition uses the green primary button variant. This visually collides with the ‘Save’ button beneath it, which uses the same variant (but is disabled anyway since the user must be Pro plan to do anything on this form)
- Same problem with the SMS MFA `UpgradeToPro` admonition
- The file storage settings Spend cap admonition (also `UpgradeToPro` under the hood) clashes with the (enabled) ‘Save’ button beneath it
- Other minor stylistic and copywriting issues with the `UpgradeToPro` admonition instances

There should only be one primary button at any one time within a view to make it clear what the primary action is.

## What is the new behavior?

- File storage settings `UpgradeToPro` admonition uses the `primary` button variant. The Save button beneath becomes default (disabled)
- Spend cap admonition uses default button variant since the sibling Save button is always functional
- SMS MFA `UpgradeToPro` admonition continues to use the `primary` button variant. The Save button below it becomes (disabled) `default` whilst that admonition is shown to distinguish the two

Other updates:

- Updated the ‘spend cap’ language to be sentence case, as per our copywriting guidelines
- Reset the global `bg` override on the admonition so it is distinguished from the form contents.

| Before | After |
| --- | --- |

## Additional context

The admonition is actually wrapped in an `UpgradeToPro` component. I don’t fully understand why. Leaving as-is since billing is related.

## Summary by CodeRabbit

* **Improvements**
  * Refined upgrade prompt wording across many settings for clearer, more consistent messaging and punctuation.
  * Call-to-action buttons for upgrade prompts and save actions now adapt their visual style based on plan/entitlement, improving emphasis and guidance.
  * Backups view now displays the upgrade prompt when backup access is missing.
  * Minor spacing and layout tweaks to upgrade panels for improved visual consistency.

### DIFF
--- a/apps/studio/components/ui/UpgradeToPro.tsx
+++ b/apps/studio/components/ui/UpgradeToPro.tsx
@@ -58,7 +58,7 @@ export const UpgradeToPro = ({
           source={source}
           featureProposition={featureProposition}
           disabled={disabled}
-          type={buttonVariant}
+          variant={buttonVariant}
         >
           {buttonText}
         </UpgradePlanButton>
--- a/apps/studio/components/ui/UpgradeToPro.tsx
+++ b/apps/studio/components/ui/UpgradeToPro.tsx
@@ -48,7 +48,7 @@ export const UpgradeToPro = ({
       className={cn(
         'mb-0',
         fullWidth &&
-          'border-0 rounded-none border-b px-6 [&>svg]:left-6 [&>div>button]:w-min [&>div>button]:mt-3 [&>div>a]:w-min [&>div>a]:mt-3',
+          'border-0 rounded-none border-b',
         className
       )}
       actions={
--- a/apps/studio/components/ui/UpgradeToPro.tsx
+++ b/apps/studio/components/ui/UpgradeToPro.tsx
@@ -45,12 +45,7 @@ export const UpgradeToPro = ({
       layout={layout}
       title={primaryText}
       description={secondaryText}
-      className={cn(
-        'mb-0',
-        fullWidth &&
-          'border-0 rounded-none border-b',
-        className
-      )}
+      className={cn(fullWidth && 'border-0 rounded-none border-b', className)}
       actions={
         <UpgradePlanButton
           plan={planToUpgrade}
--- a/apps/studio/components/ui/UpgradeToPro.tsx
+++ b/apps/studio/components/ui/UpgradeToPro.tsx
@@ -19,7 +19,7 @@ interface UpgradeToProProps {
   disabled?: boolean
   fullWidth?: boolean
   layout?: 'vertical' | 'horizontal'
-  buttonVariant?: 'default' | 'primary'
+  variant?: 'default' | 'primary'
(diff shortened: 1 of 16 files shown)


apps/studio/components/ui/UpgradeToPro.tsx
```diff
@@ -18,8 +18,9 @@ interface UpgradeToProProps {
   source?: string
   disabled?: boolean
   fullWidth?: boolean
-  className?: string
   layout?: 'vertical' | 'horizontal'
+  buttonVariant?: 'default' | 'primary'
```
dnywh: New prop needed in order to target the button.

dnywh: I’ve named it `buttonVariant` because:
1. `type` means something else in TypeScript
2. We are moving towards `variant` more broadly to mean a stylistic type
	a. Shadcn does [the same thing](https://ui.shadcn.com/docs/components/radix/button)

apps/studio/components/ui/UpgradeToPro.tsx
```diff
@@ -44,7 +46,7 @@ export const UpgradeToPro = ({
       title={primaryText}
       description={secondaryText}
       className={cn(
-        'bg-surface-100 mb-0',
```
dnywh, line 47: Removed this `bg` override. It should be set on any children that specifically need it.

I did a brief check against other `UpgradeToPro` instances, with nothing looking out of place. I’ll do another sweep in a second and make any per-child changes as needed.

apps/studio/components/ui/UpgradeToPro.tsx
```diff
@@ -56,6 +58,7 @@ export const UpgradeToPro = ({
           source={source}
           featureProposition={featureProposition}
           disabled={disabled}
+          type={buttonVariant}
```
dnywh: We could also change the `type` prop on this `UpgradePlanButton` to match the `variant` naming. This button component is used in 13 other places, so that would make this PR a bit more full-on. I’m happy to do it, though.

dnywh: Done

apps/studio/components/ui/UpgradeToPro.tsx
dnywh: - Also removed other custom `className` overrides
- Swept everything
- All okay

apps/studio/components/ui/UpgradeToPro.tsx
joshenlim: nit - should we just call this prop `variant` instead? since that's a common prop name that we're using rather than `buttonVariant` (from a consistency POV)

happy to keep this as is too - since its clearer what the variant is referring to (the button)

joshenlim (approved):
Verified changes on preview!

PR #42080, opened 2026-01-22
fix(studio): use paginated buckets endpoint for command menu

charislam:

YES

Bug fix

The command menu search uses the unpaginated buckets endpoint, which times out or causes jankiness when users have tens of thousands of buckets.

## What is the new behavior?

Uses the paginated buckets endpoint. Only the first 10 results are shown, and users need a more specific search query to find other buckets (same UX as before, but with better performance).

## Additional context

Towards FE-2118

## Summary by CodeRabbit

* **Performance Improvements**
  * Debounced search queries to reduce API calls during storage searches
  * Paginated fetching for bucket lists and server-side filtering for file buckets

* **New Features**
  * Display estimated total bucket count with an “(estimate)” indicator in results
  * Improved empty-state and footer behavior to align with the debounced search flow

### DIFF
--- a/apps/studio/components/interfaces/App/CommandMenu/StorageSearchResults.tsx
+++ b/apps/studio/components/interfaces/App/CommandMenu/StorageSearchResults.tsx
@@ -1,23 +1,29 @@
 'use client'
 
-import { useMemo, useCallback } from 'react'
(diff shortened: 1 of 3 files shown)


awaseem (approved):
Tested in staging!

PR #42082, opened 2026-01-22
fix: add description of column on header name

awaseem:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES/NO

## What kind of change does this PR introduce?

Minor: add description of column on header name

## Demo
Screenshot 2026-01-22 at 3 16 49 PM

## Summary by CodeRabbit

* **New Features**
  * Column headers can now show optional comments as tooltips beneath the column name for extra context; headers without comments render as before.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #42090, opened 2026-01-23
feat(studio): show SQL snippet description

dnywh:

- Feature
- Resolves DEPR-332

Some users add descriptions to their SQL snippets. We don’t show this description anywhere.

## What is the new behavior?

We show this optional description when hovering over the snippet’s name on the sidebar.

| Before | After |
| --- | --- |
| before | after3 |

## Additional context

- A cleaner approach might be to deprecate the description field but people are actively using it
	- It might come in handy later too, e.g. with improved sharing
- I chose to append the description to the snippet name as the `title` element already is in use for that
	- This `title` is needed because some (many of my) snippets are truncated
- Our built-in `Tooltip` component is a too heavy-handed for this
	- The HTML `title` element has the perfect weight, I think, for this type of affordance

## Summary by CodeRabbit

* **Improvements**
  * Tree view items in the SQL editor now show clearer titles and optional descriptions, improving readability.
  * Tooltips updated: item hover text now combines title and description when present; redundant per-element tooltips removed for a cleaner UI.
  * Names can display richer content (not limited to plain text), ensuring more flexible item rendering across navigation and search.

### DIFF
--- a/apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
+++ b/apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
@@ -115,13 +115,9 @@ export const SearchList = ({ search }: SearchListProps) => {
                     ...element,
                     name: (
                       <span className="flex flex-col py-0.5">
-                        <span className="truncate">
-                          {element.name}
-                        </span>
+                        <span className="truncate">{element.name}</span>
                         {!!visibility && (
-                          <span className="text-foreground-lighter text-xs">
-                            {visibility}
-                          </span>
+                          <span className="text-foreground-lighter text-xs">{visibility}</span>
                         )}
                       </span>
                     ),
--- a/apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
+++ b/apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
@@ -115,17 +115,18 @@ export const SearchList = ({ search }: SearchListProps) => {
                     ...element,
                     name: (
                       <span className="flex flex-col py-0.5">
-                        <span title={element.name} className="truncate">
+                        <span className="truncate">
                           {element.name}
                         </span>
                         {!!visibility && (
-                          <span title={visibility} className="text-foreground-lighter text-xs">
+                          <span className="text-foreground-lighter text-xs">
                             {visibility}
                           </span>
                         )}
                       </span>
                     ),
                   }}
+                  nameForTitle={element.name}
                   isBranch={false}
                   isOpened={isOpened && !isPreview}
                   isSelected={isActive}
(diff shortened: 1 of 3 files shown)


apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
```diff
@@ -115,17 +115,18 @@ export const SearchList = ({ search }: SearchListProps) => {
                     ...element,
                     name: (
                       <span className="flex flex-col py-0.5">
-                        <span title={element.name} className="truncate">
+                        <span className="truncate">
                           {element.name}
                         </span>
                         {!!visibility && (
-                          <span title={visibility} className="text-foreground-lighter text-xs">
```
dnywh, line 122: Redundant as it's always 'Private' or 'Shared'; never truncated.

apps/studio/components/layouts/SQLEditorLayout/SQLEditorNavV2/SearchList.tsx
```diff
@@ -115,17 +115,18 @@ export const SearchList = ({ search }: SearchListProps) => {
                     ...element,
                     name: (
                       <span className="flex flex-col py-0.5">
-                        <span title={element.name} className="truncate">
```
dnywh, line 118: This and the below `title` collide with that on the parent row. Removing.

packages/ui/src/components/TreeView/TreeView.tsx
```diff
@@ -63,7 +63,11 @@ const TreeViewItem = forwardRef<
     /** The horizontal padding of the item */
     xPadding?: number
     /** name of entity */
-    name: string
+    name: string | ReactNode
+    /** Optional description of entity */
+    description?: string
+    /** String name to use for title attribute when name is a ReactNode */
```
dnywh, line 69: Name comes through as a ReactNode on the `SearchList`.

packages/ui/src/components/TreeView/TreeView.tsx
```diff
@@ -254,7 +265,7 @@ const TreeViewItem = forwardRef<
               />
             )
           )}
-          <span className={cn(isEditing && 'hidden', 'truncate text-sm')} title={name}>
```
dnywh, line 257: As above: removed to avoid clashing with `title` on parent row.

packages/ui/src/components/TreeView/TreeView.tsx
```diff
@@ -99,7 +105,8 @@ const TreeViewItem = forwardRef<
     },
     ref
   ) => {
-    const [localValueState, setLocalValueState] = useState(name)
+    const nameString = nameForTitle ?? (typeof name === 'string' ? name : '')
+    const [localValueState, setLocalValueState] = useState(nameString)
```
dnywh, line 109: As above: needs to be checked as sometimes `name` is a ReactNode.

SaxonF (approved):
Tested in design system and in studio on snippets and edge function code

joshenlim (approved):
Verified changes on preview - although separately i reckon we look into consolidating how we wanna handle description tooltips like these - via title or Tooltip component

this and [this PR](https://github.com/supabase/supabase/pull/42082) does similar things, but deviates in how its presented - am presuming we might also opt to do something similar for table comments too

PR #42093, opened 2026-01-23
Validate against empty name in FileExplorerRowEditing.tsx

joshenlim:
## Context

Patches a dashboard bug whereby:
- Users can rename a folder to an empty string, should not be allowed
- Renaming a folder throws a toast error (2nd level onwards, not at the root)

## Changes involved
- Am using MacOS finder as a prior art - setting the folder name to an empty string and saving (either by hitting enter or clicking out of the input) will just reset the name to its original name

## To test
- [ ] Verify that you cannot set a folder name (or item name) to an empty string
- [ ] Verify that you can rename folders (at the root, or nested folders)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented saving empty names when renaming items/folders; trimming now falls back to the original name.
  * Improved folder-rename guard to avoid runtime errors and related UI issues.
* **Tests**
  * Added end-to-end tests ensuring empty-name renames are canceled and the rename UI closes appropriately.

### DIFF
--- a/apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerRow.tsx
+++ b/apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerRow.tsx
@@ -17,7 +17,6 @@ import {
 import { useContextMenu } from 'react-contexify'
 import SVG from 'react-inlinesvg'
 
-import { useParams } from 'common'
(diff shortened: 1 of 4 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #42096, opened 2026-01-23
Adjust disabled state while project is paused

joshenlim:
## Context

The original issue this PR is trying to address is to make edge functions accessible irregardless of the project's DB status.
But figured that we also look into the dashboard's state while the project isn't active healthy.

## Current behaviour 

The current behaviour is that only certain pages of the project are accessible while the project is not active healthy, but I realised this isn't clearly communicated as the side nav looks like everything is accessible while the project is paused. Clicking the side nav for each item just goes to the URL, but shows the `ProjectPausedState` still, e.g:
image

Am thinking the ideal UX is to show disabled states, rather than hiding UI so the changes here are trying to address that
Happy to discuss about this further tho or even push this back as I'm not 100% about this

## Changes involved

- Project side navigation bar shows disabled state for all pages that are not accessible while project is paused (Reckon this makes is clearer _what is accessible_ in this state)
image

- Always show the product menu, but disable the links that are not accessible while project is paused, rather than hiding the entire product menu which is the current behaviour
  - From a DX POV, am thinking this also makes us more intentional when deciding which pages are not dependent on the database status (i reckon there's more pages that we can let users access e.g API keys)
image

## Long term goal

- Ideal state will be that dashboard UI doesn't differ greatly irregardless of project's status
- Opt to show disabled states where appropriate over hiding UI as a general design rule of thumb
  - Allows us to reveal what actions are available, if the preceding condition is met - existing example would be auth forms like the performance one that's gated behind a paid plan


alaister (approved):
LGTM!

- Tested on preview

PR #42148, opened 2026-01-25
docs(skills): agent skills

Rodriguespn:

YES

Docs update — Rewrite the Agent Skills documentation as a single flat page with a skills table.

The Agent Skills docs have a nested structure with an index page and individual dynamic pages for each skill (fetched from the `supabase/agent-skills` repo). Skills also inject sub-items into the sidebar navigation.

## What is the new behavior?

This PR replaces the nested skill pages with a single, flat Agent Skills page that:

- Lists all skills in a **table** with name, description, and a **copy-to-install button** (fetched dynamically from the [supabase/agent-skills](https://github.com/supabase/agent-skills) repo)
- Includes **installation commands** for both the skills CLI (`npx skills add`) and Claude Code plugins
- Links skill names directly to their source on GitHub instead of rendering full skill content inline
- Removes the dynamic `[slug]` route, sidebar nav injection, and local skill example files

### Navigation Structure

```
Start
  > AI Tools
      > Agent Skills   (new)
      > Prompts        (existing)
      > Supabase MCP server (existing)
```

Closes [AI-361](https://linear.app/supabase/issue/AI-361/create-skills-documentation-page-in-ai-tooling-docs)


ChrisChinchilla (approved):
Looks good ot me, but @jeremenichelli should also take a quick look, mostly as we added some better error handling around federated page content recently that we should also probably add here.

apps/docs/app/guides/getting-started/ai-skills/AiSkills.utils.ts
```diff
@@ -0,0 +1,101 @@
...
+      headers: {
+        Accept: 'application/vnd.github.v3+json',
+        'User-Agent': 'Supabase-Docs',
+      },
+      next: { revalidate: 3600 },
+    })
+
+    if (!response.ok) {
+      throw new Error(`GitHub API error: ${response.status}`)
+    }
```
jeremenichelli: With this error handling pattern, the build might silently fail and we won't know, or could cause timeouts. I suggest to follow this pattern we've recently implemented to catch build errors correctly:

```ts
try {
  response = await fetch(`https://raw.githubusercontent.com/...`)
} catch (err) {
  throw new Error(`Failed to fetch <insert the project related docs> from GitHub (network error): ${err}`)
}

if (!response.ok) {
  throw new Error(
    `Failed to fetch <insert the project related docs> docs from GitHub: ${response.status} ${response.statusText}`
  )
}
```

apps/docs/app/guides/getting-started/ai-skills/AiSkillsIndex.tsx
```diff
@@ -0,0 +1,7 @@
+import { getAiSkills } from './AiSkills.utils'
+import { AiSkillsTable } from './AiSkillsTable'
+
+export async function AiSkillsIndex() {
+  const skills = await getAiSkills()
+  return <AiSkillsTable skills={skills} />
```
jeremenichelli: This file adds unnecessary volume to the PR, why isn't the table all in this component? I don't see how we will use it outside the boundaries of AI Skills being fetched.

apps/docs/app/guides/getting-started/ai-skills/AiSkillsIndex.tsx
```diff
@@ -0,0 +1,7 @@
+import { getAiSkills } from './AiSkills.utils'
+import { AiSkillsTable } from './AiSkillsTable'
+
+export async function AiSkillsIndex() {
+  const skills = await getAiSkills()
```
jeremenichelli: The fetching, preferably should be done within a route.ts file so it happens at server time.

You can then pass it as metadata to consume within the page.

It is better to keep components usage agnostic and not to fire network calls.

To make it clear since this might be connected with other comments:
 - Move fetching to a `route.ts` file (this might make you create a `layout.tsx` file just because of it)
 - Eliminate the `AISkillsIndex` file and change the table to a more agnostic usage table. In this table data should be injected, and probably the addition of a copy button should be optionable (I can help you design that component API if needed).
 - Add the table component in the MDX file and use the data fetched in `route.ts`.

apps/docs/components/Breadcrumbs.tsx
```diff
@@ -176,6 +176,16 @@ function useBreadcrumbs() {
     return breadcrumbs
   }
 
+  const isAiSkillsPage = pathname.startsWith('/guides/getting-started/ai-skills')
```
jeremenichelli, line 181: I know our Breadcrumbs are kind of broken, but not a fun of just _patching it_, if we do, let's leave a comment so we fix it and remove it.

apps/docs/app/guides/getting-started/ai-skills/AiSkills.utils.ts
Rodriguespn: Good catch. Updated both `fetchGitHubDirectory` and `fetchGitHubFile` to follow the established pattern: network errors are caught and re-thrown with context, and HTTP errors are thrown separately with status and statusText. No more silent failures.

apps/docs/app/guides/getting-started/ai-skills/AiSkillsIndex.tsx
Rodriguespn: Agreed. Merged the table rendering directly into `AiSkillsIndex.tsx` and deleted `AiSkillsTable.tsx`. Thanks for the suggestion

apps/docs/app/guides/getting-started/ai-skills/AiSkillsIndex.tsx
Rodriguespn: For now I've kept the async server component pattern (same as `AiPromptsIndex`), since the data is consumed via an MDX component and there's no existing `route.ts` pattern. Happy to refactor to a `route.ts`/`layout.tsx` approach if you'd prefer

apps/docs/components/Breadcrumbs.tsx
Rodriguespn: Added a `// TODO: Remove this patch once Breadcrumbs are properly fixed to handle these routes` comment above the workaround

jeremenichelli:
Two more comments:

- The `<query>` notation kind of breaks the bash syntax: 
Screenshot 2026-03-20 at 15 02 41

I would use something different, or maybe just the word query in uppercase, not sure how we do it in other pages.

-Introducing a new component just for one row, doesn't make that much sense, also de person doesn't know what they are copying which in terms of UX is bad in my opinion.

Screenshot 2026-03-20 at 15 03 46

Why not just use markdown and do:

```raw
### `supabase-postgres-best-practices`

Postgres performance optimization and best practices from Supabase. Use this skill when writing, reviewing, or optimizing Postgres queries, schema designs, or database configurations.

``..
the command... our code components already have a copy option.
``..

[Check this skill in GitHub](link to GitHub)
```

apps/docs/app/guides/getting-started/ai-skills/AiSkills.utils.ts
```diff
@@ -0,0 +1,102 @@
...
+    response = await fetch(url, {
+      headers: {
+        Accept: 'application/vnd.github.v3+json',
+        'User-Agent': 'Supabase-Docs',
+      },
+      next: { revalidate: 3600 },
+    })
+  } catch (err) {
+    throw new Error(`Failed to fetch agent skills directory from GitHub (network error): ${err}`)
+  }
```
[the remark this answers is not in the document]
jeremenichelli: That is actually a valid point @Rodriguespn, I will take this an even update the ones I migrated to this pattern.

apps/docs/app/guides/getting-started/ai-skills/AiSkillsIndex.tsx
jeremenichelli: It's fine for now. Have you checked the message in the end of the PR? I am proposing an alternative to the table. 

I think this approach obscures the command which is not good both for people and for agents reading the content.

apps/docs/app/guides/getting-started/ai-skills/CopyButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Check, Copy } from 'lucide-react'
+import { cn } from 'ui'
+
+export function CopyButton({ text }: { text: string }) {
+  const [copied, setCopied] = useState(false)
+
+  const handleCopy = async () => {
+    await navigator.clipboard.writeText(text)
+    setCopied(true)
+    setTimeout(() => setCopied(false), 2000)
```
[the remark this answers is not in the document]
jeremenichelli: ➕ 1️⃣

apps/docs/content/guides/getting-started/ai-skills.mdx
```diff
@@ -0,0 +1,48 @@
...
+npx skills add supabase/agent-skills
+```
+
+To install a specific skill from the repository:
+
+```bash
+npx skills add supabase/agent-skills --skill <skill-name>
+```
+
+### Claude code plugin
```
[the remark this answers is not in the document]
jeremenichelli: Yeap.

Rodriguespn:
> The `<query>` notation kind of breaks the bash syntax. I would use something different, or maybe just the word query in uppercase.

Changed both `<query>` and `<skill-name>` to uppercase bare words (`QUERY` and `SKILL_NAME`). The docs don't enforce a single convention, but uppercase without brackets is the safest choice for CLI args since `<value>` reads as a bash redirect — you can see the same pattern used across the docs (e.g. `YOUR_PROJECT_ID` in the quickstart guide, `YOUR_PASS` in the replication guide).

> Introducing a new component just for one row doesn't make sense — the person also doesn't know what they are copying, which in terms of UX is bad.

Agreed on the UX problem. I don't want to go with full `###` headings per skill though, because as the catalogue grows the page gets very noisy.

Updated the table instead: the "Install command" column now shows the full command as inline `code` alongside a small copy icon. The column scrolls horizontally within the cell so it doesn't eat into the description column. Users can read the exact command before copying, and the layout stays compact as the skill list grows.

image

apps/docs/app/guides/getting-started/ai-skills/AiSkillsIndex.tsx
Rodriguespn: Replied in the bottom

jeremenichelli (approved):

PR #42154, opened 2026-01-26
feat(studio): recents menu

dnywh:

- Experimental feature

Aside from browser history, there is no way to jump to recently-viewed routes. This can be annoying, especially in such a dense and nested product.

## What is the new behavior?

The five most recently viewed routes are saved to local storage. The command menu displays these.


packages/ui-patterns/src/CommandMenu/api/CommandMenu.tsx
```diff
@@ -199,7 +199,7 @@ function CommandMenuTriggerInput({
             strokeWidth={1.5}
             className="group-hover:text-foreground-light transition-colors"
           />
-          <p className="flex text-sm pr-2 text-foreground-muted">{placeholder}</p>
+          <p className="flex text-xs pr-2 text-foreground-muted">{placeholder}</p>
```
dnywh, line 202: Just matching the font size with other items in the top navigation area.

dnywh:
Prototype that is now outdated by @fsansalvadore’s better nav work!

PR #42197, opened 2026-01-27
Add amdonition about schema to choose

smeubank:
- what schema should i choose when configuring pg_cron?
- what is internal workings that is not really relevant to 90% of decs?
- does it matter if i choose extension or some public schema i created?
- realistically what would i functionally do with the extension schema for cron? Do I need to think about this?

YES

docs update

## What is the new behavior?

Docs don't explain the step when instlling pg_cron that i need to choose a a schema. But there is mention of the cron schema. Which is not an option to choose but does explain some implementation details.

## Additional context

image


Hallidayo:

PR #42198, opened 2026-01-27
chore: Bump next versions to non-vulnerable numbers

ivasilov:
Bumping next to avoid the latest vulnerability. We should be safe because we're not using latest React or RSC.

## Summary by CodeRabbit

* **Chores**
  * Updated Next.js framework dependencies to recent patch versions.


alaister (approved):
LGTM!

- Tested on preview

PR #42239, opened 2026-01-28
Shift replication alpha access request CTA into DestinationPanel

joshenlim:
## Context

Part of unifying read replicas with database replication page

Shifts the Alpha Access request CTA for database replication into the Destination Panel if the feature flag for the new UI is toggled on. So that users can still manage their read replicas

image

## Release Notes

* **New Features**
  * PostgreSQL replication is now available as a private alpha feature. Users can request early access directly from the destination replication panel with a dedicated request button.
  * Enhanced replication configuration interface with new quick-action buttons for accessing documentation and plan upgrade options.

* **UI Improvements**
  * Refined replication panel layout for better navigation and feature discoverability.


alaister (approved):
LGTM!

- Tested on preview

PR #42249, opened 2026-01-28
Fix sidebar highlight sync in the reference docs

Ellba:

YES

Bug fix

On reference documentation pages (/reference/cli/*, /reference/javascript/*, etc.), the sidebar navigation sometimes gets "stuck"; clicking a different section scrolls the page correctly, but the sidebar highlight remains on a previously selected item instead of updating to the newly selected one.

This issue is intermittent and typically occurs after the page has been idle for 2-3 minutes. It appears to be caused by Next.js's usePathname() hook returning stale data after the internal router state becomes desynced from the actual browser URL (which is updated via history.pushState/replaceState).

## What is the new behavior?

The sidebar now reliably highlights the correct active section, both when:
- Clicking sidebar links
- Scrolling through the page (scroll spy)

This is achieved by introducing a custom useCurrentPathname() hook that:
1. Reads directly from window.location.pathname (the source of truth)
2. Subscribes to history.pushState and history.replaceState changes
3. Uses React 18's useSyncExternalStore for proper external state synchronization

The implementation uses a singleton pattern to ensure history methods are only patched once, even when multiple components subscribe.

## Additional context

https://github.com/user-attachments/assets/87da13e0-6b01-44d9-8ce5-f12d0dcaa8fb

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved navigation reliability by enhancing active reference tracking in menus when navigating between documentation pages.
  * Fixed inconsistent pathname updates to ensure more reliable navigation state synchronization across the documentation interface.


ChrisChinchilla (approved):
Looks good, in honesty I want to redo all the reference sidebars as they jump back and forth confusingly.

PR #42259, opened 2026-01-28
feat(telemetry): add dev telemetry toolbar

seanoliver:
## Problem
We need a local-only UI to inspect client and server telemetry events and override flags during development without touching non-local env behavior. This work is intended to be shared across Studio, Docs, and WWW.

## Changes
- Introduced a shared `dev-tools` package with the Dev Telemetry Toolbar UI, trigger, and provider.
- Wired the toolbar into Studio, Docs, and WWW app shells (local-only gating).
- Added a local-only `devTelemetry()` opt-in with storage gating and SSE subscription.
- Wired client PostHog events into a local listener and re-exported types.
- Added local flag override cookie support in the UI and CODEOWNERS for the new package.
- Added unit tests covering local/non-local behavior and flag utilities.

## Testing
Manual (local only):
- Start each app locally: `pnpm dev:studio`, `pnpm dev:docs`, `pnpm dev:www`
- Open the app, run `devTelemetry()` in the browser console
- Click around and confirm both client and server events appear (client will be page views only)
- Verify feature flag overrides (PostHog + ConfigCat) persist and restore correctly
- Confirm dismissing the toolbar clears local storage and hides the trigger

Unblocked by https://github.com/supabase/platform/pull/29172

Resolves GROWTH-591

Demo:
[github.com/user-attachments/assets/60b376db-7440-4ada-82f5-d1bd4af4db3b](https://github.com/user-attachments/assets/60b376db-7440-4ada-82f5-d1bd4af4db3b)

Screenshots:

## Summary by CodeRabbit

* **New Features**
  * Developer Toolbar (local only): view client/server telemetry, inspect events, and manage/override feature flags with persistent overrides, filtering, and clear/reload.
  * Client-side telemetry hooks: surface structured events to dev tooling for realtime inspection.

* **Bug Fixes**
  * Fixed end-of-file newline in shared code.

* **Chores**
  * Added dev-tools package, integrated provider and trigger across Studio, Docs, and marketing sites, and added CODEOWNERS entry.

* **Tests**
  * Added comprehensive tests and test setup for the DevToolbar.

### DIFF
--- a/apps/studio/components/ui/DevTelemetryToolbar/DevTelemetryToolbar.tsx
+++ b/apps/studio/components/ui/DevTelemetryToolbar/DevTelemetryToolbar.tsx
@@ -1,0 +1,445 @@
+'use client'
+
+import { useFeatureFlags } from 'common'
+import { Activity, ChevronDown, ChevronUp, Flag } from 'lucide-react'
+import {
+  type ChangeEvent,
+  type Dispatch,
+  type KeyboardEvent,
+  type SetStateAction,
+  useCallback,
+  useEffect,
+  useState,
+} from 'react'
+import {
+  Badge,
+  Button,
+  Input_Shadcn_ as Input,
+  Sheet,
+  SheetContent,
+  SheetDescription,
+  SheetHeader,
+  SheetTitle,
+  Switch,
+  Tabs_Shadcn_ as Tabs,
+  TabsContent_Shadcn_ as TabsContent,
+  TabsList_Shadcn_ as TabsList,
+  TabsTrigger_Shadcn_ as TabsTrigger,
+  cn,
+} from 'ui'
+
+import { type DevTelemetryEvent, useDevTelemetryToolbar } from './DevTelemetryToolbarContext'
+import {
+  CC_ORIGINALS_KEY,
+  PH_ORIGINALS_KEY,
+  deleteCookie,
+  getCookie,
+  parseOverrideValue,
+  readOriginals,
+  safeJsonParse,
+  setCookie,
+  valuesAreEqual,
+  writeOriginals,
+} from './utils'
+
+const IS_LOCAL_DEV = process.env.NEXT_PUBLIC_ENVIRONMENT === 'local'
+
+function EventCard({ event }: { event: DevTelemetryEvent }) {
+  const [isExpanded, setIsExpanded] = useState(false)
+
+  const handleToggle = () => setIsExpanded((prev) => !prev)
+
+  const handleKeyDown = (e: KeyboardEvent<HTMLDivElement>) => {
+    if (e.key === 'Enter' || e.key === ' ') {
+      e.preventDefault()
+      handleToggle()
+    }
+  }
+
+  return (
+    <div className="border rounded-md p-3 bg-surface-100">
+      <div
+        role="button"
+        tabIndex={0}
+        aria-expanded={isExpanded}
+        aria-label={`${event.eventName} event details`}
+        className="flex items-start justify-between cursor-pointer gap-4 focus:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 rounded-sm"
+        onClick={handleToggle}
+        onKeyDown={handleKeyDown}
+      >
+        <div className="flex flex-col gap-1 min-w-0 flex-1">
+          <div className="flex items-start gap-2 flex-wrap">
+            <Badge variant={event.source === 'client' ? 'default' : 'success'} className="shrink-0">
+              {event.source}
+            </Badge>
+            <Badge variant="secondary" className="shrink-0">
+              {event.eventType}
+            </Badge>
+            <span className="font-mono text-sm break-all">{event.eventName}</span>
+          </div>
+          {event.distinctId && (
+            <div
+              className="text-xs text-foreground-muted font-mono truncate"
+              title={event.distinctId}
+            >
+              ID: {event.distinctId}
+            </div>
+          )}
+        </div>
+        <div className="flex items-center gap-2 text-foreground-muted shrink-0">
+          <span className="text-xs whitespace-nowrap">
+            {new Date(event.timestamp).toLocaleTimeString()}
+          </span>
+          {isExpanded ? <ChevronUp className="w-4 h-4" /> : <ChevronDown className="w-4 h-4" />}
+        </div>
+      </div>
+
+      {isExpanded && (
+        <pre className="mt-3 p-2 bg-surface-200 rounded text-xs overflow-x-auto max-h-[300px] overflow-y-auto">
+          {JSON.stringify(event.properties, null, 2)}
+        </pre>
+      )}
+    </div>
+  )
+}
+
+function FlagCard({
+  flagName,
+  currentValue,
+  originalValue,
+  isOverridden,
+  onToggle,
+}: {
+  flagName: string
+  currentValue: unknown
+  originalValue: unknown
+  isOverridden: boolean
+  onToggle: (value: unknown) => void
+}) {
+  const valueType = typeof originalValue
+  const isNull = originalValue === null
+  const inputProps = {
+    value: String(currentValue),
+    onChange: (event: ChangeEvent<HTMLInputElement>) => onToggle(event.target.value),
+    className: 'w-32',
+  }
+
+  return (
+    <div className={cn('border rounded-md p-3', isOverridden && 'border-warning bg-warning/5')}>
+      <div className="flex items-center justify-between">
+        <div className="flex items-center gap-2 flex-1 min-w-0">
+          <span className="font-mono text-sm truncate">{flagName}</span>
+          {isOverridden && (
+            <Badge variant="warning" className="shrink-0">
+              Overridden
+            </Badge>
+          )}
+          {isNull && (
+            <Badge variant="secondary" className="shrink-0">
+              null
+            </Badge>
+          )}
+        </div>
+
+        {isNull ? (
+          <Input value="null" disabled className="w-32 opacity-50" />
+        ) : valueType === 'boolean' ? (
+          <Switch
+            checked={currentValue as boolean}
+            onCheckedChange={(checked) => onToggle(checked)}
+          />
+        ) : valueType === 'number' ? (
+          <Input type="number" {...inputProps} />
+        ) : (
+          <Input {...inputProps} />
+        )}
+      </div>
+
+      {isOverridden && (
+        <div className="mt-2 text-xs text-foreground-muted">
+          Original:{' '}
+          <code className="bg-surface-200 px-1 rounded">{JSON.stringify(originalValue)}</code>
+        </div>
+      )}
+    </div>
+  )
+}
+
+export function DevTelemetryToolbar() {
+  const { isEnabled, isOpen, setIsOpen, events, setEvents } = useDevTelemetryToolbar()
+  const [activeTab, setActiveTab] = useState<string>('events')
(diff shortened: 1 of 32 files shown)


apps/studio/components/ui/DevTelemetryToolbar/DevTelemetryToolbar.tsx
```diff
@@ -0,0 +1,448 @@
...
+    })
+
+    return unsubscribe
+  }, [isEnabled])
+
+  useEffect(() => {
+    if (!isEnabled || !isOpen) return
+
+    const sessionId = getCookie('session_id')
+    const url = `${API_URL}/telemetry/stream${sessionId ? `?session_id=${sessionId}` : ''}`
```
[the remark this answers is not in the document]
seanoliver: Good call! Done!

apps/studio/components/ui/DevTelemetryToolbar/DevTelemetryToolbar.tsx
```diff
@@ -0,0 +1,448 @@
...
+          title="Dev Telemetry Toolbar"
+        >
+          <Activity className="w-5 h-5" />
+          {events.length > 0 && (
+            <span className="absolute -top-1 -right-1 h-5 w-5 flex items-center justify-center rounded-full bg-destructive text-destructive-foreground text-xs">
+              {events.length > 99 ? '99+' : events.length}
+            </span>
+          )}
+        </button>
+      </div>
```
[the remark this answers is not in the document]
seanoliver: Good catch — added relative to the trigger button so the badge positions against the button, not the fixed container.

apps/studio/components/ui/DevTelemetryToolbar/DevTelemetryToolbar.tsx
```diff
@@ -0,0 +1,450 @@
...
+  }, [isEnabled, isOpen])
+
+  useEffect(() => {
+    const saved = getCookie('x-ph-flag-overrides')
+    if (saved) {
+      try {
+        setFlagOverrides(JSON.parse(saved))
+      } catch {}
+    }
+  }, [])
```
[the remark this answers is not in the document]
seanoliver: We’re keeping this silent for now to avoid extra console noise in local dev. If this causes confusion in practice, we can add a debug log later.

packages/dev-tools/DevToolbar.tsx
```diff
@@ -0,0 +1,448 @@
...
+            checked={currentValue as boolean}
+            onCheckedChange={(checked) => onToggle(checked)}
+          />
+        ) : (
+          <Input
+            value={String(currentValue)}
+            onChange={(e) => onToggle(e.target.value)}
+            className="w-32"
+          />
+        )}
```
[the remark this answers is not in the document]
seanoliver: Good point. For now we’re keeping overrides simple and string‑based for non‑booleans; this is local‑only and primarily for quick debugging. If numeric flags become common, we can add type‑aware parsing in a follow‑up.

apps/studio/pages/_app.tsx
```diff
@@ -190,6 +191,7 @@ function CustomApp({ Component, pageProps }: AppPropsWithLayout) {
                     </RouteValidationWrapper>
                   </TooltipProvider>
                   <Telemetry />
+                  <DevTelemetryToolbar />
```
ivasilov: I think it would be better if we do `{process.env.NEXT_PUBLIC_ENVIRONMENT === 'local' && <DevTelemetryToolbar />}`. Not sure if there's other issues with this approach.

seanoliver: Good call. I’ll add a gate on `NEXT_PUBLIC_ENVIRONMENT === 'local'` so it’s a guaranteed no-op in staging and prod.

charislam (commented):
Wild idea — can we move the entire thing to a `dev-tools` package for tree-shaking.

Basically make a new `packages/dev-tools`, set `{ side-effect: false }` in `package.json` for optimal tree-shaking, and then when `process.env.NODE_ENV !== 'development'`, we short-circuit the returned functions.

Happy to help on this if you want!

seanoliver:
> Wild idea — can we move the entire thing to a `dev-tools` package for tree-shaking.
> 
> Basically make a new `packages/dev-tools`, set `{ side-effect: false }` in `package.json` for optimal tree-shaking, and then when `process.env.NODE_ENV !== 'development'`, we short-circuit the returned functions.
> 
> Happy to help on this if you want!

Great idea! As we discussed via Slack, a shared package is also better because it allows us to put the toolbar on all of the frontend apps (including `docs` and `www` as well).

I've updated the PR to move this to a shared `pacakges/dev-tools` package with `{ side-effect: false }`, and integrated it into all of the frontend apps so we can use it to inspect feature flags and event logging everywhere instead of just in `studio`.

apps/studio/pages/_app.tsx
```diff
@@ -173,19 +173,22 @@ function CustomApp({ Component, pageProps }: AppPropsWithLayout) {
...
-                                {getLayout(<Component {...pageProps} />)}
-                              </MainScrollContainerProvider>
-                              <StudioCommandMenu />
-                              <FeaturePreviewModal />
-                            </FeaturePreviewContextProvider>
-                            <SonnerToaster position="top-right" />
-                            <MonacoThemeProvider />
-                          </CommandProvider>
-                        </AiAssistantStateContextProvider>
+                        <DevToolbarProvider apiUrl={API_URL}>
```
ivasilov, line 176: Do you actually need the provider here? It looks like the only place where you're using the context is in `DevToolbar` and `DevToolbarTrigger`. You can probably merge those two components together and just render them in the header components?

seanoliver: Good question. The main reason for keeping the provider is to preserve the event history and SSE connection across route changes, including pages that don't use the standard header where the `DevToolbarTrigger` is rendered, like sign-in and onboarding pages.

Since we will likely use this to debug auth and onboarding flows, I wanted to avoid losing accumulated events as you move through those screens.

That said, the upside here is pretty small, so I am totally open to collapsing it into a single component if you think the provider is overkill.

packages/dev-tools/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "main": "./index.ts",
+  "types": "./index.ts",
+  "exports": {
+    ".": "./index.ts"
+  },
+  "scripts": {
+    "preinstall": "npx only-allow pnpm",
+    "clean": "rimraf node_modules",
+    "test": "vitest run",
+    "test:watch": "vitest"
```
charislam: Probably want to include a typecheck so it's included in the global typecheck run

packages/dev-tools/DevToolbarContext.tsx
```diff
@@ -0,0 +1,223 @@
...
+        dismissToolbar,
+      }}
+    >
+      {children}
+    </DevToolbarContext.Provider>
+  )
+}
+
+export function useDevToolbar() {
+  if (!IS_LOCAL_DEV) {
```
charislam: This check needs to be moved outside of the core implementation so it can be tree-shaken

packages/dev-tools/DevToolbar.tsx
```diff
@@ -0,0 +1,445 @@
...
+  const handleKeyDown = (e: KeyboardEvent<HTMLDivElement>) => {
+    if (e.key === 'Enter' || e.key === ' ') {
+      e.preventDefault()
+      handleToggle()
+    }
+  }
+
+  return (
+    <div className="border rounded-md p-3 bg-surface-100">
+      <div
```
charislam: Would be more semantically correct and accessible to make this a button instead of a div, but since this is a dev-only feature, won't block on it

packages/dev-tools/DevToolbar.tsx
```diff
@@ -0,0 +1,445 @@
...
+              <Activity className="w-4 h-4" />
+              Events ({filteredEvents.length})
+            </TabsTrigger>
+            <TabsTrigger value="flags" className="flex items-center gap-2 px-4">
+              <Flag className="w-4 h-4" />
+              Flags {totalOverrideCount > 0 && `(${totalOverrideCount} overrides)`}
+            </TabsTrigger>
+          </TabsList>
+
+          {activeTab === 'events' && (
```
charislam: more semantically correct to have these wrapped in TabContent as documented in [Radix UI](https://www.radix-ui.com/primitives/docs/components/tabs#content) but again this is dev only so not going to block on it

charislam (commented):
I pushed a change to make sure the tree-shaking works (to test I just grep for strings that only show up in the Dev Tools UI in the final bundle).

I can't seem to make the feature flag override work — I tried overriding a flag, but it didn't change the UI that was dependent on that flag even after I reloaded

packages/dev-tools/DevToolbarContext.tsx
charislam: I pushed a fix for tree-shaking so this branch can just be deleted now

seanoliver: Nice, this is a clever pattern! I haven't run into it before but it's exactly what we need here. Thanks, Charis! 🙌  Stripped out the redundant `IS_LOCAL_DEV` check from `useDevToolbar`.

packages/dev-tools/package.json
seanoliver: Good call! Done!

packages/dev-tools/DevToolbar.tsx
seanoliver: You're right, swapped to a `<button>`.

packages/dev-tools/DevToolbar.tsx
seanoliver: Good call. Wrapped the panels in `TabsContent` to follow the Radix pattern properly.

packages/dev-tools/vitest.setup.ts
```diff
@@ -0,0 +1,25 @@
...
+    onchange: null,
+    addListener: vi.fn(), // deprecated
+    removeListener: vi.fn(), // deprecated
+    addEventListener: vi.fn(),
+    removeEventListener: vi.fn(),
+    dispatchEvent: vi.fn(),
+  })),
+})
+
+vi.mock('next/navigation', () => require('next-router-mock'))
```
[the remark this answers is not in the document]
seanoliver: Good catch! Updated to use the /navigation entry point so App Router hooks are mocked correctly.

packages/dev-tools/package.json
seanoliver: Ah actually I tried adding a typecheck script, but it fails on CI due to workspace dependency issues (`tsc` follows imports into `ui` which has CSS modules it can't resolve). 

The `common` package actually appears to have the same issue and its typecheck is commented out as `typecheck_CURRENTLY_IGNORED`: https://github.com/supabase/supabase/blob/20a89084277cd28a2b45a2614df606d706c6f070/packages/common/package.json#L12

So for now I left it out. LMK what you think! Happy to revisit if there's a known workaround!

packages/dev-tools/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  setIsOpen: () => {},
+  events: [],
+  setEvents: () => {},
+  dismissToolbar: () => {},
+}
+
+export const DevToolbarProvider =
+  process.env.NODE_ENV !== 'development'
+    ? ({ children }: { children: ReactNode }) => children
+    : DevToolbarContextModule.DevToolbarProvider
```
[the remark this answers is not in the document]
seanoliver: Good catch! Added `apiUrl` to the stub props so the types match across environments.

charislam (approved):

CoolAssPuppy (approved):

PR #42274, opened 2026-01-29
Update vector bucket SQL queries for wrappers ext 0.5.7

joshenlim:
## Context

There's some changes for Vector Buckets with the wrappers extension on version 0.5.7 and above

The "Query from Postgres" CTA (which runs an import foreign schema command) no longer accepts the bucket name as part of the schema options (ref [changelog](https://github.com/supabase/wrappers/releases/tag/v0.5.7)), and instead takes the bucket name as the remote schema name.

So changes here are to update the SQL for importing foreign schema for vector buckets if wrappers extension is above 0.5.7

Also - the `::embd` type no longer seems to exist in 0.5.7 as well, so updated the example code too

## To test

Can probably test on a new US East 1 project on the preview
- [ ] Verify creating a vector bucket table + can query from postgres (via the Insert vectors CTA after creating a table)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed breadcrumb navigation links for vector bucket pages.

* **Improvements**
  * Unified loading state for schema workflows and disabled actions while loading.
  * Enhanced version compatibility checks for foreign schema imports to support newer extension syntax.
  * Adjusted generated SQL for vector data to use the appropriate embedding syntax based on extension version.

### DIFF
--- a/apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
+++ b/apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
@@ -1,8 +1,4 @@
 import type { PostgresExtension } from '@supabase/postgres-meta'
-import { Database, ExternalLinkIcon, Plus } from 'lucide-react'
-import { useEffect, useState } from 'react'
-import { toast } from 'sonner'
-
 import { DocsButton } from 'components/ui/DocsButton'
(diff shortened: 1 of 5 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #42276, opened 2026-01-29
Chore/connect upgrade

SaxonF:
image

Adds a Connect Sheet that replaces the existing Connect Dialog . The Sheet is behind a feature flag and will need to be tested. As part of this new Sheet I have also refactored how fields are generated and content is rendered using a connect.schema file. This schema file defines modes, fields and steps. Each mode has a set of fields. Each field can be dependent on another field. The steps generated are then based off the values of the modes and fields. Each step can also render varying content dynamically using a template `{{framework}}/{{frameworkVariant}}/{{library}}` which just replaces those values with what's in state (fields selected).


Hallidayo:

PR #42281, opened 2026-01-29
Joshen/fe 2410 add default label to pg cron extension schema selector

joshenlim:
## Context

Addresses a UX confusion where users are unsure of which schema to install database extensions in. Am opting to show a "Default" badge on the extensions schema just to lower the cognitive load for this step if the extension doesn't have a specific schema it needs to be installed in

image

If the extension has a recommended schema, we'll show a badge as well (Related [PR](https://github.com/supabase/supabase/pull/40968))
image

Otherwise, if there's a fixed schema, then no badge
image

## Other changes
- Refactored `EnableExtensionModal` to use react hook form + Dialog component
- Refactored checking of default schema to use React Query

## To test

- [ ] Verify the UI changes
- [ ] Verify that you can still enable / disable extensions - and ensure that they're being installed in the right schemas

## Summary by CodeRabbit

* **New Features**
  * Enhanced extension enable dialog with form-driven schema selection, read-only default-schema display, and "create custom schema" flow
  * "Recommended" and "Default" badges next to schema choices; platform-specific warnings and callouts
  * Fetches and surfaces extension default/recommended schema to guide selection

* **Refactor**
  * Dialog-based UI with integrated form validation and improved enable workflow


apps/studio/data/sql/queries/get-extension-default-schema.ts
```diff
@@ -0,0 +1,7 @@
+export const getDatabaseExtensionDefaultSchemaSQL = ({ extension }: { extension: string }) => {
+  const sql = /* SQL */ `
+select name, version, schema from pg_available_extension_versions where name = '${extension}' limit 1;
+`.trim()
```
avallete, line 6: I'll recommend using the `literal` escape utils.

```suggestion
import { literal } from '@supabase/pg-meta/src/pg-format'
export const getDatabaseExtensionDefaultSchemaSQL = ({ extension }: { extension: string }) => {
  const sql = /* SQL */ `
select name, version, schema from pg_available_extension_versions where name = ${literal(extension)} limit 1;
`.trim()
```

apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
```diff
@@ -124,6 +117,8 @@ const EnableExtensionModal = ({ visible, extension, onCancel }: EnableExtensionM
           ? values.name
           : values.schema
 
+    console.log(extension.name, schema)
```
avallete:
```suggestion
```

apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx
```diff
@@ -50,7 +70,24 @@ const EnableExtensionModal = ({ visible, extension, onCancel }: EnableExtensionM
...
+        projectRef: project?.ref,
+        connectionString: project?.connectionString,
+        extension: extension.name,
+      },
+      { enabled: visible }
+    )
+  const defaultSchema = extensionMeta?.schema
+
+  const isLoading = fetchingSchemaInfo || isSchemasLoading
+  console.log(extension.name, { fetchingSchemaInfo, isSchemasLoading })
```
alaister: Sneaky one:
```suggestion
```

alaister (approved):
LGTM!

One missed console.log:

soedirgo (approved):

apps/studio/components/interfaces/Database/Extensions/Extensions.constants.ts
```diff
@@ -24,3 +24,8 @@ export const SEARCH_TERMS: Record<string, string[]> = {
 export const EXTENSION_DISABLE_WARNINGS: Record<string, string> = {
   pg_cron: 'Disabling this extension will delete all scheduled jobs. This cannot be undone.',
 }
+
+// Extensions that have recommended schemas (rather than required schemas)
+export const extensionsWithRecommendedSchemas: Record<string, string> = {
+  wrappers: 'extensions',
```
soedirgo, line 30: We should use `pg_catalog` for pg_cron here per the discussion:

```suggestion
  pg_cron: 'pg_catalog',
  wrappers: 'extensions',
```

joshenlim: gotcha on this - opting to update this differently [here](https://github.com/supabase/supabase/blob/joshen/fe-2410-add-default-label-to-pg_cron-extension-schema-selector/apps/studio/components/interfaces/Database/Extensions/EnableExtensionModal.tsx#L92) instead

for context
- given that pg_cron is automatically enforced to be installed in pg_catalog irregardless of schema selection, this should then be a "default" schema, rather than a recommendation
- by default we check the "default" schema of an extension via `pg_available_extension_versions` which i understand returns `schema` as null for pg_cron
- so just hardcoding `defaultSchema` for pg_cron here (we may remove the schema selection entirely as well eventually so this should do for now)

soedirgo: Gotcha 👍

PR #42283, opened 2026-01-29
Fix queue operations stale state + refactor to reuse existing methods

joshenlim:
## Context

Related to this PR [here](https://github.com/supabase/supabase/pull/42265)

If you were to update a cell twice, then save, the value that get's saved is the first edited value.

There's a number of changes in this PR but the main thing is just `useOperationQueueActions` to declare `operations` outside of `handleSave` since `snap` wouldn't trigger the re-render of that function in the dependency array.

## Other changes

- Refactored `useOperationQueueShortcuts.ts`
  - Reuse an existing `useHotKey` method to handle the event listeners
  - Remove unnecessary prop drilling, call the necessary hooks within it instead
  - Simplify its calling by only calling this hook within `SaveQueueActionBar` (removed calling from `OperationQueueSidePanel`
- Add tooltip for OperatioItem "Remove operation"
- (Unrelated) Fix Table Editor sort "Apply sort" CTA not working for large tables when applying sort on an indexed column (e.g primary key)

## Release Notes

* **New Features**
  * Operation queue view now toggles open and closed with a single action
  * Simplified keyboard shortcuts: press **S** to save operations, **.** to toggle the queue view

* **Bug Fixes**
  * Sort operations on large tables with non-primary-key sorting now show a confirmation warning and require explicit user approval before applying

* **Improvements**
  * Operation queue panel automatically closes after successfully saving operations
  * Enhanced UI with improved delete action feedback and refined visual separators

### DIFF
--- a/apps/studio/components/grid/hooks/useOperationQueueActions.ts
+++ b/apps/studio/components/grid/hooks/useOperationQueueActions.ts
@@ -1,10 +1,9 @@
 import { useQueryClient } from '@tanstack/react-query'
-import { useCallback } from 'react'
-import { toast } from 'sonner'
-
 import { tableRowKeys } from 'data/table-rows/keys'
 import { useOperationQueueSaveMutation } from 'data/table-rows/operation-queue-save-mutation'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
+import { useCallback } from 'react'
+import { toast } from 'sonner'
 import { useGetImpersonatedRoleState } from 'state/role-impersonation-state'
 import { useTableEditorStateSnapshot } from 'state/table-editor'
 import { QueuedOperation } from 'state/table-editor-operation-queue.types'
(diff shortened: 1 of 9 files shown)


apps/studio/components/grid/hooks/useOperationQueueActions.ts
```diff
@@ -40,12 +40,10 @@ export function useOperationQueueActions(options: UseOperationQueueActionsOption
     })
 
   const isSaving = snap.operationQueue.status === 'saving' || isMutationPending
+  const operations = snap.operationQueue.operations as readonly QueuedOperation[]
```
alaister, line 43: @awaseem this is the main fix if you're interested. The previous code wasn't causing `snap.operationQueue.operations` to be tracked correctly with valtio

alaister (approved):
LGTM!

apps/studio/components/grid/hooks/useOperationQueueActions.ts
awaseem: interesting...

awaseem:
Nice clean up 👨‍🍳

PR #42288, opened 2026-01-29
Feat: Insert and delete rows for batch operations on table editor

awaseem:

YES

Completion of batch edits on the table editor

## Demo

https://github.com/user-attachments/assets/ab5a7112-3dcc-456a-a5fc-1c9a99fccf34

## Summary by CodeRabbit

* **New Features**
  * Queued add/edit/delete operations with optimistic UI, conflict resolution, and queue-based flows
  * Side-panel items showing queued add/delete row previews

* **UI**
  * Pending-add placeholders plus a visible "DEFAULT" marker in grid cells
  * Visual row states: green for pending adds, red with strike-through for pending deletes
  * Queue-based deletes can bypass confirmation when queue mode is enabled

* **Tests**
  * Expanded tests covering queue conflict resolution and queue utilities


apps/studio/components/grid/types/table.ts
```diff
@@ -34,3 +35,25 @@ export interface SupaTable {
...
+export type PendingDeleteRow = SupaRow & PendingDeleteMarker
+
+export function isPendingAddRow(row: SupaRow): row is PendingAddRow {
+  return '__tempId' in row && typeof (row as PendingAddRow).__tempId === 'string'
+}
+
+/** Check if row is pending deletion (has __isDeleted marker) */
+export function isPendingDeleteRow(row: SupaRow): row is PendingDeleteRow {
+  return '__isDeleted' in row && (row as PendingDeleteRow).__isDeleted === true
+}
```
awaseem, line 59: These are different than the operation, these are the actual rows being rendered on the table. We add some values to them to know that they are essentially fake and need some UI applied to them. They also come in handy when applying conflict resolution later on

apps/studio/components/grid/utils/queueConflictResolution.ts
```diff
@@ -0,0 +1,175 @@
+import { isPendingAddRow } from '../types'
```
awaseem, line 1: a lot of this logic was going to live in the table editor state. Decided to refactor and test because its a bit crazy due to all the different edges cases with editing fake rows

apps/studio/components/grid/utils/queueOperationUtils.ts
```diff
@@ -1,49 +1,55 @@
...
-  columnName?: string
-  rowIdentifiers?: Record<string, unknown>
+  payload: {
+    columnName: string
+    rowIdentifiers: Dictionary<unknown>
+  }
 }
 
-export function generateTableChangeKeyFromOperation(operation: NewQueuedOperation): string {
+export function generateTableChangeKey(
```
awaseem, line 25: Cleaned this up a ton, no need for multiple methods now that the types are proper unions. When you check for the `type` it'll automatically turn the payload into the right value

apps/studio/components/grid/utils/queueOperationUtils.ts
```diff
@@ -1,49 +1,55 @@
...
+    const { rowIdentifiers } = operation.payload
+    const rowIdentifiersKey = Object.entries(rowIdentifiers)
+      .sort(([a], [b]) => a.localeCompare(b))
+      .map(([key, value]) => `${key}:${value}`)
+      .join('|')
+    return `${operation.type}:${operation.tableId}:${rowIdentifiersKey}`
+  }
+
+  // Exhaustive check - TypeScript will error if we miss a case
+  const _exhaustiveCheck: never = operation
```
awaseem, line 51: Now that all the types need to be checked, we should never run into these cases

apps/studio/state/table-editor-operation-queue.types.ts
```diff
@@ -24,23 +23,72 @@ export interface EditCellContentPayload {
...
+
+export interface NewDeleteRowOperation extends NewQueuedOperationBase {
+  type: QueuedOperationType.DELETE_ROW
+  payload: DeleteRowPayload
+}
+
+export type NewQueuedOperation =
+  | NewEditCellContentOperation
+  | NewAddRowOperation
+  | NewDeleteRowOperation
```
awaseem, line 92: This is by far the best change, it doesn't change the structure of the original types but makes them proper unions. So when you check for a type and it satisfies that constraint it'll the payload is the correct value. No more fake casting something as X type

apps/studio/components/grid/utils/queueOperationUtils.ts
```diff
@@ -120,6 +177,52 @@ export function queueCellEditWithOptimisticUpdate({
...
+  queueOperation,
+  projectRef,
+  tableId,
+  table,
+  rowData,
+  enumArrayColumns,
+}: QueueRowAddParams) {
+  // Generate unique idx and tempId for this pending row
+  const idx = -Date.now()
+  const tempId = String(idx)
```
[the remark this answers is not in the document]
awaseem: If that becomes the case we can change it, no need to move it now

soedirgo (approved):

alaister (approved):
LGTM!

- Tested on preview

PR #42293, opened 2026-01-29
fix(studio): copy schema behavior on safari

charislam:

YES

Bug fix

The clipboard copy operation for table schema would show the success toast before the copy actually completed.

## What is the new behavior?

Move the toast showing to the callback so it appears only after the copy finishes

## Summary by CodeRabbit

* **Bug Fixes**
  * Clipboard success confirmation now appears reliably after the copy completes.

* **New Features**
  * Copied table schemas are automatically formatted for improved readability.

* **Improvements**
  * Enhanced clipboard behavior for better reliability across browsers.

### DIFF
--- a/packages/ui/src/lib/utils/clipboard.ts
+++ b/packages/ui/src/lib/utils/clipboard.ts
@@ -22,17 +22,13 @@ export const copyToClipboard = async (str: string | Promise<string>, callback = 
       const text = new ClipboardItem({
         'text/plain': Promise.resolve(str).then((text) => new Blob([text], { type: 'text/plain' })),
       })
-      // [Joshen] Adding a timeout based on this comment here about a workaround
-      // https://stackoverflow.com/questions/62327358/javascript-clipboard-api-safari-ios-notallowederror-message
-      setTimeout(() => {
-        navigator.clipboard.write([text]).then(callback)
-      }, 0)
+      return navigator.clipboard.write([text]).then(callback)
     } else {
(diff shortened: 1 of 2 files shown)


packages/ui/src/lib/utils/clipboard.ts
```diff
@@ -22,11 +22,7 @@ export const copyToClipboard = async (str: string | Promise<string>, callback =
       const text = new ClipboardItem({
         'text/plain': Promise.resolve(str).then((text) => new Blob([text], { type: 'text/plain' })),
       })
-      // [Joshen] Adding a timeout based on this comment here about a workaround
-      // https://stackoverflow.com/questions/62327358/javascript-clipboard-api-safari-ios-notallowederror-message
-      setTimeout(() => {
-        navigator.clipboard.write([text]).then(callback)
-      }, 0)
+      navigator.clipboard.write([text]).then(callback)
```
charislam: Tested on Safari and seems like the `setTimeout` might not be needed? The `ClipboardItem` thing seems to solve it, and the timeout makes the behavior with awaits unexpected

jordienr (approved):

dnywh (approved):
Tested on staging on both Chrome and Safari. Works great on Chrome. Safari now flashes an error toast if focus is blurred before the copy is complete, but weirdly only if I press a key on the keyboard.

Approving because it’s still a big improvement on what’s currently in prod!

PR #42306, opened 2026-01-30
Set isStatementTimeoutDisabled for EditorPanel

### LINKED ISSUE #42304: Menu bar / right panel sql editor ---> errors out on 'vacuum analyze' and 'vacuum full'
however when I run these commands from the dedicated SQL Editor page >> success

the error i persistently receive from the Menu bar / right panel sql editor: "Error running SQL query
Failed to run sql query: ERROR:  25001: VACUUM cannot run inside a transaction block"

Image

versus:

Image

joshenlim:
## Context

Resolves https://github.com/supabase/supabase/issues/42304

Realised that the Inline Editor behaves differently from the SQL Editor in which [`isStatementTimeoutDisabled`](https://github.com/supabase/supabase/pull/36367) is not set to true for the former which hence runs queries within a transaction.

PR here sets that to true which should align the behaviour between both editors

IMO though - the long term fix is just consolidating the editors so that the behaviours are the same

## Summary by CodeRabbit

* **New Features**
  * SQL statements in the editor can now execute without timeout restrictions, allowing long-running queries to complete successfully.

### DIFF
--- a/apps/studio/components/ui/EditorPanel/EditorPanel.tsx
+++ b/apps/studio/components/ui/EditorPanel/EditorPanel.tsx
@@ -1,7 +1,3 @@
-import { Book, Maximize2, X } from 'lucide-react'
-import { useRouter } from 'next/router'
-import { useState } from 'react'
-
 import { LOCAL_STORAGE_KEYS, useParams } from 'common'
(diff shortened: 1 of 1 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #42307, opened 2026-01-30
Joshen/fe 2441 backup restore dialog shows incorrect utc timestamp

joshenlim:
## Context

Clicking "Restore" on a scheduled backup shows the confirmation modal, in which the backup date does not align with what's shown on the back up row (the latter is the correct one)

image

image

Am opting to use the same `TimestampInfo` component in the confirmation modal - which will align the values shown in both UI

image


alaister (approved):
LGTM!

- Tested on preview

PR #42319, opened 2026-01-30
blog: fixing ordering and removing fallbacks

stylessh:

YES


apps/www/app/api-v2/cms-posts/route.ts
```diff
@@ -713,7 +714,7 @@ export async function GET(request: NextRequest) {
           formattedDate,
           readingTime,
           authors,
-          imgThumb: imgThumbUrl || imgSocialUrl || '',
+          imgThumb: imgThumbUrl || '',
```
dnywh: The 	`imgSocialUrl` was a fallback, no? Or does it not matter since this CMS is going to be deprecated?

apps/www/app/blog/[slug]/page.tsx
```diff
@@ -152,9 +151,9 @@ export async function generateMetadata({ params }: { params: Promise<Params> }):
     }
   }
 
-  // Fallback to imgThumb or imgSocial if no meta image
+  // Only use imgSocial for OG image, no fallback to imgThumb
```
Comment (dnywh) [inline apps/www/app/blog/[slug]/page.tsx]: Similar to above: why not keep the fallback?

dnywh:
Noticed a few things that could be improved so closed this in favour of #43842

PR #42331, opened 2026-01-31
docs: update Refine tutorial to use getClaims for auth check

ghost:
📝 Summary
This PR updates the "Build a User Management App with Refine" tutorial to replace the unrecommended auth.getSession() method with auth.getClaims().

🚩 Problem
The current documentation uses getSession() for authentication guards. As noted in the Supabase Auth security guidelines, getSession() retrieves data directly from local storage without verifying the JWT signature. This can lead to security vulnerabilities where a user could potentially bypass frontend guards by manipulating local storage data.

✅ Improvement
I have updated the check method in the authProvider to use auth.getClaims().

Security: getClaims() verifies the JWT signature against the project's public keys (JWKS), ensuring the session is authentic.

Performance: Unlike getUser(), which always performs a network request to the Auth database, getClaims() uses a cached key set, making it ideal for the high-frequency "gatekeeper" checks required by Refine.

🛠️ Changes
Updated src/authProvider.ts in the with-refine.mdx file.

Replaced supabaseClient.auth.getSession() with supabaseClient.auth.getClaims().

Added error handling to catch invalid or expired claims.

🧪 Testing
Verified that the logic correctly identifies an active session.

Confirmed that removing the sb-access-token from local storage correctly triggers a redirect to /login.

YES/NO

## What is the new behavior?

## Additional context


ghost (commented):
Nice improvement to the tutorial and good alignment with the current Supabase auth guidance.

Switching from getSession() to getClaims() makes the example safer and more realistic for production-style guard checks. The explanation around signature verification and the JWKS cache is clear and helps readers understand why this approach is preferable for frequent auth validation in Refine.

The added error handling for invalid or expired claims is also important, since it prevents silent failures and makes the redirect-to-login behavior explicit.

PR #42338, opened 2026-01-31
docs(realtime): add note about sync event behavior in Presence

### LINKED ISSUE #41175: Enhance presence sync event documentation
# Improve documentation

## Link 

https://supabase.com/docs/guides/realtime/presence

## Describe the problem

When using Realtime Presence, I noticed that during a sync event, the client receives join and leave events at the same time even though no actual users are joining or leaving.

This looked like unexpected behavior, and it took a bit of digging to realize that this is intentional and happens because Presence reconciles its local state with the server.

Since this can be confusing for people who are new to Presence, it might be helpful to mention this explicitly in the docs.

## Describe the improvement

A small note in the Presence section explaining that:
	• sync emits leave / join
	• this is normal and expected
	• it reflects state reconciliation, not real user movement

Thanks!

## Additional context

https://github.com/orgs/supabase/discussions/26748?utm_source=chatgpt.com#:~:text=%ED%95%9C%EA%B5%AD%EC%96%B4%EB%A1%9C%20%EB%B2%88%EC%97%AD-,I%20figured%20out%20that%20whenever%20the%20presence%20is%20updated%2C%20a%20leave%20event%20and%20a%20join%20event%20happen%20at%20the%20same%20time.%20This%20is%20expected%20behavior%20even%20though%20it%20is%20a%20bit%20confusing,-.%20You%20should%20process

jonathan-fulton:

Documentation improvement for Realtime Presence.

The Presence documentation doesn't explain that during a `sync` event, clients may receive `join` and `leave` events simultaneously even when no users are actually joining or leaving. This can be confusing for developers new to Presence.

## What is the new behavior?

Added an explanatory note clarifying that:
- During `sync`, you may receive `join` and `leave` events at the same time
- This is normal and expected
- It reflects state reconciliation with the server, not real user movement

This helps developers understand this behavior upfront rather than having to discover it through debugging.

## Additional context

As mentioned in the issue, this behavior is discussed in [this community discussion](https://github.com/orgs/supabase/discussions/26748) where the solution had to be discovered by users.

Fixes #41175

## Summary by CodeRabbit

* **Documentation**
  * Clarified behavior during sync events in the Realtime Presence guide. Added explanation that simultaneous join and leave events may occur due to state reconciliation rather than actual user movement changes.


ChrisChinchilla (approved):
Thanks @jonathan-fulton

PR #42340, opened 2026-01-31
fix(studio): support pg_cron $ syntax for last day of month in schedule parsing

jonathan-fulton:
Bug fix

Cron jobs using pg_cron's `$` syntax (representing 'last day of month') show 'Unable to parse next run for job' in the dashboard's Next run column, even though these are valid pg_cron schedules.

## What is the new behavior?
The dashboard now correctly parses schedules using `$` and displays the proper next run time.

## Root Cause
pg_cron uses `$` for 'last day of month', but the `cron-parser` library used by Studio uses `L` for the same purpose. The `$` character was causing the parsing to fail.

## Fix
Normalize pg_cron's `$` syntax to cron-parser's `L` syntax before parsing. Both represent 'last day of month' - it's just a syntax difference between the two systems.

Fixes #42176

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cron job scheduling to correctly calculate next-run times for pg_cron-compatible schedule expressions.


charislam (approved):
Thanks @jonathan-fulton!

PR #42342, opened 2026-01-31
fix(studio): improve column type visibility in table editor

jonathan-fulton:
Bug fix

The column type dropdown in the table editor is always truncated, making it impossible to differentiate between similar types like 'timestamp' and 'timestamptz'. The placeholder text 'Choose a column type...' is also truncated.

## What is the new behavior?
- Column types are more visible with increased column width (25% → 28%)
- Full type name is shown on hover via title attribute
- Text overflow is handled gracefully with proper truncation

## Changes Made
- Added `title` attribute to the column type button for full text visibility on hover
- Restructured button content with proper flex layout and `min-w-0`
- Added `shrink-0` to icon and `truncate` to text for proper overflow handling
- Increased type column width from 25% to 28%
- Adjusted default value column width accordingly (25%/30% → 22%/27%)

Fixes #41579


djhi:
Closed in favor of #43630. Thanks!

PR #42343, opened 2026-01-31
fix(studio): prevent entity dropdown from being disabled on empty search results

jonathan-fulton:
Bug fix

1. When creating an index, if you search for entities that don't exist, the entity dropdown gets disabled - even though the schema has tables
2. The schema and entity dropdowns share the same search term, so typing in one affects the other

## What is the new behavior?
1. Entity dropdown only gets disabled when the schema truly has no tables
2. Each dropdown has independent search state

## Changes Made
- Separated `schemaSearchTerm` and `tableSearchTerm` state variables
- Added `allEntities` query to check if schema has any tables (unfiltered by search)
- Updated `isSelectEntityDisabled` to use `allEntityTypes` instead of filtered `entityTypes`
- Reset `tableSearchTerm` when schema changes
- Fixed schema dropdown placeholder to say 'Find schema...' instead of 'Find table...'

Fixes #41537


7ttp (commented):
hi @jonathan-fulton 
unfotunately already solved in: https://github.com/supabase/supabase/pull/43906

closing this up!
appreciate ur efforts towards contributing to Supabase 💚😁

PR #42349, opened 2026-01-31
feat(studio-local): functions management api - function blob artifacts

kallebysantos:

Feature

Functions page on self-hosted differs from Platform

## What is the new behavior?

> [!NOTE]  
> This PR only add readonly operations. Function edit and deploy should be implemented in a future one.

Adds the possibility to download and see function code in Self-Host version.

<details>

image

</details>

## Summary by CodeRabbit

* **New Features**
  * API endpoint to stream function files/artifacts as multipart/form-data.
  * New function file entry type and server-side file listing for functions.

* **Improvements**
  * Edge Functions "Code" navigation item always visible.
  * Download popover reworked: ZIP download always available; CLI section shown only on supported platforms.
  * Editor set to read-only and file actions disabled on unsupported environments.

* **Editor**
  * Added JavaScript, TypeScript, and Markdown language modules for the embedded editor.


apps/studio/pages/api/v1/projects/[ref]/functions/[slug]/body.ts
```diff
@@ -0,0 +1,57 @@
...
+  })
+
+  const multipartResponse = new Response(form)
+
+  res.setHeader(
+    'Content-Type',
+    multipartResponse.headers.get('content-type') ?? 'multipart/form-data'
+  )
+
+  return res.status(200).send(await multipartResponse.text())
```
Comment (charislam) [inline apps/studio/pages/api/v1/projects/[ref]/functions/[slug]/body.ts]: This loads everything into memory which could be significant, let's try streaming instead

apps/studio/lib/api/self-hosted/functions/fileSystemStore.ts
```diff
@@ -29,6 +29,24 @@ export class FileSystemFunctionsArtifactStore {
 
     return parseFolderToFunctionArtifact(functionFolder)
   }
+
+  async getBlobArtifactsBySlug(slug: string): Promise<FunctionBlobArtifact[]> {
+    if (slug === 'main') return []
+
+    const functionFolderPath = path.join(this.folderPath, slug)
```
charislam: possible path traversal here?

charislam (approved):

PR #42381, opened 2026-02-02
fix(studio): prevent sidebar text overlap on project cards

7ttp:
## Problem
When hovering over the sidebar on the projects page, project card text (names, regions, status) was rendering on top of the expanded sidebar, 
making both the sidebar and project information difficult to read.

## Solution
Increased sidebar z-index from `z-10` to `z-50` to ensure the sidebar renders above all project card content.

<details>
<summary>Before</summary>
image

</details>

<details>
<summary>After</summary>
image

</details>

## Related
- Closes https://github.com/supabase/supabase/issues/42376


stylessh:
hey! let's change the z-50 in the instance and not the original component itself as we might mess with other layers in other parts

7ttp:
> hey! let's change the z-50 in the instance and not the original component itself as we might mess with other layers in other parts

done 💚🙌

djhi (approved):
Thanks!

PR #42389, opened 2026-02-02
chore(ui-patterns): reset command menu after selection

fsansalvadore:
## What is the new behavior?

Reset the command menu state after a selection has been made (either action or route).

https://github.com/user-attachments/assets/d17b4780-17b4-4e1e-948c-fa8c50236ae3

## To test

- Open command menu
- navigate to and select any final command action/route (so not if you select a command menu sub-page - eg. "Create...")
- reopening the command menu you should always see the initial command menu state

## Summary by CodeRabbit

* **Bug Fixes**
  * Command menu now automatically resets after copying API keys and URLs, and when navigating between pages, ensuring a clean state for repeated use.


awaseem (approved):
Tested on staging

PR #42396, opened 2026-02-02
fix: AI filter re-render by seperating URL logic syncing and local state

awaseem:

YES

For the new AI filters we should take a totally different approach. Rather than using the URL as the source of truth, it should rather act as a secondary state store. The first being local state in Snap that gets loaded from the URL. By forcing this clear line we can always make sure the URL is backed up with the right values and ensure state is updated quickly. 

## Summary by CodeRabbit

* **New Features**
  * Toggleable next‑gen filter bar with lifecycle that initializes and syncs filters to the URL.
  * Centralized filter API in table state for setting, clearing, and sharing filters across components.

* **Improvements**
  * Grid queries are project‑scoped, use placeholder previous data, and reapply optimistic updates on refresh.
  * Added MS SQL sorting warning and feature‑flag controls for queue/operation behavior.

## Testing
- Make sure the old filters still work as expected
- Enable the new Headers, test out new filters approach (count is still broken FYI, will fix that later)


alaister (approved):
LGTM!

- Tested on preview

PR #42408, opened 2026-02-03
docs: fix inconsistent key naming in schedule-functions

zent7x:
## Summary
Fixed inconsistent key naming in the Edge Functions scheduling documentation.

## Problem
The vault setup example stored the key with the name `publishable_key`:
```sql
select vault.create_secret('YOUR_SUPABASE_PUBLISHABLE_KEY', 'publishable_key');
```

But the cron job example retrieved it using `anon_key`:
```sql
(select decrypted_secret from vault.decrypted_secrets where name = 'anon_key')
```

## Solution
Updated the vault setup to use `anon_key` consistently:
```sql
select vault.create_secret('YOUR_SUPABASE_ANON_KEY', 'anon_key');
```

Fixes #42398

## Documentation
- Updated Edge Function integration guide to clarify the correct authentication key required for API request authorization headers.

### DIFF
--- a/apps/docs/content/guides/functions/schedule-functions.mdx
+++ b/apps/docs/content/guides/functions/schedule-functions.mdx
@@ -31,7 +31,7 @@ Store `project_url` and `anon_key` in Supabase Vault:
 
 ```sql
 select vault.create_secret('https://project-ref.supabase.co', 'project_url');
-select vault.create_secret('YOUR_SUPABASE_PUBLISHABLE_KEY', 'publishable_key');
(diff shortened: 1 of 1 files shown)


mintydev789:
I believe the anon key is indeed the correct one, based on the fact that it's used when calling edge functions normally.

mintydev789:
Wait, now that I think about it, isn't this supposed to be service_role_key? Cause anon_key is public, right? We probably don't want it to be possible for anyone to trigger functions.

Issue #42323, opened 2026-01-30
[Auth] URLs reset in non-production branches


r-bart:
# Bug report

  

  - [x] I confirm this is a bug with Supabase, not with my own application.
  - [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub
  [Discussions](https://github.com/supabase/supabase/discussions), and
  [Discord](https://discord.supabase.com).

  ## Describe the bug

  Email confirmation redirect URLs configured in Authentication > URL Configuration get reset
  unexpectedly on preview branches. After configuring custom redirect URLs, they revert to
  `localhost:3000`, causing confirmation emails to contain incorrect redirect links.

  This issue occurs specifically on preview branches (not the main production branch).

  **Note:** This bug is difficult to reproduce on demand. The reset happens intermittently and
   I haven't been able to identify the exact trigger. It could be related to branch syncing,
  deployments, or some scheduled background process.

  ## To Reproduce

  1. Create a preview branch in Supabase Dashboard
  2. Go to Authentication > URL Configuration
  3. Set custom Site URL and Redirect URLs (e.g., `https://my-preview-app.vercel.app`)
  4. Save the configuration
  5. Wait some time (the reset timing is unpredictable — could be hours or days)
  6. Check the URL Configuration again — values have reverted to `localhost:3000`
  7. Trigger a confirmation email — the redirect URL in the email points to `localhost:3000`
  instead of the configured URL

  **Important:** The exact conditions that trigger the reset are unknown. The issue has been
  observed multiple times but not consistently reproducible.

  ## Expected behavior

  Once configured, the Site URL and Redirect URLs should persist indefinitely and not reset to
   default values. Confirmation emails should always use the configured redirect URLs.

  ## Screenshots

  ## System information

  - OS: macOS
  - Browser: Chrome
  - Version of supabase-js: 2.90.1
  - Version of Node.js: 22.13.1
  - Supabase Project: Preview branch (not main/production)

  ## Additional context

  - This does NOT occur on the main production branch — only on preview branches
  - The reset seems to happen silently without any user action or notification
  - This breaks the email confirmation flow for preview/staging environments, making it
  difficult to test auth flows properly
  - Multiple occurrences have been observed, but the trigger remains unclear

Siddx6:
I'm experiencing this same issue. Happy to help investigate and fix it.
Questions that might help diagnose:

Does it happen after deployments/commits to the preview branch?
Related to branch syncing from production?
[...]

PR #42441, opened 2026-02-04
fix(studio): prevent auth settings from submitting stale site URL

abhisek343:

Bug fix.

`BasicAuthSettingsForm` and `ProtectionAuthSettingsForm` load `SITE_URL` into form state even though neither form renders or owns that setting. Saving an unrelated auth setting can therefore include the previously fetched `SITE_URL` in the update payload and overwrite a newer Site URL value.

This is a narrow dashboard form-state bug. It may contribute to symptoms reported in #42323, but this PR does not claim to explain every preview-branch URL reset described there.

## What is the new behavior?

- Remove `SITE_URL` from the schemas, default values, and reset state of the User Signups and Bot and Abuse Protection forms.
- Keep Site URL updates scoped to the UI that actually manages Site URL.
- Add regression coverage that starts with an auth config containing `SITE_URL`, changes an unrelated setting in each form, submits it, and verifies the mutation payload does not contain `SITE_URL`.

## Why

A form should only submit settings it owns. Excluding `SITE_URL` from unrelated forms prevents stale form state from overwriting a newer value saved elsewhere.

## Validation

Regression tests were added for both affected forms. CI results are pending after refreshing this branch onto current `master`.

Related to #42323.

## Summary by CodeRabbit

- **Bug Fixes**
  - Authentication settings no longer require or include the `SITE_URL` field when configuring signup and CAPTCHA options.
  - Other authentication configuration fields continue to work as expected.
  - Updates now submit only the relevant authentication settings, preventing unrelated site URL values from being included.

- **Tests**
  - Added coverage to verify that authentication updates submit only the intended settings.


abhisek343:
Refreshed this PR onto current `master` and narrowed the scope to the form-state bug: these two settings forms should not submit `SITE_URL`, which they do not own. I also added regression coverage for both forms that starts with `SITE_URL` present in the fetched auth config and verifies it is absent from the mutation payload after saving an unrelated setting.

The branch is now a single commit and conflict-free. The Supabase Actions workflows are currently showing `action_required` with no jobs started, so they appear to be waiting for maintainer approval before CI can run.

PR #42445, opened 2026-02-04
fix(docs): pass headers to redirect in Server OAuth example

### LINKED ISSUE #30900: Subtle bug in Remix guide
# Improve documentation

## Link 

 https://supabase.com/docs/guides/auth/social-login/auth-github?queryGroups=environment&environment=server&queryGroups=framework&framework=remix

## Describe the problem

This code is wrong:
```ts
  provider,
  options: {
    redirectTo: 'http://example.com/auth/callback',
  },
})

if (data.url) {
  redirect(data.url) // use the redirect API for your server framework
}
```
It causes a bug where the cookie is not set and the auth does not work

## Describe the improvement

```ts
const headers = new Headers();

 const supabase = createServerClient(process.env.SUPABASE_URL!, process.env.SUPABASE_API_KEY!, { 
	cookies: {
		getAll() {
			return parseCookieHeader(request.headers.get("Cookie") ?? "")
		},
		setAll(cookiesToSet) {
			for (const { name, value, options } of cookiesToSet) {
				headers.append("Set-Cookie", serializeCookieHeader(name, value, options))
			}
		},
	},
})
  provider,
  options: {
    redirectTo: 'http://example.com/auth/callback',
  },
})

if (data.url) { 
  redirect(data.url, { headers }) // use the redirect API for your server framework
}
```

## Additional context

The previous docs cause a subtle and hard to catch bug that cost my employee 2 days of work, and me 2 hours to figure out. It doesn't append the cookie headers to the redirect and doesn't set the cookies in the browser making the subsequent request to the callback url NOT have the verifier code or whatever it's called and making supabase throw an error

Rayyan-Oumlil:
Fixes #30900

In the **Server** tab of the OAuth PKCE flow docs, the example now uses `createServerClient` from `@supabase/ssr` with cookie handlers that write to a `Headers` object, and passes `redirect(data.url, { headers })` so cookies are set and auth works after redirect (fixes the subtle Remix/auth bug where cookies were not sent).

## Summary by CodeRabbit

* **Documentation**
  * Updated OAuth PKCE flow docs to demonstrate server-side client initialization and header-based cookie preservation across redirects.
  * Revised framework-specific examples (Next.js, SvelteKit, Astro, Remix, Express) to show how to read incoming cookies and emit Set-Cookie headers so session state is preserved during OAuth redirects.

### DIFF
--- a/apps/docs/content/_partials/oauth_pkce_flow.mdx
+++ b/apps/docs/content/_partials/oauth_pkce_flow.mdx
@@ -29,15 +29,32 @@ await supabase.auth.signInWithOAuth({
 
 <TabPanel id="server" label="Server">
 
-In the server, you need to handle the redirect to the OAuth provider's authentication endpoint. The `signInWithOAuth` method returns the endpoint URL, which you can redirect to.
+In the server, you need to handle the redirect to the OAuth provider's authentication endpoint. The `signInWithOAuth` method returns the endpoint URL, which you can redirect to. You must use a server client with cookie handlers that write to a `Headers` object, and pass those headers to your framework's redirect so that cookies are set in the browser.
 
 ```js
-import { createClient, type Provider } from '@supabase/supabase-js'
-const supabase = createClient('url', 'anonKey')
-const provider = 'provider' as Provider
-const redirect = (url: string) => {}
+import { createServerClient, parseCookieHeader, serializeCookieHeader } from '@supabase/ssr'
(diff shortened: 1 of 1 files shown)


aantti:
@Rayyan-Oumlil, landing this in #50722, with you as co-author - your `as string` catch in the SvelteKit callback block was still on master and is fixed there. Thanks.

PR #42446, opened 2026-02-04
chore(studio): update Cursor rule to use `queryOptions`

ivasilov:
## Summary

- Updates the Cursor rule at `.cursor/rules/studio/queries/RULE.md` to recommend the `queryOptions` pattern from TanStack React Query instead of custom `useXQuery` wrapper hooks
- Provides an example implementation by refactoring `thirdPartyAuthIntegrationsQuery` from a `useThirdPartyAuthIntegrationsQuery` hook to `thirdPartyAuthIntegrationsQueryOptions`

## Why this pattern?

The `queryOptions` factory pattern from TanStack Query v5 offers several benefits:

1. **Type safety** - Query keys and return types are properly inferred
2. **Reusability** - The same query config can be used with both `useQuery()` in components and `queryClient.fetchQuery()` for imperative fetching
3. **Consistency** - Aligns with TanStack Query's official recommendations
4. **Simplicity** - Removes the need for custom generic wrapper hooks

## Test plan

- [ ] Verify the ThirdPartyAuthForm component works correctly with the new pattern
- [ ] Check that the Cursor rule provides clear guidance for writing new queries

## Summary by CodeRabbit

* **Documentation**
  * Updated query-pattern guidance and examples to the new query-options surface; adjusted mutation and component usage wording.

* **Refactor**
  * Switched third‑party auth integrations data fetching to the new query-options approach and updated component usage accordingly.

* **Chores**
  * Marked a legacy React Query helper as deprecated in types.

### DIFF
--- a/apps/studio/data/third-party-auth/integrations-query.ts
+++ b/apps/studio/data/third-party-auth/integrations-query.ts
@@ -1,17 +1,20 @@
-import { useQuery } from '@tanstack/react-query'
+import { queryOptions } from '@tanstack/react-query'
 import { components } from 'api-types'
-import { get, handleError } from 'data/fetchers'
-import type { ResponseError, UseCustomQueryOptions } from 'types'
-import { keys } from './keys'
 
-export type GetThirdPartyAuthIntegrationsVariables = {
+import { keys } from './keys'
+import { get, handleError } from '@/data/fetchers'
+import type { ResponseError } from '@/types'
+
+export type ThirdPartyAuthIntegrationsVariables = {
   projectRef?: string
 }
 
+export type ThirdPartyAuthIntegrationsError = ResponseError
+
 export type ThirdPartyAuthIntegration = components['schemas']['ThirdPartyAuth']
 
-export async function getThirdPartyAuthIntegrations(
-  { projectRef }: GetThirdPartyAuthIntegrationsVariables,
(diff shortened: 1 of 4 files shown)


apps/studio/data/third-party-auth/integrations-query.ts
```diff
@@ -29,16 +32,13 @@ export type ThirdPartyAuthIntegrationsData = Awaited<
...
-  { projectRef }: GetThirdPartyAuthIntegrationsVariables,
-  {
-    enabled = true,
-    ...options
-  }: UseCustomQueryOptions<ThirdPartyAuthIntegrationsData, ResponseError, TData> = {}
-) =>
-  useQuery<ThirdPartyAuthIntegrationsData, ResponseError, TData>({
+export const thirdPartyAuthIntegrationsQueryOptions = (
+  { projectRef }: ThirdPartyAuthIntegrationsVariables,
+  { enabled = true }: { enabled?: boolean } = { enabled: true }
```
charislam, line 37: Can we still allow passing the other fields in queryOptions? In case say we want to do override of staleTime, etc

charislam (approved):

apps/studio/data/third-party-auth/integrations-query.ts
ivasilov: Answered on slack.

PR #42476, opened 2026-02-05
chore(studio): filter out impact none incidents

dnywh:

UI improvement

We show a non-dismissable, scary, _We are investigating a technical issue_ banner to all users for very low priority things, like a telecom-specific outage for users in Yemen. These `"impact": "none"` events should not trigger the banner.

## What is the new behavior?

These `"impact": "none"` events no longer trigger the banner.

They _do_ still however show up in the support form and inform the AI Assistant, so users can self-debug if relevant.

## To test

Override your `incident-status` response with the following:

```json
[{
    "id": "79kw380212jb",
    "name": "Regional network issues in Yemen",
    "status": "identified",
    "impact": "none",
    "active_since": "2026-02-03T21:38:44Z"
}]
```

No banner should show. But the support form and AI Assistant should still show their respective awareness of the Yemen outage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced incident notification system to improve accuracy and reduce alert fatigue. Incident banners will now only display for incidents with meaningful impact, filtering out inconsequential incidents from the notification feed. This change provides users with clearer visibility into issues that actually affect the platform.

### DIFF
--- a/apps/studio/components/interfaces/App/AppBannerWrapper.tsx
+++ b/apps/studio/components/interfaces/App/AppBannerWrapper.tsx
@@ -12,10 +12,12 @@ export const AppBannerWrapper = ({ children }: PropsWithChildren<{}>) => {
   const { data: allStatusPageEvents } = useIncidentStatusQuery()
   const { maintenanceEvents = [], incidents = [] } = allStatusPageEvents ?? {}
 
+  // Only show incident banner for incidents with real impact (not "none")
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/App/AppBannerWrapper.tsx
dnywh, line 1: Scoping this check here where the banner is rendered, not in `incident-status-query`. The latter also hides it from view in the Support form.

joshenlim (approved):
Verified changes on preview!

PR #42502, opened 2026-02-05
Chore/adjust instance node for replicas

joshenlim:
## Context

Replaces the provider name with the region code in the infrastructure settings overview node diagram
- The latter is more important, and we don't really specify the cloud provider anywhere in the dashboard
Screenshot 2026-02-05 at 18 43 11

Also made the same changes in the upcoming database replication page too in the form of tooltips
And swapped the title + description to be more consistent (Top: type of data, bottom: description), same for the table below

Before:
image

image

After:
image

image

## Summary by CodeRabbit

* **New Features**
  * Added region metadata tooltips to replication and read replica nodes

* **Improvements**
  * Reorganized information hierarchy in replication diagram for better clarity
  * Expanded zoom range to improve diagram visibility
  * Increased node spacing for clearer visualization
  * Enhanced icon styling with improved color support


silentworks:
Could the aws region be aligned below instead of the side? it looks a bit weird at the side like it is.

PR #42512, opened 2026-02-05
Chore/fix connect sheet and dialog

joshenlim:
## Context

Connect dialog no longer shows up for local / self-host as we changed it a bit to do some testing with a new UI

PR here fixes it + add some e2e tests to ensure that either the connect sheet or dialog opens when clicking on the connect button

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a display issue where the Connect panel was not appearing in certain states; it now appears when expected.
* **Tests**
  * Added end-to-end tests covering Connect behavior: opening via URL, opening via header button, dismissing with Escape, and URL state updates.


awaseem (approved):
LGTM

PR #42546, opened 2026-02-06
Update enable data api copy

SaxonF:
Slightly modifies copy of data api warning during project creation

## Summary by CodeRabbit

* **Documentation**
  * Clarified the warning shown during project creation when configuring the Data API: the message now explicitly states that disabling the Data API prevents client libraries (e.g., supabase-js and similar) from querying or mutating your database, making the setup impact clearer for users.


joshenlim (approved):
Verified changes on preview

PR #42550, opened 2026-02-06
Chore/adjust instance node for replicas

joshenlim:
## Context

Replaces the provider name with the region code in the infrastructure settings overview node diagram
- The latter is more important, and we don't really specify the cloud provider anywhere in the dashboard
Screenshot 2026-02-05 at 18 43 11

Also made the same changes in the upcoming database replication page too in the form of tooltips
And swapped the title + description to be more consistent (Top: type of data, bottom: description), same for the table below

Before:
image

image

After:
image

image

## Summary by CodeRabbit

* **New Features**
  * Added region metadata with interactive tooltips across replication and database views.
  * Actions that modify replicas now refresh the list automatically.
  * Replication diagram layout updated with increased node spacing and a wider zoom-out range.

* **Style**
  * Destination and replica name/ID display reorganized for clearer, two-line presentation.
  * BigQuery icon now inherits color from CSS for better visual consistency.
  * Provider/region labels refined for clearer wording.


alaister (approved):
LGTM!

- Tested on preview

PR #42551, opened 2026-02-06
chore(studio): Consolidate incident banner + maintenance banner, make incident banner dismissible

joshenlim:
## Context

PR mainly adds a change to allow users to dismiss an incident banner

Am also taking this time to consolidate `IncidentBanner` and `MaintenanceBanner` into a single component `StatusPageBanner` so that the logic for rendering either component is more clearer.

## Intended behaviour (should be covered by e2e tests)
- Incident banner shows if there are incidents with impact !== 'none'
- Incident banner still takes precedence over Maintanence banner (if there's both incident + maintenence events)
  - Dismissing the incident banner will subsequently show the maintenance banner if the maintenance hasn't been dismissed before (just to make sure that the user doesn't miss any comms)
- Incident banner will not be dismissible if there's an override (via feature flag or env var)

## Summary by CodeRabbit

* **New Features**
  * Unified status page banner showing incident or maintenance notifications, with dismissible banners that persist across sessions; incident notices take precedence over maintenance.

* **Refactor**
  * Replaced separate incident and maintenance banners with a single consolidated StatusPageBanner component.

* **Tests**
  * Added end-to-end tests covering banner visibility, precedence, and dismissal behavior.


apps/studio/components/layouts/AppLayout/StatusPageBanner.tsx
```diff
@@ -0,0 +1,68 @@
...
+  if (showIncidentBannerOverride || (highImpactIncident && !dismissedIncident)) {
+    return (
+      <HeaderBanner
+        variant="warning"
+        title="We are investigating a technical issue"
+        description={BANNER_DESCRIPTION}
+        onDismiss={
+          showIncidentBannerOverride || !highImpactIncident
+            ? undefined
+            : () => setDismissedIncident(true)
```
matlin, line 50: Could also store a timestamp instead of a boolean so we could later treat it more like a "snooze" e.g. by dismissing it, it hides it for 24hrs. Dismissed state would just then be `isDismissed = dismissal && dismissal < (now() - 24hrs)`

joshenlim: hmm is there a case whereby a user would wanna temporarily snooze the banner tho? 

just thinking that users who close the banner likely don't wanna see it again, and it might come across to users as "another incident" when the banner shows up again

the support form will still have reference to the status page and any ongoing incidents nonetheless if we're afraid that closing the banner will lose reference to the ongoing incident

matlin: Yeah thats probably right I was just thinking in the sense of future proofing but agree that a dismissed banner retuning would feel like a bug if it wasn't clear that it was a snooze.

joshenlim: gotcha - we can revisit the banner dismiss behaviour down the line nonetheless if we do get any feedback around it! 🙂 happy to adjust! 🙏

alaister (approved):
LGTM!

- Tested minimally on preview

PR #42575, opened 2026-02-07
docs: update schedule functions guide

hizzuu:

YES

Docs update - Fix inconsistent variable name in code example

The documentation for scheduling Edge Functions has an inconsistency in the code example. The variable is stored as 'publishable_key' in Vault but referenced as 'anon_key' when building the Authorization header, which would cause the SQL query to fail.

## What is the new behavior?

The code example now correctly references 'publishable_key' throughout, matching the variable name used when storing the secret in Vault. This ensures the code example works as expected when users copy and use it.

Changed line 50 from:
```sql
'Authorization', 'Bearer ' || (select decrypted_secret from vault.decrypted_secrets where name = 'anon_key')
To:
'Authorization', 'Bearer ' || (select decrypted_secret from vault.decrypted_secrets where name = 'publishable_key')
Additional context
This is a small but important documentation fix that helps users avoid errors when following the guide to schedule Edge Functions with pg_cron.


ChrisChinchilla:
@hizzuu Functions is still abit edge case when it comes to the new key names, but it will change soon and then I will get to tyour PR :)

ChrisChinchilla:
Thanks @hizzuu, but it looks like other PRs ended up fixing this.

PR #42583, opened 2026-02-07
Fix incorrect port forwarding from Kong to Storage when external port differs

### LINKED ISSUE #42582: Kong forwards internal port to Storage when external port differs, breaking Storage URLs
# Bug report

## Describe the bug

When the external Kong port differs from the internal container port, Kong forwards the internal port to the Storage service.
As a result, Supabase Storage generates URLs (via the Location header) using the wrong port, causing subsequent requests to fail.

This happens because Kong passes its internal listening port (e.g. 8000) instead of the externally exposed port (e.g. 80).

## To Reproduce

1. Go to the Supabase repository:

```bash
cd docker
cp .env.example .env
```

2. Change the external Kong port to 80:

```bash
sed -i 's/^KONG_HTTP_PORT=8000/KONG_HTTP_PORT=80/' .env
```

3. Start supabase:

```bash
docker compose up -d
```
4. Access Storage: http://localhost/project/default/storage/files

5. Create a bucket named test.

6. Upload a file (~10MB).

7. Observe the error in the response.

Image

## Expected behavior

Storage should generate URLs using the external Kong port (the port exposed to the client), not the internal container port.

The Location header should reflect the correct externally accessible URL so that subsequent requests succeed.

## Actual behavior

Kong forwards its internal port (8000) to the Storage service.
Storage then uses this port to build the URL in the Location header.

Because the client is accessing Kong through port 80, all subsequent requests fail due to the incorrect port in the generated URL.

## System information

- OS: Arch Linux
- Browser: Zen Browser Mozilla Zen 1.18.3b

## Additional context

This issue only occurs when the external Kong port differs from the internal container port.
If both ports are the same, the problem does not happen.

The root cause seems to be that Kong forwards the internal listening port instead of respecting the externally exposed port when proxying requests to Storage.

luizfelmach:

YES

This PR fixes an issue where Supabase Storage generates invalid URLs when Kong is exposed on a different external port than the one it listens to internally.

When `KONG_HTTP_PORT` (external) differs from Kong’s internal container port (default `8000`), Kong was forwarding the **internal port** to Storage.  
As a result, Storage used this incorrect port when building URLs (via the `Location` header), causing subsequent requests to fail.

- Kong forwarded its **internal listening port** instead of the **externally exposed port**
- Storage relied on this value to construct redirect URLs
- Generated URLs pointed to an unreachable port (e.g. `:8000` instead of `:80`)
- Uploads would fail after the initial request

## What is the new behavior?

Fixes #42582

## Summary by CodeRabbit

* **Chores**
  * Updated Docker configuration for port mapping setup.


aantti:
See the following:
- PR #43837 (add `KONG_PORT_MAPS` when proxy is used)
- PRs [storage#900](https://github.com/supabase/storage/pull/900), [storage#914](https://github.com/supabase/storage/pull/914), and PR #43826 (adding `STORAGE_PUBLIC_URL`)

luizfelmach:
Closing as completed. Thanks @aantti.

PR #42606, opened 2026-02-09
Prevent double click action in ReferenceRecordPeak

joshenlim:
## Context

Tiny UX behaviour that we came across - just prevents the double click action when peeking a foreign row in the table editor

Feels odd that double clicking the `ReferenceRecordPeek` opens the side panel to select a new foreign row to reference

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced double-click handling to prevent unintended actions in the foreign key selection popover.


alaister (approved):
LGTM!

- Tested on preview

PR #42613, opened 2026-02-09
bug: fix refetch loop

jordienr:
Problem:
- new filters are creating an infinite fetch loop of ph.supabase.io 

#  Before:
  snap.filters change → effect → setParams() → router.push()
  → URL updates → filters/sorts recalc → setParams recreated
  → effect re-runs → INFINITE LOOP 🔄

 # After:
  snap.filters change → effect → comparison shows change → timeout scheduled
  → setParams() → router.push(shallow: true) → URL updates
  → filters/sorts recalc → setParams recreated → effect runs
  → comparison check: no content change → early return → LOOP BROKEN ✓

  The deep comparison prevents the effect from re-executing when only the dependency references change
  (due to URL updates), not the actual filter content. Test it out and the PostHog spam should stop!

## Bug Fixes

* **Prevented unnecessary URL updates** when filter state remains unchanged, improving application responsiveness.
* **Optimized filter and sort parameter handling** with enhanced callback efficiency and routing behavior.


awaseem (approved):
Tested on staging

PR #42615, opened 2026-02-09
Chore/fix redeem code UI

joshenlim:
## Context

Main fix was to address the size of the `OrganizationCard` on the redeem code page - was taking up the height of the flex div when the number of cards is <= 3 (Main fix is `max-h-min` in `OrganizationCard`)

### Before
image

### After
image

## Other changes
- Small improvement to perceived loading speed by only rendering the cards as loaders, text content on the left should render immediately on page load
- Loading state to consider if the feature flag has been loaded as well to prevent flicker of UI
- Hide the "Go to org" button when redeemed on the org page already

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading/gating so the redemption UI waits for feature flags and profile data before showing content.
  * Disabled submit when the redemption form is invalid to prevent empty submissions; submit button respects form validity, permissions, and loading state.

* **UI**
  * Conditionally shows upgrade and "Go to organization" options based on organization plan and current path.
  * Adjusted card layout to better constrain content height.

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/CreditCodeRedemption.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/CreditCodeRedemption.tsx
@@ -232,9 +232,25 @@ export const CreditCodeRedemption = ({
                     Upgrade organization
                   </UpgradePlanButton>
                 )}
-                <Button asChild type="default">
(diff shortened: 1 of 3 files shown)


awaseem (approved):
Tested on staging ✅

PR #42633, opened 2026-02-10
docs: fix Express OAuth callback example variable references

yogeshwaran-c:
Documentation fix

The Express callback example in the OAuth PKCE flow documentation has two bugs:

1. **Wrong variable references**: Uses `context.req` and `context.res` instead of `req` and `res`. The Express route handler receives `req` and `res` as parameters, not a `context` object.
2. **Brittle redirect path**: Uses `` `/${next.slice(1)}` `` which incorrectly strips the first character regardless of whether it's a `/`. For example, `next = "dashboard"` would become `/ashboard`.

## What is the new behavior?
1. Fixed `context.req.headers.cookie` → `req.headers.cookie` and `context.res.appendHeader` → `res.appendHeader`
2. Replaced brittle redirect with safe normalization: `next.startsWith('/') ? next : '/${next}'`

## Additional context
The other framework examples (Next.js, SvelteKit, Astro, Remix) in the same file correctly reference their request/response objects. Only the Express example had this issue.

## Summary by CodeRabbit

* **Documentation**
  * Updated OAuth PKCE flow examples to correct cookie handling and response header behavior.
  * Hardened redirect guidance to ensure only safe internal paths are used, preventing open-redirects.

### DIFF
--- a/apps/docs/content/_partials/oauth_pkce_flow.mdx
+++ b/apps/docs/content/_partials/oauth_pkce_flow.mdx
@@ -240,11 +240,11 @@ app.get("/auth/callback", async function (req, res) {
       process.env.SUPABASE_PUBLISHABLE_KEY, {
     cookies: {
       getAll() {
-        return parseCookieHeader(context.req.headers.cookie ?? '')
+        return parseCookieHeader(req.headers.cookie ?? '')
       },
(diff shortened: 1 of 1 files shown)


Hallidayo:

PR #42635, opened 2026-02-10
fix(studio): allow add column button to work on empty tables

### LINKED ISSUE #42590: Add new column ("+") button not working when a table has no rows
# Bug report

## Describe the bug

In the Supabase managed platform (supabase.com) **Table Editor UI**, the **"+" button** in the grid header used to add a new column is **not clickable when the table has no rows** (0 rows).

The "+" button only becomes clickable/functional after there is **at least 1 row** in the table.

This prevents users from adding new columns to empty tables via the grid header shortcut.

## To Reproduce

1. Go to the Supabase Dashboard (supabase.com)
2. Navigate to the **Table Editor**
3. Create a new table (or use an existing table with no rows)
4. Ensure the table has **0 rows**
5. Try to click the **"+" button** in the grid header to add a new column
6. Observe that the button is **not clickable / does not respond**

7. Now insert **1 row** into the table
8. Try clicking the **"+" button** again
9. Observe that it **now works** and opens the "Add new column" panel

## Expected behavior

The **"+" button to add a new column** should be clickable and functional **regardless of whether the table has rows or not**. Users should be able to add columns to empty tables via the grid header.

## Workaround

- Add a column by clicking the table name dropdown → **"Edit Table"** → add columns from there, or
- Insert a row first, then use the "+" button to add columns

## System information

- **Supabase version:** Supabase managed (supabase.com)
- **Browser:** Firefox on Ubuntu

yogeshwaran-c:
Bug fix

The "+" button in the grid header to add a new column is not clickable when a table has no rows. The button only becomes functional after at least 1 row is inserted.

The root cause: the empty state overlay uses `inset-0` which positions it from `top: 0`, covering the entire grid container including the header row. Even though the overlay has `pointer-events-none`, the combination of the overlay's `z-[1]` stacking context and the DataGrid's `contain: strict` (which creates its own stacking context) prevents click events from reaching the header in certain browsers.

Closes #42590

## What is the new behavior?
The empty state overlay now starts at `top: 35px` (matching the grid header row height) instead of `top: 0`, so it no longer covers the header row. The "+" add column button is now clickable regardless of whether the table has rows.

## Additional context
The fix is a single CSS class change: replacing `inset-0` with explicit `left-0 right-0 bottom-0 top-[35px]` to skip the header area. The `35px` value matches react-data-grid's default header row height used by this component.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed vertical positioning of the empty state placeholder in the grid when no rows are present.

### DIFF
--- a/apps/studio/components/grid/components/grid/Grid.tsx
+++ b/apps/studio/components/grid/components/grid/Grid.tsx
@@ -237,7 +237,7 @@ export const Grid = memo(
           {(rows ?? []).length === 0 && (
             <div
               className={cn(
-                'absolute w-full inset-0 flex flex-col items-center justify-center p-2 z-[1] pointer-events-none',
+                'absolute w-full left-0 right-0 bottom-0 top-[35px] flex flex-col items-center justify-center p-2 z-[1] pointer-events-none',
                 isTableEmpty && isDraggedOver && 'border-2 border-dashed',
                 isValidFileDraggedOver ? 'border-brand-600' : 'border-destructive-600'
               )}


7ttp (commented):
Hey @yogeshwaran-c 
Thanks for Contributing! 💚
 this was already fixed in #42618 which added pointer-events-none to allow clicks through the empty state. 

Closing this!

PR #42636, opened 2026-02-10
Remove default freeze on PK column in table editor

joshenlim:
## Context

Just removes the default column freeze behaviour on primary key columns in the table editor. Behaviour should still persist if freezing manually.

## Summary by CodeRabbit

* **Updates**
  * Grid columns are no longer frozen based on primary key designation.


alaister (approved):
LGTM!

- Tested on preview

PR #42645, opened 2026-02-10
Fix data api exposed schemas to handle spaces in schema names

joshenlim:
## Context

Fixes Data API settings, exposed schemas parameter to handle schemas with spaces in the name - was buggy due to the global removal of all spaces in the `db_schema` string (presumably old logic as the input field used to a free input in the past, which now is a multi select)

Also adjusts the admonition when the public schema is not exposed

Before:
image

After:
image

## To test

- [ ] Have a schema with a space in it's name
- [ ] Verify that you can add + save , remove + save that schema in the data api settings exposed schemas field with no issues


awaseem (approved):
Nice!

Issue #17168, opened 2023-09-05
MFA example is missing RLS policy for mfa_factors


KirioXX:
# Improve documentation

## Link 

https://supabase.com/docs/guides/auth/auth-mfa

## Describe the problem

I tried to implement MFA and enforce only for users that have opted-in.
But when I try to fetch now I get this error:
```sh
permission denied for table mfa_factors
```
I tried a couple combinations to make mfa_factors readable but I can't get it to work.

## Describe the improvement

It would be great to have also an example RLS policy for the mfa_factors 
to make them accessible only for the MFA check.

## Additional context

Add any other context or screenshots that help clarify your question.

itsajay1029:
I'd like to work on this if you guide me a little @KirioXX

itsajay1029:
Is this something I can refer and add to the docs : [MFA AUTH VIA RLS]( https://supabase.com/blog/mfa-auth-via-rls#enforcement-via-row-level-security 
)

KirioXX:
Hey @itsajay1029,

Thanks for responding! The docs actually already have some examples for RLS policies here:
https://supabase.com/docs/guides/auth/auth-mfa#enforce-rules-for-mfa-logins

I think those are the same as the one in the article. But they miss a policy for the auth.mfa_factors table.
Without that the other policies are useless and will fail.

Collin0810:
Hi @KirioXX,

I encountered a similar issue and managed to resolve it by creating a separate `check_mfa` function set as a `SECURITY DEFINER`.

This function enforces MFA when the user has at least one MFA factor verified:
```
CREATE OR REPLACE FUNCTION public.check_mfa()
RETURNS boolean
LANGUAGE plpgsql
SECURITY DEFINER
AS $$
BEGIN
    RETURN (
        array[auth.jwt()->>'aal'] <@ (
            SELECT
                CASE
                    WHEN count(id) > 0 THEN array['aal2']
                    ELSE array['aal1', 'aal2']
                END
            FROM auth.mfa_factors
            WHERE auth.uid() = user_id AND status = 'verified'
        )
    );
END;
$$;
```

Once created, you can implement the checks in your RLS policies as follows:
```
check_mfa() AND auth.uid() = user_id
```

Hope this helps!

kiwicopple:
Closing the issue as a part of large repository cleanup, due to it being inactive and/or outdated. There's a large chance that this issue has been solved since.
[...]

PR #42659, opened 2026-02-10
docs: fix MFA opt-in RLS policy example to use security definer function

yogeshwaran-c:
Documentation fix (bug fix)

The MFA documentation's "Enforce only for users that have opted-in" RLS policy example directly queries `auth.mfa_factors`. However, the `authenticated` role does not have SELECT permission on this table, causing every request to fail with:

```
ERROR: 42501: permission denied for table mfa_factors
```

This has been reported as a bug multiple times (see also #17168 which was closed for inactivity).

Closes #36024

## What is the new behavior?
The example now uses a two-step approach:

1. A `security definer` function (`public.mfa_is_required()`) that safely queries `auth.mfa_factors` with postgres-level permissions
2. A simplified RLS policy that calls this function

The function uses `set search_path = ''` for security hardening, following Supabase best practices for security definer functions.

The RLS policy logic is also simplified from the previous array-based approach to a more readable boolean expression:
```sql
(select auth.jwt()->>'aal') = 'aal2' or not (select public.mfa_is_required())
```

## Additional context
The issue author provided a working workaround using the same `security definer` approach but placed the function in the `auth` schema. Since creating functions in the `auth` schema is no longer allowed (as of CLI v2.22.15), this fix places the function in the `public` schema instead.

## Summary by CodeRabbit

* **Documentation**
  * Updated MFA authentication guide with improved security verification approach.

* **Security**
  * Enhanced MFA verification mechanism for more reliable authentication factor validation.

### DIFF
--- a/apps/docs/content/guides/auth/auth-mfa.mdx
+++ b/apps/docs/content/guides/auth/auth-mfa.mdx
@@ -236,28 +236,37 @@ create policy "Policy name."
 Users that have enrolled MFA on their account are expecting that your
 application only works for them if they've gone through MFA.
 
+Since the `authenticated` role does not have direct access to `auth.mfa_factors`, you need to wrap the check in a `security definer` function. This function runs with the permissions of the owner (`postgres`), which can query the `auth` schema.
+
+```sql
+create or replace function public.mfa_is_required()
+returns boolean as $$
+  select exists (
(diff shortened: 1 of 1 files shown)


Hallidayo:

PR #42664, opened 2026-02-10
fix(studio): remove incident banner automation

charislam:
## Summary by CodeRabbit

* **Refactor**
  * Status banner now only appears for ongoing incidents (controlled by feature flag/env), maintenance events are no longer shown
  * Removed persistent dismissal behavior for the banner; dismissal state no longer stored locally

* **Tests**
  * Simplified/removed legacy tests around maintenance, impact levels, and dismissal flows; test suite now only verifies absence when incident flag is unset


awaseem (approved):
LGTM!

PR #42667, opened 2026-02-11
chore(studio): hide password managers from edge functions

dnywh:

- Chore
- Resolves DEPR-163

1Password and other password managers’ widgets pop up in the _Name_ `input` despite that being irrelevant to password managers.
 
## What is the new behavior?

We tell that field to ignore password mangers.

## Additional context

Apple’s own password manager does not have an attribute and the secret value is password-like, so I’m unable to hide theirs.

## Summary by CodeRabbit

* **Chores**
  * Improved form behavior with password managers by updating input configuration settings on the edge function secrets form.


alaister (approved):
LGTM!

- Tested on preview

PR #42668, opened 2026-02-11
fix(studio): colour contrast for selected file storage item

dnywh:

- Fix
- Resolves DEPR-334

Not enough colour contrast on selected item. It’s hard to tel what’s selected, especially on light mode.

## What is the new behavior?

Updated the storage explorer row styling to use `bg-selection` for selected, opened, and previewed rows, which increases contrast in light mode.

| Before | After |
| --- | --- |
| Buckets  Supabase-60C5BBB5-FB94-44AA-9435-E9610ACB25AA | Buckets  Supabase |

## Summary by CodeRabbit

* **Style**
  * Updated file explorer row background styling to provide consistent visual feedback for opened, previewed, and selected item states in the storage interface.


alaister (approved):
LGTM!

- Tested on preview

PR #42674, opened 2026-02-11
fix(studio): remove tooltip transition-opacity to prevent drift animation

### LINKED ISSUE #36360: Dashboard Logs & Analytics - Tooltip Placement and Animation Issue
In the "Logs & Analytics" section, when viewing charts that display log statuses (errors, warnings, etc.), hovering the cursor over a chart bar to reveal the detailed tooltip causes this tooltip to initially appear on the far left side of the chart area and then begin to "float" or "drift" from this far-left position towards the hovered bar, covering a significant distance.

https://github.com/user-attachments/assets/3d56154d-7aab-4241-8ed7-230423ca045d

yogeshwaran-c:

Bug fix

Chart tooltips in Logs & Analytics drift/slide from the left side of the chart when hovering over bars. This happens because Recharts initially positions the tooltip at `left: 0` before repositioning it to the cursor location, and the `transition-opacity` CSS class on the `CustomTooltip` component makes this repositioning visible as a smooth drift animation.

Closes #36360

## What is the new behavior?

The tooltip appears instantly at the correct position without any drift animation. The opacity toggle between `opacity-0` and `opacity-100` still works for showing/hiding the tooltip when entering/leaving the chart, but without the CSS transition that caused the visible repositioning.

## Additional context

The fix removes the `transition-opacity` Tailwind class from the tooltip wrapper div in `ComposedChart.utils.tsx`. The opacity is still toggled based on `isActiveHoveredChart`, but the change is now instant rather than animated.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined tooltip visibility handling in chart components while maintaining existing behavior.

### DIFF
--- a/apps/studio/components/ui/Charts/ComposedChart.utils.tsx
+++ b/apps/studio/components/ui/Charts/ComposedChart.utils.tsx
@@ -221,8 +221,8 @@ export const CustomTooltip = ({
     return (
       <div
         className={cn(
-          'grid min-w-[8rem] items-start gap-1.5 rounded-lg border border-border/50 bg px-2.5 py-1.5 text-xs shadow-xl transition-opacity opacity-100',
-          !isActiveHoveredChart && 'opacity-0'
+          'grid min-w-[8rem] items-start gap-1.5 rounded-lg border border-border/50 bg px-2.5 py-1.5 text-xs shadow-xl',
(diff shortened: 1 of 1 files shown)


7ttp (commented):
i dont think its fading/drifting anymore :)
closing this up!

ref:

Image

PR #42677, opened 2026-02-11
fix(studio): allow sorting auth users table when not actively searching

### LINKED ISSUE #39658: Authentication Users UI can't be sorted by other than user ID
# Bug report

## Describe the bug
In the frontend auth/users/ columns can't be sorted. It's sorted by user ID by default, but can't be changed. It says "Sorting cannot be changed when searching on a specific column. If you'd like to sort on other columns, change the search to all columns from the header." but it is already set to "all columns". See also screenshot.

## To Reproduce

1. Go to Authentication -> Useres
2. Try to sort the table

## Expected behavior

The table UI can be sorted by other columns than user ID.

## Screenshots

Image

## System information

- OS: macOS and Windows
- tested on Chrome and Safari

yogeshwaran-c:

Bug fix

The sort dropdown in the Authentication > Users page is always disabled when a specific column filter (e.g., "Email address") is selected, even when no search keywords have been entered. Since the default filter is "Email address", users cannot change the sort order on initial page load.

The disabled state shows: "Sorting cannot be changed when searching on a specific column."

Closes #39658

## What is the new behavior?

Sorting is now only restricted when the user is **actively searching** (has entered keywords) with a specific column filter selected. When no search keywords are entered, the sort dropdown is fully functional regardless of the selected filter column.

This is safe because:
- Column-based cursor pagination (which requires fixed sort order) is only used when there are actual search keywords
- Without keywords, the query falls back to page-based pagination which supports arbitrary sorting

## Changes

- **`SortDropdown.tsx`**: Added `filterKeywords` prop; only disables sort when keywords is non-empty
- **`UsersV2.tsx`**: 
  - Pass `filterKeywords` to `SortDropdown`
  - Only use column-based search when keywords are present (or improved search is enabled)
  - Only reset sort to `id:asc` when changing filter column with active keywords

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the interaction between search and sort functionality in the Auth Users interface. The sort dropdown now properly respects active filter keywords, and automatic sort reset behavior has been refined when changing filters.

### DIFF
--- a/apps/studio/components/interfaces/Auth/Users/SortDropdown.tsx
+++ b/apps/studio/components/interfaces/Auth/Users/SortDropdown.tsx
@@ -15,6 +15,7 @@ import {
 
 interface SortDropdownProps {
   specificFilterColumn: string
+  filterKeywords: string
   sortColumn: string
   sortOrder: string
   sortByValue: string
@@ -26,6 +27,7 @@ interface SortDropdownProps {
 
 export const SortDropdown = ({
   specificFilterColumn,
+  filterKeywords,
   sortColumn,
   sortOrder,
   sortByValue,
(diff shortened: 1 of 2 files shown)


7ttp (commented):
seems to be expected behaviour as of now..

ref: (as stated by gary too)
https://x.com/stojaaan/status/1979523885449031726

will update here, if anything changes :) 💚

PR #42678, opened 2026-02-11
docs: replace deprecated getSession with getUser in Refine tutorial

veeceey:

YES

Docs update

The Refine tutorial's `authProvider.check` method uses the unrecommended `getSession()` to verify if a user is authenticated.

Related issue: #42193

## What is the new behavior?

Switched to `getUser()` which contacts the Supabase Auth server to verify the user, matching the recommended approach. This is also consistent with `getIdentity` in the same file already using `getUser()`.

## Additional context

Kept the change minimal -- only the `check` method needed updating since the rest of the authProvider was already using the right methods.

## Release Notes

* **Documentation**
  * Updated the getting started tutorial with improved authentication error messaging for clearer feedback on authentication failures.

### DIFF
--- a/apps/docs/content/guides/getting-started/tutorials/with-refine.mdx
+++ b/apps/docs/content/guides/getting-started/tutorials/with-refine.mdx
@@ -233,15 +233,14 @@ const authProvider: AuthProvider = {
   },
   check: async () => {
     try {
-      const { data } = await supabaseClient.auth.getSession()
-      const { session } = data
+      const { data, error } = await supabaseClient.auth.getUser()
 
-      if (!session) {
+      if (error || !data?.user) {
         return {
           authenticated: false,
-          error: {
+          error: error || {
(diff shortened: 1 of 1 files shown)


veeceey:
Updated `authProvider.check` in the Refine tutorial to use `getUser()` instead of the deprecated `getSession()`.

Changes:
- `supabaseClient.auth.getSession()` → `supabaseClient.auth.getUser()`
- Now destructures `{ data, error }` and checks for `error || !data?.user` instead of checking for `!session`
- Error name updated from "Session not found" to "User not found"

The `getIdentity` method in the same file was already using `getUser()`, so this makes the whole authProvider consistent.

veeceey:
Hi team, just a gentle bump on this docs PR — it replaces the deprecated getSession() with getUser() in the Refine tutorial. Small change, no code logic affected. Would appreciate a review\!

PR #42693, opened 2026-02-11
fix: prevent AIAssistant from submitting on IME confirmation

pastadev3:

YES

Bug fix

In the AI Assistant panel, pressing the Enter key while composing text with an IME (e.g., Japanese, Chinese, or Korean) triggers a message submission. This makes it impossible to confirm character conversion without accidentally sending an incomplete message.

## What is the new behavior?

The PR adds a check for `event.nativeEvent.isComposing`. Now, the message is only submitted when the Enter key is pressed outside of an IME composition. This allows users to use the Enter key to confirm character conversions without triggering a submission.

## Additional context

This is a standard fix for React-based text inputs to support international users who rely on IMEs.

## Release Notes

* **Bug Fixes**
  * Updated the AI Assistant chat input so pressing Enter only submits when the user is not composing with IME, helping prevent accidental message sends.


pastadev3:
@coderabbitai full review

Hallidayo:

PR #42698, opened 2026-02-11
fix(studio): disable retries on GitHub authorization check

yogeshwaran-c:

Bug fix

The `useGitHubAuthorizationQuery` hook uses React Query's default retry behavior (3 retries with exponential backoff). This means when checking if a user has authorized the GitHub integration, a failed authorization check is retried 3 times before settling. This is unnecessary since a single check is sufficient to determine authorization status.

There was an existing FIXME comment from a maintainer:
```
// FIXME(kamil): Do not retry, a single check is fine.
```

## What is the new behavior?

- Added `retry: false` to the query configuration so authorization checks are not retried on failure
- Removed the FIXME comment since the issue is now resolved

This is a minimal, targeted fix that addresses exactly what the FIXME requested.

## Additional context

File changed: `apps/studio/data/integrations/github-authorization-query.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated GitHub authorization query to prevent automatic retries on failure, providing more predictable error handling.

### DIFF
--- a/apps/studio/data/integrations/github-authorization-query.ts
+++ b/apps/studio/data/integrations/github-authorization-query.ts
@@ -4,7 +4,6 @@ import { get } from 'data/fetchers'
 import type { ResponseError, UseCustomQueryOptions } from 'types'
 import { integrationKeys } from './keys'
 
-// FIXME(kamil): Do not retry, a single check is fine.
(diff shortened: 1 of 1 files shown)


Hallidayo:

PR #42706, opened 2026-02-11
fix(studio): replace hardcoded hover color with semantic token in ProjectUsage

yogeshwaran-c:

Bug fix (styling consistency)

`ProjectUsage.tsx` uses `hover:text-gray-1200` for the hover state on panel header links. This hardcoded color does not respect theme changes.

## What is the new behavior?

Replaced `hover:text-gray-1200` with `hover:text-foreground`, which is the semantic token for primary text color and correctly adapts to both light and dark themes.

## Additional context

File changed: `apps/studio/components/interfaces/Home/ProjectUsage.tsx`

`text-gray-1200` is the darkest gray variant, which maps directly to `text-foreground` in the semantic token system.


7ttp (commented):

PR #42726, opened 2026-02-12
fix(studio): replace manual dark mode colors with bg-foreground in UserOverview

yogeshwaran-c:
Bug fix (styling)

The boolean indicator icons (checkmark/X) in `UserOverview.tsx` use `bg-black dark:bg-white` to manually handle dark mode, duplicating logic that the design system's semantic tokens already provide.

## What is the new behavior?
Replaces `dark:bg-white bg-black` with the semantic `bg-foreground` token. The `bg-foreground` token resolves to black in light mode and white in dark mode, which is the exact same behavior but expressed through the design system.

The component already uses `text-contrast` for the icon color inside these circles, confirming that the design system's color tokens are the intended approach.

## Additional context
`apps/studio/components/interfaces/Auth/Users/UserOverview.tsx`

This is a two-line change affecting the boolean property indicator icons (Check and X) in the user details panel.


7ttp (commented):

PR #42727, opened 2026-02-12
fix(studio): replace hardcoded bg-white and add text-foreground to IntegrationCard icons

yogeshwaran-c:
Bug fix (styling / dark mode)

The `IntegrationCard` component has two issues:

1. **Hardcoded `bg-white`** on the icon container (line 90) and loading state shimmer (line 26), which does not adapt to dark mode
2. **Missing `text-foreground`** override on the `icon()` call (line 91) -- the featured card variant correctly passes `text-foreground` to the icon, but the non-featured variant does not, causing the icon to use its default `text-black` which is invisible on the hardcoded white background in dark mode

## What is the new behavior?

| Element | Before | After |
|---------|--------|-------|
| Icon container background | `bg-white` | `bg-surface-100` |
| Loading shimmer background | `bg-white` | `bg-surface-100` |
| Icon color override | `icon()` (no override) | `icon({ className: 'w-full h-full text-foreground' })` |

This makes the non-featured card consistent with the featured card variant and ensures proper dark mode support.

## Additional context
`apps/studio/components/interfaces/Integrations/Landing/IntegrationCard.tsx`

The featured card already correctly uses `text-foreground` for the icon (line 66). This PR aligns the non-featured card to the same pattern.

### DIFF
--- a/apps/studio/components/interfaces/Integrations/Landing/IntegrationCard.tsx
+++ b/apps/studio/components/interfaces/Integrations/Landing/IntegrationCard.tsx
@@ -23,7 +23,7 @@ export const IntegrationLoadingCard = () => {
   return (
     <div className={cn(INTEGRATION_CARD_STYLE, 'pl-5 pr-6 py-3 gap-3 inline-flex h-[110px]')}>
       <div className="w-10 h-10 relative">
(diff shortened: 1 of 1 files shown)


7ttp (commented):

PR #42734, opened 2026-02-12
Chore/deprecate use query state with select part 03

joshenlim:
## Context

Related to FE-2461

More refactoring to clean up usage of `useQueryStateWithSelect`, mainly in the database pages
- Roles
- Triggers
- Functions
- Enumerated Types

## To test

In each of those pages, verify that
- [ ] Clicking the "new" cta updates the URL params, and refreshing should re-open the sheet
- [ ] Editing an existing item should update URL params, and refreshing should re-open the sheet with the right item
  - Verify that if the URL param has the wrong id, page should not open the sheet, show a toast, and reset the URL param
- [ ] Deleting an existing item should update URL params, and refreshing should re-open the sheet with the right item
  - Verify that if the URL param has the wrong id, page should not open the sheet, show a toast, and reset the URL param


awaseem (approved):
Tested on staging ✅

PR #42768, opened 2026-02-13
fix: persist first referrer across app boundaries

seanoliver:
## Summary

Fixes GROWTH-625.

Preserves first-touch attribution across app boundaries by persisting external referrer context at the edge and consuming it on Studio's initial pageview.

When users come from an external source to www/docs and then navigate to Studio, Studio often only sees the internal `supabase.com` hop. This change preserves the original external context so first-touch attribution is retained.

## What changed

- **Shared first-referrer cookie utilities** (`packages/common/first-referrer-cookie.ts`):
  - `isExternalReferrer`, `buildFirstReferrerData`, `serializeFirstReferrerCookie`, `parseFirstReferrerCookie`
  - `hasPaidSignals` — detects click IDs (gclid, fbclid, etc.) and paid utm_medium values
  - `shouldRefreshCookie` — centralizes stamp-or-skip decision for all apps
  - `stampFirstReferrerCookie` — shared middleware helper used by all apps (extracted from duplicated inline logic)
- **Edge middleware on all apps** — stamps cookie for external visitors, refreshes on paid signals:
  - `apps/www/middleware.ts` (simplified to use shared helper)
  - `apps/docs/middleware.ts` (simplified to use shared helper)
  - `apps/studio/proxy.ts` (integrated into existing proxy file)
- **Docs middleware matcher** — broadened from `/reference/:path*` to all non-static paths so the first-referrer cookie is stamped on all docs pages, not just reference paths
- **Telemetry** — Studio consumes cookie on initial pageview (`packages/common/telemetry.tsx`). `handlePageTelemetry` refactored from 7 positional params to an options object for readability.
- **Tests** — 22 unit tests covering all utilities and edge cases (including direct-navigation scenario)

## Behavior

- Writes `_sb_first_referrer` cookie when:
  - cookie is not already set and request has an external referrer, OR
  - cookie exists but incoming URL has paid traffic signals (click IDs or paid utm_medium)
- Cookie: 365-day TTL, `domain=supabase.com`, `sameSite=lax`, `secure=true` in production
- On first Studio pageview, if current referrer is internal and cookie has external context:
  - use persisted external referrer
  - apply persisted UTM/click-id/landing-url attribution props
- Measurement properties: `first_referrer_cookie_present`, `first_referrer_cookie_consumed`

## Manual testing

1. Visit `supabase.com/pricing?utm_source=google&utm_medium=cpc` from an external referrer (or use DevTools to set a `Referer` header)
2. Check `_sb_first_referrer` cookie is set in Application > Cookies
3. Navigate to Studio (`supabase.com/dashboard`)
4. In PostHog (or browser network tab), verify the first `$pageview` event has:
   - `first_referrer_cookie_present: true`
   - `first_referrer_cookie_consumed: true`
   - `$utm_source: "google"`, `$utm_medium: "cpc"`
   - `$referrer` points to the external source, not `supabase.com`
5. Verify subsequent route changes do NOT include `first_referrer_cookie_*` properties

## Review feedback addressed

- Added `secure: true` flag on production cookies (Pam's first comment)
- Fixed inaccurate JSDoc on `utms` field — keys retain `utm_` prefix (Pam's fourth comment)
- Added test coverage for edge cases: malformed URLs, multi-cookie headers, http:// referrers (Pam's sixth comment)
- Docs matcher broadening: fast-path exit on cookie-exists check keeps overhead minimal, exclusion list is correct
- Extracted shared middleware helper to eliminate duplication across 3 apps
- Refactored `handlePageTelemetry` from positional params to options object
- Removed redundant null check in `hasPaidSignals`
- Added direct-navigation test case
- Deleted dead `apps/learn/middleware.ts`
- Fixed studio build: integrated cookie stamping into existing `proxy.ts` (Next.js 16 rejects both middleware.ts and proxy.ts)


pamelachia (commented):
Ran this through claude code review and it came back with some stuff. Also is there a way to test this in preview? Couldn't really find a good way to do it. 

Important
  1. Missing secure flag on cookie — Both apps/www/middleware.ts and apps/docs/middleware.ts set sameSite: 'lax' but omit secure: true. Since the cookie carries attribution data scoped to domain: supabase.com, it should be secure on
  production. Can be conditional: secure: request.nextUrl.protocol === 'https:' to keep localhost working.
  2. Docs middleware matcher broadened from /reference/:path* to catch-all — Now every docs page request runs through middleware. The fast path (cookie already set) exits after a single cookies.has() check so overhead is minimal, and the
  reference-rewrite logic is still gated correctly. But worth confirming with the docs team that no other path-specific behaviors depend on middleware NOT running. The exclusion list (api|_next/static|_next/image|favicon.ico|__nextjs)
  looks correct.
  3. getFirstTouchAttributionProps return type mismatch — The function returns { $utm_source: string | undefined } | false patterns via && spreads, but the result gets spread into Record<string, string>. Not a runtime bug (truthiness
  guards prevent undefined), but the explicit type annotation could mask future issues. Pre-existing, not introduced by this PR.

Suggestions
  4. Inaccurate JSDoc — FirstReferrerData.utms comment says "keys without the utm_ prefix" but keys retain the prefix (utm_source, etc.). Should fix the comment.
  5. Cookie size bound — Payload is ~400-1100 bytes (comfortably under 4KB limit), but consider truncating referrer/landing_url to a max length as a safety net for extreme URLs.
  6. Test coverage gaps — Tests are good but could add: buildFirstReferrerData with malformed landingUrl, empty string input to parseFirstReferrerCookie, cookie header with multiple cookies, and isExternalReferrer with http:// referrers.
  7. undefined vs null for firstReferrerData — The undefined (not passed) vs null (checked but not found) distinction for controlling measurement property emission is correct but subtle. The inline comment documents it well.

seanoliver:
Hey @pamelachia, addressed your feedback in 9831d45:

1. **secure flag** — Added `secure: true` in the production domain conditional across all four middleware files. On localhost/previews it's omitted so dev still works.

2. **Docs matcher broadening** — The fast path (cookie-exists + no paid signals) exits after a single `shouldRefreshCookie()` call so overhead is negligible. The exclusion list covers all Next.js internals. Happy to check with the docs team if you think it's worth a Slack thread.

3. **getFirstTouchAttributionProps return type** — Pre-existing, agreed. Can clean up separately.

4. **JSDoc fix** — Updated the utms comment to say "e.g. utm_source, utm_medium" instead of claiming keys drop the prefix.

5. **Cookie size bound** — Skipping for now since payloads are well under 4KB, but happy to add a truncation if you feel strongly.

6. **Test coverage** — Added tests for: malformed landingUrl, empty string input, multi-cookie headers, and http:// referrers. Now at 21 tests.

Also extended coverage to `apps/studio`, and added paid traffic refresh so returning visitors who click a paid ad get fresh attribution instead of stale organic data.

pamelachia:
Fixed the typecheck error but it seems testing in preview it didn't work. After going to the studio 2 pageview events for that page occured and the second one had properties cookie_present is false, cookie_consumed is false, and it shows direct referrer.

charislam (approved):

seanoliver:
> Fixed the typecheck error but it seems testing in preview it didn't work. After going to the studio 2 pageview events for that page occured and the second one had properties cookie_present is false, cookie_consumed is false, and it shows direct referrer.

Hey @pamelachia, thanks for testing this — the preview environment unfortunately can't support the cross-app cookie handoff. Here's what's happening:

Each Vercel preview deploy gets its own unique `.vercel.app` hostname (e.g. `zone-www-dot-com-git-...vercel.app` for www, `studio-staging-git-...vercel.app` for studio). Since `vercel.app` is on the Public Suffix List, browsers won't let you set a cookie that's shared across those hostnames — so the cookie stamped on the www preview is completely invisible to the studio preview.

In production this isn't an issue because everything runs under `*.supabase.com` and the cookie is set with `domain=supabase.com`, which makes it readable across all our apps.

To test locally:
1. Run `pnpm dev:studio` and `pnpm dev:www`
2. Visit `localhost:3000/pricing?utm_source=google&utm_medium=cpc` (www)
3. Check that `_sb_first_referrer` is set in Application > Cookies
4. Navigate to `localhost:8082` (studio)
5. The cookie should be visible and consumed on the first pageview

Since both apps share `localhost` as the hostname, the cookie crosses the app boundary the same way it would under `supabase.com` in production.

pamelachia (approved):

PR #42775, opened 2026-02-13
fix: card paddings

fsansalvadore:
Uniform card paddings to use the `--card-padding-x` css var.

## Before
Screenshot 2026-02-13 at 12 21 57

## After
Screenshot 2026-02-13 at 12 22 11

## Release Notes

* **New Features**
  * Enhanced form validation in billing settings with structured data handling

* **Style**
  * Standardized padding and spacing across components for visual consistency
  * Updated layout spacing in observability and metrics components

* **Chores**
  * Reorganized imports and optimized Tailwind configuration with new padding token

### DIFF
--- a/apps/studio/components/interfaces/Observability/ServiceHealthCard.tsx
+++ b/apps/studio/components/interfaces/Observability/ServiceHealthCard.tsx
@@ -81,7 +81,7 @@ export const ServiceHealthCard = ({
         </div>
       </CardHeader>
 
-      <CardContent className="py-4 px-[var(--card-padding-x)] pt-4 flex-1 overflow-hidden max-h-[200px]">
+      <CardContent className="p-[var(--card-padding-x)] pt-4 flex-1 overflow-hidden max-h-[200px]">
         <Loading isFullHeight active={isLoading}>
           <LogsBarChart
             isFullHeight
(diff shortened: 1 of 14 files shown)


apps/studio/components/interfaces/Observability/ServiceHealthCard.tsx
```diff
@@ -1,10 +1,10 @@
 import NoDataPlaceholder from 'components/ui/Charts/NoDataPlaceholder'
 import Link from 'next/link'
-import { Button, Card, CardContent, CardFooter, CardHeader, CardTitle, Loading, cn } from 'ui'
+import { Button, Card, CardContent, CardFooter, CardHeader, CardTitle, cn, Loading } from 'ui'
```
kemaldotearth, line 3: @fsansalvadore actually for these, we could replace them with our `<Chart />` component which should have the correct paddings already applied. I matched them up with the homepage charts. Might be worth giving them a go?

kemaldotearth (changes requested):

apps/studio/components/interfaces/Observability/ServiceHealthCard.tsx
fsansalvadore: I actually noticed this ServiceHealthCard component isn't used anywhere....

kemaldotearth: Maybe @jordienr can answer better here, but if it's not used a possible candidate to remove?

kemaldotearth (approved):
:shipit:

PR #37185, opened 2025-07-16
fix(docker): access buckets via s3 client

singh-inder:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

Bug fix

This PR adds the necessary configuration to access buckets via `/storage/v1/s3` endpoint in self-hosted instance. Also, storage image version is updated to match the version in docker-compose.yml

## What is the current behavior?

Fixes #37184

## Summary by CodeRabbit

* **Chores**
  * Upgraded storage service to the latest version with improved functionality and stability
  * Enhanced deployment configuration with environment-driven S3 credentials and storage settings for greater flexibility

PR #42778, opened 2026-02-13
fix(docker): use 127.0.0.1 for storage healthcheck in S3 compose

### LINKED ISSUE #42776: Self-hosted Docker: storage healthcheck fails due to IPv6 localhost resolution
## Describe the bug

When running the self-hosted Docker compose stack (with or without S3 override), the `supabase-storage` container is reported as `unhealthy` even though the service is running correctly.

The healthcheck uses:

  http://localhost:5000/status

Inside Docker containers, `localhost` resolves to `::1` (IPv6).  
However, the storage API binds to IPv4 (`0.0.0.0`), not IPv6.

As a result:
- The healthcheck attempts to connect to `::1:5000`
- No listener exists on IPv6
- Healthcheck fails
- Container remains marked `unhealthy`
- Service itself works normally

This creates confusion for self-hosted users.

## To Reproduce

1. Clone the Supabase repo
2. Run:
   docker compose -f docker-compose.yml -f docker-compose.s3.yml up
3. Run:
   docker ps
4. Observe `supabase-storage` marked as `unhealthy`

5. Exec into container:
   docker exec -it supabase-storage wget -S -O- http://localhost:5000/status

   → Connection refused

6. Test IPv4 explicitly:
   docker exec -it supabase-storage wget -S -O- http://127.0.0.1:5000/status

   → Returns successfully

## Expected behavior

Healthcheck should pass when the service is running.

## Proposed fix

Replace `localhost` with `127.0.0.1` in storage healthcheck:

From:
  http://localhost:5000/status

To:
  http://127.0.0.1:5000/status

This ensures deterministic IPv4 resolution inside containers and avoids IPv6 ambiguity.

## System information

- OS: macOS
- Docker: Compose v2
- Supabase: current main branch

7ttp:
## Summary
Storage container incorrectly shows as unhealthy in S3 Docker setup due to IPv6/IPv4 mismatch.

## Problem
`localhost` resolves to IPv6 `[::1]` inside containers, but storage API binds to IPv4 only.
 Healthcheck fails with connection refused.

## Solution
Use `127.0.0.1` instead of `localhost` in healthcheck for explicit IPv4 resolution.

## Related
- Closes #42776

## Summary by CodeRabbit

* **Chores**
  * Updated storage service configuration to improve internal health-check reliability.

### DIFF
--- a/docker/docker-compose.s3.yml
+++ b/docker/docker-compose.s3.yml
@@ -50,7 +50,7 @@ services:
           "--no-verbose",
           "--tries=1",
           "--spider",
-          "http://localhost:5000/status"
+          "http://127.0.0.1:5000/status"
         ]
       timeout: 5s
       interval: 5s


aantti:
This has been addressed in PR #37185

PR #42780, opened 2026-02-13
Fix grid error not clickable

joshenlim:
## Context

As per PR title - currently the error CTA here is not clickable

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error message display styling and layout in the grid component.

* **New Features**
  * Enhanced query identification during SQL execution preflight checks.


awaseem (approved):
Works!

PR #42795, opened 2026-02-14
fix(studio): use console.debug for upload timing log in storage explorer

yogeshwaran-c:

Bug fix (logging improvement)

The storage explorer's file upload function in `apps/studio/state/storage-explorer.tsx` uses `console.log` to output the total upload time, which clutters the browser console with debug-level information during normal operation.

## What is the new behavior?

Replaces `console.log` with `console.debug` for the upload timing message, keeping it filterable in browser dev tools and out of the default console output.

## Additional context

Single-line change. No functional behavior is modified — only the console severity level.

### DIFF
--- a/apps/studio/state/storage-explorer.tsx
+++ b/apps/studio/state/storage-explorer.tsx
@@ -1361,7 +1361,7 @@ function createStorageExplorerState({
       }
 
       const t2 = new Date()
-      console.log(
+      console.debug(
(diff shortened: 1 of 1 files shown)


7ttp (commented):
will fold this into the bulk PR 
am planning to open in a few weeks and will credit you there

(Closing this here for now since it’s a small improvement)

PR #42797, opened 2026-02-14
fix(studio): replace hardcoded hover:border-white with semantic token in EmptyStates

yogeshwaran-c:

Bug fix (styling consistency)

The `Header` component in `EmptyStates.tsx` uses a hardcoded `hover:border-white` class on the Supabase logo image border. This doesn't adapt to the current theme and breaks the project's semantic color token convention.

## What is the new behavior?

Replaces `hover:border-white` with `hover:border-stronger`, which is the semantic token for a stronger border on hover and respects the active theme.

**File changed:**
- `apps/studio/components/interfaces/Home/ProjectList/EmptyStates.tsx`

## Additional context

Per the project's styling guidelines, Tailwind classes should use semantic color tokens instead of hardcoded colors.


7ttp (commented):

PR #42851, opened 2026-02-16
Fix resizeable panel error in LogTable

joshenlim:
## Context

There's an error from the `LogTable` component `Previous layout not found` that's caused by trying to set `defaultSize` as a dynamic value for the `ResizablePanel` component. Theres 2 ways to address this really

Option 1: Set `id` param on `ResizablePanel` to be a dynamic: this came from a suggestion by the library's author [here](https://github.com/bvaughn/react-resizable-panels/issues/401#issuecomment-2351010630)

Option 2: Avoid dynamic values entirely

Opting for option 2 as the cleaner way to fix this. Just note as well that there's a couple of console warnings from the library about `Invalid layout total size` - this one's safe to ignore and is seemingly an issue with the current version of `react-resizable-panels` that we're using (v2, there's currently v4 already)

## To test

Tbh I couldn't reproduce this locally actually, but I happened to come across this same bug while debugging the `Invalid layout total size` warning in `DefaultLayout` by trying to set a dynamic value for `defaultSize` which made me realise the dynamic value was the one causing that error

image

## Release Notes

* **Refactor**
  * Improved logs viewer layout with resizable panels for better content management and log selection display.
  * Reorganized internal component structure and module exports for improved maintainability.

* **Chores**
  * Updated test imports to reflect internal module restructuring.

### DIFF
--- a/apps/studio/components/interfaces/Settings/Logs/LogTable.tsx
+++ b/apps/studio/components/interfaces/Settings/Logs/LogTable.tsx
@@ -1,4 +1,10 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
+import { IS_PLATFORM, useParams } from 'common'
+import { ButtonTooltip } from 'components/ui/ButtonTooltip'
+import { DownloadResultsButton } from 'components/ui/DownloadResultsButton'
+import { useSelectedLog } from 'hooks/analytics/useSelectedLog'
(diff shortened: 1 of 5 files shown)


jordienr (approved):

awaseem:
Merging, lets see if this reduces the error count

PR #42856, opened 2026-02-16
Opt to use connection string of read replica if available to power the table editor

joshenlim:
## Context

Part of dashboard scalability project

Opting to use the connection string of the project's read replica (if available) for read queries on the database.

Trialing with the Table Editor as a first pass - changes involved will opt to use replica connection string for `useTableRowsQuery`, `useTableRowsCountQuery`, and `useForeignKeyConstraintsQuery`

There's definitely optimizations to be done for deciding which replica to use - but am starting off with a rather naive logic to prioritize replicas in the same region as the project.

## Changes involved

- We're no longer passing `connectionString` as a param into the affected hooks, the `connectionString` is derived from within those hooks instead
- Change is feature flagged, so things should be status quo if flag is off (use primary database's connection string)
- Added `useConnectionStringForReadOps` hook which returns the replica's connection string if (Otherwise defaults to primary database connection string)
  - Feature flag is on
  - Project has a replica available

## To test

- [ ] Verify that the table editor works as expected for a project that has read replicas (There shouldn't be any change really)
  - [ ] Also just double check that updating cells in the table editor works as well (There's no change there, we're using the primary DB's connection string for mutation ops)
- [ ] ^ Same thing for a project that doesn't have read replicas
- [ ] ^ Same thing for local / self-host


jordienr (approved):

alaister (approved):
LGTM!

apps/studio/data/table-rows/table-rows-count-query.ts
```diff
@@ -60,20 +50,17 @@ export const getTableRowsCountSql = ({
...
     select reltuples as estimate
     from pg_class
     where oid = ${table.id}
 )
 select 
   case 
-    when estimate = -1 then (select pg_temp.count_estimate('${selectBaseSql.replaceAll("'", "''")}'))
-    when estimate > ${THRESHOLD_COUNT} then ${filters.length > 0 ? `pg_temp.count_estimate('${selectBaseSql.replaceAll("'", "''")}')` : 'estimate'}
+    when estimate > ${THRESHOLD_COUNT} then estimate
     else (${countBaseSql})
```
avallete: **question**

This will cause the estimate count for filtered tables to never update. Before we didn't relied on the `retuples` but on an estimate of the result of the filtered set.

If that's intended to simplify things over (which arguably is a acceptable alternative) then, LGTM, otherwise it'll cause a BC in a non feature flagged way.

apps/studio/data/table-rows/table-rows-count-query.ts
```diff
@@ -60,20 +50,17 @@ export const getTableRowsCountSql = ({
...
 with approximation as (
     select reltuples as estimate
     from pg_class
     where oid = ${table.id}
 )
 select 
   case 
-    when estimate = -1 then (select pg_temp.count_estimate('${selectBaseSql.replaceAll("'", "''")}'))
-    when estimate > ${THRESHOLD_COUNT} then ${filters.length > 0 ? `pg_temp.count_estimate('${selectBaseSql.replaceAll("'", "''")}')` : 'estimate'}
+    when estimate > ${THRESHOLD_COUNT} then estimate
```
avallete: **issue**

I think we still wanna keep the `pg_temp.count_estimate` here, for both case (-1 / filter).
I don't see the benefit for dropping it off. Trying out on the preview, I can see this cause quite a regression in my "stress-editor-project", over a partitionned table (140M+ rows):

Image

Image

This run the exact count (giving me the 148M rows as a result in preview), while keeping things with a fast estimate (~147M rows) on current staging.

Since this is not feature flagged, this could cause issues to the dashboard of some users, since this query is always run when clicking on a table.

I know we can't `CREATE OR REPLACE` function on a read replica. But maybe then, this should be passed as part of our default `supabase.X` postgres schema image, so it's always available. Or we wanna feature flag this, and only run the hard count when there is an estimate:
 (`when estimate = 1 OR estimate > ${THRESHOLD_COUNT} then estimate` (since users can force an exact count from UI))

Or, have a way to make this query "read-replica aware", removing the use of the `count_estimate` in such case. Keeping it otherwise.

avallete (changes requested):

apps/studio/data/table-rows/table-rows-count-query.ts
joshenlim: ohhhhhhhhhh you're righhttttt that's why we did `explain` to get the rows estimate ahh 🤦

apps/studio/data/table-rows/table-rows-count-query.ts
joshenlim: great catch on this one 🙏 🙏  i'll have a think! have reverted the latest change for now

joshenlim:
Just a quick update as to the latest changes from the commit [here](https://github.com/supabase/supabase/pull/42856/commits/547c1d681d7c3e2ae395d0e929bba0398f7c2aae)

Have decided to tweak the UX of the Grid footer a little
- Show pagination controls irregardless of whether the count has successfully loaded or not (There's still a loading indicator beside the pagination controls while the count is loading)
- If we're using the read replica's connection string to fetch the count data, avoid using the `pg_temp.count_estimate` function and just omit the count value if the table's total row count (based on reltuples) is below 50K (the existing threshold) ([ref](https://github.com/supabase/supabase/pull/42856/changes#r2857912851))
  image
- Note that if there's no count value, users can still paginate between the pages 
  -  This is the same behaviour to what we already have for foreign tables which don't have the count value accessible
- No change nonetheless if we're using the primary database's connection string (which is the existing behaviour, even after this PR goes through as the feature flag is off still)

Am thinking this is probably the best balance for now, but happy to discuss further if any feedback!

apps/studio/data/table-rows/table-rows.sql.ts
```diff
@@ -0,0 +1,89 @@
...
+    let countQueryChains = countQuery.from(table.name, table.schema ?? undefined).count()
+    filters
+      .filter((x) => x.value && x.value != '')
+      .forEach((x) => {
+        const value = formatFilterValue(table, x)
+        countQueryChains = countQueryChains.filter(x.column, x.operator, value)
+      })
+    const countBaseSql = countQueryChains.toSql().slice(0, -1)
+
+    if (isUsingReadReplica) {
```
joshenlim, line 55: @avallete JFYI the main change in SQL query is here - i've decided to avoid using `pg_temp.count_estimate` in this scenario, and return count as -1 as an indication to omit the count value in the UI

avallete (approved):
LGTM !

alaister (approved):
LGTM!

- Tested on preview

PR #42880, opened 2026-02-16
fix: remove duplicate text-sm class in FunctionsEmptyState

yogeshwaran-c:

Bug fix (duplicate CSS class cleanup)

In `FunctionsInstructionsLocal`, two `<p>` elements have `text-sm` specified twice in their `className`:

```tsx
className="text-sm text-foreground-light mt-1 mb-4 prose [&>code]:text-xs text-sm max-w-full"
//         ^^^^^^^^                                                       ^^^^^^^^
```

While Tailwind deduplicates at runtime, the duplicate is clearly unintentional and adds noise.

## What is the new behavior?

The second `text-sm` is removed from both elements, leaving:

```tsx
className="text-sm text-foreground-light mt-1 mb-4 prose [&>code]:text-xs max-w-full"
```

No visual change — purely a code cleanliness fix.

### DIFF
--- a/apps/studio/components/interfaces/Functions/FunctionsEmptyState.tsx
+++ b/apps/studio/components/interfaces/Functions/FunctionsEmptyState.tsx
@@ -219,7 +219,7 @@ export const FunctionsInstructionsLocal = () => {
                 <Code size={20} />
                 <h4 className="text-base text-foreground">Create an Edge Function</h4>
               </div>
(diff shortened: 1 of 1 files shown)


7ttp (commented):
thanks for this aswell @yogeshwaran-c 

will add this in the BULK PR and credit you !

PR #42881, opened 2026-02-16
fix: correct typos in organization-not-found error messages

yogeshwaran-c:

Bug fix (text typos in user-facing error messages)

The organization-not-found admonition displays:

> ...does not exist or you do not have permission to access **to** it. Contact **the the** owner if you believe this is a mistake.

Two issues:
1. **"the the owner"** — duplicated word
2. **"access to it"** — incorrect preposition (should be "access it")

This appears in both `apps/studio/pages/organizations.tsx` and `apps/studio/components/interfaces/Organization/OrgNotFound.tsx`.

## What is the new behavior?

> ...does not exist or you do not have permission to access it. Contact the owner if you believe this is a mistake.

## Additional context

Purely textual fix — no logic changes.

### DIFF
--- a/apps/studio/components/interfaces/Organization/OrgNotFound.tsx
+++ b/apps/studio/components/interfaces/Organization/OrgNotFound.tsx
@@ -27,8 +27,8 @@ export const OrgNotFound = ({ slug }: { slug?: string }) => {
               ) : (
                 <>This organization </>
               )}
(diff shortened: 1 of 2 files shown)


7ttp (commented):
thanks @yogeshwaran-c 💚

will include this, in the bulk PR 
since its a small change (will give u the credit)

PR #42884, opened 2026-02-16
fix: move React key prop to outermost Dialog in template list

yogeshwaran-c:

Bug fix (React key prop placement)

In `FunctionsInstructionsLocal`, the "Explore our templates" section renders a list of `<Dialog>` components via `.map()`, but the `key` prop is placed on the inner `<ResourceItem>` instead of the outermost `<Dialog>`:

```tsx
{templates.map((template) => (
  <Dialog>                          {/* <-- no key here */}
    <DialogTrigger asChild>
      <ResourceItem
        key={template.name}        {/* <-- key on wrong element */}
```

React requires the `key` to be on the outermost element returned by `.map()` for correct list reconciliation. Without it, React may log a "Each child in a list should have a unique key prop" warning and fail to properly reconcile items when the list changes.

## What is the new behavior?

```tsx
{templates.map((template) => (
  <Dialog key={template.name}>     {/* <-- key on outermost element */}
    <DialogTrigger asChild>
      <ResourceItem
        media={...}                 {/* key removed from here */}
```

## Additional context

Single line change — moves `key` from `ResourceItem` to `Dialog`.

### DIFF
--- a/apps/studio/components/interfaces/Functions/FunctionsEmptyState.tsx
+++ b/apps/studio/components/interfaces/Functions/FunctionsEmptyState.tsx
@@ -321,10 +321,9 @@ curl --request POST 'http://localhost:54321/functions/v1/hello-world' \\
         <ScaffoldSectionTitle className="text-xl mt-12">Explore our templates</ScaffoldSectionTitle>
         <ResourceList>
           {templates.map((template) => (
-            <Dialog>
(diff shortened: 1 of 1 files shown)


7ttp (commented):
resolved by #42957 Thanks!💚

PR #42896, opened 2026-02-16
fix: replace deprecated scale-600 tokens and remove duplicate hover class

yogeshwaran-c:

Bug fix (styling cleanup)

1. **`BlockField.tsx`**: Has conflicting hover classes `hover:bg-accent/50` and `hover:bg-surface-400` on the same element, plus an extra space in the className
2. **`ChooseChannelPopover.tsx`**: Uses deprecated `border-scale-600` token (no longer in Tailwind config) plus an extra space
3. **`PrivilegesHead.tsx`**: Uses deprecated `border-scale-600` token

## What is the new behavior?

1. Removed redundant `hover:bg-accent/50` (kept `hover:bg-surface-400` which is the semantic token)
2. Replaced `border-scale-600` with `border-default` (the semantic equivalent)
3. Same replacement for PrivilegesHead

## Additional context

`border-scale-600` is a deprecated token that is no longer defined in the Tailwind config. The semantic replacement is `border-default` which adapts to both light and dark themes.


7ttp (commented):

PR #42905, opened 2026-02-17
docs: update outdated app.supabase.io URL in local development guide

yogeshwaran-c:

Documentation fix

The local development overview doc (`apps/docs/content/guides/local-development/overview.mdx`) contains a link to the "Supabase Platform" pointing to `https://app.supabase.io/`, which is the old domain.

## What is the new behavior?

Updated the URL to `https://supabase.com/dashboard`, which is the current correct URL for the Supabase dashboard. The same document already uses `supabase.com/dashboard` elsewhere (line 238).

## Additional context

The old `app.supabase.io` domain redirects to `supabase.com/dashboard`, so links still work but it's better to point directly to the canonical URL.

### DIFF
--- a/apps/docs/content/guides/local-development/overview.mdx
+++ b/apps/docs/content/guides/local-development/overview.mdx
@@ -7,7 +7,7 @@ video: 'https://www.youtube-nocookie.com/v/vyHyYpvjaks'
 tocVideo: 'vyHyYpvjaks'
 ---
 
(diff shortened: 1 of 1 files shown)


7ttp (changes requested):
(same here)
Hey @yogeshwaran-c thanks for this!💚
I’ll include it in the batch PR I’m preparing and will credit you there

PR #42913, opened 2026-02-17
fix: use console.error for error logging in useLocalStorage hooks

yogeshwaran-c:

Bug fix (logging severity)

The `useLocalStorage` hook in both `apps/learn` and `apps/ui-library` uses `console.log()` to log errors from localStorage operations (JSON parsing failures, quota exceeded errors, etc.). This makes real errors indistinguishable from informational logs.

## What is the new behavior?

Both error catch blocks now use `console.error()` instead of `console.log()`, ensuring localStorage failures are logged at the correct severity level and are easier to identify during debugging.

## Additional context

Files changed:
- `apps/learn/components/use-local-storage.tsx` (2 occurrences)
- `apps/ui-library/components/use-local-storage.tsx` (2 occurrences)

This is consistent with the same fix previously applied to the Studio app's `useLocalStorage` hook.

### DIFF
--- a/apps/learn/components/use-local-storage.tsx
+++ b/apps/learn/components/use-local-storage.tsx
@@ -17,7 +17,7 @@ export function useLocalStorage<T>(key: string, initialValue: T) {
       return item ? JSON.parse(item) : initialValue
     } catch (error) {
       // If error also return initialValue
-      console.log(error)
+      console.error(error)
       return initialValue
     }
   })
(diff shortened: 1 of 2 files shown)


7ttp (commented):

> PS: if u think it's still a valid fix, please open an issue with a screenshot/detailed explanation and a pr linked to it w a before/after..

PR #42924, opened 2026-02-17
fix(studio): rename initalSelectedTable to initialSelectedTable in TriggerSheet

yogeshwaran-c:

Bug fix (variable name typo)

In `TriggerSheet.tsx`, the local variable is named `initalSelectedTable` (missing the second 'i' in "initial").

## What is the new behavior?

The variable is renamed to `initialSelectedTable` across all 4 occurrences in the file. This is a purely local variable, so the change has no external API impact.

## Additional context

`apps/studio/components/interfaces/Database/Triggers/TriggerSheet.tsx`

All 4 references are within the same `useEffect` block (lines 168-174), so this is a safe rename with no risk of breaking other code.


7ttp (commented):
(same here)
closing this up!

will add this onto the bulk pr and credit you there! thanks

7ttp:
(mb double clicked)

PR #42925, opened 2026-02-17
fix(studio): add missing rel="noreferrer" to target="_blank" links

yogeshwaran-c:
## What kind of change does this PR introduce?

Security/accessibility improvement

## What is the current behavior?

Several `<Link>` components with `target="_blank"` are missing the `rel="noreferrer"` attribute. Without it, the opened page receives the `Referer` header containing the current URL, which may leak sensitive information (e.g., project refs in dashboard URLs).

## What is the new behavior?

Added `rel="noreferrer"` to 9 link instances across 8 files, consistent with the existing pattern used in 113+ other `target="_blank"` links throughout Studio.

### Files changed:
- `apps/studio/components/ui/GrafanaPromoBanner.tsx` (2 links)
- `apps/studio/components/ui/BannerStack/Banners/BannerMetricsAPI.tsx`
- `apps/studio/components/ui/ObservabilityLink.tsx`
- `apps/studio/components/ui/Charts/ChartHeader.tsx`
- `apps/studio/pages/new/[slug].tsx`
- `apps/studio/pages/integrations/vercel/install.tsx`
- `apps/studio/components/layouts/LogsLayout/LogsSidebarMenuV2.tsx`
- `apps/studio/pages/project/[ref]/observability/storage.tsx`

## Additional context

The `rel="noreferrer"` pattern is already the standard in this codebase (113 existing instances). This PR brings the remaining outliers into alignment.

PR #42927, opened 2026-02-17
fix: add missing rel="noreferrer" to anchor tags with target="_blank"

yogeshwaran-c:

Bug fix (security hardening)

Three `<a target="_blank">` tags are missing the `rel="noreferrer"` attribute:

- `apps/studio/components/interfaces/Organization/BillingSettings/Subscription/EnterpriseCard.tsx` (2 instances — desktop and mobile CTA links)
- `apps/www/components/Nav/GitHubButton.tsx` (1 instance — global navigation GitHub star button)

Without `rel="noreferrer"`, the browser sends the `Referer` header to the target site when opening links in a new tab.

## What is the new behavior?

All three anchor tags now include `rel="noreferrer"`, consistent with the existing pattern used across the codebase (113+ instances already use `rel="noreferrer"`).

## Additional context

This follows the same pattern as #42925 which addressed missing `rel="noreferrer"` on Next.js `<Link>` components in Studio. These are native `<a>` tags that were missed in that pass.

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/EnterpriseCard.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/EnterpriseCard.tsx
@@ -56,7 +56,7 @@ export const EnterpriseCard = ({ plan, isCurrentPlan }: EnterpriseCardProps) => 
             })
           }
         >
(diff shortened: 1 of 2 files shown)


Hallidayo:

PR #43004, opened 2026-02-19
Joshen/fe 2573 table editor allow users to run query with cost warning

joshenlim:
## Context

Related to this previous PR [here](https://github.com/supabase/supabase/pull/42321)

Table Editor: Adding a CTA to the `HighQueryCost` UI to allow users to proceed with fetching data despite the high query cost warning, to prevent completely blocking the users from their workflows (realised that certain heavy queries are required and this safeguard shouldn't be creating dead-ends for users)

image

Clicking "Load more" will open a confirmation dialog, in which proceeding to load the data will thereafter suppress this preflight check for the table, for the rest of the browser session

image

## Other changes

- Also bumped the query cost threshold from 100,000 to 200,000 - the former might have been too aggressive 😓 
- (Unrelated) Added query cost tooltip for cron jobs high query cost warning
  image


alaister (approved):
LGTM!

- Tested on preview

PR #43010, opened 2026-02-19
Ensure logs pages are accessible irregardless of project's status

joshenlim:
## Context

Ensure that project's logs pages are accessible irregardless of project's status, given that logs are mutually exclusive from the project's database and doesn't require the database to be online for it to be functional.

## Other changes
Mainly around simplifying some logic within ProjectLayout
- Remove `disabled` prop in `SideBarNavLink` component, given that `route` prop already has a `disabled` param
  - Define `disabled` within `NavigationBar.utils.tsx` -> just have one place to define the "configuration" of a route
- Minimize manually declaring specific routes to render specific UI (for Restore failed)
  - All of these can be grouped under `routesToIgnoreDBConnection`

## To test
- [ ] Verify that you can access the logs page while the project is coming up
- [ ] Verify that you can access the logs page on a paused project (and also use it)


alaister (approved):
LGTM!

- Tested on preview

PR #43014, opened 2026-02-19
Joshen/fe 2609 make project filtersort controls more visible

joshenlim:
## Context

Makes the status filter on the projects page more visible by using the `FilterPopover` component + persist the filter status in local storage such that it'll be re-loaded when landing back on `org/[slug]`

### Before
image

### After
image

## To test
- [ ] Verify that status filter works
- [ ] Verify that status filter persists in local storage


awaseem (approved):
Tested this on our shared infra org!

PR #43017, opened 2026-02-19
docs: Rework Expo React native social auth

ChrisChinchilla:

YES

Reworks Expo React native social to use appropriate auth methods and restructures to use external code.


mandarini (commented):
@ChrisChinchilla great work!!! :D 
One small thing from me in `AuthProvider`: inside the `onAuthStateChange` callback, the session's `access_token` is already available but currently discarded, which means `getClaims()` ends up calling `getSession()` internally to retrieve it again:                                                                         

```typescript
  // current
  supabase.auth.onAuthStateChange(async (_event, _session) => {
    const { data } = await supabase.auth.getClaims()
    setClaims(data?.claims ?? null)
  })
```

I think we should be doing: 
```typescript
  supabase.auth.onAuthStateChange(async (_event, session) => {
    const { data } = session
      ? await supabase.auth.getClaims(session.access_token)
      : { data: null }
    setClaims(data?.claims ?? null)
  })
```

Passing the token directly avoids the redundant internal `getSession()` call and makes the intent clearer. Same applies to the duplicate `AuthProvider` in the second example file.

ChrisChinchilla:
Made that change @mandarini

mandarini (approved):

marceldSL:
I am getting 404 for the docs now at this page: https://supabase.com/docs/guides/auth/quickstarts/with-expo-react-native-social-auth

I was able to load it this morning before this merge, so something has probably gone wrong here :(

ChrisChinchilla:
@marceldSL Fix for this building now.

PR #43026, opened 2026-02-19
[DOCS-454] fix(Docs): Fix images from causing CLS on pages

jeremenichelli:

Bug fix that prevents images from causing a layout shift. This hurts UX since people using hashes to go to specific parts of our pages lose reference of the heading they land, when an image loads the scroll reference gets lost.

Images pushes content down eliminating #... behavior for links.

## What is the new behavior?

Images now have a defined width and height and use the Next Image to prevent layout shifts.

## Additional context

Sadly, I had to do something I try not to, but I bypassed our `ui` package `Image` component. The reason is that the component adds a "mounting effect" which achieves nothing and seems like tech debt to patch some other library behavior.

Code reference: https://github.com/supabase/supabase/blob/master/packages/ui/src/components/Image/Image.tsx#L64

Also, the only thing that gives us is "zoomable" which is really not giving much since in mobile images don't expand and in desktop they shrink. _Pics or didn't happen..._

https://github.com/user-attachments/assets/ce42b4cd-c583-477c-b09d-fc803d62135c

As a following task and best practice to prevent layout shifts, we should consider filling the width and height of all images so that `next/image` can produce aspect ratio config for all.


charislam (changes requested):
Thanks for a super clear PR description @jeremenichelli!

There is one other thing we get from our existing Image component that needs to be ported over — dark mode support, for example see: https://github.com/supabase/supabase/blob/d5cc70560da9cdcca19194525f0c251f87bd9f16/apps/docs/content/guides/auth/architecture.mdx?plain=1#L13-L19

jeremenichelli:
@charislam thanks for the pointer! I actually realize not only the light/dark thing, but the fact we have captions, so I created a reduced version of the component within the docs to unblock us.

This also made image and width required properties on all pages and attach the corresponding dimensions so dev builds don't error locally, and also we fix the CLS on all pages with images which is definitely good.

Take another look when you can :)

charislam (approved):
Nice catch with the captions. LGTM! 🚀

PR #43052, opened 2026-02-20
fix: remove entitlement checks from realtime settings

ignaciodob:
The Realtime configurable values are not hard limits, so we shouldn't model them as entitlements in the Realtime Settings.


awaseem (approved):
LGTM

PR #43064, opened 2026-02-21
ci: set explicit empty permissions for tokenless workflows

Rohan5commit:
## Summary\n- add explicit `permissions: {}` to workflows that do not require default `GITHUB_TOKEN` scopes\n\n### Updated workflows\n- `.github/workflows/trigger-nimbus-sync.yml`\n- `.github/workflows/validate-pr.yml`\n\n## Why\nBoth workflows rely on either custom token inputs or purely local checks. Explicitly setting empty permissions avoids unnecessary default token scopes and documents intent.\n\n## Notes\n- configuration-only changes\n- left `publish_image.yml` untouched because it chains into a reusable publish workflow with broader scope requirements\n

### DIFF
--- a/.github/workflows/trigger-nimbus-sync.yml
+++ b/.github/workflows/trigger-nimbus-sync.yml
@@ -5,6 +5,8 @@ on:
     branches: [master]
   workflow_dispatch: # Allow manual triggering
 
+permissions: {}
+
 jobs:
   trigger-nimbus-sync:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


eyrebuild (commented):
Lgtm 👍

7ttp (commented):
hi @Rohan5commit 💚

I don't think this is needed. Both workflows either use custom secrets (not GITHUB_TOKEN) or don't make any calls that require permissions.
Adding `permissions: {}` is unnecessary and doesn't improve security in any case😅

Thnx for thinking about security though! closing this up.

PR #43083, opened 2026-02-23
Fix table editor prefetch

joshenlim:
## Context

Noticed that the table editor prefetch wasn't working as intended as the table's rows were getting fetched again when opening the table.

Fix is that `preflightCheck` should've been excluded from the `table-rows` query key

## To test
Within the Table Editor
- [ ] Verify that `table-rows` is getting prefetched when hovering over a table in the side menu
- [ ] Verify that `table-rows` doesn't get fetched again when opening the table subsequently (There shouldn't be a UI loader too - rows should render immediately)


alaister (approved):
LGTM!

- Tested on preview

PR #43085, opened 2026-02-23
feat: require current password on reset password

staaldraad:

YES

feature

Change password only requires the new password

## What is the new behavior?

The current password is required and must be valid before the password can be changed.

Also hides a revealed password if the user clicks out of the input field.

## Additional context

Needs the backend flag to be changed, but is safe to merge _before_ backend change is made. Is unsafe to merge _after_ backend flag is changed (there will be a gap where users can't reset password through the dashboard).

before:
Screenshot 2026-02-23 at 09 58 04 
after:
Screenshot 2026-02-23 at 09 58 56

before:
Screenshot 2026-02-23 at 09 59 53

after: 
Screenshot 2026-02-23 at 10 00 10


apps/studio/components/interfaces/SignIn/ResetPasswordForm.tsx
```diff
@@ -20,42 +20,54 @@ import {
...
+const recoveryPasswordSchema = z.object({
+  currentPassword: z.string().optional(),
+  password: passwordValidation,
 })
 
 type FormData = z.infer<typeof passwordSchema>
 
 const ResetPasswordForm = () => {
   const router = useRouter()
+  const isRecoveryFlow = router.query.type === 'recovery'
```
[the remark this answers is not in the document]
staaldraad: the backend validates the `currentPassword` requirement

apps/studio/components/interfaces/SignIn/ForgotPasswordWizard.tsx
```diff
@@ -66,7 +66,7 @@ const ConfirmResetCodeForm = ({ email }: { email: string }) => {
       } else {
         await router.push({
           pathname: '/reset-password',
-          query: router.query,
+          query: { ...router.query, type: 'recovery' },
```
joshenlim: guessing we should be doing this change in `SignInMfaForm` as well?

[edit] ignore this one - will be doing [this](https://github.com/supabase/supabase/pull/43085#discussion_r2857064419)

apps/studio/components/interfaces/SignIn/ResetPasswordForm.tsx
joshenlim: the current password field will flash for a moment if the URL has `type=recovery` - looking into how we can make this better

joshenlim: i'd flip the logic around i think - set the query param if accessing this reset password page from "reset password" CTA

so account/me -> reset password -> /reset-password?type=something

but forgot password -> /reset-password

joshenlim (approved):
Verified both forgot password + change password flows on staging preview, both LGTM!

Just a heads up one issues noticed that needs to be patched on the BE
- Putting an invalid password as the current password returns an error message from the BE `Current password required` which is inaccurate

PR #43086, opened 2026-02-23
chore(studio): Implement auto retry logic for renaming folders in Storage Explorer

joshenlim:
## Context

Renaming folders on the Storage Explorer has a tendency to be flaky as we're using the `POST /objects/move` storage API to move files to the renamed folder which handles one file per request.

For a big folder (e.g 100+ files), users will easily hit rate limits and run into 429s with the API, in which the current behaviour just shows toast errors as the error handler, leaving an odd state upon running through every file in the folder where some files are in the old folder, and others in the new one

## Changes involved

Main solution is to implement an automatic retry logic for `renameFolder`, but changes involved are:
- Remove batching logic for renaming folders (Which fires `n` requests per batch)
  - The batch logic was negligible anyways since our batch size was only 2 😅 
  - Instead we now handle moving of files sequentially
- If a file runs into an error during the `POST` request, retry the request after `n` seconds
  - If the error was a 429, retry after 60 seconds, otherwise just 5 seconds
  - Will attempt the request for a total of 3 times
  - Toast progress description will also be updated to show the message (timer will also count down to keep the UI interactive and not look stuck)
    image
- After the whole process is completed, if any file ran into an error, toast description will add a CTA to retry the rename

## Demo

https://github.com/user-attachments/assets/d65b3c22-f9f8-4984-9521-6fc83a69f82e

## To test

Have a folder in your project's storage bucket with 100+ files (can stress test with nested folders)
- [ ] Verify that you can rename the folder, and the dashboard automatically handles errors


djhi (approved):
Tested in staging

kemaldotearth (approved):

PR #43088, opened 2026-02-23
fix: only enable install/uninstall button when user has permissions

imor:
Enable the install/uninstall integration button on the stripe sync engine only if the user has permissions to create/delete edge function secrets. This fixes a problem when non-admin/non-owner users try to install the integration but it fails due to lack of permissions. For a user without appropriate permissions a tool tip will be shown:

image

### DIFF
--- a/apps/studio/components/interfaces/Integrations/templates/StripeSyncEngine/InstallationOverview.tsx
+++ b/apps/studio/components/interfaces/Integrations/templates/StripeSyncEngine/InstallationOverview.tsx
@@ -1,9 +1,11 @@
 import { zodResolver } from '@hookform/resolvers/zod'
+import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { ButtonTooltip } from 'components/ui/ButtonTooltip'
 import { useStripeSyncInstallMutation } from 'data/database-integrations/stripe/stripe-sync-install-mutation'
 import { useStripeSyncUninstallMutation } from 'data/database-integrations/stripe/stripe-sync-uninstall-mutation'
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Integrations/templates/StripeSyncEngine/InstallationOverview.tsx
```diff
@@ -328,12 +336,14 @@ export const StripeSyncInstallationPage = () => {
...
                   onClick={() => setShouldShowInstallSheet(true)}
-                  disabled={!canInstall}
+                  disabled={!canInstall || !canManageSecrets}
                   tooltip={{
                     content: {
                       text: !canInstall
                         ? 'Your database already uses a schema named "stripe"'
-                        : undefined,
+                        : !canManageSecrets
+                          ? 'You need to be an administrator or owner to install the Stripe Sync Engine.'
```
joshenlim: just a tiny nit but for consistency, we typically use the language 'you need additional permissions' for tooltips of this context

joshenlim (approved):
Changes LGTM!

apps/studio/components/interfaces/Integrations/templates/StripeSyncEngine/InstallationOverview.tsx
imor: I can open a follow up PR to make the tooltips consistent but isn't it more informative for users to know which permission are the missing exactly instead of a generic message? Also I'm not sure about our auth model but can users be assigned individual permissions instead of just the roles admin, developer etc.?

PR #43090, opened 2026-02-23
feat: custom report snippets from inline sql editor

jordienr:
- adds option to create snippets from inline sql editor
- removes the date range from the header, its noisy

## to test
- go to custom report
- add block → snippets → add snippet
- create a new snippet in the inline SQL Editor
- back to add block → snippets → select your snippet
- your snippet should show in the report
- now try updating the snippet in the inline SQL Editor
- the chart/block should auto update
- amazing!


kemaldotearth (approved):
To unblock :shipit:

PR #43091, opened 2026-02-23
fix: remove connectionString check

### LINKED ISSUE #43081: This table cannot be accessed via the data api
# Bug report

## Describe the bug

In the RLS policy overview, tables are incorrectly labeled as:

- "API disabled"
- "This table cannot be accessed via the Data API as no permissions exist for the anon or authenticated roles."

This occurs even though valid `authenticated` and/or public policies are defined on the table.

The tables are accessible through the Data API and function correctly, but the dashboard UI displays the wrong status.

The issue appears in a local environment.

## To Reproduce

1. Run Supabase locally.
2. Open the local dashboard.
3. Navigate to the RLS overview for any table (e.g. in the `public` schema).
4. Observe the status shown for the table.

## Expected behavior

If valid policies exist for `anon` or `authenticated`, the UI should reflect that the Data API is enabled and accessible.

## System information

- OS: Windows 11 Pro (23H2)
- Browser: Brave 1.86.142 (Chromium-based)
- Supabase Version: 2.75
- Node.js Version: 21.6
- Local setup: Supabase CLI running via Docker Desktop

7ttp:
## Problem
In local/self-hosted mode, tables incorrectly show "API disabled" even when they have valid RLS policies. 

This happens because:
1. `project.connectionString` is `null` for local/self-hosted (defaults to null per API types)
2. `useTablesRolesAccessQuery` has `!!connectionString` in its enabled condition
3. The query never runs, so the UI can't determine actual API access status

## Solution
Removed `!!connectionString` check from `useTablesRolesAccessQuery`'s enabled condition.

This is safe cause:
-  likewise `executeSql` already handles null/empty connectionString via `connectionString ?? ''`
 For local/self-hosted, the query works without connectionString (pg-meta uses direct database connection)
The query now runs and correctly works

Before:
Before fix - API disabled shown incorrectly

After:
After fix - API status displays correctly

- Closes #43081


7ttp:
@aantti when available, can u have a look at this?! 😅💚

IdellHeaney (commented):
Working

awaseem:
@7ttp We might also need to add the loading state of this query, that can be done later because it just flickers in

awaseem (approved):
LGTM tested in staging and production

7ttp:
> @7ttp We might also need to add the loading state of this query, that can be done later because it just flickers in

bet bet! will push up another PR with the loading state aswell :) 💚

PR #43096, opened 2026-02-23
docs: Fix breaking rendering issues

ChrisChinchilla:

YES


examples/user-management/react-user-management/.env.example
```diff
@@ -0,0 +1,2 @@
+VITE_SUPABASE_URL=
+VITE_SUPABASE_PUBLISHABLE_DEFAULT_KEY=
```
jeremenichelli, line 2: @ChrisChinchilla do we really want to delete these? 🤔

jeremenichelli (approved):

examples/user-management/react-user-management/.env.example
ChrisChinchilla: @jeremenichelli The file didn't exist at all, which is what broke the page render, as it was a file inclusion. I just left them blank for people to fill in. I am not sure if these .env files should even be part of those transcluded files really, but let's get this fixed now and think about it later.

PR #43101, opened 2026-02-23
Add a context menu item to copy a column name

djhi:
Small QOL improvement: a new context menu that allows to copy a column name.


apps/studio/components/grid/components/menu/ColumnMenu.tsx
```diff
@@ -96,6 +96,16 @@ export const ColumnMenu = ({ column, isEncrypted }: ColumnMenuProps) => {
         {snap.editable && (
           <>
             <DropdownMenuSeparator />
+            <DropdownMenuItem
```
ivasilov: You should move this in the dropdown items above `snap.editable` since it doesn't require the table to be editable.

djhi: Missed that! I wanted to have it below the separator

ivasilov (approved):

PR #43112, opened 2026-02-23
feat(studio): smart incident banner targeting

charislam:

YES

Feature enhancement — smarter incident banner targeting logic

Displaying the incident banner requires toggling a flag or environment variable. Banners are shown to all users regardless of whether their projects are in affected regions or whether the incident affects project creation.

## What is the new behavior?

Banner visibility is now driven by `show_banner` metadata from the StatusPage API — no manual flag or env var toggle needed. Per-user targeting is then applied:
- Users with projects only see the banner when they have a database in an affected region
- Users without projects only see the banner when the incident affects project creation

Incident responses are enriched with cache data (`affected_regions`, `affects_project_creation`) fetched from a Supabase table. Visibility logic is extracted into a dedicated hook and pure utility function, backed by unit tests.

## Additional context

It'll be easiest to test locally:

- Use this platform branch to get the right status page: https://github.com/supabase/platform/pull/29939
- Then change your environment variables so `NEXT_PUBLIC_SUPABASE_URL` and `LIVE_SUPABASE_SECRET_KEY` point to your local Supabase instance (make sure all migrations are pushed)
- Then you can toggle metadata directly in the database to observe the banner showing/not showing as appropriate

Resolves FE-2562

### DIFF
--- a/apps/studio/components/layouts/AppLayout/NoticeBanner.tsx
+++ b/apps/studio/components/layouts/AppLayout/NoticeBanner.tsx
@@ -1,10 +1,10 @@
-import { useRouter } from 'next/router'
-
-import { useLocalStorageQuery } from '@/hooks/misc/useLocalStorage'
 import { LOCAL_STORAGE_KEYS } from 'common'
-import { HeaderBanner } from 'components/interfaces/Organization/HeaderBanner'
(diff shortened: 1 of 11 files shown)


jordienr (approved):
cannot smoke test but overall code and tests look good

PR #43115, opened 2026-02-24
Show last migration's name instead of timestamp in new home page

joshenlim:
## Context

In the new project home page, we have a stat for "Last migration" which we're showing _when_ the last migration was applied. However:
- The timestamp for the migration is derived from the "version" column of the migration (in the `supabase_migrations` table) which afaik is derived from the migration's file name
  - It'll be alright if the migration was generated via the CLI, but we can't really enforce the name of the migration file if say they were generated via AI, so this is technically a point of flakiness
- Reckon that it's more value to show _what_ was the last migration rather than _when_ so opting to change the value here to show the name of the last migration instead

### Before
image

### After
image


apps/studio/components/interfaces/HomeNew/ActivityStats.tsx
```diff
@@ -37,14 +36,16 @@ export const ActivityStats = () => {
       )[0]
   }, [branchesData])
 
-  const { data: migrationsData, isPending: isLoadingMigrations } = useMigrationsQuery({
+  const { data: migrationsData = [], isPending: isLoadingMigrations } = useMigrationsQuery({
```
alaister: Lovely React footgun here
```suggestion
  const { data: migrationsData = EMPTY_ARR, isPending: isLoadingMigrations } = useMigrationsQuery({
```

alaister (approved):
LGTM!

- Tested on preview

One tiny nit:

PR #43116, opened 2026-02-24
Render email in UserDropdown if username is empty

joshenlim:
## Context

Just a tiny fix - the user dropdown doesn't show any user info if there's no username set, opting to show the email minimally

### Before
image

### After
image


dnywh (approved):
Tested locally. Looks good!

PR #43117, opened 2026-02-24
New home page, show service status as coming up if service health is unhealthy + project status is coming up

joshenlim:
## Context

With the new home page, we have a stat to show the overall status of the project + individual service health, both of which data is pulled from an endpoint
image

At the time of project creation, some of the services start of as UNHEALTHY, which results in the status stat here showing a label of "Unhealthy" (while accurate in the sense that we're showing what's coming from the API, its inaccurate as the services are still spinning up)

## Changes involved

For each service, am opting to show their status as "COMING_UP" if the service status is UNHEALTHY + the project's status is COMING_UP or UNKNOWN. The overall status label should also show "Coming up" if the project's status is COMING_UP or UNKNOWN as well

### DIFF
--- a/apps/studio/components/interfaces/HomeNew/ServiceStatus.tsx
+++ b/apps/studio/components/interfaces/HomeNew/ServiceStatus.tsx
@@ -291,7 +291,7 @@ export const ServiceStatus = () => {
                     isLoading={service.isLoading}
                     isProjectNew={isProjectNew}
                     status={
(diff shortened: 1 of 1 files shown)


dnywh (approved):
Tested locally. Looks good.

For AWS Revamped I saw “Coming up...”. For AWS I saw “Checking...” followed by “Coming up...”.

PR #43118, opened 2026-02-24
feat(studio): sortable projects

dnywh:

Feature. Resolves DEPR-390

Projects aren’t sortable in either the card or table view.

## What is the new behavior?

Projects are sortable in both:

- Card view: sort dropdown
- Table view: sort dropdown or table column headers

| Before | After |
| --- | --- |
| Supabase-4D1BFE40-875D-494C-8F17-A68D92826458 |  Supabase-D8C0AC7C-A28D-4AA6-BA7C-0FCD61DB5D11  |
| Supabase-0A545C5C-40B5-47F7-9ACD-2200879BB95E |  Supabase-0F7AB608-2E86-4F0C-BB60-C85D9B7F3D57 |

## Additional context

I wonder if this is overkill given most folks only have 1–2 projects. Some ideas:

- Only sortable in table view via column headers
- Conditional rendering for folks with 2+ projects
- Opt-in feature
- Feature-flag

I’ve opted to make it global and synced for now.


djhi (approved):

apps/studio/components/interfaces/Home/ProjectList/ProjectListSort.utils.ts
```diff
@@ -0,0 +1,53 @@
...
+] as const
+
+export type ProjectListSort = (typeof PROJECT_LIST_SORT_VALUES)[number]
+export type ProjectListSortableColumn = 'name' | 'created'
+
+export const PROJECT_LIST_SORT_LABELS: Record<ProjectListSort, string> = {
+  name_asc: 'Project name (A–Z)',
+  name_desc: 'Project name (Z–A)',
+  created_desc: 'Creation date (Newest first)',
+  created_asc: 'Creation date (Oldest first)',
```
joshenlim: Just a nit but these feel pretty verbose imo - wondering if we should do a nested dropdown for sort? we've got this for auth users
Image

we should also probably standardise our sort dropdowns too - e.g for storage explorer we've got it like this
Image

apps/studio/components/interfaces/Home/ProjectList/ProjectListSortDropdown.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const ProjectListSortDropdown = ({ value, onChange }: ProjectListSortDropdownProps) => {
+  return (
+    <DropdownMenu>
+      <DropdownMenuTrigger asChild>
+        <Button
+          type="default"
+          size="tiny"
+          icon={
+            value.endsWith('desc') ? (
+              <ArrowDownWideNarrow size={14} />
```
joshenlim: afaik we don't need to provide size for Button `icon` btw - can remove

apps/studio/components/interfaces/HomePageActions.tsx
```diff
@@ -91,6 +101,15 @@ export const HomePageActions = ({
           onSaveFilters={(options) => setFilterStatusStorage(options)}
         />
 
+        {showViewToggle && (
```
joshenlim: nit but i don't think we need this `showViewToggle` param tbh

`HomePageActions` is being used for the projects page + projects wildcard route - both UI could benefit with a sort action + card/list view i reckon

apps/studio/components/interfaces/Home/ProjectList/ProjectListSort.utils.ts
dnywh: Agreed. Discussing elsewhere

apps/studio/components/interfaces/HomePageActions.tsx
```diff
@@ -35,6 +40,10 @@ export const HomePageActions = ({
     'status',
     parseAsArrayOf(parseAsString, ',').withDefault([])
   )
+  const [sort, setSort] = useQueryState(
```
joshenlim, line 38: we should also persist this in local storage and load it up when the user comes back to the projects page

similar for how we handle the project status filter

apps/studio/components/interfaces/Home/ProjectList/ProjectListSort.utils.ts
```diff
@@ -0,0 +1,53 @@
...
+): 'ascending' | 'descending' | 'none' => {
+  if (column === 'name') {
+    if (currentSort === 'name_asc') return 'ascending'
+    if (currentSort === 'name_desc') return 'descending'
+    return 'none'
+  }
+
+  if (currentSort === 'created_asc') return 'ascending'
+  if (currentSort === 'created_desc') return 'descending'
+  return 'none'
```
joshenlim: this seems like it can be simplified - dont need to rely on column

apps/studio/components/interfaces/Home/ProjectList/ProjectListSort.utils.ts
```diff
@@ -0,0 +1,53 @@
...
+): ProjectListSort => {
+  if (column === 'name') {
+    if (currentSort === 'name_asc') return 'name_desc'
+    if (currentSort === 'name_desc') return 'name_asc'
+    return 'name_asc'
+  }
+
+  if (currentSort === 'created_desc') return 'created_asc'
+  if (currentSort === 'created_asc') return 'created_desc'
+  return 'created_desc'
```
joshenlim: hmm also feels like can be simplified to not have to rely on column here

joshenlim (approved):
Verified changes on staging - all looks good! 😄

PR #43123, opened 2026-02-24
feat(etl): Improve copy for advanced settings

iambriccardo:


joshenlim (approved):
Verified changes on preview! The advanced settings area is getting rather text heavy UX wise but alright for now. reckon might be better to have the in depth descriptions in our docs instead, and we add a link to the docs page from the dashboard

PR #43128, opened 2026-02-24
feat(docs): `token_endpoint_auth_method` in OAuth server docs

cemalkilic:

YES

Docs update

## Summary
The OAuth server supports three token endpoint authentication methods (`none`, `client_secret_basic`, `client_secret_post`), but the docs only showed `client_secret_post` implicitly without labeling it, and never mentioned client_secret_basic (the actual default for confidential clients per RFC 7591).

- Add `token_endpoint_auth_method` explanation with defaults/constraints to the client registration section in getting-started.mdx
- Update registration examples (JS, Python, cURL) and response JSON to include token_endpoint_auth_method
- Restructure token exchange and refresh token sections in oauth-flows.mdx to show all three auth methods with clear labels
- Add `client_secret_basic` examples using HTTP Basic auth header

### DIFF
--- a/apps/docs/content/guides/auth/oauth-server/getting-started.mdx
+++ b/apps/docs/content/guides/auth/oauth-server/getting-started.mdx
@@ -491,6 +491,22 @@ Store the client secret securely. It will only be shown once. If you lose it, yo
 
 </Admonition>
 
+#### Token endpoint authentication method
+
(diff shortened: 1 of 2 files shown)


fadymak (approved):
The changes LGTM, though I think we should consider breaking out the oauth-flows page into separate sub pages since it's gotten quite long and has repeated information

PR #43131, opened 2026-02-24
fix: universal link redirects to general

jordienr:
Currently `/org/_` redirects to `/general` (org settings) after picking an org, this makes it redirect to the project list

## to test
- go to /dashboard/org/_
- select org
- redirects to project picker 

## test 2
- go to /dashboard/org/_/general
- select org
- redirects to /org/whatever/general (settings)


apps/studio/components/interfaces/Organization/Organization.utils.ts
```diff
@@ -1,3 +1,23 @@
+/**
+ * Builds an org-scoped URL for the wildcard org route (/org/_).
+ *
+ * When `slug` is undefined the destination is the org project list at
+ * /org/<orgSlug>. When `slug` is an array the sub-path is preserved so the
+ * user lands on the equivalent page for the chosen org.
+ */
+export function buildOrgUrl(
```
ivasilov: I would expect the `orgSlug` to be a first param, then slug (maybe rename to `path` or `subPath`) and then params.

ivasilov (approved):

PR #43133, opened 2026-02-24
fix: remove disable SSL banner from UI

imor:
We now support SSL connections when installing stripe sync engine, so removing the banner asking users to disable SSL enforcement.


imor:
Closing as this change is included in https://github.com/supabase/supabase/pull/43289

PR #43137, opened 2026-02-24
[FE-2635] fix: fail build if assets upload fails

alaister:
Adds `set -eo pipefail` so any failed upload immediately exits the script with a non-zero code, failing the Vercel build.

### DIFF
--- a/scripts/upload-static-assets.sh
+++ b/scripts/upload-static-assets.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -eo pipefail
 
 #######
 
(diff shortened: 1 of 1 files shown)


scripts/upload-static-assets.sh
```diff
@@ -82,13 +83,11 @@ aws s3 sync "$PUBLIC_DIR" "s3://$BUCKET_NAME/$SITE_NAME/${VERCEL_GIT_COMMIT_SHA:
     --region auto \
     --only-show-errors
 
-if [ $? -eq 0 ]; then
```
alaister, line 85: I think this if is redundant because if we've made it this far we good

ivasilov (approved):

PR #43159, opened 2026-02-25
Update add-ons design

SaxonF:
**Old**

image

**New**

image

This updates the add-ons design to remove the Compute Size section which has now been moved and simplify the add-on list design.


apps/studio/components/interfaces/Settings/Addons/Addons.tsx
```diff
@@ -73,588 +70,323 @@ export const Addons = () => {
...
+      : projectUpdateDisabled
+        ? 'Project updates are currently disabled'
+        : !canUpdateIPv4 && !ipv4Enabled
+          ? 'You can only add IPv4 when your project network configuration is set to IPv6'
+          : undefined
 
-  return (
-    <>
-      <ScaffoldDivider />
+  const pitrDisabledReason = !isProjectActive
```
awaseem: These ternary operators are super hard to read, same with the ones above. Here is what we should do:
- Separate them out to utility functions and use if/else or early return statements
- Then write tests for all the cases

awaseem (changes requested):
Ternary operators need some clean up, everything else looks good!

apps/studio/components/interfaces/Settings/Addons/Addons.tsx
SaxonF: @awaseem should be good for a re-look pretty please

awaseem (approved):
LGTM

seanoliver (approved):
LGTM!

PR #43160, opened 2026-02-25
chore(studio): remove physical backups disabled download button 

dnywh:

UI update

- The vast majority of projects now have physical backups, not logical
- Physical backups are not downloadable
- But we show a disabled `Download` button for each physical backup
	- This button has confusing multi-step instructions about `pg_dump` that are [inaccurate](https://supabase.slack.com/archives/C02BJ2239GA/p1771979586829419?thread_ts=1771887878.203199&cid=C02BJ2239GA)

## What is the new behavior?

- We only show the (enabled) `Download` button for logical backups

| Before | After |
| --- | --- |
| AWS Healthy  Toolshed  Supabase-6FE7C23F-29D6-47B6-819A-84BC49927F3A | AWS Healthy  Toolshed  Supabase-C9E763B7-B447-468E-B928-BB9AD5ABC588 |

## Additional context

Support believes removing the button entirely will have less of a burden than a better tooltip explanation.

### DIFF
--- a/apps/studio/components/interfaces/Database/Backups/BackupItem.tsx
+++ b/apps/studio/components/interfaces/Database/Backups/BackupItem.tsx
@@ -88,8 +88,9 @@ export const BackupItem = ({ index, isHealthy, backup, onSelectBackup }: BackupI
 
   return (
     <div
-      className={`flex h-12 items-center justify-between px-6 ${index ? 'border-t border-default' : ''
-        }`}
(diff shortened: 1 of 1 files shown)


joshenlim (approved):
Changes LGTM on preview! JFYI tho, most of my projects only have physical backups (as expected) so i can't visibly check for logical backups, although code wise looks right

PR #43165, opened 2026-02-25
fix: theme switcher discoverability

dnywh:

Fix

The theme switcher in `CommandMenu` is not showing up when typing in “dark mode”, “toggle theme” or similar commands.

## What is the new behavior?

Added hidden root commands: `Toggle theme`, `Use dark theme`, `Use light theme`, `Use system theme`, plus existing `Switch theme...` with search aliases.

Also added `value` aliases to the theme items so searches pop up.


joshenlim (commented):
2 issues found while testing on staging preview:

- (Unrelated) The hover state of the CMD K item is likely the same color (?) as the item background color on light mode?
  Image
  - I can’t see the difference when my mouse is hovered over as compared to dark mode 
    Image

- Hitting toggle theme twice on the same browser session doesn’t change the theme on the 2nd time
  - e.g if im on dark mode originally and i hit toggle theme, it flips to light mode. but if i hit toggle theme again, it stays on light mode

joshenlim (approved):

SaxonF (approved):

PR #43167, opened 2026-02-25
Remove org slug and project ref filter for GET notifications request

joshenlim:
## Context

Since moving notifications to the Advisors Panel, we've been sending `org_slug` and `project_ref` to the GET notifications endpoint, which resulted in certain notifications not being returned such as those that are user specific (no org slug nor project ref)

Am opting to remove both slug and ref filters for the notifications as the notifications should be on a user level (irregardless if you're within a project or not) - the Advisor's Panel's button in the layout header would also suggest that notifications in there are not tied to an org or project

## To test

This one's a bit tricky to test unless you have notifications on staging, but i've double checked on prod with a curl command that removing the org slug and project ref filters returns the correct notifications


alaister (approved):
LGTM!

- Tested on preview

PR #43170, opened 2026-02-25
fix: add compact formatting to Y axis numbers in db charts

jordienr:
In the db charts, numbers in the Y axis are not formatted, sometimes these get too big and get cut-off. 

## before
(in this example it is not cutoff because the number is not big enough)
CleanShot 2026-02-25 at 11 20 35@2x

## after
CleanShot 2026-02-25 at 11 21 07@2x


joshenlim (approved):
Changes LGTM! I don't have big numbers on my staging projects either hahaha so not able to verify visually but code changes look good

PR #43171, opened 2026-02-25
chore: split tests running from coveralls upload

jordienr:
this is so we can easily tell when tests failed vs when coveralls failed in the PR checks / github action breakdown.

CleanShot 2026-02-25 at 11 28 00@2x


joshenlim (approved):
Approving to unblock - just need to double check though if the coverall results are getting uploaded properly as I expected the coveralls action to fail due to coverall's incident

PR #43173, opened 2026-02-25
Speed CI e2e tests by avoiding reseting the database

djhi:
## Problem

We currently initialise a local Supabase instance first (which apply migrations) and then reset it. Although that makes sense in a local development environment, it doesn't for CI as the instance will be completely new.

## Solution

Add a condition to avoid reseting the database when the `CI` environment variable isn't set.

### Locally

Run the following command on your machine:
```
pnpm run e2e:setup:cli
```

You should see the migrations being applied twice if the local instance wasn't running yet.

### In CI

The _Selfhosted Studio E2E Tests_ job in this PR checks should only apply the migrations once


djhi:
I need to modify something in the studio to make the job run

djhi:
Closed in favor of #43211

PR #43182, opened 2026-02-25
fix iops chart max value

jordienr:
The Y axis domain/range was being calculated incorrectly. 

Testing this is a bit painful. You need to: 
- using tweak-extension or similar overwrite the API response for the IOPS chart and add a really big value 
- reload the page
- the Y axis should adapt to that really big value even if it goes over the disk max iops

See screenshots below

## before
- big data point wouldn't change the chart range 
- Y Axis max is 6k even tho we have a data point over 20k
CleanShot 2026-02-25 at 18 47 14@2x

## after
- the Y axis shows the correct range
CleanShot 2026-02-25 at 18 47 57@2x


charislam (approved):
Tested on staging with a network override

PR #43183, opened 2026-02-25
fix(Config): Change strong tag font weight to supported font spec

jeremenichelli:

Changes the base font weight for strong tags within the tailwind/typography configuration.

I noticed that texts within our strong tags had an irregular stroke, and problems on rendering. The main problem is we are using the default 600 weight value coming from the config of tailwind, while our Circular custom font supports 400 and 600.

_See the first text in the list from the screen just, like "Multi Protocol"._

Screenshot 2026-02-25 at 16 10 29

## What is the new behavior?

Strong tags now render with 500 font weight to prevent bad text rendering.

Screenshot 2026-02-25 at 16 12 56


ChrisChinchilla:
Fine with me, we may want @supabase/design to check too?

ChrisChinchilla (approved):

kemaldotearth (approved):

PR #43189, opened 2026-02-25
fix(proxy): avoid unconditional NextResponse.next() in Studio middleware

ivasilov:
Summary
- Fixes full page reloads on every client-side navigation in Studio caused by the first-referrer cookie middleware (#43153)
- In the multi-zone production setup (www proxies /dashboard/* → Studio), returning `NextResponse.next()` unconditionally processes every response through Next.js's middleware pipeline, interfering with client-side navigation
- Now only returns NextResponse.next() when a cookie actually needs to be stamped; returns undefined otherwise so Next.js handles the request untouched
- Also fixes valid API routes (e.g. /api/ai/sql/generate-v4) falling through to cookie-stamping code instead of passing through cleanly

### DIFF
--- a/apps/studio/proxy.ts
+++ b/apps/studio/proxy.ts
@@ -1,7 +1,12 @@
-import { stampFirstReferrerCookie } from 'common/first-referrer-cookie'
-import { IS_PLATFORM } from 'lib/constants'
+import {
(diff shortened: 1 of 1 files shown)


charislam (approved):
🤞🏼

PR #43201, opened 2026-02-26
chore(studio): standardise discard changes behaviour 

dnywh:

UX consistency improvement. Updates DEPR-355.

Discard-confirm close behavioir is implemented inconsistently across Studio forms:

- some sheets/dialogs used `useConfirmOnClose`
- some duplicated local `CloseConfirmationModal` components
- some (e.g. `CreateHookSheet`) closed unconditionally and could lose unsaved changes

## What is the new behavior?

Extracts and validates a reusable discard-close pattern for dialogs/sheets
- enhances `useConfirmOnClose` with `handleOpenChange(open)` for `Dialog`/`Sheet` `onOpenChange`
- adds shared `DiscardChangesConfirmationDialog` (`AlertDialog`-based, override-able copy)
- migrates:
	- `InviteMemberButton`
	- `CreateHookSheet`
	- `EditSecretSheet`

This standardizes close-guard behavior for backdrop/escape/close-button/cancel-button flows without trying to block route changes or arbitrary unmounts.

## Additional context

`CreateHookSheet` now also marks the generated secret action as dirty (`setValue(..., { shouldDirty: true })`) so the discard guard behaves correctly.

- Added tests for `useConfirmOnClose` covering:
	- clean vs dirty close
	- handleOpenChange(true|false)
	- confirm/cancel behavior
	- latest callback ref behavior

A follow-up PR is needed to migrate remaining duplicated `CloseConfirmationModal` usages and older `useConfirmOnClose` call sites to the shared `DiscardChangesConfirmationDialog` + `handleOpenChange` pattern.


apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
dnywh, line 1: Codex spotted some additional flaws in this dialog’s dirty-state edge case handling. Hence the additional touches.

> Findings
> 
> High: async role initialization can clear dirty state and disable the discard prompt
> form.reset({ ...form.getValues(), role }) in [InviteMemberButton.tsx (line 231)](app://-/index.html#) (specifically [InviteMemberButton.tsx (line 235)](app://-/index.html#)) will re-baseline the form if roles finish loading after the dialog is already open.
> Result: if a user starts typing emails before isSuccess flips, reset() clears formState.isDirty, so a subsequent backdrop/escape close can skip the new “Discard changes?” guard and lose typed input.
> 
> High: role/scope-only edits may bypass the discard prompt because isDirty may never flip
> The new close logic depends on form.formState.isDirty ([InviteMemberButton.tsx (line 244)](app://-/index.html#)), but role and applyToOrg are updated with [form.setValue(...)](app://-/index.html#) without shouldDirty: true at [InviteMemberButton.tsx (line 346)](app://-/index.html#) and [InviteMemberButton.tsx (line 387)](app://-/index.html#).
> In RHF, that can leave isDirty false, so changing only the role/switch can still close immediately and lose changes.

apps/studio/components/interfaces/Organization/TeamSettings/InviteMemberButton.tsx
```diff
@@ -72,13 +66,19 @@ export const InviteMemberButton = () => {
...
 
   const [isOpen, setIsOpen] = useState(false)
-  const [isDiscardConfirmOpen, setIsDiscardConfirmOpen] = useState(false)
   const [projectDropdownOpen, setProjectDropdownOpen] = useState(false)
 
   const { data: members } = useOrganizationMembersQuery({ slug })
   const { data: allRoles, isSuccess } = useOrganizationRolesV2Query({ slug })
   const orgScopedRoles = allRoles?.org_scoped_roles ?? []
 
+  const getInviteFormDefaults = () => ({
```
joshenlim: this can be a static object yeah?

joshenlim (approved):
Verified changes locally!

PR #43204, opened 2026-02-26
Tiny fix for notifications red dot

joshenlim:
## Context

Should've removed the `status` prop when fetching notifications in `AdvisorButton` for a more accurate indication of `hasCriticalNotifications` as we should show that dot for read critical notifications too (not just new)


alaister (approved):
LGTM!

- Tested on preview

PR #43210, opened 2026-02-26
fix: custom report gets stuck block after deleting sql snippet

jordienr:
- Deleted SQL Snippets leave a hanging block that loads forever in custom reports, and its not possible to delete them. 
- Now you can delete blocks if they get stuck loading
- Also shows correct error state when a block couldn't load because the sql snippet was removed
## before
- stuck forever 
CleanShot 2026-02-26 at 13 23 25@2x

## after
- show error state
- allow user to delete snippet
CleanShot 2026-02-26 at 13 23 45@2x


joshenlim (approved):
Verified changes locally!

PR #43218, opened 2026-02-26
feat: copy multiple logs

jordienr:
Allows users to select logs and copy their contents for debugging, or pass them on to assistant with one click.

## To test
- go to logs or log explorer
- select some logs
- try copying as json, markdown, or sending it to assistant.


packages/ui/src/components/shadcn/ui/checkbox.tsx
```diff
@@ -23,7 +23,7 @@ const Checkbox = React.forwardRef<
     {...props}
   >
     <CheckboxPrimitive.Indicator className={cn('flex items-center justify-center text-current')}>
-      <Check className="h-3 w-3" strokeWidth={4} />
+      <Check className="h-3 w-3 text-background" strokeWidth={4} />
```
jordienr, line 26: the ✔︎ wont show up without this

apps/studio/components/interfaces/Settings/Logs/Logs.utils.ts
```diff
@@ -756,3 +756,91 @@ export function role(metadata: any) {
...
+        lines.push(`**Timestamp:** ${tsString}`)
+      }
+      if (row.event_message) {
+        lines.push(`**Message:** ${row.event_message}`)
+      }
+      const { id: _id, timestamp: _ts, event_message: _msg, ...rest } = row as any
+      if (Object.keys(rest).length > 0) {
+        lines.push('', '**Details:**', '```json', JSON.stringify(rest, null, 2), '```')
+      }
+      return lines.join('\n')
```
awaseem, line 774: nit: Can we refactor this innner map function out? Its hard to read. Then we can test a single unit instead of the full array operations

apps/studio/components/interfaces/Settings/Logs/MultiSelectActionBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+  sqlQuery,
+}: MultiSelectActionBarProps) {
+  const { openSidebar } = useSidebarManagerSnapshot()
+  const aiSnap = useAiAssistantStateSnapshot()
+
+  function handleOpenAiAssistant() {
+    const prompt = buildLogsPrompt(selectedRowsData, queryType, sqlQuery)
+    openSidebar(SIDEBAR_KEYS.AI_ASSISTANT)
+    aiSnap.newChat({ initialMessage: prompt })
+  }
```
awaseem, line 44: nit: lets use annon function here? Its the standard in other components

apps/studio/components/interfaces/Settings/Logs/LogTable.tsx
```diff
@@ -1,15 +1,16 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
```
awaseem, line 1: @jordienr This component is quite big and needs a clean. Couple of things:
- Refactor some of the rendercell as new components
- Multiple use effects and use callbacks and be refactored to hooks and then referenced here

Its more work, but we should do it post merge IMO

awaseem (approved):
Minor changes, some we can do post

kemaldotearth (approved):
Works as expected :shipit:

PR #43222, opened 2026-02-26
fix: resolve undefined project ref in TabsStateContextProvider

ivasilov:
This pull request refactors how the `TabsStateContextProvider` receives the project reference and updates related imports for consistency and maintainability. The main change is to pass `projectRef` explicitly as a prop instead of fetching it internally, which improves context control and makes the component easier to test and reuse. Additionally, the PR updates import paths to use absolute aliases and removes an unused function.

This fixes an issue which you can replicate by:
1. Go to SQL editor
2. Open any snippet
3. Delete the local storage `supabase_studio_tabs_{project ref}`
4. Refresh the page while still the snippet is open

This will make the snippet to enter in a ghost state where the tab name is not visible but you see the content.


joshenlim (approved):
Verified changes locally for both SQL Editor and Table Editor where the tabs UI is being used

PR #43228, opened 2026-02-27
chore(studio): move organization settings nav to sidebar layout

dnywh:

UI change

Organization settings are currently spread across a NavMenu. This is getting unweildy as the amount of children within Organization settings grows.

## What is the new behavior?

Organization settings are now on a sidebar, just like Project Settings and most other surfaces.

| Before | After |
| --- | --- |
| Supabase | General  Supabase |

## Additional context

@fsansalvadore and the design team have have other efforts going on to reduce interface clutter, e.g. nested sidebars.

### DIFF
--- a/apps/studio/components/interfaces/Organization/AuditLogs/AuditLogs.tsx
+++ b/apps/studio/components/interfaces/Organization/AuditLogs/AuditLogs.tsx
@@ -163,7 +163,7 @@ export const AuditLogs = () => {
 
   if (isLogsNotAvailableBasedOnPlan) {
     return (
-      <ScaffoldContainer>
+      <ScaffoldContainer className="px-6 xl:px-10">
         <ScaffoldSection isFullWidth>
           <UpgradeToPro
             plan="Team"
@@ -179,7 +179,7 @@ export const AuditLogs = () => {
 
   return (
     <>
-      <ScaffoldContainer>
+      <ScaffoldContainer className="px-6 xl:px-10">
         <ScaffoldSection isFullWidth>
           <div className="space-y-4 flex flex-col">
             {showFilters && (
--- a/apps/studio/components/interfaces/Sidebar.tsx
+++ b/apps/studio/components/interfaces/Sidebar.tsx
@@ -372,6 +372,14 @@ const OrganizationLinks = () => {
   const showBilling = useIsFeatureEnabled('billing:all')
 
   const activeRoute = router.pathname.split('/')[3]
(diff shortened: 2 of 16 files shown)


apps/studio/components/interfaces/Sidebar.tsx
```diff
@@ -409,7 +409,7 @@ const OrganizationLinks = () => {
         ]
       : []),
     {
-      label: 'Organization settings',
+      label: 'Organization Settings',
```
dnywh, line 420: We use title case on all sidebars. This one was inconsistent

apps/studio/components/interfaces/Sidebar.tsx
```diff
@@ -428,11 +436,7 @@ const OrganizationLinks = () => {
               i === 0
                 ? activeRoute === undefined
                 : item.key === 'settings'
-                  ? router.pathname.includes('/general') ||
-                    router.pathname.includes('/apps') ||
-                    router.pathname.includes('/audit') ||
-                    router.pathname.includes('/documents') ||
-                    router.pathname.includes('/security')
+                  ? organizationSettingsRoutes.has(activeRoute ?? '')
```
dnywh, line 439: Fixes an issue where the SSO page was deselecting the Organisation Settings sidebar item

apps/studio/components/layouts/ProjectSettingsLayout/SettingsMenu.utils.tsx
dnywh, line 1: Renamed section titles to match Organization Settings. _Configuration_ makes more sense than _Project Settings_ since that's already the section title... and page title for /genereal.

apps/studio/components/interfaces/Organization/AuditLogs/AuditLogs.tsx
```diff
@@ -163,7 +163,7 @@ export const AuditLogs = () => {
 
   if (isLogsNotAvailableBasedOnPlan) {
     return (
-      <ScaffoldContainer>
+      <ScaffoldContainer className="px-6 xl:px-10">
```
dnywh, line 166: This change and the same ones repeated throughout are necessary to accommodate new `PageLayout` wrapper around it. Otherwise their paddings don’t match and these bits stick out horizontally.

The proper solution is replacing `ScaffoldContainer` with newer components, but I don’t want to make a huge PR. One for another time.

djhi (approved):

jordienr (approved):

PR #43184, opened 2026-02-25
feat(studio): mobile sheet nav

fsansalvadore:
This is the first step towards a new Mobile navigation experience. See [Linear](https://linear.app/supabase/project/mobile-navigation-v2-38776138da76/overview).

Main changes:
- use `MobileSheetNav` for all `LayoutSidebar` panels on mobile viewport
- move `Help` to LayoutSidepanel for a more cohesive experience
- you can now open the help panel using params `?sidebar=help-panel`

Out of scope for this pr is changing the mobile nav layout and positioning of the icons, which will come in follow-up PRs.

https://github.com/user-attachments/assets/a84e56f0-f2e2-4217-89a2-ba895bb7d352

PR #43239, opened 2026-02-27
fix: mobile sheet nav close

fsansalvadore:
Mobile sheet nav doesn't close on route change and viewport resizing after #43184.

This PR fixes that and also makes it optional to close on nav and resizing, because the sidepanels need them _not_ to resize, while mobile navigation menu does.

Also added some tests. 🪄 

## Before

https://github.com/user-attachments/assets/da13bbb6-9ba4-403a-a153-4734458fc5c1

## After

https://github.com/user-attachments/assets/32bf1e66-d3a0-45c5-8d35-2e5345064bcf

To test that it works:
- open dashboard on desktop and see that if you open sidepanels, they remain open both on desktop and mobile viewports, just changing components (right sidepanel vs bottom mobile sheet)
- navigating using the mobile viewport navigation menus should
  - close the sheet after a successful navigation
  - close the sheet on viewport resize


kemaldotearth (approved):
Works as described :shipit:

PR #43244, opened 2026-02-27
chore: mobile input font-size

fsansalvadore:
Bump font-size to 16px for inputs and textareas on mobile viewports to avoid auto zoom-in when focusing the inputs.

The alternative would be to add `user-scalable=no;user-scalable=0;` to the "viewport" meta html tag, but it wouldn't be a good ux practice and it would worsen a11y.

Left the `tiny` sizes out as the're a minor case and would need more design considerations.
Tiny should probably never be used on mobile but I'll leave that out of the scope of this PR.

This PR fixes the issue on the most important CUFs.

## Before

https://github.com/user-attachments/assets/1779e05e-ff1a-411a-9997-1ad129b1e84a

## After

https://github.com/user-attachments/assets/ed7dbb0b-99b4-46b5-8cc7-9ee9adad5851


kemaldotearth (approved):
Approving, but may need deeper testing before merge 🙏

PR #43249, opened 2026-02-27
feat: restore ability to dismiss incident banners

charislam:

YES

Feature restoration

Incident banners cannot be dismissed by users.

## What is the new behavior?

Users can dismiss incident banners again.

## Additional context

### DIFF
--- a/apps/studio/components/layouts/AppLayout/StatusPageBanner.tsx
+++ b/apps/studio/components/layouts/AppLayout/StatusPageBanner.tsx
@@ -16,5 +16,12 @@ export const StatusPageBanner = () => {
 
   if (!banner) return null
 
-  return <HeaderBanner variant="warning" title={banner.title} description={BANNER_DESCRIPTION} />
+  return (
+    <HeaderBanner
+      variant="warning"
+      title={banner.title}
+      description={BANNER_DESCRIPTION}
+      onDismiss={banner.dismiss}
+    />
+  )
 }
--- a/apps/studio/components/layouts/AppLayout/StatusPageBanner.utils.ts
+++ b/apps/studio/components/layouts/AppLayout/StatusPageBanner.utils.ts
@@ -1,6 +1,6 @@
 import type { IncidentCache } from 'lib/api/incident-status'
 
-type BannerIncident = { cache?: IncidentCache | null }
+type BannerIncident = { id: string; cache?: IncidentCache | null }
 
 /**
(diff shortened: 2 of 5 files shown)


apps/studio/components/layouts/AppLayout/StatusPageBanner.utils.ts
joshenlim, line 33: (Unrelated to this PR)

Reckon we should show this only if the user is on the new project page? Might be misleading if the user has projects, and is trying to create more projects nonetheless

apps/studio/components/layouts/AppLayout/StatusPageBanner.tsx
```diff
@@ -16,5 +16,12 @@ export const StatusPageBanner = () => {
 
   if (!banner) return null
 
-  return <HeaderBanner variant="warning" title={banner.title} description={BANNER_DESCRIPTION} />
+  return (
+    <HeaderBanner
+      variant="warning"
+      title={banner.title}
+      description={BANNER_DESCRIPTION}
+      onDismiss={banner.dismiss}
```
joshenlim, line 24: (nit, maybe unnecessary) am wondering if this needs to be conditional in the case if the incident banner was toggled without a corresponding incident from the API and FE doesnt have context of incident IDs

In which case hitting dismiss wouldnt do anything

e.g if we're using the override (flag or env var) because the API to retrieve incidents is somehow down, or if configcat is down and we need to use the env var

joshenlim (approved):
Code changes LGTM, but not able to test on staging

apps/studio/components/layouts/AppLayout/StatusPageBanner.utils.ts
```diff
@@ -42,3 +42,27 @@ export function shouldShowBanner({
     return affectedRegions.some((region) => userRegions.has(region))
```
joshenlim, line 42: just considering a bit more nuance RE new project page - thinking we should somehow surface on the new projects page if the user is creating a project in a region that's affected? (not necessarily if the incident is affecting project creation)

imo it's a bit of an odd UX if say
- there's an incident in ap-south-1, but i have no projects in that region so i don't see that banner
- I create a project in ap-south-1
- Once i reaching the project building state, i now see a banner "we're investigating a technical issue"

It might help as well to render the incident title in the banner? instead of just "investigating a technical issue"

PR #43280, opened 2026-03-02
[FE-2158] – feat(studio): Add exposed tables config to Postgrest settings

alaister:
Adds a feature flagged exposed tables config to postgrest settings:
Screenshot 2026-03-02 at 17 04 13

To test:
- make sure the existing (feature flag off) settings work as expected
- make sure exposing and removing schemas works in the new mode
- make sure exposing and removing tables works in the new mode
- ideally try with a lot of tables to check search and infinite scroll work as expected
- try exposing a schema without any tables (like `graphql_public`) and make sure it doesn't get removed randomly when editing tables
- try with a table with custom permissions (for example `REVOKE SELECT ON public.posts FROM anon;`) and make sure the user is informed with the tooltip


apps/studio/components/interfaces/Settings/API/ExposedSchemaSelector.tsx
```diff
@@ -0,0 +1,152 @@
...
+  })
+
+  const schemas = useMemo(
+    () =>
+      (allSchemas ?? [])
+        .filter((s) => {
+          if (s.name === 'graphql_public') return true
+          return !INTERNAL_SCHEMAS.includes(s.name)
+        })
+        .sort((a, b) => a.name.localeCompare(b.name)),
```
ivasilov, line 56: This is mutating the RQ internal state, right? use `[...allSchemas ?? []]`

ivasilov (approved):

apps/studio/components/interfaces/Settings/API/ExposedSchemaSelector.tsx
charislam: Shouldn't be, because the .filter returns a new array

apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -77,89 +100,177 @@ export const PostgrestConfig = () => {
...
+
+  const isLoading =
+    isLoadingConfig || isLoadingSchemas || (isApiGrantTogglesEnabled && isLoadingExposeMode)
+
+  const schemas = useMemo(
+    () =>
+      allSchemas
+        .filter((x) => {
+          const find = indexOf(INTERNAL_SCHEMAS, x.name)
+          if (find < 0) return x
```
charislam:
```suggestion
          return !INTERNAL_SCHEMAS.some((schema) => schema === x.name)
```

not sure my syntax is 100% right without intellisense lol, but should do the same thing

apps/studio/data/privileges/exposed-tables-infinite-query.ts
```diff
@@ -0,0 +1,151 @@
...
+    page = 0,
+    limit = EXPOSED_TABLES_PAGE_LIMIT,
+  }: ExposedTablesVariables & { page?: number; limit?: number },
+  signal?: AbortSignal
+): Promise<ExposedTablesResponse> {
+  if (!projectRef) throw new Error('projectRef is required')
+
+  const offset = page * limit
+
+  const sql = /* SQL */ `
```
charislam: we should move SQL to `data/**/*.sql.ts` so it's co-owned by postgres team and support team can scrape it for their docs

alaister: Done 🙏🏻

apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -77,89 +94,135 @@ export const PostgrestConfig = () => {
...
+        })
+      }
+
+      await updatePostgrestConfig(
+        {
+          projectRef,
+          dbSchema,
+          maxRows: values.maxRows,
+          dbExtraSearchPath: values.dbExtraSearchPath.join(','),
+          dbPool: values.dbPool ? values.dbPool : null,
```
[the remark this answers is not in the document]
joshenlim: this is alright - we validate the form against a 0 value for this field in particular

apps/studio/data/privileges/update-exposed-tables-mutation.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import type { ConnectionVars } from '../common.types'
+import { getExposedSchemasSql } from './privileges.sql'
+
+export type UpdateExposedTablesVariables = ConnectionVars & {
+  tableIdsToAdd: number[]
+  tableIdsToRemove: number[]
+}
+
+const buildTablePrivilegesSql = (oids: number[], action: 'grant' | 'revoke') => {
```
joshenlim, line 14: nit but these could be in privilege.sql.ts too

apps/studio/components/interfaces/Settings/API/ExposedSchemaSelector.tsx
```diff
@@ -0,0 +1,152 @@
...
+} from 'ui'
+import { ShimmeringLoader } from 'ui-patterns/ShimmeringLoader'
+
+import { useSchemasQuery } from '@/data/database/schemas-query'
+import { useSelectedProjectQuery } from '@/hooks/misc/useSelectedProject'
+import { INTERNAL_SCHEMAS } from '@/hooks/useProtectedSchemas'
+import { pluralize } from '@/lib/helpers'
+
+interface ExposedSchemaSelectorProps {
+  className?: string
```
joshenlim: this isn't being used right? iinw - only PostgrestConfig is importing this and isn't passing a class name

apps/studio/components/interfaces/Settings/API/ExposedSchemaSelector.tsx
```diff
@@ -0,0 +1,152 @@
...
+        })
+        .sort((a, b) => a.name.localeCompare(b.name)),
+    [allSchemas]
+  )
+
+  const selectedSet = useMemo(() => new Set(selectedSchemas), [selectedSchemas])
+  const selectedCount = schemas.filter((s) => selectedSet.has(s.name)).length
+
+  return (
+    <div className={className}>
```
joshenlim: from [this](https://github.com/supabase/supabase/pull/43280/changes#r2894309580) - then this div is unnecessary then

joshenlim:
this one's just a UI feedback though - any reason why we aren't using MultiSelector component for `ExposedSchemaSelector` and `ExposedTablesSelector`? Otherwise i'd need to open the popover to know what schemas are exposed, whereas it'll be better to have that info upfront (the x of y exposed text could be a description instead)

Could be same UX as per Extra search path field tho probably needs to be extended for the disabled state and tooltips
image

joshenlim:
Unrelated to your changes but noticed this overflow hidden behaviour happening - if it might be an easy fix
image

apps/studio/data/privileges/update-exposed-tables-mutation.ts
alaister: Will address in https://github.com/supabase/supabase/pull/43478 👍🏻

apps/studio/data/privileges/privileges.sql.ts
```diff
@@ -0,0 +1,161 @@
...
+
+export function getExposedTableCountsSql({ selectedSchemas }: { selectedSchemas: string[] }) {
+  const schemasList =
+    selectedSchemas.length > 0 ? selectedSchemas.map((s) => `'${s}'`).join(', ') : "''"
+
+  return /* SQL */ `
+    with ${getTableGrantsCTEs()}
+    select
+      (select count(*)::int from table_grants) as total_count,
+      (select count(*)::int from table_grants where status = 'granted' and schema_name in (${schemasList})) as grants_count;
```
pgnickb: this can be simplified to:
```suggestion
 select
      count(*)::int as total_count,
      (count(*)filter where status = 'granted' and schema_name in (${schemasList})))::int as grants_count
  from table_grants
```

It's a bit cheaper, as the whole thing does a single pass on the CTE table.

apps/studio/data/privileges/privileges.sql.ts
```diff
@@ -0,0 +1,161 @@
...
+  `
+}
+
+export function getExposedSchemasSql() {
```
pgnickb, line 131: As per offline discussion, to be removed completely.

apps/studio/data/privileges/update-exposed-tables-mutation.ts
```diff
@@ -0,0 +1,87 @@
...
+        select quote_ident(n.nspname) || '.' || quote_ident(c.relname) as relation
+        from pg_class c
+        join pg_namespace n on n.oid = c.relnamespace
+        where c.oid in (${oids.join(', ')})
+      loop
+        execute format('${privilegeClause}', rec.relation);
+      end loop;
+    end $$;
+  `
+}
```
pgnickb, line 38: nit:
```suggestion
  const privilegeClause =
    action === 'grant'
      ? 'grant select, insert, update, delete on table %I.%I to anon, authenticated, service_role'
      : 'revoke all on table %I.%I from anon, authenticated, service_role'

  return /* SQL */ `
    do $$
    declare
      relname name;
      nspname name;
    begin
      for nspname, relname in
        select n.nspname, c.relname
        from pg_class c
        join pg_namespace n on n.oid = c.relnamespace
        where c.oid in (${oids.join(', ')})
      loop
        execute format('${privilegeClause}', relname, nspname);
      end loop;
    end $$;
  `
}
```

pgnickb (approved):

PR #43281, opened 2026-03-02
fix(studio): add option to treat empty CSV cells as NULL on import

### LINKED ISSUE #43258: It's not possible to import CSV with empty string value
# Bug report

## Describe the bug

It's not currently possible to import a CSV file in Supabase Studio with a genuine empty string value, which is not the same as a NULL value. It treats empty cells as null, but it also treats NULL as null, so there's two ways to enter a null value and no way to enter an empty string.

## To Reproduce

1. Go to Supabase Studio.
2. Open the table editor with some table.
3. Insert -> import CSV and import a file with some empty cells.
4. See that the imported data has empty cells set to NULL (even in the preview).

## Expected behavior

It should treat empty cells as empty strings or at least treat EMPTY as empty strings (which is how they're displayed in the table editor). But in the latter case you'd have to have a way to escape that to actually enter the value "EMPTY".

usamanadeemdeveloper:
When importing a CSV file, empty cells were always imported as empty strings with no way to import NULL values instead. This adds a "Treat empty cells as NULL" checkbox to the import configuration panel.

When enabled, PapaParse's transform option converts empty strings to null before the data is parsed and previewed, so the imported rows correctly contain NULL rather than empty strings.

Fixes #43258.

YES

Bug fix

Empty cells in a CSV file are always imported as empty strings with no way to import NULL values instead. Fixes #43258.

## What is the new behavior?

"Treat empty cells as NULL" checkbox is added to the import configuration panel. When enabled, empty cells are imported as NULL instead of empty strings. Toggling the checkbox instantly re-parses the preview.

## Additional context

The fix uses PapaParse's transform option to convert empty strings to null before parsing. Applies to both file uploads and pasted text.

 ## Screenshots                                                                
  
**Checkbox enabled — empty cells show as NULL in preview:** 
csv-import-step2-checkbox-with-null-preview


usamanadeemdeveloper:
Hi team, I believe all the CodeRabbit feedback has been addressed. Could someone take a look when you get a chance? Happy to make any further changes.

Thanks!

usamanadeemdeveloper:
@coderabbitai review

usamanadeemdeveloper:
Hi team 👋

Just a quick update, this PR is ready for review. I've added a screenshot to the description showing the new "Treat empty cells as NULL" checkbox in action, with the live preview correctly displaying NULL for empty cells.

To summarise what's been done:
- Added a **"Treat empty cells as NULL"** checkbox to the CSV import configuration panel
- Uses PapaParse's `transform` option to convert empty strings to `null` before parsing
- Works for both file uploads and pasted text
- Toggling the checkbox instantly re-parses the preview

This fixes #43258. Happy to make any adjustments if needed. Thanks for your time and for maintaining such a great project!

usamanadeemdeveloper:
@coderabbitai resume

charislam (approved):
Thanks @usamanadeemdeveloper! Took the opportunity to do some refactoring of this area, but kept your original changes with the fix!

mintydev789:
Was this implemented as a global thing for thr whole table? Because I think it should be a per-column configuration.

usamanadeemdeveloper:
> Was this implemented as a global thing for thr whole table? Because I think it should be a per-column configuration.

I dont understand what you are saying can you clarify?

mintydev789:
> > Was this implemented as a global thing for thr whole table? Because I think it should be a per-column configuration.
> 
> I dont understand what you are saying can you clarify?

So when you import a CSV, can you set if empty cells are treated as null or as an empty string per-column? Or is it applied to all columns in the csv table?

usamanadeemdeveloper:
> > > Was this implemented as a global thing for thr whole table? Because I think it should be a per-column configuration.
> > 
> > 
> > I dont understand what you are saying can you clarify?
> 
> So when you import a CSV, can you set if empty cells are treated as null or as an empty string per-column? Or is it applied to all columns in the csv table?

Thanks for clarifying currently this is implemented as a global setting, so it applies to all columns in the CSV.

The goal here was to address the immediate limitation where empty cells could only be imported as empty strings, and provide a simple toggle without adding too much complexity to the import flow.

I agree that a per-column configuration would be more flexible, especially for cases where empty strings are meaningful in some fields but not others.

That said, implementing per-column control would require extending the import UI and handling column-level transforms, which felt a bit out of scope for this fix.

Happy to explore that as a follow-up improvement if that aligns with the expected behavior.

mintydev789:
> > > > Was this implemented as a global thing for thr whole table? Because I think it should be a per-column configuration.
> > > 
> > > 
> > > I dont understand what you are saying can you clarify?
> > 
> > So when you import a CSV, can you set if empty cells are treated as null or as an empty string per-column? Or is it applied to all columns in the csv table?
> 
> Thanks for clarifying currently this is implemented as a global setting, so it applies to all columns in the CSV.
> 
> The goal here was to address the immediate limitation where empty cells could only be imported as empty strings, and provide a simple toggle without adding too much complexity to the import flow.
> 
> I agree that a per-column configuration would be more flexible, especially for cases where empty strings are meaningful in some fields but not others.
> 
> That said, implementing per-column control would require extending the import UI and handling column-level transforms, which felt a bit out of scope for this fix.
> 
> Happy to explore that as a follow-up improvement if that aligns with the expected behavior.

Yeah, I think it would be good to extend it, as it currently relies on the same setting working for all columns in the table

usamanadeemdeveloper:
> > > > > Was this implemented as a global thing for thr whole table? Because I think it should be a per-column configuration.
> > > > 
> > > > 
> > > > I dont understand what you are saying can you clarify?
> > > 
> > > 
> > > So when you import a CSV, can you set if empty cells are treated as null or as an empty string per-column? Or is it applied to all columns in the csv table?
> > 
> > 
> > Thanks for clarifying currently this is implemented as a global setting, so it applies to all columns in the CSV.
> > The goal here was to address the immediate limitation where empty cells could only be imported as empty strings, and provide a simple toggle without adding too much complexity to the import flow.
> > I agree that a per-column configuration would be more flexible, especially for cases where empty strings are meaningful in some fields but not others.
> > That said, implementing per-column control would require extending the import UI and handling column-level transforms, which felt a bit out of scope for this fix.
> > Happy to explore that as a follow-up improvement if that aligns with the expected behavior.
> 
> Yeah, I think it would be good to extend it, as it currently relies on the same setting working for all columns in the table

OK I will work on it

mintydev789:
> > > > > > Was this implemented as a global thing for thr whole table? Because I think it should be a per-column configuration.
> > > > > 
> > > > > 
> > > > > I dont understand what you are saying can you clarify?
> > > > 
> > > > 
> > > > So when you import a CSV, can you set if empty cells are treated as null or as an empty string per-column? Or is it applied to all columns in the csv table?
> > > 
> > > 
> > > Thanks for clarifying currently this is implemented as a global setting, so it applies to all columns in the CSV.
> > > The goal here was to address the immediate limitation where empty cells could only be imported as empty strings, and provide a simple toggle without adding too much complexity to the import flow.
> > > I agree that a per-column configuration would be more flexible, especially for cases where empty strings are meaningful in some fields but not others.
> > > That said, implementing per-column control would require extending the import UI and handling column-level transforms, which felt a bit out of scope for this fix.
> > > Happy to explore that as a follow-up improvement if that aligns with the expected behavior.
> > 
> > Yeah, I think it would be good to extend it, as it currently relies on the same setting working for all columns in the table
> 
> OK I will work on it

Thanks!

usamanadeemdeveloper:
@mintydev789 do we need to open issue again or what? are you gonna open it?

usamanadeemdeveloper:
@charislam I think for the fix what @mintydev789 is suggesting we need a multi select dropdown where we list down all column names and checking them true will set that column only null, so that way we have control over the column level!
What you suggest on this?

something like the below image

col-level-null

mintydev789:
> @mintydev789 do we need to open issue again or what? are you gonna open it?

You can probably reopen the same issue

charislam:
@usamanadeemdeveloper that makes sense, feel free to submit another PR if you want to tackle these changes!

usamanadeemdeveloper:
@charislam
This is to inform that the new enhacement has already been submitted here
[Corrected Link!  https://github.com/supabase/supabase/pull/44515](https://github.com/supabase/supabase/pull/44515)
the new change is like below
feature_tutorial

PR #43282, opened 2026-03-02
fix(Docs): Make code tags sync generated

jeremenichelli:

The main problem is back ticks that generate an inline code tag in headings. It seems the MDX parser we use, or somewhere in our configuration, forces that code tag to be processed via our CodeBlock component which has a top level await and generates a React.lazy component. When that heading with a lazy component as children gets evaluated, as it's not rendered yet it hasn't resolved its children yet and the anchor generation for the heading fails.

Heading with back ticks inside fail to render.

## What is the new behavior?

Heading with back ticks inside successfully render.

## Additional context

You can confirm that manually by replacing the h3 assignation in the `MDXBase.shared.tsx` file with this which forces the async resolution of all children:

```
h3: async (props: any) =>  {
  await Promise.all([...props.children])
   
  return <Heading tag="h3" {...props}>
      {props.children}
    </Heading>
}
```

I am gonna opt-in for another solution for now though and leave a comment for posterity.


ChrisChinchilla:
@jeremenichelli Nice catch!

aantti (approved):

PR #43292, opened 2026-03-02
feat(etl): Add new flags per org

iambriccardo:
This PR adds new feature flags for controlling etl destinations visibility based on the org slug.


joshenlim:
Just to reiterate changes i made here:
- Reduce the usage of 6 feature flags down to 2
  -  `etlEnableBigQueryPrivateAlpha` and `etlEnableIcebergPrivateAlpha`
- Access to ETL UI (`useIsETLPrivateAlpha`) will just be a derivative of those 2 flags via `OR`
- Override to allow access to all users is just by setting those flags to a value of `all`

joshenlim (approved):
Verified changes on both staging preview and local

PR #43300, opened 2026-03-02
feat: add tooltip displaying active incidents to banner

charislam:

YES

Feature

The incident banner exists but does not display active incidents.

## What is the new behavior?

The incident banner now displays a tooltip showing active incidents.

## Additional context

Test on local, with the latest env vars from platform it should have an active incident banner

CleanShot 2026-03-02 at 13 34 15@2x

Resolves FE-2651
(the onDismiss issue from that incident already works, just addressing the listing incidents part)


joshenlim:
Hmm so sorry just thinking this twice now - i know i initially suggested showing the incident title in the banner, but realised that perhaps we shouldn't distract users from the primary CTA in this banner which is directing users to the status page + keep the banner light weight

Will close this one for now! 🙏

PR #43301, opened 2026-03-02
chore: disk size config entitlement

ignaciodob:
Adds entitlements check to the `DiskSizeConfiguration` and `DiskSizeConfigurationModal` components

### Database Settings
- You'll need an Org in the Pro Plan with spend cap disabled, and a Project with Fly as the cloud provider.
- Head to `/project/_/database/settings`
- Assert that the Disk Size configuration section is visible:
image

- With a Free Project with Fly as the cloud provider, assert that the Disk Size configuration is gated:
image

### Observability > Database

- You'll need an Org in the Pro Plan with spend cap disabled, and a Project with Fly as the cloud provider.
- Head to `project/_/observability/database`, and scroll to the `Database Size` section
- Assert that clicking the `Increase disk size` button shows this modal:
image

- With a Free Project with Fly as the cloud provider, assert that the increase size modal doesn't allow you to increase the size
image


awaseem (approved):
LGTM

PR #43308, opened 2026-03-02
feat(studio): Display region-specific incidents in RegionSelector

charislam:

YES

Feature

Region-specific incidents are only shown in the global StatusPageBanner, which doesn't clearly indicate that an incident only affects specific regions during project creation.

## What is the new behavior?

Region-specific incidents are now displayed inline in the RegionSelector with smart region matching to show which regions are affected. The StatusPageBanner logic is updated to avoid duplicate incident notices for region-specific incidents when creating projects.

## Additional context

CleanShot 2026-03-02 at 16 32 34@2x

Resolves FE-2652


joshenlim (approved):
Verified changes locally

Also checked staging preview, although unable to test properly but minimally verified that we're not showing the warning incorrectly

PR #43309, opened 2026-03-02
chore: organization mfa entitlement

ignaciodob:
Adds the entitlement check for organization MFA.

### Testing 
- With an Org on the Free Plan, head to `/org/_/security`
- Assert that the upgrade prompt is shown:

image

- With an Org on the Pro Plan or above, head to `/org/_/security`
- Assert that you can enable MFA:
image


awaseem (approved):
LGTM

PR #43314, opened 2026-03-02
chore: auth sessions entitlement

ignaciodob:
This PR adds the entitlement check for configuring auth user sessions.

### Testing
- Head to `/project/_/auth/sessions` with an org on the Free Plan 
- Assert that the upgrade prompt is shown and that u:
image

- Head to `/project/_/auth/sessions` with an org on the Pro Plan or above
- Assert that you're able to configure the User Sessions section
image


awaseem (approved):
LGTM

PR #43322, opened 2026-03-03
chore: auth performance settings entitlement

ignaciodob:
Adds an entitlement check to the Auth Performance Settings page.

### Testing
- Head to `/project/_/auth/performance` with an org on the Free Plan
- Assert that you can't edit any values in the page
- Assert that the upgrade prompt is shown
image

- Head to `/project/_/auth/performance` with an org on the Pro Plan or above
- Assert that the upgrade prompt is not rendered
- Assert that you're able to modify and save values


awaseem (approved):
LGTM, there is something wrong with unit tests so worth looking into why

PR #43323, opened 2026-03-03
Persist dismissal of table filter bar banner if already enabled or enabling feature preview

joshenlim:
## Context

Persists the dismissal of the table filter banner promo if
- Already enabled
- After enabling feature preview

image


alaister (approved):
LGTM!

- Tested on preview

PR #43325, opened 2026-03-03
Joshen/fe 2453 post rollout dashboard clean up

joshenlim:
## Context

Just cleaning up the `unifiedReplication` feature flag now that this is fully rolled out

Also addresses some bugs that I came across while cleaning up, specifically for the new home page
- Add enabled features flag check for read replicas in database replication
- Infra diagram, read replicas
  - View connection string CTA to open connect UI with the right database selected
  - Replaced all other actions with a "Manage replica" CTA which links to the replica details page


alaister (approved):
LGTM!

- Tested on preview

PR #43332, opened 2026-03-03
fix: banning / unbanning a user does not update user pane ban status

jordienr:


joshenlim (approved):

PR #43333, opened 2026-03-03
feat: nested mobile nav menu

fsansalvadore:
- consolidate `top level menu` and `contextual menu` into nested menu on mobile
- removed previous mobile submenu

### Before
Screenshot 2026-03-03 at 11 29 48

### After
https://github.com/user-attachments/assets/de5e9fc8-466a-4417-9205-243d2239f437

For testing:
- on mobile
  - on `/organizations`, you shouldn's see any mobile menu button
  - on `/org/[orgId]`, you should see mobile menu button
  - org mobile menu should navigate directly to all menu items except "Oranization Settings" which should open relevant sub menu with nav items
  - in project context, the mobile menu button should open the Project main menu in the homepage
  - All sections with sub menus should open the sub menu in the mobile sheet except for "sql editor" and "table editor" which should navigate directly (because otherwise you wouldn't be able to navigate from the submenu)
  - when inside a product section, the mobile menu should open up directly on the product sub menu
  - you should be able to open to the main menu from the top of the mobile sheet
  - account settings menu should be available on mobile
- on desktop the navigation should be the same as before


apps/studio/components/layouts/Integrations/layout.tsx
```diff
@@ -1,18 +1,98 @@
...
 import { withAuth } from 'hooks/misc/withAuth'
 import { Menu, Separator } from 'ui'
 import { GenericSkeletonLoader } from 'ui-patterns'
 
+/**
+ * Menu-only component for the Integrations section. Used by the desktop sidebar and by the
+ * mobile sheet submenu. Must not wrap ProjectLayout so that opening the integrations submenu
+ * in the mobile sheet does not overwrite registerOpenMenu and break the menu button.
+ */
+export const IntegrationsProductMenu = () => {
```
awaseem: I don't think its right to create a component in this file, rather this should be its own file. Its cleaner and easier to read

apps/studio/components/layouts/Integrations/layout.tsx
```diff
@@ -1,18 +1,98 @@
...
+  const router = useRouter()
+  const { ref: projectRef } = useParams()
+  const { data: project } = useSelectedProjectQuery()
+  const { integrationsWrappers: showWrappers } = useIsFeatureEnabled(['integrations:wrappers'])
+
+  const segments = router.asPath.split('/')
+  const page = `${segments[3]}${segments[4] ? `/${segments[4]}` : ''}`
```
awaseem: This scares me a bit because what if the path for whatever reason is less than the length of 5. I think we need to be a bit defensive here and check if these values exist. Better to split this into a util method and test it

apps/studio/components/layouts/Integrations/layout.tsx
```diff
@@ -1,18 +1,98 @@
...
+
+  const urlParams = new URLSearchParams(router.asPath.split('?')[1] || '')
+  const categoryParam = urlParams.get('category')
```
awaseem: Same goes for the comment above. All this array logic should be type checked correctly and split into its own method rather than be in the component

apps/studio/components/layouts/Integrations/layout.tsx
```diff
@@ -1,18 +1,98 @@
...
+    label: integration.status,
+    key: `integrations/${integration.id}`,
+    url: `/project/${project?.ref}/integrations/${integration.id}/overview`,
+    icon: (
+      <div className="relative w-6 h-6 bg-white border rounded flex items-center justify-center">
+        {integration.icon({ className: 'p-1' })}
+      </div>
+    ),
+    items: [],
+  }))
```
awaseem: Very minor nit, wonder if it would be better to just map to the component `ProductMenuItem`rather than an object and than another map below in the render function. Get a list of components instead and display them

apps/studio/components/layouts/ProjectLayout/LayoutHeader/MobileMenuContent/MobileMenuContent.tsx
```diff
@@ -0,0 +1,294 @@
...
+  const sectionLabel =
+    sectionKeyToShow &&
+    (sectionKeyToShow === currentSectionKey
+      ? currentProduct
+      : allTopLevelRoutes.find((r) => r.key === sectionKeyToShow)?.label ?? sectionKeyToShow)
+
+  const SectionMenuContent = sectionKeyToShow ? getProductMenuComponent(sectionKeyToShow) : null
+  const pageSegment = pathname.split('/')[4]
+
+  const renderTopLevelRoute = useCallback(
```
awaseem: This doesn't need to be a function, this is its own component. The arguments of this function are the props and its returning JSX. This should be refactored to its won component

apps/studio/components/layouts/ProjectLayout/LayoutHeader/MobileMenuContent/MobileMenuContent.tsx
```diff
@@ -0,0 +1,294 @@
...
+    setViewLevel('top')
+    setSelectedSectionKey(null)
+  }, [])
+
+  const sectionKeyToShow = viewLevel === 'section' ? selectedSectionKey ?? currentSectionKey : null
+  const sectionLabel =
+    sectionKeyToShow &&
+    (sectionKeyToShow === currentSectionKey
+      ? currentProduct
+      : allTopLevelRoutes.find((r) => r.key === sectionKeyToShow)?.label ?? sectionKeyToShow)
```
awaseem: Logic here is confusing, to many ternary. Should refactor to a util method and test

apps/studio/components/layouts/ProjectLayout/LayoutHeader/MobileMenuContent/MobileMenuContent.tsx
```diff
@@ -0,0 +1,294 @@
...
+        onCloseSheet?.()
+      }
+    },
+    [hasSubmenu, router, onCloseSheet]
+  )
+
+  const handleBackToTop = useCallback(() => {
+    setViewLevel('top')
+    setSelectedSectionKey(null)
+  }, [])
```
awaseem: All the methods above with callbacks is super hard to parse. Instead we should refactor these out into a single hook or multiple hooks. Then reference the hook here with all the methods extracted. Otherwise this becomes are to read

apps/studio/components/layouts/ProjectLayout/LayoutHeader/MobileMenuContent/MobileMenuContent.utils.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+export function getSectionKeyFromPathname(pathname: string): string | null {
+  const segments = pathname.split('/').filter(Boolean)
+  const projectIndex = segments.indexOf('project')
+  if (projectIndex === -1 || segments.length <= projectIndex + 1) return null
+  const refSegment = segments[projectIndex + 1]
+  if (!refSegment || refSegment.startsWith('[')) return null
+  const sectionSegment = segments[projectIndex + 2]
+  if (!sectionSegment) return null
+  return sectionSegment
```
awaseem, line 17: nit this method needs some new lines. Really hard to read

apps/studio/components/layouts/ProjectLayout/LayoutHeader/MobileMenuContent/OrgMenuContent.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const router = useRouter()
+  const { slug } = useParams()
+  const organizationSlug: string = slug ?? (router.query.orgSlug as string) ?? ''
+  const { data: org } = useSelectedOrganizationQuery()
+  const isUserMFAEnabled = useIsMFAEnabled()
+  const disableAccessMfa = org?.organization_requires_mfa && !isUserMFAEnabled
+  const showBilling = useIsFeatureEnabled('billing:all')
+
+  const pathname = router.asPath?.split('?')[0] ?? router.pathname
+  const activeRoute = pathname.split('/')[3]
```
awaseem: Again this can lead to some issues, we should be very defensive here when checking array indexed items. We should rip out, create conditions and test

apps/studio/components/layouts/ProjectLayout/LayoutHeader/MobileMenuContent/OrgMenuContent.tsx
```diff
@@ -0,0 +1,125 @@
...
+                  <>
+                    <span className="flex size-5 shrink-0 items-center justify-center [&>svg]:size-5 [&>svg]:shrink-0">
+                      {item.icon}
+                    </span>
+                    <span className="truncate">{item.label}</span>
+                  </>
+                )
+                const menuButtonClass = cn(
+                  sidebarMenuButtonVariants({ size: 'default', hasIcon: true }),
+                  disableAccessMfa && 'opacity-50 pointer-events-none'
```
awaseem: The inside of this map function can be its own component, then when you map over the items you can just pass the whole object. IsActive checks should be a pure function and tested

apps/studio/components/layouts/ProjectLayout/NavigationBar/StudioMobileSheetNav.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { useMobileSheet } from './MobileSheetContext'
+
+function isSidebarId(content: unknown): content is TYPEOF_SIDEBAR_KEYS {
+  return (
+    typeof content === 'string' &&
+    Object.values(SIDEBAR_KEYS).includes(content as TYPEOF_SIDEBAR_KEYS)
+  )
+}
+
+/** Resolve sheet content: sidebar panel (by id) or custom ReactNode (menu, etc.). */
```
awaseem: Don't need this comment

apps/studio/components/layouts/ProjectLayout/index.tsx
```diff
@@ -96,7 +104,11 @@ export const ProjectLayout = forwardRef<HTMLDivElement, PropsWithChildren<Projec
     const router = useRouter()
     const { data: selectedOrganization } = useSelectedOrganizationQuery()
     const { data: selectedProject } = useSelectedProjectQuery()
-    const { mobileMenuOpen, showSidebar, setMobileMenuOpen } = useAppStateSnapshot()
+    const { showSidebar } = useAppStateSnapshot()
+    const { setContent: setMobileSheetContent, registerOpenMenu } = useMobileSheet()
+
+    const pathname = router.asPath?.split('?')[0] ?? router.pathname
```
awaseem: Again this logic is used in a bunch of other place, unify, split out and refactor please

awaseem (changes requested):
Needs work on refactor, array checks like [0] or [x] need to be handled for undefined values

apps/studio/components/layouts/Integrations/Integrations.utils.ts
```diff
@@ -0,0 +1,36 @@
+/**
```
awaseem: nit: we can reduce some of these comments here

apps/studio/components/layouts/Integrations/IntegrationsMenu.utils.ts
```diff
@@ -0,0 +1,44 @@
...
+        {
+          name: 'Postgres Modules',
+          key: 'integrations-postgres_extension',
+          url: `/project/${projectRef}/integrations?category=postgres_extension`,
+          pages: ['integrations?category=postgres_extension'],
+          items: [],
+        },
+      ],
+    },
+  ].filter(Boolean)
```
awaseem: Why do we need this filter? I don't think this is needed right?

apps/studio/components/layouts/OrganizationLayout/OrganizationLayout.utils.ts
```diff
@@ -0,0 +1,28 @@
+/**
```
awaseem: same comment here, just a minor nit for clean up of comments

apps/studio/components/layouts/OrganizationLayout/OrganizationLayout.utils.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Organization layout path and scope utilities.
+ * Defensive helpers for pathname parsing and route scope checks.
+ */
+
+/**
+ * Extracts pathname without query string. Safely handles undefined inputs.
+ */
+export function getPathnameWithoutQuery(
+  asPath: string | undefined,
+  fallback: string | undefined
```
awaseem: can do the following:

```suggestion
  asPath?: string
  fallback?: string
```

apps/studio/components/layouts/ProjectLayout/LayoutHeader/MobileMenuContent/OrgMenuContent.utils.ts
```diff
@@ -0,0 +1,58 @@
+import type { ReactNode } from 'react'
```
awaseem, line 1: nit comments are no longer needed

awaseem (approved):
Looks good, minor comments. Would double check if no errors are showing up and aria labels are set

staaldraad (approved):

kemaldotearth (approved):

myerekapan (approved):

apps/studio/components/layouts/Integrations/IntegrationsMenu.utils.ts
fsansalvadore: indeed we don't, removed 👍

PR #43336, opened 2026-03-03
Make e2e tests faster

djhi:
## Problem

The e2e tests suite could be faster

## Solution

- [x] Disable animations while running the tests with `prefers-reduced-motion`
- [x] Use database queries for tests setup when possible
- [x] Introduce a test cleanup mechanism


djhi:
Closed in favor of smaller PR

PR #43345, opened 2026-03-03
docs: Update auth-helpers usage from examples

ChrisChinchilla:

YES

### DIFF
--- a/examples/realtime/nextjs-auth-presence/lib/supabase-context.tsx
+++ b/examples/realtime/nextjs-auth-presence/lib/supabase-context.tsx
@@ -1,0 +1,34 @@
+import { createBrowserClient } from '@supabase/ssr'
+import { User } from '@supabase/supabase-js'
+import { createContext, useContext, useEffect, useState } from 'react'
+
+const supabase = createBrowserClient(
+  process.env.NEXT_PUBLIC_SUPABASE_URL!,
+  process.env.NEXT_PUBLIC_SUPABASE_PUBLISHABLE_KEY!
+)
+
+type SupabaseContextType = {
+  supabase: typeof supabase
+  user: User | null
+}
+
+const SupabaseContext = createContext<SupabaseContextType>({ supabase, user: null })
+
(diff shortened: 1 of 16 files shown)


mandarini (changes requested):
I think there are two bugs in the `setAll` cookie handler used in `pages/index.tsx` and `pages/profile.tsx` (same pattern in both files).

**Bug 1: Cookie options are silently dropped**

Current implementation ignores the options argument passed to each cookie item:

```
  setAll(cookiesToSet) {
    ctx.res.setHeader(
      'Set-Cookie',
      cookiesToSet.map(({ name, value }) => `${name}=${value}`) 
    )
  }
```

`@supabase/ssr` passes full `CookieOptions` (httpOnly, secure, path, sameSite, maxAge, etc.) as the third field of each item. Discarding them means any cookies set during a token refresh will be missing security attributes, which can lead to session exposure or cookies that expire at the wrong time.

**Bug 2: setHeader overwrites existing Set-Cookie headers**

`res.setHeader('Set-Cookie', ...)` replaces any previously set `Set-Cookie` header on the response, rather than appending to it. If anything else on the page is also setting a cookie, it will be silently lost.

To fix both we could use the `serialize` function from the cookie package (already a dependency of `@supabase/ssr`) and `res.appendHeader`:

```
import { serialize } from 'cookie'

  setAll(cookiesToSet) {
    const existing = ctx.res.getHeader('Set-Cookie') ?? []
    ctx.res.setHeader('Set-Cookie', [
      ...(Array.isArray(existing) ? existing : [String(existing)]),
      ...cookiesToSet.map(({ name, value, options }) => serialize(name, value, options)),
    ])
  }
```

This would apply to both `pages/index.tsx` and `pages/profile.tsx` . The rest of the migration all look correct! Still would like to hear @supabase/auth 's opinion!

ChrisChinchilla:
Ready for a new look @mandarini and @supabase/auth

mandarini (approved):

PR #43348, opened 2026-03-03
fix: time editor erases value as you type

djhi:
## Problem

When using chrome, it's impossible to enter a time value

## Solution

Avoid controlling the input value and only bubble up the change when the value is a valid time or is empty

## How to test

- Create a table with a `time` column
- Add a new row and start entering a value using the keyboard


awaseem (approved):
Tested on staging!

PR #43352, opened 2026-03-03
feat(studio): wrap table creation in single atomic transaction

7ttp:
## Problem

Tables are partially created when errors occur during table creation.

## Solution

Introduces `noTransaction` flag to pg-meta's `tables.create()` and `columns.create()` methods (defaults to `false` for BC). 
When set to `true`, these methods return plain SQL without transaction wrappers, 
allowing the studio to wrap all operations in a single atomic transaction.

If any operation fails, everything rolls back cleanly now.

## Changes

- Pass `noTransaction: true` to `pgMeta.tables.create()`
- Pass `noTransaction: true` to `pgMeta.columns.create()`
- Wrap all SQL statements in single `BEGIN...COMMIT` block

## Related

- Depends on https://github.com/supabase/postgres-meta/pull/1049
- Closes #42089
-  follows Gildas's Suggestion


djhi:
Closed in favor of #43909. Thanks!

PR #43286, opened 2026-03-02
add oauth and otp how-to guides for self-hosted

aantti:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

- Add two new how-to guides covering OAuth and OTP/MFA configuration for [self-hosted Supabase](https://supabase.com/docs/guides/self-hosting)
- Add placeholder environment variables to `.env.example` and `docker-compose.yml`

PR #43357, opened 2026-03-03
fix: reformat yaml and clarify testing in self-hosted oauth how-to

aantti:

YES

A follow-up to PR #43286:

- Reformat yaml blocks for clarity
- Add missing `apikey` header to test `curl`
- Add additional notes about testing OAuth via a local HTML page


apps/docs/content/guides/self-hosting/self-hosted-oauth.mdx
```diff
@@ -94,7 +97,7 @@ docker compose up -d --force-recreate --no-deps auth
 Check that the provider is enabled:
 
 ```sh
-curl https://<your-domain>/auth/v1/settings
+curl -H 'apikey: your-anon-key' https://<your-domain>/auth/v1/settings
```
ChrisChinchilla, line 100: @aantti Unless there's a specific reason to use anon keys, I think we're trying to steer people towards publishable keys. While we're in transition, I have been generalising to something along the lines of just generic key instead.

ChrisChinchilla (commented):
Small comment, maybe it's somehting specific to self-hosted though

ChrisChinchilla (approved):

PR #43368, opened 2026-03-04
feat(studio): add stuck pausing and restoring escalations

dnywh:

Bug fix / UX improvement for long-running project transitions. Resolves DEPR-362.

- `PausingState` does not preserve elapsed time across refreshes, so the stuck escalation can disappear for the same user.
- `RestoringState` relies on weaker frontend heuristics and always showed a support CTA in the footer even before the restore was clearly long-running.

## What is the new behaviour?

- `PausingState`
  - Persists a per-project pause start time in local storage so the stuck CTA survives refreshes in the same browser.
  - Escalates after 10 minutes.
  - Clears the stored timer when pausing succeeds or fails.

- `RestoringState`
  - Persists a per-project restore start time in local storage so the stuck CTA survives refreshes in the same browser.
  - Removes the always-visible footer CTA and only escalates once restoration is genuinely long-running.
  - Computes the long-running threshold from volume size using a shared restore estimate: `max(10, ceil(estimateRestoreTime(sizeGb) * 1.5))`.
  - Clears the stored timer when restoration succeeds or fails.

- Shared changes
  - Extracts reusable transition timing helpers and restore estimate helpers with unit tests.
  - Reuses the same restore estimate formula for branch restore timing and restore escalation, so the two do not drift.

| `PausingState` | `RestoringState` |
| --- | --- |
| Krosno  Toolshed  Supabase-C6D7E29F-C38D-43E1-8AF9-C612B6A2FD8D | Krosno  Toolshed  Supabase-51F4763D-B798-4B41-A92D-43B3CF8ECDAF |

## Additional context

- This PR intentionally stays frontend-only.
- We are not exposing backend lifecycle timestamps here; local storage is the stopgap to improve the same-browser experience now.
- If you need to test the frontend blocker states locally, use [`dnywh/chore/depr-362-blocker-preview-mocks`](https://github.com/supabase/supabase/tree/dnywh/chore/depr-362-blocker-preview-mocks) and append one of the following query params to a project URL:
  - `?mockProjectBlockingState=pausing`
  - `?mockProjectBlockingState=pausing-long-running`
  - `?mockProjectBlockingState=restoring`
  - `?mockProjectBlockingState=restoring-long-running`
- I know these two views are quite differently stylistically, and will consolidate later
  - References DEPR-434


awaseem (approved):
LGTM looks good to me!

apps/studio/components/layouts/ProjectLayout/RestoringState.tsx
```diff
@@ -42,13 +48,36 @@ export const RestoringState = () => {
...
+
+    if (remainingThresholdMs === 0) {
+      setIsTakingLongerThanExpected(true)
+      return
+    }
+
+    setIsTakingLongerThanExpected(false)
+    const timeoutId = setTimeout(() => setIsTakingLongerThanExpected(true), remainingThresholdMs)
+    return () => clearTimeout(timeoutId)
+  }, [longRunningThresholdMs, maxPersistedTransitionAgeMs, restoreStateStartStorageKey])
```
awaseem: This looks identical to `PausingState`, maybe we can reuse the effects in a hook that both components can resuse

awaseem (approved):
Minor changes

apps/studio/components/layouts/ProjectLayout/RestoringState.tsx
dnywh: Thanks. Fixed in latest commit

PR #43370, opened 2026-03-04
fix: storage view is not reset when switching buckets

djhi:
## Problem

When navigating between buckets after having selected a folder, the storage view shows the previous folder content.

## Solution

Set the selected bucket before navigating to it, preventing a race condition with Valtio updates and the effect that loads the bucket content.

## How to test

1. Create 2 buckets
2. In one of them, create a folder and upload some files
3. Go back to the bucket list
4. Open the other bucket

Results:
- It should not show the previous bucket content anymore
- It should not show an empty 2nd column either


joshenlim (approved):
Verified changes on preview! just pushed up some very minor nits

apps/studio/components/interfaces/Storage/FilesBuckets/BucketTable.tsx
```diff
@@ -83,6 +85,7 @@ export const BucketTableRow = ({
 }: BucketTableRowProps) => {
   const router = useRouter()
   const { getPolicyCount } = useBucketPolicyCount()
+  const snap = useStorageExplorerStateSnapshot()
```
ivasilov: You don't need this change, right?

djhi: Yeah I actually don't need to change that file at all

PR #43383, opened 2026-03-04
fix(query-performance): optimize pg_stat_statements query DEBUG-28

jordienr:
- Filter out never-run queries early: previously the queries didn't check if a statement had ever actually been executed, so PostgreSQL would process and return those useless rows. Adding WHERE calls > 0 skips
  them upfront.
- Replace window functions with a pre-computed total: two queries were using a SQL window function to calculate the "% of total time" column, which forces PostgreSQL to load every single row into memory before
  returning results. Replaced with a small CTE that computes the total once separately, so PostgreSQL can apply the row limit without buffering everything first. This was the main cause of page crashes on databases
   with a large query history.
 - slowQueriesCount bug fix: added missing as statements alias
 - queryMetrics WHERE clause fix: ${where} was producing WHERE ... WHERE ..., now uses .replace(/^WHERE/, 'AND')

### DIFF
--- a/apps/studio/components/interfaces/Reports/Reports.queryPerformance.test.ts
+++ b/apps/studio/components/interfaces/Reports/Reports.queryPerformance.test.ts
@@ -1,0 +1,110 @@
+import { describe, it, expect } from 'vitest'
+import { PRESET_CONFIG } from './Reports.constants'
+import { Presets } from './Reports.types'
+
+const queries = PRESET_CONFIG[Presets.QUERY_PERFORMANCE].queries as Record<
+  string,
+  { sql: (...args: any[]) => string }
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Reports/Reports.queryPerformance.test.ts
```diff
@@ -0,0 +1,116 @@
+import { describe, expect, it } from 'vitest'
```
awaseem, line 1: These are kinda interesting, just tests on the query builder. Unsure how I feel about them, but better than nothing

awaseem (approved):
LGTM

PR #43387, opened 2026-03-04
Adjust header + project list for wildcard routes

joshenlim:
## Context

Small adjustments to wildcard routes

- Header logo to be consistent with other layouts
  - Before:
    image
  - After:
    image
- Project wildcard route list view to be consistent with org/[slug] page - project list shouldn't take full height if there's no need to, scroll behaviour should be on the page
  - Before:
    image
  - After:
    image


awaseem (approved):
Tested on staging ✅

PR #43408, opened 2026-03-04
fix: updated filter bar logic to support selection of columns once set

awaseem:

YES

- Allow selecting the column, it respects different types and updates the values correctly. If the values are different it gets wiped due to mismatches
- Updated E2E tests and unit tests


charislam (approved):

kemaldotearth (approved):
Works as expected on preview.

PR #43410, opened 2026-03-04
chore: prevention of used leaked passwords entitlement

ignaciodob:
### Changes
- Replaces the isPaid plan-based check on the "Prevent use of leaked passwords" (PASSWORD_HIBP_ENABLED) setting with a proper entitlement check using the `password_hibp` entitlement key
- Adds a new `useHasEntitlementAccess` hook that returns a reusable checker function for any entitlement key, backed by the same cached entitlements query

### Testing
- Head to `/project/_/auth/providers?provider=Email` with an Org on the Free Plan
- Assert that the "Prevent use of leaked passwords" toggle is disabled.
- Head to `/project/_/auth/providers?provider=Email` with an Org on the Pro Plan
- Assert that the "Prevent use of leaked passwords" toggle is enabled and can be toggled and saved.

image


awaseem (approved):
LGTM

PR #43417, opened 2026-03-05
chore(design-system): preserve icon fill or stroke

dnywh:

Bug fix: custom icons now respect their source SVG’s stroke/fill and stroke-width instead of always getting a global stroke.

Every custom icon gets `stroke="currentColor"` and `strokeWidth={2}` from `createSupabaseIcon`, so fill-only logos get an extra stroke and icons designed at `stroke-width="1"` (e.g. postgres, auth) render too thick. Call sites and the design-system grid needed workarounds (`strokeWidth={0}`, `FILL_ONLY_ICONS`, `STROKE_WIDTH_FROM_SOURCE`).

## What is the new behavior?

The icon build reads root SVG attributes (`fill`, `stroke`, `stroke-width`, etc.) and passes them as per-icon defaults. Fill-only icons (chatgpt, claude, axiom, last9) have `stroke="none"` in source and render with no stroke; stroke icons keep their source stroke-width (e.g. postgres at 1). No `strokeWidth={0}` or allowlists needed at call sites or in the icon grid.

| Before | After |
| --- | --- |
| CleanShot 2026-03-05 at 10 53 31@2x-3D63CEF0-1132-44F5-A382-730346432F1E | CleanShot 2026-03-05 at 10 53 40@2x-514FC8BD-D30F-4D23-B209-0ECD6D13A184 |

## Additional context

- Added `stroke="none"` to the four fill-only source SVGs; sentry/grafana/otlp/datadog already had `stroke-width="0"`.  
- `createSupabaseIcon` only applies `color`/`strokeWidth` when the consumer passes them, so `svgDefaults` from the build are used otherwise.  
- Removed workarounds from `icons.tsx` and GuidesSidebar; updated icons.mdx (defaults + fill-only guidance).


kemaldotearth (approved):
Didn't spot any broken icons, think we're good! :shipit:

PR #43437, opened 2026-03-05
chore(studio): graduate homeNew experiment

pamelachia:
## Summary

The `homeNew` PostHog experiment has concluded. This PR graduates it by making the new homepage (`ProjectHome`, formerly `HomeV2`) the permanent default for all users, and removes all dead code from the old experiment.

## Changes

- Remove `homeNew` PostHog feature flag checks and `home_new` experiment exposure tracking from 3 files
- Rename `HomeNew/` → `ProjectHome/` directory and `HomeV2` → `ProjectHome` export
- Delete old `Home/Home.tsx` component (shared components like `ProjectList/` are kept — still used by org pages)
- Delete `pages/project/[ref]/building.tsx` and add a server-side redirect from `/project/:ref/building` → `/project/:ref` to prevent 404s during rollout (old cached JS bundles may still route to `/building`)
- Simplify `ContentWrapper` building-state logic in `ProjectLayout` — always redirect building projects to home, always suppress building interstitial on home page
- Always route to `/project/{ref}` after project creation (remove `/building` path)
- Update all Observability imports from `HomeNew` → `ProjectHome`

## Self-hosted behavior change

Self-hosted Studio previously showed the old `Home` component (client libraries + example projects) since PostHog flags don't load. This PR changes self-hosted to show `ProjectHome` (TopSection with service status + instance diagram, advisor, custom reports). All sections query backend APIs that exist on self-hosted. E2E tests pass against the self-hosted build.

## Testing

- [x] `pnpm turbo run build --filter=studio` passes
- [x] No remaining references to `homeNew`, `home_new`, or `HomeNew` in codebase
- [x] No broken imports to deleted files
- [x] Self-hosted E2E tests pass (145 passed, 1 flaky, 4 skipped)
- [x] `/building` redirect added to both platform and self-hosted config blocks

**Quick test:**
1. Navigate to any project homepage — should render the ProjectHome component
2. Create a new project — should redirect to `/project/{ref}` (not `/building`)
3. Visit a project in `COMING_UP` state on a non-home route — should redirect to home
4. Visit `/project/{ref}/building` directly — should 302 redirect to `/project/{ref}`

## Linear

- fixes GROWTH-671


apps/studio/next.config.js
```diff
@@ -120,6 +125,11 @@ const nextConfig = {
               destination: '/project/default',
               permanent: false,
             },
+            {
```
ivasilov, line 128: You have this entry twice, probably not intentional.

awaseem:
@pamelachia would you mind resolving conflicts and then open again for review? @ivasilov also had minor feedback

apps/studio/next.config.js
pamelachia: This is intentional — the ternary has two branches: platform config (IS_PLATFORM=true) and self-hosted config (else). Each branch needs its own /building redirect since both environments may have cached JS bundles routing to /building. The PR description notes 'added to both platform and self-hosted config blocks'.

ivasilov (approved):

PR #43443, opened 2026-03-05
chore: logs entitlements badges

ignaciodob:
### Changes
- Replace hardcoded plan-to-days mapping (`getAvailableInForDays`) with entitlement-driven checks.
- Show a lock icon on date picker helpers that exceed the user's entitled log retention, instead of a plan name badge.
- Remove availableIn from DatetimeHelper type — the entitlement API is now the single source of truth.

### Testing
- Head to `/project/_/logs` on a free plan project and verify lock icons appear on helpers beyond 1 day
- Verify clicking a locked helper still triggers the upgrade prompt
image


awaseem (approved):
Nice clean up

PR #43450, opened 2026-03-05
fix: cannot resize table editor columns on safari

djhi:
## Problem

On Safari, it's impossible to resize a column.

## Solution 

This is actually a bug in `react-data-grid` that has been fixed in more recent versions but we can't use them.
Patch the package

## How to test

- Open the any table in the _Table editor_
- Resize a column


alaister (approved):
LGTM!

- Tested on preview on Safari

PR #43463, opened 2026-03-05
chore(Docs): Run format on docs folder to update files

jeremenichelli:
Running format on our docs folder to prevent accumulating diffs when running it local over other branches.


ChrisChinchilla (approved):
Ah yes, I created https://github.com/supabase/supabase/pull/43145 last week.

PR #43469, opened 2026-03-05
chore: add entitlement check to observability log retention

ignaciodob:
### Changes
- Replace plan-based checks (`LOG_RETENTION`, `availableIn`) with `useCheckEntitlements('log.retention_days')` in the chart interval dropdown
- `ChartIntervalDropdown` now calls the entitlements hook internally instead of receiving plan props
- Remove `LOG_RETENTION` constant and `availableIn` from `CHART_INTERVALS`
- Update consumer components (`ProjectUsage`, `ProjectUsageSection`, `ObservabilityOverview`) to remove plan prop passing

### Testing
- Head to `/project/_/observability` with a Free plan org.
- Click on the time selector "Last 7 days" is disabled, default is "Last 60 minutes", tooltip shows retention limit and upgrade link

image

- With a Pro Org and above the "Last 7 days" option should be enabled


djhi (approved):
Tested on staging 👍

PR #43470, opened 2026-03-05
feat(sentry): enable Session Replay in Studio

seanoliver:
## Problem

We have no session recording to help debug user-reported issues in Studio. We're running a time-limited Sentry enterprise trial and this adds Session Replay as part of Phase 1 instrumentation.

## Changes

- **`apps/studio/instrumentation-client.ts`** — Enabled Sentry Session Replay (10% of all sessions, 100% of error sessions) with `maskAllInputs: true` and `maskAllText: true`. Refactored the `integrations` setup from early-return to accumulator pattern to support multiple integrations cleanly; both `thirdPartyErrorFilterIntegration` and `replayIntegration` are guarded with existence checks so the config doesn't crash in test environments (jsdom has no browser APIs).
- **`packages/dev-tools/DevToolbar.test.tsx`** — Switched `process.env.NODE_ENV = '...'` to `vi.stubEnv` / `vi.unstubAllEnvs` — Next.js 16 types mark `NODE_ENV` as readonly so direct assignment is a type error.
- **`pnpm-lock.yaml`** — Resolves a pre-existing catalog version mismatch (unrelated to this feature).

## Testing

- `pnpm test:studio` — all tests pass including the instrumentation-client and DevToolbar suites

Linear: GROWTH-674


seanoliver:
Closing — reverting scope. Session Replay belongs in a dedicated PR with proper privacy review. Platform-side Sentry instrumentation continues in the platform repo PR.

PR #43476, opened 2026-03-06
chore(studio): refocus editor after Run button tap

dnywh:

Resolves FDBK-40065

Using the _Run_ button on either inline or “full” SQL Editor removes focus from the Monaco editor.

## What is the new behavior?

The text caret remains in the Monaco editor even after button press.

## To test

Try running a few queries both with keyboard (⌘ enter) and mouse click on the _Run_ button:

```sql
-- 1) Create a mock table + seed data
drop table if exists public.test;

create table public.test (
  id bigserial primary key,
  name text not null,
  created_at timestamptz not null default now()
);

insert into public.test (name)
values
  ('alpha'),
  ('beta'),
  ('gamma'),
  ('delta');

select * from public.test order by id;
```

```sql
-- 2) Deletion query that should trigger the warning modal (destructive op)
delete from public.test where id <= 2;

-- Verify remaining rows
select * from public.test order by id;
```

## Additional context

- [ ] This was FDBK-40065 that we should reply to


djhi (approved):

djhi:
tested on staging 👍

PR #43478, opened 2026-03-06
[FE-2158] (part 2) – feat(studio): Add exposed functions config to Postgrest settings

alaister:
**Changes:**
- Add function exposure controls to Data API settings
- Adds a per-function GRANT/REVOKE UI to the Data API settings page, alongside the existing table selector.
- Empty state: shows "No tables/functions available" instead of "0 of 0 X exposed" when there's nothing to show
- Renamed update-exposed-tables-mutation → update-exposed-entities-mutation to handle both tables and functions in one save
                                                                                                                                                                                                      
**Decisions of note:**
- Functions are grouped by schema.name across overloads — toggling one entry grants/revokes `EXECUTE` on all overloads
- Only normal and window functions are shown (`prokind in ('f', 'w')`) — triggers and aggregates are excluded

---

**To test:**

- Table and schema selectors still works end-to-end
- Toggling a function and saving updates the count label and applies the grant (check via SQL Editor: select * from information_schema.role_routine_grants where routine_schema = 'public')
- Functions in unexposed schemas are dimmed and unclickable
- With no functions in the selected schemas, the selector label reads "No functions available" and the dropdown shows the same


apps/studio/data/privileges/update-exposed-entities-mutation.ts
```diff
@@ -0,0 +1,134 @@
...
+import { getExposedSchemasSql } from './privileges.sql'
+
+export type UpdateExposedEntitiesVariables = ConnectionVars & {
+  tableIdsToAdd: number[]
+  tableIdsToRemove: number[]
+  functionNamesToAdd: string[]
+  functionNamesToRemove: string[]
+}
+
+const buildTablePrivilegesSql = (oids: number[], action: 'grant' | 'revoke') => {
```
joshenlim: similar feedback to the previous PR - these could be in `privileges.sql.ts`

apps/studio/components/interfaces/Settings/API/ExposedFunctionSelector.tsx
```diff
@@ -0,0 +1,310 @@
...
+  )
+
+  useEffect(() => {
+    if (!isPending && !isFetching && entry?.isIntersecting && hasNextPage && !isFetchingNextPage) {
+      fetchNextPage()
+    }
+  }, [entry?.isIntersecting, hasNextPage, isFetching, isFetchingNextPage, isPending, fetchNextPage])
+
+  return (
+    <div className={className}>
```
joshenlim: similar feedback to previous PR - this div and className seems unnecessary (unused)

apps/studio/components/interfaces/Settings/API/ExposedFunctionSelector.tsx
```diff
@@ -0,0 +1,310 @@
...
+      <Popover_Shadcn_ open={open} onOpenChange={setOpen} modal={false}>
```
joshenlim: also same feedback to previous PR - any reason why we aren't using the MultiSelector component?

apps/studio/components/interfaces/Settings/API/ExposedFunctionSelector.tsx
```diff
@@ -0,0 +1,310 @@
...
+                                          'shrink-0 flex items-center justify-center hover:text-foreground-light',
+                                          isCustomNeutral && isSchemaExposed
+                                            ? 'text-warning'
+                                            : 'text-foreground-muted'
+                                        )}
+                                      >
+                                        <CircleAlert size={14} />
+                                      </div>
+                                    </TooltipTrigger>
+                                    <TooltipContent
```
joshenlim: one thing to note for this one - i've got a function here that i created via the dashboard
which makes me wonder if all functions created via the dashboard will have this tooltip - which makes it look concerning at first glance (also because of the warning style)

am also wondering why functions created via the dashboard will be labelled as `custom` + why isn't it currently exposed if `public` schema is exposed? Will users need to manually exposed tables + functions after creating them via the dashboard ?
Image

apps/studio/components/interfaces/Settings/API/ExposedFunctionSelector.tsx
```diff
@@ -0,0 +1,310 @@
...
+                                        type="button"
+                                        tabIndex={-1}
+                                        aria-label="Schema not exposed"
+                                        className="inline-flex items-center text-foreground-muted hover:text-foreground-light"
+                                      >
+                                        <Info size={14} />
+                                      </button>
+                                    </TooltipTrigger>
+                                    <TooltipContent side="right" className="max-w-[320px] text-xs">
+                                      {`The schema "${fn.schema}" must be exposed before enabling this function.`}
```
joshenlim: very very nit but don't need to use `{}` for string interpolation here

apps/studio/components/interfaces/Settings/API/ExposedFunctionSelector.tsx
joshenlim: There's also a consistency thing here whereby the warning style goes away if i click on the option twice (toggle on and off)

https://github.com/user-attachments/assets/412aa309-129e-4e76-8fd1-72f98c4f5d53

joshenlim: the double tooltip here is also a bit trippy haha

i think we can prioritize the schema warning one, and if that one's not applicable then render the custom grant one instead - so keep to just one tooltip

Image

apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -223,6 +238,8 @@ export const PostgrestConfig = () => {
   const watchedDbSchema = form.watch('dbSchema')
   const watchedTableIdsToAdd = form.watch('tableIdsToAdd')
   const watchedTableIdsToRemove = form.watch('tableIdsToRemove')
+  const watchedFunctionNamesToAdd = form.watch('functionNamesToAdd')
```
ivasilov: You should use `useWatch_shadcn_` here, `form.watch` is not reactive (doesn't trigger a rerender).

apps/studio/data/privileges/update-exposed-entities-mutation.ts
```diff
@@ -0,0 +1,134 @@
...
+export const useUpdateExposedEntitiesMutation = ({
+  onSuccess,
+  onError,
+  ...options
+}: Omit<
+  UseCustomMutationOptions<UpdateExposedEntitiesData, Error, UpdateExposedEntitiesVariables>,
+  'mutationFn'
+> = {}) => {
+  return useMutation<UpdateExposedEntitiesData, Error, UpdateExposedEntitiesVariables>({
+    mutationFn: (vars: UpdateExposedEntitiesVariables) => updateExposedEntities(vars),
```
ivasilov, line 63: Shouldn't this invalidate postgrest config?

alaister: The invalidation is handled separately all in one go in PostgrestConfig -> onSubmit. If each handles it's own, the UX feels very jank

apps/studio/components/interfaces/Settings/API/ExposedFunctionSelector.tsx
alaister: How did you create your function via the dashboard and which version of postgres are you on? The functions I create from the dashboard are exposed by default.

> There's also a consistency thing here...

If you notice the tooltip actually changes after you select/deselect the table telling you what it's going to do, which is also why it doesn't go back to the default state. The user can hit cancel if they want to keep it in its current state. Do you think this is too confusing though?

> the double tooltip here is also a bit trippy haha

Fixed by shifting one of the tooltips to the left!

pgnickb (approved):
Mostly minor optimizations to SQL can be done, similar to the previous PR. Didn't find any real issues.

apps/studio/components/interfaces/Settings/API/ExposedTableSelector.tsx
```diff
@@ -92,197 +90,193 @@ export const ExposedTableSelector = ({
   }, [entry?.isIntersecting, hasNextPage, isFetching, isFetchingNextPage, isPending, fetchNextPage])
 
   return (
-    <div className={className}>
```
alaister, line 95: Just removing this className, no other changes here

apps/studio/components/interfaces/Settings/API/ExposedFunctionSelector.tsx
joshenlim: > How did you create your function via the dashboard and which version of postgres are you on? The functions I create from the dashboard are exposed by default.

Mentioned in Slack but replying here for visibility too - it happened when I used my database function in an auth hook and thereafter disabled the auth hook

> Do you think this is too confusing though?

hmm from a UX POV i think its a little confusing - this is kind of a "toggle" behaviour when i click on the select item, then click on it again - would expect the behaviour to revert to its original. The nuance between the states is also very subtle to catch (being hidden behind a tooltip)

> Fixed by shifting one of the tooltips to the left

it's still weird though seeing 2 icon circles 😅  I reckon we only show one at a time - with the schema tooltip taking precedence over custom grants
image

Separately as well ^ just wondering are we manually applying an opacity-50? the disabled state is quite hard to see (AAA contrast is quite low) - so the tooltips are easy to miss too

joshenlim (approved):
Approving to unblock - given that the changes here are feature flagged

There's some comments that I reckon will still need to be addressed nonetheless [here](https://github.com/supabase/supabase/pull/43478#discussion_r2910699499) - but we can follow up separately, keep the PRs small

PR #43480, opened 2026-03-06
feat(billing): show billing address modal for paid orgs with missing address

kanishkdudeja:
### Summary

This PR adds a dismissable modal that prompts paid org users to fill in their billing address. We need billing addresses for tax compliance and the existing banner hasn't been as effective (~10k orgs still missing).

The modal shows on page load when all of these are true:

- ConfigCat flag `enableBillingAddressModal` is on
- Org is on a paid plan (not free)
- Org does not have billing address set
- Org is not partner-managed
- User has BILLING_WRITE permission on stripe.customer

The modal can be dismissed via the close button, but re-opens on every page navigation.

Once the user saves a valid address, the organizations query is invalidated and organization_missing_address flips to false - the modal stops appearing permanently.

Screenshot 2026-03-05 at 4 22 11 PM

#### Automated Tests

I've added unit tests in this PR

#### Manual testing

Manual test cases:

- Free org - modal does not appear
- Partner-managed org - modal does not appear
- User with "developer role" modal does not appear
- Configcat Flag disabled - modal does not appear
- Dismiss modal, navigate to another page - modal re-opens
- Fill form and save - modal closes and does not reappear
- API failure loading billing data - shows error state with Retry button


apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+    },
+  })
+
+  return (
+    <Dialog
+      open={open}
+      onOpenChange={(value) => {
+        if (!value) setDismissed(true)
+      }}
+    >
```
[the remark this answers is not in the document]
kanishkdudeja: We don't need tracking for this

apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
```diff
@@ -0,0 +1,191 @@
...
+        await invalidateOrganizationsQuery(queryClient)
+
+        toast.success('Successfully updated billing address')
+        setDismissed(true)
+      } catch (error: any) {
+        toast.error(`Failed to update billing address: ${error.message}`)
+      } finally {
+        setIsSubmitting(false)
+      }
+    },
```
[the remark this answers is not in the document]
kanishkdudeja: We're calling `handleError(error)` from within the `updateOrganizationTaxId` mutation, so I think we should be good

apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
```diff
@@ -0,0 +1,198 @@
...
+    org.organization_missing_address &&
+    !org.billing_partner &&
+    permissionsLoaded &&
+    canBillingWrite
+
+  const open = shouldShow && !dismissed
+
+  // Reset dismissed state on navigation so modal re-opens
+  useEffect(() => {
+    setDismissed(false)
```
kevcodez: I think the modal should only pop up on initial hard page load not on every single navigation (to start)

apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
```diff
@@ -0,0 +1,198 @@
...
+      <DialogContent
+        size="medium"
+        onInteractOutside={(e) => e.preventDefault()}
+        onEscapeKeyDown={(e) => e.preventDefault()}
+      >
+        <DialogHeader>
+          <DialogTitle>Billing address required</DialogTitle>
+          <DialogDescription>
+            Please provide a billing address for your organization. If you are a registered
+            business, please add a Tax ID too.
```
kevcodez:
```suggestion
            business and have a Tax ID, please add your Tax ID too.
```

apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
```diff
@@ -0,0 +1,198 @@
...
+          </DialogDescription>
+        </DialogHeader>
+
+        {profileError || taxIdError ? (
```
kevcodez: We shouldnt be rendering the modal at all if there are any errors related to tax id / profile fetching

apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
kanishkdudeja: Thanks, addressed in https://github.com/supabase/supabase/pull/43480/commits/03ed536bceac50872325aaa3690464de9ec9c1e9

apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
kanishkdudeja: Thanks, addressed in https://github.com/supabase/supabase/pull/43480/commits/03ed536bceac50872325aaa3690464de9ec9c1e9

apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
kanishkdudeja: Thanks, addressed in https://github.com/supabase/supabase/pull/43480/commits/03ed536bceac50872325aaa3690464de9ec9c1e9

charislam (approved):
Tested on staging, LGTM

Minor nit:
- Should remove top banner for customers who see the modal

Note:
- This behaviour is not implemented as described in PR description: "Dismiss modal, navigate to another page - modal re-opens". A dismiss will persist for the entire session, only clearing when user refreshes or returns with a new session. But personally I think this is better behaviour. If we want a stronger nudge, we should make it non-dismissible, making it both dismissible and reopening on every navigation is unnecessarily infuriating 😅

kanishkdudeja:
Thank you so much for the review @charislam! 🙇 

> * This behaviour is not implemented as described in PR description: "Dismiss modal, navigate to another page - modal re-opens". A dismiss will persist for the entire session, only clearing when user refreshes or returns with a new session. But personally I think this is better behaviour. If we want a stronger nudge, we should make it non-dismissible, making it both dismissible and reopening on every navigation is unnecessarily infuriating 😅

Yeah, it was implemented previously and we removed it since it seemed overly aggressive :)

ignaciodob:
I tested this locally. Just one nitpick:

- After updating the billing address and tax id inside the modal the banner at the top stays active. Clicking on "add a billing address" takes me to the billing page correctly but the Billing Address information is not updated yet. Even though we clear the local organizations cache after saving the address the server still returns a `304 Not Modified` in the next call to `organizations`. A refresh does clear out the banner and fetches the updated billing address.

https://github.com/user-attachments/assets/ce59416e-c542-48d0-9d1d-ac1c65f46e77

PR #43485, opened 2026-03-06
Chore/add get restore initiated event endpoint

joshenlim:
## Context

Resolves FE-2709
Dependent on API PR: https://github.com/supabase/platform/pull/30564

While a project is in restoring, we should a "Contact support" CTA in the footer, but this has led to a high number of tickets coming in despite the project's status being `ACTIVE_HEALTHY` when we check on our end.

Will be adjusting the behaviour of this CTA to only show up if the project's status is RESTORING and it's `updated_at` is at than 30 minutes ago, and update the CTA a bit as well

image

image


apps/studio/components/layouts/ProjectLayout/RestoringState.tsx
```diff
@@ -70,6 +76,12 @@ const RestoringState = () => {
     if (ref) await invalidateProjectDetailsQuery(ref)
   }
 
+  useEffect(() => {
+    if (isProjectStatusSuccess && projectStatusData.status === PROJECT_STATUS.ACTIVE_HEALTHY) {
+      setIsCompleted(true)
+    }
+  }, [isProjectStatusSuccess, projectStatusData, ref, invalidateProjectDetailsQuery])
```
charislam: Does invalidateProjectDetailsQuery need to be a dependency?

charislam (approved):

apps/studio/components/layouts/ProjectLayout/RestoringState.tsx
joshenlim: oh right good catch!

joshenlim: ref neither actually

PR #43497, opened 2026-03-06
fix: added not found as errors when filter is active

awaseem:

YES

fix: small issue when a filter was active and does not found error popped up, we didn't allow users to reset the value. This handles the edge cases where a filter is applied and the column is deleted


apps/studio/components/grid/components/grid/GridError.utils.ts
```diff
@@ -6,6 +6,7 @@ const FILTER_ERROR_PATTERNS = [
   'malformed array literal',
   'invalid byte sequence',
   'syntax error',
+  'does not exist',
```
awaseem, line 9: Even though the value is set here, this function only gets called when a filter is active. So FK errors should still come through

charislam (approved):

PR #43499, opened 2026-03-06
graduate auto-RLS project creation experiment (test variant wins)

seanoliver:
## Problem

The `projectCreationEnableRlsEventTrigger` A/B experiment tested showing an opt-in "Enable automatic RLS" checkbox in the project creation flow, which sets up a Postgres event trigger to auto-enable RLS on every new table in the public schema. The test variant passed with a +3.3pp lift.

## Changes

- `SecurityOptions.tsx` — removed PostHog flag check, checkbox is now always rendered
- `pages/new/[slug].tsx` — removed experiment flag reads, exposure tracking (`useTrackExperimentExposure`), and the conditional `rlsOptionVariant` telemetry property

The underlying feature logic (`enableRlsEventTrigger` form field, `AUTO_ENABLE_RLS_EVENT_TRIGGER_SQL`, submission handling) is unchanged — we're just removing the scaffolding that was gating it.

## Testing

Verified `pnpm typecheck` passes clean. The "Enable automatic RLS" checkbox now shows unconditionally in the Security Options section of project creation.

GROWTH-653

### DIFF
--- a/apps/studio/components/interfaces/ProjectCreation/SecurityOptions.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/SecurityOptions.tsx
@@ -1,5 +1,4 @@
 import Panel from 'components/ui/Panel'
-import { usePHFlag } from 'hooks/ui/useFlag'
 import Link from 'next/link'
 import { UseFormReturn } from 'react-hook-form'
 import {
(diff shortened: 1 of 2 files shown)


awaseem (approved):
LGTM needs code formatting, approving to unblock

PR #43516, opened 2026-03-07
fix(studio): support enum creation in custom schemas (#30632)

### LINKED ISSUE #30632: Creating enum in custom schema fails
# Bug report

I am experiencing this bug in local and hosted versions in slightly different ways - but seems to be related:

# Local version issue
For a brand new Supabase project, whenever I try to create an enum for a custom schema, either via the Studio UI or the Studio's SQL editor, it never creates the enum (even though in both cases it claims it has). The same project hosted on Supabase creates them fine.

For example via the Studio UI, I create the enum, the toast appears saying it's successfully created, but the inline table with the created enums never shows my new one.

# Hosted version issue
For a brand new Supabase project, I am able to create an enum for a custom schema fine and it shows up, however whenever I try to then use that enum in a new column of a table within that custom schema, I get the error: `failed to create pg.columns: type "my_enum_name" does not exist`.

This issue in the hosted version again does not happen if I'm using the `public` schema.

dumko2001:

YES

Bug fix

Currently, if a user creates an enumerated type (enum) in a custom (non-`public`) schema, the Studio UI successfully retrieves and lists the type in the column creation dropdown. However, when the user selects it and clicks save, the UI silently strips the schema prefix and downcases the value (due to Shadcn's internal combobox filtering logic). The API call attempts to create the column using the unqualified enum string, resulting in the Postgres error `type "my_enum" does not exist` (as it assumes the `public` schema).

Furthermore, in `pg-meta`, if a user attempts to define a column as an array of a custom enumerated type (`"custom_schema"."my_enum"[]`), the `typeIdent` logic strips the `[]` but incorrectly double-escapes the already-qualified string, mangling it into `"""custom_schema"".""my_enum"""[]` and causing an immediate crash.

Resolves #30632

## What is the new behavior?

1. **Studio (`ColumnType.tsx`)**: Reconstructed the enum option mappers and selectors to strictly build and pass the correctly quoted, fully-qualified identifier (e.g. `"custom_schema"."my_enum"`) when the selected option belongs to a non-public schema. This completely shields the payload from Shadcn's format-stripping UI bugs.
2. **pg-meta (`pg-meta-columns.ts`)**: Patched `typeIdent` to safely recognize schema-qualified (period-containing) array formats without redundantly triggering `ident()`, guaranteeing seamless compatibility with `custom_schema.my_enum[]` declarations.

## Additional context

This follows a strict first-principles approach. By addressing the root formatting mismatches directly at the component bounds (`ColumnType.tsx`) and the SQL schema stringifier (`pg-meta-columns.ts`), it requires zero architectural changes and introduces no regressions to the default `public` schema workflows.

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnType.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/ColumnEditor/ColumnType.tsx
@@ -73,6 +73,10 @@ const ColumnType = ({
   const listboxId = useId()
   const availableTypes = POSTGRES_DATA_TYPES.concat(
     enumTypes.map((type) => type.format.replaceAll('"', ''))
+  ).concat(
+    enumTypes.map((option) =>
+      option.schema === 'public' ? option.name : `"${option.schema}"."${option.name}"`
+    )
   )
   const isAvailableType = value ? availableTypes.includes(value) : true
   const recommendation = RECOMMENDED_ALTERNATIVE_DATA_TYPE[value]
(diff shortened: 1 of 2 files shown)


dumko2001:
Hey @coderabbitai, excellent catch! 

I've pushed a follow-up commit that completely refactors `typeIdent()` so it explicitly parses the schema-qualified string, extracts the individual schema and type components, and securely applies `ident()` to each of them separately. 

This guarantees that any crafted dotted values are safely escaped, preventing arbitrary DDL injection while preserving full support for custom-schema enumerations and arrays. Thanks for the rigorous review!

PR #43532, opened 2026-03-08
fix(studio): wrap project actions

dnywh:

- UI fix

- The _New project_ button overflows to the right on smaller (phone-sized) breakpoints
- Due to addition of _Sort by_ dropdown last week

## What is the new behavior?

- All actions now wrap on smaller breakpoint

| Before | After |
| --- | --- |
| Supabase-383C523F-0BF1-48CD-8052-F6D68301DA4E | Supabase-ED7AD923-E837-4D48-B7D1-6DA3C6E45BFA |


joshenlim (approved):
Verified changes on preview - both on desktop + mobile viewport widths

PR #43534, opened 2026-03-09
feat(studio): page titles for project surfaces

dnywh:
Stacked PR 2/5 for page title improvements. Includes Auth, Database, editors, and Edge Functions layout title adoption.

_Base: [dnywh/feat/page-titles](https://github.com/supabase/supabase/pull/43538)_

---

- Resolves FE-1960
- Resolves FE-1983
- Resolves DEPR-207

Page titles between surfaces are inconsistent and vague. Sometimes they say the product name:

```
My Project | My Org | Supabase
```

...even when on a specific surface like Database > Tables.

Other times they show the entity name but skip over the project or org name :

```
Edge Functions | Supabase
```

## What is the new behavior?

Auth, Database, editors, and Edge Functions adopt the layout title format introduced in https://github.com/supabase/supabase/pull/43538:

```
users | Table Editor | My Project | My Org | Supabase
hello-world | Logs | Edge Functions | My Project | My Org | Supabase
Backups | Database | My Project | My Org | Supabase
Authentication | My Project | My Org | Supabase
```

That format is:

entity, section, surface, project, org, brand

## Additional context

Related stacked PRs also based on the the original [dnywh/feat/page-titles](https://github.com/supabase/supabase/pull/43538) branch:

- https://github.com/supabase/supabase/pull/43535
- https://github.com/supabase/supabase/pull/43536
- https://github.com/supabase/supabase/pull/43537


apps/studio/components/layouts/EdgeFunctionsLayout/EdgeFunctionsLayout.tsx
```diff
@@ -35,11 +35,21 @@ const EdgeFunctionsProductMenu = () => {
...
+
+const EdgeFunctionsLayout = ({
+  children,
+  title = 'Edge Functions',
+  browserTitle,
+}: PropsWithChildren<EdgeFunctionsLayoutProps>) => {
   return (
     <ProjectLayout
-      title="Edge Functions"
+      title={title}
```
joshenlim, line 50: i just realised that this is bit of a confusing DX that we have a `title` property and a `browserTitle` property

was checking ProjectLayout and see that `title` is just being used in `buildStudioPageTitle` which also tries to read from `browserTitle`

could we just use browserTitle? and deprecate `title`? although do let me know if theres something that i might be overlooking!

dnywh: Hmm, that is confusing DX. I just spiked this change (which I agree with) and it adds a bit of heft to the stacked PRs.

Perhaps we can mark `title` as deprecated in `ProjectLayoutProps` (e.g. `@deprecated use browserTitle.section`) and migrate wrappers/pages to `browserTitle` incrementally? I can do that after these stacked PRs?

apps/studio/components/layouts/AuthLayout/AuthLayout.tsx
```diff
@@ -8,22 +8,45 @@ import { PropsWithChildren } from 'react'
 import { ProjectLayout } from '../ProjectLayout'
 import { useGenerateAuthMenu } from './AuthLayout.utils'
 
+const AUTH_SECTION_TITLE_BY_PAGE: Record<string, string> = {
```
joshenlim: just a DX feedback on this one too - instead of having the AuthLayout keep a mapping of page route to title, am thinking we let the parent pass this value as a prop instead

so we'd render the layout as <AuthLayout title="...." /> instead

that way its easier to keep in mind to set a title when we create new pages and we don't need to manually set page titles via splitting on the router pathname

this is just like what we're doing for EdgeFunctionsLayout - so also a factor for consistency
ensure that title is a required prop into the layout components

joshenlim: will be the same for database layout and edge function details layout

joshenlim: e.g for restore to new project - we left out its title in DatabaseLayout

so reckon it'll be better to declare at the page level instead, make title a necessary prop

joshenlim (approved):

apps/studio/pages/project/[ref]/functions/index.tsx
```diff
@@ -178,7 +178,7 @@ const EdgeFunctionsPage: NextPageWithLayout = () => {
 EdgeFunctionsPage.getLayout = (page: React.ReactElement) => {
   return (
     <DefaultLayout>
-      <EdgeFunctionsLayout>
+      <EdgeFunctionsLayout title="Edge Functions">
```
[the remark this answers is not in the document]
Comment (dnywh) [reply apps/studio/pages/project/[ref]/functions/index.tsx:181]: Has no visible impact right now. Will defer polish until after stacked PRs are in.

PR #43540, opened 2026-03-09
chore(studio): mention project OAuth app connections

dnywh:

- UI change
- Resolves DEPR-324

- Projects that are connected to OAuth applications such as Lovable, Bolt, Figma Make **do not** indicate this fact to the user
- User changes in the dashboard may conflict with what those OAuth apps expect

## What is the new behavior?

- `ProjectOAuthIntegrationsBanner` added within `ContentWrapper` in order to give user notice when this case is true
- Banner styled and acts much like the “_This organization is managed by_” banner in [`OrganizationLayout`](https://github.com/supabase/supabase/blob/c26c6400381db38a963aa7cf6d44d66f6771db9a/apps/studio/components/layouts/OrganizationLayout.tsx)
- Banner links out to OAuth app management in org settings
- Icon change  if passed
- Feature preview and matching flag to control all this for Supabase folks only

Screenshots taken with maintenance banner for stress-testing:

| After |
| --- |
| AWS Healthy  Toolshed  Supabase-72366DA4-D649-4BAE-A119-845187F6787E |
| AWS Healthy  Toolshed  Supabase |
| AWS Healthy  Toolshed  Supabase-3CEB0655-F7E4-4EE0-A018-BCB62D5D1ED3 |

## Additional context

I was unable to create and therefore test OAuth apps locally. Perhaps we should release behind feature flag or figure out a way to test in staging.


dnywh:
Cancelling because there is currently no way to determine which Supabase projects are connected to what OAuth apps. That connection happens at the org level.

dnywh:
Re-adding the same sort of thing in https://github.com/supabase/supabase/pull/44328 for Stripe Projects. We should generalise that for OAuth use should we ever be able to get that data in the endpoint.

PR #43541, opened 2026-03-09
Chore/fix storage explorer when switching buckets

joshenlim:
## Context

Taking a slightly different approach to [this PR](https://github.com/supabase/supabase/pull/43370)

Original problem was that if you opened some folders while in a bucket and then switched to a different bucket, the folder UI will persists (folders from Bucket A will render when landing on Bucket B)

## Changes involved
- Shift `StorageExplorerStateContextProvider` into `[bucketId].tsx]` instead of `ProjectContext`
  - The valtio store here only applies for the storage explorer so having it so high in the project's context was unnecessary
  - This also just implies that the valtio store will automatically reset whenever the bucket changes
- Simplify storage explorer valtio store by initializing the store with the bucket
  - We'll initialize the selected bucket with the store now (Same as previous PR)
  - Removes unnecessary `setSelectedBucket` method which required a separate `useEffect` in `StorageExplorer.tsx`

## To test
- [ ] Verify that the original is resolved
- [ ] General smoke test of the storage explorer - i've also re-added the e2e test that Gildas wrote up in his PR

### DIFF
--- a/apps/studio/components/interfaces/Storage/StorageExplorer/StorageExplorer.tsx
+++ b/apps/studio/components/interfaces/Storage/StorageExplorer/StorageExplorer.tsx
@@ -1,14 +1,13 @@
-import { compact, get, isEmpty, uniqBy } from 'lodash'
-import { useEffect, useRef, useState } from 'react'
-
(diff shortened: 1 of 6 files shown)


djhi (approved):
👍 Tested on staging

PR #43547, opened 2026-03-09
chore: e2e general improvements

djhi:
## Problem

e2e tests are still flaky and not as fast as they could be

## Solution

- [x] Reset supabase instance: this makes the database visualiser tests faster as there are less tables to screenshot
- [x] Improve the single file setup utilities so that they never block local tests by cleaning them up before starting
- [x] Disable animations while running the tests (less time waiting for animations to complete
- [x] Add utility functions that help reproducing flaky tests locally


awaseem (approved):
LGTM

PR #43563, opened 2026-03-09
feat: Implement telemetry on Search commands

jeremenichelli:

YES

This PR adds telemetry to `DocsCommandProvider` so we gather data about Search usage and define ways to improve it.

Search lacks analytics.

## What is the new behavior?

We send data to our analytics dashboard about Search usage.

## Additional context

### DIFF
--- a/apps/docs/features/command/DocsCommandProvider.tsx
+++ b/apps/docs/features/command/DocsCommandProvider.tsx
@@ -1,12 +1,26 @@
 'use client'
 
-import type { PropsWithChildren } from 'react'
+import { useEffect, type PropsWithChildren } from 'react'
 
+import { posthogClient, useUser } from 'common'
+import { useConsentToast } from 'ui-patterns/consent'
 import { CommandProvider } from 'ui-patterns/CommandMenu'
 import { useDocsCommandMenuTelemetry } from '@/hooks/useDocsCommandMenuTelemetry'
 
 export function DocsCommandProvider({ children }: PropsWithChildren) {
   const { onTelemetry } = useDocsCommandMenuTelemetry()
+  const user = useUser()
+  const { hasAcceptedConsent } = useConsentToast()
+
+  useEffect(() => {
+    posthogClient.init(hasAcceptedConsent)
+  }, [hasAcceptedConsent])
+
+  useEffect(() => {
+    if (user?.id) {
+      posthogClient.identify(user.id, { gotrue_id: user.id }, hasAcceptedConsent)
+    }
+  }, [user?.id, hasAcceptedConsent])
 
   return (
     <CommandProvider app="docs" onTelemetry={onTelemetry}>
--- a/apps/docs/features/command/DocsCommandProvider.tsx
+++ b/apps/docs/features/command/DocsCommandProvider.tsx
@@ -1,26 +1,12 @@
 'use client'
 
-import { useEffect, type PropsWithChildren } from 'react'
+import { type PropsWithChildren } from 'react'
 
-import { posthogClient, useUser } from 'common'
-import { useConsentToast } from 'ui-patterns/consent'
 import { CommandProvider } from 'ui-patterns/CommandMenu'
 import { useDocsCommandMenuTelemetry } from '@/hooks/useDocsCommandMenuTelemetry'
 
 export function DocsCommandProvider({ children }: PropsWithChildren) {
   const { onTelemetry } = useDocsCommandMenuTelemetry()
-  const user = useUser()
-  const { hasAcceptedConsent } = useConsentToast()
-
-  useEffect(() => {
-    posthogClient.init(hasAcceptedConsent)
-  }, [hasAcceptedConsent])
-
-  useEffect(() => {
-    if (user?.id) {
-      posthogClient.identify(user.id, { gotrue_id: user.id }, hasAcceptedConsent)
-    }
-  }, [user?.id, hasAcceptedConsent])
 
   return (
     <CommandProvider app="docs" onTelemetry={onTelemetry}>
--- a/apps/docs/features/command/DocsCommandProvider.tsx
+++ b/apps/docs/features/command/DocsCommandProvider.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { type PropsWithChildren } from 'react'
+import type { PropsWithChildren } from 'react'
 
 import { CommandProvider } from 'ui-patterns/CommandMenu'
 import { useDocsCommandMenuTelemetry } from '@/hooks/useDocsCommandMenuTelemetry'
--- a/apps/docs/hooks/useDocsCommandMenuTelemetry.ts
+++ b/apps/docs/hooks/useDocsCommandMenuTelemetry.ts
@@ -8,12 +8,10 @@ import type {
   CommandMenuOpenedEvent,
   CommandMenuSearchSubmittedEvent,
 } from 'common/telemetry-constants'
-import { useConsentToast } from 'ui-patterns/consent'
 import { useSendTelemetryEvent } from '@/lib/telemetry'
 
 export function useDocsCommandMenuTelemetry() {
   const sendTelemetryEvent = useSendTelemetryEvent()
-  const { hasAcceptedConsent } = useConsentToast()
   const onTelemetry = useCallback(
     (
       event:
@@ -24,7 +22,7 @@ export function useDocsCommandMenuTelemetry() {
     ) => {
       sendTelemetryEvent(event)
     },
-    [sendTelemetryEvent, hasAcceptedConsent]
+    [sendTelemetryEvent]
   )
 
   return { onTelemetry }
--- a/apps/docs/hooks/useDocsCommandMenuTelemetry.ts
+++ b/apps/docs/hooks/useDocsCommandMenuTelemetry.ts
@@ -23,8 +23,6 @@ export function useDocsCommandMenuTelemetry() {
         | CommandMenuCommandClickedEvent
         | CommandMenuSearchSubmittedEvent
     ) => {
-      console.log('--- hasAcceptedConsent ---', hasAcceptedConsent)
-      console.log('EVENT', event)
       sendTelemetryEvent(event)
       posthogClient.capture(event.action, event.properties, hasAcceptedConsent)
     },
--- a/apps/docs/hooks/useDocsCommandMenuTelemetry.ts
+++ b/apps/docs/hooks/useDocsCommandMenuTelemetry.ts
@@ -3,25 +3,32 @@
 import { useCallback } from 'react'
 
 import type {
-  CommandMenuOpenedEvent,
+  CommandMenuClosedEvent,
   CommandMenuCommandClickedEvent,
+  CommandMenuOpenedEvent,
   CommandMenuSearchSubmittedEvent,
 } from 'common/telemetry-constants'
+import { posthogClient } from 'common'
+import { useConsentToast } from 'ui-patterns/consent'
 import { useSendTelemetryEvent } from '@/lib/telemetry'
 
 export function useDocsCommandMenuTelemetry() {
   const sendTelemetryEvent = useSendTelemetryEvent()
-
+  const { hasAcceptedConsent } = useConsentToast()
   const onTelemetry = useCallback(
     (
       event:
-        | CommandMenuOpenedEvent
-        | CommandMenuCommandClickedEvent
-        | CommandMenuSearchSubmittedEvent
+      | CommandMenuOpenedEvent
+      | CommandMenuClosedEvent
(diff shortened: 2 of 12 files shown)


apps/docs/hooks/useDocsCommandMenuTelemetry.ts
```diff
@@ -3,25 +3,32 @@
...
   const onTelemetry = useCallback(
     (
       event:
         | CommandMenuOpenedEvent
+        | CommandMenuClosedEvent
         | CommandMenuCommandClickedEvent
         | CommandMenuSearchSubmittedEvent
     ) => {
+      console.log('--- hasAcceptedConsent ---', hasAcceptedConsent)
+      console.log('EVENT', event)
```
seanoliver: These `console.log` calls fire on every telemetry event — including every debounced keystroke via `command_menu_search_submitted`. They'll spam the browser console for all users. Please remove both before merge.

```ts
// remove these two lines
console.log('--- hasAcceptedConsent ---', hasAcceptedConsent)
console.log('EVENT', event)
```

apps/docs/features/command/DocsCommandProvider.tsx
```diff
@@ -1,12 +1,26 @@
...
+
+  useEffect(() => {
+    posthogClient.init(hasAcceptedConsent)
+  }, [hasAcceptedConsent])
+
+  useEffect(() => {
+    if (user?.id) {
+      posthogClient.identify(user.id, { gotrue_id: user.id }, hasAcceptedConsent)
+    }
+  }, [user?.id, hasAcceptedConsent])
```
seanoliver: `<PageTelemetry />` is already mounted in the docs app via `app.providers.tsx` and is responsible for both `posthogClient.init()` (on consent change) and `posthogClient.identify()` (on user change). Adding a second init/identify path here creates a race condition where both codepaths compete to call `init()` with potentially different consent values, and `identify()` may fire twice per session.

These two `useEffect` blocks (and the `user`, `hasAcceptedConsent`, `posthogClient`, `useUser`, and `useConsentToast` imports) should be removed from `DocsCommandProvider`. PostHog lifecycle shouldn't be owned here.

packages/ui-patterns/src/CommandMenu/api/CommandList.tsx
```diff
@@ -32,21 +32,27 @@ const CommandList = forwardRef<
...
+
+          const visibleCommands = section.commands.filter(
+            (command) => !command.defaultHidden || query
+          )
+
+          return (
+            <CommandGroup key={section.id} heading={section.name} forceMount={section.forceMount}>
+              {visibleCommands.map((command, index) => {
+                return (
+                  <CommandItem key={command.id} command={command} commandIndex={index}>
```
seanoliver: The `index` here resets to `0` for each section. So a user clicking the 3rd item in the 2nd section gets `result_index: 2` — identical to the 3rd item in the 1st section. This contradicts the JSDoc in `telemetry-constants.ts` which documents `result_index` as "Zero-based index of the clicked item **across all visible results**", and makes the property useless for ranking analysis.

To fix this, you'd need to track a running counter across sections before passing it down to `CommandItem`. Something like accumulating `visibleCommands.length` per section as you iterate.

packages/ui-patterns/src/CommandMenu/prepackaged/DocsSearch/DocsSearchPage.tsx
```diff
@@ -211,6 +230,11 @@ const DocsSearchPage = () => {
                           'ml-3 mb-3'
```
seanoliver, line 232: The `i` passed to `trackResultClicked` here (and for sections at line 230) is the index within the current page group, not the global position across all results. A user clicking the first section result of the second page gets `result_index: 0` — same as the first result of the first page. Same semantics problem as in `CommandList.tsx`.

To get a meaningful global index, you'd need to accumulate a running offset across pages and their sections as you iterate.

packages/ui-patterns/src/CommandMenu/api/hooks/useCommandMenuTelemetry.ts
seanoliver, line 1: In `useCommandMenuOpenedTelemetry` (line ~72 in the full file), the opened event is being emitted with `location: 'user_dropdown_menu'`. But `CommandMenuOpenedEvent` in `telemetry-constants.ts` defines the property as `trigger_location`, not `location`. TypeScript isn't catching this because the callback typing is loose. The value silently becomes a non-standard PostHog property and won't appear under `trigger_location` in your PostHog queries.

This is a pre-existing bug in this file, but since it's touched by this PR it's worth fixing here:

```ts
// current (wrong key)
location: 'user_dropdown_menu',

// should be
trigger_location: 'user_dropdown_menu',
```

seanoliver (changes requested):
Hey @jeremenichelli! Thanks for putting this together. The overall intent and architecture look solid, and the new events are well scoped and nicely MECE.

I’ve left a few comments calling out some blockers that we should address before merging. Happy to discuss any of them in more detail if helpful.

apps/docs/hooks/useDocsCommandMenuTelemetry.ts
jeremenichelli: Oh definitely, I left it because I wanted to have a clear way to debug these events in the preview environment, also why I kept this as a draft. Definitely removing it after feedback.

packages/ui-patterns/src/CommandMenu/api/CommandList.tsx
jeremenichelli: Oh yes, you are totally right. This bit was done via Claude Code, and I didn't catch this one. Seriously considering ignoring the `result_index` at all, since I don't know if it's totally necessary for the analysis we need to do right now.

packages/ui-patterns/src/CommandMenu/prepackaged/DocsSearch/DocsSearchPage.tsx
jeremenichelli: Yeap, same as above. I think I will try to keep the scope down and remove this property for now.

packages/ui-patterns/src/CommandMenu/api/CommandList.tsx
```diff
@@ -32,21 +32,27 @@ const CommandList = forwardRef<
...
-              .filter((command) => !command.defaultHidden || query)
-              .map((command) => (
-                <CommandItem key={command.id} command={command}>
-                  <TextHighlighter>{command.name}</TextHighlighter>
-                </CommandItem>
-              ))}
-          </CommandGroup>
-        )
-      })}
+      {(() => {
```
seanoliver: This IIFE was probably left over from when `result_index` needed an accumulator across sections. Now that `result_index` is removed, the IIFE wrapper doesn't serve a purpose — this can be simplified back to a plain `.map()`:

```tsx
{commandSections.map((section) => {
  if (section.commands.every((command) => command.defaultHidden) && !query) return null

  const visibleCommands = section.commands.filter(
    (command) => !command.defaultHidden || query
  )

  return (
    <CommandGroup key={section.id} heading={section.name} forceMount={section.forceMount}>
      {visibleCommands.map((command) => (
        <CommandItem key={command.id} command={command}>
          <TextHighlighter>{command.name}</TextHighlighter>
        </CommandItem>
      ))}
    </CommandGroup>
  )
})}
```

apps/docs/hooks/useDocsCommandMenuTelemetry.ts
```diff
@@ -3,25 +3,30 @@
...
   const onTelemetry = useCallback(
     (
       event:
         | CommandMenuOpenedEvent
+        | CommandMenuClosedEvent
         | CommandMenuCommandClickedEvent
         | CommandMenuSearchSubmittedEvent
     ) => {
       sendTelemetryEvent(event)
+      posthogClient.capture(event.action, event.properties, hasAcceptedConsent)
```
seanoliver: This call to `posthogClient.capture()` is actually redundant with the line above. `sendTelemetryEvent()`, already POSTs to our `/telemetry/event` backend, which forwards to PostHog server-side. Calling `posthogClient.capture()` again here would send the same event again directly via the client-side SDK.

packages/common/telemetry-constants.ts
```diff
@@ -2291,14 +2291,22 @@ export interface CommandMenuSearchSubmittedEvent {
  * @page any
  */
 export interface CommandMenuCommandClickedEvent {
-  action: 'command_menu_command_clicked'
+  action: 'command_menu_item_clicked'
```
seanoliver: This is kind of a nit but the interface is still named `CommandMenuCommandClickedEvent` while the action is now called `'command_menu_item_clicked'`. We try to keep these the same for readability.

packages/ui-patterns/src/CommandMenu/api/CommandList.tsx
seanoliver: Now that you've removed the accumulator for `result_index`, I don't think you need this IIFE anymore.

packages/ui-patterns/src/CommandMenu/prepackaged/DocsSearch/DocsSearchPage.tsx
```diff
@@ -75,10 +76,26 @@ const DocsSearchPage = () => {
   const setQuery = useSetQuery()
   const query = useQuery()
 
+  const telemetryContext = useCommandMenuTelemetryContext()
+
   const initialLoad = useRef(true)
   const inputRef = useRef<HTMLInputElement>(null)
   const router = useCrossCompatRouter()
 
+  function trackResultClicked(name: string, path: string) {
```
seanoliver: Nit: this function gets recreated on every render. Consider wrapping it with `useCallback()`. (This is also the pattern we use elsewhere.)

Not critical since it only fires on click, but keeps things consistent.

packages/common/telemetry-constants.ts
```diff
@@ -2307,6 +2315,24 @@ export interface CommandMenuCommandClickedEvent {
...
+export interface CommandMenuClosedEvent {
+  action: 'command_menu_closed'
+  properties: {
+    /**
+     * In which app the command menu was closed
+     */
+    app: 'studio' | 'docs' | 'www'
+  }
+  groups: Partial<TelemetryGroups>
+}
```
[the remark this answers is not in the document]
seanoliver: @pamelachia so there actually is precedent for the verb "closed" in line 1849 (`home_getting_started_closed`). Is it worth updating coderabbit to allow `closed` — seems like a reasonable addition to the nomenclature!

packages/ui-patterns/src/CommandMenu/api/hooks/useCommandMenuTelemetry.ts
seanoliver: TY!

seanoliver (changes requested):
This looks great! Really the only thing here that is still blocking is the duplicate telemetry event capture call. Everything else are just minor style nits.

packages/common/telemetry-constants.ts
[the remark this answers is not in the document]
pamelachia: @seanoliver sounds good. You can update coderabbit and our other docs to include closed as a verb too.

apps/docs/hooks/useDocsCommandMenuTelemetry.ts
jeremenichelli: I added the `capture` event because I didn't see a clear method for it, but now I am guessing I could delete it and just use `sendTelemetry`. I will also delete that function.

jeremenichelli (commented):
Thanks for the review @seanoliver, I realized there was a lot of unwanted changes not properly cleaned from a previous coding session. I think all is cleaned up right now.

packages/ui-patterns/src/CommandMenu/api/CommandMenu.tsx
```diff
@@ -249,7 +262,7 @@ function CommandMenu({ children, trigger }: CommandMenuProps) {
         onInteractOutside={() => setOpen(false)}
```
seanoliver: One last minor thing I just noticed. This still calls `setOpen(false)` directly, so clicking outside the menu to dismiss it won't fire `command_menu_closed`. The escape key path was updated to use `handleOpenChange(false)` so we just need the same treatment here:

```suggestion
        onInteractOutside={() => handleOpenChange(false)}
```

seanoliver (approved):
Looks great — thanks @jeremenichelli! I left one last comment in here which I only just noticed but approving to unblock. 🙌

fsansalvadore (approved):

ivasilov (approved):

PR #43564, opened 2026-03-09
feat: improve db overload debugging UX

jordienr:
When the dashboard hits a DB connection timeout, users currently see a raw error message with no
  path forward. This PR adds an inline troubleshooting system that detects known error types and
  surfaces contextual next steps — restart the DB, read the docs, or debug with AI.

##  Changes

-  New ErrorDisplay component (packages/ui-patterns) — styled error card with a title, monospace error
   block, optional troubleshooting slot, and a "Contact support" link that always renders. Accepts
  typed supportFormParams to pre-fill the support form.

-  Error classification in handleError (data/fetchers.ts) — on every API error, the message is tested
  against ERROR_PATTERNS. If matched, handleError throws a typed subclass (ConnectionTimeoutError
  extends ResponseError) instead of a plain ResponseError. Stack traces now show the exact error
  class. All existing instanceof ResponseError checks continue to work.

-  ErrorMatcher component — reads errorType from the thrown class instance, does an O(1) lookup into
  ERROR_MAPPINGS, and renders the matching troubleshooting accordion as children of ErrorDisplay.
  Falls back to plain ErrorDisplay for unclassified errors.

-  Connection timeout mapping — first error type wired up, with three troubleshooting steps: restart
  the database, link to the docs, and "Debug with AI" (opens the AI assistant sidebar with a
  pre-filled prompt).

-  Telemetry — three new typed events track when the troubleshooter is shown, when accordion steps are
   toggled, and which CTAs are clicked.

##  Adding a new error type

  1. Add a class to types/api-errors.ts
  2. Add { pattern, ErrorClass } to data/error-patterns.ts
  3. Create a troubleshooting component in errorMappings/
  4. Add an entry to error-mappings.tsx


apps/studio/components/layouts/TableEditorLayout/TableEditorMenu.tsx
```diff
@@ -276,10 +276,23 @@ export const TableEditorMenu = () => {
...
 
+          {/* 🚨 DELETE BEFORE MERGING */}
+          {process.env.NEXT_PUBLIC_ENVIRONMENT !== 'prod' && (
+            <ErrorMatcher
+              title="Failed to load tables"
+              error="ERROR: FAILED TO RUN SQL QUERY: CONNECTION TERMINATED DUE TO CONNECTION TIMEOUT."
+              supportFormParams={{ projectRef }}
+              className="mx-4 mt-3"
+            />
+          )}
```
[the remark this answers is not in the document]
jordienr: i know man

packages/ui-patterns/src/admonition.tsx
```diff
@@ -55,11 +55,11 @@ const InfoIcon = () => (
   </svg>
 )
 
-const WarningIcon = () => (
+export const WarningIcon = ({ className = 'w-6 h-6' }: { className?: string }) => (
   <svg
     xmlns="http://www.w3.org/2000/svg"
     viewBox="0 0 22 20"
-    className="w-6 h-6"
+    className={className}
```
ivasilov: You can just do `cn('w-6 h-6', className)` and the classes can be overridden by `className`

apps/studio/types/api-errors.ts
```diff
@@ -0,0 +1,36 @@
+import { ResponseError } from './base'
+import type { ErrorMetadata } from './base'
+
+export type KnownErrorType = 'connection-timeout'
+
+export class ConnectionTimeoutError extends ResponseError {
+  readonly errorType = 'connection-timeout' as const
```
ivasilov, line 7: Add `errorType` as prop to `ResponseError`.

apps/studio/components/interfaces/ErrorHandling/ErrorMatcher.utils.ts
```diff
@@ -0,0 +1,13 @@
+import type { KnownErrorType } from 'types/api-errors'
+
+import { ERROR_MAPPINGS, type ErrorMapping } from './error-mappings'
+
+function isKnownErrorType(type: unknown): type is KnownErrorType {
+  return typeof type === 'string' && type in ERROR_MAPPINGS
+}
+
+export function getMappingForError(error: unknown): ErrorMapping | null {
```
charislam, line 5: Since we're using classes to tag errors, wouldn't it make more sense to use an `instanceof` check rather than a string match? That way it will only ever match on the class, and we don't have to worry about keeping strings in sync

apps/studio/data/error-patterns.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      'idle connection timeout',
+      '',
+    ],
+  },
+}
+
+describe('ERROR_PATTERNS registry', () => {
+  it('has no duplicate ErrorClass entries', () => {
+    const classes = ERROR_PATTERNS.map((p) => p.ErrorClass)
+    const unique = new Set(classes)
```
charislam: What if ERROR_PATTERNS is a Map<Class, ReactNode>, so it deduplicates by default?

apps/studio/data/error-patterns.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { ResponseError } from 'types/base'
+
+interface ErrorPattern {
+  pattern: RegExp
+  ErrorClass: new (...args: ConstructorParameters<typeof ResponseError>) => ClassifiedError
+}
+
+export const ERROR_PATTERNS: ErrorPattern[] = [
+  {
+    pattern: /CONNECTION\s+TERMINATED\s+DUE\s+TO\s+CONNECTION\s+TIMEOUT/i,
```
charislam: Rather than a global matcher, could we do this on a per-query basis? so e.g., it doesn't make sense to match db errors against auth calls
```
handleError(error, (error) => /connection/i.matches(error) ? new ConnectionError('bad', { cause: error }) : new ResponseError())
```

ivasilov: I think it's ok if we match for all API calls. It's still exactly the same as before (`ResponseError`) but we can add specific UI where we want.

charislam: Not dying on this hill so happy to accept it either way, but my thinking is it allows us to match even when error messages from the API are non-specific, maybe the same error message should be handled differently if it comes back from auth vs. storage as an example

packages/common/telemetry-constants.ts
```diff
@@ -2566,7 +2566,69 @@ export interface DashboardErrorCreatedEvent {
     /**
      * Source of the error
      */
-    source?: 'admonition' | 'toast'
+    source?: 'admonition' | 'toast' | 'error_display'
+    /**
+     * Type of error matched (for error_display source)
+     */
+    errorType?: string
```
seanoliver: nit: `errorType` and `hasTroubleshooting` are defined here but i don't see anywhere in the PR that actually fires `DashboardErrorCreatedEvent` with `source: 'error_display'`.

apps/studio/components/interfaces/ErrorHandling/ErrorMatcher.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <ErrorDisplay
+      title={title}
+      errorMessage={message}
+      supportFormParams={supportFormParams}
+      className={className}
+      onRender={
+        mapping
+          ? () => track('inline_error_troubleshooter_shown', { errorType: mapping.id })
```
seanoliver: oops — i originally suggested `shown` in slack but `inline_error_troubleshooter_exposed` would be more consistent with our approved verbs list for event names. could we rename `_shown` to `_exposed`?

also worth thinking about deduplication here — `ErrorDisplay` already has `hasFired.current` which prevents duplicate fires within a single mount, but that won't survive a page refresh. page refreshes might be an edge case in this flow, but if you're worried about it you could borrow `sessionStorage` pattern we use in `fireExposureIfNew` in `packages/common/posthog-client.ts`. (but fine to ignore if that's overkill for this use case)

apps/studio/components/interfaces/ErrorHandling/TroubleshootingAccordion.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <Accordion
+      type="single"
+      collapsible
+      defaultValue={defaultValue}
+      className={cn('w-full', className)}
+      onValueChange={(value) => {
+        const expanded = Boolean(value)
+        const step = expanded ? parseInt(value.replace('step-', ''), 10) : 0
```
seanoliver: nit: when a step collapses, `value` is `""` so `step` ends up as `0`. since steps are 1-indexed, any PostHog query filtering by step number will see a `0` for all close events, which is a bit odd. might be cleaner to use `null` here (or just omit `step` entirely when `expanded` is false).

apps/studio/components/interfaces/ErrorHandling/TroubleshootingAccordion.tsx
```diff
@@ -0,0 +1,48 @@
...
+        track('inline_error_troubleshooter_accordion_toggled', {
```
seanoliver: oops again! i had suggested `toggled` in slack but `inline_error_troubleshooter_step_clicked` would be more consistent with our naming conventions. sorry for the back and forth!

also keeping `expanded` as a property is totally fine so you can still distinguish opens from closes.

packages/common/telemetry-constants.ts
seanoliver, line 2573: nit: `errorType` and `hasTroubleshooting` are defined here but i don't see anywhere in the PR that actually fires `DashboardErrorCreatedEvent` with `source: 'error_display'`.

packages/common/telemetry-constants.ts
```diff
@@ -2566,7 +2566,69 @@ export interface DashboardErrorCreatedEvent {
...
+}
+
+/**
+ * Triggered when the inline error troubleshooter is shown to the user.
+ *
+ * @group Events
+ * @source studio
+ */
+export interface InlineErrorTroubleshooterShownEvent {
+  action: 'inline_error_troubleshooter_shown'
```
seanoliver: same as above RE event naming — suggest renaming to `inline_error_troubleshooter_exposed` for consistency

packages/common/telemetry-constants.ts
```diff
@@ -2566,7 +2566,69 @@ export interface DashboardErrorCreatedEvent {
...
+}
+
+/**
+ * Triggered when a user opens or closes a troubleshooting accordion step.
+ *
+ * @group Events
+ * @source studio
+ */
+export interface InlineErrorTroubleshooterAccordionToggledEvent {
+  action: 'inline_error_troubleshooter_accordion_toggled'
```
seanoliver: same as above RE event naming — suggest renaming to `inline_error_troubleshooter_step_clicked` for consistency

packages/common/telemetry-constants.ts
```diff
@@ -2566,7 +2566,69 @@ export interface DashboardErrorCreatedEvent {
...
+ * @group Events
+ * @source studio
+ */
+export interface InlineErrorTroubleshooterActionClickedEvent {
+  action: 'inline_error_troubleshooter_action_clicked'
+  properties: {
+    /** ID of the matched error mapping */
+    errorType: string
+    /** Which action was clicked */
+    action: 'restart_db' | 'troubleshooting_guide' | 'ask_ai' | 'contact_support'
```
seanoliver: tiny nit: this `action` property shares a name with the outer `action` field (the event name) in the same interface. not a runtime problem, but could be confusing when both show up as "action" in PostHog queries. maybe rename to `ctaType`?

seanoliver (commented):
Looking good from a growth eng / telemetry POV! I just dropped in a few minor notes on naming consistency that I noticed!

apps/studio/components/interfaces/ErrorHandling/ErrorMatcher.tsx
jordienr: All good, updated the names, thanks for the review!

On deduplication, I think it makes sense to leave it as is, 1 exposure to the event 1 event fired even on refresh. Then you see stickiness of errors.

packages/common/telemetry-constants.ts
jordienr: I missed it ✅ thanks!

apps/studio/components/interfaces/ErrorHandling/ErrorMatcher.utils.ts
jordienr: checked!

apps/studio/data/error-patterns.test.ts
jordienr: done!

apps/studio/data/error-patterns.ts
jordienr: since we only handle 1 error for now i think we can leave it as is, if we need to refactor some of it in the near future it should be easy

packages/ui-patterns/src/ErrorDisplay/ErrorDisplay.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    return (
+      <Card
+        ref={ref}
+        className={cn('rounded-lg border border-default', className)}
+        role="alert"
+        aria-labelledby="error-display-title"
+        {...props}
+      >
+        <CardHeader className="flex-row items-center gap-2.5 space-y-0 py-2 px-3">
```
kemaldotearth: Vertical padding is too snug, I believe I had it `12px` / `p-3` in the design file. Gives it bit more room to breathe.

kemaldotearth (commented):
Is it possible when we do have space above a certain viewport we adhere to the alignment below?

Image

One thing I was wondering with the the placement in the sidebar, I think we could potentially do something similar to our feature preview enable popover. E.g. we have a smaller version of our error card, a button that says "Troubleshoot" (or similar) and we can open the error display as a dialog?

Image

jordienr:
> Is it possible when we do have space above a certain viewport we adhere to the alignment below?

@kemaldotearth 
we cant do this without tailwind container queries i'm afraid, since it's not the size of the viewport that matters but the parent container of the error. viewport could be 2xl but the error renders inside the sidebar (for now at least)

packages/ui-patterns/src/ErrorDisplay/ErrorDisplay.tsx
jordienr: fixed

kemaldotearth:
> > Is it possible when we do have space above a certain viewport we adhere to the alignment below?
> 
> @kemaldotearth we cant do this without tailwind container queries i'm afraid, since it's not the size of the viewport that matters but the parent container of the error. viewport could be 2xl but the error renders inside the sidebar (for now at least)

Ahh I seem to remember `@container` being used on the new layout stuff. Either way we can do as a follow up. What do you think to a :smol: version for sidebar usage? I can help run some quick mock ups in Figma if need be 🙏

jordienr:
> > > Is it possible when we do have space above a certain viewport we adhere to the alignment below?
> > 
> > 
> > @kemaldotearth we cant do this without tailwind container queries i'm afraid, since it's not the size of the viewport that matters but the parent container of the error. viewport could be 2xl but the error renders inside the sidebar (for now at least)
> 
> Ahh I seem to remember `@container` being used on the new layout stuff. Either way we can do as a follow up. What do you think to a :smol: version for sidebar usage? I can help run some quick mock ups in Figma if need be 🙏

@kemaldotearth yea 100% go for it, this is just v1 we can keep improving it and adding it to more places.

kemaldotearth (approved):
✅ to keep it moving!

seanoliver (approved):
Telemetry side looks great!

PR #43566, opened 2026-03-09
chore - allow cron jobs e2e tests to run in parallel and make them stable

djhi:
Follow up of #43547 and #43560

## Problem

The cron job tests run in serial mode and can be flaky.

## Solution

- Use simple query to install the `pg_cron` extension (way faster than using UI)
- Make sure all tests can run in parallel


awaseem (approved):
LGTM

PR #43569, opened 2026-03-09
chore - make database e2e tests run in parallel and more stable

djhi:
## Problem

- database e2e tests run in serial mode, which is slower
- they also are a bit flaky

## Solution

- Ensure they can run in parallel
- Make them more stable by using UI checks when waiting
- Use assertions that automatically wait/retry in playwright


awaseem (approved):
NICE

PR #43571, opened 2026-03-09
chore - make index advisor e2e tests faster

djhi:
## Problem

Index advisor tests are slow because they install the extension via UI

## Solution

Use an SQL query to enable the extension so that we can just test its features


awaseem (approved):
LGTM this is cleaner, enable via database call and then check ui if its on for elements

djhi:
> LGTM this is cleaner, enable via database call and then check ui if its on for elements

This one alone saves between 1 and 2 min 🎉

PR #43572, opened 2026-03-09
fix: return old data when rows are changed rather than empty data

awaseem:

YES

Pending review from the team, should fix this strange flashing of the table editor


alaister:
Opting to include in https://github.com/supabase/supabase/pull/43627 instead

PR #43580, opened 2026-03-09
feat(billing): tax id dismissible banner

ignaciodob:
### Changes
- Adds a dismissible banner prompting paid org users to add a Tax ID to their billing settings
- Only shown to users with billing read permissions on orgs with a paid plan and no Tax ID set
- Dismissal is persisted per-org via localStorage; banner also auto-hides when a Tax ID is added

### Testing
- Log in as a user with no billing permissions into a Free Plan Org with no tax id: no banner is shown
- Log in as a user with billing permissions into a Free Plan Org with no tax id: no banner is shown
- Log in as a user with billing permissions into a Paid Plan Org with no tax id
- Assert banner appears at the top
- Head to `/org/_/billing` with a paid Org. Change your billing country to a not support one (like Paraguay) and click save. Assert that the banner disappears. Change it back to continue testing.
- Click dismiss: banner disappears and stays dismissed across navigation/refresh
- Log in as a user with billing permissions into a different Paid Plan Org with no tax id
- Add a Tax ID via billing settings, assert that the banner disappears immediately

image

### DIFF
--- a/apps/studio/components/layouts/AppLayout/TaxIdBanner.tsx
+++ b/apps/studio/components/layouts/AppLayout/TaxIdBanner.tsx
@@ -21,7 +21,11 @@ export const TaxIdBanner = () => {
   )
 
   const shouldFetch =
-    !!slug && org?.plan?.id !== 'free' && isDismissLoaded && !isDismissed && !!org?.organization_missing_tax_id
+    !!slug &&
+    org?.plan?.id !== 'free' &&
+    isDismissLoaded &&
+    !isDismissed &&
+    !!org?.organization_missing_tax_id
 
   const { data: customerProfile } = useOrganizationCustomerProfileQuery(
     { slug },
--- a/apps/studio/components/layouts/AppLayout/TaxIdBanner.tsx
+++ b/apps/studio/components/layouts/AppLayout/TaxIdBanner.tsx
@@ -1,11 +1,14 @@
 import { LOCAL_STORAGE_KEYS } from 'common'
 import { useRouter } from 'next/router'
 
+import { TAX_IDS } from '@/components/interfaces/Organization/BillingSettings/BillingCustomerData/TaxID.constants'
 import { HeaderBanner } from '@/components/interfaces/Organization/HeaderBanner'
 import { InlineLink } from '@/components/ui/InlineLink'
 import { useLocalStorageQuery } from '@/hooks/misc/useLocalStorage'
 import { useSelectedOrganizationQuery } from '@/hooks/misc/useSelectedOrganization'
-import { useOrganizationTaxIdQuery } from 'data/organizations/organization-tax-id-query'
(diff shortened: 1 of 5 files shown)


apps/studio/components/layouts/AppLayout/TaxIdBanner.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const router = useRouter()
+  const { data: org } = useSelectedOrganizationQuery()
+  const slug = org?.slug
+
+  const [isDismissed, setIsDismissed, { isSuccess: isDismissLoaded }] = useLocalStorageQuery(
+    LOCAL_STORAGE_KEYS.TAX_ID_BANNER_DISMISSED(slug ?? ''),
+    false
+  )
+
+  const { data: taxId, isSuccess: isTaxIdLoaded } = useOrganizationTaxIdQuery(
```
kanishkdudeja: I think we should also check for WRITE permission on `stripe.customer` since the read permission may not be enough?

apps/studio/components/layouts/AppLayout/TaxIdBanner.tsx
```diff
@@ -0,0 +1,54 @@
...
+    { slug },
+    {
+      enabled: !!slug && org?.plan?.id !== 'free' && !isDismissed,
+      staleTime: 1000 * 60 * 30,
+    }
+  )
```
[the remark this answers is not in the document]
kanishkdudeja: +1

apps/studio/components/layouts/AppLayout/TaxIdBanner.tsx
ignaciodob: I think read permission is enough - the banner doesn't get shown to non-owner and non-admin users

charislam (changes requested):
Banner doesn't disappear immediately when Tax ID is added (requires a refresh), the problem is that the Tax ID update path doesn't call `invalidateOrganizationsQuery`, which is what the `org?.organization_missing_tax_id` reads from

ignaciodob:
Thanks for catching that @charislam ! 

I missed that when changing the approach to use the new `organization_missing_tax_id`. I ran into an issue, though. After saving the tax ID, the server responds `304 Not Modified` on the organizations endpoint for a short while, so invalidation/refetch returns stale data. Since updating the tax id involves an async call to Orb and a sync back to our databases there's a delay to update the `organization_missing_tax_id`

What I did now is optimistically update `organization_missing_tax_id `directly in the cached organizations list (using exact: true to avoid prefix-matching the tax ID query key, which would cause the form to refetch and lose its data).

I don't think this is ideal, but it's going to be short-lived until we get all the current orgs to put down their tax ID.

charislam (approved):
LGTM!

PR #43583, opened 2026-03-10
[SEC-385] feat(studio): Add default privileges for new entities toggle

alaister:
Adds a new toggle in:
Screenshot 2026-03-10 at 17 17 06

Other changes:
- form submissions with no table/function changes were failing because an empty string got passed to executeSql. Added an early return when there's nothing to execute

To test:
- Ensure the form is still in working order
- Create some tables and functions with the toggle on add off and make sure your selected default applies

### DIFF
--- a/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
+++ b/apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
@@ -1,6 +1,6 @@
 import { zodResolver } from '@hookform/resolvers/zod'
 import { PermissionAction } from '@supabase/shared-types/out/constants'
-import { useQueryClient } from '@tanstack/react-query'
+import { useQuery, useQueryClient } from '@tanstack/react-query'
 import { useParams } from 'common'
 import { Lock } from 'lucide-react'
 import Link from 'next/link'
@@ -19,6 +19,7 @@ import {
   Input_Shadcn_,
   PrePostTab,
   Skeleton,
+  Switch,
   useWatch_Shadcn_,
 } from 'ui'
 import { GenericSkeletonLoader, PageSection, PageSectionContent } from 'ui-patterns'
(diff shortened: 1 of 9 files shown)


apps/studio/data/privileges/update-default-privileges-mutation.ts
```diff
@@ -0,0 +1,48 @@
...
+  onSuccess,
+  onError,
+  ...options
+}: Omit<
+  UseCustomMutationOptions<UpdateDefaultPrivilegesData, Error, UpdateDefaultPrivilegesVariables>,
+  'mutationFn'
+> = {}) => {
+  return useMutation<UpdateDefaultPrivilegesData, Error, UpdateDefaultPrivilegesVariables>({
+    mutationFn: (vars: UpdateDefaultPrivilegesVariables) => updateDefaultPrivileges(vars),
+    onError(error: Error) {
```
ivasilov, line 41: Should invalidate the `defaultPrivileges`, right?

avallete (approved):

apps/studio/data/privileges/update-default-privileges-mutation.ts
awaseem: On the error handler though? 

```
        queryClient.invalidateQueries({
          queryKey: privilegeKeys.defaultPrivileges(projectRef),
        }),
```

it does to it for the mutation function, but why on error?

apps/studio/components/interfaces/Settings/API/PostgrestConfig.tsx
```diff
@@ -376,6 +414,28 @@ export const PostgrestConfig = () => {
                         />
                       </FormItemLayout>
 
+                      {watchedDbSchema.includes('public') && (
+                        <FormItemLayout
+                          isReactForm={false}
+                          layout="flex-row-reverse"
+                          label="Default privileges for new entities"
+                          description="When enabled, new tables and functions in the public schema are automatically accessible via the Data API. We recommend disabling this and manually granting access to each new entity."
```
ivasilov: The `public` part should be in `<code>` so it's clear it has special meaning.

apps/studio/data/privileges/update-default-privileges-mutation.ts
ivasilov: it shouldn't on error, I meant in `onSuccess`. I think you're right, it does it separately.

ivasilov (approved):

PR #43584, opened 2026-03-10
Add warning in SQL editor when running alter db commands which prevents connections to the DB

joshenlim:
## Context

Adds a warning if running an `ALTER DATABASE` command that prevents connections to the database. This would lock the dashboard out of the database, and re-configuring the setting will require a direct connection to the DB

image

## To test

- Could try running the following command in the SQL editor to check the warning, this will show all the warnings
```
ALTER TABLE colors2 drop column sss;
update colors set name = 'test';
alter database postgres connection limit 0;
```


djhi (approved):

djhi:
Tested on staging

PR #43606, opened 2026-03-10
fix(common): decode URL-encoded mw_diag cookie before parsing

seanoliver:
## Problem

The Phase 1 mw_diag monitoring dashboard wasn't showing any data despite the middleware PR (#43413) being live in production.

Next.js `response.cookies.set()` URL-encodes cookie values, so `hit=1&would_stamp=0&has_cookie=0` becomes `hit%3D1%26would_stamp%3D0%26has_cookie%3D0` in `document.cookie`. Our `parseMwDiagCookie` was passing that encoded string directly to `URLSearchParams`, which splits on literal `&` and `=` — the encoded `%26`/`%3D` delimiters were invisible to it. Result: `params.get('hit')` always returned `null`, so every diagnostic cookie was silently dropped.

## Changes

- Add `decodeURIComponent()` before `URLSearchParams` in `parseMwDiagCookie`
- Add 3 test cases covering URL-encoded cookie values (matching what Next.js actually produces in production)

## Testing

- All 34 unit tests pass (31 existing + 3 new)
- Verified locally with Playwright: manually set the URL-encoded cookie format on localhost:8082, confirmed `parseMwDiagCookie` now returns `{hit: true, ...}` and the data flows through to `PageTelemetry`

GROWTH-625

### DIFF
--- a/packages/common/first-referrer-cookie.test.ts
+++ b/packages/common/first-referrer-cookie.test.ts
@@ -232,6 +232,33 @@ describe('first-referrer-cookie', () => {
         has_existing_cookie: false,
       })
     })
+
+    it('parses URL-encoded value from Next.js response.cookies.set()', () => {
+      const header = `${MW_DIAG_COOKIE_NAME}=hit%3D1%26would_stamp%3D1%26has_cookie%3D0`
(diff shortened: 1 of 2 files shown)


awaseem (approved):
LGTM

PR #43616, opened 2026-03-10
docs: comment out Kong ports key in reverse proxy HTTPS guide

anix-lynch:
## Summary
- Comment out the ports key itself in the reverse proxy guide snippet
- Keep port mapping lines commented under it

## Why
The existing snippet comments only the list entries but leaves ports active, which can produce:
services.kong.ports must be a array
when users follow the guide literally.

## Reference
- Closes #43567


7ttp (commented):
hi @anix-lynch 💚
unfortunately this has already been fixed by 
- https://github.com/supabase/supabase/pull/43853

Thanks for ur efforts in making Supabase Better 😁

PR #43626, opened 2026-03-11
Joshen/fe 2642 project settings allow to select a replica to power

joshenlim:
## Context

Related to dashboard scalability
Previous PR: https://github.com/supabase/supabase/pull/42856

Note: Changes are all feature flagged still and I'm still not entirely convinced with the current UX
Will iterate as as go along, and only make this publicly available when we're satisfied with the behaviour

Adds a "Dashboard preference" section to the project settings
image

In which users can then select which database they'd like to use for read queries run from the dashboard
Note: Everything is local storage for now, but we'd need middleware support if we want to make this setting persist for all users on the project
image

Added a dialog as well to further explain what this implies
image

## To test

- Choosing a replica in dashboard preferences will only affect the table editor as thats the only place that's set up so far to use a replica for read queries (I'll need to follow up for other parts of the dashboard in subsequent PRs)


alaister (approved):
LGTM!

- Tested on preview

PR #43627, opened 2026-03-11
[FE-2738] fix(studio): use read replica identifier in table rows keys

alaister:
Previously `connectionString` was passed into the query key for `table-rows`. Since `connectionString` is unstable, it would cause the query key to change randomly, often making the user experience brief random "no rows found" states while data loaded for the new key.

This PR uses `readReplicaIdentifier` as a stable version of `connectionString`, maintaining the existing functionality while avoiding the unnecessary data reloads.

Note that there are still quite a few places where `connectionString` is passed into the query key. We'll fix these as follow up PRs.

I'm opting to also include the fix from https://github.com/supabase/supabase/pull/43572 in here for consistency.

**To test:**
- Ensure table editor functions normally
- Use the table editor with a read replica selected

### DIFF
--- a/apps/studio/data/prefetchers/project.$ref.editor.$id.tsx
+++ b/apps/studio/data/prefetchers/project.$ref.editor.$id.tsx
@@ -22,6 +22,7 @@ interface PrefetchEditorTablePageArgs {
   queryClient: QueryClient
   projectRef: string
   connectionString?: string | null
+  readReplicaIdentifier?: string
   id: number
   sorts?: Sort[]
   filters?: Filter[]
@@ -32,6 +33,7 @@ export function prefetchEditorTablePage({
   queryClient,
   projectRef,
   connectionString,
+  readReplicaIdentifier,
   id,
   sorts,
   filters,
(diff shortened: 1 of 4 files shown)


apps/studio/data/prefetchers/project.$ref.editor.$id.tsx
```diff
@@ -90,7 +94,7 @@ export function usePrefetchEditorTablePage() {
         // eat prefetching errors as they are not critical
       })
     },
-    [connectionString, project, queryClient, roleImpersonationState, router]
+    [connectionString, readReplicaIdentifier, project, queryClient, roleImpersonationState, router]
```
djhi, line 97: Isn't `connectionString` unstable? Can't we wrap the result of `useConnectionStringForReadOps` in a `useMemo`?

alaister: It's only a little bit unstable – most re-renders it'll be the same, it just refreshes every few minutes. So it's more unstable long term rather than being unstable in the regular coding sense haha

awaseem (approved):

PR #43629, opened 2026-03-11
feat(studio): connectSection experiment — replace Getting Started with Connect section

pamelachia:
## Summary

Re-ports PR #43119 against the current `ProjectHome` codebase (the original PR targeted `HomeNew/` which was removed during the `homeNew` graduation in #43437). Also unifies the `connectSheet` and `connectSection` feature flags into a single `connectSection` flag — the `connectSheet` flag is removed entirely, so both the ConnectSheet (header) and ConnectSection (homepage row) are controlled by one experiment.

Screenshot 2026-03-11 at 9 59 47 PM
Screenshot 2026-03-11 at 9 59 23 PM

- Adds `connectSection` PostHog experiment flag that controls two things:
  1. Swaps the Getting Started section for a Connect section on the project homepage for new projects (< 10 days old)
  2. Swaps the legacy Connect dialog for the new ConnectSheet panel in the header
- **Control** (`getting-started`): existing Getting Started section + legacy Connect dialog in header
- **Treatment** (`connect`): new 4-tile Connect section + ConnectSheet in header
- `undefined` (loading): neither section renders, avoiding flash
- Tiles filtered by the same `useIsFeatureEnabled` flags as ConnectSheet (`show_app_frameworks`, `show_mobile_frameworks`, `show_orms`)
- Connect tile clicks tracked via `home_connect_action_clicked`; section render tracked via `home_connect_section_exposed`; sheet opens tracked via `connect_sheet_opened` with source attribution (`header_button` or `connect_section`)

## Changes

- `packages/common/telemetry-constants.ts` — new `home_connect_section_exposed`, `home_connect_action_clicked`, and `connect_sheet_opened` event types
- `ConnectSheet/ConnectSheet.tsx` — read `connectTab` query param and sync to active mode on open; `handleModeChange` keeps param in sync on tab switch; fire `connect_sheet_opened` event with source attribution on open
- `ConnectButton/ConnectButton.tsx` — set `connectSource=header_button` query param on click
- `ProjectHome/ConnectSection.tsx` — new component (4-tile connect card); set `connectSource=connect_section` on tile click
- `ProjectHome/Home.tsx` — experiment flag wiring
- `LayoutHeader/LayoutHeader.tsx` — read `connectSection` flag instead of `connectSheet` to toggle ConnectSheet vs legacy Connect dialog

## Test plan

To test on the Vercel preview, set `connectSection=connect` to 100% in PostHog (or override via cookie `ph_override_connectSection=connect`). No separate `connectSheet` flag is needed — `connectSection` controls both features.
- [x] Treatment (`connect` variant) — "Get connected" section renders on new project, Getting Started hidden
- [x] Control (`getting-started`) — Getting Started renders, Connect section hidden
- [x] Mature project (> 10 days) — neither section regardless of flag
- [x] Clicking each tile opens ConnectSheet on the correct tab (Framework / Direct / ORM / MCP)
- [x] Switching tabs inside sheet updates `connectTab` URL param
- [x] Closing sheet clears `connectTab` param
- [x] Direct URL deep-link (`?showConnect=true&connectTab=orm`) opens sheet on correct tab
- [x] Dark mode — background gradient renders correctly
- [x] Light mode — background switches to light gradient
- [x] Responsive layout — 4 cols (xl), 2×2 (md), stacked (mobile)
- [x] Telemetry — `home_connect_section_exposed` fires once on load; `home_connect_action_clicked` fires with correct `mode`
- [x] Treatment — header Connect button opens ConnectSheet (not legacy Connect dialog)
- [x] Control — header Connect button opens legacy Connect dialog (not ConnectSheet)
- [x] Telemetry — clicking a ConnectSection tile fires `connect_sheet_opened` with `source: 'connect_section'`
- [x] Telemetry — clicking header Connect button fires `connect_sheet_opened` with `source: 'header_button'`


apps/studio/components/interfaces/ConnectSheet/ConnectSheet.tsx
```diff
@@ -3,34 +3,45 @@ import { useParams } from 'common'
...
+  const track = useTrack()
+  const prevShowConnect = useRef(false)
+
+  useEffect(() => {
+    if (showConnect && !prevShowConnect.current) {
+      track('connect_sheet_opened', { source: connectSheetSource })
+      setConnectSheetSource('header_button')
+    }
+    prevShowConnect.current = showConnect
+  }, [showConnect, connectSheetSource, track, setConnectSheetSource])
```
awaseem: Why do we need these use effects? this seems like its just setting values after handleModeChange is fired. We can just do this in the callback rather than using a useeffect to listen to changes no?

apps/studio/components/interfaces/ConnectSheet/ConnectSheet.tsx
```diff
@@ -61,23 +72,22 @@ export const ConnectSheet = () => {
...
   const availableModes = useMemo(
     () => schema.modes.filter((m) => availableModeIds.includes(m.id)),
     [schema.modes, availableModeIds]
   )
 
+  useEffect(() => {
+    if (!showConnect || !connectTab || !isConnectMode(connectTab)) return
+    if (!availableModeIds.includes(connectTab) || state.mode === connectTab) return
+    setMode(connectTab)
+  }, [showConnect, connectTab, availableModeIds, state.mode, setMode])
```
awaseem: Same comment here, I don' think this use effect is needed. It might be doing to much work here

apps/studio/components/interfaces/ConnectSheet/ConnectSheet.tsx
```diff
@@ -61,23 +72,22 @@ export const ConnectSheet = () => {
...
+
+  const handleModeChange = (mode: ConnectMode) => {
```
awaseem, line 93: All those use effect logic can be added here

apps/studio/components/interfaces/ConnectSheet/useAvailableConnectModes.ts
```diff
@@ -0,0 +1,24 @@
...
+    'project_connection:show_orms',
+  ])
+
+  return useMemo(() => {
+    const modes: ConnectMode[] = []
+    if (showAppFrameworks || showMobileFrameworks) modes.push('framework')
+    modes.push('direct')
+    if (showOrms) modes.push('orm')
+    modes.push('mcp')
+    return modes
```
awaseem: nit: new lines please to separate out if statements

apps/studio/components/interfaces/ProjectHome/ConnectSection.tsx
```diff
@@ -0,0 +1,132 @@
...
+    subheading: 'Third-party library',
+    icon: <Cable size={16} strokeWidth={1.5} />,
+  },
+  {
+    mode: 'mcp',
+    heading: 'MCP',
+    subheading: 'Connect your agent',
+    icon: <Sparkles size={16} strokeWidth={1.5} />,
+  },
+]
```
awaseem: nit: I would move these out to a separate file rather than in the component `ConnectSection.config.tsx`

apps/studio/components/interfaces/ProjectHome/ConnectSection.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+interface ConnectSectionProps {
+  variant: string
```
awaseem: Techincally there are 2 values for the variant right? can we make this a union

apps/studio/components/interfaces/ProjectHome/Home.tsx
```diff
@@ -73,6 +80,39 @@ export const ProjectHome = () => {
...
+
+  const shouldShowConnectSection =
+    isConnectSectionFlagResolved && isConnectSectionEnabled && !isMatureProject && !!project
+
+  const shouldShowGettingStarted =
+    isConnectSectionFlagResolved &&
+    !isConnectSectionEnabled &&
+    !isMatureProject &&
+    !!project &&
+    gettingStartedState !== 'hidden'
```
awaseem: this ternary is really hard to read, it should be split out into a function with if/else or if return statements

apps/studio/components/interfaces/ProjectHome/Home.tsx
```diff
@@ -73,6 +80,39 @@ export const ProjectHome = () => {
...
+        )
+        if (!nextKnownId) {
+          merged.push(id)
+          return
+        }
+        merged.splice(merged.indexOf(nextKnownId), 0, id)
+      })
+      return merged
+    })
+  }, [setSectionOrder])
```
awaseem: Also really hard to read and should be split out and unit tested

awaseem (changes requested):
Needs some clean up, reach out if you have any questions

apps/studio/components/interfaces/ProjectHome/Home.tsx
pamelachia: i also agree it is hard to read. 😂 i didn't know how to make it nicer but i will ask claude

apps/studio/components/layouts/ProjectLayout/LayoutHeader/LayoutHeader.tsx
```diff
@@ -75,9 +75,9 @@ export const LayoutHeader = ({
   const { setMobileMenuOpen } = useAppStateSnapshot()
   const gitlessBranching = useIsBranching2Enabled()
 
-  const connectSheetFlag = usePHFlag<string | boolean>('connectSheet')
-  const isFlagResolved = connectSheetFlag !== undefined
-  const isConnectSheetEnabled = connectSheetFlag === true || connectSheetFlag === 'variation'
+  const connectSectionVariant = usePHFlag<string | false>('connectSection')
```
seanoliver: nit: In `Home.tsx` this flag is typed as `usePHFlag<ConnectSectionVariant | false>`, but here it's `usePHFlag<string | false>`. Using the same `ConnectSectionVariant` type here would keep the two call sites consistent and make the `=== 'connect'` check self-documenting.

apps/studio/components/interfaces/ConnectSheet/ConnectSheet.tsx
```diff
@@ -3,34 +3,62 @@ import { useParams } from 'common'
...
+import type { ConnectMode, ProjectKeys } from './Connect.types'
 import { ConnectConfigSection, ModeSelector } from './ConnectConfigSection'
 import { ConnectStepsSection } from './ConnectStepsSection'
 import { useConnectState } from './useConnectState'
+import { useAvailableConnectModes } from './useAvailableConnectModes'
 import { useProjectApiUrl } from '@/data/config/project-endpoint-query'
-import { useIsFeatureEnabled } from '@/hooks/misc/useIsFeatureEnabled'
+import { useAppStateSnapshot } from '@/state/app-state'
+import { useTrack } from 'lib/telemetry/track'
+
```
seanoliver, line 18: nit: `CONNECT_MODES` here duplicates the set of values that `ConnectMode` already defines as a type. If `ConnectMode` were derived from a const array in `Connect.types.ts` (e.g. `export const CONNECT_MODES = ['framework', 'direct', 'orm', 'mcp'] as const` `export type ConnectMode = typeof CONNECT_MODES[number]`), both this guard and `useAvailableConnectModes` could import from one source of truth.

packages/common/telemetry-constants.ts
```diff
@@ -1900,6 +1900,60 @@ export interface HomeNewExperimentExposedEvent {
...
 }
 
+/**
+ * Connect section was shown to the user as part of the connectSection experiment.
+ *
+ * @group Events
+ * @source studio
+ * @page /project/{ref}
+ */
+export interface HomeConnectSectionExposedEvent {
```
seanoliver, line 1950: Consider: The `variant` property here is typed as `string` but the actual values are constrained to `ConnectSectionVariant` (`'connect' | 'getting-started'`). Using the union type would be more precise and consistent with how `mode` is typed on `HomeConnectActionClickedEvent` below (which does use a union).

seanoliver (approved):
Just a few nits to add. Otherwise LGTM! Approving to unblock!

Excited to see how this performs!

myerekapan:
Looks good as well! Just one idea, should we default to Claude Code here? Feel free to ignore.
Arc 2026-03-11 20 47 16

pamelachia:
> Looks good as well! Just one idea, should we default to Claude Code here? Feel free to ignore. Arc 2026-03-11 20 47 16

Great idea @myerekapan i'll add this

packages/common/telemetry-constants.ts
pamelachia: narrowed to 'connect' | 'getting-started' instead of string.

apps/studio/components/interfaces/ConnectSheet/ConnectSheet.tsx
pamelachia: Good call. Moved the const array into Connect.types.ts and derived the type from it. Both import from that single source

apps/studio/components/layouts/ProjectLayout/LayoutHeader/LayoutHeader.tsx
pamelachia: fixed

apps/studio/components/interfaces/ConnectSheet/ConnectSheet.tsx
pamelachia: I cleaned up the multiple useEffects into one in a follow-up commit. Is the current version better? 

I talked to claude and I am not sure how to move it into handleModeChange since they trigger on different things. handleModeChange fires when the user picks a mode inside the sheet, not when the sheet opens from an external source (like the ConnectSection tiles or header button). Since the sheet open state comes from a URL param, there's no callback I can hook into for "sheet just opened." help 🙏

apps/studio/components/interfaces/ConnectSheet/useAvailableConnectModes.ts
pamelachia: using some filter/map elegance instead 👀

apps/studio/components/interfaces/ProjectHome/ConnectSection.tsx
pamelachia: Moved to ConnectSection.config.tsx

apps/studio/components/interfaces/ProjectHome/Home.tsx
pamelachia: claude told me we can extract the logic into a getSectionVisibility() function in Home.utils.ts and the render uses if/return statements. Best I can do so open to different better ways!

apps/studio/components/interfaces/ProjectHome/Home.tsx
pamelachia: extracted it to the same file as above and added tests.

apps/studio/components/interfaces/ConnectSheet/ConnectSheet.tsx
awaseem: HmmmI think I might have misunderstood this. The url param got me

awaseem (approved):
LGTM

PR #43664, opened 2026-03-11
fix(Docs): Use internal Tooltip, prevent hydration issues on homepage

jeremenichelli:

This is a bug fix on the Docs homepage. Current in production we can see a flicker that shows black bars under `IconPanel` components.

Screenshot 2026-03-06 at 15 30 24

In development, this actually stays since we rely on server rendering instead of static generation, this stays until the dev refreshes and the cache prevents this broken look.

## What is the new behavior?

The problem is within the `react-tooltip`, a package we only use in here, that seems to not support RSC situations in modern-ish React.

The solution: Use the `Tooltip` family of components we already have in our `'ui'` package. This eliminates the issue plus shows a _tooltip_ that it's actually aligned with our current visuals. The former tooltip in this icon component was quite broken.

_Current tooltip (if you wanna call that a tooltip)_
Screenshot 2026-03-11 at 17 17 14

_Tooltip after this PR_
Screenshot 2026-03-11 at 17 17 08

### DIFF
--- a/apps/docs/components/HomePageCover.tsx
+++ b/apps/docs/components/HomePageCover.tsx
@@ -141,11 +141,7 @@ const HomePageCover = (props) => {
 
                 return (
                   <Link key={i} href={framework.href} passHref className="no-underline">
-                    <IconPanel
-                      iconSize={iconSize}
-                      tooltip={framework.tooltip}
-                      icon={iconToUse}
-                    />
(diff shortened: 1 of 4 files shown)


kemaldotearth (approved):

ChrisChinchilla (approved):
That black line has annoyed me for sooo long.

PR #43680, opened 2026-03-12
feat(design-system): hit area on table actions

dnywh:

Feature

- Platform webhooks have a ‘tap target on tap target’ pattern with `DropdownMenu` being on top of a clickable `TableRow`
- Misclicks are common (at least for me)
- @jordienr recently added the `hit-area` Tailwind plugin that helps with misclicks
	- https://github.com/supabase/supabase/pull/43636

## What is the new behavior?

- The aforementioned `DropdownMenu` in platform webhooks now uses `hit-area` to avoid misclicks
- Design system documentation
	- Extracted `hit-area` Tailwind plugin to shared packages so it can be used in design-system as well as studio

## Preview

https://github.com/user-attachments/assets/89f9110a-6c99-4eed-a386-a6f646b1f4f6


apps/design-system/registry/default/example/table-actions.tsx
```diff
@@ -41,16 +54,29 @@ export default function TableActions() {
...
+                    <DropdownMenuItem className="gap-x-2">
+                      <Pencil size={14} />
+                      <span>Edit user</span>
+                    </DropdownMenuItem>
+                    <DropdownMenuItem className="gap-x-2">
+                      <Trash2 size={14} />
+                      <span>Delete</span>
+                    </DropdownMenuItem>
+                  </DropdownMenuContent>
+                </DropdownMenu>
```
dnywh, line 79: Adding an actual `DropdownMenu` to this demo component (and others below) so the `hit-area` can be shown more clearly in action.

apps/studio/tailwind.config.js
```diff
@@ -8,7 +8,7 @@ module.exports = config({
     './../../packages/ui/src/**/*.{tsx,ts,js}',
     './../../packages/ui-patterns/src/**/*.{tsx,ts,js}',
   ],
-  plugins: [require('@tailwindcss/container-queries'), require('./tailwind-plugins/hit-area')],
+  plugins: [require('@tailwindcss/container-queries')],
```
dnywh, line 11: Moved to common packages.

jordienr (approved):

PR #43683, opened 2026-03-12
chore(studio): nits

dnywh:

UI nit changes

- We use big dots `•` to separate semantically different sets of text that are in the same row
  - This looks a bit odd given how large the dot is to text
- We put _Project Settings_ in its own sidebar section but without a separator like all other sidebar sections

## What is the new behavior?

- Use the smaller dots `·` to separate text
- Place a separator above _Project Settings_ in sidebar

| Before | After |
| --- | --- |
| AWS Healthy  Toolshed  Supabase-70BD1825-8077-4D1B-84C1-831A4ADAFC0B | AWS Healthy  Toolshed  Supabase |


joshenlim (approved):
LGTM! Verified on preview

PR #43692, opened 2026-03-12
Chore/shift manual queries into pg meta

joshenlim:
## Context

Related to FE-2557

Part of shifting manually written dashboard queries into packages/pg-meta where
- pg-meta can be code owners of
- we can write tests for the queries 

This PR just shifts all the `.sql.ts` files that we previously created into packages/pg-meta

There's still other areas where we need to shift over as well which I'll address in subsequent PRs

## Notable changes

- `getTableRowsCountSql` -> Opted to shift `formatFilterValue` logic out before calling this method (ref `table-rows-count-query`)
- `getDeleteOldCronJobRunDetailsByCtidSql` -> Opted to shift `validatePageNumber` logic out before calling this method (ref `CronJobsTab.useCleanupActions`)


soedirgo (approved):

djhi (approved):
💯

PR #43693, opened 2026-03-12
feat: Add table and column menus in schema visualiser

djhi:
## Problem

- The schema visualiser lacks editing capabilities which leads to a lot of navigation (ediing tables, columns)
- ReactFlow prevents users from selecting table and column names (to copy them). Diasbling drag and pan on those texts would make moving items cumbersome
- Long table and column names are hidden and even hide other elements

## Solution

- Add menus for both tables and columns
- Truncate long names with ellipsis and add a tooltip
- Hide menus when exporting to png/svg

[Screen Recording 2026-03-12 at 10.10.08.webm](https://github.com/user-attachments/assets/b2780266-e874-41d1-ac82-7c2c4ba5abf2)


awaseem (approved):
Tested on staging and self hosted! Nice tests

PR #43697, opened 2026-03-12
feat(studio): add pagination to query performance page FE-2774

jordienr:
## Problem

The Query Performance page loaded all results in a single query with a fixed limit of 20 rows, giving users no way to browse beyond the first page. There was also no way to control how many rows were shown at once.

## Fix

adds pagination 

## How to test

- Navigate to `/observability/query-performance` in Studio
- scroll to bottom
- should automatically load more results


kemaldotearth (approved):
Nice! 🚢  

The only thing I wondered was if we want a second indicator at the bottom of the list, but it's not necessary I guess with the border animation to indicate load!

PR #43704, opened 2026-03-12
[FE-2479] chore(studio): revoke default grants in public on project creation

alaister:
When the `tableEditorApiAccessToggle` feature flag is enabled, project creation now appends SQL to revoke default privileges for `anon`, `authenticated`, and `service_role` on the `public` schema. This runs after the base image init script's default grants. This is temporary while we're still using a feature flag. Eventually it'll be moved into the base image.

Applies to both the main project creation flow and the Vercel deploy button flow.

Part of the "Secure by Default" initiative – new projects created under this flag won't automatically expose tables/functions/sequences to the Data API via default privileges. Users can still opt in at a table level.

## Notes

Reusing the existing `useDataApiGrantTogglesEnabled()` flag here rather than creating a new one – it's the same feature surface area and avoids unnecessary flag proliferation.

## To test

1. **With flag enabled:**
   - Enable the `tableEditorApiAccessToggle` flag in PostHog for your user
   - Create a new project via the dashboard
   - Create a new table
   - Confirm in `/project/_/integrations/data_api/settings` that the new table is not exposed by default

2. **With flag disabled:**
   - Disable the flag (or use a different user without it)
   - Create a new project
   - Verify default privileges are intact and tables are accessible via the Data API as usual

3. **With RLS event trigger enabled too:**
   - Enable both the feature flag and the "enable RLS event trigger" checkbox during project creation
   - Verify both SQL statements run correctly on the new project


apps/studio/pages/new/[slug].tsx
```diff
@@ -34,25 +34,26 @@ import { useAuthorizedAppsQuery } from 'data/oauth/authorized-apps-query'
 import { useFreeProjectLimitCheckQuery } from 'data/organizations/free-project-limit-check-query'
 import { useOrganizationAvailableRegionsQuery } from 'data/organizations/organization-available-regions-query'
 import { useOrganizationsQuery } from 'data/organizations/organizations-query'
+import { buildDefaultPrivilegesSql } from 'data/privileges/privileges.sql'
```
Comment (joshenlim) [inline apps/studio/pages/new/[slug].tsx]: this import should now point to pg-meta as the queries are shifted over

joshenlim (approved):
Approving to unblock - but a few separate issues noticed

1. Auth Policies is incorrectly showing that the table is exposed via the Data API even if the table is not
  - The check [here](https://github.com/supabase/supabase/blob/chore/fe-2479/apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/index.tsx#L64) is incorrect as its only checking exposed schemas, not tables

2. Data API under integrations show exposed tables are none even if I've exposed one
Image

imor (approved):

PR #43706, opened 2026-03-12
fix: updated project level settings with incorrect fetching

awaseem:

YES

WIP


joshenlim (approved):
Fix looks correct, can verify that [API fix](https://github.com/supabase/infrastructure/pull/25634) is already through a while back which makes the FE patch irrelevant

It's quite hard to reproduce this issue exactly though as it would require an org to have more than 96 projects (to get past the pagination limit), but I did do a simple test on the preview to verify that adding / remove project scoped roles still works as expected

PR #43713, opened 2026-03-12
fix: studio docker build fails because of package patches

djhi:
## Problem

The docker build for studio fails because pnpm patches are not included

## Solution

- bump pnpm version in Dockefile to match the one used by the repo
- bump turbo version in Dockerfile to use one that handles pnpm patches
- add a script to test the build locally
- add a github action to validate the build on each PR


awaseem (approved):
LGTM

PR #43730, opened 2026-03-13
chore(design-system): centralise MCP client assets

dnywh:

Chore

- We use the MCP client connect dialog in two places: Studio and Docs
- We duplicate image assets for each client in each of those two places

## What is the new behavior?

- Centralised assets
- Consolidation and simplification as a result of it all being a single source-of-truth now

## To test

- [ ] Everything works as it did before across both Studio and [Docs](https://supabase.com/docs/guides/getting-started/mcp)
- [ ] All MCP client images load as expected


mattrossman (approved):

joshenlim (approved):
Verified on both staging and docs previews

kemaldotearth (approved):

PR #43733, opened 2026-03-13
fix(studio): fall back to legacy connect dialog when connectSection flag is unresolved

pamelachia:
## Summary

When the `connectSection` PostHog flag is unresolved (`undefined`) — e.g. during loading or if PostHog is unreachable — the header Connect button renders nothing (`null`), making it a no-op. This changes the fallback to render the legacy `<Connect />` dialog instead, so the button always works.

## Changes

- Remove `isFlagResolved` gate in `LayoutHeader` — render `<Connect />` (legacy) whenever `connectSection !== 'connect'`, including when `undefined`

## Testing

Tested on Vercel preview:
- [x] Flag returns `'connect'` — ConnectSheet opens
- [x] Flag returns `'getting-started'` or `false` — legacy Connect dialog opens
- [x] Flag is `undefined` (unresolved) — legacy Connect dialog opens instead of nothing


alaister (approved):
LGTM!

- Tested on preview

PR #43734, opened 2026-03-13
Change URL param string for destination panel type as it conflicts with Connect Modal

joshenlim:
## Context

Small one - while on the database replication page, if you open the Connect Dialog and change the "type" field, the destination panel opens up. Happening cause the string for `useQueryState` conflicts here - both are using `type`

Fix is just to update the string for `useQueryState` for the DestinationPanel component, changed it to `destinationType`


alaister (approved):
LGTM!

- Tested on preview

PR #43736, opened 2026-03-13
Disable incident banner query in prod, only status page query

joshenlim:
## Context

Realised that we're calling both /incident-status and /incident-banner on prod
Which the latter is only meant for local/staging for now but its creating a lot of 500 requests

Changes here patches that


joshenlim:
closing in favor of https://github.com/supabase/supabase/pull/43783/changes

PR #43740, opened 2026-03-13
Add source filter in query performance page

joshenlim:
## Context

Adds a source filter for the query performance advisor so you can filter out queries from the dashboard, or not from the dashboard

image

For transparency how this works:
- Queries fired via the dashboard through the /query endpoint get enriched with metadata from the API to include a comment like `-- source: dashboard`
  - That's mainly how this filter works atm, to check if this comment exists if the source "Dashboard" is selected, and the inverse if the source "Non dashboard" is selected


apps/studio/components/interfaces/Reports/Reports.queries.ts
awaseem: This isn't related, but this is super hard to parse. I think if you can follow up with a PR to separate the logic and write some unit tests around this if we already don't have any

awaseem (approved):
Tested on staging!

apps/studio/components/interfaces/Reports/Reports.queries.ts
joshenlim: i'll get this fixed up in this PR actually

monicakh:
#44201

PR #43742, opened 2026-03-13
refactor(studio): clean up ProjectLayout title API

dnywh:

Refactor, cleanup, and docs update.

After the page-title rollout, `ProjectLayout` is still in a transitional state:

- it accepts a deprecated `title` prop
- it still supports a separate `browserTitle.surface`
- wrapper layouts are split between passing `title` directly and passing `browserTitle.section`

That makes the API harder to reason about than it needs to be, even though the rendered titles are already correct.

## What is the new behavior?

This cleanup finishes the API simplification that came out of the stacked PR review:

- wrapper layouts stay `title`-first for DX
- `ProjectLayout` no longer accepts `title`
- `product` is now the single source of truth for the project-surface title segment
- `browserTitle` is now only used for extra browser-title metadata (`entity`, `section`, `override`)
- the remaining project-scoped callers now pass `browserTitle.section` when they need a section label
- docs now reflect the final pattern instead of the transitional one

Rendered page titles stay the same.

## Additional context

Checks run:

- `pnpm --filter studio exec vitest --run lib/page-title.test.ts components/layouts/ProjectLayout/index.test.tsx`
- `pnpm --filter studio typecheck`
- `pnpm exec prettier --check ...` on touched files

This is intended as the post-rollout cleanup PR based on Joshen's review feedback across the stacked title changes.


apps/studio/components/layouts/BranchLayout/BranchLayout.tsx
```diff
@@ -27,8 +27,8 @@ const BranchProductMenu = () => {
 const BranchLayout = ({ children }: PropsWithChildren<{}>) => {
   return (
     <ProjectLayout
-      title="Branching"
       product="Branching"
+      browserTitle={{ section: 'Branching' }}
```
joshenlim: guessing this isn't required technically? since its a project-level surface segment and `product` can handle

apps/studio/components/layouts/EdgeFunctionsLayout/EdgeFunctionDetailsLayout.tsx
```diff
@@ -220,9 +220,8 @@ const EdgeFunctionDetailsLayout = ({
   if (!isLoading && !canReadFunctions) {
     return (
       <ProjectLayout
-        title={title || 'Edge Functions'}
         product="Edge Functions"
-        browserTitle={browserTitle}
+        browserTitle={{ ...browserTitle, section: title || 'Edge Functions' }}
```
joshenlim: this seems unnecessary - title cannot be undefined + is already being declared in browserTitle above as section

apps/studio/components/layouts/ReportsLayout/ReportsLayout.tsx
```diff
@@ -18,8 +18,8 @@ const ReportsLayout = ({ title, children }: PropsWithChildren<ReportsLayoutProps
   if (reportsAll) {
     return (
       <ProjectLayout
-        title={title}
         product="Reports"
+        browserTitle={title ? { section: title } : undefined}
```
joshenlim: nit but bit of a consistency thing

apps/studio/components/layouts/StorageLayout/StorageLayout.tsx
```diff
@@ -11,7 +11,11 @@ export interface StorageLayoutProps {
 
 const StorageLayout = ({ title, children }: StorageLayoutProps) => {
   return (
-    <ProjectLayout title={title || 'Storage'} product="Storage" productMenu={<StorageMenuV2 />}>
+    <ProjectLayout
+      product="Storage"
+      browserTitle={{ section: title || 'Storage' }}
```
joshenlim: unnecessary - title cannot be undefined

joshenlim (approved):
All good now! ✅

PR #43744, opened 2026-03-13
fix(ui-patterns): Eliminate broken ease tailwind class

jeremenichelli:
In this PR:

Currently our Docs build produces this warning:

```
[0] warn - The class `ease-[.76,0,.23,1]` is ambiguous and matches multiple utilities.
[0] warn - If this is content and not a class, replace it with `ease-&lsqb;.76,0,.23,1&rsqb;` to silence this warning.
```

Tracking it down, it belongs to the `MultiSelect` component in our `ui-patterns` package. The duration of the modal containing the options is so fast (150ms) that custom easings do nothing in visual difference, so I replaced it for an in-out easing.


fsansalvadore (approved):
Looks good 👍

PR #43745, opened 2026-03-13
feat(Docs): Implement Accept header for agents markdown requests

jeremenichelli:

Takes `Accept: text/markdown` request header into account, and returns markdown version of the route.

Agents and users need to add `.md` to actually get the markdown version of a docs page.

## Additional context

**Read:** Useful for LLM agents - https://www.skeptrune.com/posts/use-the-accept-header-to-serve-markdown-instead-of-html-to-llms/

### DIFF
--- a/apps/docs/middleware.ts
+++ b/apps/docs/middleware.ts
@@ -11,9 +11,12 @@ const GUIDES_PATH = `${BASE_PATH ?? ''}/guides`
 export function middleware(request: NextRequest) {
   const url = new URL(request.url)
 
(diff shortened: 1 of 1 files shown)


ChrisChinchilla (approved):
Nice! This could help with metrics too maybe 🤔🤔

jeremenichelli:
> Nice! This could help with metrics too maybe 🤔🤔

In theory, checking the 'Accept" header could be a signal of agent fetching the content. But that filtering needs to be done at server analytics level.

PR #43746, opened 2026-03-13
feat(Docs): Track AI tools usage in Docs

jeremenichelli:

Adds event to AI tooling in docs.

We are unaware of usage of these tools.

## What is the new behavior?

We send telemetry events when people click on "Ask..." links or "Copy as Markdown" button.

### DIFF
--- a/apps/docs/components/GuidesSidebar.tsx
+++ b/apps/docs/components/GuidesSidebar.tsx
@@ -27,19 +27,20 @@ function AiTools({ className }: { className?: string }) {
 
   async function copyMarkdown() {
     const mdUrl = `/docs/${path}.md`
-
+    
     try {
       const res = await fetch(mdUrl)
       const text = await res.text()
       await navigator.clipboard.writeText(text)
       setCopied(true)
       setTimeout(() => setCopied(false), 2000)
-      sendTelemetryEvent({
-        action: 'copy_as_markdown_clicked',
-      })
     } catch (error) {
       console.error('Failed to copy markdown', error)
     }
+
+    sendTelemetryEvent({
+      action: 'copy_as_markdown_clicked',
+    })
   }
 
   return (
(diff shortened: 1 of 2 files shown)


packages/common/telemetry-constants.ts
```diff
@@ -747,6 +747,32 @@ export interface DocsFeedbackClickedEvent {
...
+ * User clicked "Ask..." to open a new window to consult an agent about the current page.
+ *
+ * @group Events
+ * @source docs
+ */
+export interface AskAIEvent {
+  action: 'ask_ai_clicked'
+  properties: {
+    /**
+     * 'yes' means clicking on the tick button, 'no' means clicking on the cross button.
```
myerekapan: I guess this is from a copy-paste? Better to fix it for future AI context

apps/docs/components/GuidesSidebar.tsx
```diff
@@ -32,6 +34,9 @@ function AiTools({ className }: { className?: string }) {
       await navigator.clipboard.writeText(text)
       setCopied(true)
       setTimeout(() => setCopied(false), 2000)
+      sendTelemetryEvent({
+        action: 'copy_as_markdown_clicked',
+      })
```
myerekapan: nitpick, but I think this fires on copy-success and not click. Up to you if you wanna be really accurate on click tracking.

myerekapan (approved):

apps/docs/components/GuidesSidebar.tsx
jeremenichelli: Hmm, true. It was def on purpose but you are making me re-think it.

packages/common/telemetry-constants.ts
jeremenichelli: Yeap. Thanks for catching it.

PR #43747, opened 2026-03-13
chore: Fix formatting in file

jeremenichelli:
@kemaldotearth I think you pushed this files. They weren't formatted correctly and keep spilling in format checks :)


kemaldotearth (approved):
:shipit:

PR #43760, opened 2026-03-13
fix(studio): distinguish NULL from "use default" when inserting rows

### LINKED ISSUE #43548: Dashboard row insert doesn't work properly due to dashboard insert logic not differentiate between "doing nothing" or inserting NULL
# Bug report

## Describe the bug
First bug is how I found out about this. Second one is the bug described in title, which is cause of all.

First.
Row inserting while impersonating fails for bool type columns when column has no privilege to insert for given role. Becuase supabase dashboard tries to insert bool without checking whether its column default.

Second.
Inserting NULL from dashboard doesnt work because dashboard thinks user wants to "do nothing" about column so column just defaults to postgres. But this is just wrong because psotgres accepts nulls.

## To Reproduce
For first bug.
Create a table with not null bool column (  default true ) and remove privilege of insert.
Impersonate a user and try to insert from dashboard.
While on insert screen bool column will show up as default value.
Try inserting. It will fail because privilege doesnt exist and it tried to insert a value.

If you make bool column nullable then you can select null from dropdown menu. And you will successfully insert with default value. Because dashboard didint even tried to insert null, it just did no insert for given column.

Second one is the root of first bug and its a big one.
Basically dashboard insertion screen doesn't differentiate between NULL and "doing nothing". You can check this with text fields. Setting to null just defaults to default column value.

## Expected behavior

Dashboard insert screen should differentiate between NULL and "doing nothing". Which would actually make inserting working properly and also fix privilege bug.

narek-keshishyan:
Bug fix

When inserting a new row via the Studio dashboard, setting a column to NULL 
results in the column being omitted from the INSERT statement, so PostgreSQL 
uses the column's default value instead. The dashboard doesn't distinguish 
between "user explicitly chose NULL" and "user didn't touch this field."

Fixes #43548

## What is the new behavior?
Added an `isUserModified` flag to `RowField` that tracks whether the user 
explicitly changed a field. When generating the INSERT payload:
- **Untouched fields** with null values are omitted → DB default applies
- **User-modified fields** with null values are kept → NULL is inserted

## Files changed
- `RowEditor.types.ts` — Added `isUserModified?: boolean` to RowField interface
- `RowEditor.utils.ts` — Updated `generateRowFields` and `generateRowObjectFromFields`
- `RowEditor.tsx` — Set `isUserModified: true` in `onUpdateField`
- `RowEditor.utils.test.ts` — Added 6 new tests (57 total, all passing)

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.tsx
@@ -95,7 +95,7 @@ export const RowEditor = ({
     const updatedProperties = Object.keys(changes)
     const updatedFields = rowFields.map((field) => {
       if (updatedProperties.includes(field.name)) {
-        return { ...field, value: changes[field.name] }
+        return { ...field, value: changes[field.name], isUserModified: true }
       } else {
         return field
       }
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.types.ts
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/RowEditor/RowEditor.types.ts
@@ -18,4 +18,5 @@ export interface RowField {
   isNullable: boolean
   isIdentity: boolean
   isPrimaryKey: boolean
+  isUserModified?: boolean
 }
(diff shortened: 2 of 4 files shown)


7ttp (commented):
hi @narek-keshishyan 😅

closing this in favour of : https://github.com/supabase/supabase/pull/44331

Thanks for ur efforts towards making Supabase better !!💚

PR #43763, opened 2026-03-13
fix(studio): accept TSV files in table editor import

mvanhorn:
## Summary

Fixes #41493

TSV files were rejected with "only CSV files are accepted" despite the UI text and constants correctly listing TSV as a supported format.

**Root cause:** The file validation used OR logic (`||`) requiring **both** MIME type and file extension to pass. Some browsers/OSes report empty or incorrect MIME types for `.tsv` files (e.g., `application/octet-stream` instead of `text/tab-separated-values`), causing rejection even when the extension was valid.

**Fix:** Changed to AND logic (`&&`) so files are accepted if **either** the MIME type or the extension matches. Also updated error messages to mention TSV alongside CSV.

## Test plan

- [ ] Upload a `.csv` file - should still work
- [ ] Upload a `.tsv` file - should now work (previously rejected on some machines)
- [ ] Upload a `.txt` file - should be rejected with "only CSV and TSV files" message
- [ ] Upload a file larger than 100MB - error message now says "CSV/TSV files"

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SpreadsheetImport/SpreadsheetImport.utils.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SpreadsheetImport/SpreadsheetImport.utils.tsx
@@ -1,13 +1,13 @@
 import dayjs from 'dayjs'
+import { DOCS_URL } from 'lib/constants'
+import { tryParseJson } from 'lib/helpers'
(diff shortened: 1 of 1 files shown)


7ttp (commented):
currently the team, wants to support CSV only
closing this up for now!...

> Thanks for ur efforts, towards contributing to Supabase !! 💚

PR #43765, opened 2026-03-13
fix(studio): prevent entities dropdown from disabling on empty search

mvanhorn:
## Summary

Fixes #41537

Two bugs in the Create Index side panel:

1. **Entities dropdown disabled on empty search:** The disabled check used `entityTypes.length === 0` which counted search-filtered results. When a search returned no matches, the dropdown became disabled and showed "No tables available in schema" even though tables existed.

2. **Shared search state between dropdowns:** The schema and entity dropdowns shared the same `searchTerm` state, so typing in one filtered the other.

**Changes:**
- Separated search state: schema dropdown uses Command's built-in filtering, entity dropdown has its own `entitySearchTerm` state
- Fixed disabled logic: only disable entity dropdown when there are no entities AND no active search term
- Reset entity search when schema changes
- Fixed schema dropdown placeholder from "Find table..." to "Find schema..."

## Test plan

- [ ] Open Create Index panel, select a schema with tables - entity dropdown should be enabled
- [ ] In entity dropdown, search for a non-existent table - dropdown stays enabled, shows "No tables found"
- [ ] Clear search - all tables reappear
- [ ] Type in schema dropdown - only schemas are filtered, entity dropdown is unaffected
- [ ] Select a genuinely empty schema - entity dropdown is correctly disabled

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/apps/studio/components/interfaces/Database/Indexes/CreateIndexSidePanel.tsx
+++ b/apps/studio/components/interfaces/Database/Indexes/CreateIndexSidePanel.tsx
@@ -1,8 +1,3 @@
-import { Check, ChevronsUpDown, Loader2 } from 'lucide-react'
-import Link from 'next/link'
-import { Fragment, useEffect, useMemo, useState } from 'react'
-import { toast } from 'sonner'
-
 import CodeEditor from 'components/ui/CodeEditor/CodeEditor'
 import { DocsButton } from 'components/ui/DocsButton'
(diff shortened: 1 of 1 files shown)


7ttp (commented):
hi @mvanhorn 
unfortunately closing in favour of https://github.com/supabase/supabase/pull/43906

Thanks for ur efforts in making supabase better 💚

PR #43777, opened 2026-03-14
chore: Add `credentials: omit` when calling into /incident API endpoints

ivasilov:
`/incident-status` and `/incident-banner` are not cached because they include auth cookies which busts Vercel cache. This PR add `credentials: omit` since they calls don't need auth cookies and they're same for all users.


joshenlim (approved):
Approving to unblock! 🙏

PR #43782, opened 2026-03-14
restore org selector loaders

fsansalvadore:
Fix org loader in the LayoutHeader.

## Before
Screenshot 2026-03-14 at 17 29 53

## After
Screenshot 2026-03-14 at 17 30 42


kemaldotearth (changes requested):
Skeleton not showing for me on mobile or desktop, on latest preview build:

https://github.com/user-attachments/assets/e956bd27-1fe9-480b-adb1-319416930175

Potentially remove the `md:` prefix on the width? Might be doing something weird with media queries.

kemaldotearth (approved):

PR #43798, opened 2026-03-16
Slight tuning of edge function page

SaxonF:
image

Switches from two to one column layout for details page and removes an inconsistent background colour in header. Moves critical detail information from page into header (url and deploy dates/counts)


alaister (approved):
LGTM!

- Tested on preview

One minor nit: The layout shift on full page reloads of the header is pretty aggressive

PR #43801, opened 2026-03-16
fix(storage): use correct prefix in fetchMoreFolderContents pagination

### LINKED ISSUE #43799: Bug: fetchMoreFolderContents uses wrong prefix in column view pagination
### Describe the bug

In the Storage Explorer column view, `fetchMoreFolderContents` computes the API prefix using all `openedFolders` instead of slicing to the target column index. This causes "load more" pagination in non-leaf columns to fetch objects from the wrong directory.

`fetchFolderContents` correctly slices:

```typescript
// storage-explorer.tsx L335-338
const prefix = state.openedFolders
  .slice(0, index + 1)
  .map((folder) => folder.name)
  .join('/')
```

`fetchMoreFolderContents` does not:

```typescript
// storage-explorer.tsx L398
const prefix = state.openedFolders
  .map((folder) => folder.name)
  .join('/')
```

Both functions accept an `index` parameter, but `fetchMoreFolderContents` ignores it for prefix computation.

### Severity

Users see files from the wrong directory when paginating in column view. If they act on those files (download, delete, move), the operation targets wrong objects.

### To reproduce

1. Open Storage Explorer in column view
2. Create or navigate into nested folders (e.g. A → B → C)
3. Ensure the root column (or folder A) has enough items to trigger pagination (100+ items, or reduce `LIMIT` constant for testing)
4. In that non-leaf column, scroll down to trigger "load more"
5. The additional items are fetched from the deepest path (A/B/C) instead of the column's actual path

### Expected behavior

`fetchMoreFolderContents` should use `.slice(0, index + 1)` like `fetchFolderContents` does, so pagination fetches objects from the correct directory.

### Suggested fix

```diff
- const prefix = state.openedFolders.map((folder) => folder.name).join('/')
+ const prefix = state.openedFolders
+   .slice(0, index + 1)
+   .map((folder) => folder.name)
+   .join('/')
```

### System information

- Reviewed on commit `77e5180c` (master)
- File: `apps/studio/state/storage-explorer.tsx`, line 398

### Additional context

This was missed during the refactor in #43541. The refactor correctly updated `fetchFolderContents` but missed the parallel function `fetchMoreFolderContents`.

karesansui-u:

Bug fix

In the Storage Explorer column view, `fetchMoreFolderContents` computes the API prefix using **all** `openedFolders` instead of slicing to the target column index. This causes "load more" pagination in non-leaf columns to fetch objects from the wrong directory.

For example, if folders A → B → C are open and the user scrolls down in the root column, the additional items are fetched from path `A/B/C` instead of the root.

Reported in #43799.

## What is the new behavior?

`fetchMoreFolderContents` now uses `.slice(0, index + 1)` to compute the prefix, matching the existing logic in `fetchFolderContents` (L335-338).

## Additional context

`fetchFolderContents` correctly slices:

```typescript
// L335-338
const prefix = state.openedFolders
  .slice(0, index + 1)
  .map((folder) => folder.name)
  .join('/')
```

`fetchMoreFolderContents` was missing the slice:

```typescript
// L398 (before fix)
const prefix = state.openedFolders
  .map((folder) => folder.name)
  .join('/')
```

This was missed during the refactor in #43541. The fix is a single `.slice(0, index + 1)` addition to align both functions.

Fixes #43799

### DIFF
--- a/apps/studio/state/storage-explorer.tsx
+++ b/apps/studio/state/storage-explorer.tsx
@@ -395,7 +395,10 @@ function createStorageExplorerState({
     }) => {
       state.setColumnIsLoadingMore(index)
 
-      const prefix = state.openedFolders.map((folder) => folder.name).join('/')
+      const prefix = state.openedFolders
+        .slice(0, index + 1)
+        .map((folder) => folder.name)
+        .join('/')
       const options = {
         limit: LIMIT,
         offset: column.items.length,


7ttp (commented):
hi @karesansui-u thanks for this!!
but unfortunately this has already been solved by pr: https://github.com/supabase/supabase/pull/44275

closing up!
Thanks again for ur effort towards making Supabase better 💚😅

PR #43811, opened 2026-03-16
feat(studio): move error code popover to shared data

kemaldotearth:

YES

As we now have the internal error code mapping in `shared-data`, thought it better to move away from the GraphQL endpoint and leave that for external use. The error code popovers in studio now rely on shared-data, meaning instant loads and no rate limiting.

### DIFF
--- a/apps/studio/components/interfaces/Settings/Logs/ErrorCodeTooltip.tsx
+++ b/apps/studio/components/interfaces/Settings/Logs/ErrorCodeTooltip.tsx
@@ -3,6 +3,8 @@ import Link from 'next/link'
 import { useState } from 'react'
 import { useTheme } from 'next-themes'
 import { ExternalLink } from 'lucide-react'
+import { ERROR_CODES, ERROR_CODE_DOCS_URLS, HTTP_ERROR_CODES } from 'shared-data'
+import type { ErrorCodeService } from 'shared-data'
 
 import { useErrorCodesQuery } from 'data/content-api/docs-error-codes-query'
 import { Service } from 'data/graphql/graphql'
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/Settings/Logs/ErrorCodeTooltip.tsx
```diff
@@ -38,16 +41,29 @@ export const ErrorCodeTooltip = ({ errorCode, service, children }: ErrorCodeTool
   const snap = useAiAssistantStateSnapshot()
   const { openSidebar } = useSidebarManagerSnapshot()
 
-  const { data, isPending } = useErrorCodesQuery({ code: errorCode, service }, { enabled: isOpen })
+  const mappedService = service ? SERVICE_MAP[service] : undefined
+  const sharedDataDef = mappedService
```
jordienr: i'd move all of this to a function so it can be tested

apps/studio/components/interfaces/Settings/Logs/ErrorCodeTooltip.tsx
```diff
@@ -38,16 +41,29 @@ export const ErrorCodeTooltip = ({ errorCode, service, children }: ErrorCodeTool
...
+    { enabled: isOpen && !sharedDataDef }
+  )
 
   const errors = data?.errors?.nodes?.filter((e) => !!e.message) ?? []
 
-  const docsUrl =
-    errors.map((e) => SERVICE_DOCS_URLS[e.service]).find(Boolean) ??
-    (service ? SERVICE_DOCS_URLS[service] : undefined)
+  const description = sharedDataDef?.description ?? errors[0]?.message
+  const docsUrl = mappedService
```
jordienr: same with this, if we move to a function and test the type cast in line 62 is not that problematic

jordienr (changes requested):
2 nits

jordienr (approved):

PR #43822, opened 2026-03-16
feat: improve readability of CPU chart O11Y-1290

jordienr:
## Problem
The CPU usage chart in the Observability dashboard could show values exceeding 100%, differing from the internal Grafana panel that support uses for debugging.

## Solution
Add a stackedPercent prop to the chart system that locks the Y-axis to [0, 100]. When enabled, the chart never visually overflows 100% while tooltip values remain completely untouched. We fill the rest of the unused with an "Idle" prop.

# Before
CleanShot 2026-03-18 at 16 04 26@2x

## After
CleanShot 2026-03-18 at 16 03 13@2x
CleanShot 2026-03-18 at 16 03 41@2x


awaseem (approved):
LGTM 🚀

PR #43554, opened 2026-03-09
add new opaque api keys and new auth to self-hosted

aantti:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

- Add support for opaque `sb_` API keys ("new API keys")
- Add support for asymmetric keys and asymmetric JWT user session tokens
- Keep it backward compatible with the legacy API keys (symmetric JWTs - `ANON_KEY` and `SERVICE_ROLE_KEY`), and symmetric JWT user session tokens
- Replicate the platform behavior for `sb_` keys substitution, per route

Other changes:

- Update Kong to 3.9.1 (latest OSS build from 2025)
- Add healthcheck for Kong
- Optimize dependencies (remove dependencies on Analytics/Logflare - not needed)
- Add `api-gw` alias for Kong service (in preparation for migration to Envoy)

To test (check all steps below carefully):

1. **Update/add:** `.env.example` (merge into `.env`), `docker-compose.yml`, `utils/add-new-auth-keys.sh`, `utils/rotate-new-api-keys.sh`, `volumes/api/kong-entrypoint.sh`, `volumes/api/kong.yml`
2. Run `utils/add-new-auth-keys.sh` and check `.env` for new env vars
3. **Uncomment** `GOTRUE_JWT_KEYS`, `API_JWT_JWKS`, and `JWT_JWKS` env vars in `docker-compose.yml` for Auth, Realtime, and Storage
4. Start containers via `docker compose up -d` and verify environment for Auth, PostgREST, Realtime, Storage, and Functions
5. (Consider using tests from PR #43573 for additional checks/debugging)

Closes #41843 (hopefully...)

PR #43837, opened 2026-03-16
fix: adjust proxy configs for new api keys and auth

aantti:

YES

- Fix graphql route
- Do not proxy directly to Storage, otherwise new `sb_` keys won't work (see PR #43554)
- Add `KONG_PORT_MAPS` to proxy overlays (without that S3 protocol will currently break)

(An alternative to `KONG_PORT_MAPS` would be using `KONG_TRUSTED_IPS: "0.0.0.0/0,::/0"` as Kong is only reachable internally - but it's also fragile, configuration-wise.)

### DIFF
--- a/docker/docker-compose.caddy.yml
+++ b/docker/docker-compose.caddy.yml
@@ -2,6 +2,8 @@ services:
 
   kong:
     ports: !reset []
+    environment:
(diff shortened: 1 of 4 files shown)


aantti:
@singh-inder, @luizfelmach fyi

jgoux (approved):

PR #43844, opened 2026-03-17
chore(studio): storage file explorer breadcrumbs

dnywh:

UI update that resolves DEPR-114. Also resolves DEPR-113.

- The breadcrumbs on the file explorer have some rough edges in column view
  - Fancy hide/show behavior
  - Hidden tap targets
- `FileExplorerHeader` actions can overflow on the x-axis
- The Navigate button is only shown on hover
- The inline Navigate flow does not work well on smaller screens

## What is the new behavior?

- Column view now shows the same in-explorer breadcrumb trail as list view
- The active breadcrumb is visually emphasized, while inactive breadcrumbs remain clickable
- The back affordance now uses a clearer arrow treatment with a stronger separator from the breadcrumb trail
- The Navigate button is permanently visible and moved to the right-side action group before Reload
- Navigate now opens a dialog on both desktop and mobile
- Added typed telemetry so we can measure `Navigate` usage before deciding whether to keep or remove it
- Fixed header overflow by letting the full header contents scroll horizontally together instead of visibly spilling out

| Before | After |
| --- | --- |
| Buckets  Storage  AWS Healthy  Toolshed  Supabase | Buckets  Storage  AWS Healthy  Toolshed  Supabase |
| Buckets  Storage  AWS Healthy  Toolshed  Supabase | Buckets  Storage  AWS Healthy  Toolshed  Supabase |

## Additional context

Also added a link to the `Buckets` portion of the `PageHeader` breadcrumb:

```text
Files > Buckets > MyBucketName
```

It goes to the same place as Files because the root Files page lists buckets, but having both links there feels more ergonomic in practice.


apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx
```diff
@@ -84,6 +84,7 @@ const BucketPage: NextPageWithLayout = () => {
           },
           {
             label: 'Buckets',
+            href: `/project/${ref}/storage/files`,
```
Comment (dnywh) [inline apps/studio/pages/project/[ref]/storage/files/buckets/[bucketId].tsx:87]: Just some nice ergonomics. Was annoying/confusing to me (the original designer) that “Buckets” wasn’t hyperlinked.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -3,16 +3,17 @@ import { useIsAPIDocsSidePanelEnabled } from 'components/interfaces/App/FeatureP
...
+  ArrowLeft,
   Check,
-  ChevronLeft,
   ChevronRight,
   Columns,
   Edit2,
   FolderPlus,
   List,
-  Loader,
+  LoaderCircle,
```
dnywh, line 16: Will match this on `FileExplorerRow` too.

dnywh: (in an upcoming PR)

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -53,41 +63,72 @@ const SORT_ORDER_OPTIONS = [
...
+interface NavigateDialogProps {
+  open: boolean
+  pathString: string
+  onOpenChange: (open: boolean) => void
+  onPathStringChange: (event: any) => void
+  onCancel: () => void
+  onSubmit: (event?: any) => void
+}
+
+const NavigateDialog = ({
```
dnywh, line 82: Using a dialog for Navigate instead of inline input because:

1. The new button entry point is now away (right) from where the input now renders (left) making the interaction confusing.
2. The current inline input is not reachable on smaller breakpoints.
3. A responsive solution feels unnecessary given we might remove this functionality in the future.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -53,41 +63,72 @@ const SORT_ORDER_OPTIONS = [
...
+            onChange={onPathStringChange}
+            placeholder="parent/child/grandchild"
+            aria-describedby={descriptionId}
+            onKeyDown={(event) => {
+              if (event.key === 'Enter') onSubmit(event)
+            }}
+          />
+          <FieldDescription_Shadcn_ id={descriptionId} className="text-foreground-lighter">
+            Enter a folder path within this bucket.
+          </FieldDescription_Shadcn_>
```
dnywh, line 118: Duplicative of `DialogDescription` but looks so much nicer.

dnywh: `DialogDescription` sticks around for `sr-only`.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -113,24 +154,37 @@ const HeaderBreadcrumbs = ({
...
+            )}
+            <p
+              className={cn(
+                'truncate text-sm transition-colors',
+                isEllipsis && 'text-foreground-light',
+                !isEllipsis && isActive && 'text-foreground',
+                !isEllipsis &&
+                  !isActive &&
+                  'cursor-pointer text-foreground-lighter hover:text-foreground'
+              )}
```
dnywh: Clearer state (selected, unselected, unselected but hovered). Matches `PageHeader` breadcrumbs.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -264,232 +320,223 @@ export const FileExplorerHeader = ({
     setIsRefreshing(false)
   }
 
+  const onOpenNavigate = () => {
+    track('storage_explorer_navigate_clicked')
+    togglePathEdit()
+  }
+
   return (
```
dnywh, line 339: Some fairly intense refactoring required here to fix the following issues:

- Scroll overflow
- Scrollbar squishing the height of contents

packages/common/telemetry-constants.ts
dnywh, line 1: Figured this is a good candidate to get my feet (re)wet with testing. Is this all we need to do @supabase/growth-eng?

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -264,232 +320,223 @@ export const FileExplorerHeader = ({
...
+              <>
                 <Button
-                  key="cancelPath"
-                  type="default"
-                  htmlType="button"
-                  onClick={cancelSetPathString}
-                >
-                  Cancel
-                </Button>,
+                  icon={<ArrowLeft size={16} strokeWidth={2} />}
```
dnywh, line 348: The previous icon—a chevron— looked too confusingly like a backward breadcrumb.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const inactiveBreadcrumb = screen.getByText('images')
+    const activeBreadcrumb = screen.getByText('2024')
+
+    expect(rootBreadcrumb).toBeInTheDocument()
+    expect(inactiveBreadcrumb).toBeInTheDocument()
+    expect(activeBreadcrumb).toBeInTheDocument()
+    expect(inactiveBreadcrumb).toHaveClass('text-foreground-lighter', 'cursor-pointer')
+    expect(activeBreadcrumb).toHaveClass('text-foreground')
+    expect(activeBreadcrumb).not.toHaveClass('text-foreground-lighter')
+    expect(rootBreadcrumb.parentElement?.parentElement).not.toHaveClass('overflow-x-auto')
```
seanoliver: This feels kind of brittle IMO. It's mainly testing DOM nesting structure rather than user-visible behavior, and it'll likely fail on any layout refactor that changes this structure even if nothing changes functionally. Probably safe to remove.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -113,32 +161,45 @@ const HeaderBreadcrumbs = ({
...
+              className={cn(
+                'truncate text-sm transition-colors',
+                isEllipsis && 'text-foreground-light',
+                !isEllipsis && isActive && 'text-foreground',
+                !isEllipsis &&
+                  !isActive &&
+                  'cursor-pointer text-foreground-lighter hover:text-foreground'
+              )}
+              style={{ maxWidth: '6rem' }}
+              onClick={() => (!isEllipsis && !isActive ? selectBreadcrumb(crumb.index) : {})}
```
seanoliver: Clickable breadcrumb items are `<p>` tags with `onClick`, so they're not reachable by keyboard (no `Tab` focus, no `Enter`/`Space` activation). Since these are newly interactive in column view, worth switching inactive crumbs to `<button type="button">` — that also lets you drop the `onClick` guard (`!isEllipsis && !isActive`) since the button just wouldn't render for those cases.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -113,32 +161,45 @@ const HeaderBreadcrumbs = ({
...
+            <p
+              className={cn(
+                'truncate text-sm transition-colors',
+                isEllipsis && 'text-foreground-light',
+                !isEllipsis && isActive && 'text-foreground',
+                !isEllipsis &&
+                  !isActive &&
+                  'cursor-pointer text-foreground-lighter hover:text-foreground'
+              )}
+              style={{ maxWidth: '6rem' }}
```
seanoliver: nit: you could replace the inline style with `max-w-24` to stay consistent with the rest of the Tailwind-only styling in this file.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -203,21 +265,22 @@ export const FileExplorerHeader = ({
...
   }
 
-  const navigateByPathString = (event: any) => {
+  const navigateByPathString = (event?: SyntheticEvent) => {
     if (event) {
       event.preventDefault()
       event.stopPropagation()
     }
-    setIsEditingPath(false)
+    track('storage_explorer_navigate_submitted')
```
seanoliver: This fires before `onSetPathByString` resolves, so it counts failed navigations (bad path, network error, etc.) the same as successful ones. If the goal is using this to decide whether to keep or remove Navigate, it might be worth tracking after the await — or at least adding a `path_depth: compact(pathString.split("/")).length` property so you can tell if people are navigating to deep paths vs. shallow ones.

seanoliver: Minor edge case: if the user clears the input and submits an empty path, this event fires but `onSetPathByString([])` actually navigates to bucket root rather than a specific folder — a meaningfully different action that gets counted the same way.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.test.tsx
seanoliver: This is testing DOM nesting structure rather than user-visible behavior. It may fail with layout refactors even if nothing functionally changes, so it might be worth considering removing or relaxing this.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    const navigateButton = screen.getByRole('button', { name: 'Navigate' })
+    const reloadButton = screen.getByRole('button', { name: 'Reload' })
+
+    expect(navigateButton.closest('div.overflow-x-auto')).toHaveClass('overflow-x-auto')
```
seanoliver: Similar to above, this assertion on the overflow container class is testing layout internals. The `compareDocumentPosition` check below seems like a solid way to verify that Navigate appears before Reload. The overflow assertion may add some fragility without covering user-facing behavior.

packages/common/telemetry-constants.ts
seanoliver: It depends! 🤪 

Curious what the intended measurement strategy is here. The PR description mentions using telemetry to decide whether to keep or remove Navigate — is this meant to be passive usage monitoring, or an actual experiment with a control group?

If it's the latter, a few things that would be needed:
- **Exposure event** — something that fires when the Navigate button becomes visible (e.g. when a user lands in column view with a bucket open). Without it you can only count clicks, not compute a conversion rate against the population who had the opportunity to click.
- **Feature flag** — to gate the behavior so you have a control group to compare against.
- **Defined success metric** — is `storage_explorer_navigate_submitted` the conversion? (i.e. would a user taking this action be enough to tell us that this change is "better" than before, or is there some downstream behavior we're actually targeting?) Also a user could navigate using breadcrumbs instead, so it might be worth taking the time to spell out the ultimate goal.

If this is just "let's see if anyone uses it" monitoring (not a formal experiment), the two events are probably enough to answer that question — just flagging in case the plan is more structured than that.

seanoliver (commented):
This looks really good!

Left a few nitpicky comments you can take or ignore haha.

Left a longer comment RE testing and telemetry since it kind of depends on what you want to learn from this and whether you want to do a test vs. just a rollout and monitor situation. These are just initial reactions but I'm happy to go deeper if you're not sure about anything!

packages/common/telemetry-constants.ts
dnywh: Kept this intentionally in the simple-monitoring bucket rather than turning it into an experiment. The PR still uses just click + submit events, but the submit event is now tighter: it only fires after a non-root folder-path navigation completes. I didn’t add exposure events, flags, or extra telemetry shape in this PR.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
dnywh: Good shout. Fixed.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
dnywh: Fixed! Clearing the input and navigating back to bucket root no longer emits `storage_explorer_navigate_submitted`, so root navigation is no longer counted the same as folder-path usage

dnywh: Fixed. `storage_explorer_navigate_submitted` now fires after the non-root folder navigation flow completes, instead of before `fetchFoldersByPath` runs.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
dnywh: Good idea. Done.

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.test.tsx
dnywh: Thanks. I removed the overflow-container assertion. The test now keeps the behaviour-level ordering check (`Navigate` before `Reload`) without depending on a particular layout wrapper or class structure

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.test.tsx
dnywh: Relaxed the tests. Now just checks the breadcrumb integration itself rather than the surrounding structure

awaseem (approved):
Love it!

apps/studio/components/interfaces/Storage/StorageExplorer/FileExplorerHeader.tsx
```diff
@@ -117,28 +165,40 @@ const HeaderBreadcrumbs = ({
...
-          </p>
-        </div>
-      ))}
+    <div className="ml-3 flex min-w-0 flex-1 items-center overflow-hidden">
+      {formattedBreadcrumbs.map((crumb, idx: number) => {
+        const isEllipsis = crumb.name === ellipsis
+        const isActive = crumb.index === breadcrumbs.length - 1
+
+        return (
+          <div className="flex shrink-0 items-center" key={crumb.name}>
```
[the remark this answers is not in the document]
seanoliver: This one seems worth addressing before we merge!

seanoliver (approved):
This looks great @dnywh! LGTM to ship! 🔥 

There's one comment from copilot about duplicate react keys on repeated folder names that's definitely an edge case but might be worth fixing as it would cause some UI funkiness if it ever happened.

PR #43848, opened 2026-03-17
Chore/shift manual table editor queries into pg meta

djhi:
Move some studio queries to `pg-meta`, related to the table editor


soedirgo (approved):

joshenlim (approved):
Verified on staging preview
- Create table (with/without RLS, with/without realtime, with foreign key)
- Duplicate table ^ same permutations
- Delete table

PR #43854, opened 2026-03-17
fix: prevent unnecessary dialog after saving a new edge function

djhi:
## Problem

After deploying a new edge function where you actually modified the code from the initial template, a dialog appears about unsaved changes.

## Solution

We must check that even though the files have changed from the initial ones, we actually saved them by deploying the function. 

However, because we trigger the navigation in the mutation `onSuccess` side effect, it happens before react can update the state that would prevent the dialog from appearing.


jordienr (approved):
good one

PR #41362, opened 2025-12-15
Fix(studio): prevent importing rows as CSV to protected entities

marsou001:
Fixes #41358 

## What kind of change does this PR introduce?

Bug fix

## What is the current behavior?

Users able to add row data as CSV to any empty entity

## What is the new behavior?

Users can't import CSV rows to any table, and instead they are shown the relevant UI message depending on the case, if:

1. The entity is not a table.
2. The entity schema is protected.
3. User is not allowed to edit tables.

## Additional context

If the schema is protected (`auth`):

Screenshot (10)

If the entity is not a table (a view in this case):

Screenshot (11)

If the user is not allowed (I switched a boolean value, just to check the UI):

Screenshot (12)

Edit: added a special case handling foreign tables:

Screenshot (15)

## Summary by CodeRabbit

* **New Features**
  * Permission- and entity-aware CSV import: import option only appears when allowed; interactive empty state supports drag-and-drop, an “Import data from CSV” button, and contextual messaging for foreign tables/non-table entities/protected schemas.
  * Telemetry: import button clicks emit usage events.

* **Bug Fixes**
  * Prevented import actions and clarified UI when importing is disallowed.

* **Refactor**
  * Improved grid interaction (column dragging, selection, row styling, and edit behaviors) for a smoother editing experience.

Issue #41358, opened 2025-12-15
Import data as CSV is allowed on protected schema and other scenarios


marsou001:
# Context

Users can add rows to tables by importing data as CSV either by:

1. Clicking on the green dropdown button "Insert", and choosing "Insert row".
2. If the table is empty, dragging and dropping a csv file, or clicking "Import data from CSV".

Some tables are not editable, either because:

1. They belong to a protected schema.
2. They are not table entities (views, materialized views...).
3. Some users don't have the required permission to edit them.

In those cases, the green "Insert" button disappears, preventing unwanted row additions to protected entities, but there is still the option to import the rows as CSV data if the entity is emptys.

That's the bug.

## Screenshots

Here is an example:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/2d93d852-7f27-4a2c-8c4c-a131fef57ab9" />

The `instances` table of `auth` schema is supposed to be locked. Yet, when I clicked on "Import data from CSV" and uploaded a row that matches the data types of the columns, the data was successfully added to the table.

This option should not exist in such a case.

PR #43857, opened 2026-03-17
fix: hide CSV import on empty table when table is not editable

mango766:

YES

Bug fix

When a table is empty, the Table Editor shows an "Import data from CSV" button and accepts CSV drag-and-drop regardless of whether the table is actually editable. This means users can import data into:

- Protected schema tables (e.g. `auth.instances`, `auth.users`)
- Views and materialized views
- Tables the user doesn't have permission to edit

The `Header` component already correctly checks `snap.editable` before showing the import button, but the empty-table view in `Grid.tsx` is missing this check.

Related: #41358

## What is the new behavior?

The empty table view now checks `snap.editable` before:
1. Showing the "Import data from CSV" button
2. Enabling CSV drag-and-drop via `useCsvFileDrop`

For non-editable tables, the empty state just shows "This table is empty" without the import option — consistent with how the Header already hides the Insert dropdown for non-editable tables.

## Additional context

This is a minimal fix (2 lines of logic changed) that reuses the same `snap.editable` flag the Header already relies on. There was a previous attempt at this (#41362) but it went stale — this approach is simpler and follows the existing pattern in the codebase.

### DIFF
--- a/apps/studio/components/grid/components/grid/Grid.tsx
+++ b/apps/studio/components/grid/components/grid/Grid.tsx
@@ -97,8 +97,10 @@ export const Grid = memo(
 
       const { mutate: sendEvent } = useSendEventMutation()
 
+      const isEditable = snap.editable
+
       const { isDraggedOver, onDragOver, onFileDrop } = useCsvFileDrop({
-        enabled: isTableEmpty && !isForeignTable,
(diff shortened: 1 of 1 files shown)


7ttp (commented):
should be resolved by https://github.com/supabase/supabase/pull/47040

thanks for ur efforts towards contributing to Supabase !! 💚

PR #43858, opened 2026-03-17
docs: Update Ionic React guide with correct auth methods, updated dependencies, and code sample component

ChrisChinchilla:

YES


mandarini (changes requested):
1. I these files appear to be leftover Ionic/CRA scaffold files rather than part of the actual tutorial flow:

- `examples/user-management/ionic-react-user-management/src/pages/Home.tsx`
- `examples/user-management/ionic-react-user-management/src/pages/Home.css`
- `examples/user-management/ionic-react-user-management/src/components/ExploreContainer.tsx`
- `examples/user-management/ionic-react-user-management/src/components/ExploreContainer.css`
- `examples/user-management/ionic-react-user-management/src/App.test.tsx`
- `examples/user-management/ionic-react-user-management/src/setupTests.ts`
- `examples/user-management/ionic-react-user-management/src/react-app-env.d.ts`

Possibly also “starter plumbing” files that are necessary for the app to run, but not really tutorial-specific:

- `examples/user-management/ionic-react-user-management/public/index.html`
- `examples/user-management/ionic-react-user-management/public/manifest.json`
- `examples/user-management/ionic-react-user-management/public/assets/icon/favicon.png`
- `examples/user-management/ionic-react-user-management/public/assets/icon/icon.png`
- `examples/user-management/ionic-react-user-management/public/assets/shapes.svg`
- `examples/user-management/ionic-react-user-management/src/theme/variables.css`

Maybe we could remove them?

2) Should we make the auth/profile refresh more explicit and reduce the “unused scaffold” feel?

We could remove unused scaffold routes/components. If the tutorial never navigates to `Home`, `ExploreContainer`, or the default starter page, we could delete those files and replace them with only the files the example actually uses.

Could we also maybe tighten the avatar callback type?

Now it's like:
```ts
onUpload: (e: any, file: string) => Promise<void>;
```

and the component calls:

```ts
onUpload(null, fileName);
```

It could be something like:

```ts
onUpload: (file: string) => Promise<void>;
```

and then:

```ts
onUpload(fileName);
```

That removes the fake event parameter.

We also could have `Avatar` just return the uploaded file name, and then let `AccountPage` handle updating state + profile persistence in one place.

These are not blockers, just I think it makes things more concise?

mandarini (approved):

PR #43865, opened 2026-03-17
fix: remove oauth apps from self hosted

awaseem:

YES

self hosted fix: remove oauth apps and add tests so this never happens again


jordienr (approved):
tested in preview

PR #43866, opened 2026-03-17
fix: Bucket policies actions menu is hidden on Safari

djhi:
## Problem

image

## Solution

image

Tested on Safari, Chrome and Firefox


awaseem (approved):
Tested in firefox and safari

PR #42364, opened 2026-02-01
fix: improve bucket link handling in storage explorer

7ttp:
## Summary

Fix  Edge case in incorrect URL generation in Storage Explorer when clicking "Get URL" on files located outside the currently opened folder.

## Problem

When inside a subfolder ( `bucket/folder/`), 
clicking "Get URL" on a file from a parent directory (`bucket/image.jpg`) generates a broken 404 link like `bucket/folder/image.jpg`
instead of the correct `bucket/image.jpg`.

## Solution

Use the file's intrinsic `path` property from the API response for URL generation, instead of constructing the path from the current UI navigation context.

## Related

- closes https://github.com/supabase/supabase/issues/42357

## Summary by CodeRabbit

* **Improvements**
  * Enhanced "Get URL / Copy URL" so it consistently uses the full file item (prefers explicit file path when present) and works uniformly across preview, row, and context menus.
  * Expiry actions now reliably apply selected expiry when copying URLs.

* **Bug Fixes**
  * Fixed inconsistent URL generation when items include an explicit path.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #43875, opened 2026-03-17
fix: storage copy url include the opened folder for files that are not in it

djhi:
Fixes #42357
Supersedes #42364

## Problem

When copying the URL of a file that is not in the currently opened folder, the folder path is still included in that file URL.

## Solution

We actually always have the file path, use it directly instead of rebuilding it from the storage explorer state

## How to test

1. Create a bucket and upload a file in it
2. Create a folder and upload a file in it

While the folder is selected, right click the root file uploaded in 1 and copy its URL. It should not contain the folder name. Same with the dropdown menu that appears when hovering the file.


awaseem (approved):
LGTM

PR #43890, opened 2026-03-18
bugfix: failing data api granular controls

doublethink:
In the Data API tab: 
* Fixed a logic issue whereby the dropdowns for tables were accurately reflected by exposed schemas but the counts in the dropdown header were not
* Fixed a syntax issue causing the NOT IN clause of the dropdown counts to generate invalid SQL if no schemas were ignored.

Fixes PRODSEC-83


doublethink:
Already being addressed in https://github.com/supabase/supabase/pull/43704

PR #43892, opened 2026-03-18
fix(studio): tighten webhook endpoint validation

dnywh:

Bug fix.

Webhook endpoint validation is inconsistent across Studio forms. The webhook sheet accepts incomplete hostnames like `https://webhook`, event type validation is not surfaced clearly, and HTTP endpoint validation differs between webhooks, log drains, cron jobs, and database hooks.

## What is the new behavior?

- Tightens webhook endpoint URL validation and rejects incomplete hostnames while still allowing localhost and IP-based endpoints.
- Surfaces the Event types validation through the standard form error styling and highlights the existing accordion item border when invalid.
- **Extracts a shared HTTP endpoint URL validator and reuses it in webhooks, log drains, cron jobs, and database hooks.**
- Adds focused regression tests for the webhook sheet and the shared/consumer validation paths.

| After |
| --- |
| Webhooks  Settings  AWS Healthy  Toolshed  Supabase-CB0D999C-D0BF-47AA-A10F-342A2E328DF9 | 

## To test

Form behaviour (in particular URL validation) on:

- Webhook endpoint
- Log drains
- Cron jobs
- Database hooks


apps/studio/components/interfaces/Database/Hooks/EditHookPanel.constants.ts
```diff
@@ -1,14 +1,14 @@
 import { z } from 'zod'
 
-import { isValidHttpUrl } from '@/lib/helpers'
```
dnywh, line 3: Didn’t delete this because it isn’t fully redundant yet. It still has one runtime use left in `PublishAppSidePanel`.

apps/studio/components/interfaces/Integrations/CronJobs/CreateCronJobSheet/CreateCronJobSheet.constants.ts
```diff
@@ -1,8 +1,8 @@
 import { toString as CronToString } from 'cronstrue'
 import z from 'zod'
 
-import { urlRegex } from 'components/interfaces/Auth/Auth.constants'
```
dnywh, line 4: Ditto: `urlRegex` is still needed for auth flows because those URLs may allow things like:

- auth redirect-style patterns
- wildcards

jordienr (approved):
looks like mostly tests and the rest lgtm!

PR #43906, opened 2026-03-18
fix: create index form dropdown behavior

7ttp:
## TL;DR
Fixes/improves the create Index form ux by fixing schema/table dropdown state coupling and dropdown scroll behavior

## Before
 https://github.com/user-attachments/assets/093a6b42-b13e-4138-ab86-45149f4894cb

## After
https://github.com/user-attachments/assets/db9cc11b-0218-4276-9ccf-9c81831e95e5

## Related
- closes https://github.com/supabase/supabase/issues/43878


vickyprus100 (commented):
Ran locally with fix seem to work perfect

awaseem (approved):
LGTM thanks for the fix!

PR #43914, opened 2026-03-18
fix: SMTP form shows incorrect rate limit message when already enabled

vj2303:
## Description

Fixes #43800

When editing SMTP settings while custom SMTP is already enabled, the form footer was incorrectly displaying "Rate limit for sending emails will be increased to 30". The rate limit is only set during the initial enable transition, so the message and submit behavior were out of sync.

## Changes

This fix aligns the footer message with the submit logic (Option A from the issue):

- **Enabling SMTP**: Shows rate limit increase message
- **Already enabled, editing settings**: Shows "Custom SMTP settings will be updated"
- **Disabling SMTP**: Shows rate limit reduction message

## Testing

1. Go to Authentication → SMTP Settings
2. Enable custom SMTP and save (rate limit is set to 30)
3. Edit any SMTP field (e.g. change host or port)
4. Footer now correctly shows "Custom SMTP settings will be updated" instead of the rate limit message

Made with [Cursor](https://cursor.com)


dnywh (approved):
Reviewed against the current SMTP submit path in SmtpForm. This matches the existing behaviour: the rate-limit bump only happens on the initial enable transition, so ordinary edits to an already-enabled SMTP config should show neutral update copy instead.

dnywh:
Small nit: just removed the _Custom SMTP settings will be updated_ from ordinary edits as it’s redundant and inconsistent with our other dirty forms. 

Just running tests before merging. Thanks again for your contribution!

vj2303:
Thanks for the review and the cleanup! Makes sense to keep it consistent with the other forms

PR #43918, opened 2026-03-18
Add banner for queue operations

joshenlim:
## Context

Similar to how we added a banner for the new table filter bar - mainly to improve visibility for this feature preview as we'd like some feedback from usage

image

## Other changes
- Also adding some typing to the `addBanner` and `dismissBanner`
- Removed banner for table editor filter bar


awaseem (approved):
LGTM

PR #43926, opened 2026-03-18
chore: split Copilot review guidelines into topic-specific files

seanoliver:
## Context

Noticed while working on #43913 that `copilot-instructions.md` is currently at ~4,600 characters. Per [GitHub's docs on Copilot code review](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review):

> Copilot code review only reads the first 4,000 characters of any custom instruction file. Any instructions beyond this limit will not affect the reviews generated by Copilot code review.

This means the testing section at the bottom of our current file isn't being read during reviews.

## Proposal

Split the single file into path-specific instruction files under `.github/instructions/`, following [GitHub's recommended pattern for repository custom instructions](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions):

> These are specified in one or more `NAME.instructions.md` files within or below the `.github/instructions` directory.

> If the path you specify matches a file that Copilot is working on, and a repository-wide custom instructions file also exists, then the instructions from both files are used.

This gives us separate files that each stay under the 4K limit and get combined automatically by Copilot during reviews:

| File | Size | Scope |
|------|------|-------|
| `copilot-instructions.md` | 929 chars | General repo context + pointers |
| `instructions/studio-telemetry.instructions.md` | 3,570 chars | Telemetry rules for `apps/studio/**` |
| `instructions/studio-testing.instructions.md` | 1,228 chars | Testing rules for `apps/studio/**` |

Note: Copilot reads instructions from the **base branch** of a PR, not the feature branch — so these won't take effect until merged to master.

### New telemetry guidance

The telemetry file adds guidance we've been missing — specifically around feature-flagged rollouts:

- Flag PRs that use `usePHFlag`/`useFlag` to gate behavior but don't capture the flag state in telemetry
- Flag rollouts that track flag state but not user response to the new behavior
- Documents the raw flag pattern (read via `usePHFlag`, not coerced wrapper hooks) to avoid the `undefined`→`false` data quality bug we hit in #43913

### What didn't change

All existing telemetry and testing rules are preserved — nothing was removed, just reorganized. The telemetry rules still reference `.claude/skills/telemetry-standards/SKILL.md` as the authoritative source.

## References

- [Adding repository custom instructions for GitHub Copilot](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions) — file structure, path-specific instructions, frontmatter format
- [Using Copilot code review](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review) — 4K character limit, base branch behavior

## Open questions

Would love the team's input on:
- Does the file split make sense, or would you prefer keeping everything in one file (and trimming to fit)?
- Are there other topics that should get their own instruction file?
- Any concerns with the new feature flag telemetry guidance?

### DIFF
--- a/.claude/skills/telemetry-standards/SKILL.md
+++ b/.claude/skills/telemetry-standards/SKILL.md
@@ -14,9 +14,9 @@ reviewing PRs that touch tracking or when implementing new tracking.
 
 **Format:** `[object]_[verb]` in snake_case
 
-**Approved verbs only:**
-opened, clicked, submitted, created, removed, updated, retrieved, intended, evaluated, added,
-enabled, disabled, copied, exposed, failed, converted
+**Approved verbs only** (canonical list — derived from `packages/common/telemetry-constants.ts`):
+opened, clicked, submitted, created, removed, updated, intended, evaluated, added,
+enabled, disabled, copied, exposed, failed, converted, closed, completed, applied, sent, moved
 
 **Flag these:**
 - Unapproved verbs (saved, viewed, seen, pressed, etc.)
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -9,88 +9,14 @@ This is a TypeScript/Next.js/React monorepo:
 - `apps/docs/` — Documentation
 - `packages/common/` — Shared code including telemetry definitions
 
----
+## Topic-Specific Guidelines
 
-## Telemetry Review Rules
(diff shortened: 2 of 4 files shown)


awaseem (approved):
LGTM

myerekapan (approved):

PR #43957, opened 2026-03-19
Add support for cmd z undo in table editor for queue operations

joshenlim:
## Context

Add support for undo-ing an operation via keyboard shortcut in Table Editor for the queue operations feature preview


djhi (approved):
Tested on staging 💯

PR #43958, opened 2026-03-19
Simplify SaveQueueActionBar positioning logic to just CSS

joshenlim:
## Context

Noticed that the position of the Save queue action bar shifts off screen after switching tables
image

Changes here just simplifies the positioning logic to be purely css


djhi (approved):
Tested on staging 💯

PR #43960, opened 2026-03-19
fix(Docs): Improve await error handling on external content sources.

jeremenichelli:

In this PR we handle the rejected Promises coming from fetched federated sources of content to avoid build `TIMEOUT` errors and get the actual reason.


ChrisChinchilla (approved):
OK, let's give it a go!

PR #43964, opened 2026-03-19
Refactor queue operations to not mutate the query client to simplify implementation

joshenlim:
## Context

If you've got a queued operation on a table, then switch tables, the dashboard stalls for a few seconds before finally rendering the selected table.

This was caused by an infinite loop in a useEffect calling `reapplyOptimisticUpdates` due to `dataUpdatedAt` being in the dependency array (which comes from react query)

Removing `dataUpdatedAt` would resolve the issue, but then it won't solve the following comment in the code:
`// This ensures pending changes remain visible when switching tabs or after data refresh`

## Changes involved

Opting to refactor the logic for rendering data in the Grid instead
- Avoid manipulating the data directly from the query client
  - Manual changes to the query client are hard to track when debugging
  - Ideally the data in the query client are exactly what is coming from the API
- Instead just compute the row data if there's operations applied, and render that in the grid
  ```
  const baseRows = data?.rows ?? EMPTY_ARR
  const rows = formatGridDataWithOperationValues({ operations, rows: baseRows })
  ```
  - Simplifies operations logic + data in react query remains as the source of truth
  - This also improves perceived performance, as previously we'd need to invalidate the query client if we're removing any operation. 
    - A lot more apparent with the undo operation introduced [here](https://github.com/supabase/supabase/pull/43957)
    - Whereas now, we'll just revert back to whatever's in the query client (and still do the invalidation behind the scenes) so things feel faster

### DIFF
--- a/apps/studio/components/grid/SupabaseGrid.tsx
+++ b/apps/studio/components/grid/SupabaseGrid.tsx
@@ -1,4 +1,4 @@
-import { keepPreviousData, useQueryClient } from '@tanstack/react-query'
+import { keepPreviousData } from '@tanstack/react-query'
 import { useParams } from 'common'
 import { isMsSqlForeignTable } from 'data/table-editor/table-editor-types'
 import { useTableRowsQuery } from 'data/table-rows/table-rows-query'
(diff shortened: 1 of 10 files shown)


awaseem (approved):
Thank you sir for cleaning up my mess LOL

PR #43974, opened 2026-03-19
Hide is unique option for column that's a primary key

joshenlim:
## Context

Table Editor - hide "Is unique" configuration for a column that's a primary key - unnecessary to configure a column as unique if it's already a primary key (redundant index is created)

## Changes involved

- TableEditor side panel: No longer shows is unique option under configuration popover if column is a PK
- ColumnEditor side panel: Disable is unique + is nullable option if column is a PK (Added tooltip)
image

Also chucked in some minor refactors:
- Refactored some `any` types
- Refactored Column to use the latest Checkbox component


awaseem (approved):
Tested on staging and self hosted

PR #43979, opened 2026-03-19
Use new sheet component + small nudges to UI

joshenlim:
## Context
Resolves FE-2826
Resolves FE-2825
Small refactors + UI nudges to the queue operations side panel

## Before

image

## After

image


awaseem (approved):
LGTM

PR #43987, opened 2026-03-19
Chore improve e2e tests

djhi:
## Problem

Some tests rely on hard coded timeouts. That makes them
- brittle if the timeout is not long enough
- take longer than necessary if the timeout is too long

## Solution

- Rely on playwright `expect` retries when possible
- Rely on UI updates when possible


awaseem (approved):
LGTM

PR #43998, opened 2026-03-19
fix(observability): use 30m step for 7-day dashboard to preserve spikes FE-2813

jordienr:
## Problem

The database observability dashboard flattens critical spikes when viewing the 7-day window. A spike clearly visible at the 3-hour view (2-minute resolution) disappears completely at the 7-day view because the Prometheus query step was 1 hour, averaging each spike into a 60-minute bucket and reducing its apparent magnitude to near zero.

## Fix

Increase the granularity in the 7 day window to prevent spikes from being flattened.
Data point counts: 7-day view goes from ~168 to ~336 points per series, more consistent with the existing 12-hour view (~360 points).

## How to test

- Open the observability database dashboard for a project that has had a recent CPU, RAM, or connection spike
- Select the "Last 3 hours" time window and confirm the spike is visible
- Switch to "Last 7 days" and confirm the same spike is now visible (previously it would disappear)
- Confirm all other time windows (10m, 30m, 1h, 24h) still load and display correctly


alaister (approved):
LGTM!

- Tested on preview

PR #44001, opened 2026-03-19
chore: remove filter feature flag

awaseem:

YES

chore: remove feature flag and make default opt in true

### DIFF
--- a/e2e/studio/features/table-editor.spec.ts
+++ b/e2e/studio/features/table-editor.spec.ts
@@ -418,6 +418,11 @@ testRunner('table editor', () => {
         await dropTable(tableName)
       }
     )
+    // Disable the new filter bar for this test since it uses the old filter UI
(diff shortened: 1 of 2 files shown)


e2e/studio/features/table-editor.spec.ts
```diff
@@ -418,6 +418,11 @@ testRunner('table editor', () => {
         await dropTable(tableName)
       }
     )
+    // Disable the new filter bar for this test since it uses the old filter UI
+    await page.goto(toUrl(`/project/${ref}/editor?schema=public`))
+    await page.evaluate(() => {
+      localStorage.setItem('supabase-ui-table-filter-bar', 'false')
+    })
```
awaseem, line 425: This wasn't failing before because the flag evaluates to false in the test env even though its been enabled for more than a few weeks. This test case is still related to the old filter bar, since users can still opt out we keep this just in case

charislam (approved):

PR #44009, opened 2026-03-20
Edge function overview

SaxonF:
image

Attempts to improve our edge function overview pages to provide stronger insights into the health of a function, including reliability (error rates), performance (execution times) and usage (cpu and memory).

As part of this work it refactors existing charts to use our new chart components.

main consideration is the collective performance of error queries https://github.com/supabase/supabase/pull/44009/changes#diff-2a79cf61c5397a8ef363c333229fa7729a2efc90a4d8e0806e49c212d5aa97e7

## To test:
1. Create an edge function that errors out randomly across requests. You can use cron to poll this function every second.
2. View the edge function and on the overview page confirm that errors are showing and grouped correctly in recent failed invocations sections.


apps/studio/components/interfaces/Functions/EdgeFunctionInvocationsSection.tsx
```diff
@@ -0,0 +1,250 @@
...
+const toAlertError = (error: unknown): { message: string } | undefined => {
+  if (typeof error === 'string') return { message: error }
+
+  if (error && typeof error === 'object') {
+    const message = (error as { message?: unknown }).message
+    if (typeof message === 'string') return { message }
+  }
+
+  return undefined
+}
```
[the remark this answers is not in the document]
awaseem: +1 on this on @SaxonF, should reuse or move to utils

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionInvocationsSection.tsx
```diff
@@ -0,0 +1,250 @@
...
+                      </div>
+                    )}
+                  </div>
+                )}
+              </Chart>
+            </div>
+          </div>
+        </PageContainer>
+      </PageSectionContent>
+    </PageSection>
```
awaseem, line 154: The chart it self can be its own component, this can reduce the size of this file down a lot and make the ternaries easier to follow

apps/studio/components/interfaces/Functions/EdgeFunctionInvocationsSection.tsx
```diff
@@ -0,0 +1,250 @@
...
+  isLoadingChart,
+  isErrorChart,
+  chartErrorMessage,
+  chartData,
+  onChartClick,
+  updateAnnotation,
+}: EdgeFunctionInvocationsSectionProps) => {
+  const dateTimeFormat = selectedInterval.format ?? 'MMM D, h:mma'
+  const emptyStateCopy = getChartEmptyStateCopy('invocations', isErrorChart, chartErrorMessage)
+  const timeRangeLabels = getChartTimeRangeLabels(chartData, dateTimeFormat)
```
awaseem: We should useMemo these right?

apps/studio/components/interfaces/Functions/EdgeFunctionInvocationsSection.tsx
```diff
@@ -0,0 +1,250 @@
...
+  formatChartTimestamp,
+  formatRate,
+  getChartEmptyStateCopy,
+  getChartTimeRangeLabels,
+  getSegmentedButtonClassName,
+  INVOCATION_CHART_CONFIG,
+} from './EdgeFunctionOverview.utils'
+import type { InvocationChartDatum, InvocationUpdateAnnotation } from './EdgeFunctionOverview.utils'
+
+type InvocationActions = ComponentProps<typeof ChartActions>['actions']
```
awaseem: Should also move this type out to the utils or resuse

apps/studio/components/interfaces/Functions/EdgeFunctionOverview.utils.ts
```diff
@@ -0,0 +1,264 @@
...
+  redirect_count?: string | number
+  client_err_count?: string | number
+  server_err_count?: string | number
+  avg_execution_time?: string | number
+  max_execution_time?: string | number
+  avg_cpu_time_used?: string | number
+  max_cpu_time_used?: string | number
+  avg_memory_used?: string | number
+  avg_heap_memory_used?: string | number
+  avg_external_memory_used?: string | number
```
awaseem: Need help understanding this, how can it be a string or a number? Most of the utility functions are converting them back to numbers so how come its both? Do we need 2 types here? One thats the base and then one that happens after the transformation?

apps/studio/components/interfaces/Functions/EdgeFunctionOverview.utils.ts
```diff
@@ -0,0 +1,264 @@
...
+  timestamp: string
+  ok_count: number
+  warning_count: number
+  error_count: number
+}
+
+export type InvocationUpdateAnnotation = {
+  timestamp: string
+  position: number
+  updatedAt: Dayjs
```
awaseem: Should be `Date`, we are bleeding libraries into types which is generally a bad idea. When we do move to Temproal or other date libraries this will be a PITA

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionOverview.utils.ts
```diff
@@ -0,0 +1,264 @@
...
+
+export const getRollingTimeRange = (
+  interval: ChartIntervals,
+  now: Dayjs = dayjs()
+): [Dayjs, Dayjs] => {
+  const end = now
+  const start = end.subtract(interval.startValue, interval.startUnit as dayjs.ManipulateType)
+
+  return [start, end]
+}
```
awaseem, line 145: These should be all dates as inputs and outputs. Its fine to use Dayjs inside the methods, but we shouldn't expose them as the function args or outputs. If what your doing is returning dayjs values then calling formatters. You should rather always return dates and then create a utility method to convert date to a formatted string. inside those methods we can call dayjs

apps/studio/components/interfaces/Functions/EdgeFunctionOverview.utils.ts
```diff
@@ -0,0 +1,264 @@
...
+export const getInvocationUpdateAnnotation = ({
+  updatedAt,
+  invocationChartData,
+  windowStart,
+  windowEnd,
+}: {
+  updatedAt?: string
+  invocationChartData: InvocationChartDatum[]
+  windowStart: Dayjs
+  windowEnd: Dayjs
```
awaseem: Same comment as above, lets not leak dayjs returns

apps/studio/components/interfaces/Functions/EdgeFunctionPerformanceSection.tsx
```diff
@@ -0,0 +1,135 @@
...
+  data: EdgeFunctionChartDatum[]
+  dateTimeFormat: string
+  isLoading: boolean
+  isError: boolean
+  errorMessage?: string
+  averageExecutionTime: number
+  maxExecutionTime: number
+}
+
+const renderExecutionTooltipDetails = (
```
awaseem: This can be its own component in a separate file, a function that return JSX should be treated like a component rather than a utility

apps/studio/components/interfaces/Functions/EdgeFunctionRecentErrors.tsx
```diff
@@ -0,0 +1,326 @@
...
+  getFunctionRuntimeLogsSql,
+  getRecentErrorGroups,
+  getRecentErrorGroupsBase,
+  getRecentErrorInvocationsSql,
+  getRelatedExecutionIds,
+  toAlertError,
+  type RecentErrorGroup,
+} from './EdgeFunctionRecentErrors.utils'
+
+dayjs.extend(relativeTime)
```
awaseem: When we split the formatters as separate utility functions, we should add this in there instead of the component layer.

apps/studio/components/interfaces/Functions/EdgeFunctionRecentErrors.tsx
```diff
@@ -0,0 +1,326 @@
...
+
+  const getStatusBadgeVariant = (statusCode?: string) => {
+    if (!statusCode) return 'destructive' as const
+
+    const status = Number(statusCode)
+    if (Number.isNaN(status)) return 'destructive' as const
+    if (status >= 500) return 'destructive' as const
+
+    return 'default' as const
+  }
```
awaseem: These are all pure functions that can be separated out into new files based on their context.

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionRecentErrors.utils.ts
```diff
@@ -0,0 +1,185 @@
...
+export const toAlertError = (error: unknown): { message: string } | undefined => {
+  if (typeof error === 'string') return { message: error }
+
+  if (error && typeof error === 'object') {
+    const message = (error as { message?: unknown }).message
+    if (typeof message === 'string') return { message }
+  }
+
+  return undefined
+}
```
awaseem, line 55: Do we not have any other utility functions in studio that does this rather than creating new ones in this file?

apps/studio/components/interfaces/Functions/EdgeFunctionRecentErrors.utils.ts
```diff
@@ -0,0 +1,185 @@
...
+cross join unnest(metadata) as metadata
+where metadata.function_id = '${escapeSqlString(functionId)}' and metadata.execution_id in (${escapedExecutionIds})
+order by timestamp desc
+limit ${limit}`
+}
+
+export const getRecentErrorGroupsBase = (
+  recentErrorInvocations: LogData[]
+): RecentErrorGroupBase[] => {
+  const grouped = recentErrorInvocations.reduce<Record<string, RecentErrorGroupBase>>(
```
awaseem: This is really hard to read as a reduce, I'm actually more in favor of making this into a for loop

apps/studio/components/interfaces/Functions/EdgeFunctionUsageSection.tsx
```diff
@@ -0,0 +1,219 @@
...
+      {formatReferenceDelta(Number(value ?? 0), averageMemoryUsage)}
+    </span>
+    <span className="text-foreground-light">
+      {getMemoryTooltipDetail(
+        Number(datum.avg_heap_memory_used ?? 0),
+        Number(datum.avg_external_memory_used ?? 0)
+      )}
+    </span>
+  </>
+)
```
awaseem: Can be react components instead rather than utility functions. Cleans up the callers as well

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionUsageSection.tsx
```diff
@@ -0,0 +1,219 @@
...
+                      />
+                    </div>
+                  </ChartContent>
+                </ChartCard>
+              </Chart>
+            </div>
+          </div>
+        </PageContainer>
+      </PageSectionContent>
+    </PageSection>
```
awaseem, line 156: Any chance we can create a resuable chart that can be used here and in performance section? Those components look identical

apps/studio/pages/project/[ref]/functions/[functionSlug]/index.tsx
```diff
@@ -125,307 +109,136 @@ const PageLayout: NextPageWithLayout = () => {
...
+        isLoading={combinedStatsResults.isLoading}
+        isError={isErrorCombinedStats}
+        errorMessage={combinedStatsError?.message ?? 'Unknown error'}
+        averageCpuTime={averageCpuTime}
+        maxCpuTime={maxCpuTime}
+        averageMemoryUsage={averageMemoryUsage}
+        totalHeapMemory={totalHeapMemory}
+        totalExternalMemory={totalExternalMemory}
+        totalMemoryByType={totalMemoryByType}
+      />
```
Comment (awaseem) [inline apps/studio/pages/project/[ref]/functions/[functionSlug]/index.tsx]: Very clean!

awaseem (changes requested):
Composition is good to start, there needs to be more clean up around the Logs query data to the transform. We also need to remove DayJS as inputs/outputs and make them generic dates, more utility methods in separate files

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionRecentErrors.tsx
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  return (
+    <PageSection>
+      <PageSectionContent>
+        <PageContainer size="full">
+          <div className="flex flex-col gap-6">
+            <PageSectionMeta>
+              <PageSectionSummary>
+                <PageSectionTitle>Recent Failed Invocations</PageSectionTitle>
```
dnywh, line 137: Really minor nit: _Recent failed invocations_ instead of _Recent Failed Invocations_? To match our sentence case uses elsewhere.

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionOverview.utils.ts
```diff
@@ -0,0 +1,297 @@
...
+  error_count: number
+}
+
+export type InvocationUpdateAnnotation = {
+  timestamp: string
+  position: number
+  updatedAt: Date
+}
+
+export const EDGE_FUNCTION_CHART_INTERVALS: ChartIntervals[] = [
```
dnywh, line 51: I noticed that this replaces the Refresh button. Should that sick around as an IconButton perhaps? To the left of these intervals?

Image

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionPerformanceSection.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const emptyStateCopy = getChartEmptyStateCopy('execution time', isError, errorMessage)
+  const tooltipDetails = useMemo(
+    () => getExecutionTooltipDetails(averageExecutionTime),
+    [averageExecutionTime]
+  )
+  const metrics = (
+    <div className="flex flex-wrap gap-x-8 gap-y-4">
+      <ChartMetric
+        label="Average Execution Time"
+        value={formatMetric(averageExecutionTime, 'ms')}
```
dnywh, line 49: The `ms` and `MB` values on the Y axis get ever-so-slightly clipped on Chrome:

| ms | MB |
| --- | --- |
| Image | Image |

dnywh (approved):
Tested on staging. Some very minor comments. Looks good!

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionMetricTooltipDetails.tsx
```diff
@@ -0,0 +1,93 @@
...
+}) => {
+  return (
+    <>
+      <span className="text-foreground">
+        {formatReferenceDelta(Number(value ?? 0), averageMemoryUsage)}
+      </span>
+      <span className="text-foreground-light">
+        {getMemoryTooltipDetail(datum.avg_heap_memory_used, datum.avg_external_memory_used)}
+      </span>
+    </>
```
Ziinc, line 55: shifting the data calculations to above the returns and setting it as const would be more readable.
If placing inline is preferred then should return the fragment without the `return` and braces to make it more concise.

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionOverview.utils.test.ts
```diff
@@ -0,0 +1,219 @@
...
+      totalInvocationCount: 23,
+      totalWarningCount: 4,
+      totalErrorCount: 5,
+    })
+  })
+
+  it('builds segmented button classes and chart range labels', () => {
+    expect(getSegmentedButtonClassName(0, 4)).toBe('rounded-tr-none rounded-br-none')
+    expect(getSegmentedButtonClassName(1, 4)).toBe('rounded-none')
+    expect(getSegmentedButtonClassName(3, 4)).toBe('rounded-tl-none rounded-bl-none')
```
Ziinc, line 91: asserting on styling increases brittleness of tests.

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionOverview.utils.test.ts
```diff
@@ -0,0 +1,219 @@
...
+      )
+    ).toEqual({
+      start: dayjs('2026-03-20T10:00:00.000Z').format('MMM D, h:mma'),
+      end: dayjs('2026-03-20T11:00:00.000Z').format('MMM D, h:mma'),
+    })
+    expect(getChartTimeRangeLabels([], 'MMM D')).toBeUndefined()
+    expect(formatChartTimestamp('2026-03-20T10:00:00.000Z', 'MMM D, h:mma')).toBe(
+      dayjs('2026-03-20T10:00:00.000Z').format('MMM D, h:mma')
+    )
+  })
```
Ziinc, line 106: testing implementation at this point.

apps/studio/components/interfaces/Functions/EdgeFunctionOverview/EdgeFunctionOverview.utils.ts
```diff
@@ -0,0 +1,297 @@
...
+  rows.map((row) => ({
+    timestamp: String(row.timestamp ?? ''),
+    success_count: toNumber(row.success_count),
+    redirect_count: toNumber(row.redirect_count),
+    client_err_count: toNumber(row.client_err_count),
+    server_err_count: toNumber(row.server_err_count),
+    avg_execution_time: toNumber(row.avg_execution_time),
+    max_execution_time: toNumber(row.max_execution_time),
+    avg_cpu_time_used: toNumber(row.avg_cpu_time_used),
+    max_cpu_time_used: toNumber(row.max_cpu_time_used),
```
Ziinc, line 175: is toNumber necessary? this could be done in the query with a coalesce and will improve performance by not needing to iterate over each data point.

Ziinc: if null values are an issue, modifying the logflare endpoint is possible (and better).
 Separate kaizen issue since this isn't related to PR feature. @jordienr

Ziinc (approved):
Image

would advise the timestamp display of the embedded logs to be more close to ISO 8601 format.
Making this embedded logs component into a reusable widget like how reports charts work will also be good for other areas

awaseem (approved):
Minor questions about post hog flag

PR #44018, opened 2026-03-20
Reinstate source param in new connect sheet to select read replicas

joshenlim:
## Context

Adds a "source" param in the new Connect sheet to allow selecting read replicas - this was present in the old Connect dialog

image

Selecting sources will update the connect steps at the bottom - connection string, host etc


charislam (approved):
I think unit tests need to be fixed but otherwise looks good

PR #44025, opened 2026-03-20
Studio SQL export emits valid PostgreSQL literals

### LINKED ISSUE #44024: Studio "Copy as SQL" / "Export as SQL" emits invalid PostgreSQL literals
## Describe the bug

Studio's table editor can generate invalid SQL when using `Copy as SQL` or `Export as SQL`.

I hit this on `storage.buckets`, where the generated output included values like:

```sql
'true'
'false'
'10485760'
ARRAY["image/*"]
```

Those are not correct PostgreSQL literals for this case.

## Expected behavior

The generated SQL should emit proper Postgres literals, e.g.:

```sql
true
false
10485760
ARRAY['image/*']
```

## Actual behavior

The SQL exporter currently wraps booleans and numbers as string literals, and emits text array elements with double quotes inside `ARRAY[...]`.

## Reproduction

1. Open Studio table editor
2. Pick a table with boolean, numeric, or `text[]` columns
3. Select one or more rows
4. Use `Copy as SQL` or `Export as SQL`
5. Run the generated SQL in Postgres

## Likely source

This appears to come from the shared formatter in:

- `apps/studio/components/interfaces/TableGridEditor/TableEntity.utils.ts`

That formatter is used by both row copy and full row export paths.

## Notes

A small fix here would be to:
- emit booleans and numbers as raw literals
- emit `text[]` members as SQL string literals inside `ARRAY[...]`

voiys:
## Summary
- fix the Studio SQL row formatter to emit raw boolean and numeric literals instead of quoted strings
- serialize text array members as SQL string literals inside `ARRAY[...]`
- escape fallback string formats outside `text`/`varchar`, and add regression coverage for that path

Closes #44024

## Test plan
- [x] Ran a direct `tsx` smoke against `formatTableRowsToSQL()` for the `storage.buckets` case and confirmed it now emits `true`, `false`, `10485760`, and `ARRAY['image/*']`
- [x] Ran focused formatter smokes for JSON escaping, text arrays, and fallback string formats like `citext`
- [x] `pnpm --filter studio test -- TableEntity.utils.test.ts`

Note: the targeted Vitest run is still blocked in this environment before the test executes (`localStorage.getItem is not a function`).


alaister (approved):
LGTM!

- Thanks, @voiys!

PR #44032, opened 2026-03-20
fix: suppress misleading toast when navigating away from a deleted project

charislam:

YES

Bug fix

When a user deletes a project, the `useProjectDetailQuery` in `RouteValidationWrapper` returns an error (HTTP 404) because the project no longer exists. This triggers the `isErrorProject` effect, which shows a "You do not have access to this project" toast and redirects to the home page — even though the user intentionally deleted the project.

## What is the new behavior?

The error's HTTP status code is checked before showing the toast. A 404 response (project not found / deleted) silently redirects without showing the misleading access-denied toast. The toast is still shown for other error codes (e.g. 403 Forbidden), where the "no access" message is accurate.

## Additional context

Resolves FE-2831


awaseem (approved):
LGTM

PR #44037, opened 2026-03-21
fix: reset identity sequences after CSV import to prevent unique key violations

majiayu000:

YES

Bug fix

When importing data via CSV into an existing table with identity columns, PostgreSQL sequences are not updated to reflect the imported values. This causes unique key violations on subsequent inserts because the sequence still starts from 1 (or its previous value), conflicting with the imported IDs.

Fixes #43993

## What is the new behavior?

After a successful CSV import into an existing table, the identity column sequences are now reset to the maximum value in the table using `setval()`. This mirrors the existing behavior in the `createTable` flow (SidePanelEditor.utils.tsx) which already correctly handles this case.

The sequence reset is non-blocking — if it fails, the import still succeeds (data is already inserted) and a warning is logged.

## Additional context

The fix reuses the existing `getUpdateIdentitySequenceSQL` utility from `@supabase/pg-meta` which is already battle-tested in the table creation flow. The only difference is that `onImportData` works with `RetrieveTableResult` columns (snake_case `is_identity`) rather than the internal editor format (camelCase `isIdentity`).


7ttp (commented):
closing in favour of: https://github.com/supabase/supabase/pull/45076

PR #44044, opened 2026-03-21
fix: only show rate limit reduction warning when disabling active SMTP

mohithshuka:

YES

Bug fix

When the SMTP form is dirty (e.g. user edits any field) and custom SMTP is not currently enabled, the footer incorrectly shows the message "Rate limit for sending emails will be reduced to 2 after disabling custom SMTP" — even though SMTP was never active in the first place.

Fixes #43800

## What is the new behavior?

The "rate limit will be reduced to 2" message now only appears when the user is actively toggling OFF an already-enabled custom SMTP configuration. If SMTP was never enabled, no misleading message is shown.

## Additional context

The fix is a single condition change in `SmtpForm.tsx`:

- Before: `) : (`
- After: `) : isSmtpEnabled(authConfig) ? (` with a `null` fallback

This ensures the warning message is only shown when it is actually relevant — i.e. when disabling a previously active SMTP setup.


themavik (commented):
Rate limit warning only when turning off active custom SMTP (isSmtpEnabled branch). nit: diff removes the final newline on SmtpForm.tsx; keep EOF newline.

dnywh:
Thanks for contributing to Supabase! Unfortunately the current diff on this PR no longer includes the conditional change described in the PR body, so I can’t verify it as a fix for the issue in its present state. 

I’m going to close this anyway as the earlier #43914 fixes #43800.

PR #44048, opened 2026-03-22
fix: include view options (security_invoker) in view definition display 

### LINKED ISSUE #35823: View definition displayed in Dashboard table editor missing critical "security_invoker" option
# Bug report

## Describe the bug

When you view the definition of a "security_invoker" view in the Supabase Dashboard, a "CREATE OR REPLACE VIEW" command is displayed with the critical security_invoker view setting missing.

This makes it easy for a developer, who opens this view definition to update the view in the SQL editor, to accidentally change the view to run with postgres permissions, which for a view in the public schema removes all RLS protection.

## To Reproduce

1. Create a "security invoker" view:
```
CREATE VIEW public.exposed_api WITH (security_invoker = TRUE) AS
SELECT * FROM public.rls_protected_table;
```
2. Go to the view in the Supabase Dashboard table editor
3. Click "Definition" in the bottom right

The definition shows a "create view " statement that is missing "WITH (security_invoker = TRUE)"

5. Click "Open in SQL Editor"
6. Change "create view" to "create or replace view" 
7. Click Run

The view is changed to a security definer view, and an anonymous user can now view the full contents of the protected table.

## Expected behavior

I expect to see "WITH (security_invoker = TRUE)" in the view definition.

## Proposed fix

You can get the full view definition, including options, using:

```
    'create or replace view ' ||
    quote_ident(n.nspname) || '.' || quote_ident(c.relname) ||
    case
        when c.reloptions is not null and array_length(c.reloptions, 1) > 0 then
            ' with (' || array_to_string(c.reloptions, ', ') || ')'
        else
            ''
    end ||
    e' as\n' ||
    pg_get_viewdef(c.oid, true)
```

I don't know the Supabase code well, but I suspect this is the code that generates the view definition:

https://github.com/supabase/supabase/blob/d8fe840627d83f23a65f36c4e186fff0403f3bef/apps/studio/components/interfaces/TableGridEditor/TableDefinition.tsx#L34

This executes the query defined here:

https://github.com/supabase/supabase/blob/d8fe840627d83f23a65f36c4e186fff0403f3bef/apps/studio/data/database/view-definition-query.ts#L16-L26

I also found code in "getEntityDefinitionsSql" that gets the view definition, but I don't know how this is used:

https://github.com/supabase/supabase/blob/d8fe840627d83f23a65f36c4e186fff0403f3bef/apps/studio/data/database/entity-definitions-query.ts#L28-L29

The "getEntityDefinitions" function already includes code to get options for regular table definitions here:

https://github.com/supabase/supabase/blob/d8fe840627d83f23a65f36c4e186fff0403f3bef/apps/studio/data/database/database-table-definition.ts#L214-L220

TheSeydiCharyyev:
## Description

  View definitions in the Dashboard Table Editor were missing `WITH (...)` options like `security_invoker`. When a user clicked "Open in SQL Editor" and re-executed the definition as
  `CREATE OR REPLACE VIEW`, the view silently lost its `security_invoker` setting, removing RLS protection.

  ## Related Issue

  Fixes #35823

  ## Changes Made

  - **`packages/pg-meta/src/sql/studio/database/views.ts`** — added `reloptions` to the SQL query and return `view_options` alongside `definition`
  - **`packages/pg-meta/src/sql/studio/database/table-definition.ts`** — added `WITH (...)` clause to entity definitions SQL for views and materialized views
  - **`apps/studio/data/database/view-definition-query.ts`** — updated return type to include `viewOptions`
  - **`apps/studio/components/interfaces/TableGridEditor/TableDefinition.tsx`** — include view options in the `CREATE VIEW` prepend
  - **`apps/studio/components/interfaces/TableGridEditor/ViewEntityAutofixSecurityModal.tsx`** — updated to use new data structure

  ## Checklist

  - [x] Linter passes (0 errors)
  - [x] No new dependencies
  - [x] Backwards compatible (views without options display unchanged)

## Summary by CodeRabbit

* **New Features**
  * Views now include optional WITH (...) clauses and view option formatting throughout the UI.
  * Utilities added to normalize, display and merge view options (affects view creation/editing and dialogs).

* **Bug Fixes**
  * View definition fetch and editors now surface view options alongside definitions.
  * Security dialogs and SQL previews accurately show merged view options (e.g., adding security_invoker).


Dlove123:
## 🎯 Claiming this Bounty!

**Plan**:
1. Fix view options display with security_invoker
2. Add tests
3. Submit PR

## 💰 Payment Information

**PayPal**: 979749654@qq.com
**ETH**: 0x31e323edC293B940695ff04aD1AFdb56d473351D
**GitHub**: Dlove123

**Quality Commitment**:
- ✅ Code Review + Tests

Let's fix this! 🔧

Dlove123:
## 💰 Payment Information

**PayPal**: 979749654@qq.com
**ETH**: 0x31e323edC293B940695ff04aD1AFdb56d473351D
**GitHub**: Dlove123

themavik (commented):
nit: View + table definition queries both run now instead of picking one branch—worth confirming that extra round-trip is OK for large schemas.

clean fix wiring `view_options` through SQL and keeping `prepend` in the `useMemo` deps.

TheSeydiCharyyev:
Thanks for the review! Actually, both hooks already existed before this PR and use the `enabled` option:

  - `useViewDefinitionQuery` has `enabled: isViewLike(entity)`
  - `useTableDefinitionQuery` has `enabled: isTableLike(entity)`

  Since `isViewLike` and `isTableLike` are mutually exclusive, only one query fires at a time — React Query skips execution when `enabled` is `false`. My change only affects how the results are destructured, not whether the queries run. So there's no extra round-trip.

TheSeydiCharyyev:
@avallete Friendly reminder — this PR has been waiting for review since March 22. It fixes a security-related issue where view definitions are missing the security_invoker option, which can lead to silent RLS bypass. Would appreciate a review when you have a chance. Thanks!

apps/studio/components/interfaces/TableGridEditor/ViewEntityAutofixSecurityModal.tsx
```diff
@@ -97,7 +97,7 @@ export default function ViewEntityAutofixSecurityModal({
             {isLoading && <GenericSkeletonLoader />}
             {isSuccess && (
               <SimpleCodeBlock>
-                {`create view ${table.schema}.${table.name} with (security_invoker = on) as\n ${data}`}
+                {`create view ${table.schema}.${table.name} with (security_invoker = on) as\n ${data?.definition}`}
```
avallete: **question**

Should that be changed ? If a view currently has other options say check_option=local or security_barrier=true the "after" preview would silently drop them, showing a statement that would actually lose those settings. The should the after-preview merge security_invoker = on into the existing options ?

Something like:
```ts
const afterOptions = mergeViewOptions(data?.viewOptions, 'security_invoker = true');
```

avallete: pg_class.reloptions stores values as key=value (e.g., security_invoker=true). We later directly interpolate them.

PostgreSQL's own pg_dump and standard docs use the form security_invoker = on (spaces, on/off instead of true/false).

This won't break anything , Postgres accepts both, but it creates a visual inconsistency between the "before" panel (raw reloptions format) and the hardcoded "after" panel (security_invoker = on). If cosmetic parity matter, we might wanna normalize those to both to` security_invoker = on`.

apps/studio/components/interfaces/TableGridEditor/TableDefinition.tsx
```diff
@@ -53,17 +53,25 @@ export const TableDefinition = ({ entity }: TableDefinitionProps) => {
...
-  const { data: definition, isLoading } = isViewLike(entity) ? viewResult : tableResult
+  const { data: viewData, isLoading: isViewLoading } = viewResult
+  const { data: tableData, isLoading: isTableLoading } = tableResult
+
+  const isLoading = isViewLike(entity) ? isViewLoading : isTableLoading
+  const definition = isViewLike(entity) ? viewData?.definition : tableData
+
+  const viewOptions = isViewLike(entity) && viewData?.viewOptions
+    ? ` with (${viewData.viewOptions})`
+    : ''
```
avallete: **nitpick**

The same ` with (${viewOptions})` interpolation pattern appears in:

TableDefinition.tsx
ViewEntityAutofixSecurityModal.tsx
table-definition.ts (SQL version)

Might consider making it an utility, especially if we handle normalization for them:
```ts
function formatViewOptionsClause(opts: string | null): string {
  return opts ? ` with (${opts})` : '';
}
```

avallete (approved):
This LGTM, only a few nitpicks, will need group Dasbhoard approval to get this trough as well.

apps/studio/components/interfaces/TableGridEditor/ViewEntityAutofixSecurityModal.tsx
TheSeydiCharyyev: Good catch! Fixed — the "after" preview now merges `security_invoker = on` into the existing options instead of replacing them. Added a `mergeViewOptions()` utility that preserves any other options like `check_option` or `security_barrier`.

TheSeydiCharyyev: Agreed. Added `normalizeViewOptions()` that converts the raw `pg_class.reloptions` format (`security_invoker=true`) to the pg_dump-compatible format (`security_invoker = on`). Both panels now show the same normalized format.

apps/studio/components/interfaces/TableGridEditor/TableDefinition.tsx
TheSeydiCharyyev: Done — extracted to `apps/studio/lib/view-options.ts` with three functions: `normalizeViewOptions()`, `formatViewOptionsClause()`, and `mergeViewOptions()`. Both `TableDefinition.tsx` and `ViewEntityAutofixSecurityModal.tsx` now use them. The SQL-level formatting in `table-definition.ts` stays as-is since it runs server-side.

avallete (approved):

7ttp (commented):
i think this has been superseded in: 
- https://github.com/supabase/supabase/pull/44936

which also closed the linked issue here :) 
marking this as closed for now!

Thanks for ur efforts towards contributing to Supabase!! 💚

PR #44064, opened 2026-03-23
Use TimestampInfo in edge functions created at column + use relative label

joshenlim:
## Context

Very tiny one - just uses TimestampInfo here
image


alaister (approved):
LGTM!

- Tested on preview

PR #44072, opened 2026-03-23
Fix missing edge when deploying a new RR, and add some small qol tidbits

joshenlim:
## Context

Edge between replica and primary DB was missing when the replica is first initialized, so this PR primarily fixes that.

## Other changes
Also just adding in some tiny QoL nudges
- Add replication lag if its a read replica
- Add a loader edge label if the target is coming up
- Add a "X" edge label if the target status failed

image

image


jordienr (approved):
amazin

PR #44094, opened 2026-03-23
Fix mcp url

saltcod:
fix mcp url in dashboard connect side panel

### DIFF
--- a/apps/studio/components/interfaces/ConnectSheet/useMcpUrl.ts
+++ b/apps/studio/components/interfaces/ConnectSheet/useMcpUrl.ts
@@ -9,7 +9,7 @@ export function useMcpUrl(
   projectKeys: StepContentProps['projectKeys']
 ): string {
   const readonly = Boolean(state.mcpReadonly)
-  const baseUrl = IS_PLATFORM ? 'https://mcp.supabase.com' : (projectKeys.apiUrl ?? '')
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/ConnectSheet/useMcpUrl.ts
```diff
@@ -8,11 +9,13 @@ export function useMcpUrl(
   state: StepContentProps['state'],
   projectKeys: StepContentProps['projectKeys']
 ): string {
+  const { ref: projectRef } = useParams()
   const readonly = Boolean(state.mcpReadonly)
-  const baseUrl = IS_PLATFORM ? 'https://mcp.supabase.com' : (projectKeys.apiUrl ?? '')
+  const baseUrl = IS_PLATFORM ? 'https://mcp.supabase.com/mcp' : (projectKeys.apiUrl ?? '')
```
mattrossman: At a glance, this `useMcpUrl` hook seems to be duplicating URL construction logic from the getMcpUrl helper we use on the docs / in the old Connect dialog. Could this be simplified to use that function to prevent this sort drift in the future?

https://github.com/supabase/supabase/blob/3f81d47723a209bf2d3f85d5cdcd2ea95f769b68/packages/ui-patterns/src/McpUrlBuilder/utils/getMcpUrl.ts#L18-L55

awaseem (approved):
Tested on staging and self hosted (not effected)!

PR #44098, opened 2026-03-23
docs: Update Functions quickstart guide

ChrisChinchilla:

YES

## Summary by CodeRabbit

* **Documentation**
  * CLI-first quickstart rewritten: Docker/local runtime required, explicit create/enter directory steps.
  * Updated starter function template and optional VSCode runtime note.
  * Guidance to secure endpoints at creation with an auth flag.
  * Streamlined local testing and curl examples to use 127.0.0.1 and apikey header with publishable keys.
  * Simplified deployment to a single deploy command; notes API-deploy fallback.


apps/docs/content/guides/functions/quickstart.mdx
```diff
@@ -47,85 +51,67 @@ Within your project, generate a new Edge Function with a basic template:
 supabase functions new hello-world
 ```
 
+{/* TODO: Link to parameter documentation */}
+
+<Admonition type="tip" label="Secure your function with Supabase Auth">
+
+When an HTTP request is sent to Edge Functions, you can use Supabase Auth to secure endpoints. By default, the `supabase functions new` command adds handling a valid publishable or secret key to the basic template. However, you can change this behavior with the `--auth` flag when creating a new function.
```
ChrisChinchilla, line 58: Would be good to link to reference docs, but they don't list parameters… That's for docs team to fix

kallebysantos: Ahh sure! the CLI docs wasn't updated yet.

kallebysantos (approved):

apps/docs/content/guides/functions/quickstart.mdx
```diff
@@ -178,39 +162,28 @@ After this step, you should have your local project authenticated and linked to
...
 ```bash
-supabase functions deploy hello-world --use-api
+supabase functions deploy
 ```
 
-</Admonition>
+<Admonition type="note" label="Docker not required">
 
-If you want to skip JWT verification, you can add the `--no-verify-jwt` flag for webhooks that don't need authentication:
+The CLI automatically falls back to API-based deployment if Docker isn't available. You can also explicitly use API deployment with the `--use-api` flag:
```
tomaspozo, line 181: wondering if this comment is needed considering the user is already using the local environment and for that they need Docker (we mention that on the prerequisites)

tomaspozo (approved):
This is a great revamp to the doc!

apps/docs/content/guides/functions/quickstart.mdx
kallebysantos: The `--use-api` its important to mention, perhaps it could be adapted from "fallback Docker ins't available" to something like "In case you would like to bundle from cloud"....

But I think its fine the way it is already.

ChrisChinchilla: What do you mean by "bundle from cloud"? We can also move this to other guides instead of the quickstart, which generally should be an opinionated "one path"

kallebysantos: When setting `--use-api` the bundle process happens on cloud — we have a dedicated bundle service for it.

Otherwise it will spin up a local docker container to generate the bundle locally to then get uploaded.

--

But for a quickstart guide I think it's already fine

ChrisChinchilla:
@kallebysantos Righto, shall we merge?

PR #44105, opened 2026-03-23
feat(billing): use Stripe AddressElement for billing address form

kanishkdudeja:
### Summary

This PR refactors the billing address form to use Stripe AddressElement on both the org billing page and the missing-address modal, adding Google Maps autocomplete and country-aware state/province inputs.

### Testing

- Address element renders on org billing page, autocomplete works, and state/province options update per country.
- Existing billing address data loads correctly into the form.
  - Need to verify this with invalid country/state as well.
- Save enables only when the form is dirty after editing address or tax ID fields.
- Cancel restores both address and tax ID to the original values.
- Submitting with missing required address fields shows a validation toast.
- Tax ID options are filtered by the selected billing country.
- Tax-ID-only edits submit successfully without interacting with the address form.
- Without billing write permission, address fields are read-only.
- Update billing address modal behaves the same as the org billing page.
- Address element appearance looks correct in both light and dark mode.
- Payment method form remains unchanged and still uses floating labels.

## Summary by CodeRabbit

* **New Features**
  * Billing form now uses Stripe Address Element via Stripe Elements with an embedded custom font and theme-aware appearance; tax-ID entry remains separate.

* **Bug Fixes**
  * Stronger address validation and submit gating, improved dirty-state tracking, explicit save confirmation and reset behavior, and clearer error toasts on update failures.

* **Tests**
  * Expanded tests covering hook behavior, address element validation, submit gating, payloads, reset, and permission scenarios.

* **Style**
  * Adjusted billing email form spacing/layout.


apps/studio/components/interfaces/Billing/Payment/Payment.utils.ts
```diff
@@ -1,4 +1,23 @@
...
+import { CUSTOM_FONT_BOOK_DATA_URL } from 'fonts/stripe-fonts'
+
+/**
+ * Custom font for Stripe Elements iframes.
+ * Stripe renders inside an iframe that can't access the parent page's CSS variables or fonts,
+ * so we pass the font explicitly (Stripe requires https:// or data:// sources).
+ * In production (HTTPS), we use a static URL. In dev (HTTP), we fall back to a base64 data URL.
+ */
+const fontSrc = process.env.NEXT_PUBLIC_SITE_URL?.startsWith('https')
+  ? `url(${process.env.NEXT_PUBLIC_SITE_URL}/fonts/CustomFont-Book.woff2)`
```
kanishkdudeja: @supabase/frontend 

Is it okay to serve our custom Studio font at a static path like https://supabase.com/fonts/CustomFont-Book.woff2 ?

I'm refactoring the existing billing address form to use Stripe address element, which renders in a cross-origin iframe. Stripe requires font sources to be absolute HTTPS or data: URLs - it can't access the parent page's CSS or _next/static/ assets (since they have build-time hashes in their filenames).

staaldraad (approved):

awaseem (approved):
LGTM

apps/studio/components/interfaces/Organization/BillingSettings/BillingCustomerData/BillingCustomerData.tsx
```diff
@@ -117,8 +114,35 @@ export const BillingCustomerData = () => {
     },
   })
 
+  useEffect(() => {
+    addressElementRef.current = null
```
ignaciodob, line 132: I think we need reset the `isAddressDirty` here (or after submitting - not sure). After submitting it seems `isAddressDirty` stays true which allows you to save/cancel again without touching the form. 

https://github.com/user-attachments/assets/60a463b2-7fe5-466f-9be1-7bb4e1b59d8d

ignaciodob (commented):
Some minor details I noticed while testing this. Might not be worth it to address, but I'll add them here.

1. The padding between the Email Recipient section and the Address section does not match
Image

2. The input highlight colors are not consistent between the two sections

|  Address  |  Email Recipient section  |
| Image  | Image  |

apps/studio/components/interfaces/Organization/BillingSettings/BillingCustomerData/BillingCustomerData.tsx
kanishkdudeja: Thanks, I've addressed all the feedback 👍

apps/studio/components/interfaces/Organization/BillingSettings/BillingCustomerData/BillingCustomerData.tsx
```diff
@@ -54,49 +61,56 @@ export const BillingCustomerData = () => {
...
 
-        await updateTaxId({ slug, taxId: data.tax_id })
+        try {
+          await updateTaxId({ slug, taxId: data.tax_id })
+        } catch (error) {
+          toast.error(
+            `Failed updating tax ID: ${error instanceof Error ? error.message : 'Unknown error'}`
+          )
+          throw error
+        }
```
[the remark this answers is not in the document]
kanishkdudeja: This has already been addressed

PR #44119, opened 2026-03-24
Fix/build issue react flow

joshenlim:
Not sure how i overlooked this one sorry 🤦 
was importing from the wrong library


apps/studio/components/interfaces/Database/Replication/ReplicationDiagram/Edges.tsx
```diff
@@ -1,13 +1,21 @@
+import { BaseEdge, EdgeLabelRenderer, getSmoothStepPath, type EdgeProps } from '@xyflow/react'
```
joshenlim, line 1: main fix is here

ivasilov (approved):

djhi (approved):

PR #44120, opened 2026-03-24
chore: remove TypeScript build error ignoring from next.config.js in `studio`

ivasilov:
This pull request makes a small configuration change to the Next.js application. The TypeScript build option that allowed build errors to be ignored has been removed from `next.config.js`. This means type errors will now cause the build to fail, improving code quality by catching issues earlier.

- Removed the `typescript.ignoreBuildErrors` option from `next.config.js`, so builds will now fail if there are TypeScript errors.


djhi (approved):

ivasilov:
It's not making the build faster.

PR #44122, opened 2026-03-24
fix(billing): clarify service restriction reset delay after billing cycle

pamelachia:
## Summary

Customers were upgrading to Pro to clear service restrictions, unaware that restrictions aren't lifted immediately when the billing cycle resets — there's a delay. This adds a clarifying note to both the docs FAQ and the Studio restricted alert so users know what to expect.

Note: upgrading your plan or disabling spend cap *does* lift restrictions immediately (unchanged). The delay only applies to the billing-cycle-reset path for usage-limit violations.

## Changes

- `billing-faq.mdx`: Removes the implication that restrictions clear the moment the billing period resets; adds "Note that there may be a short delay after your billing period resets before restrictions are fully lifted"
- `Restriction.tsx`: Updates the restricted alert to make clear that upgrading lifts restrictions immediately, while the billing-reset path may have a delay

## Testing

To test the Studio banner, find or create an org with `restriction_status === 'restricted'`, or temporarily hardcode `shownAlert = 'restricted'` in `Restriction.tsx` to preview:

- [x] Restricted alert reads clearly and distinguishes immediate (upgrade) vs delayed (billing reset) paths
- [x] Docs FAQ answer under "How can I remove restrictions" reads correctly with the new caveat

## Linear
- fixes GROWTH-704

### DIFF
--- a/apps/docs/content/guides/platform/billing-faq.mdx
+++ b/apps/docs/content/guides/platform/billing-faq.mdx
@@ -140,7 +140,7 @@ The Fair Use Policy is generally applied to all projects of the restricted organ
 
 To remove restrictions, you will need to address the issue that caused the restriction. This could be reducing your usage, paying overdue invoices, updating your payment method, or any other issue that caused the restriction. Once the issue is resolved, the restriction will be lifted.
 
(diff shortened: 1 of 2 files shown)


alaister (approved):
LGTM!

PR #44127, opened 2026-03-24
chore(billing): make update billing address modal non-dismissable

kanishkdudeja:
### Summary

This PR removes the visible close button from the billing address required modal to make it non dismissable, since the majority of users on paid orgs are still not filling their billing address.

The modal already prevents dismissal via Esc and outside/backdrop interaction, and this change preserves that behavior. The existing submit flow is unchanged, so the modal still closes after a successful billing address update.

### Manual testing:

1. Open Studio in a state where the billing address required modal appears.
2. Confirm the top-right close button is no longer shown.
3. Press Esc and verify the modal remains open.
4. Click outside the modal / on the backdrop and verify the modal remains open.
5. Submit a valid billing address and verify the modal closes after a successful save.

For testing whether an update successfully closes the modal, you can:

- You can create a free org without a billing address
- You will need to tweak this logic (on the frontend)

```
const shouldShow = Boolean(
  IS_PLATFORM &&
  // showMissingAddressModal &&
  org &&
  // org.plan.id !== 'free' &&
  org.organization_missing_address &&
  !org.billing_partner &&
  permissionsLoaded &&
  canBillingWrite
)
```

- Tweak the backend logic which computes `organization_missing_address` to exclude free orgs

### DIFF
--- a/apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
+++ b/apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
@@ -128,6 +128,7 @@ export function UpdateBillingAddressModal() {
     >
       <DialogContent
         size="medium"
+        hideClose
         onInteractOutside={(e) => e.preventDefault()}
         onEscapeKeyDown={(e) => e.preventDefault()}
       >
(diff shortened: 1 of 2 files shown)


joshenlim (approved):
Verified locally! (Couldn't trigger the modal on staging)
Just pushed to fix the failing unit tests

kanishkdudeja:
Thank you so much @joshenlim! 🙇

PR #43951, opened 2026-03-19
Chore/shift manual queries into pg meta 03

joshenlim:
## Context

Shifting more dashboard queries into pg-meta so that we centralize all manually written queries in one place
Having them in packages/pg-meta also allows us to write tests for them

## To test

Just needs a smoke test on
- Table Editor
  - Fetching entities
  - Viewing definition 
- SQL Editor
  - View ongoing queries
  - Abort queries
- Integrations
  - Queues
- Database
  - Migrations
  -Triggers (Updating)

PR #44128, opened 2026-03-24
fix: webhook regression

### LINKED ISSUE #44099: Webhook URL not saving correctly – reverts to incomplete URL after saving
I am experiencing an issue with webhooks in Supabase. I had an existing (old) webhook configured in my project. When I try to edit the webhook and change only the URL, I click "Save" and Supabase shows that the webhook is saved successfully. However, when I open the webhook again, the URL has reverted to a different/old value instead of the one I just saved.

This caused issues in production because the webhook was still sending requests to the old URL even though the dashboard showed it was saved.

To Reproduce

Go to Supabase Dashboard → Database → Webhooks
Open an existing webhook
Change the webhook URL
Click Save
Reopen the webhook
The URL is different from the one that was saved (reverted to old URL)

Expected behavior
The webhook URL should remain the same as the one entered after clicking Save.

Actual behavior
The webhook URL changes back to a previous/incorrect URL after saving.

Workaround
Deleting the webhook and creating a new one with the correct URL works. The issue only happens when editing an existing webhook.

Please let me know if you need any logs or additional information.

7ttp:
## TL;DR

Fixes webhook update regression when editing url,
 causes values like headers etc to disappear

## Before:

https://github.com/user-attachments/assets/3cb93640-79f2-45a8-930f-e1e4c5067476

## After:

https://github.com/user-attachments/assets/60772c0b-800f-4a14-9d08-8e4b172035cb

## Related

- closes https://github.com/supabase/supabase/issues/44099
- ig the regression was introduced in https://github.com/supabase/supabase/pull/43951


awaseem (approved):
LGTM

awaseem:
Thanks for the E2E test

steve-chavez (approved):
LGTM

PR #44132, opened 2026-03-24
feat(billing): show billing address modal to users other than owners/admins

kanishkdudeja:
### Summary

This PR updates the billing address required modal so it also shows for users with billing read access but without billing write access, such as developers and users with read-only acess.

For users with BILLING_WRITE, the existing blocking behavior remains in place:

- the modal has no visible close button
- it closes only after a successful billing address update

For users with BILLING_READ but not BILLING_WRITE, the modal now:

- the modal includes a visible close button
- no editable billing form is shown
- the copy tells the user to ask an organization administrator or owner to update the billing address

Screenshot 2026-03-24 at 7 18 49 PM

#### Admins/owners

1. As a user with BILLING_WRITE, open an org that is missing a billing address.
2. Confirm the modal appears with the billing address form.
3. Confirm there is no visible close button.
4. Submit a valid billing address and verify the modal closes after a successful save.

For testing whether an update successfully closes the modal, you can:

-  You can create a free org without a billing address
- You will need to tweak this logic (on the frontend)

```
const shouldShow = Boolean(
  IS_PLATFORM &&
  // showMissingAddressModal &&
  org &&
  // org.plan.id !== 'free' &&
  org.organization_missing_address &&
  !org.billing_partner &&
  permissionsLoaded &&
  canViewModal
)
```

- Tweak the backend logic which computes `organization_missing_address` to exclude free orgs

#### Developers/users with readp-only access

1. As a user with BILLING_READ but not BILLING_WRITE, open an org that is missing a billing address.
2. Confirm the modal appears with informational copy only and no editable form.
3. Click the close button and verify the modal dismisses.

### DIFF
--- a/apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
+++ b/apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
@@ -158,7 +158,8 @@ export function UpdateBillingAddressModal() {
                 Tax ID is also required.
                 <br />
                 <br />
-                Please ask an organization administrator or owner to update it.
+                Please ask an organization administrator or owner to update it as soon as possible
+                to avoid service restrictions.
               </>
             )}
           </DialogDescription>
--- a/apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
+++ b/apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
@@ -40,10 +40,18 @@ export function UpdateBillingAddressModal() {
   const { data: org } = useSelectedOrganizationQuery()
   const slug = org?.slug
 
-  const { can: canBillingWrite, isSuccess: permissionsLoaded } = useAsyncCheckPermissions(
+  const { can: canBillingRead, isSuccess: billingReadLoaded } = useAsyncCheckPermissions(
+    PermissionAction.BILLING_READ,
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/App/UpdateBillingAddressModal.tsx
```diff
@@ -128,45 +143,55 @@ export function UpdateBillingAddressModal() {
...
         <DialogHeader>
           <DialogTitle>Billing address required</DialogTitle>
           <DialogDescription>
-            Please provide a billing address for your organization. If you are a registered business
-            and have a Tax ID, please add your Tax ID too.
+            {canManageBillingAddress ? (
+              'Please provide a billing address for your organization. If you are a registered business and have a Tax ID, please add your Tax ID too.'
+            ) : (
+              <>
+                Your organization requires a billing address. If you are a registered business, a
```
ignaciodob, line 157: I think we can make this more urgent

kanishkdudeja: thanks, addressed in https://github.com/supabase/supabase/pull/44132/commits/d4011db3b2db6e0747744090f550c0857aa3e075

awaseem (approved):
LGTM

PR #44134, opened 2026-03-24
docs(auth): Add OAuth PKCE example for SSR

benogeorge:
This surfaces the PKCE OAuth example in the SSR client guide so SvelteKit users can
see the `redirect(data.url)` pattern directly in the docs.

It keeps the guidance in the SSR auth flow where readers already are, instead of
burying the callback behavior in a less visible partial.

Fixes #23618


Hallidayo:

PR #44137, opened 2026-03-24
fix: annoying and confusing bugs with queue operations in table editor

awaseem:

YES

  - Index mismatch on ADD then DELETE/EDIT: formatGridDataWithOperationValues was searching the original rows array for DELETE_ROW and EDIT_CELL_CONTENT operations, then using those indices on the modified formattedRows array (which had  been shifted by ADD_ROW's unshift). Both now search formattedRows directly.
  - Cross-table operation leaking: The entire operation queue was passed to formatGridDataWithOperationValues without filtering by the current table, causing pending ADD_ROW and DELETE_ROW operations from one table to appear in other  tables. Operations are now filtered by tableId before rendering.

### DIFF
--- a/apps/studio/components/grid/SupabaseGrid.tsx
+++ b/apps/studio/components/grid/SupabaseGrid.tsx
@@ -93,7 +93,9 @@ export const SupabaseGrid = ({
     if (!mounted) setMounted(true)
   }, [])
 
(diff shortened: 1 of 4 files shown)


joshenlim (approved):
Verified changes on preview!

PR #44154, opened 2026-03-25
fix(studio): keep editor browser titles in sync after renames

dnywh:

Bug fix that resolves DEPR-415.

Browser titles in the table/SQL editors can keep showing an old entity name after a rename.

## What is the new behavior?

- `EditorBaseLayout` now prefers the live tab label when composing the browser title entity, so existing stale tab metadata no longer wins.
- Updating a tab label now also syncs the persisted `metadata.name`, so renamed entities recover cleanly on reload instead of carrying stale names in local storage.
- Adds a regression test covering the stale-metadata case in the editor layout.


joshenlim (approved):
Verified changes on preview!

PR #44155, opened 2026-03-25
fix: keep recent items labels in sync after renames

dnywh:

Bug fix.

The `Recent items` list shown in the table editor empty-state can keep showing an old label after a table or SQL snippet is renamed.

That UI is mounted from:
- `/project/[ref]/editor`
- `/project/[ref]/editor/new`

## What is the new behavior?

- Re-adding an existing recent item now refreshes its label and metadata instead of only bumping the timestamp.
- Updating an open tab label now also updates the matching recent item, so rename flows stay aligned with the recent-items list.
- The fix is applied in the shared tabs store, so table and SQL editor tab state stay consistent even though the visible `Recent items` UI currently appears in the table editor empty-state.
- Adds a small store-level regression test for both cases.

| Before | After |
| --- | --- |
| test  Table Editor  Mallet  Toolshed  Supabase-DA49998A-FE09-4197-8EE7-5D5366FABDC7 | test  Table Editor  Mallet  Toolshed  Supabase-8F2B6CC9-0721-4348-A094-EE94D6919B08 |

### DIFF
--- a/apps/studio/state/tabs.test.ts
+++ b/apps/studio/state/tabs.test.ts
@@ -1,0 +1,62 @@
+import { beforeEach, describe, expect, it } from 'vitest'
+
+import { createTabsState } from './tabs'
+import { ENTITY_TYPE } from '@/data/entity-types/entity-type-constants'
+
+describe('tabs recent items', () => {
+  beforeEach(() => {
+    localStorage.clear()
+  })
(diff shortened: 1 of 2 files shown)


joshenlim (approved):
Verified changes on preview!

PR #44163, opened 2026-03-25
chore(studio): database tables UI improvements

dnywh:

UI improvements

- The database tables list and columns list use inconsistent page shells and table primitives
- The child columns page has weaker information hierarchy and row actions than the parent tables page
- Responsive column priority on the tables list does not reflect the most important data on smaller breakpoints
- Table actions and counts are harder to scan than they should be

## What is the new behavior?

- Both pages now use `PageLayout` with matching large-width content containers
- `ColumnList` now uses the latest `ui` Table primitives instead of the legacy cleaned-up-later table
- Both pages now show totals in a table footer
- `ColumnList` now uses a tiny filter input, case-insensitive filtering, inline descriptions under the name, and a primary `Edit` button with overflow actions
- `TableList` now has improved responsive column priority:
  - smallest breakpoint keeps `Rows`
  - `Columns` appears from `sm`
  - `Size` appears from `lg`
  - `Realtime Enabled` appears from `2xl`
- `TableList` now uses `View columns` as the CTA, removes the ambiguous icon from that CTA, restores the entity icon from `sm` upwards only, and tightens the name column on the smallest breakpoint only
- Boolean icon columns are right-aligned consistently, with the same Realtime icon tones applied to both `Realtime Enabled` and `Nullable`
- The columns detail page now uses breadcrumbs for navigation back to Tables instead of an inline back button

| Before | After |
| --- | --- |
| Tables  Database  Mallet  Toolshed  Supabase-0E0E3DE0-4EA1-407F-88D4-B85664D26D8E | Tables  Database  Mallet  Toolshed  Supabase-C8FC339C-E9DA-4ADB-8458-C7EFF55F2AEC |
| Tables  Database  Mallet  Toolshed  Supabase-FE9196A0-BEAF-4BA5-8A2C-06F934A62C38 | Tables  Database  Mallet  Toolshed  Supabase-36E93C1E-7943-4C98-8119-CAF48E2FE5BA |

### DIFF
--- a/apps/studio/components/interfaces/Database/Tables/ColumnList.tsx
+++ b/apps/studio/components/interfaces/Database/Tables/ColumnList.tsx
@@ -1,12 +1,6 @@
-import { PermissionAction } from '@supabase/shared-types/out/constants'
-import { noop } from 'lodash'
-import { Check, ChevronLeft, Edit, MoreVertical, Plus, Search, Trash, X } from 'lucide-react'
-import Link from 'next/link'
-import { useState } from 'react'
-
 import { PostgresColumn } from '@supabase/postgres-meta'
+import { PermissionAction } from '@supabase/shared-types/out/constants'
 import { useParams } from 'common'
-import Table from 'components/to-be-cleaned/Table'
 import AlertError from 'components/ui/AlertError'
 import { ButtonTooltip } from 'components/ui/ButtonTooltip'
 import { NoSearchResults } from 'components/ui/NoSearchResults'
@@ -15,18 +9,30 @@ import { isTableLike } from 'data/table-editor/table-editor-types'
 import { useAsyncCheckPermissions } from 'hooks/misc/useCheckPermissions'
 import { useSelectedProjectQuery } from 'hooks/misc/useSelectedProject'
 import { useIsProtectedSchema } from 'hooks/useProtectedSchemas'
+import { noop } from 'lodash'
+import { Check, Edit, MoreVertical, Plus, Search, Trash, X } from 'lucide-react'
+import { useState } from 'react'
 import {
   Button,
+  Card,
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
   DropdownMenuTrigger,
-  Input,
+  Table,
+  TableBody,
+  TableCell,
+  TableFooter,
+  TableHead,
+  TableHeader,
+  TableRow,
   Tooltip,
   TooltipContent,
   TooltipTrigger,
 } from 'ui'
+import { Input } from 'ui-patterns/DataInputs/Input'
 import { GenericSkeletonLoader } from 'ui-patterns/ShimmeringLoader'
+
 import { ProtectedSchemaWarning } from '../ProtectedSchemaWarning'
 
 interface ColumnListProps {
@@ -40,7 +46,7 @@ export const ColumnList = ({
   onEditColumn = noop,
   onDeleteColumn = noop,
 }: ColumnListProps) => {
-  const { id: _id, ref } = useParams()
+  const { id: _id } = useParams()
   const id = _id ? Number(_id) : undefined
 
   const { data: project } = useSelectedProjectQuery()
@@ -62,7 +68,9 @@ export const ColumnList = ({
   const columns =
     (filterString.length === 0
       ? (selectedTable?.columns ?? [])
-      : selectedTable?.columns?.filter((column) => column.name.includes(filterString))) ?? []
+      : selectedTable?.columns?.filter((column) =>
+          column.name.toLowerCase().includes(filterString.toLowerCase())
+        )) ?? []
 
   const { isSchemaLocked } = useIsProtectedSchema({ schema: selectedTable?.schema ?? '' })
   const { can: canUpdateColumns } = useAsyncCheckPermissions(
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Database/Tables/ColumnList.tsx
```diff
@@ -72,16 +80,13 @@ export const ColumnList = ({
...
-      <div className="flex items-center justify-between">
-        <div className="flex items-center space-x-4">
-          <Button asChild type="outline" icon={<ChevronLeft />} style={{ padding: '5px' }}>
-            <Link href={`/project/${ref}/database/tables`} />
-          </Button>
+      <div className="flex flex-col gap-2 lg:flex-row lg:items-center lg:justify-between">
+        <div className="w-full lg:w-52">
           <Input
-            size="small"
+            size="tiny"
```
dnywh, line 86: We use `tiny` everywhere else in this context.

apps/studio/components/interfaces/Database/Tables/ColumnList.tsx
```diff
@@ -108,115 +113,163 @@ export const ColumnList = ({
...
+                            className="max-w-md truncate text-foreground-lighter"
+                            title={column.comment}
+                          >
+                            {column.comment}
+                          </span>
+                        ) : null}
+                      </div>
+                    </TableCell>
+                    <TableCell>
+                      <code className="text-code-inline">{column.data_type}</code>
```
dnywh, line 197: Still not convinced we need `text-code-inline` but leaving that for another time.

apps/studio/components/interfaces/Database/Tables/ColumnList.tsx
```diff
@@ -108,115 +113,163 @@ export const ColumnList = ({
...
+                            tooltip={{
+                              content: {
+                                side: 'bottom',
+                                text: !canUpdateColumns
+                                  ? 'Additional permissions required to edit column'
+                                  : undefined,
+                              },
+                            }}
+                          >
+                            Edit
```
dnywh, line 229: Brought this out. Feels like we should have at least one common action outside the dropdown here.

apps/studio/components/interfaces/Database/Tables/TableList.tsx
```diff
@@ -384,80 +375,80 @@ export const TableList = ({
...
                             </div>
                           )}
                         </TableCell>
                         <TableCell>
                           <div className="flex justify-end gap-2">
-                            <Button
-                              asChild
-                              type="default"
-                              iconRight={<Columns size={14} className="text-foreground-light" />}
-                              className="whitespace-nowrap hover:border-muted"
```
dnywh, line 456: These classes were causing a strange hover state.

apps/studio/components/interfaces/Database/Tables/TableList.tsx
```diff
@@ -317,18 +306,20 @@ export const TableList = ({
...
                     Rows (Estimated)
                   </TableHead>
-                  <TableHead key="size" className="hidden text-right xl:table-cell">
+                  <TableHead key="size" className="hidden text-right lg:table-cell">
                     Size (Estimated)
                   </TableHead>
-                  <TableHead key="realtime" className="hidden xl:table-cell text-right">
+                  <TableHead key="realtime" className="hidden text-right 2xl:table-cell">
                     Realtime Enabled
                   </TableHead>
```
joshenlim, line 324: just wondering but should we just show this column irregardless? i find it odd that we're hiding columns based on viewport width

apps/studio/components/interfaces/Database/Tables/ColumnList.tsx
```diff
@@ -108,115 +113,163 @@ export const ColumnList = ({
...
+                            </DropdownMenuTrigger>
+                            <DropdownMenuContent side="bottom" align="end" className="w-32">
+                              <Tooltip>
+                                <TooltipTrigger>
+                                  <DropdownMenuItem
+                                    disabled={!canUpdateColumns || isSchemaLocked}
+                                    onClick={() => onDeleteColumn(column)}
+                                    className="space-x-2"
+                                  >
+                                    <Trash stroke="red" size={12} />
```
joshenlim: nit but lets remove the red stroke here for consistency

joshenlim (approved):
Changes LGTM although a couple of feedback that's not related to the changes here

- Am thinking that we shouldn’t hide columns based on viewport width for `TableList`? just show all columns instead
- There’s a lot more that we can do with the ColumnList table
  - We only show data type, format, and nullable
    - Am thinking data type isn’t useful to be showing as a column? format is more important perhaps
  - We could show more information up front as a summary, e.g
    - All constraints (not just nullable)
    - Marking columns that are primary keys
    - Marking columns that have foreign keys etc
-  UX feedback (probably a design pattern thing) - am wondering if the count here could be higher up (maybe below the header?)
   - imo i’d want to know how many columns the table has (or just the count information in general) without having to scroll all the way to the bottom
   Image

apps/studio/components/interfaces/Database/Tables/ColumnList.tsx
dnywh: Agree, thanks

apps/studio/components/interfaces/Database/Tables/TableList.tsx
dnywh: Fair!

PR #44172, opened 2026-03-25
fix: use entitlement check for project pausing

kevcodez:
Project pausing is available on platform plan, so we should not do a check based on Free Plan


ivasilov (approved):

joshenlim (approved):
Verified changes on preview! both free and paid plan

kanishkdudeja (approved):

PR #44173, opened 2026-03-25
fix(Docs): Apply retry strategy, improve error throwing

jeremenichelli:
In this PR, we add a plugin to retry Octokit requests to try to compensate for 500 errors from the GitHub Contents API.

Also, we noticed errors weren't thrown correctly when using this lib.


ChrisChinchilla (approved):
🐙

PR #44174, opened 2026-03-25
fix(project-list): remove duplicate resource exhaustion badge DEBUG-48

jordienr:
## Problem

Project cards on the home page showed resource exhaustion warnings twice: once as a badge in the card header, and again as an alert in the card footer. This created a confusing, redundant UI as seen when a project has disk IO, CPU, or other resource warnings active.

## Fix

Removed `resourceWarnings` from the footer `ProjectCardStatus` render in `ProjectCard.tsx`. Resource exhaustion state now only appears as the header badge (which already shows a tooltip with the description on hover). Project status indicators such as paused or restarting still appear in the footer since those do not depend on resource warnings.

Also removed the dead `showResourceExhaustionWarnings = false` flag from `ProjectCardStatus.tsx`, which was hardcoded to false and only gated description text that was no longer reachable.

## How to test

- Navigate to the org home page with a project that has an active resource warning (disk IO, disk space, CPU, memory, or read-only mode).
- Confirm the warning badge appears in the card header with the correct color (warning or critical).
- Confirm no duplicate alert appears in the card footer area.
- Confirm that paused or restarting projects still show the status alert in the footer.
- Confirm the table view still shows the badge correctly in the status column.

## Release Notes

**Style**
- Project status information now displays exclusively in the project card footer, removing redundant header display for a cleaner interface.
- Simplified resource warning notification logic for more consistent and straightforward warning presentation.


kemaldotearth (changes requested):
Think for exhaustion the bottom row message on cards works better than the pill. Otherwise looking tip top!

kemaldotearth (approved):
:shipit:

PR #44190, opened 2026-03-25
fix: cannot create nor edit foreign keys from schema visualiser

djhi:
## Problem

Users cannot create nor edit foreign keys when editing a table or a column from the schema visualiser

## Solution

The foreign key edition component is always reading the current table from the URL parameters. However, in the context of the schema visualiser, the parameter does not exist.

Allows to override it from the side panel editor context

## How to test

- Create two tables
- Go to the schema visualiser and edit one of the table
- Click the _Add foreign key relation_ and add a foreign key to the other table

- Edit one of the foreign key column
- You should be able to edit the relation


awaseem (approved):
LGTM tested on staging and self hosted

PR #44195, opened 2026-03-25
fix: hide tax id banner for billing partner orgs

ignaciodob:
### Changes
- (shouldFetch): `!org.billing_partner` - prevents the unnecessary API call to fetch the customer profile for partner orgs.
- (early return guard): `org.billing_partner` - returns null (hides the banner) when the org has a billing partner.


kanishkdudeja (approved):

alaister (approved):
LGTM!

PR #44196, opened 2026-03-25
fix: revert changes when the value is the same as old

awaseem:

YES

Revert changes when the value is the same as the old value


joshenlim (approved):
Verified changes on preview!

PR #44212, opened 2026-03-26
refactor(ui-patterns): Standardise TanStack sort headers

dnywh:

Component update.

TanStack tables in the repo are split between the shared `TableHeadSort` primitive and the older Studio-local `DataTableColumnHeader` helper, which makes the sorting UI and integration path inconsistent.

If you were to just use `DataTableColumnHeader` in `ui-patterns/Table`, you’d get a very different visual result to the  `TableHeadSort` UI you see in most other tables.

## What is the new behaviour?

Adds a shared `TanStackTableHeadSort` adapter in `ui-patterns/Table`, backed by the existing `TableHeadSort` primitive, and switches the webhook table plus the design-system TanStack demo to that canonical path. `DataTableColumnHeader` stays as a deprecated wrapper for now, Studio gets a lint guard to block new imports of it, and the table docs now point TanStack tables at the shared adapter explicitly.

## To test

Check out column sorting on the Platform Webhook endpoint deliveries table.


joshenlim (approved):
Changes LGTM! Did a quick check for the platform webhooks UI

fsansalvadore (approved):
Looks good ✅

PR #44214, opened 2026-03-26
Refactor ColumnEditor to use latest UI components + fix some UI oddities

joshenlim:
## Context

Refactors the Table Editor's `ColumnEditor` to use the latest UI components, and fix some UI oddities along the way

## Bug fixes

- Fix header text vertical alignment
  - Before:
    image
  - After:
    image
- Fix closing a toast would close the panel as well
  - Can verify by creating a new column, then hitting save without entering anything. Will trigger some error toasts and closing them will close the panel too
- Fix tooltips on "is nullable" and "is unique" showing up irregardless if "is primary key" is toggled on or off


dnywh (approved):
Tested on staging. Looks good!

PR #44232, opened 2026-03-26
chore: use `react-hook-form` in SQL snippets modals

djhi:
## Problem

- SQL snippets modals still uses `formik` and we want to remove it in favour of `react-hook-form` to keep only one form library

## Solution


awaseem (approved):
Tested locally at its fine!

PR #44233, opened 2026-03-26
chore: use `react-hook-form` in disk size configuration modal

djhi:
## Problem

- Disk size configuration modal still uses `formik` and we want to remove it in favour of `react-hook-form` to keep only one form library

## Solution

## How to test

Haven't found a way to test locally except by temporarily inverting conditions in `apps/studio/pages/project/[ref]/database/settings.tsx` at L57 and checking the API request payload in the network tab.


awaseem (approved):
Tested on staging!

PR #44258, opened 2026-03-26
feat(billing): show tax in subscription upgrade preview

ignaciodob:
This PR moves the client-side "Charge Today" calculation to the backend - relying on the `preview` endpoint data to show the `Charge Today`, `Prorated Credits` and `Customer Balance` in the Subscription Upgrade Preview.

It also includes fields for displaying Tax information in the preview, only for enabled orgs. Do note that at the moment the Tax Preview will not change if the address in the Subscription Preview changes, that will be tackled in a follow-up PR.

## Changes

- Replace client-side charge calculation with backend data: The upgrade dialog previously computed the prorated credit and total charge locally (using subscription period timestamps and plan prices). This PR uses the `upfront_charge` object returned by the subscription preview API instead.
- Display itemized tax breakdown: When the backend returns tax data, the dialog now shows a line-by-line breakdown: plan cost → subtotal (if different) → tax (with rate %) → total charged today.
- Use `tax_status` from the subscription preview to conditionally show tax details, and display a warning when tax calculation fails.

### No taxes - Upgrade from Free to Paid Plan
- With a Free Org with a billing address in Canada or any other non-enabled jurisdiction, start the upgrade to the Pro Plan
- Assert that only the Charge Today field is shown in the summary
image

### Taxes - Upgrade Plan
- Update an Org's Orb Customer on the Free or Pro Plan with the `automatic_tax_enabled: true` flag.
```
curl --location --request PUT 'https://api.withorb.com/v1/customers/external_customer_id/{ORG_SLUG}' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'Authorization: ••••••' \
--data '{
  "tax_configuration": {
    "tax_exempt": false,
    "tax_provider": "numeral",
    "automatic_tax_enabled": true
  }
}'
```
- Update the Org address to a valid taxable address, like: 
```
  "address": {
    "country": "US",
    "line1": "100 Congress Ave",
    "city": "Austin",
    "state": "TX",
    "postal_code": "78701"
  }
```
- Assert that the tax information is shown in the preview, including Tax Rate, Prorated credits (if applicable), Customer balance (if applicable) and the price of the intended plan.

image 

## Summary by CodeRabbit

* **New Features**
  * Added itemized breakdown of subscription charges, including plan cost, unused time credits, subtotal, and applicable taxes.
  * Enhanced billing estimates with detailed tax information display.

* **Improvements**
  * Updated "Charge today" calculations to reflect real-time preview data for greater accuracy.
  * Improved billing estimate UI layout and clarity with better organization of charge details.


apps/studio/data/organizations/organization-billing-subscription-preview.ts
```diff
@@ -43,6 +43,20 @@ export type OrganizationBillingSubscriptionPreviewResponse = {
     ref: string
   }[]
   billed_via_partner?: boolean
+  upfront_charge: {
```
kanishkdudeja: I think we need to mark `upfront_charge` as nullable to match the backend type?

apps/studio/data/organizations/organization-billing-subscription-preview.ts
```diff
@@ -43,6 +43,20 @@ export type OrganizationBillingSubscriptionPreviewResponse = {
     ref: string
```
kanishkdudeja: When a user enters a new billing address in the payment form during upgrade, the preview doesn't seem to be refetched. Are we planning to tackle this as a follow up?

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
```diff
@@ -211,23 +211,20 @@ export const SubscriptionPlanUpdateDialog = ({
...
-  const totalSeconds = subscription?.current_period_end - subscription?.current_period_start
-  const remainingRatio = remainingSeconds / totalSeconds
+  const proratedCredit = subscriptionPreview?.upfront_charge?.prorated_credit ?? 0
 
-  // Calculate prorated credit for current plan
-  const currentPlanMonthlyPrice = currentPlanMeta?.price ?? 0
-  const proratedCredit = currentPlanMonthlyPrice * remainingRatio
-
-  // Calculate new plan cost
   const newPlanCost = Number(subscriptionPlanMeta?.priceMonthly) ?? 0
```
kanishkdudeja: Not related to this PR but this could be an issue

Number(undefined) returns NaN, and NaN ?? 0 is still NaN

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
```diff
@@ -211,23 +211,20 @@ export const SubscriptionPlanUpdateDialog = ({
         })
       : []
 
-  // Calculate remaining days in current billing cycle
-  const now = Math.floor(Date.now() / 1000) // current time in seconds
-  const remainingSeconds = subscription?.current_period_end - now
-  const totalSeconds = subscription?.current_period_end - subscription?.current_period_start
-  const remainingRatio = remainingSeconds / totalSeconds
+  const proratedCredit = subscriptionPreview?.upfront_charge?.prorated_credit ?? 0
```
kanishkdudeja: This isn't a blocker and this isn't introduced by your PR, but given that we're refactoring this I thought of bringing this up.

The rendering logic is difficult to reason because the visibility rules are spread across several unrelated conditional (hasBreakdownItems, changeType !== 'downgrade', proratedCredit > 0, hasTax, taxFailed, customerBalance > 0 etc).

It makes it hard to answer a question like “which rows should show for this charge?”.

I think this would be clearer if the component derived an explicit `breakdownItems` array from `upfront_charge` and rendered from that.

Then the rule becomes: build the rows that apply, render them if any exist, and always render Charge today.

Something like this:

```
const breakdownItems = useMemo(() => {
    const adjustmentItems: Array<
      | { type: 'amount'; label: string; amount: number }
      | { type: 'notice'; label: string }
    > = []

    if (currentPlan !== 'free' && proratedCredit > 0) {
      adjustmentItems.push({
        type: 'amount',
        label: `Unused Time on ${currentPlanName} Plan`,
        amount: -proratedCredit,
      })
    }

    if (hasTax && tax) {
      if (taxableAmount !== newPlanCost) {
        adjustmentItems.push({ type: 'amount', label: 'Subtotal', amount: taxableAmount })
      }

      adjustmentItems.push({
        type: 'amount',
        label: `Tax (${tax.tax_rate_percentage}%)`,
        amount: tax.tax_amount,
      })
    }

    if (taxFailed) {
      adjustmentItems.push({
        type: 'notice',
        label: 'Tax could not be estimated and may be applied separately',
      })
    }

    if (customerBalance > 0) {
      adjustmentItems.push({ type: 'amount', label: 'Credits', amount: -customerBalance })
    }

    if (changeType !== 'downgrade' && adjustmentItems.length > 0) {
      return [{ type: 'amount' as const, label: `${planName} Plan`, amount: newPlanCost }, ...adjustmentItems]
    }

    return adjustmentItems
}, [])
```

apps/studio/data/organizations/organization-billing-subscription-preview.ts
ignaciodob: Yep - I mentioned this in the description, thanks for checking though!

> Do note that at the moment the Tax Preview will not change if the address in the Subscription Preview changes, that will be tackled in a follow-up PR.

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
ignaciodob: Great catch! Changed it to `Number(subscriptionPlanMeta?.priceMonthly) || 0`

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
ignaciodob: Excellent suggestion @kanishkdudeja! I went with something similar - but I think simpler.

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/SubscriptionPlanUpdateDialog.tsx
```diff
@@ -211,23 +215,96 @@ export const SubscriptionPlanUpdateDialog = ({
...
+
+  const newPlanCost = Number(subscriptionPlanMeta?.priceMonthly) || 0
+
+  const customerBalance = subscriptionPreview?.upfront_charge?.customer_balance ?? 0
+
+  const taxStatus = subscriptionPreview?.upfront_charge?.tax_status
```
kanishkdudeja: May want to have an object for upfront charge so that we don't need to repeat computing `subscriptionPreview?.upfront_charge?` multiple times

ivasilov (approved):

PR #44277, opened 2026-03-27
Migrate Database Webhook integrations to new integration sUI

joshenlim:
## Context

Related to marketplace integrations

Shifts Database Webhooks integration to the new Integrations UI. This one's a bit different from the previous PRs as this involves a full SQL installation query instead of only just a database extension. So am tweaking `Integrations.constants` a little.

For context eventually all the integrations will be pulled remotely from a database, so am still trying to figure out an optimal data structure, but requirements will be clearer as we build out the UI

RE installing integrations:
- For now, if the integration has a provided SQL installation command, that'll take precedence
- Else, if the integration has a provided SQL installation query, we'll use that on the /query endpoint
- Otherwise, if the integration only requires database extensions, dashboard will generate the queries to install the extensions
- In the case of the former tho, we won't allow users to choose which schema to install the extension in too

Just ping me if any clarification's required!

## To test
- [ ] Verify that you can install the database webhooks with the new integration UI
- [ ] Verify that behaviour is status quo without the new integration UI


charislam (approved):

avallete (commented):
The query looks correct to me, just to make sure I get it right, this is because we wanna be able to have the same UI between our "remote project" that use the "mgmt-api" for such extensions enabling right now. And the "local project ui" that doesn't have access to the mgmt-api ? So we bring the logic within studio instead of keeping it into our mgmt-api ?

apps/studio/components/interfaces/Integrations/Landing/Integrations.constants.tsx
```diff
@@ -266,6 +274,12 @@ const SUPABASE_INTEGRATIONS: Array<IntegrationDefinition> = [
       }
       return null
     },
+    installationSql: getEnableWebhooksSQL(),
+    installationCommand: async ({ ref }: { ref: string }) => {
+      const queryClient = getQueryClient()
+      await enableDatabaseWebhooks({ ref })
```
joshenlim, line 280: Still using mgmt API behind the scenes, no change to current behaviour

joshenlim:
> The query looks correct to me, just to make sure I get it right, this is because we wanna be able to have the same UI between our "remote project" that use the "mgmt-api" for such extensions enabling right now. And the "local project ui" that doesn't have access to the mgmt-api ? So we bring the logic within studio instead of keeping it into our mgmt-api ?

Ahh for context - the changes to the UI here are related to the marketplace integrations, in which we have a centralized and consistent "install integration" button for all integrations. This button will then toggle a panel that will show exactly what gets exactly installed on the project's database for transparency. (Hence the SQL tab under "Installs")


So for e.g if the extension is just a database extension, the SQL is just enabling the DB extension then.

So for Database Webhooks, its a similar concept, just that i understand the SQL cant be run from the dashboard due to permissions. So in the SQL tab, dashboard is just showing the SQL that the mgmt API runs, but under the hood we're still using the mgmt API's endpoint to enable the integrations (https://github.com/supabase/supabase/pull/44277#discussion_r3008633744)

^ Altho if we reckon that this might be a concern, then I can adjust things a bit as a workaround

soedirgo (approved):

PR #44278, opened 2026-03-27
chore: Clean the `ui` package from `next` imports

ivasilov:
This PR moves several components which rely on `next` out of the `ui` package to the `ui-patterns` package. 

`ui-patterns` package is intented to be imported with specific imports so it's ok if there are components reliant on `next` in there.

The `SonnerToaster` component has removed its dependency by requiring a prop for `theme`.


awaseem (approved):
LGTM mostly import changes

kemaldotearth (approved):

charislam (approved):

PR #44284, opened 2026-03-27
fix: detect prior consent via uc_user_interaction for fast cross-app navigation

seanoliver:
## Problem

Follow-up to #44252. The previous fix handled the case where the user waits on the page after accepting (GTM writes `ucData`), but there's a second scenario: if the user accepts and navigates to another app quickly (before GTM finishes), `ucData` hasn't been written yet. The SDK on the new app overwrites `uc_user_interaction: "true"` with `"false"` and shows the banner again.

## What changed

`hasPreviousConsentInUcData` is now `detectPriorConsent` and checks two signals before `UC.init()` overwrites them:

1. **Slow navigation** (existing): `ucData` contains all services with `consent: true`
2. **Fast navigation** (new): `uc_user_interaction` is `"true"` — the SDK on the previous app wrote this but GTM hasn't had time to replace it yet

Also adds unit tests covering both scenarios, edge cases (empty services, malformed JSON, mixed signals), and the combined behavior.

## Testing

- 13 unit tests for `detectPriorConsent()` covering all localStorage state combinations
- Can't fully reproduce on staging previews (CSP blocks GTM), verified root cause via production console monitoring

Closes FE-2648

### DIFF
--- a/packages/common/consent-state.test.ts
+++ b/packages/common/consent-state.test.ts
@@ -1,0 +1,137 @@
+// @vitest-environment jsdom
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+
+import { detectPriorConsent } from './consent-state'
+
+// jsdom's localStorage can be flaky in vitest, so ensure it's available
+const storage = new Map<string, string>()
+const mockLocalStorage = {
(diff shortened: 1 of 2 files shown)


awaseem (approved):
LGTM

PR #44292, opened 2026-03-27
fix(studio): fix infinite pagination loop in queue messages

oniani1:

YES

Bug fix

The `getNextPageParam` function in `database-queue-messages-infinite-query.ts` uses `<=` instead of `>=`, which means pagination never stops. Since the result length is always less than or equal to the page size, `hasNextPage` is always `true`, causing infinite API requests when viewing queue messages in Studio.

Resolves #44291

## What is the new behavior?

Uses `>=` to correctly detect when there are more pages, consistent with every other infinite query in the codebase:

- `database-cron-jobs-infinite-query.ts` uses `>=`
- `users-infinite-query.ts` uses `>=`
- `database-cron-jobs-runs-infinite-query.ts` uses `< PAGE_SIZE` to return `undefined` (equivalent logic)

## Additional context

One character change: `<=` to `>=` on line 112.


ivasilov (approved):

ivasilov:
@oniani1 ty for your fix!

PR #44297, opened 2026-03-28
Parallelise CI stages (fix Run Dependabot bottleneck)

urjitupadhya:
**Fix Type:** parallelize_ci
**Source:** Forked at urjitupadhya/supabase
*


7ttp (changes requested):
Closing this for now, since it only adds a metadata file and doesn’t include any actual CI changes

If you’d like, 
please open an issue/discussion with more detail and the team can take a look if it makes sense

PR #44309, opened 2026-03-28
fix(www): redirect unauthenticated "Start for Free" pricing CTA to sign-up

abdulbaqui17:

Yes

Bug fix

Clicking "Start for Free" on the [pricing page](https://supabase.com/pricing) redirects unauthenticated users to `/dashboard/new?plan=free`, which is an authenticated route.

The page immediately triggers a 401 from platform API calls (e.g. `/platform/notifications`) and bounces the user to `/dashboard/sign-in?plan=free&returnTo=%2Fnew` — the **sign-in** page, not sign-up. New users must manually find "Don't have an account? Sign up" to proceed.

Fixes #44280

## What is the new behavior?

The Free plan `href` now points to `/dashboard/sign-up?plan=free` so unauthenticated visitors land directly on the sign-up form.

`packages/shared-data/plans.ts` — one line changed:

```diff
-    href: 'https://supabase.com/dashboard/new?plan=free',
+    href: 'https://supabase.com/dashboard/sign-up?plan=free',
```

The `plan` query param is preserved so downstream logic can detect which plan the user came from. This pattern is consistent with existing unauthenticated CTAs in `apps/www/components/Nav/` which already route to `/dashboard/sign-up`.

### DIFF
--- a/packages/shared-data/plans.ts
+++ b/packages/shared-data/plans.ts
@@ -4,7 +4,7 @@ export interface PricingInformation {
   id: string
   planId: PlanId
   name: string
-  nameBadge?: string
+  nameBadge?: strin
   costUnit?: string
   href: string
   priceLabel?: string
@@ -25,7 +25,7 @@ export const plans: PricingInformation[] = [
     name: 'Free',
     nameBadge: '',
     costUnit: '/ month',
-    href: 'https://supabase.com/dashboard/new?plan=free',
+    href: 'https://supabase.com/dashboard/sign-up?plan=free',
     priceLabel: '',
     priceMonthly: 0,
     description: 'Perfect for passion projects & simple websites.',
--- a/packages/shared-data/plans.ts
+++ b/packages/shared-data/plans.ts
@@ -4,7 +4,7 @@ export interface PricingInformation {
   id: string
   planId: PlanId
   name: string
-  nameBadge?: strin
+  nameBadge?: string
   costUnit?: string
   href: string
   priceLabel?: string
@@ -25,7 +25,7 @@ export const plans: PricingInformation[] = [
     name: 'Free',
     nameBadge: '',
     costUnit: '/ month',
-    href: 'https://supabase.com/dashboard/sign-up?plan=free',
+    href: 'https://supabase.com/dashboard/new?plan=free',
     priceLabel: '',
     priceMonthly: 0,
     description: 'Perfect for passion projects & simple websites.',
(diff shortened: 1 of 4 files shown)


packages/shared-data/plans.ts
```diff
@@ -4,7 +4,7 @@ export interface PricingInformation {
   id: string
   planId: PlanId
   name: string
-  nameBadge?: string
+  nameBadge?: strin
```
kevcodez: Looks like you accidentally broke types here

packages/shared-data/plans.ts
```diff
@@ -25,7 +25,7 @@ export const plans: PricingInformation[] = [
     name: 'Free',
     nameBadge: '',
     costUnit: '/ month',
-    href: 'https://supabase.com/dashboard/new?plan=free',
+    href: 'https://supabase.com/dashboard/sign-up?plan=free',
```
kevcodez: This is not really fixing it unfortunately - as you are now breaking this for a logged in user. A logged in user is now no longer redirected to the organization creation

The proper fix is to handle both cases: logged in and non-logged in users
- Non-logged in users should be redirected to login and be redirected to org creation after logging in with the plan preselected
- Logged in users should be redirected to org creation with plan preselected

kevcodez (changes requested):

mahdirajaee (commented):
The redirect fix makes sense — routing unauthenticated users to `/dashboard/sign-up?plan=free` instead of `/dashboard/new?plan=free` eliminates the sign-in redirect loop described in #44280. The `plan` query param preservation is a good detail.

However, there is a **breaking change in this diff that needs to be fixed before merge**:

**Line 7 introduces a type error:**
```diff
-  nameBadge?: string
+  nameBadge?: strin
```

`strin` is not a valid TypeScript type. This will fail type checking. Looks like an accidental deletion of the trailing `g`. This needs to be reverted to `string` — the only intentional change in this PR should be the `href` value on line 28.

Other than that accidental edit, the actual fix is straightforward and correct. The pattern is consistent with other unauthenticated CTAs in the codebase that already route to `/dashboard/sign-up`.

abdulbaqui17:
Addressed the requested changes in follow-up commit `54b5a6274a`.

What I changed:
- Restored the accidental type regression in `packages/shared-data/plans.ts`:
  - `nameBadge?: strin` -> `nameBadge?: string`
- Implemented auth-aware routing for the Free plan pricing CTA in `www`:
  - Logged-in users: `/dashboard/new?plan=free`
  - Logged-out users: `/dashboard/sign-in?returnTo=/new?plan=free`
- Applied this consistently in both pricing CTA surfaces:
  - `apps/www/components/Pricing/PricingPlans.tsx`
  - `apps/www/components/Pricing/PricingComparisonTable.tsx`
- Added helper:
  - `apps/www/components/Pricing/getPricingPlanHref.ts`

This keeps logged-in org creation flow intact while fixing the unauthenticated redirect flow.

Hallidayo:

PR #44317, opened 2026-03-29
fix: correct inverted comparison operator in Queue Messages infinite pagination

mohithshuka:
## What changed
Fixed an inverted comparison operator in `getNextPageParam` inside `database-queue-messages-infinite-query.ts`.

## Problem
The condition `lastPage.length <= QUEUE_MESSAGES_PAGE_SIZE` was always `true` because the SQL query uses `LIMIT 30`, so results can never exceed 30. This caused an **infinite pagination loop** — the query kept fetching forever and never stopped.

## Fix
Changed `<=` to `>=`:
```ts
// Before (broken)
const hasNextPage = lastPage.length <= QUEUE_MESSAGES_PAGE_SIZE

// After (fixed)
const hasNextPage = lastPage.length >= QUEUE_MESSAGES_PAGE_SIZE
```
Now it correctly stops paginating when the last page has fewer than 30 results.

## Testing
- Queue with >30 messages → infinite scroll loads next pages correctly
- Queue with <30 messages → loads once and stops (no infinite loop)
- Empty queue → loads once and stops

Closes #44291

### DIFF
--- a/apps/studio/data/database-queues/database-queue-messages-infinite-query.ts
+++ b/apps/studio/data/database-queues/database-queue-messages-infinite-query.ts
@@ -109,7 +109,7 @@ export const useQueueMessagesInfiniteQuery = <TData = DatabaseQueueData>(
     enabled: enabled && typeof projectRef !== 'undefined',
     initialPageParam: undefined,
     getNextPageParam(lastPage) {
-      const hasNextPage = lastPage.length <= QUEUE_MESSAGES_PAGE_SIZE
+      const hasNextPage = lastPage.length >= QUEUE_MESSAGES_PAGE_SIZE
       if (!hasNextPage) return undefined
       return last(lastPage)?.enqueued_at
     },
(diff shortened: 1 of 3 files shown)


mahdirajaee (commented):
The one-character fix is correct — `<=` was effectively always true since the SQL query caps results at `LIMIT 30`, so every page satisfied `length <= 30`, producing the infinite loop described in #44291.

Changing to `>=` means: "if the last page returned a full page of results, there might be more." When the page has fewer than 30 rows, pagination stops. Standard TanStack Query infinite pagination pattern.

A few notes on the rest of the diff:

1. **Core fix is clean and minimal.** One operator flip, correct semantics. Good.

2. **Unrelated additions in the same PR.** The `getUpdateSerialSequenceSQL` function in `packages/pg-meta/src/sql/studio/table-editor/identity.ts` and its tests in `SpreadsheetImport.utils.test.ts` don't appear related to the queue pagination bug. Mixing unrelated changes into a bug-fix PR makes backporting and bisecting harder. Consider splitting the serial-sequence work into its own PR.

3. **Test file placement.** The `getUpdateSerialSequenceSQL` tests are appended to `SpreadsheetImport.utils.test.ts`, which is a test file for spreadsheet import utilities. These tests should live in a file that corresponds to the identity module, not the spreadsheet import module — otherwise future maintainers won't find them.

The pagination fix itself is correct and should resolve the infinite loop.

7ttp (commented):
hi @mohithshuka thanks for this! 💚😅

but closing this as a duplicate/in favour of : https://github.com/supabase/supabase/pull/44292


Thanks for ur efforts towards making Supabase better !!

PR #43807, opened 2026-03-16
fix: browser back navigation on observability page

7ttp:
## Problem
The observability pages has a browser navigation issue where pressing the back button would get stuck in a redirect loop between `/observability` and `/observability/api-overview`, making it difficult to navigate away using browser controls

<details>
<summary>vid:</summary>

https://github.com/user-attachments/assets/4b48651b-143d-4857-831b-67d1b98e237d

</details>

## Solution
Changed automatic redirects from `router.push()` to `router.replace()` in the routing logic.

## Related
Same pattern as https://github.com/supabase/supabase/commit/f8419ef9ae7c43ea033617e5544c565dfaaf87acc

PR #44322, opened 2026-03-29
fix(studio): use router.replace for observability auto-redirects to fix back-navigation loop

abdulbaqui17:
Yes

Bug fix

The /observability index page auto-redirects using router.push(), which adds entries to the browser history. Pressing back returns to /observability, which immediately redirects forward again — creating an infinite back-navigation loop making it impossible to leave the page.

Fixes #43807

## What is the new behavior?
Changed both router.push() calls to router.replace() so the intermediate /observability URL is replaced in history rather than pushed. The back button now correctly navigates away.

## Additional context
Same pattern as commit f8419ef which applied the identical fix elsewhere in the codebase. One-line change, no UI impact.


7ttp (commented):
hi @abdulbaqui17  :)
what u are fixing is already a PR itself,

closing this as a duplicate of: https://github.com/supabase/supabase/pull/43807

PR #44331, opened 2026-03-30
Refactor table editor logic for handling null and undefined values

### LINKED ISSUE #43548: Dashboard row insert doesn't work properly due to dashboard insert logic not differentiate between "doing nothing" or inserting NULL

joshenlim:
## Context

Resolves https://github.com/supabase/supabase/issues/43548

There's currently an issue with the Table Editor where if you have, for example, a nullable `text` column with a default value, inserting a new row and selecting "Set to NULL" doesn't do anything, and saving will insert the row with the default value
image

This stems from a legacy logic in the Table Editor whereby we treat `null` values as "no input" - which is incorrect as `null` values are also valid values. So the PR here changes a few things to resolve this properly:

## Changes involved
Main fix: 
- `undefined` will be the "no input" value instead, and it'll be the default value when generating the row object for inserting a new row
  - `NULL` or even empty string like `''` will be treated as they are (valid inputs)

Secondary adjustments:
- (Queue operations) Queueing an insert with no value but default value is NULL, will show the placeholder as `DEFAULT` instead of `NULL` for better accuracy in representation
  image
- Added a `Set to Default` CTA here, but will only show up if adding a new row or updating a queued insert row operation, which will set the value of the input field back to `undefined` for PG to handle it as the default value
  image

## To test
Verify within the Table Editor (along with queue operation feature preview)
- For inserting a new row, setting value to NULL and setting value to Default works
- For updating a row, setting value to NULL works


alaister (approved):
LGTM!

- Tested on preview

PR #44332, opened 2026-03-30
fix: Set the default branch if it's not already set on Github integration

ivasilov:
Previously, the POST to `/v1/projects/{ref}/branches` didn't work on free projects, which left some Supabase projects setup without a default branch. Easy fix for those projects would be to disable/enable the whole integration. 

This PR fixes that so that when the user wants to setup a production sync, the code checks if there's branches already setup and if not, sets the selected branch as default.


joshenlim (approved):
Verified changes locally!

PR #44345, opened 2026-03-30
fix auth providers string fields are all considered sensitive

djhi:
## Problem

All Auth Providers string fields are considered sensitive

## Solution

Restore check of the `sensitive` property that was mistakenly removed in #44095

## Screenshots

image


awaseem (approved):
Tested on staging!

PR #44346, opened 2026-03-30
feat: promote new banner endpoint to prod

charislam:
New incident banner workflow should now be the standard in prod as well as preview.


awaseem (approved):
LGTM

PR #44349, opened 2026-03-30
Joshen/fe 2899 migrate wrappers integration to new UI

joshenlim:
## Context
Related to marketplace integrations, now shifting wrappers over to the new UI (feature flagged)

## Changes involved
- Added a "Installed" indicator if the integration is installed
  image
- New UI will only show the "Add new wrapper" + "Recent wrappers" table only if the required extensions have been installed (Current UI shows them both irregardless)
  image
- Once required extensions are installed, the wrappers tab will also show up
  - Currently the wrappers tab only shows up after a wrapper is created
  - This change will affect the existing behaviour too (reckon it makes more sense this UX)
  image
- In the install integration sheet, only extensions that have yet to be installed will users be allowed to adjust schema selection for
  image
- Likewise, if the extension has already been installed, it will be omitted from the SQL code output
  image
- Updated docs URL for some of the integrations (vault, database webhooks, graphql)
- Nit: This one's bugged me for the longest time but remove extra border bottom in "Recent wrappers" table
  image

FWIW i think the integrations UI for wrappers is due for an update, but we'll leave things as they are for now - at least until we're bit more clearer on the requirements for marketplace integrations (less we make changes now which then have to be redone again later)

## To test
- [ ] Verify that you can install + create wrappers still with the new UI
- [ ] Verify that behaviour is status quo with flag off (except the ones mentioned explicitly)

### DIFF
--- a/apps/studio/components/interfaces/Integrations/Integration/IntegrationOverviewTabV2/InstallIntegrationSheet.tsx
+++ b/apps/studio/components/interfaces/Integrations/Integration/IntegrationOverviewTabV2/InstallIntegrationSheet.tsx
@@ -108,7 +108,7 @@ export const InstallIntegrationSheet = ({ integration }: InstallIntegrationSheet
   const { mutateAsync: enableExtension } = useDatabaseExtensionEnableMutation({ onError: () => {} })
 
   const enableExtensionsSQL = getEnableExtensionsSQL({
-    extensions: requiredExtensions,
+    extensions: requiredExtensions.filter((ext) => !ext.installed_version),
     extensionsSchema,
   })
   const installationSQLContent = installationSql ?? enableExtensionsSQL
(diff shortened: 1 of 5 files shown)


apps/studio/components/interfaces/Integrations/Integration/IntegrationOverviewTabV2/InstallIntegrationSheet.tsx
```diff
@@ -156,26 +156,28 @@ export const InstallIntegrationSheet = ({ integration }: InstallIntegrationSheet
...
+        .filter((ext) => !ext.installed_version)
+        .map((ext) => {
+          const { name, default_version: version } = ext
+          const createSchema = extensionsSchema[name].schema === 'custom'
+          const schema =
+            name === 'pg_cron'
+              ? 'pg_catalog'
+              : createSchema
+                ? (extensionsSchema[name].value as string)
+                : extensionsSchema[name].schema
```
charislam, line 169: I recall having seen this multiple times now, is it worth extracting a utility function for it?

apps/studio/components/interfaces/Integrations/Wrappers/OverviewTab.tsx
```diff
@@ -68,92 +109,108 @@ export const WrapperOverviewTab = () => {
...
-              disabled={!canCreateWrapper}
-              tooltip={{
-                content: {
-                  text: !canCreateWrapper
-                    ? 'You need additional permissions to create a foreign data wrapper'
-                    : undefined,
-                },
-              }}
+  return !!wrapperMeta && isWrappersExtensionInstalled && !hasRequiredVersion ? (
+    <div className="">
```
charislam: unnecessary div?

charislam (approved):

apps/studio/components/interfaces/Integrations/Integration/IntegrationOverviewTabV2/InstallIntegrationSheet.tsx
joshenlim: yup! i'll probably get to this down the road with the next PR as this logic still needs tweaking a little 🙏

apps/studio/components/interfaces/Integrations/Wrappers/OverviewTab.tsx
joshenlim: ahh thanks for the catch! i'll refactor the alert as well actually

PR #44351, opened 2026-03-30
fix: forward cache-control headers in SSR cookie handlers

mandarini:
## Summary

Updates all `setAll` cookie handlers across the codebase to accept and forward the `headers` parameter introduced in `@supabase/ssr` v0.10.0. The dependency bump itself already landed on master -- this PR updates the code examples, UI library clients, and Studio snippets to actually use the new signature.

The `headers` parameter carries cache-control headers (`Cache-Control: private, no-cache, no-store`, `Pragma: no-cache`, `Expires: 0`) that prevent CDNs from caching responses containing refreshed auth tokens. Without forwarding these headers, a CDN could serve a cached authenticated response to the wrong user.

### Next.js UI library clients: per-context file split

The single `server.ts` file was split into two context-specific files:

- **`server-component.ts`** -- Read-only client for Server Components. Only implements `getAll` (no `setAll`, no silent try/catch). Server Components cannot set cookies or response headers, so providing `setAll` there was misleading and silently discarded headers.
- **`server-route.ts`** -- Full read-write client for Route Handlers and Server Actions. Accepts a `NextRequest`, returns `{ supabase, response }`, and properly forwards both cookies and cache-control headers via the response object.

All blocks and example pages updated to import from the correct new file based on their context (server components use `server-component`, route handlers use `server-route`).

### Next.js middleware

Updated `setAll` to accept and forward the `headers` parameter via `supabaseResponse.headers.set()`.

### React Router client

Renamed the local `headers` variable to `responseHeaders` to avoid shadowing the new `headers` callback parameter, and forwards cache-control headers via `responseHeaders.append()`.

### TanStack Start client

Updated `setAll` to accept `headers` and forward them using `setResponseHeader()` from `@tanstack/react-start/server`.

### Studio Connect/ConnectSheet snippets

Updated all Next.js and Remix code snippets shown in the Studio Connect UI to reflect the new `setAll` signatures and header forwarding.

### `apps/www`

Updated `supabaseContribute.server.ts` to accept `_headers` in `setAll` (server component context, param accepted but not applied).

## Notable details

- The `SetAllCookies` type change in `@supabase/ssr` v0.10.0 is non-breaking at the TypeScript level (functions with fewer params satisfy the type), which is why existing code compiled without errors despite missing the new parameter.
- Route handlers (`confirm/route.ts`, `oauth/route.ts`) were switched from `redirect()` (next/navigation) to `NextResponse.redirect()` so that cookies and cache-control headers propagate correctly on the redirect response.
- See [Railway incident report](https://blog.railway.com/p/incident-report-march-30-2026-accidental-cdn-caching) for real-world context on why CDN caching of authenticated responses is a security concern.

Related: supabase/ssr#176


awaseem (commented):
LGTM

apps/ui-library/registry/default/clients/nextjs/lib/supabase/server.ts
```diff
@@ -16,7 +16,7 @@ export async function createClient() {
         getAll() {
           return cookieStore.getAll()
         },
-        setAll(cookiesToSet) {
+        setAll(cookiesToSet, _headers) {
```
fadymak: Why aren't we setting the headers in this case?

mandarini: In the server component context, `cookies()` from next/headers only exposes cookie methods, so there's no API to set arbitrary response headers like `Cache-Control`. That's only available in middleware via `NextResponse`. So `_headers` is accepted to satisfy the type but there's nowhere to forward it here.

mandarini:
Closing after discussing internally with @fadymak .

PR #44354, opened 2026-03-30
fix: correct Next.js publishable key environment variable usage

mohithshuka:
## Description
This PR fixes the handling of the Next.js publishable key environment variable.

The previous implementation had issues with how the environment variable was referenced, which could cause incorrect configuration when initializing the Supabase client.

## Changes
- Updated the environment variable reference for the Next.js publishable key.
- Ensured compatibility with standard Next.js environment variable conventions.

## Why this change is needed
Next.js requires public environment variables to follow the `NEXT_PUBLIC_` prefix convention. Without this, the variable may not be available on the client side.

## Testing
- Verified that the Supabase client initializes correctly with the updated environment variable.
- Confirmed that the environment variable is accessible in the Next.js application.

## Type of change
- Bug fix


7ttp (commented):
seems resolved!

> Thanks for ur efforts towards contributing to Supabase!! 💚

PR #44372, opened 2026-03-31
feat(studio): advisor signal items for banned IPs

dnywh:

Feature. Resolves DEPR-430.

The homepage Advisor summary, shared Advisor panel, and top-nav Advisor indicator only surface lints and notifications. Banned IPs are not represented as dismissible Advisor items, so network bans are easy to miss unless a user visits Database Settings directly.

The `public bucket allows listing` warning is no longer part of this PR. That warning will move to a follow-up Splinter `WARN` lint so it can flow through the standard lint surfaces instead of a bespoke Studio signal path.

## What is the new behaviour?

- adds a new Advisor `signal` source for banned IPs on the platform homepage, in the shared Advisor panel, and in the top-nav Advisor indicator
- keeps dismissals client-side only for now, scoped by project and exact IP fingerprint
- keeps banned IP signals at `warning` severity because they still indicate suspicious traffic and remain actionable if a user wants to review or remove a ban
- leaves `/project/[ref]/advisors/security` as follow-up work because that surface is still lint-native, and banned IPs are management-plane signals rather than Splinter lints

| After |
| --- |
| Mallet  Toolshed  Supabase-65A60B4A-107E-4D79-B9A8-23F754BEAB08 |

## Reviewer testing notes

1. Use a throwaway project.
2. Get the database connection string for that project.
3. Attempt to connect with the wrong password 3-4 times until you hit an `ECONNREFUSED`-style error, which should mean your IP has been banned.
4. Refresh Studio and confirm the project overview shows the new `Banned IP address` signal.
5. Open the Advisor Center and confirm:
   - the top-nav Advisor dot turns warning yellow
   - the signal detail shows `Entity`, `Issue`, and `Resolve`
   - `Edit network bans`, `Dismiss`, and `Learn more` are present
6. Open Database Settings > Network bans and confirm your banned IP appears there and can be unbanned.
7. Note that `/project/[ref]/advisors/security` will not show this item. That page is still lint-only, and this banned IP work is a short-term client-side signal rather than a true lint.

Longer term, we likely want a more durable event model here so banned IPs can power notifications, webhooks, emails, and other project-level alerts.


apps/studio/components/ui/AdvisorPanel/useAdvisorSignals.ts
```diff
@@ -0,0 +1,51 @@
...
+  enabled?: boolean
+}
+
+export const useAdvisorSignals = ({
+  projectRef,
+  enabled = true,
+}: UseAdvisorSignalsOptions = {}) => {
+  // TODO(DEPR-430): Remove this local-only shim once network bans can be created for local testing.
+  const debugBannedIPs = getAdvisorDebugBannedIPs(process.env.NEXT_PUBLIC_ADVISOR_DEBUG_BANNED_IPS)
+
```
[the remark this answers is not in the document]
dnywh: Whoops. Must remove before merging.

kemaldotearth:
Apologies, you've ended up with some merge conflicts due to my PR's, let me know if you need a second set of eyes to review/fix it!

kemaldotearth (changes requested):
I know you got some bits to resolve on this first, but locally, after going through your setup instructions, I see the advisors but not the mocked banned IP's in the network bans. Maybe some propagation thing going on there?

Image

apps/studio/components/interfaces/ProjectHome/AdvisorSection.tsx
```diff
@@ -26,47 +36,75 @@ export const AdvisorSection = ({ showEmptyState = false }: { showEmptyState?: bo
...
+            : undefined
+        const advisorType = item.source === 'lint' ? item.original.name : item.title
+        const advisorLevel = item.source === 'lint' ? item.original.level : undefined
+
+        track('advisor_detail_opened', {
+          origin: 'homepage',
+          advisorSource: item.source,
+          advisorCategory,
+          advisorType,
+          advisorLevel,
```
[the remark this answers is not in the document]
dnywh: Fair point. Added in latest commit FYI @supabase/growth-eng. Signal opens now emit `advisor_detail_opened` telemetry with `advisorSource: 'signal'`.

pamelachia (approved):
approving the telemetry part only

kemaldotearth (approved):
@dnywh Approving from design pov, all seemed above board (despite the IPV4 thing). Would get someone from @supabase/dashboard to have a second look too!

apps/studio/hooks/misc/useLocalStorage.ts
joshenlim, line 1: any context for why the changes in this file is required?
this hook is being used in many places so I'm a bit more wary of making changes

apps/studio/data/banned-ips/banned-ips-query.ts
```diff
@@ -11,12 +11,28 @@ export type BannedIPVariables = {
...
+  if (typeof AbortSignal.any === 'function') {
+    return AbortSignal.any([signal, timeoutSignal])
+  }
+
+  // AbortSignal.any unavailable (older browsers): prefer the timeout signal so
+  // the 5 s cap is always enforced, even at the cost of losing React Query's
+  // own cancellation signal on this path.
+  return timeoutSignal
+}
+
```
joshenlim: any context on why we're doing this?

apps/studio/components/ui/AdvisorPanel/useAdvisorSignals.ts
```diff
@@ -0,0 +1,53 @@
...
+  enabled = true,
+}: UseAdvisorSignalsOptions = {}) => {
+  const bannedIPsQuery = useBannedIPsQuery(
+    { projectRef },
+    {
+      enabled,
+    }
+  )
+  const { dismissSignal, dismissedFingerprintSet, pruneDismissedSignals } =
+    useAdvisorSignalDismissals(projectRef)
```
joshenlim: it seems odd that we're separating `useAdvisorSignals` and `useAdvisorSignalDismissals` into 2 separate hooks? why not consolidate them into one within `useAdvisorSignals` given that they're both the same context?

e.g `dismissSignal` that we're returning is just passing through the method from `useAdvisorSignalDismissals`

apps/studio/components/ui/AdvisorPanel/useAdvisorSignalDismissals.ts
```diff
@@ -0,0 +1,58 @@
+import { useCallback, useMemo } from 'react'
+
+import { createAdvisorSignalDismissalStorageKey } from './AdvisorPanel.utils'
+import { useLocalStorageQuery } from '@/hooks/misc/useLocalStorage'
+
+export const useAdvisorSignalDismissals = (projectRef?: string) => {
+  const storageKey = projectRef
+    ? createAdvisorSignalDismissalStorageKey(projectRef)
```
joshenlim: given that this is the only file using this function - we can just have this util sit in this file instead of `AdvisorPanel.utils`

apps/studio/components/ui/AdvisorPanel/useAdvisorSignals.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const signalItems = useMemo(
+    () =>
+      createAdvisorSignalItems({
```
joshenlim: likewise - given that this is the only file using `createAdvisorSignalItems` - we can just have this util sit in this file instead of `AdvisorPanel.utils` so that things are more contextually grouped

apps/studio/components/ui/AdvisorPanel/AdvisorPanel.utils.ts
```diff
@@ -1,10 +1,151 @@
...
+  projectRef?: string
+  bannedIPsData?: IPData
+}): AdvisorSignalItem[] => {
+  if (!projectRef) return []
+
+  const bannedIPs = bannedIPsData?.banned_ipv4_addresses ?? []
+
+  const bannedIpSignals = bannedIPs.map((ip) => ({
+    id: createBannedIPSignalFingerprint(ip),
+    fingerprint: createBannedIPSignalFingerprint(ip),
```
joshenlim: bit confused - what does fingerprint mean here? (if we can opt for another noun to improve clarity/readability)

it feels like an ID, but there's already another `id` prop

apps/studio/components/ui/AdvisorPanel/AdvisorPanel.utils.ts
```diff
@@ -1,10 +1,151 @@
...
+    source: 'signal' as const,
+    signalType: 'banned-ip' as const,
+    severity: 'warning' as const,
+    tab: 'security' as const,
+    title: 'Banned IP address',
+    description: `The IP address \`${ip}\` is temporarily blocked because of suspicious traffic or repeated failed password attempts.`,
+    detailDescription:
+      'This IP address is temporarily blocked because of suspicious traffic or repeated failed password attempts. If this block is expected, you can dismiss this signal or remove the ban.',
+    learnMoreHref: 'https://supabase.com/docs/reference/cli/supabase-network-bans',
```
joshenlim: these are very tiny nits so feel free to ignore but just thinking that some of the properties can be renamed to be more succinct / clearer

`signalType` -> `type`
`description` -> `summary`
`detailDescription` -> `description`
`learnMoreHref` -> `docsUrl` (we use this language commonly)

apps/studio/components/interfaces/ProjectHome/AdvisorSection.tsx
```diff
@@ -1,16 +1,27 @@
...
+import type { AdvisorItem } from '../../ui/AdvisorPanel/AdvisorPanel.types'
+import {
+  createAdvisorLintItems,
+  getAdvisorItemDisplayTitle,
+  MAX_HOMEPAGE_ADVISOR_ITEMS,
+  severityBadgeVariants,
+  severityColorClasses,
+  sortAdvisorItems,
+} from '../../ui/AdvisorPanel/AdvisorPanel.utils'
+import { useAdvisorSignals } from '../../ui/AdvisorPanel/useAdvisorSignals'
```
joshenlim: lets use `@/` import syntax here?

apps/studio/hooks/misc/useLocalStorage.ts
dnywh: Claude reckons that the current hook is doing unnecessary background work on every render, and that removing that dependency list fixes it (without change to existing callers):

> We're calling `useLocalStorageQuery(key, [])` — that `[]` is created fresh on every render (two empty arrays are never equal in JS). The old code had it in a dependency list, so the hook was doing unnecessary background work on every single render.

apps/studio/data/banned-ips/banned-ips-query.ts
dnywh: Because the `AdvisorButton` calls this on every page load (to check if the nav badge dot should be rendered). If that network bans endpoint is slow or down (without a timeout) the request just hangs.

I believe this is timeout is also needed for the `AdvisorSection` on Project Overview. That big row would just show a shimmer indefinitely if the endpoint stalls (which it did on staging).

apps/studio/hooks/misc/useLocalStorage.ts
joshenlim: hmm okay gotcha on this - i'll just add some unit tests to make sure

apps/studio/data/banned-ips/banned-ips-query.ts
joshenlim: hmm just to clarify - so its to cancel the request after 5 seconds and not let it stall?
feels like there might be something on react query that we can use instead of manually implementing - will check

[edit] ahh okay - has to be implemented manually
although im wary tho - this feels like a new pattern that we're introducing RE timing out requests manually on the dashboard

> that big row would just show a shimmer indefinitely if the endpoint stalls 

this would apply to all UI across the dashboard tho - even for the table editor for e.g if the request to fetch rows takes really long its just showing a big loader then (from a design pattern POV)

joshenlim: what happens if the request hangs? is there a behaviour that's affected because of this?

dnywh: > is there a behaviour that's affected because of this?

Yep. See [this recording](<https://supabase.slack.com/archives/C0429V78ACX/p1775729636742939?thread_ts=1774908680.707569&cid=C0429V78ACX>). I’m happy to remove and retest if you like?

apps/studio/components/ui/AdvisorPanel/AdvisorSignalDetail.tsx
```diff
@@ -0,0 +1,96 @@
...
+import type { AdvisorSignalItem } from './AdvisorPanel.types'
+import { SIDEBAR_KEYS } from '@/components/layouts/ProjectLayout/LayoutSidebar/LayoutSidebarProvider'
+import { AiAssistantDropdown } from '@/components/ui/AiAssistantDropdown'
+import { InlineLink } from '@/components/ui/InlineLink'
+import { useAiAssistantStateSnapshot } from '@/state/ai-assistant-state'
+import { useSidebarManagerSnapshot } from '@/state/sidebar-manager-state'
+
+interface AdvisorSignalDetailProps {
+  item: AdvisorSignalItem
+  onDismiss: (dismissalKey: string) => void
```
joshenlim: nit but since the logic for onDismiss can all be retrieved from hooks, can just call the hooks directly here rather than prop drill

apps/studio/components/ui/AdvisorPanel/AdvisorPanel.tsx
```diff
@@ -70,15 +43,24 @@ export const AdvisorPanel = () => {
...
+  } = useProjectLintsQuery({ projectRef: project?.ref }, { enabled: shouldLoadProjectAdvisorData })
+
+  const {
+    data: signalItems,
+    dismissSignal,
+    isPending: isSignalsPending,
+    isError: isSignalsError,
+  } = useAdvisorSignals({
+    projectRef: project?.ref,
+    enabled: shouldLoadProjectAdvisorSignals,
```
joshenlim: fwiw because we've got multiple places calling this hook - the enabled logic needs to be consistent or banned ips will get fetched anyways

joshenlim: not a biggie tho for this case specifically i think

apps/studio/components/layouts/AppLayout/AdvisorButton.tsx
```diff
@@ -3,15 +3,21 @@ import { useMemo } from 'react'
...
+import { IS_PLATFORM } from '@/lib/constants'
 import { useSidebarManagerSnapshot } from '@/state/sidebar-manager-state'
 
 export const AdvisorButton = ({ projectRef }: { projectRef?: string }) => {
   const { toggleSidebar, activeSidebar } = useSidebarManagerSnapshot()
 
   const { data: lints } = useProjectLintsQuery({ projectRef })
+  const { data: signalItems } = useAdvisorSignals({
+    projectRef,
+    enabled: IS_PLATFORM && Boolean(projectRef),
```
joshenlim: not required - these are all checked within the banned-ips-query already

joshenlim (approved):
LGTM!

apps/studio/components/ui/AdvisorPanel/AdvisorSignalDetail.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const { ref: projectRef } = useParams()
+  const dismissSignal = useAdvisorSignalDismissal(projectRef)
+  const { setSelectedItem } = useAdvisorStateSnapshot()
+  const snap = useAiAssistantStateSnapshot()
+  const { openSidebar } = useSidebarManagerSnapshot()
+
+  const handleDismiss = () => {
+    dismissSignal(item.dismissalKey)
+    setSelectedItem(undefined)
+  }
```
joshenlim: this is duplicated code - there's already a dismiss function below, will fix

apps/studio/components/ui/AdvisorPanel/useAdvisorSignals.ts
```diff
@@ -43,6 +42,30 @@ const createBannedIPSignalItems = ({
...
 }
 
+/**
+ * Lightweight hook that only provides `dismissSignal`.
+ * Use this inside detail views (e.g. `AdvisorSignalDetail`) instead of the
+ * full `useAdvisorSignals` to avoid creating a second `useBannedIPsQuery`
+ * instance and a second pruning effect, which would cause an infinite update
+ * loop when both hooks share the same localStorage query key.
+ */
+export const useAdvisorSignalDismissal = (projectRef?: string) => {
```
joshenlim: this seems unnecessary - `useBannedIPsQuery` is a react query hook so it doesn't really matter how many times we call it since the data is shared

joshenlim: will remove this and revert the change

joshenlim (approved):
Double approve!

PR #44378, opened 2026-03-31
fix(studio): parameterize database queue SQL queries

### LINKED ISSUE #44375: Dashboard: Queue Messages queries use string interpolation for SQL parameters
### Describe the bug

Several files in `apps/studio/data/database-queues/` construct SQL queries by directly interpolating user-controlled values into template literals instead of using parameterized queries. While `queueName` is validated via regex (`isQueueNameValid`), other values like `payload` in the send mutation and `afterTimestamp` in the pagination query are interpolated without sanitization.

### Affected files

- `database-queue-messages-infinite-query.ts` — `afterTimestamp` is interpolated directly:
  ```ts
  query += ` WHERE enqueued_at > '${afterTimestamp}'`
  ```
- `database-queue-messages-send-mutation.ts` — `payload` is interpolated directly:
  ```ts
  sql: `select * from pgmq.send( '${queueName}', '${payload}', ${delay})`
  ```
- `database-queue-messages-delete-mutation.ts` — `messageId` (number type, lower risk but still not parameterized)
- `database-queue-messages-archive-mutation.ts` — same pattern with `messageId`

### Impact

The `payload` field in the send mutation is the most concerning since it accepts arbitrary user-provided JSON content. A payload containing a single quote (`'`) will at minimum break the query; a crafted payload could potentially execute arbitrary SQL. The `afterTimestamp` comes from `enqueued_at` values returned by previous queries, which is lower risk but still not safe to interpolate.

Even though `executeSql` runs through the project's SQL execution pipeline (which may use a service role connection), the defense-in-depth principle still applies — these should use parameterized queries or at minimum escape the interpolated values.

### Expected behavior

SQL parameters should be passed through parameterized query mechanisms rather than string interpolation. For the `pgmq` function calls, this could look like:
```ts
sql: `SELECT * FROM pgmq.send($1, $2::jsonb, $3)`,
params: [queueName, payload, delay]
```

Or at minimum, the values should be escaped before interpolation using a function like `pg-meta`'s `ident()` or a dedicated escape utility.

### System information

- Affected area: Studio Dashboard — Queues feature
- All files under: `apps/studio/data/database-queues/`

vj2303:
## Summary
Extends `executeSql` with optional `parameters` (forwarded to pg-meta `RunQueryBody`) and uses PostgreSQL placeholders for Studio database queue operations instead of string interpolation.

## Changes
- **execute-sql-query**: optional `parameters` on requests (including preflight EXPLAIN); `fetcherOverride` and deprecated `useExecuteSqlQuery` updated accordingly.
- **Queue messages**: parameterized `pgmq.send`, `delete`, `archive`, `set_vt`, and pagination `enqueued_at` filter.

## Context
Fixes defense-in-depth concerns from user-controlled `payload` and timestamps (GH-44375).

Closes #44375

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/studio/data/database-queues/database-queue-messages-archive-mutation.ts
+++ b/apps/studio/data/database-queues/database-queue-messages-archive-mutation.ts
@@ -27,7 +27,8 @@ export async function archiveDatabaseQueueMessage({
   const { result } = await executeSql({
     projectRef,
     connectionString,
-    sql: `SELECT * FROM pgmq.archive('${queueName}', ${messageId})`,
+    sql: `SELECT * FROM pgmq.archive($1::text, $2::bigint)`,
+    parameters: [queueName, messageId],
     queryKey: databaseQueuesKeys.create(),
   })
 
--- a/apps/studio/data/database-queues/database-queue-messages-delete-mutation.ts
+++ b/apps/studio/data/database-queues/database-queue-messages-delete-mutation.ts
@@ -28,7 +28,8 @@ export async function deleteDatabaseQueueMessage({
   const { result } = await executeSql({
     projectRef,
     connectionString,
-    sql: `SELECT * FROM pgmq.delete('${queueName}', ${messageId})`,
+    sql: `SELECT * FROM pgmq.delete($1::text, $2::bigint)`,
+    parameters: [queueName, messageId],
     queryKey: databaseQueuesKeys.create(),
   })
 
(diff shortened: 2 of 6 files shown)


7ttp (commented):
closing in favour of: https://github.com/supabase/supabase/pull/44451
(which closed the specific issue u linked)


Thanks for ur efforts towards contributing to Supabase!! 💚

PR #44381, opened 2026-03-31
fix(studio): allow native Apple sign-in without OAuth secret

abdulbaqui17:
## What changed
- Updated Apple auth validation in `AuthProvidersFormValidation.tsx` to allow an empty `EXTERNAL_APPLE_SECRET` when a client ID is provided by using `excludeEmptyString: true` on the JWT format check.
- Added a unit regression test for Apple provider validation.

## Why
Issue #44315 reports that native Apple sign-in (bundle ID only) incorrectly requires a secret key in Studio. The schema already intends to allow empty values, but the JWT regex rejected empty strings before downstream tests could allow that path.

## Validation
- `pnpm exec vitest run tests/unit/auth-providers-form-validation.test.ts`
- `pnpm exec eslint components/interfaces/Auth/AuthProvidersFormValidation.tsx tests/unit/auth-providers-form-validation.test.ts`

Closes #44315

### DIFF
--- a/apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
+++ b/apps/studio/components/interfaces/Auth/AuthProvidersFormValidation.tsx
@@ -478,7 +478,10 @@ const EXTERNAL_PROVIDER_APPLE = {
         },
         then: (schema) =>
           schema
-            .matches(/^[a-z0-9_-]+([.][a-z0-9_-]+){2}$/i, 'Secret key should be a JWT.')
+            .matches(/^[a-z0-9_-]+([.][a-z0-9_-]+){2}$/i, {
+              message: 'Secret key should be a JWT.',
+              excludeEmptyString: true,
+            })
             .test({
               message: 'Secret key is not a correctly generated JWT.',
               test: (value?: string): boolean => {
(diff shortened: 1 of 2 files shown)


anatoly-os:
Please take a look at it asap. There is no way to enable Apple Sign in for native apps right now!

anatoly-os:
Same issue in the dashboard...

Screenshot 2026-03-31 at 20 05 53

abdulbaqui17:
Validated against the dashboard behavior shown in the screenshot.

Root cause: Apple secret validation treated empty string as a JWT format failure even for native client ID setup.

Fix in this PR:
- allow empty `EXTERNAL_APPLE_SECRET` by skipping JWT regex on empty values
- keep non-empty invalid secrets rejected
- add a regression unit test for both cases (`tests/unit/auth-providers-form-validation.test.ts`)

Local verification run:
- `pnpm exec vitest run tests/unit/auth-providers-form-validation.test.ts` (passes)

7ttp (commented):
hi @abdulbaqui17 thanks for this!

but unfortunately, this most likely will get landed in: https://github.com/supabase/supabase/pull/44386

thanks again for ur efforts towards making Supabase better!! 💚😅

PR #44383, opened 2026-03-31
fix: Dont fetch databases and migrations for coming up projects

kamilogorek:
Those queries will always error out anyway as there's no data in place yet.


alaister (approved):
LGTM!

- Tested on preview

PR #44385, opened 2026-03-31
Joshen/fe 2921 extension integrations need to check for default schemas

joshenlim:
## Context

For database extensions, previously dashboard would fire a separate call just to retrieve the "default schema" for an extension via `useDatabaseExtensionDefaultSchemaQuery` from the `pg_available_extension_versions` table (the `schema` from this table implies where the extension will be installed in)

## Changes involved

Am updating the `useDatabaseExtensionsQuery` to use a custom studio SQL that will fetch this data in one request via a `LEFT JOIN`, so dashboard no longer needs to fire a request to `pg_available_extension_versions` each time we open the `EnableExtensionModal` since all the info we need is loaded up front.

Have also validated that the cost of the custom studio SQL is low (6.8, via explain analyze) so performance wise on the project's DB should be okay.

This will then also allow us to correctly render the "default schema" of the extensions in the new Install Integration Sheet now that we have that information up front.

## Misc fix

Also fixed a small issue on the database extensions page whereby if you searched for an extension that's hidden (e.g pg_tle), there's no "No results" UI state showing up
image

## Summary by CodeRabbit

* **New Features**
  * Per-extension default schema detection surfaced across install flows; default schema options added to selectors when applicable.

* **Bug Fixes**
  * Hidden extensions filtered out earlier so they no longer appear in lists.
  * Install button now correctly disables when required extensions are missing.

* **Refactor**
  * Consolidated extensions metadata retrieval and simplified schema selection/validation logic; UI text formatting standardized.


djhi (approved):

soedirgo (approved):

joshenlim:
im just gonna add some e2e tests here just to be sure before i merge

packages/pg-meta/src/sql/studio/database/extensions.ts
```diff
@@ -1,14 +1,23 @@
...
-select name, version, schema from pg_available_extension_versions where name = ${literal(extension)} limit 1;
+export const getDatabaseExtensionsSQL = () =>
+  `
+SELECT
+  e.name,
+  n.nspname AS schema,
+  e.default_version,
+  x.extversion AS installed_version,
+  e.comment,
+  ev.schema AS default_schema
```
soedirgo: I think this should be `default_version_schema`. Postgres extensions don't have default schemas, the schema is either explicit (i.e. it's required to be in that schema) or it can be installed in any schema (default being the first schema in `search_path`)

PR #44386, opened 2026-03-31
fix: make Apple OAuth client secret optional for native sign-in

cemalkilic:

YES

Bug fix

 When enabling Apple Sign-in in Studio, the form requires a valid JWT secret key whenever a client ID is provided. This blocks users who only use Apple native sign-in (iOS, macOS, watchOS, tvOS), where only the client ID (bundle ID) is needed and no secret is required.

Resolves AUTH-1138

## What is the new behavior?

The secret key field is now optional, matching Google's provider behavior. JWT format validation still applies when a secret is provided, but leaving it empty is allowed. This supports native-only Apple sign-in configurations.

## Additional context

  The validation was simplified from two `.when` clauses (dependent on both `ENABLED` and `CLIENT_ID`) to a single `.when` (dependent only on `ENABLED`), matching the pattern used by the Google provider.


awaseem (approved):
LGTM

PR #44282, opened 2026-03-27
chore: add shadcn input-group components

djhi:
## Screenshots

On a number input with units:
image

focused state:
image

On a textarea:
image

PR #44387, opened 2026-03-31
fix inputs width and spacing

djhi:
## Problem

#44282 introduced a change in design for inputs used outside a form layout such as search inputs on the project list page

## Solution

Fix padding and spacing


fsansalvadore (approved):
looks good

PR #44388, opened 2026-03-31
Fix create s3 access keys

joshenlim:
## Context

Resolves FE-2923

There's an issue with creating s3 keys in the storage settings at the moment, where naming your key throws a form field validation error. This PR fixes that - not sure when this was introduced though, but the main fix was just using `field` from `render` instead of calling `register`

## To test

- [ ] Verify that you can create an s3 access key via the storage settings

### DIFF
--- a/apps/studio/components/interfaces/Storage/StorageSettings/CreateCredentialModal.tsx
+++ b/apps/studio/components/interfaces/Storage/StorageSettings/CreateCredentialModal.tsx
@@ -21,6 +21,7 @@ import {
   DialogTitle,
   DialogTrigger,
   Form_Shadcn_,
+  FormField_Shadcn_,
   Tooltip,
   TooltipContent,
   TooltipTrigger,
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Storage/StorageSettings/CreateCredentialModal.tsx
```diff
@@ -126,22 +127,10 @@ export const CreateCredentialModal = ({ visible, onOpenChange }: CreateCredentia
...
               </FormItemLayout>
               <FormItemLayout label={'Secret access key'} isReactForm={false}>
-                <Input
-                  className="input-mono"
-                  readOnly
-                  copy
-                  disabled
-                  value={createS3KeyData?.secret_key}
-                />
+                <Input className="input-mono" readOnly copy value={createS3KeyData?.secret_key} />
```
joshenlim, line 133: just removing `disabled` for better cosmetic - otherwise the text constrast is too low

ivasilov (approved):

alaister (approved):
LGTM!

- Tested on preview

PR #44394, opened 2026-03-31
fix: remove plan checks from SSO Config Query

ignaciodob:
`useOrgSSOConfigQuery` had an internal plan-based gate (`canSetupSSOConfig`) that permanently disabled the query for non-team/enterprise orgs. 

The fix removes the redundant plan gate from the query hook, plan-based access is already controlled upstream via the entitlement check in the component.


awaseem (approved):
Tested in staging!

PR #44397, opened 2026-03-31
fix(studio): coerce index advisor cost values to numbers

awaseem:
## Summary

- Coerces `before`/`after` cost values to `Number()` in `QueryPanelScoreSection` and `calculateImprovement` before any comparison or arithmetic
- Fixes contradictory index advisor display where correct cost numbers showed 0% improvement and wrong arrow direction

## Root Cause

When `index_advisor_result` is prefetched from the Reports SQL query (via `json_build_object`), cost values can arrive as strings instead of numbers. JavaScript string comparison is lexicographic, producing wrong results:

| Expression | Numbers | Strings |
|---|---|---|
| `after > before` (arrow) | `50 > 100` → `false` ✅ | `"50" > "100"` → `true` ❌ |
| `costBefore <= costAfter` (improvement calc) | `100 <= 50` → `false` ✅ | `"100" <= "50"` → `true` ❌ |

The direct fetch path (`retrieve-index-advisor-result-query.ts`) validates through Zod and is unaffected. Only the prefetched path lacks validation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numeric value handling in query performance calculations to ensure more accurate and reliable improvement metrics.

* **Refactor**
  * Enhanced type safety and numeric coercion for query performance score comparisons, resulting in more consistent and robust metric calculations.


alaister (approved):
LGTM!

PR #44404, opened 2026-03-31
fix(webhooks): unblock enable button in old integrations UI

awaseem:
## Summary
- Fixes the "Enable webhooks" button being greyed out in the non-marketplace UI
- PR #44277 added `requiredExtensions: ['pg_net']` for the new UI, which inadvertently caused the old UI's actions area to get disabled via `hasToInstallExtensions`
- Skips the disabled styling when `hideRequiredExtensionsSection` is set, since webhooks handles `pg_net` installation server-side

Companion to #44402 which fixed the same issue in the new (V2) UI.


joshenlim (approved):
approving to unblock

PR #44447, opened 2026-04-01
feat(studio): add admonition for public bucket rls

kemaldotearth:

YES

Similar to the Advisor, this adds an admonition enabling the user to disable a public RLS select via dashboard. 

## Summary by CodeRabbit

* **New Features**
  * Detect public buckets that have removable SELECT policies and show a contextual warning on the bucket page.
  * Let users remove the policy via a confirmation dialog that previews the removal action.
  * Show success/error feedback and automatically refresh storage views after removal.
  * Adjust page layout to surface the warning above the storage explorer.


apps/studio/data/storage/public-buckets-with-select-policies-query.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+async function getPublicBucketsWithSelectPolicies({
+  projectRef,
+  connectionString,
+}: PublicBucketsWithSelectPoliciesVariables) {
+  const { result } = await executeSql<PublicBucketWithSelectPolicy[]>({
+    projectRef,
+    connectionString,
+    sql: `
+      SELECT DISTINCT b.id AS bucket_id, b.name AS bucket_name, p.policyname
```
charislam: For users with tens of thousands of buckets, this could be a heavy query -- it lists every bucket, and for every bucket iterates through the policies list with an unindexable join. Since we only show the admonition on each individual bucket page, we could fetch per-bucket as needed

charislam (changes requested):

charislam (approved):

apps/studio/components/interfaces/Storage/PublicBucketWarning.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  const { policyName, isRemovingPolicy, onRemovePolicy, isModalVisible, onShowModal, onHideModal } =
+    props
+
+  return (
+    <>
+      <Admonition
+        type="warning"
+        layout="horizontal"
+        title="Object listing is enabled on this bucket"
```
dnywh: Title idea in a bit simpler English for our less technical users: “Anyone can list all files in this bucket”.

Description is good but here’s another option:

> A `SELECT` policy on `storage.objects` allows clients to retrieve a full list of files. Public buckets don’t need this and it may expose more data than intended.

dnywh:
Restacking note: I've now moved #44372 to sit on top of this branch so the ownership split is clearer for reviewers.

- #44447 owns the bucket-page warning flow: the inline admonition, the single-bucket lookup, and the one-click single-policy removal path
- #44372 now only carries the extra global layer on top: the project-level listable-bucket query, homepage/Advisor/top-nav signal surfacing, dismissals, and related tests

Once this PR lands, #44372 can be retargeted back to `master`.

apps/studio/components/interfaces/Storage/PublicBucketWarning.tsx
kemaldotearth: Cheers for this, updating the copy!

PR #44457, opened 2026-04-01
fix(docs): clarify Free Plan 500 MB limit is database size, not disk size

Nedunchezhiyan-M:
## Summary

- The Free Plan section in `database-size.mdx` was ambiguous — it said "when you exceed the 500 MB limit" without specifying which limit
- Clarified that the 500 MB threshold is the **database size** limit (actual Postgres data), while Free Plan projects include 1 GB of total disk space
- This explains why read-only mode triggers before total disk usage is reached

Closes #43487

## Test plan

- [ ] Documentation renders correctly at `/docs/guides/platform/database-size`
- [ ] Wording is unambiguous about database size vs disk size distinction

## Summary by CodeRabbit

* **Documentation**
  * Clarified Free Plan read-only mode: the 500 MB threshold applies to database size (Postgres data), not overall disk usage.
  * Made Pro Plan instructions explicit: upgrading increases the database size quota rather than a generic disk limit.

### DIFF
--- a/apps/docs/content/guides/platform/database-size.mdx
+++ b/apps/docs/content/guides/platform/database-size.mdx
@@ -109,9 +109,9 @@ Due to restrictions on the underlying cloud provider, disk modifications are lim
 
 ### Free Plan behavior
 
-Free Plan projects enter [read-only](#read-only-mode) mode when you exceed the 500 MB limit. Once in read-only mode, you have these options:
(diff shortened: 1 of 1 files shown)


monicakh (approved):

ChrisChinchilla (approved):
Thanks @Nedunchezhiyan-M

PR #44470, opened 2026-04-01
feat: refresh tax preview on address change

ignaciodob:
## Summary

- Passes billing address and tax ID from the payment form to the subscription preview endpoint, so taxes are recalculated live as the user updates their details
- Debounces address/tax ID changes (1s) in `NewPaymentMethodElement` to avoid excessive API calls while typing
- Decouples the preview refetch from the payment form's mount state - uses `keepPreviousData` so the form stays mounted and the breakdown dims with `opacity-50` instead of unmounting/remounting during refetches. Shimmer skeleton only shows on initial load.
- Disables the "Confirm upgrade" button while a refetch is in progress to prevent submitting with stale tax data
- Respects the "Use address as my org's billing address" checkbox: the preview should mirror what will actually happen - if the checkbox is unchecked, the address won't be saved to Orb, so it shouldn't be used for the tax estimate either. Otherwise the user sees one price and gets charged another. The logic for this lives in `PlanUpdateSidePanel`

``` mermaid
flowchart TD
    A[User opens upgrade dialog] --> B[Fetch subscription preview]
    B --> C[Show payment form + price breakdown]

    C --> D[User edits address or tax ID]
    D -->|1s debounce| E{Use as billing address?}

    E -->|Yes| F[Re-fetch preview with new address]
    E -->|No| G[Re-fetch preview without address override]

    F --> H[Update breakdown — form stays mounted]
    G --> H

    H --> C

```

## Test plan

- [x]  Open the plan upgrade dialog, verify the payment form and breakdown load normally on first open
- [x]  Change the billing address country - verify the breakdown dims briefly and updates with new tax amounts without the payment form unmounting
- [x]  Toggle the tax ID on/off and change its value - verify the preview refreshes after ~1s debounce
- [x]  Confirm the upgrade button is disabled while the preview is refetching
- [x]  Uncheck "Use address as my org's billing address" - verify the preview refetches without address/tax overrides
- [x]  Re-check the checkbox - verify the preview refetches again with the current form address
- [x]  Assert that adding a new billing address in the CreditTopUp form works and saves the address

## Summary by CodeRabbit

* **New Features**
  * Payment and subscription flows now propagate billing address and tax ID via new callbacks; subscription preview requests include these values and preserve prior results while fetching.
  * Preview updates are debounced to reduce noise; loading state disables confirm actions and visually dims charge breakdown.

* **UX**
  * Address input emits complete normalized address updates (empty second line cleared).
  * Tax ID input emits updates and explicit clears (null).
  * “Use address as my org's billing address” is now controlled and reports changes.

### DIFF
--- a/apps/studio/components/interfaces/Billing/Payment/PaymentMethods/NewPaymentMethodElement.tsx
+++ b/apps/studio/components/interfaces/Billing/Payment/PaymentMethods/NewPaymentMethodElement.tsx
@@ -85,12 +85,16 @@ export const NewPaymentMethodElement = forwardRef(
       currentAddress,
       currentTaxId,
       customerName,
+      onAddressChange,
+      onTaxIdChange,
     }: {
       email?: string | null | undefined
       readOnly: boolean
       currentAddress?: CustomerAddress | null
       currentTaxId?: CustomerTaxId | null
       customerName?: string | undefined
+      onAddressChange?: (address: CustomerAddress) => void
+      onTaxIdChange?: (taxId: CustomerTaxId | null) => void
     },
     ref
   ) => {
@@ -123,13 +127,25 @@ export const NewPaymentMethodElement = forwardRef(
       form.setValue('tax_id_name', name)
     }
 
-    const { tax_id_name } = form.watch()
+    const { tax_id_name, tax_id_value } = form.watch()
     const selectedTaxId = TAX_IDS.find((option) => option.name === tax_id_name)
 
     const [purchasingAsBusiness, setPurchasingAsBusiness] = useState(currentTaxId != null)
     const [stripeAddress, setStripeAddress] = useState<
       StripeAddressElementChangeEvent['value'] | undefined
     >(undefined)
+    useEffect(() => {
+      if (!onTaxIdChange) return
+      if (purchasingAsBusiness && selectedTaxId && tax_id_value) {
+        onTaxIdChange({
+          country: getEffectiveTaxCountry(selectedTaxId),
+          type: selectedTaxId.type,
+          value: tax_id_value,
+        })
+      } else {
+        onTaxIdChange(null)
+      }
+    }, [purchasingAsBusiness, selectedTaxId, tax_id_value, onTaxIdChange])
 
     const availableTaxIds = useMemo(() => {
       const country = stripeAddress?.address.country || null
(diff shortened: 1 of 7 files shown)


apps/studio/components/interfaces/Billing/Payment/PaymentMethods/NewPaymentMethodElement.tsx
```diff
@@ -123,13 +127,25 @@ export const NewPaymentMethodElement = forwardRef(
       form.setValue('tax_id_name', name)
     }
 
-    const { tax_id_name } = form.watch()
+    const { tax_id_name, tax_id_value } = form.watch()
```
ivasilov, line 130: You should use `use_Watch_shadcn` here, when the values change, it won't cause a rerender otherwise.

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
```diff
@@ -61,6 +63,26 @@ export const PlanUpdateSidePanel = () => {
   const [showUpgradeSurvey, setShowUpgradeSurvey] = useState(false)
   const [showDowngradeError, setShowDowngradeError] = useState(false)
   const [selectedTier, setSelectedTier] = useState<'tier_free' | 'tier_pro' | 'tier_team'>()
+  const [latestAddress, setLatestAddress] = useState<CustomerAddress>()
```
ivasilov, line 66: I'd recommend refactoring this into a Reach hook form and just passing the `form` around, might be simpler.

ivasilov (approved):

apps/studio/components/interfaces/Billing/Payment/PaymentMethods/NewPaymentMethodElement.tsx
ignaciodob: `form.watch()` only subscribes to fields registered with this react-hook-form instance (`tax_id_name`, `tax_id_type`, `tax_id_value`). The Stripe AddressElement isn't registered with the form so I think this is safe to leave as is

apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
ignaciodob: The main values here (address, tax ID) come from Stripe's `AddressElement` which has its own `onChange`, I don't think it can't be registered as a form field. We'd still need the debounce logic either way, so a form wouldn't simplify much here. Happy to refactor this in the future though if we need to!

staaldraad (approved):

PR #44494, opened 2026-04-02
feat(studio): adding upgrade button to header experiment

myerekapan:

YES

- Add an always-visible "Upgrade to Pro" button in the dashboard header for free-plan users (GROWTH-615)
- Button is gated behind a PostHog experiment (`headerUpgradeCta`) with `control` and `test` variants
- Experiment exposure is tracked for both variants; click events are tracked when the button is clicked
- Button reuses existing `UpgradePlanButton` component for routing, permissions, and billing logic

Arc 2026-04-02 16 36 22

## What is the new behavior?

Arc 2026-04-02 16 36 12

## Additional context

## Summary by CodeRabbit

* **New Features**
  * Added an upgrade call-to-action in the header and mobile navigation (visible on medium+ screens in platform builds) shown to free-plan users as part of a controlled experiment.
  * The CTA records experiment exposures and sends analytics for impressions and clicks, including the user's current plan, to measure engagement and upgrade interest.


charislam (approved):

packages/common/telemetry-constants.ts
```diff
@@ -3021,6 +3021,36 @@ export interface PricingPageExperimentExposedEvent {
...
 }
 
+/**
+ * User was exposed to the header upgrade CTA experiment in studio.
+ * GROWTH-615: always-visible upgrade button in dashboard header for free-plan users.
+ *
+ * @group Events
+ * @source studio
+ */
+export interface HeaderUpgradeCtaExperimentExposedEvent {
```
pamelachia: useTrackExperimentExposure is something we already have and actually the experiment exposure events in this file are mostly dead code, could we remove this and use the useTrackExperimentExposure instead?

apps/studio/components/layouts/Navigation/NavigationBar/MobileNavigationBar.tsx
```diff
@@ -91,6 +92,7 @@ const MobileNavigationBar = ({
               </button>
             </CommandMenuTrigger>
           )}
+          {IS_PLATFORM && <HeaderUpgradeButton />}
```
pamelachia, line 95: this is just a repeat of if (!IS_PLATFORM) return null in L28 of apps/studio/components/layouts/Navigation/LayoutHeader/HeaderUpgradeButton.tsx so keep this and remove the other or the other way around?

pamelachia (changes requested):

apps/studio/components/layouts/Navigation/LayoutHeader/HeaderUpgradeButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+  className?: string
+}
+
+export const HeaderUpgradeButton = ({ className }: HeaderUpgradeButtonProps) => {
+  const track = useTrack()
+  const { data: organization } = useSelectedOrganizationQuery()
+  const flagValue = usePHFlag<HeaderUpgradeCtaVariant | false>(EXPERIMENT_ID)
+
+  const isFreePlan = organization?.plan?.id === 'free'
+  const isInExperiment = flagValue === 'control' || flagValue === 'test'
```
seanoliver, line 20: ❤️

apps/studio/components/layouts/Navigation/LayoutHeader/HeaderUpgradeButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  if (!isFreePlan) return null
+  if (!showButton) return null
+
+  const handleClick = () => {
+    track('header_upgrade_cta_clicked', { currentPlan: organization?.plan?.id ?? 'free' })
+  }
+
+  return (
+    <div onClick={handleClick} className={className}>
```
seanoliver: The `onClick` is on a wrapping `div` rather than the button itself — this means:
1. Clicks on padding around the button will fire the telemetry event without actually navigating
2. It's a nested interactive element (`div[onClick]` > `Link`/`Button`), which is an accessibility concern
3. The click event fires independently from the actual navigation, so if `UpgradePlanButton` prevents navigation (e.g. permissions check, disabled state), the telemetry event still fires

Consider passing an `onClick` callback into `UpgradePlanButton` instead (it already accepts arbitrary props via `PropsWithChildren`), or tracking the click at the point where navigation actually occurs.

Also per my other comment below, let's make sure we use the backend billing event to actually measure successful conversions!

apps/studio/components/layouts/Navigation/LayoutHeader/HeaderUpgradeButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+  // Track experiment exposure for all free-plan users in the experiment (both control and test)
+  const variant = isFreePlan && isInExperiment ? (flagValue as string) : undefined
+  useTrackExperimentExposure(EXPERIMENT_ID, variant)
+
+  if (!isFreePlan) return null
+  if (!showButton) return null
+
+  const handleClick = () => {
+    track('header_upgrade_cta_clicked', { currentPlan: organization?.plan?.id ?? 'free' })
+  }
```
seanoliver, line 32: `currentPlan` will always be `'free'` here — line 27 already returns `null` if `!isFreePlan`, so by the time this click handler can fire, the plan is guaranteed to be `'free'`. If there's a strong reason to keep it, it might be worth leaving a comment for clarity, otherwise, I'd recommend just removing it.

apps/studio/components/layouts/Navigation/LayoutHeader/HeaderUpgradeButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+type HeaderUpgradeCtaVariant = 'control' | 'test'
+
+interface HeaderUpgradeButtonProps {
+  className?: string
+}
+
+export const HeaderUpgradeButton = ({ className }: HeaderUpgradeButtonProps) => {
+  const track = useTrack()
+  const { data: organization } = useSelectedOrganizationQuery()
+  const flagValue = usePHFlag<HeaderUpgradeCtaVariant | false>(EXPERIMENT_ID)
```
seanoliver, line 17: nit: Before the org query and PostHog flag resolve, `flagValue` will be `undefined` and `isFreePlan` will be `false`, so the component returns `null` briefly. For a test-group user this means a tiny flash where the button isn't rendered before it appears. Probably fine 99% of the time since the button takes minimal space and the queries resolve fast, but worth being aware of if you notice any layout shift in testing.

packages/common/telemetry-constants.ts
```diff
@@ -3021,6 +3021,21 @@ export interface PricingPageExperimentExposedEvent {
...
 }
 
+/**
+ * User clicked the "Upgrade to Pro" CTA in the dashboard header.
+ * GROWTH-615: always-visible upgrade button in dashboard header for free-plan users.
+ *
+ * @group Events
+ * @source studio
+ */
+export interface HeaderUpgradeCtaClickedEvent {
```
seanoliver, line 3031: Not feedback for this PR, but flagging for awareness: I think we should use `organization_subscription_updated` (the backend billing event implemented in the `platform` repo) as the conversion metric for this test, rather than a button click event.

You may already be planning this, but I didn’t see it mentioned in the PR description and wanted to make sure it’s considered.

apps/studio/components/layouts/Navigation/NavigationBar/MobileNavigationBar.tsx
seanoliver: This looks good now — the `IS_PLATFORM` guard was removed from `HeaderUpgradeButton` itself since in `LayoutHeader.tsx` it's already inside the `IS_PLATFORM` ternary (line 229). The explicit `{IS_PLATFORM && <HeaderUpgradeButton />}` in `MobileNavigationBar` is still there since that component doesn't have a parent guard. So both call sites are covered.

pamelachia (approved):

seanoliver (approved):
LGTM! Approving to unblock. The only thing I'd consider maybe fixing is attaching the onClick event to a div wrapper rather than to the button itself. It's partially a semantic thing but also I've seen it lead to data quality issues depending on how the layout renders.

PR #44515, opened 2026-04-03
fix(studio): allow selecting columns for empty-cell NULL conversion on CSV import

usamanadeemdeveloper:

Bug fix / enhancement to an existing Studio import flow.

Studio currently supports converting empty CSV cells to `NULL`, but the behavior applies across all imported columns. That makes it impossible to preserve empty strings in some columns while still treating empty cells as `NULL` in others.

This follows up on the earlier CSV import fix discussed in #43281, which addressed the underlying issue reported in #43258.

## What is the new behavior?

This PR updates the CSV import flow so users can choose exactly which imported columns should convert empty cells to `NULL`.

The import flow now:
- keeps all imported columns selected by default for backwards-compatible behavior
- lets users narrow that selection down to specific columns
- keeps preview behavior aligned with the actual inserted data

## Additional context

I manually verified the full create-table-from-CSV flow locally in Studio, including saving the imported data and confirming that only the selected columns were persisted as `NULL` while other empty cells remained empty values.

Formatting was checked with Prettier.
A direct production build run was started locally, but the Next.js build process remained running without producing a final result during this session, so I am not claiming a completed local build verification here.

## Related issues

- Fix follow-up for #43281
- Original issue: #43258

## Release Notes

* **New Features**
  * Added a "Set empty cells as NULL" selector during spreadsheet import, allowing users to specify which columns should treat empty strings as NULL values rather than converting all nullable columns automatically.
  * Enhanced import preview to display NULL representation for selected empty-string columns.

* **Tests**
  * Added test coverage for selective empty-string-to-NULL conversion behavior.

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
@@ -842,7 +842,7 @@ export const SidePanelEditor = ({
       return console.error('no project or table selected')
     }
 
-    const { file, rowCount, selectedHeaders, resolve } = importContent
+    const { file, rowCount, selectedHeaders, emptyStringAsNullHeaders, resolve } = importContent
     const toastId = toast.loading(
       `Adding ${rowCount.toLocaleString()} rows to ${selectedTable.name}`
     )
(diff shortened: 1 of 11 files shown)


usamanadeemdeveloper:
@charislam this is the follow-up enhancement we discussed on #43281, which was the fix for #43258.

This update keeps the original empty-cell-to-`NULL` support, but narrows it so users can choose exactly which imported columns should treat empty CSV cells as `NULL` while preserving empty values in the others.

@mintydev789 this now supports the selective-column behavior you suggested.
feature_tutorial

mintydev789:
@usamanadeemdeveloper thank you!

charislam (approved):

PR #44518, opened 2026-04-03
docs: fix SSR cookie header omission in OAuth PKCE flow (fixes #30900)

### LINKED ISSUE #30900: Subtle bug in Remix guide

King-Austin:

Docs fix.

The shared server-side OAuth snippet in the PKCE flow docs redirects using `data.url` but does not forward `Set-Cookie` HTTP headers that are captured by `@supabase/ssr`.

In frameworks such as Remix, this drops the PKCE `code_verifier` cookie, resulting in an uncatchable authentication failure during the callback endpoint exchange flow.

## What is the new behavior?

- Updates the shared server snippet in `oauth_pkce_flow.mdx` to use `createServerClient` and demonstrate capturing the `Set-Cookie` in a `Headers` instance.
- Explicitly passes `{ headers }` into the `redirect(data.url)` to show the PKCE cookie preservation requirement.
- Simplifies guidance for Server-Side authentication.

## Additional context

Closes #30900

## Summary by CodeRabbit

* **Documentation**
  * Updated Server-Side Auth example in OAuth PKCE flow: now demonstrates SSR-oriented client usage, explicit cookie serialization/handling, and performing redirects that include constructed headers for correct cookie propagation.

### DIFF
--- a/apps/docs/content/_partials/oauth_pkce_flow.mdx
+++ b/apps/docs/content/_partials/oauth_pkce_flow.mdx
@@ -1,5 +1,3 @@
----
-
 For a PKCE flow, for example in Server-Side Auth, you need an extra step to handle the code exchange. When calling `signInWithOAuth`, provide a `redirectTo` URL which points to a callback route. This redirect URL should be added to your [redirect allow list](/docs/guides/auth/redirect-urls).
 
 <Tabs
--- a/apps/docs/content/_partials/oauth_pkce_flow.mdx
+++ b/apps/docs/content/_partials/oauth_pkce_flow.mdx
@@ -1,3 +1,5 @@
+---
+
 For a PKCE flow, for example in Server-Side Auth, you need an extra step to handle the code exchange. When calling `signInWithOAuth`, provide a `redirectTo` URL which points to a callback route. This redirect URL should be added to your [redirect allow list](/docs/guides/auth/redirect-urls).
 
 <Tabs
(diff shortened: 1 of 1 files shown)


aantti:
Closing as a duplicate of #30900's fix in #50722, which is being finalized. Thanks for the report and the fix - the root cause was correct. Note for anyone reading: stubs above `// ---cut---` in this partial render verbatim, since `@supabase/ssr` isn't resolvable by the docs type-checker, so the final version avoids the cut marker.

PR #44557, opened 2026-04-06
feat(docs): Allow opening search links in new tab

thewheat:

YES

A simple way to allow for opening doc search results in a new tab

image

## Summary by CodeRabbit

* **Improvements**
  * Documentation search now respects Ctrl/Cmd-click to open links in a new tab.
  * Documentation, reference, and troubleshooting pages open in-app or in a new tab depending on modifier key; same-tab navigation closes the search menu, new-tab does not.
  * Integration and external discussion links follow the same modifier-key behavior; the search menu only closes for same-tab navigation.

### DIFF
--- a/packages/ui-patterns/src/CommandMenu/prepackaged/DocsSearch/DocsSearchPage.tsx
+++ b/packages/ui-patterns/src/CommandMenu/prepackaged/DocsSearch/DocsSearchPage.tsx
@@ -100,33 +100,57 @@ const DocsSearchPage = () => {
   )
 
   async function openLink(pageType: PageType, link: string) {
+    // A simple way to achieve opening links in new tab but room for improvement including support for middle clicks
+    const openInNewTab = window.event?.metaKey || window.event?.ctrlKey;
+
     switch (pageType) {
       case PageType.Markdown:
       case PageType.Reference:
       case PageType.Troubleshooting:
         if (BASE_PATH === '/docs') {
-          router.push(link)
-          setIsOpen(false)
+          if(openInNewTab){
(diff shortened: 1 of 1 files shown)


packages/ui-patterns/src/CommandMenu/prepackaged/DocsSearch/DocsSearchPage.tsx
```diff
@@ -100,33 +100,55 @@ const DocsSearchPage = () => {
...
+            setIsOpen(false)
+          }
         }
         break
       case PageType.GithubDiscussion:
         window.open(link, '_blank', 'noreferrer,noopener')
-        setIsOpen(false)
+        if (!openInNewTab) {
+          setIsOpen(false)
+        }
```
jeremenichelli: Ideally we would just use links as Tim Berners-Lee invision this in 1989 😆 but this is of course old approach we are living with. We are also "kinda" close to re-doing this.

In the meantime I am fine with patching it, but let's improve it a bit. We are doing a lot this:

```js
if (openInNewTab) {
  window.open(`/docs${link}`, '_blank', 'noreferrer,noopener')
} else {
  router.push(`/docs${link}`)
  setIsOpen(false)
}
```

I see this is because sometimes we are modifying the final link value.

I think we could create a variable to build the final link string. and then at the end of the switch statement just do one of those if conditional blocks.

Let me know if this is clear enough @thewheat :)

jeremenichelli (approved):

fsansalvadore (approved):
Tested and seems to be working as intended in the preview 👍

PR #44559, opened 2026-04-06
fix: do not query backups for project in unknown state

kevcodez:
Unknown state is explicitly set before a project is coming up, leading to a bunch of invalid backups list requests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the backups feature to properly handle cases where the project status is unknown. The system now avoids executing backup-related queries when the project state is indeterminate, resulting in more stable and predictable behavior while preventing potential issues that could arise from executing operations during uncertain project states.


alaister (approved):
LGTM!

- Tested on preview

PR #44565, opened 2026-04-06
fix(studio): only render MobileNavigationBar on mobile viewports

alaister:
MobileNavigationBar was always mounted on desktop, triggering unnecessary org/project queries even though it's visually hidden with `md:hidden`. Now only rendered when the viewport is at or below the md breakpoint.

**Changed:**
- Conditionally render `MobileNavigationBar` using `useBreakpoint('md')` so it only mounts on mobile

## To test

- Open Studio on a desktop viewport – verify the mobile nav bar is not in the DOM and no unnecessary requests fire
- Resize to a mobile viewport – verify the mobile nav bar appears and works normally

## Bug Fixes
* Mobile navigation bar now correctly displays only on mobile devices based on screen size, improving responsive design behavior.


fsansalvadore (approved):
👍

ivasilov (approved):

PR #44588, opened 2026-04-06
fix(studio): respect enabled prop in index advisor query hook

oniani1:
Fixes #44586

The `enabled` field in `useGetIndexAdvisorResult` had an OR branch that bypassed the `enabled` prop for queries starting with `with pgrst_source`. This meant the query could fire even when `enabled: false` was passed (e.g. when the index_advisor extension isn't installed or the result is already prefetched).

Restructures the logic to match the sibling hook in `retrieve-index-from-select-query.ts`, which correctly ANDs `enabled` with all conditions using an extracted `isValidQueryForIndexing` variable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query validation in index advisor to more reliably handle SQL query normalization and ensure consistent query processing.


awaseem (approved):
LGTM thanks for the contribution <3

PR #44589, opened 2026-04-06
fix(studio): escape SQL identifiers in index create/delete mutations

oniani1:
Fixes #44587

index-delete-mutation.ts and index-create-mutation.ts construct SQL using raw string interpolation for identifiers (schema, table, column names) without escaping double quotes. Uses ident() from @supabase/pg-meta/src/pg-format, same utility already used in the database-queues mutations.

## Summary by CodeRabbit

* **Refactor**
  * Improved identifier quoting/escaping for creating and dropping database indexes to ensure consistent, safer index operations.


oniani1:
pushed a follow-up commit that validates \ against an allowlist of supported index methods (btree, hash, gist, spgist, gin, brin). The list matches what's already defined in \.

charislam (approved):
Thanks @oniani1!

PR #44597, opened 2026-04-06
feat(www): show confirmation message after newsletter signup

stylessh:

YES

Feature

After subscribing to the newsletter in the footer, users only see "Thanks for subscribing!" with no indication of what to expect next.

## What is the new behavior?

After a successful newsletter signup, a secondary message is shown: "You'll hear from us when we publish our next newsletter issue." This sets expectations that they will receive emails when new issues are published.

## Additional context

Small UX improvement to the footer newsletter form success state.

## Release Notes

* **New Features**
  * Enhanced newsletter subscription confirmation message with additional details about when the next issue will be published.


dnywh (approved):
Tested on staging

PR #44637, opened 2026-04-07
fix: remove API Docs nav item after transition period

awaseem:
## Summary

- Removes the "API Docs" navigation item from the sidebar and mobile menu
- Removes the `UI_PREVIEW_API_SIDE_PANEL` feature preview flag since the feature is fully rolled out
- Makes API docs buttons unconditionally visible across Auth Users, Storage, Edge Functions, and SecondLevelNav

## Test plan

- [x] `NavigationBar.utils` tests pass (26 tests)
- [x] `FileExplorerHeader` tests pass (6 tests)
- [x] TypeScript compiles with no errors
- [ ] Verify sidebar no longer shows "API Docs" nav item
- [ ] Verify API docs buttons still appear in Auth Users, Storage, and Edge Functions pages
- [ ] Verify feature preview modal no longer lists "Project API documentation"

Resolves FE-2759

## Summary by CodeRabbit

* **New Features**
  * APIDocs button can optionally display a label and use a custom tooltip.

* **Chores**
  * Removed the API docs side-panel feature flag and its localStorage key.
  * “API Docs” navigation entry removed; sidebar no longer special-cases that route.
  * Back links and API Docs buttons now render consistently across the app (no flag gating).

* **Tests**
  * Tests updated to stop depending on the removed feature-flag behavior.


apps/studio/components/layouts/EdgeFunctionsLayout/EdgeFunctionDetailsLayout.tsx
```diff
@@ -332,16 +330,16 @@ const EdgeFunctionDetailsLayout = ({
...
-                  <APIDocsButton
-                    section={
-                      functionSlug !== undefined
-                        ? ['edge-functions', functionSlug]
-                        : ['edge-functions']
-                    }
-                    source="edge-functions"
-                  />
-                )}
+                <APIDocsButton
```
charislam: This one should be removed rather than made visible everywhere, I think

charislam (approved):

PR #44438, opened 2026-04-01
feat(studio): add advisor for public bucket rls

kemaldotearth:
## I have read the [CONTRIBUTING.md](https://github.com/supabase/supabase/blob/master/CONTRIBUTING.md) file.

YES

## What kind of change does this PR introduce?

This connects the Advisor from the splinter repo. Below are the list of changes:

- Register advisor which should appear in both the legacy Advisors and Advisors sidebar.
- Adds a "Dismiss" button to the admonition inside the bucket view.
- Makes the check for select policy on public buckets tiny bit stricter/truer. 

This is awaiting the [PR](https://github.com/supabase/splinter/pull/152) in splinter going in!

## Summary by CodeRabbit

* **New Features**
  * Added a security lint that flags public storage buckets allowing listing, with a direct "View bucket" link.
  * Users can dismiss public-bucket warnings per project+bucket for 14 days via a Dismiss button.

* **Improvements**
  * Tightened policy detection to better target bucket-scoped select policies and avoid unrelated matches.

* **Telemetry**
  * Added events for policy removal and dismiss-button clicks.

PR #44648, opened 2026-04-08
chore(studio): wire public bucket listing advisor lint

dnywh:

Chore.

The new `public_bucket_allows_listing` warning exists in Splinter, but Studio's embedded advisor SQL does not surface it in the `supabase/supabase` flow yet. That means local and self-hosted paths can miss the warning even though the underlying lint now exists, and Studio also lacks the lint metadata needed to label it cleanly and link to the affected bucket.

## What is the new behavior?

This wires the new `public_bucket_allows_listing` advisor query into the embedded pg-meta lint SQL and adds the Studio lint metadata entry so the warning renders with a storage bucket CTA.

This keeps the global/project-wide Advisor behaviour aligned with the new Splinter source of truth while preserving the existing inline bucket-page admonition.

## How has this been tested?

- Applied the same SQL logic that now ships in `supabase/splinter`.
- Smoke-tested the Studio integration patch in a throwaway local worktree.
- Confirmed the new lint metadata entry resolves cleanly after fixing the `bucket_id` metadata access in the CTA link.
- Manual product test to run on local or staging:

```sql
insert into storage.buckets (id, name, public)
values ('listable.bucket+1', 'listable.bucket+1', true);

create policy "listable_bucket_select"
on storage.objects
for select
to authenticated
using (bucket_id = 'listable.bucket+1');
```

Expected result:
- Security Advisor shows one warning for that bucket.
- The bucket page continues to show the existing inline admonition.

Cleanup:

```sql
drop policy "listable_bucket_select" on storage.objects;
```

Expected result:
- The Advisor warning disappears.
- The bucket remains public, but no longer counts as listable.

### DIFF
--- a/apps/studio/components/interfaces/Linter/Linter.utils.tsx
+++ b/apps/studio/components/interfaces/Linter/Linter.utils.tsx
@@ -320,6 +320,21 @@ export const lintInfoMap: LintInfo[] = [
     docsLink: `${DOCS_URL}/guides/database/database-linter?lint=0024_permissive_rls_policy`,
     category: 'security',
   },
+  {
+    name: 'public_bucket_allows_listing',
+    title: 'Public Bucket Allows Listing',
+    icon: <Box className="text-foreground-muted" size={15} strokeWidth={1.5} />,
+    link: ({ projectRef, metadata }) => {
+      const bucketId = (metadata as Record<string, string | undefined> | undefined)?.bucket_id
+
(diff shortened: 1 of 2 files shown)


dnywh:
Closing in favour of @kemaldotearth’s https://github.com/supabase/supabase/pull/44438/

Moving some of the stuff from this PR over there:

- Use `bucket_id` not `name` for the link target (they can differ)
- Wrap with `encodeURIComponent` to handle special characters in bucket IDs

PR #44650, opened 2026-04-08
fix(assistant): skip DPA signed check when tracing is disabled

pamelachia:
## Summary
- Skip the `GET /dpa-signed` API call in `getOrgAIDetails()` when `IS_TRACING_ENABLED` is false
- The endpoint requires org-level permissions, but gets called for every user on every AI assistant request, generating ~43 403 errors/day (80% of all DPA errors)
- Since `isDpaSigned` is only consumed by `isTracingAllowed()`, and tracing is currently disabled in production, the call is unnecessary
- When skipped, `isDpaSigned` defaults to `undefined`, which `isTracingAllowed()` already treats as "tracing not allowed" (privacy-protective default)

## Context
- Investigation: https://app.hex.tech/supabase/thread/019d6b95-4af6-777d-a37e-242204bcb0f8
- Linear: GROWTH-755
- The 403s come from project-scoped users (org role `None`) hitting the `ProjectPermissionsDisabledFeatureGuard`

## Test plan
- [x] Existing tests pass with `IS_TRACING_ENABLED` mocked as `true` (no behavior change when tracing enabled)
- [x] New test verifies DPA call is skipped when `IS_TRACING_ENABLED` is `false`
- [x] `isDpaSigned` returns `undefined` when call is skipped (correct for tracing-disabled path)

## Summary by CodeRabbit

* **Tests**
  * Added coverage for conditional DPA retrieval, including a test that simulates tracing being disabled to verify no DPA fetch occurs and resulting state remains unset.

* **Bug Fixes**
  * DPA status fetch now respects tracing settings and safely handles the no-fetch case, preventing unnecessary network calls and avoiding undefined access.

### DIFF
--- a/apps/studio/lib/ai/ai-details.ts
+++ b/apps/studio/lib/ai/ai-details.ts
@@ -6,6 +6,7 @@ import { getOrganizations } from '@/data/organizations/organizations-query'
 import { getProjectDetail } from '@/data/projects/project-detail-query'
 import { getOrgSubscription } from '@/data/subscriptions/org-subscription-query'
 import { getAiOptInLevel } from '@/hooks/misc/useOrgOptedIntoAi'
+import { IS_TRACING_ENABLED } from '@/lib/ai/braintrust-logger'
 
 export const getOrgAIDetails = async ({
   orgSlug,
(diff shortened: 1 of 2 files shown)


apps/studio/lib/ai/ai-details.ts
```diff
@@ -19,14 +20,18 @@ export const getOrgAIDetails = async ({
     ...(authorization && { Authorization: authorization }),
   }
 
+  const dpaPromise = IS_TRACING_ENABLED
```
mattrossman, line 23: Note in #44644 we're about to remove the production kill switch, so `IS_TRACING_ENABLED` will soon be true even for users who aren't owners.

I'm thinking until the "DPA signed" check is removed in ~2 weeks, that we'll want to condition the frontend request on whether the user is an org owner instead.

pamelachia: closed this!

PR #44651, opened 2026-04-08
chandana-postgrest-multiple-schema-objects

canumula:

YES

New troubleshooting guide for PostgREST

## Summary by CodeRabbit

* **Documentation**
  * Added a troubleshooting guide explaining why unqualified object names can fail when multiple schemas are exposed and how to resolve it (reorder schemas, initialize the client with a schema, or set schema per request).
  * Added a troubleshooting entry for the "schema pg_pgrst_no_exposed_schemas does not exist" PostgREST error, describing typical logs, causes, and when it appears if the Data API is disabled.

### DIFF
--- a/apps/docs/content/troubleshooting/postgrest-not-recognizing-objects-in-schema
+++ b/apps/docs/content/troubleshooting/postgrest-not-recognizing-objects-in-schema
@@ -16,7 +16,7 @@ Could not find the table 'X' in the schema cache
 
 ## Why this happens
 
-This happens because when multiple schemas are exposed, the first entry in the list is treated as the default, so PostgREST resolves unqualified object names only against that schema. 
+When multiple schemas are exposed, the first entry in the list is treated as the default, so PostgREST resolves unqualified object names only against that schema. 
 
 This behavior is intentional, as it prevents conflicts when different schemas contain objects with the same name.
 
--- a/apps/docs/content/troubleshooting/postgrest-not-recognizing-objects-in-schema
+++ b/apps/docs/content/troubleshooting/postgrest-not-recognizing-objects-in-schema
@@ -1,0 +1,23 @@
+---
+title = "PostgREST not recognizing the objects(tables/functions/views) in a schema even after adding it to the exposed schemas"
+topics = [
+  "database",
+  "rest-api",
+  "platform"
+]
+keywords = []
+---
+
+PostgREST is returning errors by not recognizing the objects(tables/tables/views) in a schema and logging errors similar to:
+
+```console
+Could not find the table 'X' in the schema cache
+```
+
+## Why this happens
+
+This happens because, when you have multiple schemas, the default schema will be the first in the list of exposed schemas.
+
+## How to fix this
+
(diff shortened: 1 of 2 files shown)


apps/docs/content/troubleshooting/schema-pg_pgrst_no_exposed_schemas-does-not-exist
```diff
@@ -0,0 +1,20 @@
...
+
+PostgREST logs showing the error:
+
+- `schema "pg_pgrst_no_exposed_schemas" does not exist`
+
+## Why this happens
+
+- This error means that the Data API is disabled on the Supabase project.
+- We don’t really shutdown postgrest when data api is disabled. You will see a schema called `pg_pgrst_no_exposed_schemas` added in the exposed schemas.
+- This should not adversely affect the project, although it may result in additional entries in your logs.
```
mkleczek, line 20: Worth adding a sentence that we're aware of the issue and are working on it?

apps/docs/content/troubleshooting/schema-pg_pgrst_no_exposed_schemas-does-not-exist
```diff
@@ -0,0 +1,20 @@
...
+message = "schema pg_pgrst_no_exposed_schemas does not exist"
+---
+
+PostgREST logs showing the error:
+
+- `schema "pg_pgrst_no_exposed_schemas" does not exist`
+
+## Why this happens
+
+- This error means that the Data API is disabled on the Supabase project.
```
mkleczek: I think this needs to be reworded. I find it awkward that it sounds like we are saying "it is a feature not a bug".

The fact is that this is a problem on our side in the way we implemented disabling Data API.

apps/docs/content/troubleshooting/postgrest-not-recognizing-objects-in-schema
```diff
@@ -0,0 +1,25 @@
...
+  "database",
+  "rest-api",
+  "platform"
+]
+keywords = []
+---
+
+PostgREST is returning errors by not recognizing the objects(tables/functions/views) in a schema and logging errors similar to:
+
+```console
```
jeremenichelli: I would remove the `console`

apps/docs/content/troubleshooting/postgrest-not-recognizing-objects-in-schema
```diff
@@ -0,0 +1,25 @@
...
+Could not find the table 'X' in the schema cache
+```
+
+## Why this happens
+
+This happens because when multiple schemas are exposed, the first entry in the list is treated as the default, so PostgREST resolves unqualified object names only against that schema.
```
jeremenichelli: Let's avoid redundancy.


```suggestion
When multiple schemas are exposed, the first entry in the list is treated as the default, so PostgREST resolves unqualified object names only against that schema. 
```

apps/docs/content/troubleshooting/schema-pg_pgrst_no_exposed_schemas-does-not-exist
```diff
@@ -0,0 +1,21 @@
+---
+title = "schema "pg_pgrst_no_exposed_schemas" does not exist"
```
jeremenichelli: The `"` inside are breaking the parser. Please test this locally before pushing 🙏

jeremenichelli (approved):

PR #44674, opened 2026-04-08
refactor(pg-meta): expand safeSql usage

charislam:
## Summary by CodeRabbit

* **Refactor**
  * Standardized SQL construction across the pg-meta package to use parameter-safe SQL fragments instead of raw string assembly, improving safety for dynamic values (filters, limits, offsets, identifiers) and unifying how exported SQL constants and query helpers are produced. No functional query behavior changes expected.


awaseem (approved):
Tested table editor, triggers filtering and sorting and a few smoke tests. Seems to be fine!

PR #44690, opened 2026-04-09
fix: clean up grouped input border rendering

dnywh:

Bug fix. Resolves FE-2959.

Grouped inputs strip the inner control border with `border-0`, while the outer `InputGroup` is responsible for the visible outline. After #44703, grouped form inputs now receive the correct validation attributes, but the primitive still removes the inner border dimensions entirely. That keeps the visual layer brittle and can lead to inconsistent error rendering, especially in light mode where the grouped password field does not match the Project name input.

The shared `ui-patterns` input wrapper also inherits the generic `inline-end` button spacing from `InputGroupAddon`, which pushes the password copy button slightly too far to the right.

## What is the new behaviour?

The shared grouped-input primitive now keeps a transparent inner border instead of removing border widths entirely. The outer `InputGroup` continues to own the visible border, focus ring and error treatment, but the inner control now has stable border metrics and no redundant dark-mode background override.

The grouped invalid state now also matches the plain input treatment more closely:

- unfocused error border uses `destructive-400`
- focused error border uses `destructive`
- error background uses `destructive-200`

The `ui-patterns` input wrapper now overrides the inherited `inline-end` negative margin for its own copy/reveal button addon, so the password copy button sits in the right place without changing spacing for other direct `InputGroup` consumers.

| Before | After |
| --- | --- |
| CleanShot 2026-04-09 at 13 43 50@2x-91489355-5EEA-4884-BAA2-B93B28F1D7CB| CleanShot 2026-04-09 at 13 41 36@2x-5D7E930B-EDB2-4002-9367-34ACE2DF4DB4 |
| CleanShot 2026-04-09 at 14 06 15@2x-05379F74-4CA5-4E6E-9250-CBA2764C1318 | CleanShot 2026-04-09 at 14 12 11@2x-11FBC519-D398-4A59-971A-97FBFEF610DE |

An easy place to test this is the Database password field in the New Project creation flow. See also the Minimum password length field in Email (Sign In / Providers).

## Summary

- update `packages/ui/src/components/shadcn/ui/input-group.tsx`
- replace `border-0` on grouped inputs/textareas with transparent borders
- remove the redundant `dark:bg-transparent`
- keep the inner grouped controls visually neutral so the outer group owns the rendered border state
- align grouped invalid border and background styling with the plain input treatment
- update `packages/ui-patterns/src/DataInputs/Input.tsx` so copy/reveal buttons do not inherit the extra right pull intended for other `inline-end` addons

### DIFF
--- a/packages/ui-patterns/src/DataInputs/Input.tsx
+++ b/packages/ui-patterns/src/DataInputs/Input.tsx
@@ -88,7 +88,11 @@ const Input = forwardRef<
         />
         {icon && <InputGroupAddon align="inline-start">{icon}</InputGroupAddon>}
         {copy || actions ? (
-          <InputGroupAddon align="inline-end">
+          <InputGroupAddon
+            align="inline-end"
+            // Override defaults
+            className="pr-1 has-[>button]:mr-0 has-[>kbd]:mr-0"
+          >
             {copy && !(reveal && hidden) ? (
               <InputGroupButton
                 size="tiny"
--- a/apps/studio/components/interfaces/ProjectCreation/DatabasePasswordInput.tsx
+++ b/apps/studio/components/interfaces/ProjectCreation/DatabasePasswordInput.tsx
@@ -1,11 +1,11 @@
 import { UseFormReturn } from 'react-hook-form'
-import { FormControl_Shadcn_, FormField_Shadcn_ } from 'ui'
-import { Input } from 'ui-patterns/DataInputs/Input'
+import { FormControl_Shadcn_, FormField_Shadcn_, Input_Shadcn_ } from 'ui'
 import { FormItemLayout } from 'ui-patterns/form/FormItemLayout/FormItemLayout'
 
 import { DATABASE_PASSWORD_REGEX } from './ProjectCreation.constants'
 import { CreateProjectForm } from './ProjectCreation.schema'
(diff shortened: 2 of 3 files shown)


packages/ui-patterns/src/DataInputs/Input.tsx
```diff
@@ -88,7 +88,11 @@ const Input = forwardRef<
         />
         {icon && <InputGroupAddon align="inline-start">{icon}</InputGroupAddon>}
         {copy || actions ? (
-          <InputGroupAddon align="inline-end">
+          <InputGroupAddon
+            align="inline-end"
+            // Override defaults
+            className="pr-1 has-[>button]:mr-0 has-[>kbd]:mr-0"
```
dnywh, line 94: Do we really need the `has-[>button]:mr-[-0.4rem] has-[>kbd]:mr-[-0.35rem]` on the underlying component? I don’t think it’s realistic that consumers will remember to pass these overrides each time. Maybe there’s a better way?

dnywh: Might be okay if people are just importing this `ui-patterns` version.

djhi (approved):

SaxonF (approved):

PR #44704, opened 2026-04-09
docs(examples): use latest v2 of supabase-js and remove lock files

mandarini:
* Use latest 2.x version of `@supabase/supabase-js` in all examples
* Use `latest` version of `@supabase/ssr` in all examples
* Remove lock files from examples
* Add examples lock files in .gitignore

The rationale is:
- Lock files are not actively maintained/updated (or the versions in the package.json files for that matter)
- They pin an arbitrary version (from the end-user perspective)
- Removes the need to manually update the versions and reinstall
- Consistency

## Release Notes

* **Chores**
  * Broadened Supabase SDK dependency version constraints across example projects for greater flexibility with compatible updates
  * Updated some SSR package dependencies to track latest releases
  * Added gitignore rules for dependency lock files in example directories


examples/caching/with-react-query-nextjs-14/package.json
```diff
@@ -10,8 +10,8 @@
   },
   "dependencies": {
     "@supabase-cache-helpers/postgrest-react-query": "^1.3.0",
-    "@supabase/ssr": "^0.0.10",
-    "@supabase/supabase-js": "^2.39.0",
+    "@supabase/ssr": "latest",
```
jeremenichelli, line 13: Are these changes safe? @mandarini, what happens if there's a major release containing breaking changes for example?

mandarini: I could use `0.x` for any pre-1 version (`^0` is not valid), but I preferred to go with latest, since we don't have a `1.x` version yet. When we do have a `1.x` version, we can revisit this.

mandarini: and it's already used like this in some other examples: https://github.com/supabase/supabase/blob/40ce2a1cdb7cd3f108504e6a8e65d07d259c0f30/examples/auth/nextjs/package.json#L9

So this is keeping consistency!

djhi (approved):

jeremenichelli (approved):

jeremenichelli (approved):

PR #44728, opened 2026-04-09
refactor: save tax ID via customer update endpoint

ignaciodob:
## Summary

- Consolidate tax ID saving into the `PUT /organizations/{slug}/customer` endpoint instead of using a separate `/tax-ids` endpoint

### Updating billing information
From the billing dashboard`/org/_/billing`:
- [ ] Manually test updating billing address + tax ID from org billing settings - single API call should save both
- [ ] Clear the Tax ID in the org, save and assert that it has been deleted
- [ ] Modify all the fields in the address form, including TaxID. Click "Cancel" and assert that everything has returned to the previous values.

## Summary by CodeRabbit

* **Refactor**
  * Billing address and tax ID updates are now sent in a single update operation.
  * Error messaging for billing updates consolidated into a single failure notification.
  * Tax ID handling clarified: you can clear tax ID explicitly or leave it unchanged; omitting address fields no longer overwrites existing address.


alaister (approved):
LGTM!

- Tested on preview

Issue #42574, opened 2026-02-07
AI Assistant is incapable of doing literally anything


avaer:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Supabse AI Assistant is useless in local dev mode. It is incapable of understanding schemas or running code.

## To Reproduce

Start supabase local dev server.
Make sure you have openai api keys set.
Go to the studio.
Go to the AI assistant.
type: `give me the table definition for npc_threads`

## Expected behavior

The AI gives the table definition.

## Screenshots

<img width="980" height="666" alt="Image" src="https://github.com/user-attachments/assets/dc3085be-aaeb-485a-953d-028faa1f319e" />

The AI fails to give a table definition, and fails to read its own SQL query's results.

So basically 

## System information

- OS: macOS
- Browser: Chrome
- Version of supabase-js: 2.72.7
- Version of Node.js: v25.3.0

## Additional context

Discussed in Discord: https://discord.com/channels/839993398554656828/1461228638966911219
I was told to file a bug.

ritikpal1122:
@avaer can i pick this ?

avaer:
@ritikpal1122 WDYM?

ritikpal1122:
@avaer i want to solve this bug

avaer:
@ritikpal1122 sure, you don't need my permission :D

aantti:
Thanks for creating this issue - apologies for the delay, I might have missed it because the title wasn't too obvious.

aantti:
Closing this one for now (fixed in PR #44757?).

PR #44757, opened 2026-04-10
fix(self-hosted): local AI assistant unable to read schemas and interpret/execute SQL

rxxbyy:

YES

Bug fix for #42574 

If you setup a self-hosted Supabase environment and ask local AI assistant it is unable to interpret the results for executed SQL queries.

## What is the new behavior?

Local AI assistant can now read schemas and interpret the results for executed SQL queries.

## Additional context

I tested it using ollama with a smaller model which does not support reasoning mode so It would be useful if someone can test it using an actual OpenAI API KEY.

The main issue was that `executeSql` returns a wrapped `{ result: {...} }` on `fallback-tools.ts` making the model think there's no result. 
The solution is just to unwrap it like it's done at: `apps/studio/data/database-functions/database-functions-query.ts`: 

```
export async function getDatabaseFunctions(...) {
...
  const { result } = await executeSql(...)
...
  return result as DatabaseFunction[]
}
```

Also added a `false` default value in `execute_sql` tool schema so zod validation does not reject a valid LLM tool call if the model omits the field (which happened with the smaller model)

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal schema metadata query handling in AI tools.

* **Chores**
  * Enhanced SQL execution tool robustness by providing a sensible default for query type specification when not explicitly provided.

### DIFF
--- a/apps/studio/lib/ai/tools/fallback-tools.ts
+++ b/apps/studio/lib/ai/tools/fallback-tools.ts
@@ -37,7 +37,7 @@ export const getFallbackTools = ({
       }),
       execute: async ({ schemas }) => {
         try {
-          const result = includeSchemaMetadata
+          const { result } = includeSchemaMetadata
             ? await executeSql(
                 {
                   projectRef,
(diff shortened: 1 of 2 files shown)


aantti:
@singh-inder Could you possibly check? 🙏

aantti:
Related? https://github.com/supabase/supabase/pull/44730

charislam (approved):

PR #44806, opened 2026-04-13
chore: use ro connstring for observability

staaldraad:

YES

function change

defaults to the read-write connection string when doing observability report queries

## What is the new behavior?

uses the read-only connection string instead

## Additional context

these should only ever be read-only operations, reporting should not have side effects and this adds a guardrail to ensure that remains the case

## Summary by CodeRabbit

**Bug Fixes**
- Corrected database replica query handling by using read-only connection strings for replica database access.


jordienr (approved):
change looks good, approving to unblock.<br><br>@staaldraad can you resolve the coderabbit comment?

staaldraad:
changes applied and tested in staging

PR #44810, opened 2026-04-13
fix: reset stale infinite query

7ttp:
fixes `UseInfiniteQuery` so it recreates when query inputs change

the previous prop change check compared values to themselves so the hook could keep stale paginated state 
after table columns page size or trailing query changed

- closes https://github.com/supabase/supabase/issues/37750

## Summary by CodeRabbit

* **New Features**
  * Added an optional trailingQueryKey prop to the infinite-query hook to better control when its internal store is refreshed.

* **Bug Fixes**
  * Pagination now reliably uses the latest trailing query and avoids inconsistent state during initial load and subsequent pages.

* **Refactor**
  * Reworked store lifecycle and initialization for the infinite-query hook to reduce unnecessary recreations, streamline startup, and improve performance and stability.


apps/ui-library/registry/default/blocks/infinite-query-hook/hooks/use-infinite-query.ts
```diff
@@ -176,29 +176,26 @@ function useInfiniteQuery<
   TData extends SupabaseTableData<T>,
   T extends SupabaseTableName = SupabaseTableName,
 >(props: UseInfiniteQueryProps<T>) {
-  const storeRef = useRef(createStore<TData, T>(props))
+  const tableName = props.tableName
+  const columns = props.columns ?? '*'
+  const pageSize = props.pageSize ?? 20
+  const trailingQuery = props.trailingQuery
+  const store = useMemo(
+    () => createStore<TData, T>({ tableName, columns, pageSize, trailingQuery }),
```
ivasilov: The `trailingQuery` may not be stable (if it's defined as a inline callback) which will trigger `useMemo` on each render.

7ttp: > The `trailingQuery` may not be stable (if it's defined as a inline callback) which will trigger `useMemo` on each render.

ahh, yh that makes sense! :)
I’ve pushed a change to keep `trailingQuery` out of `useMemo` deps
 (so inline callbacks don’t recreate on every render) and 

added `trailingQueryKey` so query shape changes still recreate/reset the store when needed

ivasilov (approved):

PR #44831, opened 2026-04-14
fix: reset identity and serial sequences after CSV import

BhavarthPathare:
YES

Bug fix

When importing data via CSV into a table with identity or serial primary key columns, PostgreSQL sequences are not updated to reflect the imported values. This causes unique key violations on subsequent inserts because the sequence still starts from its previous value, conflicting with the imported IDs.

Fixes #43993

## What is the new behavior?
After a successful CSV import, sequences are reset for both column types:
- **IDENTITY columns** (`GENERATED ALWAYS/BY DEFAULT AS IDENTITY`) via the existing `getUpdateIdentitySequenceSQL`
- **SERIAL columns** (`smallserial`, `serial`, `bigserial`) via the new `getUpdateSerialSequenceSQL`, detected by `nextval(...)` default value

The new `getUpdateSerialSequenceSQL` uses the 3-argument form of `setval(seq, val, is_called)` to correctly handle empty tables — setting `is_called=false` when the table has no rows, so the next insert gets ID 1 rather than 2.

## Additional context
Two existing PRs address this issue but each only covers one column type:
- #44037 handles IDENTITY columns only
- #44304 handles SERIAL columns only, but the function was never wired into the import flow (no call site), and used the 2-arg `setval` which skips ID 1 on empty tables

This PR covers both column types in a single pass and includes unit tests for `getUpdateSerialSequenceSQL`.

**Files changed:**
- `packages/pg-meta/src/sql/studio/table-editor/identity.ts` — adds `getUpdateSerialSequenceSQL`
- `apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx` — wires sequence reset into `onImportData` for both column types
- `apps/studio/tests/components/Editor/SpreadsheetImport.utils.test.ts` — adds unit tests

## Release Notes

* **New Features**
  * Auto-incrementing columns now have their sequences automatically updated after CSV/spreadsheet imports.

* **Bug Fixes**
  * Sequence reset failures no longer block data imports; a warning notification is shown instead.

* **Tests**
  * Added test coverage for sequence reset functionality.

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.tsx
@@ -1,4 +1,6 @@
+// AFTER — add these two imports:
 import * as Sentry from '@sentry/nextjs'
+import { getUpdateIdentitySequenceSQL, getUpdateSerialSequenceSQL } from '@supabase/pg-meta'
 import type { PostgresColumn, PostgresTable } from '@supabase/postgres-meta'
 import { useQueryClient } from '@tanstack/react-query'
 import { useParams } from 'common'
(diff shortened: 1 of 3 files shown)


7ttp (commented):
closing in favour of: https://github.com/supabase/supabase/pull/45076

PR #44843, opened 2026-04-14
ref(etl): Improve UX around table resets

iambriccardo:
This PR enables users to restart tables when the pipeline is stopped or failed and in addition it improves the reliability of the state.

### Screenshots

image

## Summary by CodeRabbit

* **New Features & Improvements**
  * Centralized, typed pipeline display model with clearer labels, tones and loading states.
  * Rollback now accepts an optional pipeline-status hint and will conditionally start/restart the pipeline after rollback; follow-up actions reliably run and show contextual toasts.
  * Action buttons and batch controls reflect refined busy/actionable logic and remain visible but disabled when appropriate.

* **Bug Fixes**
  * Restart dialog copy correctly handles singular/plural and omits counts when zero.

* **UX**
  * Improved empty-state text, tooltips, and disabled-state messaging for clarity.


apps/studio/data/replication/rollback-tables-mutation.ts
```diff
@@ -65,32 +68,72 @@ export const useRollbackTablesMutation = ({
   'mutationFn'
 > = {}) => {
   const queryClient = useQueryClient()
-  const { restartPipeline } = useRestartPipelineHelper()
+  const { restartPipeline } = useRestartPipelineHelper({ suppressChildErrorToasts: true })
+  const { mutateAsync: startPipeline } = useStartPipelineMutation({ onError: () => {} })
```
joshenlim: we shouldn't be calling another mutation within a mutation
each mutation should only be doing one thing (separation of concerns)

in this case i'd opt to shift this logic into `rollbackTables` above instead of within this hook
[edit] one sec just thinking twice

joshenlim: this is getting quite complex here

- double invalidation of RQ cache - in `startPipeline` we're invalidating the pipelineStatus key, but in here we're also then invalidating the same key in the finally block
- the optimistic update within `startPipeline` will subsequently get overridden too

apps/studio/data/replication/rollback-tables-mutation.ts
```diff
@@ -65,32 +68,72 @@ export const useRollbackTablesMutation = ({
...
-        }),
-        queryClient.invalidateQueries({
-          queryKey: replicationKeys.pipelinesReplicationStatus(projectRef, pipelineId),
-        }),
-      ])
-
-      await onSuccess?.(data, variables, context)
+      if (resumeError === null) {
+        await onSuccess?.(data, variables, context)
+      }
```
joshenlim: its odd that we're calling a mutation within a mutation - rollback could've technically succeeeded but if the pipeline restart logic didn't, the onSuccess is never called then

apps/studio/data/replication/rollback-tables-mutation.ts
```diff
@@ -65,32 +68,72 @@ export const useRollbackTablesMutation = ({
...
+          pipelineStatusName === PipelineStatusName.STARTED ||
+          pipelineStatusName === PipelineStatusName.FAILED
+
+        const postRollbackAction = canStartPipeline
+          ? 'start'
+          : canRestartPipeline
+            ? 'restart'
+            : 'unknown'
+
+        action = postRollbackAction === 'unknown' ? 'resume' : postRollbackAction
```
joshenlim: this feels confusing - there's 2 actions here?

joshenlim: imo rewritting as this is clearer from a readability POV

```
if (pipelineStatusName === PipelineStatusName.STOPPED) {
    await startPipeline({ projectRef, pipelineId })
  } else if (
    pipelineStatusName === PipelineStatusName.STARTED ||
    pipelineStatusName === PipelineStatusName.FAILED
  ) {
    await stopPipeline({ projectRef, pipelineId })
    await startPipeline({ projectRef, pipelineId })
  } else {
    throw new Error(
      `Cannot apply rollback while pipeline status is ${
        pipelineStatusName || 'unknown'
      }. Retry once the pipeline status is started, failed, or stopped.`
    )
  }
```

apps/studio/data/replication/rollback-tables-mutation.ts
joshenlim: it seems like we only want the startPipeline action and not the onSuccess events from it (i.e let this rollback logic here determine the onSuccess events)

apps/studio/components/interfaces/Database/Replication/Pipeline.utils.ts
```diff
@@ -19,100 +16,143 @@ export const PIPELINE_ERROR_MESSAGES = {
...
+  | 'stopping'
+  | 'restarting'
+  | 'failed'
+  | 'stopped'
+  | 'running'
+  | 'unknown'
 
-const PIPELINE_STATE_MESSAGES = {
-  enabling: {
+export type PipelineDisplayTone = 'failure' | 'loading' | 'success' | 'idle'
```
joshenlim: nit but why not just call it `type` since we're doing `const { tone: type, message, label } = displayState` in PipelineStatus.tsx

apps/studio/components/interfaces/Database/Replication/Pipeline.utils.ts
```diff
@@ -19,100 +16,143 @@ export const PIPELINE_ERROR_MESSAGES = {
...
+export type PipelineDisplayAccent = 'brand' | 'warning' | 'danger' | 'neutral' | 'success'
+
+export interface PipelineDisplayState {
+  key: PipelineDisplayStateKey
+  label: string
+  title: string
+  message: string
+  badge: string
+  tone: PipelineDisplayTone
+  accent: PipelineDisplayAccent
```
joshenlim: this property doesn't seem like its being consumed anywhere? can remove

apps/studio/components/interfaces/Database/Replication/BatchRestartDialog.tsx
```diff
@@ -91,8 +94,9 @@ export const BatchRestartDialog = ({
           description: (
             <div className="space-y-3 text-sm">
               <p>
-                This will restart replication for <strong>all {totalTables} tables</strong> in this
-                pipeline from scratch:
+                This will restart replication for{' '}
+                <strong>{totalTables > 0 ? `all ${totalTables} tables` : 'all tables'}</strong> in
```
joshenlim: this ternary feels like an overkill - could just show the number by default

apps/studio/components/interfaces/Database/Replication/ReplicationPipelineStatus/ReplicationPipelineStatus.tsx
```diff
@@ -389,46 +413,63 @@ export const ReplicationPipelineStatus = () => {
...
+                      icon={<ChevronDown />}
+                      className="w-7 rounded-l-none -ml-[1px]"
+                      disabled={showDisabledState || isPipelineError}
+                    />
+                  </DropdownMenuTrigger>
+                  <DropdownMenuContent align="end" className="w-44">
+                    <Tooltip>
+                      <TooltipTrigger asChild>
+                        <span>
+                          <DropdownMenuItem
```
joshenlim: we have a DropdownMenuItemTooltip component to handle this

joshenlim (approved):
Verified changes on staging preview - minimally stopped the pipeline then hit reset tables

PR #44853, opened 2026-04-14
fix(docs): update Auth0 third-party auth guide for custom claims

mandarini:
Update TypeScript example to use createAuth0Client factory function and ID tokens as the primary approach. Add warning about Auth0 silently stripping non-namespaced custom claims from access tokens.                                                   

Ref: supabase/supabase-js#1770     

## Summary by CodeRabbit

* **Documentation**
  * Updated Auth0 integration to a newer client initialization pattern.
  * Switched examples to use the ID token instead of access tokens for Supabase and added explicit failure behavior when the ID token is missing.
  * Updated Auth0 Action guidance to set the claim on the ID token.
  * Added a caution that Supabase requires the literal "role" claim and that non-namespaced custom claims are stripped from access tokens.


apps/docs/content/guides/auth/third-party/auth0.mdx
```diff
@@ -145,14 +143,30 @@ Your Supabase project inspects the `role` claim present in all JWTs sent to it,
 
 By default, Auth0 JWTs (both access token and ID token) do not contain a `role` claim in them. If you were to send such a JWT to your Supabase project, the `anon` role would be assigned when executing the Postgres query. Most of your app's logic will be accessible by the `authenticated` role.
 
-A recommended approach to do this is to configure the [`onExecutePostLogin` Auth0 Action](https://auth0.com/docs/secure/tokens/json-web-tokens/create-custom-claims#create-custom-claims) which will add the custom claim:
+A recommended approach to do this is to configure the [`onExecutePostLogin` Auth0 Action](https://auth0.com/docs/secure/tokens/json-web-tokens/create-custom-claims#create-custom-claims) which will add the custom claim.
+
+If you are using **ID tokens** (recommended for the TypeScript SPA example above):
+
```
jeremenichelli: It feels redundant a bit to always use the "recommended" signal of clarification. I would be more direct.

```suggestion
If you are using **ID tokens** from the example above:

```

apps/docs/content/guides/auth/third-party/auth0.mdx
```diff
@@ -145,14 +143,30 @@ Your Supabase project inspects the `role` claim present in all JWTs sent to it,
...
+```javascript
+exports.onExecutePostLogin = async (event, api) => {
+  api.idToken.setCustomClaim('role', 'authenticated')
+}
+```
+
+If you are using **access tokens** instead, you must use a namespaced claim because [Auth0 strips non-namespaced custom claims from access tokens](https://auth0.com/docs/troubleshoot/product-lifecycle/past-migrations/custom-claims-migration):
```
jeremenichelli:
```suggestion
For **access tokens**, you must use a namespaced claim because [Auth0 strips non-namespaced custom claims](https://auth0.com/docs/troubleshoot/product-lifecycle/past-migrations/custom-claims-migration):
```

apps/docs/content/guides/auth/third-party/auth0.mdx
```diff
@@ -145,14 +143,30 @@ Your Supabase project inspects the `role` claim present in all JWTs sent to it,
...
 ```javascript
 exports.onExecutePostLogin = async (event, api) => {
-  api.accessToken.setCustomClaim('role', 'authenticated')
+  api.accessToken.setCustomClaim('https://your-app.com/role', 'authenticated')
 }
 ```
 
+<Admonition type="caution">
+
+Using `api.accessToken.setCustomClaim('role', 'authenticated')` without a namespace may appear to work but Auth0 can silently strip the claim. Use ID tokens or namespaced access token claims to avoid this issue.
```
jeremenichelli:
```suggestion
Auth0 can silently strip claims without a namespace when using `api.accessToken.setCustomClaim`. We recommend ID tokens or namespaced access token claims to avoid this issue.
```

jeremenichelli (approved):
Minor text edits, but all good 👍

apps/docs/content/guides/auth/third-party/auth0.mdx
```diff
@@ -36,12 +36,10 @@ const supabase = createClient(
...
-
-      // Alternatively you can use (await auth0.getIdTokenClaims()).__raw to
-      // use an ID token instead.
-
-      return accessToken
+      // Use the ID token which reliably includes custom claims.
+      const idToken = (await auth0.getIdTokenClaims())?.__raw
+      if (!idToken) throw new Error('Missing ID token')
+      return idToken
     },
```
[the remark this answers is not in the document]
mandarini: Verified: idToken is non-nullable (String) in both auth0_flutter and Auth0.Android. The null guard is unnecessary, adding one would be dead code. The TypeScript example needs it because getIdTokenClaims() returns IdToken | undefined, but the mobile SDKs guarantee a non-null idToken on their Credentials objects.

PR #44870, opened 2026-04-14
fix(studio): show empty org warning when there are no active projects

### LINKED ISSUE #44872: Empty organization warning shows for orgs with 5 active projects instead of 0
In the subscription upgrade dialog, the "Empty organization" admonition checks \`.length === 5\` instead of \`.length === 0\`:

\`\`\`tsx
{projects.filter(
  (it) =>
    it.status === PROJECT_STATUS.ACTIVE_HEALTHY ||
    it.status === PROJECT_STATUS.COMING_UP
).length === 5 &&
\`\`\`

The warning text says "This organization has no active projects" but it only renders for orgs with exactly 5 active projects. Introduced in #44494.

oniani1:
Closes #44872

The "Empty organization" admonition in the subscription upgrade dialog checks `.length === 5` instead of `.length === 0`. The warning text says "This organization has no active projects" but only shows for orgs with exactly 5 active projects.

Introduced in #44494 (b9e83b25).

The adjacent `.length === 1` check for the single-project note is correct, confirming this is a count comparison that should be `0`.


djhi (approved):
Thanks!

PR #44871, opened 2026-04-14
fix(studio): stop cron job creation when name validation fails

### LINKED ISSUE #44873: Cron job created even when name validation fails
When creating a new cron job, \`getDatabaseCronJob\` is called to check if the name already exists. If that call throws (network error, DB connection issue, etc.), the catch block shows an error toast but doesn't \`return\`:

\`\`\`tsx
} catch (error: any) {
  toast.error(\`Failed to validate cron job name: \${error.message}\`)
  // no return here
} finally {
  setIsLoadingGetCronJob(false)
}

// execution falls through, job gets created without validation
const command = \`$$\${values.snippet}$$\`
const query = buildCronQuery(name, schedule, command)
upsertCronJob(...)
\`\`\`

oniani1:
Closes #44873

When creating a new cron job, `getDatabaseCronJob` is called to check if the name already exists. If that call throws (network error, DB connection issue), the catch block shows an error toast but doesn't return. Execution falls through to `upsertCronJob`, creating the job without validating name uniqueness.

The fix adds `return` in the catch block so the mutation doesn't fire after a failed validation check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cron job validation error handling to properly halt processing when name validation fails and display appropriate error feedback.


djhi (approved):
Thanks!

PR #44880, opened 2026-04-14
feat: add api-catalog to .well-known

gregnr:
Adds a `/.well-known/api-catalog` to advertise our management API. Motivation is for agents, but this is generally part of [RFC 9727](https://www.rfc-editor.org/rfc/rfc9727).

## Testing

Make an HTTP request to `/.well-known/api-catalog` and confirm it returns the JSON content with content type `application/linkset+json`:

```shell
curl -i https://zone-www-dot-com-git-feat-agent-catalog-supabase.vercel.app/.well-known/api-catalog
```

## Summary by CodeRabbit

* **New Features**
  * Added a public API catalog (linkset) exposing API metadata, OpenAPI endpoint, documentation and status links.
  * Added HTTP headers to advertise the API catalog from the site root and to serve the catalog with the correct content type.

### DIFF
--- a/apps/www/next.config.mjs
+++ b/apps/www/next.config.mjs
@@ -89,7 +89,10 @@ const nextConfig = {
       },
       {
         source: '/.well-known/api-catalog',
-        headers: [{ key: 'content-type', value: 'application/linkset+json' }],
+        headers: [
+          { key: 'content-type', value: 'application/linkset+json' },
+          { key: 'access-control-allow-origin', value: '*' },
+        ],
       },
       {
         source: '/.well-known/vercel/flags',
--- a/apps/www/next.config.mjs
+++ b/apps/www/next.config.mjs
@@ -79,6 +79,19 @@ const nextConfig = {
   async headers() {
     return [
       {
+        source: '/',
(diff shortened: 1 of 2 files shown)


pamelachia:
Nice, RFC 9727 is the right standard for this.

One suggestion: consider adding `Access-Control-Allow-Origin: *` to the `/.well-known/api-catalog` header rule in `next.config.mjs`. The catalog is a static advertisement pointing to already-public URLs, so there's no security concern. Without it, browser-based agents or tools making cross-origin fetches will be blocked by CORS.

Separately, `https://api.supabase.com/api/v1-json` (the OpenAPI spec this catalog points to) is already publicly accessible with no auth, but also has no CORS headers. That's outside this PR's scope but worth a follow-up if the goal is end-to-end browser-based agent discovery.

pamelachia (approved):
approving to unblock, you can make the changes then merge

gregnr:
@pamelachia Added wildcard CORS header. Out of curiosity, which browser agents are you aware of that would need this? Regardless I agree it doesn't hurt to have. Will merge now!

PR #44887, opened 2026-04-15
fix(auth): remove banner from email notification templates

fadymak:
Removes the "New" banner from the email notification templates section as the features has been GA-ed for ~6 months now.

CleanShot 2026-04-15 at 10 30 33@2x

## Summary by CodeRabbit

* **Refactor**
  * Removed the security notifications acknowledgement feature from the email templates interface, including the dismissible notification tip and associated state management.


dnywh (approved):
Tested on staging. Thanks!

PR #44670, opened 2026-04-08
feat(docs,www): make supabase.com machine-readable for LLM agents

pamelachia:
## Summary

Makes supabase.com machine-readable for LLM agents and AI crawlers. Adds a combined `llms-full.txt` (product overview + docs), curated product overview `.txt` files, auto-generated pricing content, and updates the `llms.txt` index to reference everything.

**Why:** supabase.com had `llms.txt` pointing to per-SDK doc files, but no `llms-full.txt` (every competitor has one), no machine-readable marketing content, and no pricing data agents could parse. Agents evaluating Supabase got docs but no product overview or pricing, leading to inaccurate comparisons.

### `llms-full.txt` (auto-generated, combines www + docs)
- Product Overview section (~500 lines): homepage, all 6 products, pricing
- Documentation section (~122K lines): guides, SDK references, CLI reference
- 4.4MB total, regenerated on every docs deploy
- Reads product `.txt` files from `apps/www/public/llms/` at build time

### Product overview `.txt` files (`apps/www/public/llms/`)
- Curated summaries: homepage, database, auth, storage, edge functions, realtime, vector
- `pricing.txt` is auto-generated from `packages/shared-data` (plans, pricing, compute add-ons) via `generateLlmsPricing.mjs`, includes full feature comparison matrix
- Other product files are hand-maintained (these pages change ~1x/year per git history)
- Reminder comments added to all 7 marketing page source files

### `llms.txt` index (auto-generated)
- Two sections: Documentation (existing SDK/guide links) and Product Overview (marketing page links)
- Links to `llms-full.txt` for bulk ingestion

### Rewrite changes
- Added `/llms-full.txt` rewrite to docs app
- Scoped `/llms/*.txt` wildcard to only match docs source slugs (guides, js, dart, etc.), so marketing `.txt` files in `www/public/llms/` are served directly

## Changes

- `apps/docs/scripts/llms.ts`: generate `llms-full.txt` combining www product content + docs, update `llms.txt` index with marketing links
- `apps/www/lib/rewrites.js`: add `llms-full.txt` rewrite, scope docs proxy to known slugs
- `apps/www/public/llms/*.txt`: 8 product overview files (7 hand-curated, 1 auto-generated)
- `apps/www/scripts/generateLlmsPricing.mjs`: build script generating pricing.txt from shared-data; uses `getPlanValue()` guard in `buildAddOnsSection` and `buildFeatureComparisonSection` to handle missing plan keys defensively
[...]

PR #44897, opened 2026-04-15
[GROWTH-773] chore(www): serve llms.txt content via API routes

alaister:
The docs build had a fragile implicit dependency on www's filesystem (`../../../apps/www/public/llms`), flagged by the docs team in #44670. Rather than formalising that dependency with a shared package, this PR eliminates it entirely by making www the sole owner of llms content assembly.

**How it works now:**

`/llms/[slug]` handles all `/llms/*.txt` requests via a 3-step cascade:
1. Dynamic content — `pricing.txt` generated at request time from `shared-data` imports
2. Local file — product overviews read from `data/llms/`
3. Docs proxy — reference docs (guides, js, dart, etc.) fetched from the docs app

No hardcoded slug lists, so adding new content just works.

**What changed:**

- `apps/docs/scripts/llms.ts` trimmed to only generate per-source reference files — www now owns `llms.txt`, `llms-full.txt`, and product overviews
- Removed `generateLlmsPricing.mjs` build script — pricing generated dynamically from `shared-data`
- Removed llms rewrites from `rewrites.js` — routes handle everything with consistent `Cache-Control: public, s-maxage=3600, stale-while-revalidate=86400`
- Product overview `.txt` files moved from `public/llms/` → `data/llms/` so all requests go through routes for consistent caching

**Docs team concerns from GROWTH-773:**

| Concern | Resolution |
|---------|-----------|
| Docs build depends on www files at a fragile relative path | Path removed — docs no longer reads from www |
| www restructuring breaks docs with no obvious connection | Eliminated — no cross-app filesystem dependency |
| No build order enforcement between www and docs | Not needed — docs doesn't depend on www's build output |

## To test

- `curl <preview>/llms.txt` — markdown index with doc + product overview links
- `curl <preview>/llms-full.txt` — combined product overviews + docs content
- `curl <preview>/llms/pricing.txt` — dynamically generated pricing tables
- `curl <preview>/llms/auth.txt` — product overview from local file
- `curl <preview>/llms/guides.txt` — proxied from docs app
- `curl <preview>/llms/nonexistent.txt` — 404
- Verify `Cache-Control` header on all responses


apps/www/app/llms/[slug]/route.ts
```diff
@@ -0,0 +1,49 @@
+import { readFile } from 'node:fs/promises'
+import { join } from 'node:path'
+
+import { generatePricingContent } from '@/lib/llms'
+
+export const dynamic = 'force-dynamic'
```
Comment (jeremenichelli) [inline apps/www/app/llms/[slug]/route.ts:6]: It might be worth using `getStaticPaths` if we already know the files we want or have a way to infer them through data.

It will save edge time requests, sadly next it's quite slow when the cache is not "warm". Not a deal breaker, but nice if possible optimization.

jeremenichelli (approved):

pamelachia (approved):

apps/www/app/llms/[slug]/route.ts
Comment (alaister) [reply apps/www/app/llms/[slug]/route.ts:6]: I think the edge requests will be pretty minimal right now anyway thanks to the cache headers. Vercel's CDN is pretty great with stale-while-revalidate.

Do scrapers prioritise speed on these paths? I think cold requests should be pretty minimal due to the large number of requests we get, but happy to add it in if cold request speed is a priority.

PR #44898, opened 2026-04-15
chore: extend format/lint hook to all apps and packages

alaister:
The Claude Code post-tool-use format/lint hook was hardcoded to only run on `apps/studio/` files. This updates it to work across the whole repo.

**Changed:**
- Prettier now runs on any supported file type in the repo (not just Studio)
- ESLint runs for any workspace that has it configured, using a simple lookup table
- Unsupported file types (e.g. `.sh`) are skipped to avoid "no parser" errors

## To test

- Edit a file in `apps/studio/` with bad formatting — should get auto-formatted and linted
- Edit a file in `packages/common/` — should get auto-formatted (no ESLint since it's not configured there)
- Edit a `.sh` file — should be skipped with no errors

### DIFF
--- a/.claude/scripts/format_and_lint.sh
+++ b/.claude/scripts/format_and_lint.sh
@@ -1,9 +1,20 @@
 #!/bin/bash
 #
-# PostToolUse hook: format and lint apps/studio/ files after Write or Edit.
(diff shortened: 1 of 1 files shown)


awaseem (approved):
LGTM

PR #44905, opened 2026-04-15
fix(FE-2782): re-enable log drain updates via PATCH, filter REDACTED headers

jordienr:
## Summary

Closes [FE-2782](https://linear.app/supabase/issue/FE-2782/log-drain-save-replaces-auth-token-with-redacted)

- **Switch PUT → PATCH** in `update-log-drain-mutation.ts` for partial config updates.
- **Re-enable the Update option** in the log drain row menu (`LogDrains.tsx`), which was commented out as a workaround while the PATCH endpoint was being built.
- **Strip REDACTED header sentinels** before submitting in update mode. The API returns `"REDACTED"` as a placeholder for secret header values. `computePatchHeaders` filters these out so the real secret is preserved server-side. Everything else (non-REDACTED headers, all other config fields) is sent as-is.
- **Add unit tests** for `computePatchHeaders` (9 cases) in `LogDrains.utils.test.ts` and 4 form-level update-mode tests in `LogDrainDestinationSheetForm.test.tsx`.

### Known backend limitation

The platform's PATCH endpoint currently forwards the request body directly to Logflare as a PUT without merging. The platform team is aware and will fix this. The frontend already sends the correct partial payload (REDACTED headers excluded), so once the backend is fixed everything will work end-to-end.

## Test plan

- [ ] Create a log drain with `Authorization: MY_TOKEN` and a second header. Verify both arrive at the destination.
- [ ] Open Update — `Authorization` shows as `REDACTED`, second header shows its real value.
- [ ] Add a new header and save. Verify requests still use the original token (not `REDACTED`).
- [ ] Change `Authorization` to a new real value and save. Verify requests use the new token.
- [ ] Update only the drain name. Verify it saves correctly.
- [ ] Delete all headers and save. Verify headers are cleared.
- [ ] Verify Delete still works.


jordienr:
Fixed in 43bf6c0cb6. Resolved the thread.

apps/studio/components/interfaces/LogDrains/LogDrainDestinationSheetForm.tsx
```diff
@@ -238,21 +239,27 @@ type LogDrainDestinationSubmitValues = z.infer<typeof submitSchema>
...
-        ? { ...rest, headers }
-        : rest
+  const rawHeaders = headerRowsToRecord(headerEntries)
+  const headers = mode === 'update' ? computePatchHeaders(rawHeaders) : rawHeaders
+
+  // In update mode: include headers whenever computePatchHeaders returned a value —
+  // even {} signals "explicitly cleared". In create mode: only include when non-empty.
+  const shouldIncludeHeaders =
+    rest.type === 'loki' ||
+    (mode === 'update' ? headers !== undefined : !!headers && Object.keys(headers).length > 0)
```
jordienr: this should be split into different constants for readability.

apps/studio/components/interfaces/LogDrains/LogDrainDestinationSheetForm.tsx
```diff
@@ -238,21 +239,29 @@ type LogDrainDestinationSubmitValues = z.infer<typeof submitSchema>
...
+  const rawHeaders = headerRowsToRecord(headerEntries)
+  const headers = mode === 'update' ? computePatchHeaders(rawHeaders) : rawHeaders
+
+  // In update mode, computePatchHeaders returns {} to signal "clear all headers" —
+  // so any non-undefined result should be sent. In create mode, skip empty headers.
+  const isLoki = rest.type === 'loki'
+  const hasHeadersInCreateMode = !!headers && Object.keys(headers).length > 0
+  const hasHeadersInUpdateMode = headers !== undefined
+  const shouldIncludeHeaders =
+    isLoki || (mode === 'update' ? hasHeadersInUpdateMode : hasHeadersInCreateMode)
```
jordienr, line 262: hate this part of the code but oh well

djhi (approved):
Unblocking. Unfamiliar with log drains though

jordienr:
converting to draft because i need to revisit this

PR #44909, opened 2026-04-15
fix(studio): use nuqs for database functions search to prevent caret jumping

awaseem:
## Summary
- Fixes the cursor jumping to end of input on every keystroke in the database functions search
- Replaced manual `useParams` + `router.push` URL sync with `useQueryState` from nuqs, which updates state synchronously — matching the pattern already used for the other filters on this page
- Removed the `setTimeout` hack in the SchemaSelector onChange that was only needed because of the async URL update

Fixes FE-3000

## Test plan
- [x] Navigate to Database > Functions
- [x] Type a search string, then click in the middle of the string and type — cursor should stay in place
- [x] Verify search filtering still works correctly
- [x] Verify switching schemas still clears the search input
- [x] Verify the `?search=` query param is still reflected in the URL

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsiveness of schema selection in the Functions list—search filters now clear immediately when switching schemas, eliminating previous delays.

* **Improvements**
  * Enhanced URL state synchronization for better consistency with browser history and navigation.

### DIFF
--- a/apps/studio/components/interfaces/Database/Functions/FunctionsList/FunctionsList.tsx
+++ b/apps/studio/components/interfaces/Database/Functions/FunctionsList/FunctionsList.tsx
@@ -1,7 +1,5 @@
 import { PermissionAction } from '@supabase/shared-types/out/constants'
-import { useParams } from 'common'
 import { Search } from 'lucide-react'
-import { useRouter } from 'next/router'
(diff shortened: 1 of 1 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #44911, opened 2026-04-15
feat(studio): add Copy as CSV option to SQL editor results

awaseem:
## Summary
- Adds a new "Copy as CSV" action to the export dropdown in the SQL editor results panel, alongside the existing Copy as Markdown, Copy as JSON, and Download CSV options
- Registers a `Shift+Cmd+C` keyboard shortcut for the action, with a toggle in Account > Preferences > Keyboard shortcuts
- Works everywhere `DownloadResultsButton` is used: SQL editor, logs, query performance, and linter pages

Fixes FE-2991

## Test plan
- [x] Run a SELECT query in the SQL editor, open the Export dropdown, and verify "Copy as CSV" appears between "Copy as JSON" and "Download CSV"
- [x] Click "Copy as CSV" and verify CSV data is copied to clipboard
- [x] Use `Shift+Cmd+C` shortcut and verify it copies CSV to clipboard
- [x] Go to Account > Preferences > Keyboard shortcuts and verify the "Copy results as CSV" toggle is present and functional
- [x] Disable the shortcut in preferences, verify `Shift+Cmd+C` no longer triggers the copy
- [x] Verify the same option appears in the logs export dropdown

## Summary by CodeRabbit

* **New Features**
  * Added "Copy as CSV" action to the results download menu, allowing users to quickly copy query results in CSV format to the clipboard.
  * Introduced a new hotkey preference setting to enable/disable the Shift+Meta+C keyboard shortcut for copying results as CSV.

### DIFF
--- a/apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityPanel.tsx
+++ b/apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityPanel.tsx
@@ -172,6 +172,12 @@ const UtilityPanel = ({
                   groups: { project: ref ?? '', organization: org?.slug ?? '' },
                 })
               }}
+              onCopyAsCSV={() => {
+                sendEvent({
+                  action: 'sql_editor_result_copy_csv_clicked',
+                  groups: { project: ref ?? '', organization: org?.slug ?? '' },
+                })
+              }}
             />
           )}
         </div>
--- a/apps/studio/components/interfaces/Account/Preferences/HotkeySettings.tsx
+++ b/apps/studio/components/interfaces/Account/Preferences/HotkeySettings.tsx
@@ -22,6 +22,7 @@ const HotkeySchema = z.object({
   inlineEditorEnabled: z.boolean(),
   copyMarkdownEnabled: z.boolean(),
   copyJsonEnabled: z.boolean(),
+  copyCsvEnabled: z.boolean(),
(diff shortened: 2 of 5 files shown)


apps/studio/components/interfaces/SQLEditor/UtilityPanel/UtilityPanel.tsx
```diff
@@ -172,6 +172,12 @@ const UtilityPanel = ({
                   groups: { project: ref ?? '', organization: org?.slug ?? '' },
                 })
               }}
+              onCopyAsCSV={() => {
+                sendEvent({
+                  action: 'sql_editor_result_copy_csv_clicked',
+                  groups: { project: ref ?? '', organization: org?.slug ?? '' },
```
seanoliver, line 178: nit: I see that we're using the existing analytics pattern in this file, but `useSendEventMutation` is deprecated — so ideally new events should use `useTrack` from `lib/telemetry/track` instead. This makes sure we're adding the group references and other properties in a standard way for analytics downstream.

Not blocking of course!

seanoliver (approved):
LGTM from a growth eng POV! Minor nit about our new `useTrack` hook but optional since it's aligned with the other tracking events in this file.

dnywh (approved):
Tested on staging. We can finesse exact shortcut as a follow-up.

PR #44919, opened 2026-04-15
fix(docker): decouple STORAGE_PUBLIC_URL for S3 protocol SigV4 verification

### LINKED ISSUE #44769: S3 protocol returns SignatureDoesNotMatch with valid S3_PROTOCOL_ACCESS_KEY_ID/SECRET on self-hosted official docker stack
## Summary

I'm running the official self-hosted Supabase Docker stack and trying to use the built-in S3 protocol endpoint:

`/storage/v1/s3`

The endpoint is reachable, the bucket exists, and invalid credentials correctly return `InvalidAccessKey`, but when using the exact `S3_PROTOCOL_ACCESS_KEY_ID` and `S3_PROTOCOL_ACCESS_KEY_SECRET` loaded in the `storage` container, all tested clients return:

`SignatureDoesNotMatch`

So it looks like credentials are recognized, but valid AWS SigV4 requests are still rejected.

---

## Environment

- Supabase storage image: `supabase/storage-api:v1.48.26`
- Self-hosted official Docker setup
- Single tenant
- Backend storage mode: `file`
- Host OS: macOS (Docker Desktop)
- Local test URL: `http://127.0.0.1:18000/storage/v1/s3`

---

## Active storage container env

From `docker inspect` / `docker exec env`:

```env
STORAGE_BACKEND=file
FILE_STORAGE_BACKEND_PATH=/var/lib/storage
TENANT_ID=storage-single-tenant
REGION=eu-west-3
S3_PROTOCOL_ACCESS_KEY_ID=625729a08b95bf1b7ff351a663f3a23c
S3_PROTOCOL_ACCESS_KEY_SECRET=850181e4652dd023b7a98c58ae0d2d34bd487ee0cc3254aed6eda37307425907
REQUEST_ALLOW_X_FORWARDED_PATH=true
STORAGE_PUBLIC_URL=XXXXXXXXXXX
The bucket exists in the database:

• bucket name: game-images

───

What works

1. Bucket listing through the normal Storage API works

Using service role auth against /storage/v1/bucket, bucket listing succeeds.

2. Missing signature is handled correctly

A plain curl request to the S3 endpoint returns Missing signature, which suggests the S3 endpoint is active.

3. Wrong credentials are rejected correctly

Using wrong credentials returns:

Invalid Access Key

This makes me think the configured S3_PROTOCOL_ACCESS_KEY_ID is being checked correctly.

───

What does not work

When I use the exact values loaded in the container:

• REGION=eu-west-3
• S3_PROTOCOL_ACCESS_KEY_ID=625729a08b95bf1b7ff351a663f3a23c
• S3_PROTOCOL_ACCESS_KEY_SECRET=850181e4652dd023b7a98c58ae0d2d34bd487ee0cc3254aed6eda37307425907

I consistently get:

SignatureDoesNotMatch

───

rclone reproduction

Command
rclone ls \
  --s3-provider Other \
  --s3-endpoint http://127.0.0.1:18000/storage/v1/s3 \
  --s3-region eu-west-3 \
  --s3-force-path-style \
  --s3-list-version 1 \
  --s3-access-key-id "625729a08b95bf1b7ff351a663f3a23c" \
  --s3-secret-access-key "850181e4652dd023b7a98c58ae0d2d34bd487ee0cc3254aed6eda37307425907" \
  :s3:game-images
Result
2026/04/10 15:19:29 NOTICE: Config file "/Users/bertrand/.config/rclone/rclone.conf" not found - using defaults
2026/04/10 15:19:30 NOTICE: Failed to ls: operation error S3: ListObjects, https response error StatusCode: 403, RequestID: , HostID: , api error SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your key and signing method.
───

Other tested regions

I also tested:

• eu-west-3
• stub
• us-east-1

All of them return:

SignatureDoesNotMatch

So this does not seem to be a simple region mismatch.

───

Storage logs

Storage logs show:

• Missing signature for unsigned requests
• Invalid Access Key for wrong credentials
• SignatureDoesNotMatch when using the current configured access key + secret

This is what makes me suspect either:

1. a bug in S3 SigV4 validation for the built-in S3 endpoint in this setup
2. undocumented requirements for region/scope/canonical request


3. a compatibility issue with clients like rclone against the self-hosted S3 protocol endpoint

───

Question

Is there a known required client configuration for the self-hosted /storage/v1/s3 endpoint in storage-api:v1.48.26?

Specifically:

• Is REGION the correct signing region for S3 protocol clients?


• Is there another required env var for S3 protocol signing?
• Is the self-hosted S3 protocol currently expected to work with standard SigV4 clients like rclone / AWS SDK in this configuration?

If helpful, I can also provide:

• full docker-compose storage section
• exact storage logs
• result from AWS SDK client test...

adityakrmishra:
Fixes #44769

Description
This PR resolves a SignatureDoesNotMatch AWS SigV4 validation error when using the built-in S3 protocol on self-hosted Docker deployments via non-default URLs (e.g., 127.0.0.1, custom ports, or dedicated subdomains).

The Root Cause:
The storage container uses STORAGE_PUBLIC_URL to generate the canonical request during SigV4 verification. Previously, this was hardwired to ${SUPABASE_PUBLIC_URL} in docker-compose.yml. If a user pointed an S3 client (like rclone or aws cli) to a custom endpoint URL, the client signed the request using the custom host, but the storage container attempted to verify it using localhost:8000, resulting in a hostname mismatch and a failed HMAC signature.

The Fix:

Updated docker-compose.yml to use a bash fallback: STORAGE_PUBLIC_URL: ${STORAGE_PUBLIC_URL:-${SUPABASE_PUBLIC_URL}}. This keeps the default behavior intact for standard deployments but allows independent overrides.

Added a documentation block to .env.example explaining that for S3 protocol clients, STORAGE_PUBLIC_URL must be explicitly set to exactly match the --endpoint-url used by the client.

Note on Podman: As noted in the existing Compose file comments (line 8), Podman does not support nested variable interpolation. Podman users utilizing the S3 protocol on non-default URLs will need to set both variables explicitly in their .env file.

Testing Instructions
Start the stack and ensure a bucket exists.

Attempt to list the bucket using an S3 client targeting http://127.0.0.1:18000/storage/v1/s3 using the loaded S3_PROTOCOL_ACCESS_KEY_ID and secret.

Observe the SignatureDoesNotMatch failure.

Add STORAGE_PUBLIC_URL=http://127.0.0.1:18000 to the .env file, restart the storage container, and repeat the request. It will now succeed.

## Summary by CodeRabbit

* **New Features**
  * Added configurable storage endpoint URL support in Docker deployments. Users can now optionally specify a custom public URL for storage endpoints to match their infrastructure requirements. When not explicitly configured, the system automatically falls back to the default public URL, maintaining backward compatibility and simplifying standard deployments.


aantti:
I'm sorry, this isn't the correct PR. Storage uses `X-Forwarded-Prefix: /storage/v1` from the request to properly handle S3 signatures.

PR #44921, opened 2026-04-15
fix(studio): invert tooltip condition for restart failed tables button

### LINKED ISSUE #44920: Inverted tooltip condition on restart failed tables button
The tooltip on the \"Restart failed tables only\" dropdown item in the replication pipeline status never displays.

In `ReplicationPipelineStatus.tsx` (line 447), the condition is:

```tsx
text: hasErroredTables ? 'No tables require manual retry' : undefined,
```

The `DropdownMenuItemTooltip` component only renders the tooltip when the item is `disabled` AND `text` is not `undefined`. The item is disabled when `!hasErroredTables`. So:

- When no errored tables exist: button is disabled, but `text` is `undefined` (no tooltip shown)
- When errored tables exist: `text` is set, but button is enabled (tooltip component ignores it)

The tooltip never appears in either case. The fix is to invert the condition to `!hasErroredTables`.

Introduced in #44843.

oniani1:
Closes #44920

The tooltip on the "Restart failed tables only" dropdown item never displays. The condition `hasErroredTables ? 'No tables require manual retry' : undefined` is inverted -- `DropdownMenuItemTooltip` only renders the tooltip when the item is disabled, and the item is disabled when `!hasErroredTables`. So the text is set when the button is enabled (ignored) and undefined when disabled (no tooltip).

Flipped to `!hasErroredTables` so the tooltip explains why the button is disabled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tooltip logic in the replication pipeline status interface to correctly display "No tables require manual retry" when no manual intervention is needed.

### DIFF
--- a/apps/studio/components/interfaces/Database/Replication/ReplicationPipelineStatus/ReplicationPipelineStatus.tsx
+++ b/apps/studio/components/interfaces/Database/Replication/ReplicationPipelineStatus/ReplicationPipelineStatus.tsx
@@ -444,7 +444,7 @@ export const ReplicationPipelineStatus = () => {
                       tooltip={{
                         content: {
                           side: 'left',
-                          text: hasErroredTables ? 'No tables require manual retry' : undefined,
+                          text: !hasErroredTables ? 'No tables require manual retry' : undefined,
                         },
                       }}
                     >


iambriccardo:
Duplicated.

Issue #44642, opened 2026-04-07
Lock "lock:jabe-auth" was not released within 5000ms causes app to freeze when switching browser tabs


FactoryCommunication:
Environment:

@supabase/supabase-js version: 2.101.1
Browser: Chrome (Memory Saver: Balanced)
Framework: React 18 + Vite
Hosting: Vercel

Problem:
When switching to another Chrome tab or window for 20-30 seconds and returning to the app, the entire React app freezes completely. The page shows a loading spinner indefinitely with no errors in console. The only way to recover is to clear all cookies and localStorage manually.
Error message found in console:
@supabase/gotrue-js: Lock "lock:jabe-auth" was not released within 5000ms. 
This may indicate an orphaned lock from a component unmount (e.g., React Strict Mode). 
Forcefully acquiring the lock to recover.
What we tried:

Removing flowType: 'pkce'
Setting detectSessionInUrl: false
Setting lock: false
Custom storage without lock
Downgrading to version 2.39.7
Singleton pattern for client instance
visibilitychange listener with refreshSession()
Keepalive every 20 seconds

None of the above resolved the issue.
Workaround:
Auto-reload page after 20 seconds of tab inactivity via visibilitychange event.
Question:
Is there a way to completely disable the lock mechanism in @supabase/supabase-js auth? The lock seems to get orphaned when Chrome suspends background tabs.

adityakrmishra:
Hi @FactoryCommunication 

Yes, there is a way to completely bypass the navigator.locks mechanism right now! You can override the lock implementation in your client config with a simple no-op function.

Try updating your client instantiation to this:
```
const supabase = createClient('URL', 'KEY', {
  auth: {
    // Custom no-op lock to completely bypass navigator.locks
    lock: async (_name, _acquireTimeout, fn) => {
      return await fn();
    },
  }
})
```
This prevents the auth client from using Web Locks entirely, which should stop the background tab freezing.

However, the fact that your app completely freezes after the 5000ms "Forcefully acquiring" warning means there is likely a deadlock bug in the recovery logic itself. I'm going to dive into the codebase and see if I can patch the forceful recovery mechanism so it stops hanging the event loop!

adityakrmishra:
@FactoryCommunication  team, I've tracked down the root cause!

The issue is an infinite "steal cascade" triggered by Chrome's tab suspension, and a slight misunderstanding of how steal: true works in the W3C Web Locks spec.

The Anatomy of the Deadlock:

[...]

PR #44922, opened 2026-04-15
fix(studio): add circuit breaker to prevent auth-js steal cascade deadlock

adityakrmishra:
Mitigates #44642 for Studio

Description
This PR implements a circuit breaker in the Studio's debuggableNavigatorLock wrapper to protect the application from an infinite Web Lock steal cascade originating in @supabase/auth-js.

The Context:
When a Chrome tab is suspended, the background JavaScript context freezes. If navigatorLock is held, it triggers a 5000ms timeout and recovers via steal: true. However, the W3C spec for steal revokes exclusion but does not cancel the original callback. When the tab resumes, both the stolen and original promises execute concurrently, queuing against each other and triggering an infinite loop of timeouts and steals that permanently freezes the UI thread.

The Fix:
We cannot modify the root steal behavior from this repository, so this PR patches the Studio's consumption of it:

Wraps the navigatorLock call in a try/catch.

Uses a dual guard (e.name + string matching) to robustly detect the NavigatorLockAcquireTimeoutError thrown when auth-js detects a steal-back.

Bypasses the lock by immediately returning await fn(), exiting the infinite queuing cascade.

Logs a console.warn and fires captureException to track real-world occurrences via Sentry without crashing the client.

Testing Instructions
Spin up Studio locally.

Open the app in Chrome, force tab suspension (or simulate a heavily throttled CPU that delays the microtask queue past 5000ms during an auth refresh).

Resume the tab.

Verify the Circuit breaker tripped warning appears in the console and the dashboard remains responsive instead of deadlocking.

## Release Notes

* **Bug Fixes**
  * Enhanced resilience to lock acquisition failures by gracefully handling specific timeout conditions, allowing the application to continue operating in edge cases where synchronization issues occur.

### DIFF
--- a/packages/common/gotrue.ts
+++ b/packages/common/gotrue.ts
@@ -174,6 +174,43 @@ async function debuggableNavigatorLock<R>(
         bc.close()
       }
     })
+  } catch (e: any) {
+    // Circuit breaker for the steal-cascade deadlock (Issue #44642).
+    //
+    // When a Chrome tab is suspended, navigator.locks can become orphaned —
+    // the lock is held at the browser level but the JS holding it is frozen.
+    // auth-js recovers by using { steal: true }, but per the Web Locks spec
+    // the original holder's fn() continues running as an orphaned task. When
+    // it resumes and calls navigatorLock() again, auth-js detects that OUR
+    // lock was stolen and throws NavigatorLockAcquireTimeoutError with the
+    // "stole it" message. If we re-throw here, the caller gets an unhandled
+    // error and any pending auth operations never complete — the app freezes.
+    //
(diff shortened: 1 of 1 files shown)


7ttp (commented):
thanks for the pr 💚
 but this needs to be fixed upstream in `supabase/supabase-js/packages/core/auth-js` rather than in the studio

A proper fix is already being worked on by @mandarini and the auth team
so closing this in favor of that upstream work

ref: 
- supabase/supabase-js#936
-  supabase/supabase-js#2106
-  supabase/supabase-js#2178
- supabase/supabase-js#2235

> Thanks again for ur efforts towards contributing to Supabase !! 🙌💚

PR #44941, opened 2026-04-16
chore: types cleanup for react 19

djhi:
## Problem

While trying to update `react` to version `19`, I noticed type related errors that can be fixed in version `18`, mostly usage of `JSX.Element` instead of `ReactNode`.

## Solution

- Use `ReactNode` instead of `JSX.Element`
- Fix some invalid usage of `rechart`

## Release Notes

* **Refactor**
  * Standardized React component type annotations across the codebase for improved type consistency and flexibility.
  * Updated component prop types to accept a broader range of renderable content.

* **Bug Fixes**
  * Adjusted chart layout positioning to improve visual alignment.

### DIFF
--- a/apps/studio/components/ui/Charts/ComposedChart.tsx
+++ b/apps/studio/components/ui/Charts/ComposedChart.tsx
@@ -664,13 +664,13 @@ export function ComposedChart({
             }}
           />
         </RechartComposedChart>
-        <ChartHighlightActions
-          chartHighlight={chartHighlight}
-          updateDateRange={updateDateRange}
-          actions={highlightActions}
-          chartId={chartId}
-        />
       </Container>
+      <ChartHighlightActions
+        chartHighlight={chartHighlight}
+        updateDateRange={updateDateRange}
+        actions={highlightActions}
+        chartId={chartId}
+      />
       {data && (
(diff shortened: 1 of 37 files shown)


apps/studio/components/ui/Charts/ComposedChart.tsx
```diff
@@ -664,13 +664,13 @@ export function ComposedChart({
...
           />
         </RechartComposedChart>
-        <ChartHighlightActions
-          chartHighlight={chartHighlight}
-          updateDateRange={updateDateRange}
-          actions={highlightActions}
-          chartId={chartId}
-        />
       </Container>
+      <ChartHighlightActions
```
ivasilov, line 668: Is this intentional?

ivasilov (approved):

fsansalvadore (approved):

apps/studio/components/ui/Charts/ComposedChart.tsx
djhi: Yes, because `ResponsiveContainer` actually only accept a single `ReactElement` as its child. I tested and the  `ChartHighlightActions` still work as expected

PR #44946, opened 2026-04-16
docs(dart): require onError handler in onAuthStateChange examples

grdsdev:
## Summary

- Adds a warning note to `onAuthStateChange()` that an `onError` handler is **required**
- Updates all three Dart code examples (spotlight, specific-event, unsubscribe) to include `onError`

## Background

Network errors (e.g. a token refresh attempted while the device is offline) are emitted as stream errors on `onAuthStateChange`. If no `onError` handler is provided, Dart rethrows them as unhandled zone exceptions, crashing the app.

This is tracked in supabase/supabase-flutter#1281. The SDK itself already handles this internally (`supabase_auth.dart` has its own `onError`), but user-facing code that calls `.listen()` without `onError` is still vulnerable.

The companion SDK fix (example app + doc comment in `gotrue_client.dart`) is in a separate PR in `supabase/supabase-flutter`.

## Test plan

- [ ] Review the rendered diff in the Dart reference docs (`/docs/reference/dart/auth-onauthstatechange`)
- [ ] Confirm all three code examples now show `onError`
- [ ] Confirm the new note is visible in the Notes section

## Summary by CodeRabbit

* **Documentation**
  * Clarified auth-state change docs to require explicit error handling in listeners and updated all examples to show an error callback to avoid unhandled exceptions.
  * Updated the documented list of auth event names to the new set (including initialSession and mfaChallengeVerified) and adjusted examples accordingly.

### DIFF
--- a/apps/docs/spec/supabase_dart_v2.yml
+++ b/apps/docs/spec/supabase_dart_v2.yml
@@ -1502,19 +1502,6 @@ functions:
             (data) {
               final AuthChangeEvent event = data.event;
               final Session? session = data.session;
-
-              print('event: $event, session: $session');
-
-              switch (event) {
-                case AuthChangeEvent.initialSession:
-                case AuthChangeEvent.signedIn:
-                case AuthChangeEvent.signedOut:
-                case AuthChangeEvent.passwordRecovery:
-                case AuthChangeEvent.tokenRefreshed:
-                case AuthChangeEvent.userUpdated:
-                case AuthChangeEvent.userDeleted:
-                case AuthChangeEvent.mfaChallengeVerified:
-              }
             },
             onError: (error, stackTrace) {
               // Network errors (e.g. offline) are emitted here.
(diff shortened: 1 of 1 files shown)


apps/docs/spec/supabase_dart_v2.yml
```diff
@@ -1507,54 +1491,69 @@ functions:
...
-          });
+          final authSubscription = supabase.auth.onAuthStateChange.listen(
+            (data) {
+              final AuthChangeEvent event = data.event;
+              final Session? session = data.session;
+
+              print('event: $event, session: $session');
+
+              switch (event) {
+                case AuthChangeEvent.initialSession:
```
jeremenichelli: Honestly, this comments seem mega redundant, considering the event naming is quite on point. If we do want to list all the event, maybe a code snippet is not the best way since it can get away from the reading? 

A remarks list with each, or a type definition might be better

ChrisChinchilla (approved):

PR #44948, opened 2026-04-16
fix(studio): resolve realtime channel state desync and stale closures

adityakrmishra:
Fixes ##44947

Description
This PR fixes critical state-sync issues in the Studio's Realtime Inspector (useRealtimeMessages.ts).

Added isChannelPrivate to the useEffect dependency array. Previously, toggling a channel between public and private updated the UI but failed to tear down and rebuild the supabase.channel(). This left users unknowingly streaming on a public channel while the UI indicated RLS was active.

Refactored setRealtimeConfig to use a functional updater. In the CHANNEL_ERROR callback, spreading the closure-captured config object was silently overwriting newer configuration state (like refreshed tokens) with stale data captured when the effect initially mounted. Using prev => ({ ...prev, log: ... }) ensures we always append to the latest state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed realtime channel configuration to properly handle errors and respect privacy setting changes.

### DIFF
--- a/apps/studio/components/interfaces/Realtime/Inspector/useRealtimeMessages.ts
+++ b/apps/studio/components/interfaces/Realtime/Inspector/useRealtimeMessages.ts
@@ -196,7 +196,7 @@ export const useRealtimeMessages = (
 
         newChannel.unsubscribe()
         setChannel(undefined)
-        setRealtimeConfig({ ...config, channelName: '', enabled: false })
+        setRealtimeConfig((prev) => ({ ...prev, channelName: '', enabled: false }))
       }
     })
 
@@ -213,6 +213,7 @@ export const useRealtimeMessages = (
     enablePresence,
     filter,
     host,
+    isChannelPrivate,
     schema,
     table,
   ])


danjak-eng:
Thanks for putting this together! I spent some time reproducing the issue locally and wanted to share a few thoughts.

The fix looks correct to me and aligns with my implementation:

1. Adding `isChannelPrivate` to the dependency array is important here. Without it, the channel config can get out of sync when toggling between public/private, since the effect doesn’t re-run.

2. Refactoring `setRealtimeConfig` to use a functional update (`prev => { ... }`) resolves the stale closure issue in the `CHANNEL_ERROR` callback. Since this callback executes asynchronously, relying on the captured `config` object can overwrite the newer state (e.g., updated tokens or channel names), so using `prev` is definitely the safer approach.

One thing I tested locally was rapidly changing channel names and forcing connection errors. With this fix, the state consistently reflects the latest input, which wasn’t the case before.

Overall, this looks like a solid and well-scoped fix for the desync issues in the Realtime Inspector. Nice work and thank you for suggesting my review!

adityakrmishra:
Thanks so much for taking the time to test this locally, @danjak-eng  I really appreciate you checking those edge cases with rapid state changes—that confirms the functional updater is doing its job correctly. It's great to have your validation on this approach.

Hallidayo:

PR #44963, opened 2026-04-17
feat: iso27001 certificate

doublethink:
Edit: Can be merged, mgmt api deployed

Dashboard addition to frontend for access to the ISO 27001 certificate.

View for Team customers:
image

Resolves SEC-799

## Summary by CodeRabbit

* **New Features**
  * ISO 27001 certificate added to Documents with a Download action, confirmation modal, new-tab open on success, and error toast on failure.
  * Users without billing permission see a no-permission view; users missing entitlement see an “Upgrade to Team” prompt.

* **Refactor**
  * Upgrade-to-Team flows for SOC2 and related upgrade UI standardized to use the shared upgrade component.

### DIFF
--- a/apps/studio/components/interfaces/Organization/Documents/Documents.tsx
+++ b/apps/studio/components/interfaces/Organization/Documents/Documents.tsx
@@ -3,6 +3,7 @@ import { Fragment } from 'react'
 import { CustomDocument } from './CustomDocument'
 import { DPA } from './DPA'
 import { HIPAA } from './HIPAA'
+import { ISO27001 } from './ISO27001'
 import { SecurityQuestionnaire } from './SecurityQuestionnaire'
 import { SOC2 } from './SOC2'
 import { TIA } from './TIA'
@@ -46,6 +47,12 @@ export const Documents = () => {
 
       <ScaffoldContainer id="soc2" className="px-6 xl:px-10">
         <SOC2 />
+      </ScaffoldContainer>
+
+      <ScaffoldDivider />
+
+      <ScaffoldContainer id="iso27001" className="px-6 xl:px-10">
+        <ISO27001 />
       </ScaffoldContainer>
 
       <ScaffoldDivider />
--- a/apps/studio/components/interfaces/Organization/Documents/ISO27001.tsx
+++ b/apps/studio/components/interfaces/Organization/Documents/ISO27001.tsx
@@ -1,0 +1,144 @@
+import { PermissionAction } from '@supabase/shared-types/out/constants'
+import { Download } from 'lucide-react'
+import Link from 'next/link'
+import { useState } from 'react'
+import { toast } from 'sonner'
+import { Button } from 'ui'
+import ConfirmationModal from 'ui-patterns/Dialogs/ConfirmationModal'
+import { ShimmeringLoader } from 'ui-patterns/ShimmeringLoader'
+
+import {
+  ScaffoldSection,
+  ScaffoldSectionContent,
+  ScaffoldSectionDetail,
+} from '@/components/layouts/Scaffold'
+import NoPermission from '@/components/ui/NoPermission'
+import { getDocument } from '@/data/documents/document-query'
+import { useSendEventMutation } from '@/data/telemetry/send-event-mutation'
+import { useCheckEntitlements } from '@/hooks/misc/useCheckEntitlements'
+import { useAsyncCheckPermissions } from '@/hooks/misc/useCheckPermissions'
+import { useSelectedOrganizationQuery } from '@/hooks/misc/useSelectedOrganization'
+
+export const ISO27001 = () => {
+  const { data: organization } = useSelectedOrganizationQuery()
+  const slug = organization?.slug
+
+  const { mutate: sendEvent } = useSendEventMutation()
+  const { can: canReadSubscriptions, isLoading: isLoadingPermissions } = useAsyncCheckPermissions(
+    PermissionAction.BILLING_READ,
+    'stripe.subscriptions'
+  )
+  const { hasAccess: hasAccessToISO27001, isLoading: isLoadingEntitlement } = useCheckEntitlements(
+    'security.iso27001_certificate'
+  )
+
+  const [isOpen, setIsOpen] = useState(false)
+
+  const fetchISO27001 = async (orgSlug: string) => {
+    try {
+      const link = await getDocument({ orgSlug, docType: 'iso27001-certificate' })
+      if (link?.fileUrl) window.open(link.fileUrl, '_blank')
+      setIsOpen(false)
+    } catch (error: unknown) {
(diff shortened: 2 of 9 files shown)


staaldraad (approved):

apps/studio/components/interfaces/Organization/Documents/ISO27001.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { ShimmeringLoader } from 'ui-patterns/ShimmeringLoader'
+
+import {
+  ScaffoldSection,
+  ScaffoldSectionContent,
+  ScaffoldSectionDetail,
+} from '@/components/layouts/Scaffold'
+import NoPermission from '@/components/ui/NoPermission'
+import { getDocument } from '@/data/documents/document-query'
+import { useSendEventMutation } from '@/data/telemetry/send-event-mutation'
```
seanoliver, line 17: nit: `useSendEventMutation` is the deprecated path — new tracking should use `useTrack` from `lib/telemetry/track`, which auto-injects `organization` and `project` group properties. That said, all the sibling document components (SOC2, HIPAA, TIA, SecurityQuestionnaire) use this same hook, so staying consistent with the family makes sense here. Worth a dedicated follow-up PR to migrate all four at once rather than making this one the odd one out.

apps/studio/components/interfaces/Organization/Documents/ISO27001.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  const handleDownloadClick = () => {
+    if (!slug) return
+
+    sendEvent({
+      action: 'document_view_button_clicked',
+      properties: { documentName: 'ISO27001' },
+      groups: { organization: slug },
+    })
+    setIsOpen(true)
```
seanoliver, line 56: `document_view_button_clicked` fires when the user clicks Download, which just opens the NDA modal. We don't fire anything when the user actually accepts the NDA (real download) or cancels. That means we can't measure:

- NDA drop-off rate (how many people bail at the NDA step)
- Actual certificate download conversion

If we want to measure "does offering ISO 27001 drive Team upgrades / retention?" we'd need the accepted/downloaded event to close the funnel. Worth a follow-up PR that adds a `document_downloaded` (or `document_nda_accepted`) event across SOC2, HIPAA, TIA, and ISO27001 uniformly.

apps/studio/components/interfaces/Organization/Documents/ISO27001.tsx
```diff
@@ -0,0 +1,144 @@
...
+      </ScaffoldSectionDetail>
+      <ScaffoldSectionContent>
+        {isLoadingPermissions || isLoadingEntitlement ? (
+          <div className="@lg:flex items-center justify-center h-full">
+            <ShimmeringLoader className="w-24" />
+          </div>
+        ) : !canReadSubscriptions ? (
+          <NoPermission resourceText="access our ISO 27001 certificate" />
+        ) : !hasAccessToISO27001 ? (
+          <div className="@lg:flex items-center justify-center h-full">
```
seanoliver: fyi / confirmation: the `source=iso27001` query param flows through `PlanUpdateSidePanel.tsx` into the `origin` property on the `studio_pricing_side_panel_opened` event, so upgrade attribution from this CTA works out of the box. The casing is inconsistent with most other sources in the app (which use camelCase like `securityQuestionnaire`), but it matches the existing SOC2 precedent within the Documents family — fine to keep as-is for internal consistency.

apps/studio/components/interfaces/Organization/Documents/ISO27001.tsx
```diff
@@ -0,0 +1,144 @@
...
+            <li>
+              This Confidentiality Notice does not substitute or supersede any agreement between us
+              and the Customer, or any internal rules or policies that the Customer requires you to
+              comply with in your access to and use of confidential information. However, your
+              failure to comply with this Confidentiality Notice may be used to determine whether
+              the Customer has complied with its confidentiality obligations to us.
+            </li>
+          </ol>
+        </ConfirmationModal>
+      </ScaffoldSectionContent>
```
seanoliver: Can you confirm legal explicitly signed off on the ISO 27001 NDA copy (and not just the feature rollout under SEC-799)? The text here is essentially the SOC2 NDA with "SOC 2" swapped for "ISO 27001". Likely fine since both are confidentiality agreements for audit materials, but flagging because NDA text re-use without explicit legal review is a common trap.

packages/common/telemetry-constants.ts
```diff
@@ -2016,7 +2016,7 @@ export interface DocumentViewButtonClickedEvent {
     /**
      * The name of the document being viewed, e.g. TIA, SOC2, Standard Security Questionnaire
      */
-    documentName: 'TIA' | 'SOC2' | 'Standard Security Questionnaire'
+    documentName: 'TIA' | 'SOC2' | 'ISO27001' | 'Standard Security Questionnaire'
```
seanoliver, line 2019: nit: the JSDoc comment a few lines above still says `e.g. TIA, SOC2, Standard Security Questionnaire` — worth adding `ISO27001` to the example list for parity with the union type.

apps/studio/components/interfaces/Organization/Documents/Documents.tsx
```diff
@@ -50,6 +51,12 @@ export const Documents = () => {
 
       <ScaffoldDivider />
 
+      <ScaffoldContainer id="iso27001" className="px-6 xl:px-10">
+        <ISO27001 />
+      </ScaffoldContainer>
+
+      <ScaffoldDivider />
```
seanoliver: Ordering question — placing ISO 27001 between SOC 2 and HIPAA puts Team-entitled certs together, which reads well. Just confirming this is intentional vs. alphabetical or by importance; either way is fine, but worth a beat of thought since the order is what users see.

seanoliver (approved):
LGTM! Some minor/optional nits called out inline.

apps/studio/components/interfaces/Organization/Documents/ISO27001.tsx
```diff
@@ -0,0 +1,144 @@
...
+          <div className="@lg:flex items-center justify-center h-full">
+            <ShimmeringLoader className="w-24" />
+          </div>
+        ) : !canReadSubscriptions ? (
+          <NoPermission resourceText="access our ISO 27001 certificate" />
+        ) : !hasAccessToISO27001 ? (
+          <div className="@lg:flex items-center justify-center h-full">
+            <Button asChild type="default">
+              <Link href={`/org/${slug}/billing?panel=subscriptionPlan&source=iso27001`}>
+                Upgrade to Team
```
joshenlim: will swap these to upgrade plan button - that component handles if the user doesn't have perms to upgrade the plan, and support sending a request to billing owners to upgrade instead

joshenlim (approved):
Verified changes on preview!

PR #45005, opened 2026-04-18
feat: show tax preview for credit top-ups

ignaciodob:
## Summary

- Add tax preview to the credit top-up flow by integrating the new `POST /organizations/{slug}/billing/credits/preview` endpoint
- Show a price breakdown (credit amount, tax line item, total) in the top-up dialog before the user confirms payment
- Handle all three `tax_status` states: show tax when `calculated`, hide the line when `not_applicable`, show an inline warning when `failed`

## Behavior

- Preview fires as soon as a valid amount (300–2000) is entered; address and tax ID are optional and refine the estimate
- Amount and address changes are debounced at 1s to avoid excessive API calls

## Test plan

- [ ]  Open credit top-up dialog - verify preview appears with default $300 amount
- [ ]  Change amount within 300–2000 - verify preview updates after debounce
- [ ]  Enter amount outside range (e.g. 100 or 3000) - verify preview hides and validation error shows
- [ ]  Add a new payment method with a billing address in a taxed region - verify tax line item appears
- [ ]  Add a new payment method with no tax jurisdiction - verify no tax line, just total
- [ ]  Complete a top-up - verify the charge goes through and dialog closes

image

## Summary by CodeRabbit

* **New Features**
  * Credit top-up preview with charge breakdown (credit, tax or tax note, total).
  * Onboarding survey endpoint and ISO 27001 certificate availability for organizations.

* **Updates**
  * Improved top-up UX: debounced address/tax inputs, enforced min/max amount validation, preview-driven form state, and submit disabled while preview is loading/stale.
  * API docs wording changed to “temporary access configuration.”

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
@@ -4,6 +4,7 @@ import { Elements } from '@stripe/react-stripe-js'
 import { loadStripe, PaymentIntentResult } from '@stripe/stripe-js'
 import { PermissionAction, SupportCategories } from '@supabase/shared-types/out/constants'
 import { useQueryClient } from '@tanstack/react-query'
+import { useDebounce } from '@uidotdev/usehooks'
 import { AlertCircle, Info } from 'lucide-react'
 import { useTheme } from 'next-themes'
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
@@ -14,6 +15,7 @@ import {
   AlertDescription_Shadcn_,
   AlertTitle_Shadcn_,
   Button,
+  cn,
   Dialog,
   DialogContent,
   DialogDescription,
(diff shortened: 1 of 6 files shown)


apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
```diff
@@ -87,6 +94,47 @@ export const CreditTopUp = ({ slug }: { slug: string | undefined }) => {
...
+    isFetching: creditPreviewIsFetching,
+    isSuccess: creditPreviewInitialized,
+  } = useCreditTopUpPreview(
+    {
+      slug,
+      amount: validAmount,
+      address: debouncedAddress,
+      taxId: debouncedTaxId ?? undefined,
+    },
+    { enabled: !!validAmount }
```
adityakrmishra: Gotta boost the CodeRabbit bot on this one! Since this component mounts silently on the billing page and `validAmount` defaults to 300 immediately, we end up POSTing to `/billing/credits/preview` the second any user opens their Org Billing settings—even if they never actually open the top-up dialog. 

Let's add `topUpModalVisible` to the `enabled` block so we don't cook our Stripe limits for free!

```tsx
{ enabled: !!validAmount && topUpModalVisible }

File: apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx [hunk 2]
```diff
@@ -87,6 +94,47 @@ export const CreditTopUp = ({ slug }: { slug: string | undefined }) => {
...
+  const [latestAddress, setLatestAddress] = useState<CustomerAddress>()
+  const [latestTaxId, setLatestTaxId] = useState<CustomerTaxId | null>()
+
+  const billingAddress = useAsDefaultBillingAddress ? latestAddress : undefined
+  const billingTaxId = useAsDefaultBillingAddress ? latestTaxId : null
+  const debouncedAddress = useDebounce(billingAddress, 1000)
+  const debouncedTaxId = useDebounce(billingTaxId, 1000)
+
+  const watchedAmount = form.watch('amount')
+  const debouncedAmount = useDebounce(watchedAmount, 1000)
```
adityakrmishra, line 106: yo! noticing a pretty high-stakes UI race condition here. Because `useDebounce` delays the preview update by 1s, there's a window where a user can change the amount to $2000 and instantly click "Top Up" before the debounce loop finishes. 

Since `onSubmit` grabs the live form data, we'll bill them for $2000, but the screen they confirmed explicitly said "Total due today: $300". 😬 

What if we disable the submit button whenever the preview is "stale" so they can only pay exactly what they see?

```
tsx
const isPreviewStale = 
  watchedAmount !== debouncedAmount || 
  billingAddress !== debouncedAddress || 
  billingTaxId !== debouncedTaxId;

// And on the Button down below:
<Button 
  loading={executingTopUp || paymentConfirmationLoading}
  disabled={isPreviewStale || creditPreviewIsFetching} 
>

**Review by adityakrmishra — COMMENTED** [review]
```md
yo! great work on the new endpoint integration! 

Left a couple of notes inline—mostly around a tricky UI race condition with the debounced payment amount. We definitely want to gate that submit button so we don't accidentally overcharge anyone while the UI is catching up. Let's iron that out!

**Review by asperpharma — COMMENTED** [review]
aprove

File: apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx [hunk 3]
```diff
@@ -324,6 +381,50 @@ export const CreditTopUp = ({ slug }: { slug: string | undefined }) => {
...
+
+                  <div className="flex items-center justify-between gap-2 text-foreground text-base">
+                    <div className="py-2">Total due today</div>
+                    <div className="py-2 text-right tabular-nums" translate="no">
+                      {formatCurrency(creditPreview.total)}
+                    </div>
+                  </div>
+                </div>
+              )}
             </DialogSection>
```
awaseem, line 402: nit: This would make for its own component, rather than having all this JSX in here it would be nice to split this out and refactor. Pass props for values needed

awaseem (approved):
Just minor comments, looks good otherwise

PR #45019, opened 2026-04-19
fix(pg-meta): require replica_identity_index when replica_identity is INDEX

Akash504-ai:

YES

Bug fix

When updating a table with `replica_identity = 'INDEX'`, it is possible to omit the `replica_identity_index`.  
This results in invalid SQL being generated:

`REPLICA IDENTITY USING INDEX`

PostgreSQL requires an index name for this operation.

No existing issue found. This PR fixes a validation gap discovered during development.

## What is the new behavior?

- Throws an error if `replica_identity` is set to `INDEX` without providing `replica_identity_index`
- Ensures generated SQL is always valid

## Additional context

- Added validation in `pg-meta-tables.ts`
- Added test case to verify an error is thrown when the index name is missing

This aligns pg-meta behavior with PostgreSQL requirements for `REPLICA IDENTITY USING INDEX`.

## Release Notes

* **Bug Fixes**
  * Enhanced validation for database replica identity configuration to prevent incomplete or invalid configurations.

* **Tests**
  * Added test coverage for replica identity validation scenarios.


Akash504-ai:
@charislam this adds validation to prevent generating invalid SQL when REPLICA IDENTITY is set to INDEX without an index name. Would appreciate your review when you get a chance 🙏

charislam (approved):
Thanks @Akash504-ai!

PR #45034, opened 2026-04-20
[FE-3023] feat(studio): default privileges toggle at project creation

alaister:
Screenshot 2026-04-20 at 3 02 37 PM

Adds a "Default privileges for new entities" checkbox under "Enable Data API" in both the main create flow and the Vercel deploy-button flow. Default checked (current behaviour). When unchecked, runs `buildDefaultPrivilegesSql('revoke')` after the base init script so new entities in `public` aren't auto-granted to `anon` / `authenticated` / `service_role`.

This PR decouples the two surfaces:

- **`tableEditorApiAccessToggle`** — unchanged; still gates only the integrations → Data API settings UI.
- **`dataApiRevokeOnCreateDefault`** (new) — controls only the default state of the new checkbox at project creation. `true` → checkbox unchecked by default (revoke runs); `false`/absent → checkbox checked by default (no behaviour change).

The new flag is already live in PostHog at **0% rollout, off for everyone**, so shipping this PR changes nothing until the flag is explicitly flipped.

## Added

- `apps/studio/hooks/misc/useDataApiRevokeOnCreateDefault.ts` — reads the new PostHog flag. Returns `false` in `IS_TEST_ENV` so existing E2E flows don't silently change default behaviour.
- Checkbox UI in `SecurityOptions.tsx` (main flow) and `pages/integrations/vercel/[slug]/deploy-button/new-project.tsx` (Vercel flow), with copy matching the integrations → Data API settings page.
- Tooltip + dimmed state for the main-flow checkbox when "Enable Data API" is unchecked (can't configure default privileges if Data API is off).
- Telemetry: `dataApiDefaultPrivilegesGranted` (raw checkbox value) and `dataApiRevokeOnCreateDefaultEnabled` (raw flag, conditionally included using the existing raw-flag pattern so undefined flag state → omitted property, not `false`).
- Vitest unit tests for the new hook.

## Changed

- `pages/new/[slug].tsx`: removed the `false &&` rollback guard. Revoke SQL now runs only when `dataApi && !dataApiDefaultPrivileges`. Dropped the now-unused `useDataApiGrantTogglesEnabled` import.
- `pages/integrations/vercel/[slug]/deploy-button/new-project.tsx`: this flow was **never rolled back** — it still ran revoke whenever `tableEditorApiAccessToggle` was on for a user. Now correctly gated on the new flag + checkbox state.
- `packages/common/telemetry-constants.ts`: added the two new properties and corrected the `tableEditorApiAccessToggleEnabled` docstring (it no longer claims to control project-creation revoke behaviour).

## Kill switch

Flipping `dataApiRevokeOnCreateDefault` to off in PostHog fully disables the revoke SQL for new projects without needing a redeploy — the checkbox just defaults to checked again.

## Follow-ups (not blockers)

- joshenlim's review comments on PR 43704: (1) Auth Policies table row incorrectly showing "exposed via Data API" based on schema-level check instead of table-level at `apps/studio/components/interfaces/Auth/Policies/PolicyTableRow/index.tsx:64`; (2) Data API integrations page showing zero exposed tables even after exposing one. Both unrelated to this PR but will be more visible once the checkbox lands.
- Once this flag fully rolls out, the old `tableEditorApiAccessToggle` docstring/comments elsewhere should stop claiming it controls project creation.

## To test

- **Flag off (default state, simulates post-merge):** create a project with and without "Enable Data API" checked. The new "Default privileges for new entities" checkbox should default to **checked**. Submitting should produce an identical result to today — new tables in `public` are reachable via the Data API.
- **Flag on (simulate rollout):** override the flag locally. The checkbox should default to **unchecked**. Creating a project with it unchecked should run the revoke SQL; create a new table in `public` afterwards and confirm it's not reachable via the Data API until grants are added.
- **Enable Data API off:** the new checkbox should render disabled + dimmed with a tooltip reading "Enable the Data API to configure default privileges." The revoke SQL should not run in this case regardless of checkbox state.
- **Vercel flow:** repeat at `/integrations/vercel/<slug>/deploy-button/new-project` — verify both checkbox states.

## Summary by CodeRabbit

* **New Features**
  * Added an "Automatically expose new tables and functions" checkbox to project creation and Vercel deploy flow; enabled only when Data API is available (disabled with tooltip otherwise) and affects initial project provisioning.

* **Telemetry**
  * Tracks exposure of the default-privileges control and includes checkbox state and feature-flag status on project-creation submissions.

* **Tests**
  * Added tests for flag behavior, exposure tracking, deduplication, and submission telemetry.

### DIFF
--- a/apps/studio/pages/integrations/vercel/[slug]/deploy-button/new-project.tsx
+++ b/apps/studio/pages/integrations/vercel/[slug]/deploy-button/new-project.tsx
@@ -317,8 +317,8 @@ const CreateProject = () => {
       <div className="py-2 pb-4">
         <Checkbox
           name="dataApiDefaultPrivileges"
-          label="Default privileges for new entities"
-          description="When enabled, new tables and functions in the public schema are automatically accessible via the Data API. We recommend disabling this and manually granting access to each new entity."
+          label="Automatically expose new tables and functions"
+          description="Grants privileges to Data API roles by default, exposing new tables and functions. We recommend disabling this to control access manually."
           checked={dataApiDefaultPrivileges}
           onChange={(e) => setDataApiDefaultPrivileges(e.target.checked)}
         />
--- a/apps/studio/pages/integrations/vercel/[slug]/deploy-button/new-project.tsx
+++ b/apps/studio/pages/integrations/vercel/[slug]/deploy-button/new-project.tsx
@@ -16,7 +16,7 @@ import { useIntegrationVercelConnectionsCreateMutation } from '@/data/integratio
 import { useVercelProjectsQuery } from '@/data/integrations/integrations-vercel-projects-query'
 import { useOrganizationsQuery } from '@/data/organizations/organizations-query'
 import { useProjectCreateMutation } from '@/data/projects/project-create-mutation'
-import { useDataApiGrantTogglesEnabled } from '@/hooks/misc/useDataApiGrantTogglesEnabled'
+import { useDataApiRevokeOnCreateDefaultEnabled } from '@/hooks/misc/useDataApiRevokeOnCreateDefault'
 import { useSelectedOrganizationQuery } from '@/hooks/misc/useSelectedOrganization'
 import { BASE_PATH, PROVIDERS } from '@/lib/constants'
 import { getInitialMigrationSQLFromGitHubRepo } from '@/lib/integration-utils'
@@ -63,7 +63,10 @@ const CreateProject = () => {
   const [dbRegion, setDbRegion] = useState(PROVIDERS.AWS.default_region.displayName)
 
   const snapshot = useIntegrationInstallationSnapshot()
-  const isDataApiGrantTogglesEnabled = useDataApiGrantTogglesEnabled()
+  const isDataApiRevokeOnCreateDefault = useDataApiRevokeOnCreateDefaultEnabled()
+  const [dataApiDefaultPrivileges, setDataApiDefaultPrivileges] = useState(
+    !isDataApiRevokeOnCreateDefault
+  )
 
   async function checkPasswordStrength(value: string) {
     const { message, strength } = await passwordStrength(value)
@@ -145,7 +148,7 @@ const CreateProject = () => {
       if (migrationSql) dbSqlParts.push(migrationSql)
       toast.success(`Done fetching initial migrations`, { id })
     }
-    if (isDataApiGrantTogglesEnabled) {
+    if (!dataApiDefaultPrivileges) {
       dbSqlParts.push(buildDefaultPrivilegesSql('revoke'))
     }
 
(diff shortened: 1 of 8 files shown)


charislam (approved):

apps/studio/pages/integrations/vercel/[slug]/deploy-button/new-project.tsx
```diff
@@ -287,6 +290,15 @@ const CreateProject = () => {
...
       </div>
+      <div className="py-2 pb-4">
+        <Checkbox
+          name="dataApiDefaultPrivileges"
+          label="Default privileges for new entities"
+          description="When enabled, new tables and functions in the public schema are automatically accessible via the Data API. We recommend disabling this and manually granting access to each new entity."
+          checked={dataApiDefaultPrivileges}
+          onChange={(e) => setDataApiDefaultPrivileges(e.target.checked)}
+        />
+      </div>
```
Comment (seanoliver) [inline apps/studio/pages/integrations/vercel/[slug]/deploy-button/new-project.tsx:325]: This file has no telemetry wired up, so the new checkbox is invisible to analytics: we won't see toggle rates for Vercel users or be able to compare flag-cohort behavior across the two flows.

apps/studio/pages/new/[slug].tsx
```diff
@@ -269,10 +271,14 @@ const Wizard: NextPageWithLayout = () => {
...
           enableRlsEventTrigger: form.getValues('enableRlsEventTrigger'),
           dataApiEnabled: form.getValues('dataApi'),
+          dataApiDefaultPrivilegesGranted: form.getValues('dataApiDefaultPrivileges'),
           useOrioleDb: form.getValues('useOrioleDb'),
           ...(tableEditorApiAccessToggleFlag !== undefined && {
             tableEditorApiAccessToggleEnabled: tableEditorApiAccessToggleFlag,
           }),
+          ...(dataApiRevokeOnCreateDefaultFlag !== undefined && {
+            dataApiRevokeOnCreateDefaultEnabled: dataApiRevokeOnCreateDefaultFlag,
+          }),
```
Comment (seanoliver) [inline apps/studio/pages/new/[slug].tsx:282]: This tracks the flag cohort + user's checkbox choice at submission time, which answers "what did completers pick?" but not "did flipping the default hurt completion rate?" If users in the rollout cohort see "We recommend disabling this" and bail out of project creation, we won't see it because they never reach this event.

We'd ideally add an exposure event (e.g. project_creation_default_privileges_exposed) that fires when the checkbox becomes visible (carrying dataApiRevokeOnCreateDefaultEnabled so we can bucket by cohort), then compare submit-rate between the two buckets. And also wrap it in a PostHog experiment with control and variation variants.

Happy to help with either/both of these!

seanoliver (commented):
LGTM overall but one thing to flag on the analytics/growth side before we start rolling it out:

My main concern is that we're adding a new checkbox to the project-creation funnel, and right now we can't measure whether it affects completion rate. Submission-only telemetry tells us what completers picked, but not whether the extra checkbox + the "we recommend disabling this" copy pushed anyone out of the funnel entirely. Same concern applies harder on the Vercel side, which has zero analytics at all.

Considering how important this funnel step is to activations, I'd treat the exposure tracking as a prerequisite for flipping the flag. Otherwise we're gambling on a conversion-sensitive change without a way to measure it.

**Here is a PR with the telemetry changes implemented (if helpful):** https://github.com/supabase/supabase/pull/45067

seanoliver (approved):
Approving to unblock!

PR #45035, opened 2026-04-20
feat(etl): Add UI elements to disable external replication

### LINKED ISSUE #44920: Inverted tooltip condition on restart failed tables button

iambriccardo:
This PR implements a new UI for disabling external replication. This is needed since ETL installs some triggers and schema in the customer DB for proper operation but we want to allow our customers to remove it easily if they stop using ETL.

It also fixes: https://github.com/supabase/supabase/issues/44920

Related infra PR: https://github.com/supabase/platform/pull/31898 

## Summary by CodeRabbit

* **New Features**
  * UI control to disable external replication with confirmation dialog.
  * Platform endpoints: preview credit top-ups and fetch ISO 27001 certificate URL.

* **Improvements**
  * Clarified copy: distinguishes read replicas vs external replication and unified “external replication” wording.

* **Bug Fixes**
  * Fixed tooltip text for “Restart failed tables only.”

* **Removed**
  * Removed the “Connect tables” / ETL flow from Analytics Buckets.


apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
joshenlim, line 1: JFYI this UI isn't being used atm (we previously decided to de-couple Analytics Buckets from ETL)
I might just go ahead to clean this up given that I don't think we're gonna pursue this anytime soon although cc @iambriccardo JFYI if you reckon that we should just leave this in!

apps/studio/data/replication/delete-tenant-mutation.ts
```diff
@@ -0,0 +1,72 @@
...
+
+async function deleteReplicationTenant(
+  { projectRef, tenantId }: DeleteReplicationTenantParams,
+  signal?: AbortSignal
+) {
+  if (!projectRef) throw new Error('projectRef is required')
+  if (!tenantId) throw new Error('tenantId is required')
+
+  const baseUrl = API_URL?.replace('/platform', '')
+  const url = `${baseUrl}/platform/replication/${projectRef}/tenants/${tenantId}`
```
joshenlim: this looks wrong - we should be using `del` from data/fetchers as it'll automatically handle the baseUrl logic here + have proper type checking for a valid mgmt API endpoint

is this a new mgmt API endpoint?

apps/studio/components/interfaces/Storage/AnalyticsBuckets/AnalyticsBucketDetails/ConnectTablesDialog.tsx
iambriccardo: Yeah, go ahead!

apps/studio/data/replication/delete-tenant-mutation.ts
iambriccardo: Yes, it's a new endpoint. I am in the process of improving and adding everything.

joshenlim: i can't seem to find the corresponding API types for this - is this dependent on an API PR? I can't seem to find the corresponding controller in the infra repo on latest develop

iambriccardo: I am now adding the endpoint in the backend and will generate the types!

apps/studio/components/interfaces/Database/Replication/DisableExternalReplicationDialog.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onOpenChange: (open: boolean) => void
+}
+
+export const DisableExternalReplicationDialog = ({
+  projectRef,
+  tenantId,
+  visible,
+  onOpenChange,
+}: DisableExternalReplicationDialogProps) => {
+  const [isSubmitting, setIsSubmitting] = useState(false)
```
joshenlim: don't need a separate useState for submitting, can use `isPending` from the mutation

apps/studio/components/interfaces/Database/Replication/Destinations.tsx
```diff
@@ -176,10 +216,41 @@ export const Destinations = () => {
...
+                  </DropdownMenu>
+
+                  <DisableExternalReplicationDialog
+                    projectRef={projectRef}
+                    tenantId={externalReplicationSource.tenant_id}
+                    visible={showDisableExternalReplicationDialog}
+                    onOpenChange={setShowDisableExternalReplicationDialog}
+                  />
+                </>
+              )}
```
joshenlim: i may opt to do this another way - a bit odd to slot in a `...` button between the "create destination" and docs button

apps/studio/data/replication/delete-tenant-mutation.ts
joshenlim: gotcha! i'll come back to this PR again later then 🙂 
no worries about the other comments that i left btw - i'll help to clean up later when the PR's ready for review again 🙏

apps/studio/components/interfaces/Database/Replication/Destinations.tsx
```diff
@@ -99,13 +103,45 @@ export const Destinations = () => {
...
+  const canDisableExternalReplication =
+    isSourcesSuccess &&
+    !isSourcesError &&
+    isDestinationsSuccess &&
+    !isDestinationsError &&
+    isPipelinesSuccess &&
+    !isPipelinesError &&
+    !!externalReplicationSource &&
+    destinations.length === 0 &&
+    pipelines.length === 0
```
joshenlim, line 128: this feels overly complicated 😅 can just rely on Success states, no need for error

joshenlim (approved):
Approving to unblock for now - unable to test on preview as the corresponding API PR has not been merged

@iambriccardo do test on the preview before merging this one 🙂 🙏

iambriccardo:
Tested and deployed!

PR #45042, opened 2026-04-20
fix: improve performances on large schema image export

djhi:
## Problem

When users export a large schema, the UI becomes unresponsive for a long time.
This is because the underlying `html-to-image` library calls `getComputedStyle` for every node.

## Solution

- Upgrade `html-to-image` to its latest version
- Use the new `includeStyleProperties` property to call `getComputedStyle` only once
- Extract the image export logic into a new hook

## How to test

- Open https://studio-staging-git-gildasgarcia-fe-2998-suggest-e7fb9e-supabase.vercel.app/dashboard/project/pdmusqfyrsascxykhlge/database/schemas?schema=auth 
- Rearrange tables so that they are all visible
- Export the schema as png
- It should takes (~10-15secs)

- Do the same in this PR preview https://studio-staging-gy13zepyf-supabase.vercel.app/dashboard/project/pdmusqfyrsascxykhlge/database/schemas?schema=auth
- It should takes ~3-5secs

## Summary by CodeRabbit

* **Improvements**
  * Improved schema export: more reliable PNG/SVG exports that better preserve visual styling, show progress state during downloads, and surface success/error notifications.
* **Chores**
  * Updated image-export library to a newer version for improved compatibility and performance.

### DIFF
--- a/apps/studio/components/interfaces/Database/Schemas/DefaultEdge.tsx
+++ b/apps/studio/components/interfaces/Database/Schemas/DefaultEdge.tsx
@@ -8,9 +8,10 @@ import {
   useReactFlow,
 } from '@xyflow/react'
 import { ArrowLeft, ArrowRight } from 'lucide-react'
-import { useEffect, useRef, useState } from 'react'
+import { useState } from 'react'
 import { Badge, cn } from 'ui'
 
+import { useSchemaGraphContext } from './SchemaGraphContext'
 import { EdgeData } from './Schemas.constants'
 import { useQuerySchemaState } from '@/hooks/misc/useSchemaQueryState'
 import { useStaticEffectEvent } from '@/hooks/useStaticEffectEvent'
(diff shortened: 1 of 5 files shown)


apps/studio/components/interfaces/Database/Schemas/DefaultEdge.tsx
```diff
@@ -46,13 +48,13 @@ export const DefaultEdge = ({
...
   })
-
   return (
     <>
       <BaseEdge
         id={id}
         path={edgePath}
-        className={cn(selected ? '!stroke-brand' : undefined)}
+        className={cn(selected ? '!stroke-brand' : isDownloading ? '!stroke-black' : undefined)}
+        stroke="#000000"
```
djhi, line 57: This is required in latest `html-to-image` version, otherwise the lines are not visible on the white background

charislam (approved):

PR #45050, opened 2026-04-20
fix(studio): don't scan dollar-quoted bodies for DDL in SQL editor

alaister:
Fixes a false positive in the CREATE-TABLE-without-RLS warning modal added in #45008. The warning was firing on `CREATE FUNCTION` statements because the `SELECT..INTO` detector was matching plpgsql variable assignments inside `$$…$$` function bodies.

Reported example that triggered the modal with no table actually being created:

```sql
create or replace function schema_checks()
returns jsonb
language plpgsql
as $$
declare
  ret jsonb;
begin
  select jsonb_build_object('value', 'ok') into ret;
  return ret;
end;
$$;
```

**Changed:**
- `SQLEventParser.match()` now strips the body of `$tag$…$tag$` blocks before running detectors. Tags are kept as markers; content is blanked out so function bodies, DO blocks, and dollar-quoted string literals are never scanned as DDL.
- Updated a pre-existing parser test that asserted the buggy behaviour (it expected `CREATE TABLE fake` inside a `$$…$$` string literal to be detected — `$$…$$` is a string literal in Postgres, not DDL).

**Added:**
- Regression tests in `SQLEditor.utils.test.ts` covering: the exact reported function, DO blocks with `select into`, `create table` text inside a function body, mixed top-level `CREATE TABLE` + function with `INTO` assignments, and custom `$body$…$body$` tags.
- Parser-level regression test in `sql-event-parser.test.ts`.

## To test

- In the SQL editor, paste the function from the Slack report and run it — the RLS warning modal should not appear.
- Run `create table foo (id int8 primary key);` on its own — modal still appears as before.
- Run `create table foo (id int8); create or replace function bar() returns int language plpgsql as $$ declare v int; begin select 1 into v; return v; end; $$;` — modal should flag only `foo`, not `v`.
- Run an existing destructive query (`drop table x`) — unaffected, modal still works.

## Summary by CodeRabbit

* **Bug Fixes**
  * Parser no longer treats DDL/DML-like text inside PL/pgSQL functions, DO blocks, or dollar-quoted bodies (including nested/custom tags) as top-level CREATE TABLE/SELECT INTO, preventing false detections and UI warnings.

* **Tests**
  * Added unit and e2e regression tests covering dollar-quoted blocks, nested dollar tags, DO blocks, SELECT INTO inside functions, and positive controls with a real top-level CREATE TABLE.

### DIFF
--- a/apps/studio/lib/sql-event-parser.ts
+++ b/apps/studio/lib/sql-event-parser.ts
@@ -49,10 +49,19 @@ export class SQLEventParser {
     return identifier?.replace(/["`']/g, '').replace(/\.$/, '')
   }
 
+  // Blank out the body of $tag$...$tag$ blocks (PL/pgSQL function bodies, DO
+  // blocks, dollar-quoted string literals) so their contents aren't scanned for
+  // DDL. A `select ... into var` inside a function body is variable assignment,
+  // not table creation, and would otherwise trip the SELECT..INTO detector.
+  private stripDollarQuoteBodies(sql: string): string {
+    return sql.replace(/(\$[a-zA-Z0-9_]*\$)[\s\S]*?\1/g, '$1$1')
+  }
+
   private match(sql: string): TableEventDetails | null {
+    const scannable = this.stripDollarQuoteBodies(sql)
     for (const { type, patterns } of SQLEventParser.DETECTORS) {
       for (const pattern of patterns) {
-        const match = sql.match(pattern)
+        const match = scannable.match(pattern)
         if (match?.groups) {
           return {
             type,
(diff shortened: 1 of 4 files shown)


apps/studio/lib/sql-event-parser.ts
```diff
@@ -113,7 +129,11 @@ export class SQLEventParser {
   }
 
   getTableEvents(sql: string): TableEventDetails[] {
-    const statements = this.splitStatements(this.removeComments(sql))
+    // Order matters: strip dollar-quote bodies first so comment syntax inside
+    // a function body (which is just literal text in Postgres) isn't treated
+    // as a comment by removeComments, and so inner semicolons inside the body
+    // can't confuse splitStatements.
+    const statements = this.splitStatements(this.removeComments(this.stripDollarQuoteBodies(sql)))
```
alaister, line 136: This file contains the actual changes; the rest are just tests.

awaseem (approved):

PR #45054, opened 2026-04-20
docs: remove claim flow admonition in docs

w3b6x9:
## Summary by CodeRabbit

* **Documentation**
  * Removed an informational notice from the Supabase for Platforms integration guide that previously mentioned access restrictions for the claim flow feature.


ChrisChinchilla (approved):
If this is indeed true, lgtm

w3b6x9:
> If this is indeed true, lgtm

@ChrisChinchilla yep, it's not gated

PR #45057, opened 2026-04-20
fix(docs): remove edge function Authorization header from cron example

fadymak:
Removes `Authorization` header from cron quickstart example to align with the new API keys.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Edge Function invocation example in the quickstart guide: the request headers now use the publishable API key header alongside Content-Type instead of an Authorization Bearer header.

### DIFF
--- a/apps/docs/content/guides/cron/quickstart.mdx
+++ b/apps/docs/content/guides/cron/quickstart.mdx
@@ -279,7 +279,7 @@ select
     select
       net.http_post(
           url:='https://project-ref.supabase.co/functions/v1/function-name',
-          headers:=jsonb_build_object('Content-Type','application/json', 'Authorization', 'Bearer ' || 'YOUR_ANON_KEY'),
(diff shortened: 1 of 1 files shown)


apps/docs/content/guides/cron/quickstart.mdx
```diff
@@ -279,7 +279,7 @@ select
     select
       net.http_post(
           url:='https://project-ref.supabase.co/functions/v1/function-name',
-          headers:=jsonb_build_object('Content-Type','application/json', 'Authorization', 'Bearer ' || 'YOUR_ANON_KEY'),
```
kallebysantos, line 282: I suggest to replace it by `apikey` header instead and using `PUBLISHABLE_KEY`

fadymak: @kallebysantos the publishable key won't be verified, right? So including it wouldn't actually do anything

kallebysantos: Yes its don't get verified, its more for convention pattern.

ChrisChinchilla (approved):

PR #45061, opened 2026-04-20
fix(studio): validate auth conn percentage input between 1 and 100

awaseem:
## Summary

- Closes [FE-3039](https://linear.app/supabase/issue/FE-3039/validate-conn-percentage-input-between-1-and-100).
- Adds a `superRefine`d `DatabaseFormSchema` in `PerformanceSettingsForm.tsx` that blocks submission when `DB_MAX_POOL_SIZE_UNIT === 'percent'` and the value falls outside 1-100. Connections (fixed pool size) path keeps the existing `min(1)` behavior per the ticket.
- Updates the number input's `min`/`max` attributes to match: percent is now `1`-`100` (was `3`-`80`); connections unchanged (`3`-`Math.floor(maxConnectionLimit * 0.8)`).

## Test plan

- [x] Unit tests added in `PerformanceSettingsForm.test.ts` covering bounds (1, 100), out-of-range (0, 101, -5, 150), mid-range, string coercion, connections path (no upper zod bound), and enum rejection. All 12 pass.
- [x] Manually verify in Auth → Performance settings that entering 0 or 101 in percent mode shows a validation error and blocks save, and that switching to absolute connections still uses the existing bounds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined database pool-size validation and input constraints; percentage unit enforces values between 1–100 and rejects zero, negative, and out-of-range entries. String inputs that represent numbers are coerced where appropriate; unknown unit values are rejected.
* **Tests**
  * Added comprehensive tests for boundary, invalid, string, and unknown-unit cases and for validation error messages.
* **Refactor**
  * Validation moved to a centralized schema and HTML min/max input attributes were removed.

### DIFF
--- a/apps/studio/components/interfaces/Auth/PerformanceSettingsForm.tsx
+++ b/apps/studio/components/interfaces/Auth/PerformanceSettingsForm.tsx
@@ -387,12 +387,6 @@ export const PerformanceSettingsForm = () => {
                               <FormInputGroupInput
                                 type="number"
                                 {...field}
-                                min={chosenUnit === 'percent' ? 1 : 3}
-                                max={
-                                  chosenUnit === 'percent'
-                                    ? 100
-                                    : Math.floor(maxConnectionLimit * 0.8)
-                                }
                                 disabled={!canUpdateConfig || promptUpgrade}
                               />
                               <InputGroupAddon align="inline-end">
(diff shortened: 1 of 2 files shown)


monicakh (approved):

apps/studio/components/interfaces/Auth/PerformanceSettingsForm.tsx
```diff
@@ -375,10 +387,10 @@ export const PerformanceSettingsForm = () => {
                               <FormInputGroupInput
                                 type="number"
                                 {...field}
-                                min={3}
+                                min={chosenUnit === 'percent' ? 1 : 3}
```
ivasilov: `min` and `max` props don't use the form schema. They're propagated to the input and a native tooltip is used (see screenshot). I'd suggest removing them so that they're validated by the schema object (we can control the messaging that way).

Image

ivasilov (approved):

PR #45064, opened 2026-04-20
fix: select fields aren't accessible with tab on Safari

djhi:
## Problem

`select` fields aren't accessible with tab on Safari.
This is a known issue on Safari that can be solved by users if they adjust their settings:
image

## Solution

We can simply add a `tabindex` attribute on the Shadcn `<SelectTrigger>` to fix it.

## How to test

- Create a new org. Tab on the form. It should highlight each field, including the _Type_ and _Plan_ select.
- Edit your _Account preferences_. You should be able to tab to the _Primary email_ field

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced keyboard navigation support for select components by improving focus handling.

### DIFF
--- a/packages/ui/src/components/shadcn/ui/select.tsx
+++ b/packages/ui/src/components/shadcn/ui/select.tsx
@@ -54,6 +54,7 @@ const SelectTrigger = React.forwardRef<
       SelectTriggerVariants({ size }),
       className
     )}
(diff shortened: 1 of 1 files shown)


charislam (approved):

dnywh (approved):
Tested on staging on Safari.

PR #45067, opened 2026-04-20
feat(telemetry): add exposure event for default privileges checkbox

seanoliver:
A stab at implementing the telemetry gaps flagged in the review on #45034. Merges into that branch.


alaister (approved):
LGTM!

PR #45076, opened 2026-04-20
fix: csv seq sync

### LINKED ISSUE #45073: CSV import does not advance owned sequences for sequence backed primary keys
# Bug report

## Describe the bug

Importing rows through the Studio CSV import flow does not advance the sequence for sequence-backed primary key columns when the CSV includes explicit `id` values.

After the import completes, the table contains the imported rows, but the backing sequence still points to its old value. The next insert that relies on the default primary key value can then generate an ID that is already below the table's current maximum ID, which leads to incorrect sequence state and can cause unique constraint violations on subsequent inserts.

This was reproduced with both identity-backed and sequence-default primary key columns.

## To Reproduce

1. Create a table with a sequence-backed primary key.

```sql
create sequence public.ticket_entries_id_seq;

create table public.ticket_entries (
  id bigint primary key default nextval('public.ticket_entries_id_seq'),
  title text
);

alter sequence public.ticket_entries_id_seq owned by public.ticket_entries.id;
```

2. Prepare a CSV file with explicit IDs:

```csv
id,title
341,First row
342,Second row
343,Third row
```

3. In Studio, open the table in Table Editor.
4. Use `Import data from CSV` and upload the file.
5. Wait for the import to complete successfully.
6. Check the imported data and the sequence state:

```sql
select max(id) from public.ticket_entries;
select last_value, is_called from public.ticket_entries_id_seq;
```

7. Insert a new row through the dashboard without manually setting `id`, or check the next generated value:

```sql
select nextval(pg_get_serial_sequence('public.ticket_entries', 'id'));
```

## Expected behavior

After the CSV import completes, the owned sequence should be aligned with the current maximum ID in the table.

For the example above:
- `max(id)` should be `343`
- the sequence should also be at `343`
- the next generated value should be `344`

7ttp:
## TL;DR
fixes csv imports with explicit id values leaving auto generated primary keys out of sync 
by resolving the primary key sequence correctly after import

## ex:

| Before | After |
| --- | --- |
| Before: stale sequence after CSV import | After: sequence synced after CSV import |

## ref:
- closes https://github.com/supabase/supabase/issues/45073
- also closed https://github.com/supabase/supabase/issues/43993

## Release Notes

* **Bug Fixes**
  * Fixed identity sequence synchronization during CSV imports. After spreadsheet data is inserted, the system now properly updates sequence values, ensuring subsequent rows receive correct identifiers.


awaseem (approved):
Nice and E2E tests WOOOO!

PR #45078, opened 2026-04-21
fix(studio): add resume project flow to project settings

dnywh:

YES

Bug fix. Resolves DEPR-511.

Paused projects in `Project Settings > General > Project availability` still present restart/pause maintenance controls, but no resume affordance. That makes the resume path hard to discover from Settings and pushes users back to the project dashboard to find the correct action.

The paused state also keeps showing a redundant disabled `Pause project` row, and the pause confirmation uses a more flexible modal than this flow needs.

DEPR-519 already covered the unhealthy-project restart guard, but not this paused-project discoverability path.

## What is the new behavior?

Project Settings is now paused-project aware. It shows a shared `Resume project` action when the project can still be restored, falls back to the project dashboard when the restore window has expired or pause status cannot be confirmed, and reuses the same resume flow, permission checks, and free-tier guardrails as the paused dashboard.

While a project is already paused, the redundant `Pause project` row is hidden so the section stays focused on the real next action. For active projects, the pause row remains in place, including the useful disabled tooltip states for plans that cannot pause.

The pause confirmation now uses `AlertDialog` with shorter, more accurate copy about the restore window, and the restart controls now behave more consistently on smaller breakpoints. The Project Settings command-menu entry is also searchable via `resume project`.

| Before | After |
| --- | --- |
| CleanShot 2026-04-24 at 18 05 25@2x  | CleanShot 2026-04-24 at 18 06 12@2x |
| CleanShot 2026-04-24 at 18 08 53@2x | CleanShot 2026-04-24 at 18 08 30@2x |
| CleanShot 2026-04-24 at 17 41 40@2x | CleanShot 2026-04-24 at 18 10 34@2x |

## Summary by CodeRabbit

* **New Features**
  * Added a tooltip-enabled "Resume project" button that handles permissions, free-plan member gating, optional Postgres version selection, and navigates to the project after restore.
* **UX**
  * Pause confirmation migrated to an alert-style dialog with updated copy and disabled controls during pausing.
  * Restart controls updated for improved responsive layout and refreshed button visuals.
  * Project settings now show appropriate resume/dashboard actions based on pause/restore eligibility.
* **Tests**
  * Added tests for active, resumable-paused, and non-resumable-paused states.


kemaldotearth:
Maybe you've already got it on the list, but if a project is paused, we should probably conditionally hide this row or?

Screenshot 2026-04-21 at 17 33 23

apps/studio/components/interfaces/Settings/General/Project.tsx
```diff
@@ -12,17 +12,43 @@ import {
...
+      ? 'Restart project'
+      : 'Restart database'
+
+  const primaryActionDescription = isPaused
+    ? isPauseStatusSuccess && !pauseStatus.can_restore
+      ? 'This project can no longer be resumed here. Open the dashboard to download backups and view recovery options.'
+      : isPauseStatusError
+        ? 'Open the dashboard to manage this paused project.'
+        : 'Bring your paused project back online.'
+    : 'Your project will not be available for a few minutes.'
```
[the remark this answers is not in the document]
dnywh: Fixed

apps/studio/components/layouts/ProjectLayout/PausedState/ProjectPausedState.tsx
joshenlim: unrelated - but can use the UpgradePlanButton here, will handle if the user doesn't have billing perms
i'll hop on

apps/studio/components/layouts/ProjectLayout/PausedState/ProjectPausedState.tsx
```diff
@@ -81,64 +53,6 @@ export const ProjectPausedState = ({ product }: ProjectPausedStateProps) => {
...
-  })
-
-  const form = useForm<z.infer<typeof FormSchema>>({
-    resolver: zodResolver(FormSchema),
-    mode: 'onChange',
-    defaultValues: { postgresVersionSelection: '' },
-  })
-
   return (
     <>
```
joshenlim: also unrelated - unnecessary fragment now that there's no longer a modal, will remove

apps/studio/components/interfaces/Settings/General/Project.tsx
```diff
@@ -86,9 +124,9 @@ export const Project = () => {
                   <div className="flex space-x-4">
                     <BarChart2 strokeWidth={2} />
                     <div>
-                      <p className="text-sm">Project usage statistics have been moved</p>
+                      <p className="text-sm">Project usage statistics have moved</p>
                       <p className="text-foreground-light text-sm">
-                        You may view your project's usage under your organization's settings
+                        Project usage is now viewable within organization settings.
```
joshenlim, line 129: Unrelated but im wondering whether this callout can just be removed now - it's been here for quite a while now cc @dnywh (okay to keep either way)

apps/studio/components/interfaces/Settings/General/Infrastructure/PauseProjectButton.tsx
```diff
@@ -91,21 +100,23 @@ const PauseProjectButton = () => {
...
-          it.
-        </p>
-      </ConfirmationModal>
+      <AlertDialog open={isModalOpen} onOpenChange={setIsModalOpen}>
+        <AlertDialogContent>
+          <AlertDialogHeader>
+            <AlertDialogTitle>Pause project?</AlertDialogTitle>
+            <AlertDialogDescription>
+              This project will be unavailable while paused. Your data stays safe, and you can
+              resume it later from the dashboard while it remains within the restore window.
```
joshenlim, line 109: @dnywh im just wondering whether to elaborate on "restore window"
can imagine users might be thinking what that is (e.g how long, what happens after)

maybe re-iterating the paused project state here might be good enough
e.g pause window - 90 days
you have up till (when) to unpause or you can't (but data still available for downloads)

joshenlim (approved):
Generally LGTM! Verified locally - just 2 comments i need to check with

apps/studio/components/interfaces/Settings/General/Project.tsx
dnywh: Agree. Let’s move it out. Can always revisit if it confuses folks.

PR #45083, opened 2026-04-21
fix: switch fields aren't accessible with tab on Safari

djhi:
## Problem

Similar to #45064, switch fields aren't accessible with tab on Safari.

Example:
- Open https://studio-staging-git-gildasgarcia-fe-2998-suggest-e7fb9e-supabase.vercel.app
- Create a new project
- Move through fields with tab, note that you can't focus _High Availability_

## Solution

- Add a `tabindex="0"` on `<Switch>`
- Add a `tabindex="0"` on `<Toggle>` (was a mistake as we don't actually use it yet but it will be needed when we do)

## How to test

- Open the staging
- Create a new project
- Move through fields with tab, note that you can focus _High Availability_

## Summary by CodeRabbit

* **Bug Fixes**
  * Switch and Toggle controls now correctly participate in keyboard tab order, improving focusability and keyboard navigation for accessibility.

### DIFF
--- a/packages/ui/src/components/shadcn/ui/switch.tsx
+++ b/packages/ui/src/components/shadcn/ui/switch.tsx
@@ -52,6 +52,7 @@ const Switch = React.forwardRef<React.ElementRef<typeof SwitchPrimitives.Root>, 
       className={cn(switchRootVariants({ size }), className)}
       {...props}
       ref={ref}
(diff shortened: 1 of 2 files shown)


dnywh (approved):
Tested on staging, on Safari.

PR #45099, opened 2026-04-21
chore(studio): migrate useHotKey to useShortcut

awaseem:
## Summary
- Migrates all 11 `useHotKey` call sites across 9 files to `useShortcut`, backed by `SHORTCUT_DEFINITIONS` in `state/shortcuts/registry.ts`.
- Adds 10 new registry entries (all `showInSettings: false` to keep behavior identical to today — these were not previously user-configurable).
- Deletes `apps/studio/hooks/ui/useHotKey.ts`.
- Simplifies `ActionBar.handleSave` — the legacy hook passed a `KeyboardEvent` the callback used for `preventDefault`/`stopPropagation` and a textarea-plain-Enter guard; all of that is redundant under `useShortcut` (TanStack handles default/propagation; `Mod+Enter` never fires on plain Enter).
- Removes a stale commented-out `useHotKey` reference in `DataTableFilterCommand.tsx`.

Part of FE-3025 (legacy hotkey hook cleanup). `useKeyboardShortcuts` in `grid/components/common/Hooks.tsx` will be migrated in a follow-up.

## Test plan

All shortcuts should still fire with **Cmd** (macOS) / **Ctrl** (Win/Linux).

**Table Editor — operation queue** (requires pending unsaved edits on a row)
- [x] `Cmd+S` saves pending edits
- [x] `Cmd+.` toggles the operation queue side panel
- [x] `Cmd+Z` undoes the latest edit and re-fetches the affected table rows
- [x] With no pending edits, none of the above fire (gated by `isEnabled`)

**Table Editor — side panel editor forms** (row, table, column, policy, etc.)
- [x] `Cmd+Enter` submits the form when the panel is visible
- [x] Does not submit if the form is disabled/loading or the panel is hidden

**Unified Logs — data table**
- [x] `Cmd+B` toggles the filter controls sidebar (desktop)
- [x] `Cmd+B` opens the filter drawer (mobile, `<sm` breakpoint)
- [x] `Cmd+Esc` resets active column filters (reset button visible)
- [x] `Cmd+U` resets column order + visibility
- [x] `Cmd+J` toggles live mode

**Unified Logs — reset focus**
- [x] `Cmd+.` blurs the currently focused element / resets focus to body

**AI Assistant panel**
- [x] While editing a message, `Cmd+Esc` cancels the edit

**Regression checks**
- [x] `pnpm --filter=studio typecheck` passes (verified locally)
[...]

PR #45100, opened 2026-04-21
chore(studio): migrate useKeyboardShortcuts to useShortcut

awaseem:
## Summary
- Follow-up to #45099 — migrates the final legacy hotkey hook (`useKeyboardShortcuts` in `apps/studio/components/grid/components/common/Hooks.tsx`) to `useShortcut`, backed by `SHORTCUT_DEFINITIONS`.
- Adds 4 registry entries (`TABLE_EDITOR_JUMP_FIRST_ROW` / `_LAST_ROW` / `_FIRST_COL` / `_LAST_COL`), all `showInSettings: false` so they stay non-configurable (same as today).
- Adds `ignoreInputs` to `ShortcutOptions` and threads it through `useShortcut` → `useHotkeySequence`. Replaces the legacy `whitelistNodes: ['INPUT', 'TEXTAREA', 'SELECT']` with the built-in TanStack option, which also covers `contenteditable`.
- Drops the `navigator.appVersion` macOS/Windows detection in `Shortcuts.tsx` (`Mod` resolves this automatically) and the manual `event.stopPropagation()` calls (TanStack's default is `stopPropagation: true`).
- Deletes `Hooks.tsx` — no remaining consumers.

Closes FE-3049.

## Why `ignoreInputs: true` on these entries
`Mod+Arrow*` is a Ctrl/Meta combo. TanStack's default for Ctrl/Meta combos is `ignoreInputs: false`, which would fire the shortcut even when focus is in a text input — not what we want. Setting `ignoreInputs: true` in the registry preserves the legacy whitelist behavior.

## Test plan

All shortcuts should still fire with **Cmd** (macOS) / **Ctrl** (Win/Linux). Test surface: **Table Editor** → open any table with rows and multiple columns.

**Grid navigation (with a cell selected — click a cell first)**
- [x] `Cmd+ArrowUp` jumps selection to row 0 in the same column
- [x] `Cmd+ArrowDown` jumps selection to the last row in the same column
- [x] `Cmd+ArrowLeft` jumps selection to the first non-frozen column in the same row
- [x] `Cmd+ArrowRight` jumps selection to the last column in the same row (not the trailing "add column" cell)

**Whitelist regression (most important — the `ignoreInputs` check)**
- [x] Open the **Filter** popover on a table → put focus in the filter value `<input>` → type with arrow keys / use `Cmd+ArrowLeft/Right` for word-jump → cursor moves in the input, grid selection does **not** jump
- [x] Same for a `<textarea>` (e.g. row editor → a text column with a textarea) — `Cmd+ArrowUp/Down` should not hijack focus away from the textarea
- [x] Same for a `<select>` (e.g. row editor → enum / boolean column) — arrow keys behave normally
- [x] Editing a cell value in-place (double-click a cell to enter edit mode) — arrow keys navigate the text, don't jump the grid

**Platform check**
- [x] On macOS, the shortcuts fire with `Cmd` (not `Ctrl`)
- [x] On Windows/Linux, they fire with `Ctrl` (not `Cmd`) — `Mod` resolution
- [x] None of the new entries appear in Account → Preferences → Keyboard shortcuts (all `showInSettings: false`)

**Regression checks**
- [x] `pnpm --filter=studio typecheck` passes (verified locally)
- [x] Existing shortcuts from #45099 (`Cmd+K`, `Cmd+I`, `Cmd+S` in op queue, results copy, etc.) still work unchanged

## Summary by CodeRabbit

* **New Features**
  * Added four keyboard shortcuts for table navigation: jump to first/last row and first/last column using Mod+Arrow keys.
  * Added a shortcut option to prevent triggers when focus is in input fields.

* **Bug Fixes**
  * Improved keyboard shortcut handling for more consistent behavior across platforms.


apps/studio/state/shortcuts/registry.ts
```diff
@@ -79,4 +83,32 @@ export const SHORTCUT_DEFINITIONS: Record<ShortcutId, ShortcutDefinition> = {
...
+    sequence: ['Mod+ArrowLeft'],
+    showInSettings: false,
+    options: { ignoreInputs: true },
+  },
+  [SHORTCUT_IDS.TABLE_EDITOR_JUMP_LAST_COL]: {
+    id: SHORTCUT_IDS.TABLE_EDITOR_JUMP_LAST_COL,
+    label: 'Jump to last column',
+    sequence: ['Mod+ArrowRight'],
+    showInSettings: false,
+    options: { ignoreInputs: true },
```
joshenlim, line 122: just wondering- given that `ignoreInputs` is true by default per tanstack hotkey, do we need to declare them as true for these then? (e.g we only need to declare if we want it to be specifically `false`)

joshenlim (approved):
LGTM - verified on preview

Just a nit comment [here](https://github.com/supabase/supabase/pull/45100/changes#r3122355715)

apps/studio/state/shortcuts/registry.ts
awaseem: We need them due to the mod + key API as per code comment

```
  /**
   * When `true`, suppresses the shortcut while focus is inside an input-like
   * element (text input, textarea, select, contenteditable). Button-type
   * inputs (type=button/submit/reset) are not ignored.
   *
   * Defaults to `undefined`, which falls through to TanStack's per-hotkey
   * default: `true` for single keys and Shift/Alt combos, `false` for
   * Ctrl/Meta/Mod shortcuts and Escape. Set explicitly to override.
   */
```

PR #45114, opened 2026-04-22
Fix cron job form does not focus invalid inputs

djhi:
## Problem

When trying to create a new CRON job with an already used name, the `name` input isn't focused. As a result, users don't see the error.

## Solution

Fix the call to `form.setError` so that it focuses the input

## Release Notes

* **Bug Fixes**
  * Improved error handling in cron job creation. When a duplicate job name is detected, the form now automatically focuses on the name field, providing clearer feedback to help you quickly identify and correct validation errors.


FatumaA:
LGTM

awaseem (approved):

PR #45116, opened 2026-04-22
fix: paste seq sync

### LINKED ISSUE #45113: Paste text import does not advance owned sequences for sequence-backed primary keys
# Bug report

## Describe the bug

Importing rows through the Studio `Import data from CSV` flow using the `Paste text` tab does not advance the owned sequence for sequence-backed primary key columns when the pasted CSV includes explicit `id` values.

After the import completes, the table contains the pasted rows, but the backing sequence still points to its old value. The next insert that relies on the default primary key value can then generate an ID that is already below the table's current maximum ID, which leaves the sequence in an incorrect state and can cause unique constraint violations on subsequent inserts.

This was reproducible with a primary key backed by a custom owned sequence using `nextval(...)`.

## To Reproduce

1. Create a table with a sequence-backed primary key.

```sql
create sequence public.ticket_entries_id_seq;

create table public.ticket_entries (
  id bigint primary key default nextval('public.ticket_entries_id_seq'),
  title text
);

alter sequence public.ticket_entries_id_seq owned by public.ticket_entries.id;
```

2. Prepare CSV text with explicit IDs:

```csv
id,title
341,First row
342,Second row
343,Third row
```

3. In Studio, open the table in Table Editor.
4. Click `Import data from CSV`.
5. Switch to the `Paste text` tab.
6. Paste the CSV text above and complete the import successfully.
7. Check the imported data and the sequence state:

```sql
select max(id) from public.ticket_entries;
select last_value, is_called from public.ticket_entries_id_seq;
```

8. Check the next generated value:

```sql
select nextval(pg_get_serial_sequence('public.ticket_entries', 'id'));
```

## Expected behavior

After the pasted import completes, the owned sequence should be aligned with the current maximum ID in the table.

For the example above:
- `max(id)` should be `343`
- the sequence should also be at `343`
- the next generated value should be `344`

## Actual behavior

The pasted rows are imported successfully, but the sequence remains at its previous value instead of being advanced to match the imported IDs.

For the example above:
- `max(id)` is `343`
- `last_value` remains at `1`
- `is_called` is `true`
- the next generated value is `2` instead of continuing at `344`

## Additional context

This appears to affect the `Paste text` import path specifically.

7ttp:
## TL;DR
- extends https://github.com/supabase/supabase/pull/45076
- closes https://github.com/supabase/supabase/issues/45113

## Summary by CodeRabbit

* **New Features**
  * Sequence values now properly sync after bulk row inserts, ensuring correct auto-increment behavior for subsequent inserts.

* **Tests**
  * Added end-to-end coverage for CSV import, including a "Paste text" import path that verifies custom-owned sequences are synchronized before the next insert.


e2e/studio/features/table-editor.spec.ts
```diff
@@ -1369,6 +1369,84 @@ testRunner('table editor', () => {
...
+        await query(`alter sequence public.${sequenceName} owned by public.${tableName}.id;`)
+      },
+      async () => {
+        await query(`drop table if exists public.${tableName} cascade;`)
+        await query(`drop sequence if exists public.${sequenceName};`)
+      }
+    )
+
+    await page.goto(toUrl(`/project/${ref}/editor?schema=public`))
+    await waitForTableToLoad(page, ref)
```
djhi:
```suggestion
    const waitForTable = waitForTableToLoad(page, ref)
    await page.goto(toUrl(`/project/${ref}/editor?schema=public`))
    await waitForTable
```

I know it's done as you did elsewhere but I'll fix them soon too as this makes the tests flaky

djhi (changes requested):
Thanks for adding tests 👍

awaseem (approved):
LGTM

djhi (approved):

PR #45163, opened 2026-04-23
docs: integrations and platforms Keys change

ChrisChinchilla:

YES/

## Summary by CodeRabbit

* **Documentation**
  * Updated Vercel Marketplace integration documentation with revised environment variable references. The guide now contains accurate information about the credentials required when connecting Supabase projects through Vercel's marketplace. This update ensures users have the correct configuration details for seamless project integration and proper synchronization of their environment settings.


apps/docs/content/guides/integrations/vercel-marketplace.mdx
```diff
@@ -46,7 +46,7 @@ POSTGRES_USER
 POSTGRES_HOST
 POSTGRES_PASSWORD
 POSTGRES_DATABASE
-SUPABASE_SERVICE_ROLE_KEY
+SUPABASE_SECRET_KEY
```
fadymak, line 49: We still set these env vars on Vercel projects — but perhaps this is the right move to deprioritize using them:

Image

apps/docs/content/guides/integrations/vercel-marketplace.mdx
```diff
@@ -46,7 +46,7 @@ POSTGRES_USER
...
 SUPABASE_PUBLISHABLE_KEY
 SUPABASE_URL
 SUPABASE_JWT_SECRET
```
fadymak, line 52: I believe this would be related to the legacy secrets which we would likely want to deprioritize as well.

ChrisChinchilla:
@fadymak So should we make these changes?

fadymak (approved):
@ChrisChinchilla yes, I think we should — I wanted to call it out to get your take after testing what we actually set in Vercel-connected projects.

ChrisChinchilla:
@fadymak Fair, I did notice a whole bunch of places in a separate task where we (the docs team) need to change our own keys, as probably do other teams… so we may want to coordinate that too, but this is a docs change, so we can merge.

PR #45173, opened 2026-04-23
Remove clone-of-clone restriction in RestoreToNewProject

Crispy1975:
## Summary

- Removes the UI guard in `RestoreToNewProject.tsx` that blocked projects cloned from another cloned project from being restored to a new project
- The API now supports a ConfigCat feature flag (`INDATA-559`) to allow specific customers to clone clones, so the client-side restriction is no longer needed
- Removes the unused `IS_CLONED_PROJECT` constant alongside the block

## Test plan

- [x] Verify a cloned project can now reach the restore-to-new-project UI without hitting the "cannot be restored" admonition
- [x] Verify non-cloned projects are unaffected
- [x] Verify the feature flag at the API level correctly gates access for customers without the flag enabled

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed unnecessary warning message that displayed during database restoration when restoring a cloned or previously restored project to a new project.


ivasilov (approved):

ivasilov (commented):
LGTM.

PR #45174, opened 2026-04-23
fix(studio): preserve ignoreInputs across re-renders in useShortcut

awaseem:
## Summary

Fixes [FE-3060](https://linear.app/supabase/issue/FE-3060/modshiftc-modshiftm-shortcuts-suppressed-when-a-table-editor-cell-is).

`Mod+Shift+C` (Copy as CSV) and `Mod+Shift+M` (Copy as Markdown) — and any other Meta/Ctrl/Escape shortcut registered via `useShortcut` without an explicit `ignoreInputs` — stopped firing when focus landed on a `react-data-grid` cell in the table editor (or any other input/contenteditable).

## Fix

In `apps/studio/state/shortcuts/useShortcut.tsx`, only include `ignoreInputs` in the options object when it's actually set, so TanStack's register-time default sticks across re-renders.

Updated the `ignoreInputs resolution` test to assert that the key is omitted (rather than passed as `undefined`) when neither caller nor registry set it.

## Test plan

- [x] SQL editor results: `Cmd+Shift+C`, `Cmd+Shift+M`, `Cmd+Shift+J`, `Cmd+Shift+D` still fire
- [x] `Mod+ArrowUp` / `Mod+ArrowDown` / `Mod+ArrowLeft` / `Mod+ArrowRight` inside a focused cell editor still blocked (registry sets `ignoreInputs: true`)
- [x] Unit tests: `pnpm vitest run state/shortcuts/useShortcut.test.tsx`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed keyboard shortcut handling to properly respect the hotkey library's default configuration values when custom options are not explicitly specified.

### DIFF
--- a/apps/studio/state/shortcuts/useShortcut.test.tsx
+++ b/apps/studio/state/shortcuts/useShortcut.test.tsx
@@ -118,9 +118,10 @@ describe('useShortcut', () => {
   })
 
   describe('ignoreInputs resolution', () => {
-    it('defaults to undefined when no registry default and no caller override', () => {
+    it('omits the key when no registry default and no caller override (library applies its per-hotkey default)', () => {
       renderHook(() => useShortcut(SHORTCUT_IDS.COMMAND_MENU_OPEN, vi.fn()))
-      expect(getLastHotkeyOptions().ignoreInputs).toBeUndefined()
+      const options = getLastHotkeyOptions()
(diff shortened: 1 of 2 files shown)


joshenlim (approved):
Verified changes on staging preview in SQL Editor!

PR #44555, opened 2026-04-05
fix(studio): apply literal() escaping to remaining queue mutation files

oniani1:
Fixes #44554

Follow-up to #44451 which added `literal()` escaping to 4 queue message files. The remaining 5 files in the same directory still use raw string interpolation.

The create mutation was the biggest gap -- no `literal()` and no `isQueueNameValid` at all. It could also interpolate `undefined` into SQL when partition config is missing.

Applied the same pattern from #44451 to all 5 files: import `literal`, wrap interpolated values. For the metrics query and create mutation, also used `ident()` for table name references.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal SQL query construction for database queue operations to enhance code reliability and maintainability.

PR #45181, opened 2026-04-23
fix(studio): escape SQL identifiers in policy create query

### LINKED ISSUE #45179: Policy create flow executes unescaped SQL identifiers
## Bug

`generateCreatePolicyQuery` in `apps/studio/components/interfaces/Auth/Policies/PolicyEditorPanel/PolicyEditorPanel.utils.ts` builds a `CREATE POLICY` statement with raw double-quote interpolation:

```ts
const querySkeleton = `create policy "${name}" on "${schema}"."${table}" as ${behavior} for ${command} to ${roles}`
```

That SQL is then passed directly to `useExecuteSqlMutation` in `PolicyEditorPanel/index.tsx` (~line 192). A policy name, schema, or table containing a `"` character breaks out of the identifier quoting in the executed statement.

## Reproduction

1. In Supabase Studio, go to Auth → Policies.
2. Click "New policy" on any table, pick a table whose name contains a `"` (same applies to schemas / policy names).
3. Enter a policy name containing a `"`.
4. Save. The executed SQL has a broken identifier and either fails with a parse error or, depending on the payload, executes a different statement than intended.

## Expected

Identifiers are escaped with `ident()` from `@supabase/pg-meta/src/pg-format`, which is the convention already used in the sibling `Policies.utils.ts:319` (`buildPoliciesForPath`) and in the recently merged fixes #44555, #44589, #44721, #44723.

## Scope

The preview-only SQL in `Policies.utils.ts` (`createSQLStatementForCreatePolicy` / `createSQLStatementForUpdatePolicy`) has the same pattern but is not executed (the policy mutation uses a structured payload). Cosmetic in the review modal only; tracking separately.

oniani1:
Closes #45179.

## Summary

`generateCreatePolicyQuery` in `PolicyEditorPanel.utils.ts` builds a `CREATE POLICY` statement with raw double-quote interpolation for name/schema/table, then executes it via `useExecuteSqlMutation` from `PolicyEditorPanel/index.tsx:192`. Any of those values containing a `\"` character breaks out of the identifier quoting in the executed statement.

Applies `ident()` to the three identifier interpolations. Same pattern as #44555 (queue), #44589 (index), #44721 (view autofix), #44723 (auth hooks). The helper is already used at `Policies.utils.ts:319`.

## Scope

The preview-only SQL in `Policies.utils.ts` (`createSQLStatementForCreatePolicy` / `createSQLStatementForUpdatePolicy`) has the same pattern but is not executed. That string is only rendered in the review modal while the mutation uses a structured payload. Tracking separately.

## Test plan

- [ ] Create a policy with a name containing a double quote. Verify it applies correctly with the fix (and produces broken SQL without).
- [ ] Create a policy on a table/schema name containing a double quote. Same check.
- [ ] Regression: create a plain-named policy. Works as before.

### DIFF
--- a/apps/studio/components/interfaces/Auth/Policies/PolicyEditorPanel/PolicyEditorPanel.utils.ts
+++ b/apps/studio/components/interfaces/Auth/Policies/PolicyEditorPanel/PolicyEditorPanel.utils.ts
@@ -1,3 +1,4 @@
+import { ident } from '@supabase/pg-meta/src/pg-format'
 import type { PostgresPolicy } from '@supabase/postgres-meta'
 import { isEqual } from 'lodash'
 
@@ -65,7 +66,7 @@ export const generateCreatePolicyQuery = ({
   using?: string
   check?: string
 }) => {
-  const querySkeleton = `create policy "${name}" on "${schema}"."${table}" as ${behavior} for ${command} to ${roles}`
+  const querySkeleton = `create policy ${ident(name)} on ${ident(schema)}.${ident(table)} as ${behavior} for ${command} to ${roles}`
   const query =
     command === 'insert'
       ? `${querySkeleton} with check (${check});`


charislam (approved):
Thanks @oniani1!

PR #45182, opened 2026-04-23
fix: clear tax ID fields on billing country change

ignaciodob:
### Summary

Mirrors the tax-ID-clearing behavior already present in `NewPaymentMethodElement` (used by `AddPaymentMethodForm`) over to `BillingCustomerDataForm`. When a user changes their billing country, the previously selected tax ID (type, value, name) is now cleared so a stale tax ID from the previous country doesn't silently persist under the new one.

### Test plan

- [x]  Open **Organization Settings → Billing → Billing Address** for an org with a saved billing address + tax ID
- [x]  Verify the saved tax ID is shown on initial load (not cleared)
- [x]  Change the country in the address element to a different country
- [x]  Verify the tax ID combobox resets to "Select tax ID" and the tax ID value input disappears
- [x]  Change the country back to the original country → tax ID still cleared (does not re-populate)
- [x]  Pick a new tax ID for the new country, enter a value, save → confirm it persists
- [x]  Reload the page → saved tax ID shows correctly and is not cleared on mount

## Summary by CodeRabbit

* **Bug Fixes**
  * Tax ID fields in organization billing settings are now automatically cleared when the billing address country is changed, preventing country-specific tax ID mismatches.

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/BillingCustomerData/BillingCustomerDataForm.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/BillingCustomerData/BillingCustomerDataForm.tsx
@@ -5,7 +5,7 @@ import type {
   StripeAddressElementOptions,
 } from '@stripe/stripe-js'
 import { Check, ChevronsUpDown, Info, X } from 'lucide-react'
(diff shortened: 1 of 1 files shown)


awaseem (approved):
LGTM tested in staging

PR #42350, opened 2026-01-31
feat(studio-local): functions management api - test functions

kallebysantos:
## What kind of change does this PR introduce?

Feature

## What is the current behavior?

Functions page on self-hosted differs from Platform

## What is the new behavior?

Adds the possibility to try/test functions in Self-Host version.

<details>

https://github.com/user-attachments/assets/7c7ddd92-42dd-47c6-bedb-f4f21d311af9

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved edge function URL validation so testing works reliably both on-platform and off-platform, including proper URL handling for local setups.

* **UI Improvements**
  * Moved the Test button in the edge functions interface for more consistent layout while preserving its behavior.

* **Tests**
  * Expanded tests to cover platform-aware URL validation scenarios.

PR #45183, opened 2026-04-23
fix(studio): generate edge function URLs that work on self-hosted deployments

### LINKED ISSUE #44907: Cron / Database Hooks edge function URLs hardcoded to cloud format  (${ref}.supabase.localhost) in self-hosted deployments
# Bug report                                                                      
                                                                                    
  ## Describe the bug                                                               
                                                                                    
  When creating a cron job that triggers an edge function (or a database hook       
  targeting an edge function) via the Supabase Studio Dashboard in a self-hosted
  Docker deployment, the UI automatically generates a URL in the format:            
                                          
  https://${projectRef}.supabase.${tld}/functions/v1/${slug}                        
                                          
  For a self-hosted setup where the project's `endpoint` is `localhost:8000`, this  
  produces:                                                                         
                                                                                    
  https://default.supabase.localhost/functions/v1/hello                             
                                         
  This URL **cannot be resolved** from within the Postgres container (where         
  `pg_cron` executes via `pg_net`), because:                                        
  1. `default.supabase.localhost` is not a valid DNS entry in the Docker network
  2. The `supabase.` subdomain pattern is cloud-specific and doesn't exist in       
  self-hosted environments                                                          
  3. HTTPS is used even though the self-hosted deployment runs on HTTP              
                                                                                    
  The cron job appears to succeed (`status = 'succeeded'` in                        
  `cron.job_run_details`), but the actual HTTP request fails with `Couldn't connect 
  to server` in `net._http_response.error_msg`.                                     
                                         
  ## To Reproduce                                                                   
                                          
                                          
  1. Set up self-hosted Supabase via Docker Compose:                                
     ```bash                             
     git clone https://github.com/supabase/supabase                                 
     cd supabase/docker                                                             
     docker compose up -d                                                           
  2. Open Studio Dashboard at http://localhost:8000                                 
  3. Navigate to Integrations → Cron                                                
  4. Click Create a new cron job                                                    
  5. Select Edge Function as the job type                                           
  6. Choose any edge function (e.g., hello)                                         
  7. The "Edge Function" dropdown auto-populates a URL like:                        
  https://default.supabase.localhost/functions/v1/hello                             
  8. Save the cron job                                                              
  9. Check the database:                                                            
  SELECT jobid, command FROM cron.job;                                              
  -- Shows: url:='https://default.supabase.localhost/functions/v1/hello'            
                                                                                    
  SELECT status, error_msg FROM net._http_response ORDER BY id DESC LIMIT 5;        
  -- Shows: error_msg = 'Couldn't connect to server'                                
                                                                                    
  Expected behavior                                                                 
                                                                                    
  In self-hosted deployments, the edge function URL should use the internal Docker  
  service name and HTTP protocol:                                                   
                                                                                    
  http://kong:8000/functions/v1/hello                                               
                                                                                    
  This URL is accessible from the Postgres container via the Docker network and     
  correctly routes to the edge functions service through Kong.                      
                                                                                    
  Actual behavior                                                                   
                                          
  The UI generates a cloud-format URL                                               
  (https://default.supabase.localhost/functions/v1/hello) that:
  - Fails to resolve in the Postgres container's DNS                                
  - Uses HTTPS when only HTTP is available                                          
  - Silently fails — cron.job_run_details shows status = 'succeeded' because the SQL
   statement itself executed, but net._http_response contains the actual connection 
  error                                                                             
                                                                                    
  System information                                                                
                                                                                    
  - OS: Docker on macOS (self-hosted)                                               
  - Version of supabase: Latest (commit 205cbe7 — studio image                      
  supabase/studio:2026.04.08-sha-205cbe7)                                           
  - PostgreSQL version: 15.8 (supabase/postgres:15.8.1.085)
  - pg_cron version: 1.6                                                            
  - pg_net version: present                                                         
                                                                                    
  Additional context                                                                
                                                                                    
  Root Cause                                                                        
                                          
  The URL is hardcoded in the Studio frontend at:                                   
                                         
  File:                                                                             
  apps/studio/components/interfaces/Integrations/CronJobs/EdgeFunctionSection.tsx
  (line 42-46)                                                                      
                                          
  const buildFunctionUrl = (slug: string, projectRef: string, restUrl?: string) => {
    const restUrlTld = restUrl ? new URL(restUrl).hostname.split('.').pop() : 'co'  
    const functionUrl =                                                             
  `https://${projectRef}.supabase.${restUrlTld}/functions/v1/${slug}`               
    return functionUrl                                                              
  }                                                                                 
                                         
  When restUrl is http://localhost:8000, the TLD extraction produces localhost,     
  resulting in the malformed URL         
  https://default.supabase.localhost/functions/v1/hello.                            
                                   
  Same issue affects these files:                                                   
                                          
  File:                                                                             
  apps/studio/components/interfaces/Integrations/CronJobs/EdgeFunctionSection.tsx
  Line: 42-46                                                                       
  Feature: Cron job creation                                                     
  ────────────────────────────────────────                                          
  File: apps/studio/components/interfaces/Database/Hooks/FormContents.tsx           
  Line: ~238                                                                     
  Feature: Database webhook creation                                                
  ────────────────────────────────────────
  File: apps/studio/components/interfaces/Database/Hooks/HTTPRequestConfig.tsx      
  Line: ~120                             
  Feature: Hook HTTP config dropdown                                                
  ────────────────────────────────────────
  File: apps/studio/components/interfaces/Database/Hooks/EditHookPanel.tsx          
  Line: 34-35                            
  Feature: Edge function URL detection                                              
  ────────────────────────────────────────
  File: apps/studio/components/interfaces/Database/Hooks/HooksList/HookList.tsx     
  Line: 58-59                      
  Feature: Hook list icon logic                                                     
  ────────────────────────────────────────
  File: apps/studio/components/interfaces/Functions/TerminalInstructions.tsx        
  Line: ~76                        
  Feature: curl example generation                                                  
  ────────────────────────────────────────
  File: apps/studio/components/ui/AIAssistantPanel/EdgeFunctionRenderer.tsx         
  Line: ~69                        
  Feature: AI assistant URL generation                                              
                                          
  All of these use the cloud URL pattern                                            
  https://${ref}.supabase.${tld}/functions/v1/ without handling self-hosted
  environments.                                                                     
                                          
  Suggested fix                                                                     
                                   
  Detect self-hosted environments (e.g., localhost, host.docker.internal) and use   
  the base URL directly instead of the ${ref}.supabase.${tld} subdomain pattern:
                                                                                    
  const buildFunctionUrl = (slug: string, projectRef: string, restUrl?: string) => {
    if (!restUrl) return `https://${projectRef}.supabase.co/functions/v1/${slug}`   
                                                                                    
    const url = new URL(restUrl)                                                    
    const hostname = url.hostname                                                   
                                                                                    
    // Self-hosted: use base URL directly                                           
    if (hostname === 'localhost' || hostname.includes('host.docker.internal')) {    
      const port = url.port ? `:${url.port}` : ''                                   
      return `${url.protocol}//${hostname}${port}/functions/v1/${slug}`             
    }                                                                               
                                                                                    
    // Cloud: use subdomain pattern                                                 
    const tld = hostname.split('.').pop()                                           
    return `https://${projectRef}.supabase.${tld}/functions/v1/${slug}`             
  }                                                                                 
                                                                                    
  Workaround (for users)                                                            
                                                                                    
  Manually update the cron job's command in the database:                           
                                          
  UPDATE cron.job                                                                   
  SET command = REPLACE(command, 'https://default.supabase.localhost',              
  'http://kong:8000')                                                               
  WHERE jobname = 'your-job-name';                                                  
                                                                                    
  Or create cron jobs via SQL directly:                                             
                                                                                    
  SELECT cron.schedule('my-job', '*/30 * * * * *', $$                               
    SELECT net.http_post(                                                           
      url := 'http://kong:8000/functions/v1/hello',                                 
      headers := jsonb_build_object('Content-Type', 'application/json'),            
      timeout_milliseconds := 5000                                                  
    );                                                                              
  $$);

oniani1:
Closes #44907.

## Summary

Studio generates edge function URLs across 7 sites by picking the TLD off `restUrl`:

```ts
const restUrlTld = restUrl ? new URL(restUrl).hostname.split('.').pop() : 'co'
const functionUrl = `https://${projectRef}.supabase.${restUrlTld}/functions/v1/${slug}`
```

On self-hosted Docker, `restUrl` is typically `http://localhost:8000`, so the TLD becomes `localhost` and the generated URL is `https://default.supabase.localhost/functions/v1/${slug}`. That host does not resolve, and HTTPS is not something the stack can serve. Cron jobs and database webhooks using this URL report `status = 'succeeded'` in `cron.job_run_details` while `net._http_response.error_msg` carries the actual connect failure.

## Change

Adds `apps/studio/lib/edge-functions.ts` with two helpers:

- `buildEdgeFunctionUrl(slug, projectRef, restUrl)` keeps the cloud subdomain pattern only when the hostname matches `${projectRef}.supabase.<tld>` exactly. Anything else (localhost, Docker service names like `kong`, custom domains, hostnames that happen to include "supabase" as a label) uses the `restUrl` origin directly so the generated URL resolves from wherever the call is made (browser, pg_cron / pg_net in the Postgres container, etc).
- `isEdgeFunctionUrl(url, projectRef, restUrl)` matches both the legacy `.functions.supabase.<tld>/` pattern and the current `.supabase.<tld>/functions/` pattern (now anchored with `startsWith` to avoid cross-ref substring hits), and any URL rooted at the project's `restUrl` origin with `/functions/v1/`.

Updates all 7 call sites to use the helpers:
- `Integrations/CronJobs/EdgeFunctionSection.tsx` (cron job URL)
- `Database/Hooks/FormContents.tsx` (default hook URL + edge-function detection)
- `Database/Hooks/HTTPRequestConfig.tsx` (dropdown)
- `Database/Hooks/EditHookPanel.tsx` (removed the locally-exported `isEdgeFunction`, inlined the check)
- `Database/Hooks/HooksList/HookList.tsx` (icon + title)
- `Functions/TerminalInstructions.tsx` (curl example)
- `ui/AIAssistantPanel/EdgeFunctionRenderer.tsx` (deploy panel URL)

## Test plan

- [x] 14 unit tests in `lib/edge-functions.test.ts` covering cloud, alt-TLD, localhost, kong, custom domain, cross-ref substring, hostname containing "supabase" as a non-Supabase label.
- [ ] Cloud regression: create a cron job / webhook against an edge function on a `*.supabase.co` project. URL unchanged.
- [ ] Self-hosted: `docker compose up` the self-hosted stack, create a cron job targeting an edge function, verify the cron command uses `http://localhost:8000/functions/v1/...` (or whatever `restUrl` resolves to) and actually reaches the function.

### DIFF
--- a/apps/studio/lib/edge-functions.ts
+++ b/apps/studio/lib/edge-functions.ts
@@ -1,0 +1,66 @@
+/**
+ * Builds the URL used to invoke a Supabase Edge Function.
+ *
+ * For Supabase Cloud (hostname matches *.supabase.<tld>) the `<ref>.supabase.<tld>`
+ * subdomain pattern is kept. For anything else — including self-hosted setups
+ * where the project's URL is `http://localhost:8000`, `http://kong:8000`, or a
+ * custom domain — the origin from `restUrl` is used directly so the generated
+ * URL resolves in the context it is being invoked from (browser, pg_cron in
+ * the Postgres container, etc.).
+ */
+export const buildEdgeFunctionUrl = (
+  slug: string,
+  projectRef: string | undefined,
+  restUrl?: string
+): string => {
+  const fallback = `https://${projectRef}.supabase.co/functions/v1/${slug}`
+  if (!restUrl) return fallback
+
+  let parsed: URL
+  try {
+    parsed = new URL(restUrl)
+  } catch {
+    return fallback
+  }
+
+  const { hostname, origin } = parsed
+  const tld = hostname.split('.').pop()
+  // Only treat the host as Supabase Cloud when it matches <projectRef>.supabase.<tld>
+  // exactly. This keeps generated URLs correct for self-hosted setups that run
+  // on localhost, Docker service names, or custom domains.
+  if (projectRef && hostname === `${projectRef}.supabase.${tld}`) {
+    return `https://${projectRef}.supabase.${tld}/functions/v1/${slug}`
+  }
+  return `${origin}/functions/v1/${slug}`
+}
+
+/**
+ * Checks whether a persisted webhook URL points at a Supabase Edge Function.
+ *
+ * Matches both the old `.functions.supabase.<tld>` and the current
+ * `.supabase.<tld>/functions/` cloud patterns, and for self-hosted setups
+ * treats any URL rooted at the project's restUrl origin with `/functions/v1/`
+ * as an edge function.
+ */
+export const isEdgeFunctionUrl = (
+  url: string,
+  projectRef: string | undefined,
+  restUrl?: string
+): boolean => {
+  if (!url || !projectRef) return false
+
+  if (
+    url.startsWith(`https://${projectRef}.functions.supabase.`) ||
+    (url.startsWith(`https://${projectRef}.supabase.`) && url.includes('/functions/'))
+  ) {
+    return true
+  }
+
+  if (!restUrl) return false
+  try {
+    const origin = new URL(restUrl).origin
+    return url.startsWith(`${origin}/functions/v1/`)
+  } catch {
+    return false
+  }
+}
--- a/apps/studio/components/interfaces/Database/Hooks/EditHookPanel.tsx
+++ b/apps/studio/components/interfaces/Database/Hooks/EditHookPanel.tsx
@@ -19,21 +19,10 @@ import { useDatabaseHooksQuery } from '@/data/database-triggers/database-trigger
 import { tableEditorQueryOptions } from '@/data/table-editor/table-editor-query'
 import { useSelectedProjectQuery } from '@/hooks/misc/useSelectedProject'
 import { useConfirmOnClose } from '@/hooks/ui/useConfirmOnClose'
+import { isEdgeFunctionUrl } from '@/lib/edge-functions'
 import { uuidv4 } from '@/lib/helpers'
 
 export type HTTPArgument = { id: string; name: string; value: string }
-
-export const isEdgeFunction = ({
-  ref,
-  restUrlTld,
-  url,
-}: {
-  ref?: string
-  restUrlTld?: string
-  url: string
-}) =>
-  url.includes(`https://${ref}.functions.supabase.${restUrlTld}/`) ||
-  url.includes(`https://${ref}.supabase.${restUrlTld}/functions/`)
 
 const FORM_ID = 'edit-hook-panel-form'
 
@@ -140,7 +129,6 @@ export const EditHookPanel = () => {
   const isSubmitting = isCreating || isUpdating || isLoadingTable
 
   const restUrl = project?.restUrl
-  const restUrlTld = restUrl ? new URL(restUrl).hostname.split('.').pop() : 'co'
 
   const form = useForm<WebhookFormValues>({
     resolver: zodResolver(FormSchema),
@@ -149,11 +137,7 @@ export const EditHookPanel = () => {
       table_id: selectedHook?.table_id?.toString() ?? '',
       http_url: selectedHook?.function_args?.[0] ?? '',
       http_method: (selectedHook?.function_args?.[1] as 'GET' | 'POST') ?? 'POST',
-      function_type: isEdgeFunction({
-        ref,
-        restUrlTld,
-        url: selectedHook?.function_args?.[0] ?? '',
-      })
+      function_type: isEdgeFunctionUrl(selectedHook?.function_args?.[0] ?? '', ref, restUrl)
         ? 'supabase_function'
         : 'http_request',
       timeout_ms: Number(selectedHook?.function_args?.[4] ?? 5000),
@@ -185,11 +169,7 @@ export const EditHookPanel = () => {
         table_id: selectedHook?.table_id?.toString() ?? '',
         http_url: selectedHook?.function_args?.[0] ?? '',
         http_method: (selectedHook?.function_args?.[1] as 'GET' | 'POST') ?? 'POST',
-        function_type: isEdgeFunction({
-          ref,
-          restUrlTld,
-          url: selectedHook?.function_args?.[0] ?? '',
-        })
+        function_type: isEdgeFunctionUrl(selectedHook?.function_args?.[0] ?? '', ref, restUrl)
           ? 'supabase_function'
           : 'http_request',
         timeout_ms: Number(selectedHook?.function_args?.[4] ?? 5000),
@@ -198,7 +178,7 @@ export const EditHookPanel = () => {
         httpParameters: parseParameters(selectedHook),
       })
     }
-  }, [visible, selectedHook, ref, restUrlTld, form])
+  }, [visible, selectedHook, ref, restUrl, form])
 
   const queryClient = useQueryClient()
   const onSubmit: SubmitHandler<WebhookFormValues> = async (values) => {
--- a/apps/studio/components/interfaces/Database/Hooks/FormContents.tsx
+++ b/apps/studio/components/interfaces/Database/Hooks/FormContents.tsx
@@ -22,7 +22,6 @@ import {
 } from 'ui'
 import { FormItemLayout } from 'ui-patterns/form/FormItemLayout/FormItemLayout'
 
-import { isEdgeFunction } from './EditHookPanel'
 import { WebhookFormValues } from './EditHookPanel.constants'
 import { AVAILABLE_WEBHOOK_TYPES, HOOK_EVENTS } from './Hooks.constants'
 import { HTTPHeaders } from './HTTPHeaders'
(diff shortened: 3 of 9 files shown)


aantti:
@kallebysantos fyi

kallebysantos:
Hey @oniani1 thx for contributing.
Please, consider looking at this PR https://github.com/supabase/supabase/pull/42350 specially [by theses lines](https://github.com/kallebysantos/supabase/blob/1f91a521f96453536fb0f590298a694dc2fd7133/apps/studio/pages/api/edge-functions/test.ts#L24-L29).

I believe you have some duplicated things that its already defined on [lib/api](https://github.com/kallebysantos/supabase/blob/master/apps/studio/lib/api/edgeFunctions.ts#L3).

apps/studio/lib/edge-functions.ts
kallebysantos, line 1: I think it can be moved to a shared place, to avoid duplication of [edgeFunction.ts](https://github.com/kallebysantos/supabase/blob/master/apps/studio/lib/api/edgeFunctions.ts#L3)

apps/studio/lib/edge-functions.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Builds the URL used to invoke a Supabase Edge Function.
+ *
+ * For Supabase Cloud (hostname matches *.supabase.<tld>) the `<ref>.supabase.<tld>`
+ * subdomain pattern is kept. For anything else — including self-hosted setups
+ * where the project's URL is `http://localhost:8000`, `http://kong:8000`, or a
+ * custom domain — the origin from `restUrl` is used directly so the generated
+ * URL resolves in the context it is being invoked from (browser, pg_cron in
+ * the Postgres container, etc.).
+ */
+export const buildEdgeFunctionUrl = (
```
kallebysantos, line 11: It can be simplified by just checking if `IS_PLATFORM` , similar on how is [been using here](https://github.com/kallebysantos/supabase/blob/1f91a521f96453536fb0f590298a694dc2fd7133/apps/studio/pages/api/edge-functions/test.ts#L24-L29)

oniani1:
Hey @kallebysantos, thanks for the pointer. Went through #42350 and `lib/api/edgeFunctions.ts`. The two files end up doing different jobs once I dug in.

`lib/api/edgeFunctions.ts` runs server-side (used by `pages/api/edge-functions/test.ts`) and reads `process.env.NIMBUS_PROD_PROJECTS_URL`. Its `isValidEdgeFunctionURL(url, isPlatform)` is a security validator deciding whether an arbitrary user-supplied URL is safe to proxy. The self-hosted regex `^https?://[^\s/?#]+/functions/v[0-9]{1}/.*$` accepts any host + `/functions/v1/...` path, which works there because the server is just guarding against URL smuggling.

`lib/edge-functions.ts` runs client-side across 7 React components (Hooks + Functions + Integrations + AI Assistant). `buildEdgeFunctionUrl` constructs the invocation URL for display, and `isEdgeFunctionUrl` decides whether a stored webhook URL points at *this project's* edge function so the hooks UI can render "Supabase Edge Function" vs "HTTP Request". That classification needs the actual project `restUrl` origin, not just an `isPlatform` flag. Without the origin, a user's third-party webhook to `example.com/functions/v1/foo` would get misclassified as an edge function on self-hosted.

So `isValidEdgeFunctionURL` answers "is this safe to proxy?" and `isEdgeFunctionUrl` answers "is this one of *this project's* edge function endpoints?". `buildEdgeFunctionUrl` doesn't overlap with anything in `lib/api/`.

If you'd prefer I fold these into `lib/api/edgeFunctions.ts` with `restUrl` threaded through as a parameter, happy to do that. Just wanted to explain the split before changing things.

apps/studio/lib/edge-functions.ts
oniani1: The `IS_PLATFORM` shortcut doesn't cover the self-hosted URL shapes this helper has to generate. A bare `IS_PLATFORM ? cloud : localhost` branch only works when the self-hosted origin is a known string, but in practice the project's actual origin can be any of:

- `http://localhost:8000` (local dev)
- `http://kong:8000` (docker networking, e.g. pg_cron invoking from inside the Postgres container)
- `https://edge.customer.com` (custom-domain self-hosted)
- `http://192.168.x.x:8000` (LAN)

That's why the helper reads the project's `restUrl` origin (or `settings.app_config.endpoint` in `EdgeFunctionRenderer`) instead of a boolean. The same cloud user can have multiple projects with different subdomains, and the same self-hosted user can point the dashboard at any host. Using the live origin is what makes the generated invoke URL actually resolvable in the context it's shown (browser, terminal copy/paste, cron expression, etc.).

The linked `test.ts` pattern you referenced is a different move: it's rewriting `SUPABASE_PUBLIC_URL -> SUPABASE_URL` server-side so the dashboard's proxy can reach Kong on the docker network. That's a server-to-server concern, not a URL-construction-for-display concern.

apps/studio/lib/edge-functions.ts
oniani1: Answered in the top-level reply above: `lib/api/edgeFunctions.ts` runs server-side and reads `process.env.NIMBUS_PROD_PROJECTS_URL`, and its `isValidEdgeFunctionURL` is a proxy-safety validator. `lib/edge-functions.ts` runs client-side and needs the live `restUrl` origin to classify/build URLs. Happy to fold them together if you still want consolidation after that context.

7ttp (commented):
closing in favour of:  https://github.com/supabase/supabase/pull/47861

> Thanks for ur efforts towards contributing to Supabase!! 💚

PR #45185, opened 2026-04-23
fix(studio): encode state.reason in replication error support URL

### LINKED ISSUE #45184: Replication error support URL drops or corrupts Postgres error messages
## Bug

`ErroredTableDetails.tsx` renders a link to the support form when a replication table enters the `no_retry` state. The link drops `state.reason`, a server-returned Postgres error message, into the URL without encoding:

```tsx
href={`/support?projectRef=${projectRef}&category=dashboard_bug&subject=Database%20replication%20error&error=${state.reason}`}
```

Postgres error messages routinely contain characters that corrupt URLs when unencoded:

- `&` splits into a separate query parameter
- `#` is parsed as a fragment, so everything after is dropped from the URL
- a newline terminates the URL entirely
- `=` changes the value boundary of the param

The support form ends up with a truncated, mis-parsed, or empty `error` value, making it harder to triage a replication failure report.

## Expected

`encodeURIComponent` the error message, same pattern already used for the replication logs URL in `PipelineStatus.tsx:39-42`.

## Scope

One-character (well, one-function-call) fix, no behavior change for well-formed error strings.

oniani1:
Closes #45184.

## Summary

`ErroredTableDetails.tsx` links the `no_retry` branch to the support form with the Postgres error message (`state.reason`) dropped into the `error` query parameter raw:

```tsx
href={`/support?projectRef=${projectRef}&category=dashboard_bug&subject=Database%20replication%20error&error=${state.reason}`}
```

Error strings routinely contain `&`, `#`, `=`, or newlines. Any of these break the URL and leave the support form with a truncated or misattributed value. Wrap with `encodeURIComponent`, matching the pattern already used in `PipelineStatus.tsx:39-42` for the replication-logs URL. Null-coalesce to `''` so the link is still well-formed if `state.reason` is absent.

## Test plan

- [ ] Trigger a replication error whose message contains `&` / `#` / newlines, open the support link, verify the full error text reaches the form.
- [ ] Normal error string. URL is unchanged beyond standard URL encoding.


awaseem (approved):
LGTM

PR #45212, opened 2026-04-24
chore: remove UI form components `_Shadcn_` suffix

djhi:
## Problem

We used to have a `_Shadcn_` suffix for all the shadcn form components because we also had `formik` form components.
This is not needed anymore.

## Solution

- Remove the suffix
- Update all usages


ivasilov (approved):

kemaldotearth (approved):
Legend.

staaldraad (approved):

PR #45239, opened 2026-04-25
fix(studio): resolve auth providers skeleton loaders in self-hosted deployments (#45024)

### LINKED ISSUE #45024: Self-hosted Studio: Auth → Sign In / Providers stuck on skeleton loaders (IS_PLATFORM + static export / platform auth API parity)
### Summary

On **self-hosted** Supabase Studio, **Authentication → Sign In / Providers** (`/project/:ref/auth/providers`) can stay on **infinite skeleton loaders** indefinitely. The UI never receives auth configuration data, so "User Signups", provider toggles, etc. never render.

This is a **self-hosted / Studio packaging** issue, not an end-user misconfiguration.

### What we observed (technical)

1. **`IS_PLATFORM` gating**  
   Auth config is loaded via a hook aligned with the cloud product (`useAuthConfigQuery` / `authKeys.authConfig`). In shipped bundles the query’s `enabled` predicate can include `… && IS_PLATFORM && …`. On self-hosted, `NEXT_PUBLIC_IS_PLATFORM` is **`false`**, so the query **never runs** and TanStack Query v5 stays pending → skeletons forever.

2. **Static export / empty `router.query`**  
   Studio is effectively served as static assets; on first paint **`router.query` may be empty**, so `projectRef` from `useParams()` is **undefined**. Some layouts also normalize with a **`"_"` placeholder**, which then produces requests like `GET …/platform/auth/_/config` instead of the real project ref.

3. **Platform API parity**  
   Cloud Studio calls **`GET /api/platform/auth/:ref/config`**. A typical self-hosted ingress stack (e.g. Kong + Studio) **does not implement** that route; if the request falls through to the Studio SPA, the client gets **HTML** instead of JSON, which also breaks the page.

### Related upstream (similar class of bug)

- #36285 — self-hosted **Edge Functions** home stuck on skeleton loaders (same “never finishes loading” symptom).
- #33557 — intentional **`IS_PLATFORM`** checks for local/self-hosted; the pattern is easy to apply in ways that **disable** queries that are still required for core Auth UI.

### Suggested direction for a proper fix

1. **Auth settings reads** that power the Sign In / Providers page should have a **first-class self-hosted path**: either  
   - relax `enabled` when `projectRef` can be derived from the URL (`/project/<ref>/…`), and/or  
   - stop gating **core** auth config fetch on `IS_PLATFORM` without an alternative.

2. **Stable contract for self-hosted**: document and support how Studio obtains the same information cloud gets from `GET /api/platform/auth/:ref/config` (e.g. direct GoTrue **`/auth/v1/settings`** or an equivalent behind the same origin), so reverse proxies do not return the SPA for that fetch.

3. **Regression test**: e2e or integration test for **self-hosted Studio image** hitting Auth → Providers with a non-cloud env.

### Environment

- Observed with official **`supabase/studio`** images used by the community Helm chart / k8s deployments.
- Repro: open Auth → Sign In / Providers on a self-hosted instance; skeletons persist while other Studio pages work.

Thank you for considering this — happy to help validate a fix on a self-hosted cluster.

harshdhankhar11:
## Summary

This PR fixes an issue where the Auth → Sign In / Providers page in Supabase Studio gets stuck on infinite skeleton loaders when running in self-hosted deployments.

## Problem

In self-hosted Studio, the `Authentication → Sign In / Providers` page (`/project/:ref/auth/providers`) displays infinite skeleton loaders and never renders the auth configuration data (User Signups, provider toggles, etc.). This occurs because:

1. **IS_PLATFORM gating**: The auth config is loaded via the `useAuthConfigQuery` hook, which is gated by `IS_PLATFORM`. In shipped self-hosted bundles, `NEXT_PUBLIC_IS_PLATFORM` is set to `false`, causing the TanStack Query to never execute and remain in a `pending` state indefinitely.

2. **Static export mismatch**: The self-hosted static export does not match the platform auth API, leading to the query being skipped on self-hosted setups.

## Solution

- Removed the `IS_PLATFORM` check from the `useAuthConfigQuery` hook to allow auth config data to be fetched on self-hosted deployments
- Ensured the query properly handles both cloud and self-hosted environments
- Verified that skeleton loaders resolve correctly and auth provider UI renders as expected

## Changes

- Modified `apps/studio/data/auth/auth-config-query.ts` to remove the platform gating condition
- Ensured auth keys and components are properly loaded in self-hosted mode

## Testing

- Tested locally with `pnpm dev:studio` and Docker setup
- Verified Auth → Sign In / Providers page loads correctly without skeleton loaders
- Confirmed user signups and provider toggles render properly

## Related Issue

Closes #45024

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed authentication configuration query to properly execute in self-hosted environments, enabling the Sign In/Providers page without additional configuration.

* **Documentation**
  * Added documentation detailing self-hosted authentication defaults and feature flag override capabilities via the API.

* **Tests**
  * Added comprehensive test coverage for the Sign In/Providers page and authentication components.


harshdhankhar11:
@aantti @sprsquish @djwhitt @neiled I have solved this issue, please merge this PR

PR #45247, opened 2026-04-25
fix: show plan name on empty orgs

ignaciodob:
## Summary

- Show the plan name row in the monthly invoice estimate tooltip even when the organization has no projects, so the section no longer appears headerless next to the tax line.

## Test plan

- [ ]  Open an org with no projects → Subscription plan update dialog → hover the "Monthly invoice estimate" tooltip → verify the plan name + price row is visible.
- [ ]  Repeat on an org with projects → verify plan, Compute row, and project breakdown all still render as before.

## Summary by CodeRabbit

* **Refactor**
  * Improved rendering of the Monthly invoice estimate table in the subscription plan update dialog, with reorganized display logic for plan and project billing rows.


kanishkdudeja (approved):

awaseem (approved):
LGTM

PR #45262, opened 2026-04-27
fix: org slug filter

Ziinc:
missing a filter on org slug in org-level audit logs, causing events to be matched to the first org in the array. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed organization label display in audit logs to correctly show the organization associated with each log entry.


joshenlim (approved):
Approving to unblock, enabling auto-merge

PR #45304, opened 2026-04-28
feat: open oauth apps in new tab to avoid losing studio context

imor:
Currently when a user clicks the **Install integration** button on an OAuth integration like Grafana, they are redirected to the partner website in the same tab in which they clicked the button. This makes them lose context in the Supabase Studio. This PR changes the behaviour such that the partner website will be opened in a new tab.

## Summary by CodeRabbit

* **Bug Fixes**
  * OAuth integration installation now opens the redirect URL in a new browser tab instead of redirecting the current window, allowing users to remain in the application while completing the integration process.


joshenlim (approved):
Changes LGTM!

PR #45310, opened 2026-04-28
docs: document filtering vector search by metadata column

### LINKED ISSUE #896: Doc Question:  How do we filter a vector with another column?
# Improve documentation
I'm having a hard time figuring out how filtering a vector with another column is done and how it works with the js client.

- There's no example with any complex filtering.  
- Additionally i'm not sure if the filtering is post filtering, cross filtering or pre filtering

## Link 

https://supabase.com/docs/guides/ai/vector-indexes/hnsw-indexes

## Describe the improvement

It would be great to have more details on how filtering works and how to use it with the js client.

mandarini:
Adds a worked example for combining a vector similarity search with a filter on another column from the JS client, and a short note explaining how filtering interacts with HNSW indexes (including iterative index scans introduced in `pgvector` 0.8.0). The new `match_documents` variant takes a typed filter parameter so the planner can apply the predicate during the index scan, and a jsonb `@>` variant covers the metadata-column case. Cross-links between `semantic-search`, `vector-columns`, and `vector-indexes/hnsw-indexes` so readers land on the answer regardless of which page they start on.

Closes supabase/supabase-js#896

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for filtering vector and semantic search results by metadata, with JS client examples.
  * Explained how filters affect vector (HNSW) queries and planner choice between index and sequential scans.
  * Clarified that selective filters can reduce returned matches and documented pgvector 0.8.0 controls for iterative/index scan behavior.

### DIFF
--- a/apps/docs/content/guides/ai/semantic-search.mdx
+++ b/apps/docs/content/guides/ai/semantic-search.mdx
@@ -152,7 +152,7 @@ In this scenario, you'll likely use a Postgres client library to establish a dir
 
 ### Filtering vector search by metadata
 
-In real applications you usually want to combine the similarity search with a filter on another column, for example only matching documents in a given category, owned by a specific user, or carrying matching metadata in a `jsonb` column. The recommended pattern is to push the filter into the SQL function so the planner can combine it with the vector predicate, rather than chaining `.eq()` after `rpc()`, which filters rows already returned to the client.
+In real applications you usually want to combine the similarity search with a filter on another column, for example only matching documents in a given category, owned by a specific user, or carrying matching metadata in a `jsonb` column. The recommended pattern is to push the filter into the SQL function so the planner can combine it with the vector predicate. Chaining `.eq()` after `rpc()` is applied by PostgREST as an outer filter on the function's result, _after_ the function has already executed its similarity ranking and `limit` — so the vector planner can't use it, and selective filters can leave you with fewer than `match_count` rows.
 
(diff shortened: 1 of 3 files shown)


apps/docs/content/guides/ai/semantic-search.mdx
```diff
@@ -150,6 +150,56 @@ from match_documents(
 
 In this scenario, you'll likely use a Postgres client library to establish a direct connection from your application to the database. It's best practice to parameterize your arguments before executing the query.
 
+### Filtering vector search by metadata
+
+In real applications you usually want to combine the similarity search with a filter on another column, for example only matching documents in a given category, owned by a specific user, or carrying matching metadata in a `jsonb` column. The recommended pattern is to push the filter into the SQL function so the planner can combine it with the vector predicate, rather than chaining `.eq()` after `rpc()` (which filters rows already returned to the client).
```
jeremenichelli: I would remove the parenthesis.

```suggestion
In real applications you usually want to combine the similarity search with a filter on another column, for example only matching documents in a given category, owned by a specific user, or carrying matching metadata in a `jsonb` column. The recommended pattern is to push the filter into the SQL function so the planner can combine it with the vector predicate, rather than chaining `.eq()` after `rpc()`, which filters rows already returned to the client.
```

jeremenichelli (approved):

PR #45325, opened 2026-04-28
[FE-3081] fix(studio): remove false "schema not exposed" warning in Realtime policies editor

alaister:
The Realtime policies editor was showing a warning banner on `realtime.messages` saying the schema isn't exposed through PostgREST. This is incorrect — the `realtime` schema is intentionally excluded from PostgREST (it's in `INTERNAL_SCHEMAS` and filtered out of the exposed schema picker), so the warning is always false in this context.

**Changed:**
- Removed `useProjectPostgrestConfigQuery` from `RealtimePolicies` — it was only used to derive `exposedSchemas`
- Hardcode `exposedSchemas` as `['realtime']` since this editor is for Realtime auth, not PostgREST access

## To test

- Go to the Realtime policies editor (`/project/_/realtime/policies`)
- Confirm the yellow "schema not exposed" warning banner no longer appears on `realtime.messages`
- Confirm policy rows still render correctly and other admonitions (e.g. publicly-readable if RLS is off) still show as expected

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of Realtime Policies schema handling by simplifying configuration logic to consistently use the realtime schema.


awaseem (approved):
LGTM

PR #45341, opened 2026-04-28
fix(growth): preserve untouched services in Privacy Settings save

seanoliver:

YES

Bug fix.

The Privacy Settings modal was submitting only the services the user toggled, not the full non-essential set the SDK expects. `serviceConsentMap` started empty and was only filled in via `handleServicesChange` on toggle, so untouched services were silently dropped from the `updateServices()` payload on Confirm.

There's also an SDK-init race: if the user opens the modal while `categories === null` (the "Unable to Load Privacy Settings" admonition state), toggles render asynchronously when categories arrive — but the map was only seeded at mount and on the open-button click. It stays empty, and the same partial-submit fires on Confirm.

GROWTH-790.

## What is the new behavior?

`serviceConsentMap` is seeded from the current non-essential consent set in three places: the `useState` initializer, the button `onClick`, and a `useEffect` keyed on `[isOpen, categories]` for the null-to-populated transition. The effect uses an empty-map guard so in-progress user toggles aren't clobbered if `categories` re-references mid-session for any other reason.

Confirm always sends the full non-essential decision set.

## Additional context

**Scope.** Only affects the Privacy Settings modal. The banner's Accept/Opt out buttons and Studio's Analytics toggle use `acceptAll`/`denyAll` (different SDK methods that already submit complete state) — not touched here.

**This probably doesn't resolve the active customer-reported re-prompts.** The Usercentrics ruleset currently exposes a single toggleable non-essential category, so for that shape, pre-fix and post-fix code emit the same payload when the user flips it. Shipping as SDK-contract correctness and defense against future ruleset changes that introduce additional non-essential categories.

**Companion PR:** #45350 — independent read-side `isHidden` filter fix.

**Testing.**

- `pnpm -F ui-patterns test --run` — 185/185 pass (two new PrivacySettings cases: the already-loaded path and the SDK-init race).
- Typecheck across `ui-patterns`, `docs`, `studio` — pass.
- Browser QA on Vercel preview covered the already-loaded path (staging short-circuit temporarily bypassed): toggled "Analytics and Marketing" off, confirmed; `localStorage.uc_settings` retained all 21 services with correct statuses. Plus four GDPR scenarios via Sweden VPN — fresh load, accept + reload, opt out + reload, Privacy Settings + reload — all healthy.
- The SDK-init race path is covered by the new unit test only; not separately exercised in-browser because the timing window is hard to hit deterministically.

GROWTH-790

### DIFF
--- a/packages/ui-patterns/src/PrivacySettings/index.test.tsx
+++ b/packages/ui-patterns/src/PrivacySettings/index.test.tsx
@@ -1,0 +1,181 @@
+// @vitest-environment jsdom
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import React from 'react'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { PrivacySettings } from './index'
+
+const updateServices = vi.fn()
+
+const categories = [
+  {
+    slug: 'essential',
+    label: 'Essential',
+    description: 'Required for site operation',
+    isEssential: true,
+    services: [{ id: 'essential-1', consent: { status: true } }],
+  },
+  {
+    slug: 'analytics',
+    label: 'Analytics',
+    description: 'Measure product usage',
+    isEssential: false,
+    services: [{ id: 'analytics-1', consent: { status: true } }],
+  },
+  {
+    slug: 'marketing',
+    label: 'Marketing',
+    description: 'Track campaign performance',
+    isEssential: false,
+    services: [{ id: 'marketing-1', consent: { status: true } }],
+  },
+]
+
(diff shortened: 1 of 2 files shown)


seanoliver:
Closing without merging. After review: neither fix addresses the active customer-reported re-prompt bug. The suspected cause is cross-app fast navigation, which is structurally untestable on Vercel preview because each app is served from its own domain. Both PRs were defense-in-depth correctness fixes for paths not reachable in the current ruleset (one toggleable non-essential category, no hidden non-essentials). Code preserved in branch history; revive if the ruleset changes to expose either bug, or alongside an actual fix for the cross-app race.

PR #45345, opened 2026-04-28
chore(studio): use new publish_dashboard column for marketplace db query

alexhall:
Updates the marketplace DB query to use the new `publish_dashboard` column that was added in https://github.com/supabase/marketplace/pull/49

## Summary by CodeRabbit

* **Chores**
  * Updated internal marketplace integration query and database type definitions to streamline listing management logic.


joshenlim (approved):
Verified changes on preview - can view integrations

PR #45348, opened 2026-04-28
docs(studio): scope studio-ui-patterns skill description to apps/studio

### LINKED ISSUE #45347: Docs: Scope studio-ui-patterns skill description to the Studio codebase to avoid false-positive triggering
# Improve documentation

## Link

https://github.com/supabase/supabase/blob/master/.claude/skills/studio-ui-patterns/SKILL.md

## Describe the problem

The `description` field of the `studio-ui-patterns` skill is too broad to
trigger reliably. Current text:

> Design system UI patterns for Supabase Studio. Use when building or updating
> pages, forms, tables, charts, empty states, navigation, cards, alerts, or
> side panels (sheets). Covers layout selection, component choice, and
> placement conventions.

Skill descriptions are matched against user requests to decide whether the
skill should be loaded. Because this description lists generic UI nouns
("pages, forms, tables, charts, ...") without scoping them to the Studio
codebase, it matches almost any frontend request — including work in
unrelated repos or apps that happen to be open in the same agent session.

This causes two problems:

1. **False positives.** A request like "help me build a settings form" in a
   non-Studio project triggers the skill and pushes Studio-specific
   conventions (`_Shadcn_` imports, `FormItemLayout`, `PageContainer`, etc.)
   onto code that should not follow them.
2. **No negative signal.** The description never tells the agent when *not*
   to apply the skill, so once it triggers, the agent has no basis to back
   off even if the surrounding code clearly isn't Studio.

The body of the SKILL.md is correctly scoped (it references
`apps/design-system/...` paths and Studio-specific components), but that
context is only read *after* the skill is loaded, which is too late to
prevent the wrong trigger.

## Describe the improvement

Tighten the `description` so the trigger criteria are explicit about the
codebase scope, and add a negative clause so the agent knows when to skip
the skill.

Before:

```yaml
description: Design system UI patterns for Supabase Studio. Use when building or updating
  pages, forms, tables, charts, empty states, navigation, cards, alerts, or side panels
  (sheets). Covers layout selection, component choice, and placement conventions.
```

After (suggested):

```yaml
description: Design system UI patterns for the Supabase Studio dashboard codebase
  (apps/studio). Use when building or updating UI inside apps/studio — including
  pages, forms, tables, charts, empty states, navigation, cards, alerts, or side
  panels (sheets) — that should follow the Studio design system. Do not use for
  external Supabase consumer apps, marketing site work, or unrelated frontend
  projects.
```

Key changes:

- Adds the concrete path signal (`apps/studio`) so the matcher has something
  unambiguous to anchor on.
- Names the actual constraint ("should follow the Studio design system")
  rather than just listing UI nouns.
- Adds an explicit negative clause covering the most likely false-positive
  cases (consumer apps, marketing site, unrelated work).

## Additional context

This pattern — scoping skill descriptions by codebase path and adding a
negative clause — is recommended in Anthropic's skill authoring guidance,
and is consistent with how other skills in the Studio repo are scoped (e.g.
skills that reference specific app directories in their description).

The body of the skill itself does not need to change for this issue; only
the frontmatter `description` field.

I'm happy to open a PR with the wording above (or whatever phrasing the
team prefers) if this is something you'd like to fix.

lactofreee:

Documentation improvement.

The `description` field of the `studio-ui-patterns` skill is too broad to
trigger reliably. It lists generic UI nouns (pages, forms, tables, charts, ...)
without scoping them to the Studio codebase, so it matches frontend requests
in unrelated repos and pushes Studio-specific conventions onto code that
should not follow them.

See #45347 for full context.

## What is the new behavior?

Tightens the description to:

- Anchor on the concrete codebase path (`apps/studio`) so the matcher has an
  unambiguous scope signal.
- State the actual constraint ("should follow the Studio design system")
  rather than just listing UI nouns.
- Add a negative clause covering the most likely false-positive cases
  (consumer apps, marketing site, unrelated frontend projects).

The body of the skill is unchanged; only the frontmatter `description`
field is updated.

Closes #45347

## Summary by CodeRabbit

* **Documentation**
  * Updated skill documentation to clarify scope and usage guidelines for the Supabase Studio dashboard codebase, with explicit boundaries for external applications.


lactofreee:
@coderabbitai Thanks for the suggestion. I considered using a folded
scalar but kept the existing multi-line plain scalar format to stay
consistent with the file's original frontmatter style — only the
description content was changed in this PR. Happy to switch to `>` if
the maintainers prefer that format across skills.

PR #45350, opened 2026-04-29
fix(growth): exclude hidden services from consent coverage check

seanoliver:

YES

Bug fix.

In `applyPriorDecisionToSDK`, the coverage check that decides whether to silently restore prior consent or re-prompt filters services from `UC.getServicesBaseInfo()` by `!s.isEssential` only.

The Usercentrics SDK also exposes `isHidden: boolean` on each service — services that exist in the ruleset but aren't surfaced in the Privacy Settings UI. Those still get counted in coverage. A user can never have a stored decision for a service the modal doesn't show them, so any ruleset containing a hidden non-essential would force every returning user to re-prompt forever.

GROWTH-790.

## What is the new behavior?

The filter excludes both `isEssential` and `isHidden`. Coverage matches what the user can actually decide about in the UI — same intent as the existing essential exclusion (skip services the user can't meaningfully consent to).

## Additional context

**Scope.** Defense-in-depth. The Usercentrics ruleset currently has no hidden non-essentials, so this isn't expected to resolve any active customer-reported re-prompts. The contract was wrong as written — adding any hidden service to the ruleset later would silently regress every returning user without this fix.

**Companion PR:** #45341 — independent write-path fix for the Privacy Settings modal.

**Testing.**

- `npx vitest run packages/common/consent-state.test.ts` — 28/28 pass (existing 27 + one new regression test asserting that a hidden non-essential service not in the user's stored decisions doesn't break coverage).
- `pnpm -F common typecheck` — pass.
- Walked the failure path mentally to confirm the new test exercises the actual bug: without `&& !s.isHidden`, the hidden service stays in `currentNonEssentialIds`, isn't in `coveredIds`, `allCovered` becomes false, banner shows. With the filter, it passes.

GROWTH-790

### DIFF
--- a/packages/common/consent-state.test.ts
+++ b/packages/common/consent-state.test.ts
@@ -311,7 +311,7 @@ describe('detectPriorConsent', () => {
 })
 
 describe('applyPriorDecisionToSDK', () => {
-  type MockService = { id: string; isEssential: boolean }
+  type MockService = { id: string; isEssential: boolean; isHidden?: boolean }
 
   const makeMockUC = (
     opts: {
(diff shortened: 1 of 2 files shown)


seanoliver:
Closing without merging. After review: neither fix addresses the active customer-reported re-prompt bug. The suspected cause is cross-app fast navigation, which is structurally untestable on Vercel preview because each app is served from its own domain. Both PRs were defense-in-depth correctness fixes for paths not reachable in the current ruleset (one toggleable non-essential category, no hidden non-essentials). Code preserved in branch history; revive if the ruleset changes to expose either bug, or alongside an actual fix for the cross-app race.

PR #45353, opened 2026-04-29
fix(studio): restore Safari table editor cell copy and context menu

dnywh:

YES

Bug fix.

- Safari Table Editor cells fail to copy from a focused cell with `⌘C`.
- Safari right-click can show the browser menu instead of the custom cell menu.
- Copy can leave RDG's copied-cell fill behind.

## What is the new behavior?

- Reuses the existing shared `copyToClipboard(value, onSuccess)` pattern, with the Safari clipboard fix inside that util.
- Handles selected-cell `⌘C` in the RDG keydown path, preventing browser/RDG defaults and showing the success toast only after copy.
- Replaces the row-level synthetic context-menu shim with RDG's `onCellContextMenu`, so we prevent Safari's browser menu at the source and select/focus the target cell.
- Keeps the selected-cell outline while the controlled menu is open.

## Additional context

- `RowRenderer` was only supporting the old context-menu shim; removing it is part of moving to RDG's cell event path.

## Release Notes

* **New Features**
  * Context menu now provides feedback with toast notifications when copying cells or rows.
  * Selected cells retain their visual styling when context menu is open.

* **Bug Fixes**
  * Improved keyboard shortcut handling for copy functionality.
  * Enhanced clipboard error handling with user-friendly error messages.


apps/studio/components/grid/components/grid/RowRenderer.tsx
dnywh, line 1: Moving from this `RowRenderer` to RGD’s cell context-menu event was the major fix. Hence why this was deleted.

awaseem (approved):
LGTM

kemaldotearth (approved):

PR #45372, opened 2026-04-29
docs: prefer publishable/secret key naming in SDK specs

mandarini:
## Description

Replaces legacy `anon` / `service_role` key references with the new `publishable` / `secret` naming across SDK spec files under `apps/docs/spec/`. Mirrors the rename table established by [supabase-js#2280](https://github.com/supabase/supabase-js/pull/2280), so the auto-generated reference docs at supabase.com/docs/reference/{js,dart,kotlin,python,swift} stay consistent with the SDK source.

## Files changed

| File | Highlights |
| --- | --- |
| `supabase_js_v2.yml` | Rename example vars `anon_key` → `publishable_key`, `service_role_key` → `secret_key`; admin notes use `secret`. |
| `supabase_dart_v2.yml` | Rename `anonKey` param → `publishableKey` (matches the live Flutter SDK — see [supabase-flutter/supabase.dart#L81](https://github.com/supabase/supabase-flutter/blob/main/packages/supabase_flutter/lib/src/supabase.dart#L81)); update call-site; example var `serviceRoleKey` → `secretKey`; admin notes use `secret`. |
| `supabase_kt_v2.yml`, `supabase_kt_v3.yml` | Replace `'publishable-or-anon-key'` placeholders with `'your-publishable-key'`; admin notes use `secret`; fix `importAuthToken(\"service_role\")` → `importAuthToken(\"your-secret-key\")` (the Kotlin SDK's `importAuthToken(accessToken: String, …)` takes a JWT, not a role name). |
| `supabase_py_v2.yml` | All `service_role` references → `secret`; example var `service_role_key` → `secret_key`. |
| `supabase_swift_v2.yml` | Replace placeholders with `'your-publishable-key'`; admin notes use `secret`; example var `serviceRoleKey` → `secretKey`. |
| `storage_v0_config.yaml` | Rewrite `ANON_KEY` / `SERVICE_KEY` env var **descriptions** in publishable/secret terms. **Env var names kept** — see below. |

## Why the storage env var names are unchanged

`storage_v0_config.yaml` documents the env vars used to configure the storage server (multi-tenant deployments). The `id` and `title` fields must match the actual env var names the binary reads. The storage server still reads its config from env vars literally named `ANON_KEY` and `SERVICE_KEY`:

- [`storage/src/config.ts#L614`](https://github.com/supabase/storage/blob/master/src/config.ts#L614) — `getOptionalConfigFromEnv('SERVICE_KEY')`
- [`storage/src/config.ts#L625`](https://github.com/supabase/storage/blob/master/src/config.ts#L625) — `getOptionalConfigFromEnv('ANON_KEY')`

There is no `PUBLISHABLE_KEY` / `SECRET_KEY` reader on master. Renaming the doc IDs would break self-hosted deployments — users following the docs would set the wrong env var and the storage server would silently fall back to generating its own JWT. Until the storage server itself adds publishable/secret env support, only descriptions can be updated.

(Same reasoning for not bumping `storage_v0_config.yaml` → `storage_v1_config.yaml`: the `_v0_` tracks the storage server's own API version, paired with `storage_v0_openapi.json` downloaded from `supabase.github.io/storage/api.json`. The server hasn't shipped a v1 API.)

## What we deliberately did NOT rename

Per the same rules established in supabase-js#2280:

- **JWT role claims** like `role: 'anon'` / `role: 'service_role'` — these are functional Postgres role names in JWT payloads, not key labels.
- **Real SDK identifiers** that haven't been renamed in the source (we only rename in the doc when the underlying SDK rename has shipped). The Dart `anonKey` rename was safe to apply because the Flutter SDK already ships `publishableKey` as the preferred named parameter (with `anonKey` `@Deprecated`).

## Out of scope

- **All `*_v1.yml` SDK spec files** (`supabase_js_v1`, `supabase_dart_v1`, `supabase_kt_v1`, `supabase_swift_v1`). Older SDK versions, not worth churning.
- **`cli_v1_commands.yaml`**. This file is auto-generated by the CLI repo's release workflow ([`cli/tools/bumpdoc/main.go`](https://github.com/supabase/cli/blob/develop/tools/bumpdoc/main.go)) and the example outputs come from [`cli/docs/templates/examples.yaml`](https://github.com/supabase/cli/blob/develop/docs/templates/examples.yaml) — embedded in the CLI binary at build time. Any edits we make here would be clobbered by the next CLI release. The fix needs to land upstream in the CLI repo (note: the CLI itself already ships publishable/secret naming in \`supabase status\` output — see [`cli/internal/status/status.go#L40-L44`](https://github.com/supabase/cli/blob/develop/internal/status/status.go#L40-L44) — but the doc-generation template is stale).
- **`supabase_csharp_v0.yml` / `supabase_csharp_v1.yml`** — checked, already neutral. Both files use a generic `SUPABASE_KEY` env var with no `anon` / `service_role` references.
- **Renaming the legacy Dart `anonKey` parameter itself** — that's an SDK-side change. The Flutter SDK already exposes `publishableKey` as the preferred parameter; full removal of `anonKey` will happen in a future major version per the SDK's own deprecation comment.

## Summary by CodeRabbit

* **Documentation**
  * Standardized API key naming across SDK docs: use publishable (client) and secret (server/admin) key terminology
  * Updated server-side admin examples and warnings to require secret keys and emphasize never exposing them in client code
  * Unified initialization examples across JavaScript, Dart, Kotlin, Python, and Swift
  * Corrected Storage spec metadata to point to the proper configuration file

### DIFF
--- a/apps/docs/spec/storage_v0_config.yaml
+++ b/apps/docs/spec/storage_v0_config.yaml
@@ -5,9 +5,9 @@ info:
   id: 'storage' # {string} A unique ID for this tool.
   version: 'next' # {string} The current version number of the tool.
   title: 'Storage' # {string} A readable name.
-  source: 'https://github.com/supabase/storage-api' # {string} Where developers can find the source code.
-  bugs: 'https://github.com/supabase/storage-api/issues' # {string} Where developers can file bugs.
-  spec: 'https://github.com/supabase/supabase/blob/master/spec/storage_v1_config.yml' # {string} Where developers can find this spec (to link directly in the docs).
+  source: 'https://github.com/supabase/storage' # {string} Where developers can find the source code.
+  bugs: 'https://github.com/supabase/storage/issues' # {string} Where developers can file bugs.
+  spec: 'https://github.com/supabase/supabase/blob/master/spec/storage_v0_config.yml' # {string} Where developers can find this spec (to link directly in the docs).
   description: |
     A sample `.env` file is located in the [storage repository](https://github.com/supabase/storage-api/blob/master/.env.sample).
 
@@ -29,7 +29,7 @@ parameters:
     required: true
     # default: '5432'
     description: |
-      A long-lived JWT with anonymous Postgres privileges.
+      A long-lived publishable JWT used for unauthenticated client requests against this Storage tenant.
 
   - id: 'SERVICE_KEY' # {string} A unique identifier for this param.
     title: 'SERVICE_KEY' # {string} Any name.
@@ -38,7 +38,7 @@ parameters:
     required: true
     # default: '5432'
     description: |
-      A long-lived JWT with Postgres privileges to bypass Row Level Security.
+      A long-lived secret JWT used to bypass Row Level Security on this Storage tenant. Never expose this key in client applications.
 
   - id: 'TENANT_ID' # {string} A unique identifier for this param.
     title: 'TENANT_ID' # {string} Any name.
--- a/apps/docs/spec/storage_v0_config.yaml
+++ b/apps/docs/spec/storage_v0_config.yaml
@@ -29,7 +29,7 @@ parameters:
     required: true
     # default: '5432'
     description: |
-      A long-lived publishable JWT used for unauthenticated client requests against this Storage tenant.
+      A long-lived JWT with anonymous Postgres privileges.
 
   - id: 'SERVICE_KEY' # {string} A unique identifier for this param.
     title: 'SERVICE_KEY' # {string} Any name.
@@ -38,7 +38,7 @@ parameters:
     required: true
     # default: '5432'
     description: |
-      A long-lived secret JWT used to bypass Row Level Security on this Storage tenant. Never expose this key in client applications.
+      A long-lived JWT with Postgres privileges to bypass Row Level Security.
 
   - id: 'TENANT_ID' # {string} A unique identifier for this param.
     title: 'TENANT_ID' # {string} Any name.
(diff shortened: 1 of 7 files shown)


ChrisChinchilla (approved):
At some point, we should get consistent naming and casing of everything, but for now, it's fine. @fadymak should also take a look.

apps/docs/spec/storage_v0_config.yaml
```diff
@@ -29,7 +29,7 @@ parameters:
     required: true
     # default: '5432'
     description: |
-      A long-lived JWT with anonymous Postgres privileges.
+      A long-lived publishable JWT used for unauthenticated client requests against this Storage tenant.
```
fadymak: The publishable key is not a JWT. I don't think we should change this description since it still points to `ANON_KEY` as the parameter right?

apps/docs/spec/storage_v0_config.yaml
```diff
@@ -38,7 +38,7 @@ parameters:
     required: true
     # default: '5432'
     description: |
-      A long-lived JWT with Postgres privileges to bypass Row Level Security.
+      A long-lived secret JWT used to bypass Row Level Security on this Storage tenant. Never expose this key in client applications.
```
fadymak: Same here, it still refers to `SERVICE_KEY`

apps/docs/spec/supabase_dart_v2.yml
```diff
@@ -94,7 +94,7 @@ functions:
           Future<void> main() async {
             await Supabase.initialize(
               url: 'https://xyzcompany.supabase.co',
-              anonKey: 'publishable-or-anon-key',
+              publishableKey: 'your-publishable-key',
```
fadymak: I believe this is pending the release of the latest SDK as `publishableKey` does not yet exist: https://github.com/supabase/supabase-flutter/pull/1360

fadymak (changes requested):

fadymak (approved):

PR #45374, opened 2026-04-29
fix(pg-meta): include schema in pg_indexes join to avoid cross-schema index collisions

Akash504-ai:

YES

Bug fix

The query in `sql/indexes.ts` joins `pg_class` with `pg_indexes` using only the index name:

```sql
JOIN pg_indexes ix ON c.relname = ix.indexname
```

This can lead to incorrect results when multiple schemas contain indexes with the same name. PostgreSQL allows identical index names across different schemas, so this join may return the wrong index_definition.

## What is the new behavior?

The join condition now includes the schema name:

```sql
JOIN pg_indexes ix 
  ON c.relname = ix.indexname 
  AND n.nspname = ix.schemaname
```

This ensures the correct index_definition is retrieved for each index, even when duplicate index names exist across schemas.

## Additional context

- Added a test case to verify behavior when the same index name exists in multiple schemas.
- This change prevents cross-schema collisions and ensures accurate index metadata retrieval.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved index listing so indexes that share the same name in different schemas are correctly distinguished and reported.

* **Tests**
  * Added a regression test ensuring indexes with identical names in separate schemas are both detected and contain expected schema-qualified references.


Akash504-ai:
Hi @charislam, This fixes a cross-schema index collision issue in pg-meta by including the schema in the pg_indexes join.
Would you mind taking a look when you have time? Happy to make any changes if needed!

charislam (approved):
Thanks @Akash504-ai!

charislam (approved):
Thanks @Akash504-ai!

PR #45391, opened 2026-04-30
fix(www): partner integration links

dnywh:

Bug fix.

The integrations listing links tiles to short partner URLs such as `/partners/vercel`, while the canonical integration pages live at `/partners/integrations/vercel`.

Those short URLs are handled by the legacy `/partners/[slug]` route, which is meant to redirect technology partners to `/partners/integrations/[slug]`. However, that route currently pre-renders partner slugs and returns `notFound` when `npm_lifecycle_event === 'build'`, which can bake and cache static 404s for URLs that are meant to redirect.

This was reportedly working as recently as last week, and the relevant partners page code has not changed recently. The most likely explanation is that old legacy-route behaviour was surfaced by a recent production build/deployment/prerender cache interaction, rather than a fresh change to the integrations tile code itself.

## What is the new behaviour?

Integration tiles now link directly to the canonical `/partners/integrations/[slug]` URL, so the listing no longer depends on the legacy short-url redirect.

The legacy short partner URL route also no longer pre-renders partner slugs as build-time 404s. It remains as a blocking fallback redirect for old external links, bookmarks, search results, and historical references that may still point at `/partners/[slug]`.

## Testing

This URL:

```
https://zone-www-dot-com-git-dnywh-fix-partner-integrat-b3fb68-supabase.vercel.app/partners/integrations/vercel
```

...should resolve to:

```
https://zone-www-dot-com-git-dnywh-fix-partner-integrat-b3fb68-supabase.vercel.app/partners/vercel
```

### DIFF
--- a/apps/www/components/Partners/TileGrid.tsx
+++ b/apps/www/components/Partners/TileGrid.tsx
@@ -27,7 +27,7 @@ export default function TileGrid({
           <h2 className="h2">Featured</h2>
           <div className="grid grid-cols-1 gap-5 lg:max-w-none lg:grid-cols-2 xl:grid-cols-3">
             {featuredPartners?.map((p) => (
-              <Link key={p.slug} href={`/partners/${p.slug}`}>
+              <Link key={p.slug} href={`/partners/integrations/${p.slug}`}>
                 <div
                   className="
                 bg-surface-100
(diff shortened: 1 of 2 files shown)


alexhall (approved):
Integrations tiles are fixed on the preview build 👍

SaxonF (approved):

ana1337x (approved):

PR #45405, opened 2026-04-30
chore(billing): disable tax ID banner

kanishkdudeja:
This PR disables the tax ID banner. I haven't removed the code just in case we need to reintroduce it later.

Will follow up with removing the code entirely or introducing a feature flag if we decide to keep it.

## Summary by CodeRabbit

* **Chores**
  * Disabled the tax identification banner display in the application interface.


kevcodez (approved):

awaseem (approved):
LGTM

PR #45421, opened 2026-04-30
fix: Fix infinite loading in the Cron jobs table

ivasilov:
This PR fixes an issue where the infinite list doesn't work for cron jobs. If the initial data (20 cronjobs) renders and you have a tall monitor to fit the entire table, the `react-data-grid` can't `onScroll` event which fetches the next page.

To test:
1. Add 40-50 cron jobs by running `SELECT cron.schedule('test 30', '0 2 * * 1', 'select 1');` repeatedly.
2. Open the /project/_/integrations/cron/jobs
3. See if you can scroll and trigger the second page.

Fixes https://linear.app/supabase/issue/FE-3117/cron-job-is-not-scrollable


awaseem (approved):
LGTM

PR #45423, opened 2026-04-30
fix: Fix the input group in database settings.

ivasilov:
The input groups in Database settings were rendered badly due to the Tailwind v4 bump.

Before:
Screenshot 2026-04-30 at 16 59 55

After: 
Screenshot 2026-04-30 at 17 04 56


awaseem (approved):
Tested on staging

PR #45430, opened 2026-04-30
fix: top-up loading state and show plan item on downgrade

ignaciodob:
## Summary

- **Subscription downgrade dialog**: The plan name row was being hidden in the charge breakdown when `changeType === 'downgrade'`, so users downgrading (e.g. Team → Pro) couldn't see which plan the cost referred to. Removed the downgrade exclusion so the plan name row renders consistently across upgrade/downgrade flows.
- **Credit Top Up dialog**: The submit button only reflected `executingTopUp`/`paymentConfirmationLoading`, but the `onSubmit` handler runs several async steps first (hCaptcha, billing profile validation, Stripe `createPaymentMethod`) before the mutation flips `isPending`. That left a clickable window where users could trigger multiple top-ups. Added `form.formState.isSubmitting` to both `loading` and `disabled` so the button is locked for the full submit lifecycle.

## Test plan

- [ ]  Downgrade from Team → Pro and confirm the Pro plan row appears in the charge breakdown
- [ ]  Open Credit Top Up, submit, and rapidly click the Top Up button — verify only one charge is initiated
- [ ]  Verify Top Up button shows a loading state immediately on click (before the mutation starts)

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistently show subscription plan cost during billing adjustments, including downgrades.
  * Improve Top Up button to reflect form submission state and prevent duplicate submissions.

* **Style**
  * Enhanced text contrast for better readability in billing information displays.

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
@@ -426,13 +426,7 @@ export const CreditTopUp = ({ slug }: { slug: string | undefined }) => {
                   loading={
                     form.formState.isSubmitting || executingTopUp || paymentConfirmationLoading
                   }
-                  disabled={
-                    form.formState.isSubmitting ||
-                    executingTopUp ||
-                    paymentConfirmationLoading ||
-                    isPreviewStale ||
-                    creditPreviewIsFetching
-                  }
+                  disabled={isPreviewStale || creditPreviewIsFetching}
                 >
                   Top Up
                 </Button>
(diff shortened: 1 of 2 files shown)


apps/studio/components/interfaces/Organization/BillingSettings/CreditTopUp.tsx
```diff
@@ -423,8 +423,16 @@ export const CreditTopUp = ({ slug }: { slug: string | undefined }) => {
                 <Button
                   htmlType="submit"
                   type="primary"
-                  loading={executingTopUp || paymentConfirmationLoading}
-                  disabled={isPreviewStale || creditPreviewIsFetching}
+                  loading={
+                    form.formState.isSubmitting || executingTopUp || paymentConfirmationLoading
+                  }
+                  disabled={
+                    form.formState.isSubmitting ||
```
joshenlim: Nit but loading state will also cover disabled state actually

So no need to repeat isSubmitting, isExecutingTopUp, and paymentConfirmationLoading here

joshenlim (approved):
Changes LGTM - just a tiny nit comment but approving to unblock!

joshenlim:
(just a tiny nit - it might just be me overthinking btw so feel free to ignore if not a concern! 🙏😅) 

wondering if this might be misleading at first glance (downgrading from team to pro) with the pro team line item

although the "charge today" does clarifies that nothing is charged
image

PR #45432, opened 2026-04-30
fix(docs): negotiate /guides/* markdown via shared helper

pamelachia:
## Summary

This brings docs `/guides/*` to full content negotiation for AI agents (GROWTH-811):
RFC 9110 q-value parsing instead of a `.includes('text/markdown')` substring match,
a 406 when the client rejects every type the route can produce, and markdown rewrites
for known LLM user agents.

I implemented it by extracting the negotiation into a shared `common/markdown-negotiation`
module consumed by both `apps/docs/middleware.ts` and `apps/www/middleware.ts`, rather than
duplicating the helpers into docs and keeping them in sync by hand with www (#45394). Single
source of truth, no re-sync burden. www is refactored onto the shared helper with no behavior
change.

### docs `/guides/*` content negotiation (GROWTH-811)

- Replace the `.includes('text/markdown')` substring match with RFC 9110 q-value parsing.
- Return 406 (`Cache-Control: no-store`, `Vary: Accept`) when Accept excludes every type the
  route serves. Bypassed for LLM user agents, the `.md` suffix, and clients sending no Accept.
- Rewrite to `/api/guides-md/<slug>` for LLM user agents (Claude-User, Claude-Web, ChatGPT-User,
  PerplexityBot) regardless of Accept.
- Preserve the existing `.md` suffix routing and the entire `/reference/*` block.

### Shared negotiation helper

- New `packages/common/markdown-negotiation.ts`: `negotiateMarkdown(signals, route)` returns
  `'markdown' | 'not-acceptable' | 'pass'`. Internalizes q-value parsing, the LLM user-agent
  match, the UA-length cap, and the markdown-vs-html preference.
- `apps/www/middleware.ts`: refactored to consume the shared helper; its duplicated copy of the
  negotiation helpers (added in #45394) is removed. `.md` early-return, changelog routing, and
  first-referrer cookie stamping are unchanged (no behavior change, covered by its existing tests).

### Tests

- New `apps/docs/middleware.test.ts`: q-value priority, the 406 path, `.md` suffix, LLM UA
  override, browser default Accept, training-crawler and substring-embed exclusion, and the
  `/reference/*` exemption.
- New `packages/common/markdown-negotiation.test.ts`: the same decision matrix at the unit level
  (q-values, 406, LLM UAs, `.md`, `*/*`, training crawlers, OWS, out-of-range q).

## Testing (Vercel preview)

After Vercel posts a preview URL, save it once then run the probe set.

```bash
echo 'PREVIEW_HOST' > /tmp/growth-811-host.txt
HOST=$(cat /tmp/growth-811-host.txt)

# 1) Browser-style Accept -> HTML 200
curl -sI -A "Mozilla/5.0" \
  -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8' \
  "https://$HOST/docs/guides/auth"

# 2) Accept: text/markdown -> markdown 200
curl -sI -H 'Accept: text/markdown' "https://$HOST/docs/guides/auth"

# 3) text/html;q=1.0, text/markdown;q=0.5 -> HTML 200
curl -sI -H 'Accept: text/html;q=1.0, text/markdown;q=0.5' "https://$HOST/docs/guides/auth"

# 4) unsupported Accept -> 406 + Cache-Control: no-store + Vary: Accept
curl -sI -H 'Accept: application/x-content-negotiation-probe' "https://$HOST/docs/guides/auth"

# 5) User-Agent: Claude-User/1.0 (any Accept) -> markdown 200
curl -sI -A 'Claude-User/1.0' "https://$HOST/docs/guides/auth"
```

### After merge

Run [acceptmarkdown.com/readiness-check](https://acceptmarkdown.com/readiness-check)
against `https://supabase.com/docs/guides/auth`: expect 100/100.

## Linear

- fixes GROWTH-811


apps/docs/middleware.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      expect(middleware(req).headers.get(REWRITE_HEADER)).toBeNull()
+    }
+  })
+
+  it('does not apply LLM UA rewrite to /reference/* (guides-only)', () => {
+    const req = makeRequest('/docs/reference/javascript/introduction', {
+      userAgent: 'Claude-User/1.0',
+    })
+    expect(middleware(req).headers.get(REWRITE_HEADER)).not.toContain('/api/guides-md/')
+  })
```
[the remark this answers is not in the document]
pamelachia: Fixed in b4ab7db0. Coalesced `headers.get(REWRITE_HEADER) ?? ''` before `.not.toContain` so the assertion fails cleanly instead of throwing on null. Kept `.not.toContain` (not `.toBeNull()`) because the test intent is "no /api/guides-md/* rewrite" — /reference/* may legitimately rewrite to /api/crawlers or a normalized SDK path.

apps/docs/middleware.ts
```diff
@@ -1,33 +1,100 @@
...
+      return new NextResponse('Not Acceptable', {
+        status: 406,
+        headers: { 'Cache-Control': 'no-store', Vary: 'Accept' },
+      })
+    }
 
-  const requestsMarkdown =
-    request.headers.get('Accept')?.includes('text/markdown') || url.pathname.endsWith('.md')
+    const wantsMarkdown =
+      isMdSuffix || isLlmAgent || (accept !== null && shouldServeMarkdown(accept))
```
jeremenichelli: Should we consolidate all this conditions better within a function helper?

We currently have in the same scope `shouldSeverMarkdown`, `wantsMarkdown`, `isMdSuffix` and they are all insufficient to enclose a condition, maybe we could clean this up a bit more so it's not so confusing.

apps/docs/middleware.ts
```diff
@@ -1,33 +1,100 @@
...
+  if (pathname.startsWith(GUIDES_PATH + '/')) {
+    const isMdSuffix = pathname.endsWith('.md')
+    const slug = pathname.replace(`${GUIDES_PATH}/`, '').replace(/\.md$/, '')
+    const acceptHeader = request.headers.get('accept') ?? ''
+    // Cap UA length to bound regex CPU on the edge hot path.
+    const userAgent = (request.headers.get('user-agent') ?? '').slice(0, 512)
+    const isLlmAgent = LLM_USER_AGENT.test(userAgent)
+    const accept = acceptHeader ? parseAccept(acceptHeader) : null
+
+    if (
```
jeremenichelli: A code comment wouldn't hurt here. Why are we returning a `406` here?

jeremenichelli:
Left some comments. Huge kudos for the tests @pamelachia 👏

apps/docs/middleware.ts
pamelachia: Done. I pulled the whole decision (the 406 branch, `wantsMarkdown`, and the `.md`/LLM checks) into a single `negotiateMarkdown()` that returns `'markdown' | 'not-acceptable' | 'pass'`. www needed the identical logic, so I moved it to `common/markdown-negotiation` and pointed both middlewares at it: the helpers are no longer duplicated across the two files. The middleware body now just maps the decision to a response.

apps/docs/middleware.ts
pamelachia: Added. The rationale now lives on the `'not-acceptable'` branch in `negotiateMarkdown`: we return 406 when the client's Accept header excludes every type the route can produce, and the comment spells out the three bypasses (LLM UAs, `.md` suffix, and no Accept header). The docs middleware now maps that decision to the 406 response.

jeremenichelli (approved):

4L3k51 (approved):

PR #45449, opened 2026-05-01
fix: table editor input

### LINKED ISSUE #45445: Editor shortcut keys can block typing into selected cells (Single click)
# Bug report

## Describe the bug

In the Table Editor, pressing some printable keys on a selected editable cell does not start editing the cell.

This happens when the key is also used by a Table Editor keyboard shortcut sequence. For example, pressing `I`, `F`, `S`, `C`, `R`, or `U` on a selected text cell can be intercepted by shortcuts instead of entering edit mode.

## To Reproduce

1. Open a project 
2. Open Table Editor.
3. Open any table with an editable text column.
4. Single-click a text cell so it is selected but not already editing.
5. Press `I`.

## Expected behavior

The selected editable cell should enter edit mode and start with the typed character, the same way other printable keys do.

## Additional context

Other printable keys that are not part of Table Editor shortcut sequences work as expected. I would expect these keys to work the same way when a cell is selected and editable.

This seems like a minor Table Editor keyboard handling issue and should be a small quality-of-life fix.

7ttp:
## TL;DR

typing into a cell after single click was broken for keys like `I`, `F`, `C`,`R`, `U` & `S` 
 because those keys could be picked up as shortcut prefixes instead of starting cell editing

## sol:

 https://github.com/user-attachments/assets/e388b79f-5334-47ef-a834-9164b255b88c

## ref:

- Closes https://github.com/supabase/supabase/issues/45445

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard interaction in grid cells: typing a single printable character now enters editable cells directly (allowing immediate edit-mode), while other registered keyboard shortcuts still take precedence and continue to block default grid behavior.


awaseem:
Nice you can escape out still and use the shortcut

awaseem (approved):
LGTM

PR #45477, opened 2026-05-02
feat(docs): add BreadcrumbList JSON-LD to guide pages

pamelachia:
## Summary

Emits `BreadcrumbList` JSON-LD on every `/docs/guides/*` page served by `GuideTemplate`. Search engines and AI crawlers get an explicit hierarchical signal for the docs site (the marketing site already shipped JSON-LD via #45451). The chain prepends `Docs > Guides` to the existing resolver output, so a page like `/docs/guides/auth/passwords` produces a 5-level chain with the leaf URL set per Google's spec.

## Changes

- New `apps/docs/lib/breadcrumbs.ts`: pure pathname → chain resolver, server-safe. Extracted from the existing client `useBreadcrumbs` hook so the same logic runs in both contexts.
- New `apps/docs/lib/json-ld.ts`: `serializeJsonLd` + `breadcrumbListSchema` mirroring `apps/www/lib/json-ld.ts`.
- `Breadcrumbs.tsx` (visual) now delegates to the shared resolver — single source of truth for visual + SEO chains.
- `GuideTemplate` takes a required `pathname` prop and emits `<script type="application/ld+json">` next to `<Breadcrumbs />`. Skipped when the chain is empty (e.g., page not in nav menu). Middle items without URLs (e.g., the "Auth" section root) omit `item`, matching the visual breadcrumb.
- 8 explicit-prop callers updated; `[[...slug]]` callers already spread `data` (which carries `pathname`).

## Scope

**Out of scope:**
- `/docs/reference/*` (SDK reference) — no breadcrumbs rendered today, would need separate traversal over spec JSON.
- `/guides/troubleshooting/*` — uses its own template, not `GuideTemplate`.
- `TechArticle` per-page schema — high maintenance for marginal value.

## Testing (Vercel preview)

```bash
curl -s https://<preview>/docs/guides/auth/passwords | grep -oE '<script type="application/ld\+json"[^>]*>[^<]+</script>'
```

Expect a script tag with the chain `Docs > Guides > Auth > Flows (How-tos) > Password-based`, leaf URL `https://supabase.com/docs/guides/auth/passwords`.

- [x] `/docs/guides/auth/passwords` — 5-item chain, leaf URL present
- [x] `/docs/guides/getting-started/features` — 4-item chain, all items have URLs
- [x] `/docs/guides/getting-started/ai-prompts/<slug>` — special-case chain (`Getting started > AI Tools > Prompts > <slug>`), leaf URL falls back to pathname
- [x] `/docs/guides/database/database-advisors` (explicit-prop caller) — chain renders
- [x] Visual breadcrumb on the same pages still renders correctly
- [ ] Validate output through [Google Rich Results Test](https://search.google.com/test/rich-results) on a deployed preview URL
- [x] `/docs/guides/troubleshooting/<slug>` — no JSON-LD emitted (different template, intentional)

## Linear

- fixes GROWTH-820

## Summary by CodeRabbit

* **New Features**
  * Added JSON-LD breadcrumb markup to guide pages to improve search/discovery.

* **Improvements**
  * Centralized breadcrumb generation for consistent, accurate breadcrumbs across guides.
  * Multiple guide pages updated to ensure breadcrumbs and page context display correctly.


apps/docs/lib/json-ld.ts
```diff
@@ -0,0 +1,45 @@
...
+  chain: BreadcrumbItem[]
+}
+
+export function breadcrumbListSchema({ pathname, chain }: BreadcrumbListSchemaInput) {
+  const fullChain: BreadcrumbItem[] = [DOCS_ROOT, GUIDES_ROOT, ...chain]
+
+  const itemListElement = fullChain.map((crumb, index) => {
+    const isLeaf = index === fullChain.length - 1
+    const name = crumb.title ?? crumb.name
+    const path = crumb.url ?? (isLeaf ? pathname : undefined)
```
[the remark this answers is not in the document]
pamelachia: Fixed in 1fea281. Changed to `isLeaf ? pathname : crumb.url` so dynamic-slug routes (`ai-prompts/[slug]`, `ai-skills/[slug]`) get the current page URL on the leaf instead of the category index. No-op for regular guide pages where `crumb.url === pathname`.

apps/docs/features/docs/GuidesMdx.template.tsx
```diff
@@ -74,6 +87,12 @@ const GuideTemplate = ({ meta, content, children, editLink, mdxOptions }: GuideT
             'col-span-12 md:col-span-9'
           )}
         >
+          {breadcrumbJsonLd && (
+            <script
+              type="application/ld+json"
+              dangerouslySetInnerHTML={{ __html: breadcrumbJsonLd }}
+            />
+          )}
```
jeremenichelli, line 95: Isn't it possible to place this on the `head` tag? Having the script on the body could pollute data and info extraction scripts for LLMs and increase body parsing.

pamelachia: Good callout @jeremenichelli 

On placement, I checked what peers in our stack do. dev-tool docs on App Router / SSG stacks ship JSON-LD inline in body:
- Vercel docs (Next.js App Router): body, WebSite + TechArticle JSON-LD
- Cloudflare developers docs: body, including a BreadcrumbList (same use case)
- Next.js docs: body, TechArticle

Going to leave it in body to match what everyone else is doing. If naive-extractor pollution matters here, one option is to render the script as the last child of the article (after the footer) so body-text extractors see real content first?

jeremenichelli (approved):

PR #45490, opened 2026-05-03
Expose self-hosted Observability through Query Performance

### LINKED ISSUE #45489: Self-hosted Studio still hard-disables Observability despite existing self-hosted analytics/log plumbing
## Summary

I was looking through the self-hosted Studio code paths and noticed that Observability is still hard-disabled for self-hosted/local, even though there is already self-hosted analytics/log plumbing in the repo.

This feels like a good mid-sized issue because the gap is pretty concrete: the backend support seems to exist in at least partial form, but the Studio product surface still blocks the feature very early.

## What I noticed

There are already self-hosted code paths for analytics/log-related behavior:

- `apps/studio/pages/api/platform/projects/[ref]/analytics/endpoints/[name].ts`
- `apps/studio/lib/api/self-hosted/logs.ts`

But the UI still seems to treat Observability as platform-only:

- `apps/studio/components/layouts/Navigation/NavigationBar/NavigationBar.utils.tsx`
  - currently has a comment saying Observability is only available on the platform, and excludes it for self-hosted
- `apps/studio/components/layouts/ObservabilityLayout/ObservabilityLayout.tsx`
  - gates the whole layout behind `reportsAll` and otherwise returns `UnknownInterface`
- `apps/studio/components/layouts/Navigation/NavigationBar/NavigationBar.utils.test.tsx`
  - has tests that explicitly assert observability should be excluded in self-hosted mode

So right now the self-hosted story seems stricter than the codebase itself actually is.

## Why this seems worth fixing

For self-hosted users, observability/debugging is one of the first places where friction shows up.

Even if full hosted parity is not the goal, it would still be useful if Studio could move from:

- "Observability is hidden entirely for self-hosted"

to something more capability-based, for example:

- expose the entry point when supported pieces are available
- allow partial/supported pages to render
- use clearer empty states for hosted-only pieces instead of hard-blocking the whole area

## Possible scope

I think a reasonable first pass could be:

1. Replace the blanket self-hosted exclusion in navigation with capability-based visibility.
2. Relax the layout-level hard block so supported self-hosted observability surfaces can render.
3. Keep hosted-only sections behind empty states or explicit messaging where needed.
4. Update the existing tests to match the more incremental behavior.

## Why I’m filing this separately

I know there has been broader discussion around self-hosted parity before, but this felt scoped enough to be actionable on its own and small enough to start as a real implementation issue instead of a general parity request.

If this direction sounds good, I’d be happy to take a first pass at it.

mandar1045:
## Summary

This is a small first pass at loosening the self-hosted Observability lockout in Studio.

Right now, self-hosted users have some analytics/log plumbing in the repo, but the UI still blocks Observability very early:

- the nav hides Observability entirely for self-hosted
- the layout returns `UnknownInterface` unless `reports:all` is enabled
- the default Observability landing page falls back to `api-overview`, which is a poor fit for self-hosted

This patch does three things:

- adds a shared helper for Observability access / entry routing
- exposes Observability in self-hosted navigation
- points the self-hosted entry path at `query-performance`, which already has explicit self-hosted support in its data hook

## Changes

- allow self-hosted access to `ObservabilityLayout` even when `reports:all` is false
- show `Observability` in the sidebar for self-hosted projects
- use `/observability/query-performance` as the self-hosted entry route
- update the self-hosted fallback redirect on `/observability`
- add focused tests around the new access and entry-route behavior

## Notes

This does **not** try to solve full hosted/self-hosted observability parity.

It is meant as a narrower step that removes the blanket lockout and makes the already-supported query performance path reachable through Studio.

Closes #45489

## Testing

I added focused test coverage for the new helper and updated the existing navigation expectations.

I was **not** able to run the Studio test suite in this checkout because the JS workspace dependencies are not installed locally, so `vitest` was not available to execute here.

## Summary by CodeRabbit

* **New Features**
  * Observability is now consistently available in self-hosted deployments and selects the appropriate entry route per environment.

* **Bug Fixes**
  * Navigation items and empty-state redirects now route users to the correct observability entry page for their deployment type.

* **Tests**
  * Added tests validating observability access rules and correct entry-route selection.

### DIFF
--- a/apps/studio/components/layouts/Navigation/NavigationBar/NavigationBar.utils.test.tsx
+++ b/apps/studio/components/layouts/Navigation/NavigationBar/NavigationBar.utils.test.tsx
@@ -129,15 +129,19 @@ describe('generateOtherRoutes', () => {
       isPlatform: false,
       showReports: true,
     })
-    expect(keys(routes)).not.toContain('observability')
+    expect(keys(routes)).toContain('observability')
+    const observabilityRoute = routes.find((r) => r.key === 'observability')
+    expect(observabilityRoute?.link).toBe(`/project/${REF}/observability/query-performance`)
   })
 
-  it('excludes observability in self-hosted mode when reports are disabled', () => {
+  it('includes observability in self-hosted mode when reports are disabled', () => {
     const routes = generateOtherRoutes(REF, activeProject, {
       isPlatform: false,
       showReports: false,
     })
-    expect(keys(routes)).not.toContain('observability')
+    expect(keys(routes)).toContain('observability')
+    const observabilityRoute = routes.find((r) => r.key === 'observability')
+    expect(observabilityRoute?.link).toBe(`/project/${REF}/observability/query-performance`)
   })
 
   it('does not include API Docs nav item', () => {
--- a/apps/studio/components/layouts/Navigation/NavigationBar/NavigationBar.utils.tsx
+++ b/apps/studio/components/layouts/Navigation/NavigationBar/NavigationBar.utils.tsx
@@ -2,6 +2,10 @@ import { Auth, Database, EdgeFunctions, Realtime, SqlEditor, Storage, TableEdito
 import { Blocks, Lightbulb, List, Settings, Telescope } from 'lucide-react'
 
 import { ICON_SIZE, ICON_STROKE_WIDTH } from '@/components/interfaces/Sidebar'
+import {
+  canAccessObservability,
+  getObservabilityEntryRoute,
+} from '@/components/layouts/ObservabilityLayout/ObservabilityAccess.utils'
 import type { Route } from '@/components/ui/ui.types'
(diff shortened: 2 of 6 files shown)


mandar1045:
A quick note on the failing checks: I dug into the run metadata and it looks like this PR is currently blocked at the GitHub Actions approval gate for external PRs. The `Validate pull request` workflow finished as `action_required` with no jobs started, so this does not look like a code/test failure in the patch itself. The three failing Vercel statuses are still showing `Authorization required to deploy.`, but the repo already has an `Authorize Vercel Deploys` workflow for that path. My read is that once a maintainer approves/reruns the validation workflow for this external PR, the normal authorization flow should be able to proceed and those preview checks should clear.

aantti:
I appreciate the effort, but there's currently no plans to add extended observability features identical the platform-specific ones.

PR #45503, opened 2026-05-04
fix(studio): surface FK constraint errors and remove false-success timeout in CSV import

mandar1045:
## Summary

This fixes two false-success paths in Studio's table editor:

1. Table updates could appear successful even when the follow-up foreign key constraint update failed.
2. CSV imports could appear successful after 30 seconds even if the underlying insert was still running.

## Why this matters

Both issues can leave users thinking their change completed when the database is actually in a different state.

That makes these failures easy to miss, especially during schema edits or slower imports, and can lead to follow-up errors that are harder to debug because Studio reported success too early.

## What changed

- Catch and surface foreign key constraint update failures during table edits.
- Mark the table update flow as a partial failure when the base table update succeeds but FK updates fail.
- Remove the 30s timeout race from CSV imports so success is only reported after the insert actually completes.
- Add tests covering:
  - FK update failures being surfaced correctly
  - Slow imports not resolving early

## Scope and edge cases

- This only affects foreign key updates triggered through the `SidePanelEditor` table update flow.
- It does not change foreign key behavior in other Studio flows.
- Large CSV imports may still take time, but they will no longer be reported as successful before the database finishes processing them.
- Import batching is unchanged, and this does not add extra database round trips or additional work per batch.

## Reviewer notes

The main behavioral change is that we now prefer accurate status over optimistic success:
- if a table update succeeds but FK constraint creation fails, the user now sees the FK failure instead of a silent success
- if an import is slow, the UI waits for the real database result instead of timing out into a false positive

## Testing

- `pnpm --filter studio exec vitest run components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.mutations.test.ts`

### Foreign key flow

Before saving:

test1_save_with_fk_config

Resulting table state:

test1_final_table_state

### Slow import flow

Import still in progress:

test3_progress_10s

Data visible after completion:

test3_data_finally_appeared

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for foreign-key update failures with clear user-facing error messages.
  * More reliable batch data insertion and consistent progress reporting during inserts.

* **Tests**
  * Added tests verifying foreign-key error handling, toast notifications, and insertion progress behavior.

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
@@ -52,7 +52,7 @@ import {
 } from '@/data/tables/table-update-mutation'
 import { getTables } from '@/data/tables/tables-query'
 import { sendEvent } from '@/data/telemetry/send-event-mutation'
-import { isObject, isObjectContainingKeys, timeout, tryParseJson } from '@/lib/helpers'
+import { isObject, isObjectContainingKeys, tryParseJson } from '@/lib/helpers'
 import type { SafePostgresColumn } from '@/lib/postgres-types'
 import type { DeepReadonly } from '@/lib/type-helpers'
 import type { SidePanel } from '@/state/table-editor'
@@ -921,13 +921,20 @@ export const updateTable = async ({
   }
 
   // Foreign keys will get updated here accordingly
-  await updateForeignKeys({
-    projectRef,
-    connectionString,
-    table: updatedTable,
-    foreignKeys: foreignKeyRelations,
-    existingForeignKeyRelations,
-  })
+  try {
+    await updateForeignKeys({
+      projectRef,
+      connectionString,
+      table: updatedTable,
+      foreignKeys: foreignKeyRelations,
+      existingForeignKeyRelations,
+    })
+  } catch (error: any) {
+    hasError = true
+    toast.error(
+      `Failed to update foreign key constraints: ${error.message}`
+    )
+  }
 
   await Promise.all([
     queryClient.invalidateQueries({ queryKey: tableEditorKeys.tableEditor(projectRef, table.id) }),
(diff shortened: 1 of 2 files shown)


mandar1045:
Good catch. I removed the `new Promise(async ...)` wrapper and switched the batch task to a plain async function, so rejections now flow naturally through `Promise.allSettled` without the async executor antipattern.

apps/studio/components/interfaces/TableGridEditor/SidePanelEditor/SidePanelEditor.utils.tsx
```diff
@@ -1141,22 +1148,10 @@ export async function insertTableRows({
...
-            insertError = error
-            reject(error)
-          }
-
-          insertProgress = insertProgress + batch.length / rows.length
-          resolve({})
-        }),
-        timeout(30_000),
-      ])
+    return async () => {
```
adityakrmishra, line 1145: As a student learning the codebase, seeing the removal of the 30s timeout race here is a great lesson in why we should prioritize accurate status over optimistic timeouts in long-running operations. Really clean refactor!

adityakrmishra (commented):
Hi! I'm a student dev exploring the Table Editor's mutation logic. I pulled this PR locally to study how we handle complex multi-step database updates and long-running imports.

I really appreciated the fix in `insertTableRows`. Removing the hardcoded 30s timeout race is a great lesson in maintaining UI/DB synchronization—it's much better to await the actual result than to risk a "false-success" state for the user. I also learned a lot from the new `SidePanelEditor.utils.mutations.test.ts`; specifically, seeing how you mocked the `@supabase/pg-meta` query builder was very helpful for understanding how we test SQL generation.

Verdict: The code is clean, well-tested, and solves the issue without side effects. I recommend merging.

mandar1045:
cc @joshenlim for review when you have a chance! (Merge conflicts are now resolved as well)

Hallidayo:

mandar1045:
> Hi there, we move pretty fast here at Supabase, and we want to make sure that PRs that are relevant to the work get reviewed and worked on.
> 
> This PR is now over 60 days old, and from our understanding, it isn't addressing a bug or other active issue, so we're going to close it for now.
> 
> We do appreciate your time and contribution to Supabase :green_heart: 

Thanks for taking the time to review it, really appreciate it! 🙌

If you have some time, could you please also take a look at my issue #48977? I’d really appreciate your feedback on that one as well.

PR #45511, opened 2026-05-04
Add admonition RE ipv4 addon for connect if direct connection selected

joshenlim:
## Context

Adds an admonition in the Connect sheet to inform users about the IPv4 addon if direct connection is selected and project doesn't have the IPv4 addon

Decided to place it below the copy prompt CTA since it's technically a secondary action (users with IPv6 networks wouldn't need this)

image

## Summary by CodeRabbit

* **New Features**
  * Displays contextual guidance in direct connection mode when the IPv4 add-on is not enabled, including quick-access links to configure IPv4 settings and to open IPv4 documentation.


apps/studio/data/subscriptions/project-addons-query.ts
```diff
@@ -44,3 +46,10 @@ export const useProjectAddonsQuery = <TData = ProjectAddonsData>(
     staleTime: 60 * 60 * 1000,
     ...options,
   })
+
+export const useIpv4AddonQuery = () => {
```
alaister: nit:

```suggestion
export const useHasIpv4AddonQuery = () => {
```

joshenlim: hmm but this returns the ipv4 add on itself though, doesnt return boolean

alaister: Oh true sorry – I was just looking at how it was being used

apps/studio/data/subscriptions/project-addons-query.ts
```diff
@@ -44,3 +45,10 @@ export const useProjectAddonsQuery = <TData = ProjectAddonsData>(
...
+  const { ref } = useParams()
+  const { data: addons } = useProjectAddonsQuery({ projectRef: ref })
+  const selectedAddons = addons?.selected_addons ?? []
+  return selectedAddons.find((addon) => addon.type === 'ipv4')
```
alaister: Sorry I know I'm still nit picking here, but I feel like this whole file isn't needed and `useProjectAddonsQuery({ projectRef: ref })` could be used directly in the component along with a `select: () => …` in the query options. That would also solve coderabbits issue with not being able to have a loading state

joshenlim: oooh great point! will change

joshenlim: just pushed up! 🙏

apps/studio/components/interfaces/ConnectSheet/ConnectStepsSection.tsx
```diff
@@ -157,6 +173,20 @@ export function ConnectStepsSection({ steps, state, projectKeys }: ConnectStepsS
 
         <CopyPromptAdmonition stepsContainerRef={stepsContainerRef} />
 
+        {state.mode === 'direct' && !ipv4Addon && (
```
alaister: Another small nit: it might be nice not to show this admonition when users select session pooler, since that's another option to connect via IPv4 without the addon

Image

This also applies when this checkbox is selected in the transaction pooler mode:
Image

alaister (approved):
LGTM!

- Tested on preview

One small comment:

PR #45512, opened 2026-05-04
fix: table hover styles are incorrect

djhi:
## Problem

When migrating to tailwind v4, we introduced a regression on table styles when hovering a row:
image

## Solution

Fix the styles:
image

## Summary by CodeRabbit

* **Style**
  * Enhanced table row and cell styling to improve hover effects and selection state visual feedback, providing clearer and more consistent interactions when working with tabular data.


joshenlim (approved):
Verified on preview!

dnywh (approved):
Tested on staging ×2!

PR #45524, opened 2026-05-04
feat(mcp): add Claude.ai and ChatGPT connectors with grouped client dropdown

Rodriguespn:
## Summary

- Adds Claude.ai and ChatGPT as MCP connectors in the existing `/guides/getting-started/mcp` dropdown

Screenshot 2026-05-04 at 14 53 24

- Groups all MCP clients into three categories: **Web Clients** (Claude.ai, ChatGPT, Goose), **AI Agent CLIs** (Claude Code, Codex, Gemini CLI, Opencode, Factory), **IDE** (Cursor, VS Code, Antigravity, Kiro, Windsurf)

image

- Both connectors use `generateDeepLink` to render a one-click "Connect" button linking to the Supabase connector directory entry
- Includes external docs links for each connector
- Sets Claude Code as the default selected client

Closes [AI-699](https://linear.app/supabase/issue/AI-699/add-claudeai-and-chatgpt-connectors-to-mcp-docs)

## Summary by CodeRabbit

* **New Features**
  * Added Claude.ai and ChatGPT as selectable client options.
  * Reorganized the client selector into categorized groups (AI Agent CLI, Web Clients, IDE) for easier discovery.
  * Improved dropdown rendering and selection behavior for more consistent visuals and interaction.

### DIFF
--- a/packages/ui-patterns/src/McpUrlBuilder/McpConfigPanel.tsx
+++ b/packages/ui-patterns/src/McpUrlBuilder/McpConfigPanel.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import React, { useMemo, useRef, useState } from 'react'
+import React, { useMemo, useState } from 'react'
 import { cn, Separator } from 'ui'
 import { CodeBlock } from 'ui-patterns/CodeBlock'
 
@@ -16,6 +16,13 @@ import {
 } from './constants'
 import type { McpClient, McpOnCopyCallback } from './types'
 import { getMcpUrl } from './utils/getMcpUrl'
+
+const CLIENT_GROUPS = MCP_CLIENT_GROUPS.map((group) => ({
+  heading: group.heading,
+  clients: group.keys
+    .map((key) => MCP_CLIENTS.find((c) => c.key === key))
+    .filter(Boolean) as (typeof MCP_CLIENTS)[number][],
+}))
 
 export interface McpConfigPanelProps {
   baseUrl?: string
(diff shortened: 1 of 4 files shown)


fsansalvadore (approved):
Looks good on the preview 👍

packages/ui-patterns/src/McpUrlBuilder/McpConfigPanel.tsx
```diff
@@ -105,6 +110,12 @@ export function McpConfigPanel({
         <ClientSelectDropdown
           label="Client"
           clients={MCP_CLIENTS}
+          groups={MCP_CLIENT_GROUPS.map((group) => ({
+            heading: group.heading,
+            clients: group.keys
+              .map((key) => MCP_CLIENTS.find((c) => c.key === key))
+              .filter(Boolean) as (typeof MCP_CLIENTS)[number][],
+          }))}
```
jeremenichelli: This is doing a lot of array operations on each render. I think it might be worth putting it on a `ref` via `useRef` so it stays memoized.

jeremenichelli: Since it's a bit niche, I will approve, you can do the modification and later merge Pedro 😉

packages/ui-patterns/src/McpUrlBuilder/components/ClientSelectDropdown.tsx
```diff
@@ -41,6 +48,33 @@ export const ClientSelectDropdown = ({
     setOpen(false)
   }
 
+  function renderClient(client: McpClient) {
```
jeremenichelli, line 51: Doesn't it make sense to just turn this into a component that supports a client array prop like `McpClient[]`?

jeremenichelli (approved):

ChrisChinchilla:
Looks nice @Rodriguespn like @jeremenichelli says I'd also see how we can get this (and other similar custom components created for single usage) into broader used components.

packages/ui-patterns/src/McpUrlBuilder/McpConfigPanel.tsx
Rodriguespn: Good catch! Moved it to a module-level constant instead ([here](https://github.com/supabase/supabase/pull/45524/changes#diff-f7d1411198abac419068e08aa5cad6e338a19347c894e79434639a37e77c163dR20)), since both `MCP_CLIENT_GROUPS` and `MCP_CLIENTS` are static, computing it outside the component means it runs once at import time rather than on every render or every mount. No hook needed at all.

packages/ui-patterns/src/McpUrlBuilder/components/ClientSelectDropdown.tsx
Rodriguespn: Agreed. Replaced the one-off `ClientDropdown` in `AgentPluginsPanel` with the shared `ClientSelectDropdown` from `ui-patterns/McpUrlBuilder`. `PluginClient` now extends `McpClient` so the dropdown just works, and the component-specific metadata (repoUrl, docsUrl) stays alongside without needing a separate lookup. Thanks for the suggestion

PR #45551, opened 2026-05-04
feat(studio): wire list-page shortcuts on database replication and migrations pages

awaseem:
## Summary

Wires the existing `list-page.*` shortcuts up to the Database → Replication and Database → Migrations pages, so they get the same hotkey behavior as Roles, Tables, Publications, etc. No new shortcut IDs were added — these reuse the already-registered bindings.

**Migrations page**
- `Shift+F` → focus the migration search input (label: "Search migrations")
- `F` then `C` → clear the search filter

**Replication / Destinations page**
- `Shift+F` → focus the destinations filter input (label: "Search destinations")
- `F` then `C` → clear the filter
- `Shift+N` → open the Add Destination panel. Wrapped with `<Shortcut>` so the keybind tooltip shows on the toolbar button on hover, and gated on `!!newDestinationDefaultType` so it stays inert when no destination type is available.

All three are surfaced in the `Cmd+K` command menu under "Shortcuts" while on the respective page (via `registerInCommandMenu: true` on the registry entries). They are intentionally **not** listed in Account → Preferences — list-page shortcuts use `showInSettings: false` so users discover them through the command menu / hover tooltips rather than a toggle UI, matching the existing behavior on Roles, Tables, Publications, etc.

Closes [FE-3141](https://linear.app/supabase/issue/FE-3141/add-shortcuts-for-database-replication-and-migration-page).

## Test plan

- [x] On the Migrations page, press `Shift+F` → search input focuses and selects existing text.
- [x] On the Migrations page, type a query then press `F` `C` → search clears.
- [x] On the Replication page, press `Shift+F` → filter input focuses and selects.
- [x] On the Replication page, press `Shift+N` → Add Destination panel opens (when a destination type is available).
- [x] On the Replication page, hover the "Add destination" button → keybind tooltip shows `Shift+N`.
- [x] On the Replication page, type a filter then press `F` `C` → filter clears.
- [x] All three shortcuts appear in `Cmd+K` under "Shortcuts" while on the respective page.

### DIFF
--- a/apps/studio/components/interfaces/Database/Migrations/Migrations.tsx
+++ b/apps/studio/components/interfaces/Database/Migrations/Migrations.tsx
@@ -1,6 +1,6 @@
 import { SupportCategories } from '@supabase/shared-types/out/constants'
 import { Search } from 'lucide-react'
-import { useState } from 'react'
+import { useRef, useState } from 'react'
 import {
   Button,
   Card,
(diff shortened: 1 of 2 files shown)


charislam (approved):
Not seeing where one disables list-page shortcuts as described in PR description, but not going to block on it. Everything else looks good in testing!

PR #45572, opened 2026-05-05
feat(studio): add auth email template reset flow

dnywh:

Feature. Resolves AUTH-879.

Studio does not provide a way to reset an Auth email template back to the default subject and body once it has been customised.

## What is the new behavior?

Studio shows a `Reset template` action when Platform reports that the selected Auth email template subject or body has been customised. The action opens a confirmation dialog, calls the dedicated Platform reset endpoint, and refreshes the editor with the default subject and body returned by the API.

The Auth config save/reset mutations now run their user-facing success handling before refreshing Auth lint data, so the success toast and local editor cleanup are not delayed by lint refetches.

## Additional context

Depends on supabase/platform#32417.

## Summary by CodeRabbit

* **New Features**
  * Reset email templates to defaults via a confirmation dialog; button appears when custom content is detected and respects update permissions.
  * Typed email-template definitions and improved template variable display and descriptions.

* **Tests**
  * Added tests covering template reset visibility, confirmation flow, state updates, permission handling, and toast notifications.

* **Documentation**
  * Example email template placeholders updated for internationalization and provider snippets.

[Review Change Stack](https://app.coderabbit.ai/change-stack/supabase/supabase/pull/45572)


apps/studio/data/auth/auth-template-reset-mutation.ts
```diff
@@ -0,0 +1,84 @@
...
+  'recovery',
+  'reauthentication',
+  'password-changed-notification',
+  'email-changed-notification',
+  'phone-changed-notification',
+  'mfa-factor-enrolled-notification',
+  'mfa-factor-unenrolled-notification',
+  'identity-linked-notification',
+  'identity-unlinked-notification',
+] as const
```
joshenlim: hmm im just wondering from a API design POV - if its okay to use the same `id` as what FE does avoid the need for this mapping?

FE uses upper camel case because of what gets returned by auth config

joshenlim: unless FE swaps to use lower kebab case like this instead

joshenlim: but either way these should be derived from the API types if possible, rather than FE keeping its own hard coded set

dnywh: Agree. Minimal-to-zero hardcoding! 

@doublethink has just put out a PR out that moves the default template copy into DB defaults: https://github.com/supabase/platform/pull/32582. So we can just call a config-key reset endpoint instead of a template-slug reset endpoint, negating the need for this `AUTH_TEMPLATE_TYPES` list. Right?

doublethink: This is what I floated at the start of the week, but Chris thought that the config endpoint is already overloaded.
I am inclined to agree

joshenlim: no worries for now btw - we'll make do with the current set up

think i found an alternative to clean things up with minimal hardcoding

apps/studio/components/interfaces/Auth/EmailTemplates/TemplateEditor.tsx
```diff
@@ -341,27 +407,64 @@ export const TemplateEditor = ({ template }: TemplateEditorProps) => {
 
             <SpamValidation spamRules={spamRules} />
 
-            <CardFooter className="flex flex-row justify-end gap-2">
-              {hasChanges && (
+            <CardFooter className="flex flex-row justify-between gap-2">
+              {hasCustomTemplate && !hasChanges && (
+                <AlertDialog>
```
joshenlim: couple more cleanup (i'll hop on) - shift this into its own component to avoid bloating this one

apps/studio/components/interfaces/Auth/EmailTemplates/TemplateEditor.tsx
```diff
@@ -164,9 +206,37 @@ export const TemplateEditor = ({ template }: TemplateEditorProps) => {
...
+    (subjectSlug !== undefined &&
+      authConfig?.MAILER_SUBJECTS_CUSTOM_CONTENTS?.[subjectSlug] === true)
   const hasFormChanges = JSON.stringify(formValues) !== JSON.stringify(baselineValues)
   const hasChanges = hasFormChanges || baselineBodyValue !== bodyValue
 
+  const resetTemplateToDefault = () => {
+    if (!projectRef) return console.error('Project ref is required')
+    if (!templateType) return console.error('Template type is required')
+
+    setIsResettingTemplate(true)
```
joshenlim: don't need a separate state for this - can use isPending from the mutation hook

joshenlim:
@dnywh btw this would be another example for async handling in AlertDialog haha if you recall the thread we had before - if we should bake that behaviour into the UI component

joshenlim:
@dnywh alrighty - code changes wise should be good now, just needs e2e testing when the BE changes are through 🙏 🙂

dnywh:
> @dnywh btw this would be another example for async handling in AlertDialog haha if you recall the thread we had before - if we should bake that behaviour into the UI component

Roger that. Handling in https://github.com/supabase/supabase/pull/45960

ChrisChinchilla (approved):
From a docs perspective ✅

doublethink (approved):

joshenlim (approved):
Verified changes on preview!

PR #45573, opened 2026-05-05
Fix multi selector content not scrollable if rendered in a sheet

joshenlim:
## Context

Realised that MultiSelector's content is not scrollable when rendered within a sheet (e.g Auth policies, Database indexes)

### Explanation from Claude:
- The issue is that Radix Dialog (Sheet) adds @radix-ui/react-remove-scroll which intercepts wheel events. The Popover portal renders outside the Sheet's DOM tree, so the scroll lock blocks wheel events on CommandList.
- The fix is to stop wheel event propagation on the CommandList so it doesn't reach the RemoveScroll handler.

### To test
- [ ] Verify that MultiSelector can be scrollable within a sheet (e.g Auth policies roles) and outside of a sheet (e.g Data API -> Exposed schemas)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scroll wheel propagation in multi-select dropdown to prevent unintended scrolling of parent elements.

* **Updates**
  * Simplified filter component interface by removing an unused configuration property.

### DIFF
--- a/apps/studio/components/grid/components/header/filter/FilterPopoverNew.tsx
+++ b/apps/studio/components/grid/components/header/filter/FilterPopoverNew.tsx
@@ -20,7 +20,6 @@ import { useSqlFilterGenerateMutation } from '@/data/ai/sql-filter-mutation'
 import { useTableEditorTableStateSnapshot } from '@/state/table-editor-table'
 
 export interface FilterPopoverProps {
-  portal?: boolean
   isRefetching?: boolean
 }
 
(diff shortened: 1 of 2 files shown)


apps/studio/components/grid/components/header/filter/FilterPopoverNew.tsx
```diff
@@ -20,7 +20,6 @@ import { useSqlFilterGenerateMutation } from '@/data/ai/sql-filter-mutation'
 import { useTableEditorTableStateSnapshot } from '@/state/table-editor-table'
 
 export interface FilterPopoverProps {
-  portal?: boolean
```
joshenlim, line 23: unrelated clean up - prop no longer used

alaister (approved):
LGTM!

- Tested on preview

dnywh (approved):
Tested on staging

PR #45574, opened 2026-05-05
RLS Tester to include policies that are applied to public

joshenlim:
## Context

For a table that has RLS enabled, but a policy with just `true` for the role `public`
The RLS tester was incorrectly reporting that `anon` doesn't have access

Was happening as we weren't considering policies that apply to the `public` role (which applies to _all_ roles)

## Summary by CodeRabbit

* **Bug Fixes**
  * RLS tester now treats explicitly-public policies as applicable regardless of the impersonated role, improving policy coverage accuracy.
* **Refactor**
  * Consolidated RLS test state computation to improve consistency of access badges and policy messaging.
* **Tests**
  * Added comprehensive tests validating RLS scenarios, badge states, and policy/role messaging.

### DIFF
--- a/apps/studio/components/interfaces/Auth/RLSTester/useTestQueryRLS.ts
+++ b/apps/studio/components/interfaces/Auth/RLSTester/useTestQueryRLS.ts
@@ -128,7 +128,8 @@ export const useTestQueryRLS = () => {
             (x) =>
               x.schema === schema &&
               x.table === table &&
-              x.roles.includes(role?.role ?? '') &&
+              (x.roles.includes(role?.role ?? '') ||
+                (x.roles.length === 1 && x.roles[0] === 'public')) &&
               x.command === data.operation
           )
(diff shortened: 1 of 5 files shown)


apps/studio/components/interfaces/Auth/RLSTester/useTestQueryRLS.ts
```diff
@@ -128,7 +128,8 @@ export const useTestQueryRLS = () => {
             (x) =>
               x.schema === schema &&
               x.table === table &&
-              x.roles.includes(role?.role ?? '') &&
+              (x.roles.includes(role?.role ?? '') ||
+                (x.roles.length === 1 && x.roles[0] === 'public')) &&
```
joshenlim, line 132: this is the main fix

alaister (approved):
LGTM!

- Tested on preview

PR #45575, opened 2026-05-05
Opt to use HoverCard instead of Popover for project status

joshenlim:
## Context

It isn't clear that the project status on the project home page is clickable to show the status of each service of the project
image

Which can cause confusion, especially when starting branches - that the project status is just "Coming up" with no additional information.

Opting to use a hover card instead of a popover to hopefully mitigate this gap

## Release Notes

* **UI Improvements**
  * Enhanced status tooltip interactions with improved hover behavior and customizable delay controls for a better user experience.


alaister (approved):
LGTM!

- Tested on preview

PR #45581, opened 2026-05-05
fix(studio): restore 3xx/4xx log status badge colors

jordienr:
## Problem

After the Tailwind v4 upgrade, the 3xx/4xx HTTP status badge in the Logs explorer rendered with a saturated yellow background and washed-out text, because `bg-amber-300`/`text-amber-1100` no longer resolve as they did under v3.

## Fix

Switch the 3xx/4xx case in `ResponseCodeFormatter` to the existing `bg-warning/10 text-warning` semantic tokens, matching the pattern used by the shared `Badge` warning variant.

## Before
CleanShot 2026-05-05 at 12 13 23@2x

## After
CleanShot 2026-05-05 at 12 12 55@2x

## How to test

- Run `pnpm dev:studio`
- Open any project's Logs explorer (API, Edge Functions, Auth, etc.)
- Find a row with a 3xx or 4xx status code
- Expected: faint yellow background with a darker amber/yellow text, consistent with other warning badges in the app
- Confirm 2xx (brand) and 5xx (red) badges still render as before

## Summary by CodeRabbit

* **Style**
  * Updated the visual styling of response codes in the logs display for improved consistency.

### DIFF
--- a/apps/studio/components/interfaces/Settings/Logs/LogsFormatters.tsx
+++ b/apps/studio/components/interfaces/Settings/Logs/LogsFormatters.tsx
@@ -106,7 +106,7 @@ export const ResponseCodeFormatter = ({ value }: { value: string }) => {
     // 4XX || 3XX responses
     case '4':
     case '3':
-      return <ResponseCodeItem className="bg-amber-300 text-amber-1100">{value}</ResponseCodeItem>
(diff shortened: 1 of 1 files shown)


joshenlim (approved):
Changes LGTM! 🙂

PR #45607, opened 2026-05-05
fix: applying '@supabase/server' sdk api changes

kallebysantos:
Towards FUNC-581

Docs update

Docs using deprecated `@supabase/server` API syntax

## What is the new behavior?

Applying latest API syntax

## Documentation
* Updated Edge Function authentication guides introducing new `auth` mode configuration system with refreshed code examples
* Enhanced documentation for setting up authentication across different access patterns: user-authenticated, service-to-service, and public access
* Clarified security best practices and considerations when disabling authentication validation in function handlers


tomaspozo (approved):
Thanks for driving this update!

PR #45615, opened 2026-05-06
fix(pg-meta): reject trigger creation with empty events

Akash504-ai:

YES

Bug fix

`pgMeta.triggers.create()` allows creating a trigger with an empty `events` array, which generates invalid SQL.

## What is the new behavior?

Adds validation to require at least one trigger event during trigger creation using Zod validation. Also adds a regression test to ensure empty `events` arrays are rejected.

## Additional context

Validation is enforced both at the schema level and at runtime inside `create()`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Trigger creation now enforces validation requiring at least one event to be specified.

* **Tests**
  * Added test coverage to validate rejection of empty event arrays.


Hallidayo:

PR #45625, opened 2026-05-06
fix: role selection 

### LINKED ISSUE #45619: BUG: Edge-Functions test anonymous broken
# Bug report

## Describe the bug

In Supabase Studio, the Edge Function tester sends a `service_role` JWT even when the tester role selector is set to `Anonymous`.

This makes the `Run edge function as a role` selector misleading and unsafe for testing authorization logic, because an anonymous test request is actually executed with elevated `service_role` claims.

## To Reproduce

1. Create or deploy an Edge Function that returns the decoded `Authorization` JWT payload:

```ts
const decodeJwtPayload = (authorization: string | null) => {
  const token = authorization?.replace(/^Bearer\s+/i, '') ?? ''
  const payload = token.split('.')[1]

  if (!payload) return { tokenLooksLikeJwt: false }

  try {
    const padded = payload
      .replace(/-/g, '+')
      .replace(/_/g, '/')
      .padEnd(Math.ceil(payload.length / 4) * 4, '=')

    return { tokenLooksLikeJwt: true, claims: JSON.parse(atob(padded)) }
  } catch (error) {
    return { tokenLooksLikeJwt: true, decodeError: String(error) }
  }
}

Deno.serve((req) => {
  const authorization = req.headers.get('authorization')

  return new Response(
    JSON.stringify(
      {
        authorizationPresent: !!authorization,
        decoded: decodeJwtPayload(authorization),
      },
      null,
      2
    ),
    { headers: { 'content-type': 'application/json' } }
  )
})
```

2. Open Supabase Studio.
3. Go to the project's Edge Functions page.
4. Open the function.
5. Click `Test`.
6. In `Run edge function as a role`, select `Anonymous`.
7. Click `Send Request`.
8. Inspect the returned decoded JWT claims.

Actual result:

```json
{
  "authorizationPresent": true,
  "decoded": {
    "tokenLooksLikeJwt": true,
    "claims": {
      "iss": "supabase",
      "ref": "<project-ref>",
      "role": "service_role",
      "iat": 1776345527,
      "exp": 2091921527
    }
  }
}
```

## Expected behavior

When `Anonymous` is selected in the Edge Function tester, the request should use an anonymous or anon JWT, and the decoded claims should contain:

```json
{
  "role": "anon"
}
```

It should not fall back to the project `service_role` key while the UI says the request is being run as anonymous.

## Screenshots

Not applicable.

## System information

- OS: Windows
- Browser: Chrome
- Version of supabase-js: Not applicable
- Version of Node.js: Not applicable

## Additional context

This appears to be a Supabase Studio UI/state issue in the Edge Function tester. The role selector displays `Anonymous`, but the request authorization header still uses the service role token.

7ttp:
## TL;DR

The edge function tester was sending service role tokens even when anonymous was selected,
Fixed by moving the role context provider to wrap both the selector and the submit handler

## sol:

| Before | After |
|--------|-------|
| Service role JWT sent when Anonymous selected | Anon JWT correctly sent when Anonymous selected |
| Selecting "Anonymous" had no effect, always sent `service_role`  | Selecting "Anonymous" correctly sends it now |

## ref:

- Closes https://github.com/supabase/supabase/issues/45619

## Summary by CodeRabbit

* **Refactor**
  * Internal code structure improvements to enhance maintainability and component organization.

### DIFF
--- a/apps/studio/components/interfaces/Functions/EdgeFunctionDetails/EdgeFunctionTesterSheet.tsx
+++ b/apps/studio/components/interfaces/Functions/EdgeFunctionDetails/EdgeFunctionTesterSheet.tsx
@@ -81,7 +81,18 @@ const FormSchema = z.object({
 
 type FormValues = z.infer<typeof FormSchema>
 
-export const EdgeFunctionTesterSheet = ({ visible, onClose }: EdgeFunctionTesterSheetProps) => {
+export const EdgeFunctionTesterSheet = (props: EdgeFunctionTesterSheetProps) => {
+  const { ref: projectRef } = useParams()
+
+  // [Alaister]: We're using a fresh context here as edge functions don't allow impersonating users.
+  return (
+    <RoleImpersonationStateContextProvider key={`role-impersonation-state-${projectRef}`}>
+      <EdgeFunctionTesterSheetContent {...props} />
+    </RoleImpersonationStateContextProvider>
+  )
+}
+
+const EdgeFunctionTesterSheetContent = ({ visible, onClose }: EdgeFunctionTesterSheetProps) => {
   const { data: org } = useSelectedOrganizationQuery()
   const { ref: projectRef, functionSlug } = useParams()
   const getImpersonatedRoleState = useGetImpersonatedRoleState()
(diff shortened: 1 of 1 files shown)


awaseem:
Yikes totally missed this! Thanks for adding @7ttp

awaseem (approved):
LGTM

PR #45630, opened 2026-05-06
Fix anonymous edge function tester authorization

Purushottam0001:
## Summary

Fixes an issue where the Edge Function tester would send a `service_role` JWT even when the role selector was set to `Anonymous`.

## Changes

* Use anon token when anonymous role is selected
* Preserve service role token for non-anonymous requests
* Remove unintended fallback to service role for anonymous testing

## Why

Previously, if `testAuthorization` was undefined, the tester silently fell back to the service role token, causing misleading and unsafe authorization behavior.

## Summary by CodeRabbit

* **New Features**
  * Enhanced edge function tester with impersonation-aware role-based authorization, allowing tests to run using impersonated roles while preserving existing access-token behavior and custom headers.
* **Chores**
  * Updated example project build metadata with an updated Gradle version/timestamp.


7ttp (commented):
i think this has been resolved in: https://github.com/supabase/supabase/pull/45625 ?

7ttp (commented):
closing this up :) 

feel free to reopen, if the issue persists

PR #44878, opened 2026-04-14
feat: advertise and serve agent skills from .well-known

gregnr:
## Summary

- Adds `apps/www/scripts/fetchAgentSkills.mjs` — at build time (`pnpm content:build`) fetches the latest `index.json` from `supabase/agent-skills` release assets and writes it verbatim to `public/.well-known/agent-skills/index.json`
- Skill URLs are absolute GitHub Release asset URLs, embedded by the agent-skills repo at release time (supabase/agent-skills#87) — no URL rewriting needed on this side

## How it works

1. Fetches latest release from `supabase/agent-skills` via GitHub API
2. Downloads `index.json` from the release assets
3. Writes it verbatim to `public/.well-known/agent-skills/index.json`

Clients discover and install skills from `supabase.com/.well-known/agent-skills/index.json` and fetch tarballs directly from GitHub Release assets.

## Dependency

Requires supabase/agent-skills#87 to be merged and released so the published `index.json` contains absolute URLs.

PR #45641, opened 2026-05-06
feat: serve agent skills from .well-known/agent-skills at build time

Rodriguespn:
## Summary

This PR makes `fetchAgentSkills.mjs` a spec-compliant client of the [agent-skills `.well-known` URI spec](https://github.com/agentskills/agentskills/pull/254), and updates the script to match the current release structure in [`supabase/agent-skills`](https://github.com/supabase/agent-skills).

---

## 1. Spec-compliant URL resolution and digest verification

`fetchAgentSkills.mjs` acts as a client consuming the `.well-known` discovery index. The [agent-skills `.well-known` spec](https://github.com/agentskills/agentskills/pull/254) is explicit on two points:

**URL resolution** — skill artifact URLs in `index.json` must be resolved per [RFC 3986 §5.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.2) using the index URL as the base URI:

> "The `url` field specifies where to fetch the skill artifact. URLs are resolved per RFC 3986 Section 5 using the index URL as the base URI."

This means `skill.url` can be relative (`supabase.tar.gz`), path-absolute (`/.well-known/agent-skills/supabase.tar.gz`), or fully absolute (e.g. a CDN URL like `https://cdn.example.com/supabase.tar.gz`). The previous implementation extracted a filename with `.split('/').pop()` which happened to work for bare relative URLs but was not doing RFC 3986 resolution.

**Digest verification** — clients must verify artifact integrity before use:

> "Clients **must** verify downloaded content against the `digest` in the index. A mismatch indicates the content is corrupted or tampered with — clients **must not** use unverified content."

The updated script uses `new URL(skill.url, githubReleaseIndexUrl)` for compliant resolution, verifies each artifact's SHA-256 digest from the in-memory buffer before any disk writes, and only writes to `public/.well-known/agent-skills/` once all digests pass.

**Acknowledged overhead**: since Supabase owns both the publisher ([`scripts/build-release.ts`](https://github.com/supabase/agent-skills/blob/main/scripts/build-release.ts) in `supabase/agent-skills`) and this consumer, the practical risk of non-compliant URL handling is currently low — the publisher always emits bare relative filenames. However, being spec-compliant here gives us full flexibility to change how skills are packaged or hosted in `supabase/agent-skills` in the future (e.g. moving artifacts to a CDN) without needing to update this script.

---

## 2. Semver release tags

#44878 referenced `supabase/agent-skills#66` (date+SHA tags). [supabase/agent-skills#77](https://github.com/supabase/agent-skills/pull/77) has since merged, moving releases to semver tags managed by Release Please. `/releases/latest` works for both formats — no code change needed, just a rebase.

### DIFF
--- a/apps/www/scripts/fetchAgentSkills.mjs
+++ b/apps/www/scripts/fetchAgentSkills.mjs
@@ -1,18 +1,19 @@
 // @ts-check
 
 /**
- * Fetches the latest agent-skills release from supabase/agent-skills and writes
- * index.json + skill archives to public/.well-known/agent-skills/.
+ * Fetches the latest agent-skills index.json from supabase/agent-skills and
+ * writes it to public/.well-known/agent-skills/index.json with skill URLs
+ * rewritten to absolute GitHub Release URLs.
  *
- * Releases are semver-tagged (e.g. v0.2.0) and managed by Release Please in
- * https://github.com/supabase/agent-skills. Each release includes:
- *   - index.json  (discovery index per agent-skills .well-known spec v0.2.0)
- *   - *.tar.gz    (one archive per skill)
+ * Tarballs are NOT downloaded or hosted here — the index points directly to
+ * GitHub Release assets. The digest in each skill entry is the trust anchor:
+ * clients verify SHA-256 after downloading from GitHub, same pattern as SRI
+ * on the web. The URL is just a location hint; the digest is the guarantee.
  *
+ * Spec: https://github.com/agentskills/agentskills/pull/254
  * Runs unauthenticated — public repo, build-time only.
  */
 
-import { createHash } from 'node:crypto'
 import { promises as fs } from 'node:fs'
 import { dirname, join } from 'node:path'
 import { fileURLToPath } from 'node:url'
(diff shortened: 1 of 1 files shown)


apps/www/scripts/fetchAgentSkills.mjs
```diff
@@ -33,12 +43,31 @@ async function main() {
...
+      const artifactName = skill.url.split('/').pop()
+      const artifactPath = join(OUT_DIR, artifactName)
+      const actual = sha256(await fs.readFile(artifactPath))
+      if (actual !== skill.digest) {
+        throw new Error(
+          `Digest mismatch for ${skill.name}: expected ${skill.digest}, got ${actual}`
+        )
+      }
+      console.log(`  ✓ ${skill.name} digest verified`)
+    }
```
gregnr: nit: we already have all of these files buffered in memory above (`index.json` and the `*.tar.gz` archives as part of `download()`), so we could calculate the checksums earlier and use this loop to verify without re-buffering the files again.

Rodriguespn: Thanks for the review! Good catch on the redundant re-buffering. I've gone a step further than just moving the digest check earlier — I've also reworked the fetching flow so that `fetchAgentSkills.mjs` behaves as a proper spec-compliant client of the [agent-skills `.well-known` URI spec](https://github.com/agentskills/agentskills/pull/254).

The two changes:

**1. RFC 3986-compliant URL resolution**

Each skill artifact URL is now resolved against the index URL as the base URI per [RFC 3986 §5.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.2), using Node's `URL` constructor:

```js
const githubReleaseArtifactUrl = new URL(skill.url, githubReleaseIndexUrl).href
```

The spec allows `skill.url` to be relative (`supabase.tar.gz`), path-absolute, or fully absolute (e.g. a CDN URL). The previous `.split('/').pop()` approach extracted a filename correctly for bare relative URLs but wasn't doing actual RFC 3986 resolution — it would silently misbehave for any other URL form. I explain my take on this increased complexity in the PR description. 

**2. In-memory digest verification before writing to disk**

As you suggested, artifacts are now verified from the in-memory buffer immediately after download — no re-reading from disk. I've also moved `fs.mkdir` and all writes to after the verification loop, so nothing lands on disk if any digest fails:

```js
// download → verify digest → buffer
const buf = await download(githubReleaseArtifactUrl)
const actual = sha256(buf)
if (actual !== skill.digest) throw new Error(...)
buffers[artifactName] = buf

// write only after all digests pass
await fs.mkdir(OUT_DIR, { recursive: true })
for (const [name, buf] of Object.entries(buffers)) {
  await fs.writeFile(join(OUT_DIR, name), buf)
}
```

apps/www/scripts/fetchAgentSkills.mjs
```diff
@@ -21,24 +27,53 @@ async function fetchJson(url) {
...
+  const indexJsonFile = JSON.parse(indexJsonFileBuf.toString('utf8'))
+
+  // Fetch each artifact from its resolved URL, verify digest, buffer before writing
+  const buffers = { 'index.json': indexJsonFileBuf }
+  for (const skill of indexJsonFile.skills ?? []) {
+    // Resolve skill.url against the index URL per RFC 3986 §5.2.2 (https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.2)
+    // as adopted by the agent-skills .well-known spec (https://github.com/agentskills/agentskills/pull/254)
+    // skill.url can be relative ("supabase.tar.gz"), path-absolute, or fully absolute (e.g. a CDN URL)
+    const githubReleaseArtifactUrl = new URL(skill.url, githubReleaseIndexUrl).href
+    const artifactName = new URL(githubReleaseArtifactUrl).pathname.split('/').pop()
```
gregnr: I don't think this complexity is worth it. IIUC, we are guarding against a theoretical scenario where the `skill.url` is not a relative path and instead is path-absolute or a full URL (which realistically is not true - we control the other end). But even so, we're grabbing the last segment of the path and using that to create a relative path on disk, but we're not updating `index.json` to also use this new relative URL.

IMO, if `skill.url` is not relative, this creates complex undefined behaviour on this end that we just shouldn't support. My vote is to revert to the simpler logic, or at most reject non-relative `skill.url`.

apps/www/scripts/fetchAgentSkills.mjs
gregnr: Thanks @Rodriguespn, buffer logic looks good. One more comment [here](https://github.com/supabase/supabase/pull/45641#discussion_r3236268541) w.r.t. the URL logic.

apps/www/scripts/fetchAgentSkills.mjs
Rodriguespn: After reading the spec a couple more times, we might not even need to host the skill tarballs at all. The `url` field can be an absolute path ([according with the spec](https://agent-skills-jh-well-known-uri.mintlify.app/well-known-uri#url-resolution)), and we're already hosting the tarballs on GitHub Releases — so we can just point to them instead of downloading and re-hosting them.

The new `supabase.com/.well-known/agent-skills/index.json` would look like:

```json
{
  "$schema": "https://schemas.agentskills.io/discovery/0.2.0/schema.json",
  "skills": [
    {
      "name": "supabase",
      "type": "archive",
      "description": "Use when doing ANY task involving Supabase...",
      "url": "https://github.com/supabase/agent-skills/releases/download/v0.1.2/supabase.tar.gz",
      "digest": "sha256:807c5a126a4c0c03fa2c96afaf6e9cc02e35f611726d8fde8001e6623fe117a1"
    },
    {
      "name": "supabase-postgres-best-practices",
      "type": "archive",
      "description": "Postgres performance optimization and best practices...",
      "url": "https://github.com/supabase/agent-skills/releases/download/v0.1.2/supabase-postgres-best-practices.tar.gz",
      "digest": "sha256:1bbc12e371a3400fb3e525b530d5ba1673343108e2ea8ae62004d723472df2b9"
    }
  ]
}
```

You'll notice we no longer verify the digest in the build script. That's because we're not acting as a client anymore — we're just pointing to the file location. Digest verification is the client's responsibility, as the spec states explicitly:

> "Clients must verify downloaded content against the `digest` in the index.
> A mismatch indicates the content is corrupted or tampered with — clients
> must not use unverified content."
>
> — [Integrity and verification](https://agent-skills-jh-well-known-uri.mintlify.app/well-known-uri#integrity-and-verification),
> agentskills/.well-known spec PR #254

The security chain: `supabase/agent-skills` CI computes the digests and uploads them to the GitHub release → our build fetches `index.json` over TLS and serves it at `supabase.com/.well-known/agent-skills/index.json` → the client downloads the tarball from GitHub and verifies the digest. Same pattern as SRI on the web.

Let me know your thoughts on about this approach

gregnr: Yep love it. Much simpler. Ship it

PR #45649, opened 2026-05-06
docs: fix CPU time limit documentation

gabrielclaudino-supabase:
Updates the documented CPU time limit from 200ms to the correct value of 2000ms (2 seconds) for Edge Functions in the shutdown guide.

**References:**
- Source: https://github.com/supabase/edge-runtime/blob/main/.cargo/config.toml#L7
- Docs: https://supabase.com/docs/guides/functions/limits#runtime-limits

## Summary by CodeRabbit

* **Documentation**
  * Updated edge function shutdown reasons documentation to reflect an increased CPU time limit of 2000 milliseconds (previously 200 milliseconds).

### DIFF
--- a/apps/docs/content/troubleshooting/edge-function-shutdown-reasons-explained.mdx
+++ b/apps/docs/content/troubleshooting/edge-function-shutdown-reasons-explained.mdx
@@ -51,7 +51,7 @@ These events are surfaced through logs and observability tools, allowing you to 
 
 ### CPUTime
 
-**What it means:** The worker consumed more CPU time than allowed. CPU time measures actual processing cycles used by your code, excluding time spent waiting for I/O or sleeping. Currently limited to 200 milliseconds.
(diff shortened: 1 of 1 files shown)


apps/docs/content/troubleshooting/edge-function-shutdown-reasons-explained.mdx
```diff
@@ -51,7 +51,7 @@ These events are surfaced through logs and observability tools, allowing you to
 
 ### CPUTime
```
jeremenichelli, line 52: Shouldn't this be "CPU Time"? Do you know why it's altogether?

jeremenichelli (approved):

apps/docs/content/troubleshooting/edge-function-shutdown-reasons-explained.mdx
gabrielclaudino-supabase: For the shutdown event, the reason code in the log is written as `CPUTime`.

PR #45664, opened 2026-05-07
docs: Auth method explanations

ChrisChinchilla:

YES

## Summary by CodeRabbit

* **Documentation**
  * Clarified anonymous-account linking guide: example now explicitly verifies the session is anonymous and exits early if not, preventing unintended linking steps.
  * Enhanced resumable uploads guide: added an admonition explaining why the example accesses the local session token to forward it for server-side validation when interacting with direct storage endpoints.

### DIFF
--- a/apps/docs/content/guides/storage/uploads/resumable-uploads.mdx
+++ b/apps/docs/content/guides/storage/uploads/resumable-uploads.mdx
@@ -32,7 +32,7 @@ Instead of `https://project-id.supabase.co` use `https://project-id.storage.supa
   <TabPanel id="js" label="JavaScript">
 
     Here's an example of how to upload a file using `tus-js-client`:
-    
+
     <Admonition type="note">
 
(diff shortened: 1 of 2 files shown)


fadymak (approved):

apps/docs/content/guides/storage/uploads/resumable-uploads.mdx
```diff
@@ -33,6 +33,12 @@ Instead of `https://project-id.supabase.co` use `https://project-id.storage.supa
 
     Here's an example of how to upload a file using `tus-js-client`:
 
+    <Admonition type="note">
+
+    Typically we advise against using `getSession`, because the session is read from local storage and you can't trust its claims for auth decisions. In this case however, the code only needs the raw access token string to forward as a credential to the S3 service, which validates the token server-side. Since no client-side auth decision is made based on the session data, `getSession` is appropriate here.
```
fadymak: I believe this example uploading to Supabase storage, not S3 right?

ChrisChinchilla: Ah sorry, copy pasta, 

```suggestion
    Typically we advise against using `getSession`, because the session is read from local storage and you can't trust its claims for auth decisions. In this case however, the code only needs the raw access token string to forward as a credential to Supabase storage, which validates the token server-side. Since no client-side auth decision is made based on the session data, `getSession` is appropriate here.
```

PR #45667, opened 2026-05-07
feat(studio): apply timezone picker to observability/reports charts

jordienr:
## Problem

The dashboard's timezone picker (#45517) propagates to log timestamps and the shared TimestampInfo component, but observability and reports charts still render their X-axis labels, range labels, and tooltip headers in the browser's local timezone. The result is jarring once a user picks a non-local timezone: hover a chart and you get one tz, hover a log row and you get another.

## Fix

Routes all display-side timestamp formatting in the chart layer through the existing picker-aware helpers (\`useFormatDateTime\` / \`formatDateTime\`) so chart UI matches the rest of the dashboard.

- **ComposedChart.utils** \`CustomTooltip\` (the hotspot — drives every observability dashboard tooltip): reads the active timezone via \`useTimezone\` for both the header label and the formatted timestamp.
- **AreaChart** / **BarChart**: introduce a \`formatChartDate\` helper that honours each component's existing \`displayDateInUtc\` prop, otherwise routes through the picker.
- **ChartBlock**: the two recharts \`labelFormatter\` arrows now close over \`useFormatDateTime\`.
- **ChartHighlightActions**: range labels in the zoom dropdown migrated to the same hook.

Intentionally untouched (must stay UTC):
- \`ChartHandler\` / \`ChartBlock\` \`startDate\`/\`endDate\` (API range params, day boundary).
- \`ChartBlock.tsx:166\` explicit \`.utc()\` data-key normalisation.
- \`useFillTimeseriesSorted\` and friends (range math, no display).

## How to test

- Sign in. Open the avatar dropdown, pick a timezone different from your browser local (e.g. Asia/Tokyo).
- Visit any project, then \`/project/<ref>/reports/database\` (or any \`/observability/...\` page).
- Hover any chart series — the tooltip header should display the chosen IANA name and times in that timezone.
- Click-drag a range on a chart to open the zoom dropdown — start/end labels in the menu should also be in the chosen timezone.
- Switch back to "Auto detect" and confirm everything reverts to browser-local.
- For an AreaChart/BarChart that uses \`displayDateInUtc\` (e.g. some legacy reports), confirm those still render in UTC regardless of picker.

## Summary by CodeRabbit

* **Refactor**
  * Standardized date/time formatting across charts, tooltips, axis labels, header/footer labels, and highlight range labels in Reports and chart components.
  * Switched to a shared, timezone-aware formatter that respects UTC display mode or the selected picker/timezone, ensuring consistent, human-readable timestamps throughout the UI.

### DIFF
--- a/apps/studio/components/interfaces/Reports/ReportBlock/ChartBlock.tsx
+++ b/apps/studio/components/interfaces/Reports/ReportBlock/ChartBlock.tsx
@@ -27,6 +27,7 @@ import {
   useProjectDailyStatsQuery,
 } from '@/data/analytics/project-daily-stats-query'
 import { METRICS } from '@/lib/constants/metrics'
+import { useFormatDateTime } from '@/lib/datetime'
 import { useDatabaseSelectorStateSnapshot } from '@/state/database-selector'
 import type { Dashboards } from '@/types'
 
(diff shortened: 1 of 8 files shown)


alaister (approved):
LGTM!

- Tested on preview

PR #45683, opened 2026-05-07
[FE-3145] feat(studio): use data_api_revoke_default_privileges flag

alaister:
Swap the project-creation revoke from custom `db_sql` over to the new `data_api_revoke_default_privileges` API field. Same behaviour, just delegated to the platform so non-studio flows (branches, CLI, terraform) can apply the same revoke logic — addresses [FE-3145](https://linear.app/supabase/issue/FE-3145/swap-frontend-to-use-revoke-default-privileges-flag).

Backend support landed in supabase/platform#32158 and supabase/platform#32493 (FUP that decoupled the flag from `data_api_use_api_schema`).

**Changed:**
- `apps/studio/data/projects/project-create-mutation.ts` — accepts `dataApiRevokeDefaultPrivileges` and forwards it as `data_api_revoke_default_privileges`
- `apps/studio/pages/new/[slug].tsx` — drops the inline `buildDefaultPrivilegesSql('revoke')` injection in `dbSql`, passes the flag instead
- `apps/studio/pages/integrations/vercel/[slug]/deploy-button/new-project.tsx` — same swap on the Vercel deploy-button flow
- `packages/api-types/types/platform.d.ts` — adds the new field to `CreateProjectBody`

**Preserved:**
- The `dataApiRevokeOnCreateDefault` PostHog flag still gates the default checkbox state and telemetry — only the SQL application changes
- `data_api_use_api_schema: false` stays as-is — projects keep `public` + `graphql_public` exposed, no project-shape change

## To test

- Project creation form (`/new/<org>`):
  - With PostHog flag off: "Automatically expose new tables" defaults to checked → request body has `data_api_revoke_default_privileges: false`
  - Manually uncheck the box → request body has `data_api_revoke_default_privileges: true`, project ends up with revoked default grants on `public`
  - With "Enable Data API" off → `data_api_revoke_default_privileges: false` (no point revoking when nothing's exposed)
- Vercel deploy-button flow (`/integrations/vercel/<slug>/deploy-button`):
  - Same checkbox behaviour as above
  - Migration SQL from the GitHub repo still runs as `db_sql` separately

## Summary by CodeRabbit

* **New Features**
  * Added support for a dedicated `dataApiRevokeDefaultPrivileges` option during project creation.

* **Refactor**
  * Simplified Data API privilege configuration by using a dedicated parameter instead of SQL-based management across project creation flows.


awaseem (approved):
Nice!

PR #45685, opened 2026-05-07
fix: avoid matching postgres casts as SQL parameters

Akash504-ai:

YES

Bug fix

The SQL parameter parser incorrectly treats PostgreSQL casts such as `::text` and `::int` as named parameters because it matches any `:word` pattern using a regex.

Examples of broken behavior:

```sql
SELECT now()::text;
```

was incorrectly detecting `text` as a parameter.

This could also cause `processParameterizedSql()` to throw:

```txt
Missing value for parameter: text
```

when processing valid PostgreSQL SQL containing casts.

## What is the new behavior?

Parameter parsing now ignores PostgreSQL cast syntax by preventing matches where the `:` is preceded by another `:`.

Regression tests were added to ensure:

* PostgreSQL casts are not detected as parameters
* Valid named parameters like `:id` still work correctly
* Cast expressions remain unchanged during parameter substitution

## Additional context

This fixes false-positive parameter detection for valid PostgreSQL syntax.

## Release Notes

* **Bug Fixes**
  * Fixed SQL parameter parsing to correctly handle PostgreSQL type casts (e.g., `::text`, `::int`), ensuring they are not incorrectly treated as query parameters.

* **Tests**
  * Added test coverage for parameter parsing and replacement behavior with PostgreSQL type casts.


apps/studio/lib/sql-parameters.ts
```diff
@@ -45,7 +45,7 @@ export const parseParameters = (sql: string | undefined) => {
   }
 
   // Find all :parameter occurrences and count them
-  const paramRegex = /:(\w+)/g
+  const paramRegex = /(?<!:):([a-zA-Z_]\w*)/g
```
adityakrmishra, line 48: As a student of enterprise architecture, I really appreciate the use of a negative lookbehind here. It's a clean, performant way to disambiguate `:word` from `::word` without needing a full-blown SQL tokenizer for this specific utility. This effectively hardens the parameter replacement logic against standard PostgreSQL syntax—excellent find!

Hallidayo:

PR #45722, opened 2026-05-08
feat(www): consolidated partner intake form on /partners (DEBR-271)

stylessh:

YES

Feature.

The "Become a Partner" CTAs on supabase.com/partners and supabase.com/partners/integrations link out to a HubSpot-hosted form at forms.supabase.com/partner. The form is fragmented across multiple pages, lacks conditional logic, and doesn't capture the qualifying info partnerships need to triage inbound. Linear: [DEBR-271](https://linear.app/supabase/issue/DEBR-271/deploy-new-consolidated-partner-intake-form-on-website).

## What is the new behavior?

- New consolidated partner intake form is rendered inline at the `#become-a-partner` anchor on `/partners`, built on the existing `MarketingForm` component with conditional sections per partner type (Technology / Solutions / Startup / Whitelabel / Events / Other).
- Adds per-provider `sendWhen` gating to the marketing CRM fan-out so that, for example, only Technology Partner submissions are mirrored to Notion while every submission still reaches HubSpot. `evaluateShowWhen` is extracted to a shared module so client and server share the same rule semantics. Also adds a new `checkbox-group` field type for multi-select questions on the Solutions section.
- Migrates `/partners` and `/partners/integrations` to the App Router (the new design + the form together) because `MarketingForm` pulls in server-only modules that Pages Router can't bundle. All pre-existing "forms.supabase.com/partner" CTAs now anchor to `#become-a-partner` instead.

## Additional context

- HubSpot form GUID and the Notion Tech Partner Intake database ID are placeholders (`00000000…`) — flagged with `TODO(DEBR-271)` in `apps/www/components/Partners/PartnerIntakeForm.tsx` and need to be filled before merge.
- The form is single-page with `showWhen`-driven conditional visibility rather than the multi-step UX of the original HubSpot form. If stepping is required we can extend `MarketingForm`.
- forms.supabase.com/partner (a HubSpot-hosted page, not in this repo) is unaffected and tracked separately.


stylessh:
Superseded — committing directly into alan/partners-pages instead.

PR #45761, opened 2026-05-10
fix: foreign key shortcut

### LINKED ISSUE #45759: Cmd+Enter/Ctrl+Enter saves column without applying foreign key in Table Editor
I confirm this is a bug with Supabase, not with my own application.
I confirm I have searched the existing bugs.

When editing a column in the Table Editor and adding a foreign key, pressing Cmd+Enter (macOS) to save causes the column to be saved before the foreign key relationship is saved/applied. The foreign key is silently dropped, resulting in the column being saved without the intended constraint.

To Reproduce:

1. Open the Table Editor and navigate to an existing table that has a text column.
2. Click Edit Column on the text column.
3. In the column editor, navigate to the Foreign Key section and begin configuring a new foreign key relationship.
3. Instead of clicking the Save button with the mouse, press Cmd+Enter (macOS).
4. Observe that the column is saved, but the foreign key constraint has not been saved/applied.

(Using the save buttons returns the correct expected behaviour... this seems to only happen with the keyboard shortcut)

Expected behavior
Pressing Cmd+Enter / Ctrl+Enter should save the foreign key before or without saving the column... I would expect having to cmd + enter twice to save both the foreign key and then the column, producing the same result as clicking the Save button.

Screenshots
N/A
System information

OS: macOS
Browser: Both Safari & Chrome (tested on both)
Version of supabase-js: N/A (Studio UI issue)
Version of Node.js: N/A

7ttp:
## TL:DR 

`cmd + enter` in the foreign key dialog was also saving the parent column too early and dropping the foreign key relation
now this fix stops the parent save shortcut while the foreign key dialog is open so the relation saves properly

## before: 

https://github.com/user-attachments/assets/5a8040a8-d527-424b-8df8-14a794a30c43

## after: 

https://github.com/user-attachments/assets/c2f058e2-e2dd-4730-b7fb-a63b181f2ac8

## ref: 
- closes https://github.com/supabase/supabase/issues/45759

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved foreign key selector state management to properly reset when the column editor becomes visible, ensuring clean state between editing sessions.
  * Enhanced action bar visibility logic to only display when the column editor is active and the foreign key selector is closed, providing a clearer editing interface.

[Review Change Stack](https://app.coderabbit.ai/change-stack/supabase/supabase/pull/45761)


awaseem (approved):
LGTM

awaseem:
approving to unblock, but if you have time @7ttp would be great to get a E2E test. Ping me in slack and I'll get it merged!

PR #45776, opened 2026-05-11
docs: clarify scheduled function Vault key usage

aqilaziz:
## I have read CONTRIBUTING.md

Yes.

Documentation update.

## Current behavior

The scheduled functions guide did not clearly state when Vault setup runs or how the cron job reads the `publishable_key` secret for the Edge Function bearer token.

Fixes #42398.

## New behavior

Clarifies that Vault setup runs once before creating the cron job, that the scheduled job reads the same `publishable_key` secret from Vault, and that the key is not hard-coded in the cron job definition.

## Additional context

Verification:

- `git diff --check`


aqilaziz:
Updated the PR description to follow the repository template sections noted by CodeRabbit.

PR #33510, opened 2025-02-11
[FE-1401] chore: fly postgres deprecation banner

alaister:
## What kind of change does this PR introduce?

Fixes FE-1401

PR #45778, opened 2026-05-11
feat(studio): Fly.io deprecation banner

pamelachia:
## Summary

Adding an in-dashboard banner for the Fly.io May 31 suspension. Banner targets users on a Fly project (or with a Fly project in their currently-selected org) and surfaces a per-project breakdown of what's affected in a dialog. Detection is self-correcting: as soon as the user migrates off Fly, the banner disappears with no follow-up.

Screenshot 2026-05-11 at 5 08 22 PM
Screenshot 2026-05-11 at 3 55 06 PM

## Changes

- **Detection hook** (`useFlyDeprecationProjects`): reads only from already-cached data — `useSelectedProjectQuery` for the current project, plus `useOrgProjectsInfiniteQuery` scoped to the selected org. Zero cross-org fan-out: worst case is one paginated query per session (the same one the project list page already makes).
- **Banner component** (`FlyDeprecationBanner.tsx`): mounted in `AppBannerWrapper`. Dynamic title (primaries / branches / both), dialog lists affected projects with org name, numbered migration steps, links to backup/restore CLI + Dashboard backup + branching docs. List truncates to 5 entries with "…and N more." tail when more are affected.
- **Telemetry**: `fly_deprecation_banner_exposed` and `fly_deprecation_banner_dismissed` events emitted via `useTrack` (auto-injects project + org groups). Properties: `primaryCount`, `branchCount`. CTA click tracking intentionally omitted — migration outcome is measured via warehouse `cloud_provider = 'FLY'` decay.
- **LocalStorage**: dated dismissal key `FLY_DEPRECATION_2026_05_31`; orphan `FLY_POSTGRES_DEPRECATION_WARNING` from PR #33510 removed in the same change so users who dismissed the Feb 2025 banner still see this one.
- **Support contact**: email `success@supabase.io` only (no support ticket link), per Brian's outreach copy in the Linear issues.

## Coverage trade-off

Banner renders on project pages (selected-project check) and pages where the selected org's projects list is cached (org overview, project list). It does **not** render on `/dashboard` home or other pages without org context. Email outreach from GROWTH-817 / GROWTH-819 handles those users. This was a deliberate trade-off to avoid cross-org fan-out load.

## Lifecycle

Banner expires `2026-06-01T00:00:00Z` (right after the May 31 deadline). Stale client bundles stop rendering it without a redeploy. Cleanup PR planned post-deadline to remove the component, hook, localStorage key, and telemetry events.

## Testing

Tested on the Vercel preview with React Query cache overrides to mock a Fly project:

- [x] Banner renders for a user with at least one project where `cloud_provider === 'FLY'`
- [x] Banner does **not** render for a user with no Fly projects
- [x] Banner does **not** render on `/sign-in`
- [x] Title varies by primaries-only / branches-only / both
- [x] Dialog lists affected projects with org name in parens
- [x] Dialog list truncates to 5 with "…and N more." for larger sets
- [x] Migration guide / Dashboard backup / branching links open in a new tab
- [x] Dismiss (×) closes the banner and persists across hard reload (localStorage `fly-deprecation-2026-05-31-dismissed`)
- [x] PostHog receives one `fly_deprecation_banner_exposed` per mount with `primaryCount` + `branchCount` and `$groups.organization` populated
- [x] PostHog receives one `fly_deprecation_banner_dismissed` on close with the same property shape

## Linear

- fixes GROWTH-817
- fixes GROWTH-819

### DIFF
--- a/apps/studio/components/layouts/AppLayout/FlyDeprecationBanner.tsx
+++ b/apps/studio/components/layouts/AppLayout/FlyDeprecationBanner.tsx
@@ -26,9 +26,12 @@ import {
 import { useLocalStorageQuery } from '@/hooks/misc/useLocalStorage'
 
 const BANNER_EXPIRES_AT = new Date('2026-06-01T00:00:00Z')
-const MIGRATION_GUIDE_URL =
+const BACKUP_RESTORE_CLI_URL =
   'https://supabase.com/docs/guides/platform/migrating-within-supabase/backup-restore'
-const SUPPORT_URL = 'https://supabase.com/dashboard/support/new'
+const DASHBOARD_RESTORE_URL =
+  'https://supabase.com/docs/guides/platform/migrating-within-supabase/dashboard-restore'
+const BRANCHING_DASHBOARD_URL = 'https://supabase.com/docs/guides/deployment/branching/dashboard'
+const SUPPORT_EMAIL = 'success@supabase.io'
 
 export const FlyDeprecationBanner = () => {
   const router = useRouter()
@@ -85,10 +88,10 @@ export const FlyDeprecationBanner = () => {
 
   const title =
     primaries.length > 0 && branches.length > 0
-      ? 'Action required: Fly.io projects and branches will be suspended May 31'
+      ? 'Action required: Fly.io project and branch suspensions begin May 31'
       : primaries.length > 0
-        ? 'Action required: Fly.io projects will be suspended May 31'
-        : 'Action required: Fly.io branches will be suspended May 31'
+        ? 'Action required: Fly.io project suspensions begin May 31'
+        : 'Action required: Fly.io branch suspensions begin May 31'
 
   return (
     <HeaderBanner
@@ -114,10 +117,10 @@ const FlyDeprecationDialog = ({
       </DialogTrigger>
       <DialogContent>
         <DialogHeader>
-          <DialogTitle>Fly.io deprecation: May 31, 2026</DialogTitle>
+          <DialogTitle>Fly.io deprecation: suspensions begin May 31, 2026</DialogTitle>
           <DialogDescription>
-            Supabase is sunsetting Fly.io infrastructure. Projects and branches still running on
-            Fly.io will be suspended on May 31, 2026.
+            Supabase will begin suspending projects and branches still running on Fly.io
+            infrastructure on May 31, 2026.
           </DialogDescription>
         </DialogHeader>
 
@@ -128,9 +131,18 @@ const FlyDeprecationDialog = ({
           items={primaries}
           note={
             <>
-              To preserve your data, follow the{' '}
-              <InlineLink href={MIGRATION_GUIDE_URL}>migration guide</InlineLink> to back up and
-              restore onto Supabase's general infrastructure.
+              <p>
+                To preserve your data, migrate each project to Supabase's general infrastructure:
+              </p>
+              <ol className="list-decimal pl-5 space-y-1">
+                <li>
+                  Back up your database using the{' '}
+                  <InlineLink href={BACKUP_RESTORE_CLI_URL}>Supabase CLI</InlineLink> (or take a{' '}
+                  <InlineLink href={DASHBOARD_RESTORE_URL}>Dashboard backup</InlineLink>).
+                </li>
+                <li>Create a new project on Supabase's general infrastructure.</li>
+                <li>Restore the backup into the new project.</li>
+              </ol>
             </>
           }
         />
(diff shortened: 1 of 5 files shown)


packages/common/telemetry-constants.ts
```diff
@@ -3072,6 +3072,36 @@ export interface ComputeBadgeUpgradeClickedEvent {
...
+ * @source studio
+ */
+export interface FlyDeprecationBannerViewedEvent {
+  action: 'fly_deprecation_banner_viewed'
+  groups: TelemetryGroups
+  properties: {
+    primaryCount: number
+    branchCount: number
+  }
+}
```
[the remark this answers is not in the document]
pamelachia: Fixed in 68cbe4e — renamed to `fly_deprecation_banner_exposed` to match the established `_exposed` verb convention used by 6 other events in this file.

apps/studio/components/layouts/AppLayout/FlyDeprecationBanner.tsx
```diff
@@ -0,0 +1,192 @@
...
+  DialogSection,
+  DialogSectionSeparator,
+  DialogTitle,
+  DialogTrigger,
+} from 'ui'
+
+import { HeaderBanner } from '@/components/interfaces/Organization/HeaderBanner'
+import { InlineLink, InlineLinkClassName } from '@/components/ui/InlineLink'
+import { useSendEventMutation } from '@/data/telemetry/send-event-mutation'
+import {
```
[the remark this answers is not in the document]
pamelachia: Fixed in 68cbe4e — switched to `useTrack` which auto-injects project/org groups from selection context.

apps/studio/hooks/misc/useFlyDeprecationProjects.ts
```diff
@@ -0,0 +1,49 @@
...
+  orgName: string
+}
+
+export function useFlyDeprecationProjects({ enabled }: { enabled: boolean }) {
+  const { data: organizations } = useOrganizationsQuery({ enabled })
+
+  const orgProjectsQueries = useQueries({
+    queries: (organizations ?? []).map((org) => ({
+      queryKey: projectKeys.bannerProjectsByOrg(org.slug),
+      queryFn: () => getOrganizationProjects({ slug: org.slug, limit: 100 }),
```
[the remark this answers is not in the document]
pamelachia: Not changing for this PR. The hook runs in `AppBannerWrapper` (mounted in `DefaultLayout`), so paginating means N orgs × multiple pages on every dashboard mount. For a 2-week deprecation campaign, missing a Fly project in an org with >100 projects is an acceptable miss — those users are also in the email outreach list from GROWTH-817/819. Documenting as a known limitation in the post-deadline cleanup ticket.

apps/studio/components/layouts/AppLayout/FlyDeprecationBanner.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+export const FlyDeprecationBanner = () => {
+  const router = useRouter()
+
+  const [acknowledged, setAcknowledged, { isSuccess }] = useLocalStorageQuery(
+    LOCAL_STORAGE_KEYS.FLY_DEPRECATION_2026_05_31,
+    false
+  )
+
+  const isExpired = Date.now() >= BANNER_EXPIRES_AT.getTime()
```
seanoliver, line 44: Nit: `pathname.includes('sign-in')` is a little brittle since it's a substring match (`/sign-in-anything` or a future route with that token would also suppress the banner). `router.pathname === '/sign-in' is a little safer. Low/no risk today since no such routes exist; flagging in case `useSelectedOrganizationQuery` ever resolves on an unauth path.

apps/studio/components/layouts/AppLayout/FlyDeprecationBanner.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const track = useTrack()
+
+  const exposedRef = useRef(false)
+  useEffect(() => {
+    if (!shouldEvaluate || !isReady || !hasFlyResources || exposedRef.current) return
+    exposedRef.current = true
+    track('fly_deprecation_banner_exposed', {
+      primaryCount: primaries.length,
+      branchCount: branches.length,
+    })
```
seanoliver, line 62: `AppBannerWrapper` is mounted in both `DefaultLayout` and `AuthenticationLayout`. Navigating between routes that use different layouts will remount this component and re-fire `fly_deprecation_banner_exposed` (the `exposedRef` only dedupes within a mount). For counting unique exposed users that's fine, but it inflates raw event count — worth noting if you plan to read totals directly. Otherwise no change needed.

apps/studio/hooks/misc/useFlyDeprecationProjects.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export function useFlyDeprecationProjects({ enabled }: { enabled: boolean }) {
+  const { data: selectedProject } = useSelectedProjectQuery()
+  const { data: selectedOrg } = useSelectedOrganizationQuery()
+
+  const orgSlug = selectedOrg?.slug
+  const orgName = selectedOrg?.name ?? ''
+
+  const { data: orgProjectsData, isFetched } = useOrgProjectsInfiniteQuery(
+    { slug: orgSlug },
```
seanoliver, line 21: This calls `useOrgProjectsInfiniteQuery` but never advances pages so only the first 96 projects are inspected. For an org with >96 projects, Fly resources on later pages won't be counted (and won't surface in the dialog list). The `selectedProject` path still catches the case where they're sitting on one, so it's a partial backstop. Probably fine most of the time, but could be an edge case worth solving for now!

seanoliver (approved):

apps/studio/hooks/misc/useFlyDeprecationProjects.ts
pamelachia: it fans out if we run for all projects and was worried it'll cause problems for sfp orgs which is why just handled for the cheapest use case

apps/studio/components/layouts/AppLayout/FlyDeprecationBanner.tsx
pamelachia: Fixed in 78add72 — switched to `pathname.startsWith('/sign-in')`. Covers all 4 sign-in routes (`/sign-in`, `/sign-in-sso`, `/sign-in-mfa`, `/sign-in-partner`) and rules out the future-substring-match case.

awaseem (approved):
LGTM

PR #45783, opened 2026-05-11
Refactor auth email template variables

joshenlim:
## Context

Reworks the template variables section of an email template a little
- Add a section header with docs button
  image
- Refactor `AuthTemplatesValidation` to remove unnecessary description parsing logic in `TemplateEditor`
- Update tooltip content to match [docs](https://supabase.com/docs/guides/local-development/customizing-email-templates#template-variables)
  - e.g `SiteURL`: Add inline link for configuration
    image
  - e.g `Token`: Mention that its swappable with `ConfirmationURL` if both parameters are available for use in the template (Counter example is the reauthentication template which there's only `Token` variable available)
    image
    image

### DIFF
--- a/apps/studio/components/layouts/AuthLayout/AuthEmailsLayout.tsx
+++ b/apps/studio/components/layouts/AuthLayout/AuthEmailsLayout.tsx
@@ -23,7 +23,7 @@ export const AuthEmailsLayout = ({ children }: PropsWithChildren<{}>) => {
   ]
 
   return (
-    <AuthLayout title="Email">
+    <AuthLayout title="Emails">
       {showEmails ? (
         <PageLayout
           title="Emails"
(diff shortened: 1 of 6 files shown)


apps/studio/components/layouts/AuthLayout/AuthEmailsLayout.tsx
```diff
@@ -23,7 +23,7 @@ export const AuthEmailsLayout = ({ children }: PropsWithChildren<{}>) => {
   ]
 
   return (
-    <AuthLayout title="Email">
+    <AuthLayout title="Emails">
```
dnywh, line 26: Just making the sidebar route match the naming on the page. It currently says "Email" when the page is called "Email**s**"

PR #45784, opened 2026-05-11
fix(studio): pre-empt React 19 regressions in tests + Support form

alaister:
Four React-19-sensitive patterns that pass on React 18 today but break under React 19 (verified on the in-flight TanStack Start branch). Landing on master now so the eventual React 19 upgrade is a no-op for tests, instead of a separate cleanup pass under upgrade pressure.

Each fix is a strict superset / less-fragile equivalent of the existing pattern, so master (React 18) stays green.

**Changed:**
- `hooks/misc/useStateTransition.ts` — fire on entry into `newTest` from any state other than `newTest`, instead of requiring exactly `prevTest → newTest`. React 18+ auto-batches dispatches across awaits (e.g. `dispatch SUBMIT` in the handler, `dispatch ERROR` in `onError`), collapsing `editing → submitting → error` into a single render where the intermediate `submitting` tick is never observed. Strict superset of the old check for our reducers — `success`/`error` are only reachable from `submitting`.
- `Support/CategoryAndSeverityInfo.tsx` — guard `onValueChange` against Radix Select's spurious `''` emission. When the controlled value transitions from `undefined` to a defined value whose `SelectItem` isn't mounted yet (dropdown closed → items haven't registered), Radix's hidden `BubbleSelect` fires `onValueChange('')` and clobbers the field. No `SelectItem` can have `value=""` (Radix throws), so any `''` is guaranteed spurious — drop it before calling `field.onChange`. ([radix-ui/primitives#3381](https://github.com/radix-ui/primitives/issues/3381))
- `EditSecretModal.test.tsx` — `getByLabelText` → `findByLabelText`. Under React 19's scheduling, the decrypted-value query resolves on a separate render tick, so form fields appear one tick after the skeleton.
- `LogsPreviewer.test.tsx` — `addEventListener('click', spy)` instead of `loadOlder.onclick = vi.fn()`. React 19 reassigns `.onclick` on managed elements as part of its event wiring, clobbering the direct-property spy.

### Unit tests
- `pnpm --filter studio test` — all unit tests pass on master (React 18)

### Support form URL prefill (Radix Select guard)
- `/support/new?category=Problem` → category dropdown reads "APIs and client libraries" on first paint
- `/support/new?category=dashboard_bug` → "Dashboard bug" (case-insensitive match)
- `/support/new?category=invalid_garbage` → falls back to "Select an issue" placeholder, no crash
- `/support/new?subject=My%20issue&message=Details%20here` → subject and message inputs are prefilled
- `/support/new?projectRef=<your-ref>&category=Problem` → both project selector and category set, library selector appears
- With a prefilled URL, click the category dropdown and pick a different option — the new value sticks (this is the path that surfaced the Radix bug, want to confirm we didn't break user selection)
- DevTools console on first load should be clean — no React hydration mismatch warning

### Support form submit (`useStateTransition` success + error branches)
- Submit a valid support form → green toast "Support request sent" appears **once**, view swaps to the success screen, one `POST /platform/feedback/send` in the network panel
- Block `POST /platform/feedback/send` in DevTools → submit → red error toast appears **once** (not twice — if you see two toasts the relaxed transition is firing more than it should), form stays editable with all inputs preserved
- Unblock and submit again → success path runs cleanly

### Sidebar support form (same reducer + `useStateTransition`, separate component)
- Open the support widget in the side nav (`SupportSidebarForm`)
- Repeat the success and error paths — should behave identically

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed category selector to prevent selected values from being unexpectedly cleared during form interactions.

* **Tests**
  * Improved test reliability for modal field rendering and event handling assertions.

* **Chores**
  * Clarified internal comments for form initialization logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/supabase/supabase/pull/45784)

### DIFF
--- a/apps/studio/components/interfaces/Integrations/Vault/Secrets/__tests__/EditSecretModal.test.tsx
+++ b/apps/studio/components/interfaces/Integrations/Vault/Secrets/__tests__/EditSecretModal.test.tsx
@@ -79,9 +79,6 @@ describe(`EditSecretModal`, () => {
 
     await screen.findByRole(`dialog`)
 
-    // The decrypted-value query resolves on a separate render tick from the
-    // vault list query that opens the dialog. Wait for the form fields to
-    // replace the skeleton before reading them.
     const nameInput = await screen.findByLabelText(`Name`)
     const descriptionInput = screen.getByLabelText(`Description`)
     const valueInput = screen.getByLabelText(`Secret value`)
--- a/apps/studio/components/interfaces/Integrations/Vault/Secrets/__tests__/EditSecretModal.test.tsx
+++ b/apps/studio/components/interfaces/Integrations/Vault/Secrets/__tests__/EditSecretModal.test.tsx
@@ -79,7 +79,10 @@ describe(`EditSecretModal`, () => {
 
     await screen.findByRole(`dialog`)
 
-    const nameInput = screen.getByLabelText(`Name`)
+    // The decrypted-value query resolves on a separate render tick from the
+    // vault list query that opens the dialog. Wait for the form fields to
+    // replace the skeleton before reading them.
+    const nameInput = await screen.findByLabelText(`Name`)
     const descriptionInput = screen.getByLabelText(`Description`)
     const valueInput = screen.getByLabelText(`Secret value`)
     const togglePasswordButton = screen.getByRole(`button`, { name: `Show secret value` })
--- a/apps/studio/hooks/misc/useStateTransition.ts
+++ b/apps/studio/hooks/misc/useStateTransition.ts
@@ -6,7 +6,10 @@ export function useStateTransition<
   NewType extends State['type'],
 >(
   state: State,
-  prevTest: PrevType,
+  // Kept in the signature for documentation / type-narrowing of `cb`'s first
+  // arg, but the entry-detection logic below no longer uses it. See the
+  // comment in the effect for why.
+  _prevTest: PrevType,
   newTest: NewType,
   cb: (
     prevState: Extract<State, { type: PrevType }>,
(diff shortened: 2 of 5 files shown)


apps/studio/components/interfaces/Integrations/Vault/Secrets/__tests__/EditSecretModal.test.tsx
```diff
@@ -79,7 +79,7 @@ describe(`EditSecretModal`, () => {
 
     await screen.findByRole(`dialog`)
 
-    const nameInput = screen.getByLabelText(`Name`)
+    const nameInput = await screen.findByLabelText(`Name`)
```
alaister, line 82: Switched `getByLabelText` → `findByLabelText` here only — under React 19's scheduling the vault list query opens the dialog one tick before the decrypted-value query resolves and the form fields replace the skeleton. The remaining inputs stay on `getBy` because once Name is mounted the rest are in the same render. On React 18 the `findBy` resolves immediately, so this is invisible there.

apps/studio/tests/features/logs/LogsPreviewer.test.tsx
```diff
@@ -132,11 +132,12 @@ test('can click load older', async () => {
     { timeout: 10000 }
   )
 
-  loadOlder.onclick = vi.fn()
+  const handleClick = vi.fn()
+  loadOlder.addEventListener('click', handleClick)
```
alaister, line 136: Switched from `loadOlder.onclick = vi.fn()` to `addEventListener('click', spy)` because React 19 reassigns `.onclick` on managed elements as part of its event wiring, clobbering the direct-property spy before the assertion runs. `addEventListener` isn't touched by React on either version, so the test is equivalent on React 18 and survives React 19.

alaister (commented):
Some notes:

apps/studio/hooks/misc/useStateTransition.ts
```diff
@@ -18,13 +21,24 @@ export function useStateTransition<
...
+    // Treat prevTest as advisory only: as long as the previous state was *not*
+    // already `newTest`, fire the callback.
+    if (savedPrevState.type !== newTest && state.type === newTest) {
+      // Pass the previous state through even if it doesn't match prevTest —
+      // callers shouldn't rely on it being prevTest in batched flows. Cast is
+      // safe at the type level because callers only consume `currState`.
       cb(
         savedPrevState as Extract<State, { type: PrevType }>,
         state as Extract<State, { type: NewType }>
       )
```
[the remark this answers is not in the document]
alaister: Good catch — fixed in 76389dbcaf. Went with the second option (widen `cb`'s `prevState` type) rather than the first (restoring `=== prevTest` to the guard), because the strict `=== prevTest` check is exactly what fails under React 18+ auto-batching: when `dispatch SUBMIT` collapses with the mutation's `dispatch ERROR`, `savedPrevState.type` is the pre-batch state (`'editing'`), not `prevTest` (`'submitting'`). Re-adding that check would silently swallow the transition the previous commit was fixing.

Diff matches what you proposed: `prevState: Extract<State, { type: PrevType }>` → `Exclude<State, { type: NewType }>`, and the cast at the call site is now an accurate assertion matching the runtime guard rather than a fiction. Both call sites (`SupportFormPage`, `SupportSidebarForm`) ignore the first arg, so the widening is a no-op at the consumer.

djhi (approved):
💯 comments!

PR #45803, opened 2026-05-11
fix(billing): block downgrade plan requests

pamelachia:
## Summary
Blocks the frontend from submitting downgrade or no-op plan-change requests to billing owners. `RequestUpgradeToBillingOwners` now checks the requested plan against the viewer's current plan and renders a disabled "Not available" state when the request would be a downgrade or a no-op, so the backend `requested_plan` enum (Pro/Team/Enterprise only) never sees a Free submission. Centralising the guard inside the component also catches three pre-existing wrong-tier surfaces (log-drains empty state, billing-metric paywall, custom-auth-providers fallback) where higher-tier members were seeing "Request upgrade to Pro" buttons.

## Changes
- Added a plan-direction guard inside `RequestUpgradeToBillingOwners` that renders a disabled `ButtonTooltip` when the requested plan is not an upgrade from the current plan
- Skipped the guard for addon-flavored requests (compute size, PITR, custom domain, IPv4, spend cap) since they target feature add-ons, not plan tiers
- Widened the `plan` prop type to include `'Free'` so the plan-switcher can pass it honestly; cast back to `PlanRequest` at the submission and telemetry call sites where the guard ensures safety
- Updated the `PlanUpdateSidePanel` callsite to pass the wider literal union

## Testing
Still verifying on the Vercel preview — unchecked items below are pending. Reviewers, please leave the PR un-merged until the boxes are checked:
- [x] Plan switcher as a non-billing-owner on Pro: Free card shows "Not available" + tooltip, Pro shows "Current plan", Team shows "Request upgrade to Team"
- [x] Plan switcher as a non-billing-owner on Team: Free and Pro both show "Not available", Team shows "Current plan"
- [x] Network tab: zero POSTs to `/platform/organizations/{slug}/billing/upgrade-request` when interacting with the disabled state
- [x] Hover the "Not available" button: tooltip appears with the "Ask a billing owner to change the plan directly" copy

If the tooltip does not fire on hover, Radix is blocking pointer events on the disabled trigger. Fix by wrapping the disabled button with `<span className="inline-flex w-full">` inside `ButtonTooltip`.

## Linear
- fixes GROWTH-833

### DIFF
--- a/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
+++ b/apps/studio/components/interfaces/Organization/BillingSettings/Subscription/PlanUpdateSidePanel.tsx
@@ -308,7 +308,10 @@ export const PlanUpdateSidePanel = () => {
                         Current plan
                       </Button>
                     ) : !canUpdateSubscription ? (
-                      <RequestUpgradeToBillingOwners block plan={plan.name as 'Pro' | 'Team'} />
+                      <RequestUpgradeToBillingOwners
+                        block
+                        plan={plan.name as 'Free' | 'Pro' | 'Team'}
+                      />
                     ) : (
                       <ButtonTooltip
                         block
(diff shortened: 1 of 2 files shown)


joshenlim:
Closing in favor of a simpler implementation [here](https://github.com/supabase/supabase/pull/45823)

PR #45814, opened 2026-05-12
feat(studio): move CLI login to connect interstitial

dnywh:

Feature / UI refactor

The CLI browser login route still uses the older API authorisation layout and redirects missing or failed sign-in session states to generic 404/500 pages.

## What is the new behaviour?

Moves `/cli/login` onto the shared connect interstitial layout as the next small stacked slice after the organisation invite work.

This keeps the real CLI login contract intact while updating the surface:
- creates the CLI login session from `session_id`, `public_key`, and optional `token_name`
- redirects to the generated `device_code`
- renders missing-parameter and session-creation failures in-card instead of redirecting away
- keeps the 8-character verification code selectable and copyable as a single string
- uses a full-width primary `Copy code` action

This also adds the small shared interstitial helpers needed by this surface and adjusts `CopyButton` so the copied check icon inherits the primary button colour instead of turning green.

This also removes the CLI version admonition:

> Browser login flow requires Supabase CLI version 1.219.0 and above.

I checked with our stats and the CLI team. The vast majority of users are on a newer version.

| Before | After |
| --- | --- |
| Authorize API access  Supabase-D1E3CF26-BD59-4BB2-B457-B552EE47E3DA | Authorize CLI  Supabase-C9977F21-88B8-441B-8A2C-09A9515935B0 |
| Supabase-F42FBEAF-F74D-4920-8A51-7C25004F66D5 | Authorize CLI  Supabase-8159A1B1-2594-4183-AC35-FEF1EFD4EA37
| Supabase-2506E468-9F42-44B9-A5B7-BC4D3777F552 | Authorize CLI  Supabase-A0EE1239-A345-427C-9CF7-997037A8FC0E |
| Authorize API access  Supabase-A7B84CA6-D230-4C3E-9227-DE21CE35375C | Authorize CLI  Supabase-F55E26B2-609B-449C-9C64-08AA90AE3D1E |

## Testing instructions

Use the Vercel preview URL for this PR once it is available. The examples below use `<preview-origin>` as a placeholder, for example `https://studio-git-dnywh-feat-cli-login-interstitial-supabase.vercel.app`.

You need to be signed in to Studio to see these states because `/cli/login` is still behind `withAuth`.

Ready state:
- Open `<preview-origin>/cli/login?device_code=ABCD1234`
- Check the page title is `Authorize CLI | Supabase`
- Check the card title is `Authorize Supabase CLI`
- Check the code fills the width, uses the normal sans font, and can be selected
- Drag-select the code and copy it; the clipboard should contain `ABCD1234`, not one character per line
- Click `Copy code`; the button should show the usual copied success state without a green check icon on the primary button

Missing parameters state:
- Open `<preview-origin>/cli/login`
- Check the card says `Missing sign-in parameters` and names the missing `session_id` and `public_key` parameters
- Open `<preview-origin>/cli/login?session_id=session-test`
- Check it still stays in-card and names the missing `public_key` parameter instead of redirecting to `/404`

Creation error state:
- Open `<preview-origin>/cli/login?session_id=not-real&public_key=not-real&token_name=local-dev`
- Check it stays in-card with `Unable to create CLI sign-in` instead of redirecting to `/500`
- The exact error detail can vary by environment; the important bit is that the failure is shown inside the interstitial card

Loading state:
- This is transient because there are no production mocks in this slice
- To inspect it manually, throttle the browser network before opening a session-creation URL such as `<preview-origin>/cli/login?session_id=not-real&public_key=not-real`

Real CLI flow:
- Run the browser login flow from Supabase CLI as usual
- When the CLI opens a Studio URL, keep the path and query string but replace the origin with the PR preview origin
- The page should create the login session and then route to `/cli/login?device_code=<8 character code>`
- Enter that 8-character code back in the CLI prompt

## Summary by CodeRabbit

* **New Features**
  * Redesigned CLI login flow with clearer state-driven screens and improved verification UI.
  * Added a small paired-logo component for centered logo pairs with a connector icon.

* **Improvements**
  * Copy button behavior and styling refined for consistent visual feedback across variants.

* **Tests**
  * New unit tests covering copy-button behavior and multiple CLI login UI flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/supabase/supabase/pull/45814)

### DIFF
--- a/apps/studio/components/layouts/InterstitialLayout.tsx
+++ b/apps/studio/components/layouts/InterstitialLayout.tsx
@@ -1,4 +1,5 @@
 import { motion } from 'framer-motion'
+import { ArrowRightLeft } from 'lucide-react'
 import type { PropsWithChildren, ReactNode } from 'react'
 import { Card, CardContent, CardHeader, cn } from 'ui'
 
@@ -85,6 +86,14 @@ export const LogoBox = ({ children, className }: { children: ReactNode; classNam
     )}
   >
     {children}
+  </div>
+)
+
+export const LogoPair = ({ left, right }: { left: ReactNode; right: ReactNode }) => (
+  <div className="flex items-center justify-center gap-2.5">
+    {left}
+    <ArrowRightLeft className="size-4 text-foreground-muted" />
+    {right}
   </div>
 )
 
--- a/apps/studio/components/ui/CopyButton.tsx
+++ b/apps/studio/components/ui/CopyButton.tsx
@@ -31,6 +31,9 @@ const CopyButton = forwardRef<HTMLButtonElement, CopyButtonProps>(
       onClick,
       copyLabel = 'Copy',
       copiedLabel = 'Copied',
+      type = 'primary',
+      icon,
+      className,
       ...props
     },
     ref
@@ -53,9 +56,17 @@ const CopyButton = forwardRef<HTMLButtonElement, CopyButtonProps>(
           onClick?.(e)
         }}
         {...props}
-        className={cn({ 'px-1': iconOnly }, props.className)}
+        type={type}
+        className={cn({ 'px-1': iconOnly }, className)}
         icon={
-          showCopied ? <Check strokeWidth={2} className="text-brand" /> : (props.icon ?? <Copy />)
+          showCopied ? (
+            <Check
+              strokeWidth={2}
+              className={cn(type === 'primary' ? 'text-inherit' : 'text-brand')}
+            />
+          ) : (
+            (icon ?? <Copy />)
+          )
         }
       >
         {!iconOnly && <>{children ?? (showCopied ? copiedLabel : copyLabel)}</>}
--- a/apps/studio/pages/cli/login.tsx
+++ b/apps/studio/pages/cli/login.tsx
@@ -30,6 +30,24 @@ const CliLogo = () => (
   </LogoBox>
 )
 
+const CliLoginInterstitial = ({
+  title,
+  description,
+  children,
+}: {
+  title: ReactNode
+  description?: ReactNode
+  children: ReactNode
+}) => (
+  <InterstitialLayout
+    logo={<LogoPair left={<CliLogo />} right={<SupabaseLogo />} />}
+    title={title}
+    description={description}
+  >
+    <div className="px-6 pb-6">{children}</div>
+  </InterstitialLayout>
+)
+
 const CliLoginPage: NextPageWithLayout = () => {
   const router = useRouter()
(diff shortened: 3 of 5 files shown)


apps/studio/components/layouts/InterstitialLayout.tsx
```diff
@@ -88,6 +89,14 @@ export const LogoBox = ({ children, className }: { children: ReactNode; classNam
...
 )
 
+export const LogoPair = ({ left, right }: { left: ReactNode; right: ReactNode }) => (
+  <div className="flex items-center justify-center gap-2.5">
+    {left}
+    <ArrowRightLeft className="size-4 text-foreground-muted" />
+    {right}
+  </div>
+)
+
```
dnywh, line 99: Reusable for future connection surfaces, like the app authorization step.

apps/studio/components/ui/CopyButton.tsx
dnywh, line 1: General changes. The green check/tick was not legible on a green button.

| After |
| --- |
| Image |

apps/studio/pages/cli/login.tsx
```diff
@@ -1,88 +1,264 @@
...
+    children: ReactNode
+  }) => (
+    <InterstitialLayout
+      logo={<LogoPair left={<CliLogo />} right={<SupabaseLogo />} />}
+      title={title}
+      description={description}
+    >
+      <div className="px-6 pb-6">{children}</div>
+    </InterstitialLayout>
   )
```
djhi: nit but I don't really get the point of having a function to render this layout. Why not simply wrap each return in the component directly?

djhi (approved):

apps/studio/pages/cli/login.tsx
dnywh: It’s abstracted because there are ~4 states to render. Can just repeat `<InterstitialLayout>` if that’s preferred?

dnywh: - loading
- missing params
- error
- ready with verification code

djhi: I mean, JSX is already an abstraction over simple functions, we probably don't need another one. Besides, it only saves you 7 lines. If you don't want to repeat the `div` inside `InterstitialLayout`, you should probably introduce a new component instead of a function. That would be better for React rerenders too

dnywh: Fair enough. Fixed in https://github.com/supabase/supabase/pull/45814/commits/c6e498357169d9b9feb7f68549e2bbbe958bd41a

PR #45828, opened 2026-05-12
fix: logs explorer field references doesn't show all sources

djhi:
## Problem

The field reference side panel doesn't show all sources: there are actually 11 items and users can't see them nor select them.

## Solution

Use a combobox instead

## Screenshots

Before:
image

After:
image
image

## Summary by CodeRabbit

* **New Features**
  * Schema selector in the Logs Query Panel is now a searchable popover and shows the selected source in the button (or prompts “Select source...”).
  * Field table now displays fields for the chosen schema only.

* **Refactor**
  * UI simplified for faster schema selection and clearer field reference browsing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/supabase/supabase/pull/45828)

### DIFF
--- a/apps/studio/components/interfaces/Settings/Logs/LogsQueryPanel.tsx
+++ b/apps/studio/components/interfaces/Settings/Logs/LogsQueryPanel.tsx
@@ -1,18 +1,27 @@
 import { IS_PLATFORM } from 'common'
-import { BookOpen, Check, ChevronDown, Copy, ExternalLink, X } from 'lucide-react'
+import { BookOpen, Check, ChevronDown, ChevronsUpDown, Copy, ExternalLink, X } from 'lucide-react'
 import Link from 'next/link'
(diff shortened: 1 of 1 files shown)


apps/studio/components/interfaces/Settings/Logs/LogsQueryPanel.tsx
```diff
@@ -220,38 +232,66 @@ const LogsQueryPanel = ({
...
+                head={[
+                  <Table.th className="text-xs p-2!" key="path">
+                    Path
+                  </Table.th>,
+                  <Table.th key="type" className="text-xs p-2!">
+                    Type
+                  </Table.th>,
+                ]}
+                body={selectedSchema.fields
+                  .sort((a: any, b: any) => a.path - b.path)
```
jordienr: old code but can use string types here instead

jordienr (approved):

PR #45385, opened 2026-04-29
fix(studio): encode special characters in database advisor lint links

oniani1:
The link builders in `apps/studio/components/interfaces/Linter/Linter.utils.tsx` interpolate `metadata.schema` and `metadata.name` directly into URL query strings. A schema or table name with `&`, `=`, `+`, or `#` breaks the destination filter on the linked page because `URLSearchParams` stops at the bare `&` and decodes `+` to a space.

The `public_bucket_allows_listing` lint at line 338 already wraps `bucket_id` in `encodeURIComponent`. The other 15 builders did not. This wraps each `metadata?.schema` and `metadata?.name` interpolation with `encodeURIComponent(value ?? '')` to match.

Added `Linter.utils.test.tsx` that constructs links with a schema `a&b=c` and a name `d e+f` and asserts `URLSearchParams` round-trips them. The bucket precedent is also covered.

Closes #45384

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL encoding for navigation links in the linter interface to ensure proper handling of special characters in database, schema, and table names.

* **Tests**
  * Added test coverage for URL generation functionality in the linter utility.

PR #45846, opened 2026-05-12
fix(studio): encode special characters in table editor policy links

### LINKED ISSUE #45845: Bug: special characters in table name or schema break links from table editor to RLS policies
### Bug

`apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx` interpolates `table.name` and `table.schema` directly into the policies URL at two places (the `Add RLS policy` button at line 233 and the `RLS policies` count button at line 263):

```tsx
href={`/project/${projectRef}/auth/policies?search=${table.name}&schema=${table.schema}`}
```

If `table.name` or `table.schema` contains `&`, `=`, `+`, or `#`, the destination link is corrupted. `URLSearchParams` stops at the bare `&`, decodes `+` to a space, and treats `=` as the key-value separator. The user lands on a policies page filtering on the wrong value.

Postgres allows these characters in identifiers when quoted, so a real table or schema like `user_data&secret=1` triggers this.

### Reproduction

```js
const name = 'user_data&secret=1'
const url = `/project/abc/auth/policies?search=${name}&schema=public`
const params = new URLSearchParams(url.split('?')[1])
params.get('search')  // 'user_data'   <- wrong, should be the full name
params.get('secret')  // '1'           <- injected, not part of the original URL
params.get('schema')  // 'public'
```

After wrapping with `encodeURIComponent`:

```js
const fixed = `/project/abc/auth/policies?search=${encodeURIComponent(name)}&schema=public`
const params = new URLSearchParams(fixed.split('?')[1])
params.get('search')  // 'user_data&secret=1'  <- correct
params.get('schema')  // 'public'
```

### Expected

`search` and `schema` should round-trip through `URLSearchParams` regardless of the characters in the table name or schema.

### Note

Same class of bug as #45385 (already fixed for the Linter shortcut links). Other Studio surfaces (project secure landing, trigger/function lists, entity sidebar, vault link, etc.) have the same pattern and will need separate fixes.

oniani1:
Closes #45845.

## Summary

`GridHeaderActions.tsx` interpolated `table.name` and `table.schema` directly into the policies URL at two `<Link href>` builders. A table or schema containing `&`, `=`, `+`, or `#` corrupted the destination and routed users to the wrong policies filter.

Extracts the URL into `getTablePoliciesUrl` in `TableEntity.utils.ts` with `encodeURIComponent` wraps, and replaces both inline interpolations. Same pattern as #45385 (Linter shortcut links).

## Test plan

Added four `getTablePoliciesUrl` cases in `TableEntity.utils.test.ts`: plain values, special chars in name, special chars in schema, special chars in both. Existing seven tests in the same file still pass.

## Summary by CodeRabbit

* **Refactor**
  * Improved Row-Level Security (RLS) policy URL handling in the table editor using a shared utility function for consistent URL building and proper parameter encoding.

* **Tests**
  * Added test coverage for RLS policy URL generation with various parameter combinations and special character handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/supabase/supabase/pull/45846)

### DIFF
--- a/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
+++ b/apps/studio/components/interfaces/TableGridEditor/GridHeaderActions.tsx
@@ -31,7 +31,10 @@ import { SecurityDefinerViewPopover } from './SecurityDefinerViewPopover'
 import { ViewEntityAutofixSecurityModal } from './ViewEntityAutofixSecurityModal'
 import { RefreshButton } from '@/components/grid/components/header/RefreshButton'
 import { useTableIndexAdvisor } from '@/components/grid/context/TableIndexAdvisorContext'
-import { getEntityLintDetails } from '@/components/interfaces/TableGridEditor/TableEntity.utils'
+import {
+  getEntityLintDetails,
+  getTablePoliciesUrl,
+} from '@/components/interfaces/TableGridEditor/TableEntity.utils'
 import { ButtonTooltip } from '@/components/ui/ButtonTooltip'
 import { useDatabasePoliciesQuery } from '@/data/database-policies/database-policies-query'
 import { useIsTableRealtimeEnabled } from '@/data/database-publications/database-publications-query'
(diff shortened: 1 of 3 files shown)


charislam (approved):
Thanks @oniani1!

PR #45866, opened 2026-05-13
fix: search on the partners/integrations page was still using misc db

imor:
The `INTEGRATIONS_MARKETPLACE_DB` was not included in the client bundle (because it doesn't start with `NEXT_PUBLIC_`) which caused the client side search functions on the https://supabase.com/partners/integrations page to keep using the older misc db for search.

Steps to reproduce:

* Go to the https://supabase.com/partners/integrations page.
* Search for `aikido`.
* Click the only tile found for this search term.

Notice that it takes you to `https://supabase.com/partners/integrations/aikido-security` which 404s. This is because that slug (`aikido-security`) is coming from the old misc db. The correct slug in the new db is `aikido`.

Other fixes:

* Corrected the tsv column name from `tsv` to `listing_tsv`.
* Fixed search debouncing.


joshenlim (approved):
Verified changes on preview!

fsansalvadore (approved):

PR #45867, opened 2026-05-13
chore: remove shadcn suffix for `Input`, `Textarea`, `Alert` and `Collapsible`

djhi:
## Problem

Now that we migrated old components to their new shadcn alternatives, we don't need the `_Shadcn_` suffix anymore.

## Solution

Remove it

image


joshenlim (approved):
Approving to unblock from FE side - changes LGTM, skimmed through studio file changes 🙏

kemaldotearth (approved):
To unblock and unlock.

pamelachia (approved):

staaldraad (approved):

PR #45880, opened 2026-05-13
Support optionally remove custom domain addon when deleting custom domain after activated

joshenlim:
## Context

Related BE PR: https://github.com/supabase/platform/pull/32693

Add support to remove the custom domain add-on when deleting a custom domain after the custom domain is activated.

image

We previously had this behaviour by default to address some customer feedback RE confusion that they were still being charged for custom domain add-on despite deleting the custom domain, but not removing the add-on.

However this was a bit of confusing UX (RE deleting the add-on implicitly), so this makes the deleting of the custom domain add-on an explicit action instead.

## To test

- [ ] Set up custom domain on a project
- [ ] Trying deleting the custom domain after activating _without_ removing the add-on
- [ ] Trying deleting the custom domain after activating _with_ removing the add-on

### DIFF
--- a/apps/studio/components/interfaces/Settings/Addons/Addons.tsx
+++ b/apps/studio/components/interfaces/Settings/Addons/Addons.tsx
@@ -3,7 +3,6 @@ import { useFlag, useParams } from 'common'
 import { Lock } from 'lucide-react'
 import { useTheme } from 'next-themes'
 import Image from 'next/image'
-import Link from 'next/link'
 import {
   Alert_Shadcn_,
   AlertDescription_Shadcn_,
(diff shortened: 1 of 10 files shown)


apps/studio/components/interfaces/Settings/Addons/Addons.tsx
```diff
@@ -268,15 +267,13 @@ export const Addons = () => {
                   <p className="m-0 text-foreground-light text-sm">
                     Reserve a dedicated IPv4 address for your project.
                   </p>
-                  <Link
+                  <InlineLink
+                    className="text-foreground-light"
                     href={`${DOCS_URL}/guides/platform/ipv4-address`}
-                    target="_blank"
```
jordienr, line 273: should we keep the target="_blank" and rel="noreferrer"?

jordienr (approved):

apps/studio/components/interfaces/Settings/Addons/Addons.tsx
joshenlim: inline link innately [handles](<https://github.com/supabase/supabase/blob/master/apps/studio/components/ui/InlineLink.tsx#L27>) that 🙂 if its an external URL, it'll do \_blank and noreferrer

PR #45889, opened 2026-05-13
fix(studio): improve memory and swap units in DB reports

jordienr:
Fixes [DEBUG-94](https://linear.app/supabase/issue/DEBUG-94)

## Summary

- **Min MB units**: Added \`formatBytesMinMB\` helper that always formats byte values in at least MB. Applied to RAM and swap y-axis tick formatters, tooltips, and chart headers.
- **Swap chart scale**: Every Supabase compute instance is provisioned with **1 GB of swap regardless of size** (per [docs](https://supabase.com/docs/guides/troubleshooting/memory-and-swap-usage-explained-aPNgm0)), so the swap y-axis always shows at least 0 to 1 GB. Low swap usage no longer fills the full chart height. Removed the show/hide limit toggle since the limit is implicit in the y-axis scale.
- **Swap units in header/tooltip**: Fixed all three formatters (initial header, hover-sync header, tooltip) to use \`formatBytesMinMB\` for swap so the value always shows a unit like "2.00 MB".
- **Reference-line defensive fix**: Added a \`customValue\` fallback in \`useAttributeQueries\` so future reference-line attributes that use \`customValue\` instead of \`value\` are not silently overwritten to 0.

## Test plan

- [ ] Open DB Reports and verify RAM/Swap y-axis labels show MB (or GB for large values), never KB or bytes
- [ ] Hover a bar in the swap chart and verify tooltip and header show the same value with a unit (e.g. "2.00 MB")
- [ ] With low or zero swap usage, the chart bars are flat and the y-axis goes 0 to 1 GB
- [ ] CPU chart limit toggle still renders at 100% (regression check on the \`useAttributeQueries\` fallback change)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


alaister (approved):
LGTM!

One thing I did notice is the bars still have height when it's 0mb swap:

Image

